PR #1725, opened 2022-05-16
Add postgresac integration tests to CI postgres unit tests

rhybrillou:
## Description

While expanding the test coverage on scoped access control for directly scoped resources, it was noticed that the tests being added were not run by the Postgres CI test targets

## Checklist

## Testing Performed

Validate that some tests from the secrets datastore layer were run in the postgres unit tests CI target

### DIFF
--- a/central/networkgraph/flow/datastore/internal/store/testcommon/flow.go
+++ b/central/networkgraph/flow/datastore/internal/store/testcommon/flow.go
@@ -174,8 +174,8 @@ func (suite *FlowStoreTestSuite) TestRemoveAllMatching() {
 	t2 := time.Now()
 	if features.PostgresDatastore.Enabled() {
 		// Round the timestamps to the microsecond
-		t1.Truncate(1000)
-		t2.Truncate(1000)
+		t1 = time.Date(t1.Year(), t1.Month(), t1.Day(), t1.Hour(), t1.Minute(), t1.Second(), (t1.Nanosecond()/1000)*1000, t1.Location())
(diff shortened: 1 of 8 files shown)


mtodor:
@rhybrillou This looks great! That was a missing link to enable SAC tests. The only thing that we should do, before we merge this PR, is to merge (or better rebase) the master branch into this branch. This branch is a bit behind the master branch and we are missing a lot of tests that already exist in the master.

central/networkgraph/flow/datastore/internal/store/testcommon/flow.go
```diff
@@ -172,6 +172,11 @@ func (suite *FlowStoreTestSuite) TestStore() {
 func (suite *FlowStoreTestSuite) TestRemoveAllMatching() {
 	t1 := time.Now().Add(-5 * time.Minute)
 	t2 := time.Now()
+	if features.PostgresDatastore.Enabled() {
+		// Round the timestamps to the microsecond
+		t1 = time.Date(t1.Year(), t1.Month(), t1.Day(), t1.Hour(), t1.Minute(), t1.Second(), (t1.Nanosecond()/1000)*1000, t1.Location())
```
mtodor, line 177: nitpick: I'm not sure if we can use `t1 = t1.Truncate()` here. https://pkg.go.dev/time#Time.Truncate

mtodor (approved):
Should we also fix `MustCreatePgSearchHelper(mappings.OptionsMap)` in search implementations. I know that it's not directly related to this PR, but since we are already fixing that in the test, maybe we should fix it for implementations too.

janisz (approved):

PR #1728, opened 2022-05-16
ROX-10203: Add a graphQL resolver that returns cluster vulns only

keyallis:
## Description

Add deprecation notices to replaced sub resolvers and queries. Add cluster specific vulnerability resolver.


central/graphql/resolvers/cluster_vulnerabilities.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/utils"
+)
+
+func init() {
+	schema := getBuilder()
+	utils.Must(
+		schema.AddType("ClusterVulnerability",
+			append(commonVulnerabilitySubResolvers,
+				"vulnerabilityType: String!",
```
md2119, line 21: Just a note that we can enforce non-nil only because the default enum string value is `UNKNOWN_VALUE`.

central/graphql/resolvers/cluster_vulnerabilities.go
```diff
@@ -0,0 +1,90 @@
...
+				"vulnerabilityTypes: [String!]!",
+			)),
+		schema.AddQuery("clusterVulnerability(id: ID): ClusterVulnerability"),
+		schema.AddQuery("clusterVulnerabilities(query: String, scopeQuery: String, pagination: Pagination): [ClusterVulnerability!]!"),
+		schema.AddQuery("clusterVulnerabilityCount(query: String): Int!"),
+	)
+}
+
+// ClusterVulnerabilityResolver represents the supported API on image vulnerabilities
+//  NOTE: This list is and should remain alphabetically ordered
```
md2119, line 31: 🙏

md2119 (approved):
LGTM.

PR #1730, opened 2022-05-16
ROX-10933: Sensor integration test

fredrb:
## Description

Draft version of sensor integration tests.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


fredrb:
Secrets in this PR are for test purposes only. No need to worry.

tools/local-sensor/main.go
```diff
@@ -0,0 +1,158 @@
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		endTime := time.Now()
+		allMessages := fakeCentral.GetAllMessages()
+		dumpMessages(allMessages, startTime, endTime, outfile)
+		os.Exit(0)
+	}()
+}
```
vikin91, line 86: How about something context-based? (if needed, one could add a timeout to the context as well)

```suggestion
func registerHostKillSignals(startTime time.Time, fakeCentral *centralDebug.FakeService, outfile string) {
	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
	defer stop()
	for {
		select {
		case <-ctx.Done():
			endTime := time.Now()
			allMessages := fakeCentral.GetAllMessages()
			dumpMessages(allMessages, startTime, endTime, outfile)
			os.Exit(0)
		}
	}
}
```

vikin91 (approved):
Looks great for start!

I would add also cmd flags instead of positional args, but we can do it later as well.

tools/local-sensor/main.go
vikin91: Sorry, the suggestion requires to start `registerHostKillSignals` in a goroutine - forgot to add this.

sensor/debugger/central/communicate_server.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// BlockReceive will read from the connection stream if the connection was already established.
+// The caller might get blocked because it waits until connection is available.
+func (s *FakeService) BlockReceive() (*central.MsgFromSensor, error) {
+	s.ConnectionStarted.Wait()
+	return s.stream.Recv()
+}
+
+// BlockSend will send from the connection stream if the connection was already established.
```
Maddosaurus, line 71: Question: Is it *"will send __TO__ the connection stream"* or does `BlockSend` really send messages it takes from the connection stream and sends it somewhere?

fredrb:
@vikin91 Thanks for the tip with the `flag` package! It made the cli much better. The only problem is that sensor also register some flags. So the help output is cluttered with a bunch of other parameter:

```
go run ./tools/local-sensor -h
Usage of /var/folders/m3/lpt8py696s9g021qj6jtnntw0000gn/T/go-build3403826662/b001/exe/local-sensor:
  -alsologtostderr
    	log to standard error as well as files
  -duration string
    	duration that the scenario should run (leave it empty to run it without timeout)
  -kubeconfig string
    	Paths to a kubeconfig. Only required if out-of-cluster.
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -output string
    	output all messages received to file (default "results.json")
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -v value
    	log level for V logs
  -verbose
    	prints all messages to stdout as well as to the output file
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
```

Other than the two parameters already introduced, there's a new `-verbose` flag, which is used to display all sensor messages in stdout other than the file output. 

Let me know what you think.

tools/local-sensor/main.go
```diff
@@ -0,0 +1,160 @@
...
+		fmt.Println("  local-sensor <minutes> <output file path>")
+		log.Fatalf("Incorrect number of arguments, expected 2 but found: %d", len(os.Args) - 1)
+	}
+
+	i, err := strconv.Atoi(os.Args[1])
+	if err != nil {
+		log.Fatalf("First parameter must be a valid integer")
+	}
+
+	return time.Duration(i) * time.Minute, path.Clean(os.Args[2])
```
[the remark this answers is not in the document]
fredrb: Done!

tools/local-sensor/main.go
fredrb: I don't understand why looping on a non-blocking `select` would be better. The suggestion you propose keeps executing the block under `default` while the os signal is not sent.

I can't see why we wouldn't want to block the goroutine here while this context is not done, like:
```go
func registerHostKillSignals(startTime time.Time, fakeCentral *centralDebug.FakeService, outfile string) {
        ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
        defer stop()
        <- ctx.Done()
        endTime := time.Now()
        allMessages := fakeCentral.GetAllMessages()
        dumpMessages(allMessages, startTime, endTime, outfile)
        os.Exit(0)

}

```

tools/local-sensor/main.go
```diff
@@ -0,0 +1,158 @@
...
+func registerHostKillSignals(startTime time.Time, fakeCentral *centralDebug.FakeService, outfile string) {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		endTime := time.Now()
+		allMessages := fakeCentral.GetAllMessages()
+		dumpMessages(allMessages, startTime, endTime, outfile)
+		os.Exit(0)
+	}()
```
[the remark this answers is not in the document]
fredrb: I'm implementing your suggestion below. But I'm failing to see the leakage here. This is only executed once, and the goroutine will be blocked until the signal is received. If no signal is received then the main goroutine will simply exit and this one gets killed.

sensor/debugger/central/communicate_server.go
fredrb: It was a typo, it sends a message to the stream. This is the fake central implementation file. Meaning that if you call `BlockSend`, sensor will receive a message.

tools/local-sensor/main.go
vikin91: Yeah, after trying that locally, I think you are right - the benefit is minuscule and to leverage it, we would need to replace `context.Background()` with something that expires after a timeout.

vikin91:
> @vikin91 Thanks for the tip with the `flag` package! It made the cli much better. The only problem is that sensor also register some flags. So the help output is cluttered with a bunch of other parameter:
> 
> ```
> go run ./tools/local-sensor -h
> Usage of /var/folders/m3/lpt8py696s9g021qj6jtnntw0000gn/T/go-build3403826662/b001/exe/local-sensor:
>   -alsologtostderr
>     	log to standard error as well as files
>   -duration string
>     	duration that the scenario should run (leave it empty to run it without timeout)
>   -kubeconfig string
>     	Paths to a kubeconfig. Only required if out-of-cluster.
>   -log_backtrace_at value
>     	when logging hits line file:N, emit a stack trace
>   -log_dir string
>     	If non-empty, write log files in this directory
>   -logtostderr
>     	log to standard error instead of files
>   -output string
>     	output all messages received to file (default "results.json")
>   -stderrthreshold value
>     	logs at or above this threshold go to stderr
>   -v value
>     	log level for V logs
>   -verbose
>     	prints all messages to stdout as well as to the output file
>   -vmodule value
>     	comma-separated list of pattern=N settings for file-filtered logging
> ```
> 
> Other than the two parameters already introduced, there's a new `-verbose` flag, which is used to display all sensor messages in stdout other than the file output.
> 
> Let me know what you think.

I will need some of these flags, otherwise we get lost quickly. This is what I have on my branch:

```go
// your params as flags
duration := flag.String("d", "60m", "observe duration")
centralOutputFile := flag.String("central-out", "central-out.json", "file to store the events that would be sent to central")
// mine added
recordTrace := flag.Bool("record", false, "whether to record a trace with k8s events")
traceOutFile := flag.String("record-out", "k8s-trace.jsonl", "a file where recorded trace would be stored")
replyTrace := flag.Bool("reply", false, "whether to reply recorded a trace with k8s events")
traceInFile := flag.String("reply-in", "k8s-trace.jsonl", "a file where recorded trace would be read from")

```

vikin91:
> > How could we separate them? Is there a way to have a subcommand or divide the flags with -- separator maybe?
> 
> What do you mean? Separate them in the `help` command?
> 
> I don't see a problem in adding a bunch of new flags but keeping sensor's flag there as well. The easiest way to add some separation might be to prefix all our commands with something like `local-` or `test-`.

I am okay with mixing and prefixing them.

I was thinking about the `--` as used in commands that run other commands.

Example from bats tests with expect:

```
run expect -f "tests/roxctl/bats-tests/local/expect/flavor-interactive.expect.tcl" -- "$roxctl_bin" development_build "$out_dir" "docker.io/stackrox"
```
Here `-f` applies to the `run` command, whereas all flags after `--` are passed to the `expect` script.

fredrb:
@vikin91 would be ok if I merge this as is now? You can rename the parameters as you wish in your PR. I'd prefer to wait and see how this evolves and maybe even wrap in a new `cobra` CLI altogether.

PR #1705, opened 2022-05-13
Use logger instead of print in qa

janisz:
## Description

This PR make use of logger in all tests.
It also prevent use of print functions.
This is follow up of work done in #1387 

## Splitted into following PRs

- [x] #1734 
- [ ] #1736 
- [x] #1737 
- [ ] https://github.com/stackrox/stackrox/pull/1760

## Testing Performed

CI

PR #1736, opened 2022-05-17
Do not print stack trace in qa test

janisz:
## Description

Prevent errors print stacktrace and use logger instead.

This is part of https://github.com/stackrox/stackrox/pull/1705

## Testing Performed

CI

### DIFF
--- a/qa-tests-backend/codenarc-rules.groovy
+++ b/qa-tests-backend/codenarc-rules.groovy
@@ -301,10 +301,10 @@ ruleset {
     // LoggerWithWrongModifiers
     LoggingSwallowsStacktrace 
     // MultipleLoggers
-    // PrintStackTrace
+    PrintStackTrace
     // Println
-    // SystemErrPrint
-    // SystemOutPrint
+    SystemErrPrint
+    SystemOutPrint
     
     // rulesets/naming.xml
     AbstractClassName 
--- a/qa-tests-backend/src/test/groovy/ComplianceTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ComplianceTest.groovy
@@ -149,8 +149,8 @@ class ComplianceTest extends BaseSpecification {
                 new Control(
                         "HIPAA_164:310_d",
                         ["Runtime support is enabled (or collector service is running) for cluster "
-                                + DEFAULT_CLUSTER_NAME
-                                + ". Network visualization for active network connections is possible."],
+                                 + DEFAULT_CLUSTER_NAME
+                                 + ". Network visualization for active network connections is possible."],
                         ComplianceState.COMPLIANCE_STATE_SUCCESS).setType(Control.ControlType.CLUSTER),
                 new Control(
                         "HIPAA_164:310_d",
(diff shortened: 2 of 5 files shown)


qa-tests-backend/src/test/groovy/ComplianceTest.groovy
```diff
@@ -394,105 +394,101 @@ class ComplianceTest extends BaseSpecification {
 
         then:
         "parse and verify export file"
-        try {
```
janisz, line 397: I don't understand what was tests in this test it's evergreen. All exceptions are catched

qa-tests-backend/codenarc-rules.groovy
```diff
@@ -301,10 +301,10 @@ ruleset {
     // LoggerWithWrongModifiers
     LoggingSwallowsStacktrace 
     // MultipleLoggers
-    // PrintStackTrace
+    PrintStackTrace
     // Println
-    // SystemErrPrint
-    // SystemOutPrint
+     SystemErrPrint
+     SystemOutPrint
```
gavin-stackrox, line 307:
```suggestion
    SystemErrPrint
    SystemOutPrint
```

qa-tests-backend/src/test/groovy/ComplianceTest.groovy
gavin-stackrox: Is there any history on the try/catch lines to suggest why it was ignoring failures? Turning it back on seems like a good idea but if the test was a WIP then it might be more headache than it is worth.

qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
```diff
@@ -268,7 +268,6 @@ class ProcessBaselinesTest extends BaseSpecification {
                 break
              }
          }
-        System.out.println("numAlertsAfterResolve .. " + numAlertsAfterResolve)
```
gavin-stackrox, line 271: log.debug?

gavin-stackrox (changes requested):

qa-tests-backend/codenarc-rules.groovy
janisz: Fixed

gavin-stackrox (changes requested):

qa-tests-backend/src/test/groovy/ComplianceTest.groovy
janisz: Unfortunately it's the only usage of this function :(
I think we can keep this test

janisz: @gavin-stackrox I reverted this change and will create a separate PR for that. Now only change println

qa-tests-backend/src/test/groovy/ComplianceTest.groovy
```diff
@@ -491,7 +491,7 @@ class ComplianceTest extends BaseSpecification {
             }
             println "Verified ${verifiedRows} out of ${rowNumber} total rows"
         } catch (Exception e) {
-            println e.printStackTrace()
+            log.error("Exception", e)
```
porridge, line 494:
```suggestion
            log.error("Exception verifying export file", e)
```

porridge (approved):

gavin-stackrox (approved):

PR #1741, opened 2022-05-17
E2E: Fail tests fast on nightly builds

janisz:
## Description

During nightly tests we often seen a pattern of multiple test failures. Sometimes there is no point in running other tests after failure as cluster might be in dirty state. This change stops tests after the first failure on nightly tag to allow us to focus on the most problematic tests and prioritize fixing them.

See: https://docs.gradle.org/7.4.2/userguide/java_testing.html#sec:test_execution

## Testing Performed

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1508,9 +1508,12 @@ commands:
             else
               export CLUSTER=K8S
             fi
-            if [[ "${CIRCLE_BRANCH}" == "master" || -n "${CIRCLE_TAG}" ]]; then
-              echo "On master, running all QA tests..."
+            if [[ "$CIRCLE_TAG" =~ .*-nightly-.* ]]; then
+              echo "On nightly tag, running all QA tests but failing fast..."
               make -C qa-tests-backend test FAIL_FAST=true || touch FAIL
+            elif [[ "${CIRCLE_BRANCH}" == "master" || -n "${CIRCLE_TAG}" ]]; then
+              echo "On master, running all QA tests..."
+              make -C qa-tests-backend test || touch FAIL
             elif .circleci/pr_has_label.sh ci-all-qa-tests; then
               echo "ci-all-qa-tests label was specified, so running all QA tests..."
               make -C qa-tests-backend test || touch FAIL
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1510,7 +1510,7 @@ commands:
             fi
             if [[ "${CIRCLE_BRANCH}" == "master" || -n "${CIRCLE_TAG}" ]]; then
               echo "On master, running all QA tests..."
-              make -C qa-tests-backend test || touch FAIL
+              make -C qa-tests-backend test FAIL_FAST=true || touch FAIL
```
msugakov, line 1513: I think we shouldn't do this for RC and release builds.

qa-tests-backend/Makefile
```diff
@@ -2,6 +2,12 @@
 all: style test
 
 GRADLE := ./gradlew
+fail_fast = ""
```
msugakov, line 5: A few nitpicks.
1. Why do you need `fail_fast`?
2. Why not reuse `GRADLE` variables in commands below: `GRADLE_TEST := $(GRADLE) test --fail-fast`?
3. Instead of overriding `GRADLE_TEST` command, I suggest to leave it as `$(GRADLE) test` at all times, but introduce `GRADLE_TEST_ARGS` that is set to `--fail-fast` on flag and empty string without it.

msugakov (changes requested):

msugakov:
I would be very grateful if you could also link background information to PRs. Such as Slack links. In this case https://srox.slack.com/archives/C0321S70YK1/p1652450732107979

It is extremely valuable for future generations of ACS developers to record not just the fact that the change was made, but why it was made and what problem it was supposed to solve.

.circleci/config.yml
janisz: Fixed

qa-tests-backend/Makefile
janisz: Fixed

janisz:
@msugakov addressed all issues.

gavin-stackrox (approved):
LGTM

msugakov (approved):

PR #1744, opened 2022-05-17
sensor/central communication: remove license check

fredrb:
## Description

As per internal slack thread: https://srox.slack.com/archives/CELUQKESC/p1652713038114379

We are no longer checking for license status on initial sensor sync. 

## Checklist

## Testing Performed
- CI might be enough

### DIFF
--- a/sensor/common/sensor/sensor.go
+++ b/sensor/common/sensor/sensor.go
@@ -318,10 +318,8 @@ func (s *Sensor) pollMetadata() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	// Metadata result doesn't matter, as long as central is reachable
-	if _, err := s.centralRestClient.GetMetadata(ctx); err != nil {
-		return errors.Wrap(err, "polling metadata")
-	}
-	return nil
+	_, err := s.centralRestClient.GetMetadata(ctx)
+	return err
 }
 
 func (s *Sensor) communicationWithCentral(centralReachable *concurrency.Flag) {
--- a/sensor/common/sensor/sensor.go
+++ b/sensor/common/sensor/sensor.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cenkalti/backoff/v3"
 	"github.com/pkg/errors"
-	v1 "github.com/stackrox/rox/generated/api/v1"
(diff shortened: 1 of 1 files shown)


vikin91 (approved):
LGTM. Would be cool to mark this in the changelog

fredrb:
@vikin91 I wouldn't add it, but I don't know for sure when should we add things to the changelog.

```
Entries in this file should be limited to:

    Any changes that introduce a deprecation in functionality, OR
    Obscure side-effects that are not obviously apparent based on the JIRA associated with the changes. Please avoid adding duplicate information across this changelog and JIRA/doc input pages.
```

This doesn't seem to cover all the cases because we also mention new features in the changelog. Since this is clean up after a field deprecation, I don't think we should add anything.

sensor/common/sensor/sensor.go
```diff
@@ -318,13 +317,9 @@ func (s *Sensor) waitUntilCentralIsReady() {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	md, err := s.centralRestClient.GetMetadata(ctx)
-
-	if err != nil {
-		return err
-	}
-	if md.GetLicenseStatus() != v1.Metadata_VALID {
-		return errors.Errorf("central license status is not VALID but %v", md.GetLicenseStatus())
+	// Metadata result doesn't matter, as long as central is reachable
```
porridge, line 320:
```suggestion
	// Metadata result doesn't matter, as long as central is reachable.
```

sensor/common/sensor/sensor.go
```diff
@@ -318,13 +317,9 @@ func (s *Sensor) waitUntilCentralIsReady() {
...
+	if _, err := s.centralRestClient.GetMetadata(ctx); err != nil {
+		return errors.Wrap(err, "polling metadata")
```
porridge, line 322: I'm not sure the `if` and `Wrap` are adding any value here, "polling metadata" is also not likely to help someone who does not already know what sensor is doing. Perhaps just `return err`?

porridge (approved):
LGTM with a couple of nitpicks. And I agree this is not a changelog-worthy change.

sensor/common/sensor/sensor.go
fredrb: Done

PR #1685, opened 2022-05-12
Replace beforeEach in Violations integration test

pedrottimark:
### Goal

* For each test, decide which helper function to call for actual or mock requests instead of letting `beforeEach` decide.
* Group tests by subject instead of by shared setup.

Find `beforeEach` in cypress/integration: from 6 results in 5 files to 5 results in 4 files.

### Changed files

The scope expanded somewhat to factor out functions and move dashboard tests.

1. Edit constants/apiEndpoints.js
    * Simplify regular expressions because `intercept` uses `minimatch`

2. Update fixtures/alerts
    * Delete alertById.json
    * Delete alertsCount.json
    * Add alert0.json
    * Add deployment0.json
    * Add deploymentWithEmptyContainerConfig.json

3. Edit helpers/main.js
    * Factor functions out from tests

4. Add helpers/violations.js
    * Export functions with sandwich of `intercept` preceding and `wait` following visit or action
    * Replace redundant alertsCount fixture with count computed from alerts fixture

5. Edit integration/dashboard.test.js
    * Move dashboard tests from general to avoid confusion during rewrite in PatternFly

6. Edit integration/general.test.js
    * Call `visitViolations` helper function
    * Delete test for violations side panel that makes request which has status 404 Not Found
    * Factor out `visitViolationsWithAlerts` helper function

7. Edit integration/violations/violations.test.js
    * Delete local `api` overrides because superseded by changes to apiEndpoints.js
    * Delete `should('be.visible')` unless the assertion is specifically visible instead of exists
    * Delete `should('exist')` unless the assertion contrasts with `should('not.exist')`
    * Replace complicated mocking with actual request in test to sort table

### Residue

1. Did not yet replace longer `ViolationsPageSelectors` with shorter `selectors` because it messed up association of tests in changed files.
2. Search to understand whether there is any difference in meaning between `'contain'` and `'have.text'` in `should` method and also how they compare to `:contains("whatever")` pseudo-selector.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Edited integration tests

## Testing Performed

Ran tests one at a time in local deployment

PR #1746, opened 2022-05-17
Simplify integration tests for violations

pedrottimark:
## Description

Follow up on constructive comments:
* https://github.com/stackrox/stackrox/pull/1685#discussion_r873787873 Delete redundant `should('exist')` assertions but keep in test which contrasts exist with not exist.
* https://github.com/stackrox/stackrox/pull/1685#discussion_r873878301 Combine tests for table columns and use inline selectors for ordinary table elements.
* Also replace `ViolationPageSelectors` with `selectors` to follow pattern in most other test files.

### Residue

* `PoliciesPageSelectors.policyDetailsPanel.detailsSection` selector is in classic but not yet in PatternFly Policies constants.
* What are pro and con for reuse of policy details to be visible in violation test versus to encapsulate in ViolationsPage.js which might import from PoliciesPage.js constants file?

## Checklist

## Testing Performed

1. `yarn lint` in ui
2. `yarn start` in ui
3. `yarn cypress-open` in ui/apps/platform
    * integration/violations/violations.test.js

### DIFF
--- a/ui/apps/platform/cypress/integration/violations/violations.test.js
+++ b/ui/apps/platform/cypress/integration/violations/violations.test.js
@@ -1,4 +1,4 @@
-import { selectors as ViolationsPageSelectors } from '../../constants/ViolationsPage';
+import { selectors } from '../../constants/ViolationsPage';
 import { selectors as PoliciesPageSelectors } from '../../constants/PoliciesPage';
 import withAuth from '../../helpers/basicAuth';
 import {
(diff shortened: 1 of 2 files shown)


ui/apps/platform/cypress/integration/violations/violations.test.js
```diff
@@ -21,22 +21,38 @@ describe('Violations page', () => {
...
-        cy.get(ViolationsPageSelectors.table.column.lifecycle);
-        cy.get(`${ViolationsPageSelectors.firstTableRow} td[data-label="Lifecycle"]`).should(
+        cy.get('th[scope="col"]:contains("Policy")');
+        cy.get('th[scope="col"]:contains("Entity")');
+        cy.get('th[scope="col"]:contains("Type")');
+        cy.get('th[scope="col"]:contains("Enforced")');
+        cy.get('th[scope="col"]:contains("Severity")');
+        cy.get('th[scope="col"]:contains("Categories")');
+        cy.get('th[scope="col"]:contains("Lifecycle")');
+        cy.get('th[scope="col"]:contains("Time")');
```
vjwilson, line 45: nit: This doesn't validate the order of the columns.
(Both plus and minus: product desire to change presentation order of columns won't break tests, but tests also can't validate that change or any other regression.)

Overall, fine to leave as-is here, but worth deciding as a team if we want to move toward that tighter testing style.

vjwilson (approved):
Just one idea for the team

ui/apps/platform/cypress/integration/violations/violations.test.js
pedrottimark: Indeed, a good point to discuss and decide.

pedrottimark: Here is rough draft of a pattern to easily change order of columns without need to edit index:

```js
[
    'Policy',
    'Entity',
    'Type',
    // and so on
].forEach((columnText, index) => {
    cy.get(`th[scope="col"]:nth(${index}):contains("${columnText}")`);
});
```

PR #1249, opened 2022-04-11
Deactivate PSPs

mtesseract:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [stackrox/openshift-docs](https://github.com/stackrox/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #1747, opened 2022-05-17
ROX-10034: Deactivate PodSecurityPolicies by default

mtesseract:
## Description

This PR is built on top of https://github.com/stackrox/stackrox/pull/1249 and toggles the defaults for generating deprecated PodSecurityPolicies effectively disabling them by default while still supporting explicit generation of them.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
+++ b/pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
@@ -20,7 +20,7 @@ tests:
     .deployments["central"] | assertThat(. != null)
     .services["central"] | assertThat(. != null)
 
-- name: "central with deprecated PodSecurityPolicies disabled"
+- name: "central with deprecated PodSecurityPolicies enabled"
   values:
     system:
       enablePodSecurityPolicies: true
--- a/pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
+++ b/pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
@@ -12,9 +12,6 @@ values:
 tests:
 - name: "central with default settings"
   expect: |
-    .podsecuritypolicys["stackrox-central"] | assertThat(. != null)
-    .rolebindings["stackrox-central-psp"] | assertThat(. != null)
-    .clusterroles["stackrox-central-psp"] | assertThat(. != null)
     .serviceaccounts["central"] | assertThat(. != null)
     .secrets["central-htpasswd"].stringData.htpasswd | assertThat(length != 0)
     .configmaps["central-config"].data.["central-config.yaml"] | assertThat(length != 0)
@@ -26,11 +23,11 @@ tests:
 - name: "central with deprecated PodSecurityPolicies disabled"
   values:
     system:
-      enablePodSecurityPolicies: false
+      enablePodSecurityPolicies: true
   expect: |
-    .podsecuritypolicys["stackrox-central"] | assertThat(. == null)
-    .rolebindings["stackrox-central-psp"] | assertThat(. == null)
-    .clusterroles["stackrox-central-psp"] | assertThat(. == null)
+    .podsecuritypolicys["stackrox-central"] | assertThat(. != null)
+    .rolebindings["stackrox-central-psp"] | assertThat(. != null)
+    .clusterroles["stackrox-central-psp"] | assertThat(. != null)
 
 - name: "central with OpenShift 3 and enabled SCCs"
   server:
--- a/pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
+++ b/pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
@@ -1,0 +1,19 @@
+tests:
+  - name: "Detect support for PodSecurityPolicies"
+    values:
+      meta:
+        apiServer:
+          overrideAPIResources: ["policy/v1beta"]
+      system:
+        enablePodSecurityPolicies: null
+    expect: |
+      .podsecuritypolicys["stackrox-sensor"] | assertThat(. != null)
+  - name: "Detect absent support for PodSecurityPolicies"
+    values:
+      meta:
+        apiServer:
+          overrideAPIResources: []
+      system:
+        enablePodSecurityPolicies: null
+    expect: |
+      .podsecuritypolicys["stackrox-sensor"] | assertThat(. == null)
--- a/pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
+++ b/pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
@@ -3,7 +3,8 @@ tests:
     values:
       meta:
         apiServer:
-          overrideAPIResources: ["policy/v1beta"]
+          overrideAPIResources:
+            - "policy/v1beta1"
       system:
         enablePodSecurityPolicies: null
     expect: |
--- a/roxctl/central/generate/generate.go
+++ b/roxctl/central/generate/generate.go
@@ -203,6 +203,15 @@ func createBundle(logger logger.Logger, config renderer.Config) (*zip.Wrapper, e
 func OutputZip(logger logger.Logger, config renderer.Config) error {
 	logger.InfofLn("Generating deployment bundle...")
 
+	if config.EnablePodSecurityPolicies {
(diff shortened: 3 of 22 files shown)


janisz:
@mtesseract Is @stackrox/merlin a proper reviewer of this change? Personally I have no context about this change

pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
```diff
@@ -26,11 +23,11 @@ tests:
 - name: "central with deprecated PodSecurityPolicies disabled"
```
SimonBaeumer, line 23: The test description is wrong right?

roxctl/scanner/generate/generate.go
```diff
@@ -109,7 +109,7 @@ func Command(cliEnvironment environment.Environment) *cobra.Command {
 		fmt.Sprintf(
 			"Generate deployment files supporting the given Istio version. Valid versions: %s",
 			strings.Join(istioutils.ListKnownIstioVersions(), ", ")))
-	c.PersistentFlags().BoolVar(&scannerGenerateCmd.enablePodSecurityPolicies, "enable-pod-security-policies", true, "Create PodSecurityPolicy resources (for pre-v1.25 Kubernetes)")
+	c.PersistentFlags().BoolVar(&scannerGenerateCmd.enablePodSecurityPolicies, "enable-pod-security-policies", false, "Create PodSecurityPolicy resources (for pre-v1.25 Kubernetes)")
```
SimonBaeumer, line 112: If we disable the PSP by default for all new roxctl generate execution do you think it could cause problems?
I.e. deploying to a cluster with enabled PSPs might fail?

SimonBaeumer (changes requested):
LGTM from a code perspective.
Can you write in the PR description why PSPs are disabled by default?

To be honest I forgot why we disable them by default immediately as kube 1.25 is not released yet and unsure about the impact it might has on roxctl based installations.

pkg/helm/charts/tests/centralservices/testdata/helmtest/central.test.yaml
mtesseract: thanks

roxctl/scanner/generate/generate.go
mtesseract: Discussed -- decided to keep PSPs by default for backwards compatibility and emit info messages.

roxctl/central/generate/generate.go
```diff
@@ -203,6 +203,15 @@ func createBundle(logger logger.Logger, config renderer.Config) (*zip.Wrapper, e
 func OutputZip(logger logger.Logger, config renderer.Config) error {
 	logger.InfofLn("Generating deployment bundle...")
 
+	if config.EnablePodSecurityPolicies {
+		logger.InfofLn("Central deployment bundle includes PodSecurityPolicies (PSPs). This is incompatible with Kubernetes >= v1.25.")
+		logger.InfofLn("Use --enable-pod-security-policies=false for disabling PodSecurityPolicies.")
+		logger.InfofLn("For the time being PodSecurityPolicies remain enabled by default in deployment bundles and need to be disabled explicitly for Kubernetes >= v1.25.")
```
SimonBaeumer, line 209: I think this message duplicates the previous line? 🤔

roxctl/central/generate/generate.go
```diff
@@ -203,6 +203,15 @@ func createBundle(logger logger.Logger, config renderer.Config) (*zip.Wrapper, e
 func OutputZip(logger logger.Logger, config renderer.Config) error {
 	logger.InfofLn("Generating deployment bundle...")
 
+	if config.EnablePodSecurityPolicies {
+		logger.InfofLn("Central deployment bundle includes PodSecurityPolicies (PSPs). This is incompatible with Kubernetes >= v1.25.")
```
SimonBaeumer, line 207: If the message excludes the deployment name it can be extracted into a function which can be called for all components and reducing duplicated code in all three places.

```suggestion
		logger.InfofLn("Deployment bundle includes PodSecurityPolicies (PSPs). This is incompatible with Kubernetes >= v1.25.")
```

roxctl/central/generate/generate.go
```diff
@@ -203,6 +203,15 @@ func createBundle(logger logger.Logger, config renderer.Config) (*zip.Wrapper, e
...
+		logger.InfofLn("Use --enable-pod-security-policies=false for disabling PodSecurityPolicies.")
```
SimonBaeumer, line 208: I am not a grammar expert, but using this form sounds cleaner to me:

```suggestion
		logger.InfofLn("Use --enable-pod-security-policies=false to disable PodSecurityPolicies.")
```

SimonBaeumer (changes requested):
Only two small changes

roxctl/central/generate/generate.go
mtesseract: tbh, I don't see this as a duplicate. l208 as a (intended!) very short explanation of what is required to be done in order to achieve a specific goal. l209 on the other hand gives 
(1) a more general explanation of the context and
(2) implies the heads-up that this default setting might be up for modification in the future.

If you have a proposal for a different wording, we can of course discuss this. Or ask Gaurauv.

SimonBaeumer: Sounds good!

roxctl/central/generate/generate.go
mtesseract: I think you are right.

roxctl/central/generate/generate.go
mtesseract: done

roxctl/central/generate/generate.go
mtesseract: done

mtesseract:
I have decided to implement auto-sensing in this PR, since otherwise master would be in a somewhat inconsistent state: deployment bundles have PSPs enabled by default and Helm charts have PSPs disabled by default.
Now the Helm charts use auto-sensing and should just work.

pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
```diff
@@ -0,0 +1,20 @@
+tests:
+  - name: "Detect support for PodSecurityPolicies"
+    values:
+      meta:
+        apiServer:
+          overrideAPIResources:
```
SimonBaeumer, line 6: Interesting, never saw this, would like to learn a bit about it. 
How does this work? Can you share a documentation link?

roxctl/central/generate/generate.go
```diff
@@ -203,6 +203,11 @@ func createBundle(logger logger.Logger, config renderer.Config) (*zip.Wrapper, e
 func OutputZip(logger logger.Logger, config renderer.Config) error {
 	logger.InfofLn("Generating deployment bundle...")
 
+	if config.EnablePodSecurityPolicies {
```
SimonBaeumer, line 206: The logic of the if-case is still duplicated.
Moving this if-case to a function `printPSPInfoLog(config.EnablePodSecurityPolicies)` and call it from all other occurrences of this clause.

roxctl/scanner/generate/generate.go
```diff
@@ -59,7 +61,13 @@ func (cmd *scannerGenerateCommand) validate() error {
 	return nil
 }
 
-func (cmd *scannerGenerateCommand) generate() error {
+func (cmd *scannerGenerateCommand) generate(logger logger.Logger) error {
+	if cmd.enablePodSecurityPolicies {
```
SimonBaeumer, line 65: Extract into one func

roxctl/sensor/generate/generate.go
```diff
@@ -130,6 +130,12 @@ func (s *sensorGenerateCommand) fullClusterCreation() error {
 	}
 	s.setClusterDefaults(env)
 
+	if s.enablePodSecurityPolicies {
```
SimonBaeumer, line 133: Extract into one func

SimonBaeumer (changes requested):
Auto-Sensing LGTM!

Small notes:
 - Extracting into one func the if-cases are missing
 - Conflict with CHANGELOG

pkg/helm/charts/tests/securedclusterservices/testdata/helmtest/psp-autosense.test.yaml
mtesseract: Malte came up with this. It's an abstraction layer on top of Helm's lookup functionality which is quite convenient to use and valuable in particular for testing purposes, allows mocking (un)availability of APIs.

I don't have a pointer to documentation for this unfortunately.

PR #1384, opened 2022-04-22
ROX-10822: Group resources, deprecating fine-grained ones.

dhaus67:
## Description

Looking at the issues observed within [ROX-8520](https://issues.redhat.com/browse/ROX-8520), our permissions set seems to be too complex to handle.

A possible way to mitigate this is reducing the number of resources by grouping them.
In this PR, the following resource groups have been created:
- `Access`, deprecating `AuthPlugin, AuthProvider, Group, Licenses, Role, User`.
- `Administration`, deprecating `AllComments, ComplianceRunSchedule, Config, DebugLogs, NetworkGraphConfig, ProbeUpload, ScannerBundle, ScannerDefinitions, SensorUpgradeConfig, ServiceIdentity`.
- `Compliance`, deprecating `ComplianceRuns`.
- `DeploymentExtensions`, deprecating `NetworkBaselin, ProcessWhitelist, Risk`.
- `Integration`, deprecating `BackupPlugins, ImageIntegration, Notifier, SignatureIntegration`.

Besides the introduction of the grouped resources, the following has been done:
- Deprecate resources in a backwards-compatible way.
- Split the deprecated resources within the UI's permission sets table, making them non-selectable.

The deprecation has been done in a backwards-compatible way, ensuring existing roles with old resources are still valid throughout the system instead of i.e. migrating the resources to the new grouped ones, as this would mean roles could potentially have more access than previously intended.

_Hint for reviewers: For now, only made changes required for the `APIToken` resource, other resources should follow in follow-up PRs after the required changes have been agreed for the single resource._

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [stackrox/openshift-docs](https://github.com/stackrox/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed
- CI

PR #1756, opened 2022-05-18
ROX-10993: Respect replaced resources within built-in scope checker.

dhaus67:
## Description

We have added grouped resources within #1384 as a means to simplify the number of resources we currently expose. The main motivation behind this was to help the efforts within postgres SAC migration as well as help users simplify the creation of access scope, as the current resource model was confusing at times.

Now, we have added the concept of replacing resources for resources which are being deprecated as a field within each `ResourceMetadata`. We also adopted the `ResourceHelper`, which is most commonly used as a pattern to check for access to the resource within datastore implementations by conditionally creating a disjunction between two scope checkers: the one of the resource and, if given, the replacing resource.

What is still missing is the adoption of `ScopeCheckerCore` implementations to respect replaced resource.

Given the following example, which is a typical pattern within our code base:
```go
hasReadAccessToAPIToken := sac.WithGlobalAccessScopeChecker(context.Background(),
		sac.AllowFixedScopes(
			sac.ResourceScopeKeys(resources.APIToken)))
```
We would expect that the `ScopeChecker` will correctly yield read access to both `APIToken` as well as its replacing resource `Integration`, but that's currently not the case.

Hence, we need to adopt the implementations of the `ScopeCheckerCore` to also take those into account.

Within this PR, the following has been done:
- Create helper function to check if a resource is contained within a permission map with the respective access.
- Adopt code to use the previously created helper function.
- Adopt the `BuiltInScopeChecker` to respect resource replacements.
- Add additional tests.

What has not been addressed within this PR and needs to be done going forward:
- Adopt `AllowFixedScopes` implementation.
- Adopt `ScopeCheckerCoreImpl` implementation.

## Testing Performed
- Unit tests (see CI).

### DIFF
--- a/pkg/auth/permissions/resource.go
+++ b/pkg/auth/permissions/resource.go
@@ -1,5 +1,7 @@
 package permissions
 
+import "github.com/stackrox/rox/generated/storage"
+
 // Resource is a string representation of an exposed set of API endpoints (services).
 type Resource string
 
@@ -43,6 +45,11 @@ func (m ResourceMetadata) GetResource() Resource {
 
 // GetScope returns the resource scope for this metadata object.
 func (m ResourceMetadata) GetScope() ResourceScope {
+	// Replacing resources _may_ have a different ResourceScope than the initial resource.
+	// This way, clients don't have to deal with replacing resources when checking for scope.
+	if m.ReplacingResource != nil && m.ReplacingResource.GetScope() < m.Scope {
+		return m.ReplacingResource.GetScope()
+	}
 	return m.Scope
 }
 
@@ -70,4 +77,20 @@ type ResourceHandle interface {
 func WithLegacyAuthForSAC(md ResourceMetadata, use bool) ResourceMetadata {
 	md.legacyAuthForSAC = &use
 	return md
+}
+
+// CheckResourceForAccess will verify whether the given ResourceMetadata is contained within the map
+// with at least the specified storage.Access.
+// Note: This will take replacing resources into account.
+func CheckResourceForAccess(resource ResourceMetadata, resourceAccessMap map[string]storage.Access,
+	access storage.Access) bool {
+	if resourceAccessMap[string(resource.GetResource())] >= access {
+		return true
+	} else if resource.ReplacingResource != nil &&
+		// Right now, we are not taking multiple replacing resources into account, i.e. Resource A has replacing
+		// resource "Resource B", which also has a replacing resource "Resource C".
+		resourceAccessMap[string(resource.ReplacingResource.GetResource())] >= access {
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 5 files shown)


rhybrillou (commented):
Good job.
Please make sure the test descriptions match the test case data.

pkg/auth/permissions/resource_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"testing"
+
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCheckResourceForAccess(t *testing.T) {
+	testResource := Resource("Test")
+	replacingResource := Resource("Test")
```
rhybrillou, line 12: Would it make sense to have a different resource name here ?

rhybrillou (approved):

pkg/auth/permissions/resource.go
```diff
@@ -43,6 +45,11 @@ func (m ResourceMetadata) GetResource() Resource {
 
 // GetScope returns the resource scope for this metadata object.
 func (m ResourceMetadata) GetScope() ResourceScope {
+	// Replacing resources _may_ have a different ResourceScope than the initial resource.
+	// This way, clients don't have to deal with replacing resources when checking for scope.
+	if m.ReplacingResource != nil && m.ReplacingResource.GetScope() < m.Scope {
```
rukletsov, line 50: I hung here for a while. What this actually does is _broadens_ the scope, i.e., from namespace to cluster if the replacing resource has the broader one. If so, could you please adjust the comment to make this clear?

pkg/auth/permissions/resource.go
```diff
@@ -71,3 +78,19 @@ func WithLegacyAuthForSAC(md ResourceMetadata, use bool) ResourceMetadata {
 	md.legacyAuthForSAC = &use
 	return md
 }
+
+// CheckResourceForAccess will verify whether the given ResourceMetadata is contained within the map
+// with at least the specified storage.Access.
+// Note: This will take replacing resources into account.
+func CheckResourceForAccess(resource ResourceMetadata, resourceAccessMap map[string]storage.Access,
```
rukletsov, line 85: Maybe take `ResourceMetadata` as a receiver and rename to `IsPermittedBy()`?

pkg/auth/permissions/resource.go
dhaus67: I've made the changes within #1834 , let's move the discussion there, if there are any additional comments.

pkg/auth/permissions/resource.go
dhaus67: I've made the changes within #1834 , let's move the discussion there, if there are any additional comments.

PR #1762, opened 2022-05-18
ROX-10081: update development build image flavor from docker to quay

jschnath:
## Description

Simple change in development flavor

## Checklist

## Testing Performed

Ran flavor_test.go

### DIFF
--- a/pkg/images/defaults/flavor.go
+++ b/pkg/images/defaults/flavor.go
@@ -102,13 +102,13 @@ type ImageFlavor struct {
 func DevelopmentBuildImageFlavor() ImageFlavor {
 	v := version.GetAllVersionsDevelopment()
 	return ImageFlavor{
-		MainRegistry:       "quay.io/stackrox-io",
+		MainRegistry:       "quay.io/rhacs-eng",
 		MainImageName:      "main",
 		MainImageTag:       v.MainVersion,
 		CentralDBImageTag:  v.MainVersion,
 		CentralDBImageName: "central-db",
 
-		CollectorRegistry:      "quay.io/stackrox-io",
+		CollectorRegistry:      "quay.io/rhacs-eng",
 		CollectorImageName:     "collector",
 		CollectorImageTag:      v.CollectorVersion + "-latest",
 		CollectorSlimImageName: "collector",
--- a/pkg/images/defaults/flavor.go
+++ b/pkg/images/defaults/flavor.go
@@ -102,13 +102,13 @@ type ImageFlavor struct {
 func DevelopmentBuildImageFlavor() ImageFlavor {
 	v := version.GetAllVersionsDevelopment()
 	return ImageFlavor{
-		MainRegistry:       "quay.io/stackrox",
+		MainRegistry:       "quay.io/stackrox-io",
 		MainImageName:      "main",
 		MainImageTag:       v.MainVersion,
 		CentralDBImageTag:  v.MainVersion,
 		CentralDBImageName: "central-db",
 
-		CollectorRegistry:      "quay.io/stackrox",
+		CollectorRegistry:      "quay.io/stackrox-io",
 		CollectorImageName:     "collector",
 		CollectorImageTag:      v.CollectorVersion + "-latest",
 		CollectorSlimImageName: "collector",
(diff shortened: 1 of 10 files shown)


pkg/images/defaults/flavor.go
```diff
@@ -102,13 +102,13 @@ type ImageFlavor struct {
 func DevelopmentBuildImageFlavor() ImageFlavor {
 	v := version.GetAllVersionsDevelopment()
 	return ImageFlavor{
-		MainRegistry:       "docker.io/stackrox",
+		MainRegistry:       "quay.io/stackrox",
```
msugakov, line 105: There's no such thing as `quay.io/stackrox`. Technically, there _used_ to be `quay.io/stackrox` but that was used for one specific purpose. I don't think we use it any more now.

Please take a look at the stage where we push images.
https://github.com/stackrox/stackrox/blob/4a47e2369065485dbc8089393b2696defc4d9d82/.circleci/config.yml#L981-L985
The function `push_main_image_set` will remind you of a location where development images get pushed.

msugakov (changes requested):
Please don't forget to modify unit tests that assert image repositories.
There's helmtest framework in use that is a bit tricky to figure. However, you may look for manifest files that are mentioned in test failures and just make adjustments there.

pkg/images/defaults/flavor.go
jschnath: I went too fast on copying the links from opensource flavor it seems, whoops. I fixed it now. As for those tests that assert image repositories I'm having trouble finding them. Where are they?

msugakov: > Where are they?

Please look for a directory called `testdata` next to `flavor_test.go`.

Regarding the current change to `quay.io/stackrox-io`: this is the same directory where images with community branding and `opensource` flavor go. Currently, `development_build` flavored images have commercial branding. I think that commercial branding shouldn't show up in `quay.io/stackrox-io`.
What are options?
1. Keep images with development_build flavor and commercial branding in quay repo different from `quay.io/stackrox-io`.
2. Remove images with development_build flavor and commercial branding altogether, remove the flavor, clean up the code that pushed them. Use images with `opensource` flavor and community branding for daily development in the engineering team.

(1) is less change and less chance for mistake - you just need to figure out the correct repo to set in this PR.
(2) is more brave, but the simplification may be beneficial.

vikin91:
I think, we need some more tests here. This is a simple change in code but it modifies behavior that we developed and corrected for ~3 months. A simple search&replace and relying on unit tests only can be dangerous here.

@jschnath could you think of the tests that you can run to validate this? It would be great if you could update the list in the PR description and then we could look together whether all important workflows are covered.

fredrb:
We do have our bats tests (which *correctly* failed in this PR). They cover at least the expectation that if we use `roxctl` to generate installation bundles it should match the images expected for each flavor. But that should be as easy as updating the bats function to expect `quay.io` images for the development flavor.

Looking purely at the technical change, I think changing the content of flavors shouldn't be a big problem. Most of our work was on switching between different flavors and supporting different images layout for each flavor. I *hope* that our tests (mainly deployer_test.go and helmtests) passing cover that this behavior is intact. As far as I remember we also had to change the RHACS image values at the last minute and it was relatively smooth.

msugakov (commented):
The change looks valid to me now, but there are some failures in `local-roxctl-e2e-tests` that need your attention.

jschnath:
Yes, I have been working on getting the bats tests to run for a bit. Most of them pass now, I'll get them all to run again soon.

msugakov (approved):
Before merging please try to deploy images built for your branch with `deploy-local.sh` on some cluster and share what you see in the output from
```
kubectl --namespace stackrox get pod -o yaml | grep image: | sort | uniq
```

jschnath:
Looks good to me
image

jschnath:
Just rebuilt and ran 
`kubectl --namespace stackrox get pod -o yaml | grep image: | sort | uniq`
this is the output

image

msugakov:
Hm. Do you know from which deployment did `stackrox/main:...` come from?

PR #1771, opened 2022-05-18
Dashrews/ROX-10955 racy upsert many

dashrews78:
## Description

There is a potential race issue using copyFrom because we have to delete the conflicting rows first and then re-add them as opposed to using ON CONFLICT as we would in an insert call.  This change adds a lock to UpsertMany to alleviate this issue.  I did a performance run to ensure that with the lock it is still fast than just upserting one at a time.  In the stats below the old run is with a lock on UpsertMany and the new run is processing the updates one at a time.  As you can tell it is still much faster to use copyFrom than to process each insert individually.  

postgres_op_duration Operation=UpdateMany Type=NetworkBaseline (old: (298204.19/4315) 69.11, new (615005.18/3699) 166.26): change: 140.5811%
postgres_op_duration Operation=UpdateMany Type=ProcessIndicator (old: (15900.29/354) 44.92, new (683955.58/284) 2408.29): change: 5261.7640%

## Checklist
- [ ] ~~Documented user facing changes (create PR based on [openshift/openshift-docs]~~(https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

Ran some scale and long running jobs with a few variations to take a look at that performance metrics to ensure there was not too much degradation by adding this lock.

### DIFF
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -98,9 +98,6 @@ type Store interface {
 
 type storeImpl struct {
     db *pgxpool.Pool
-
-    // Lock since copyFrom requires a delete first before being executed we can get in odd states if
-    // multiple processes are trying to work on the same subsets of rows.
     mutex sync.Mutex
 }
 
@@ -401,6 +398,9 @@ func (s *storeImpl) UpsertMany(ctx context.Context, objs []*{{.Type}}) error {
     }
     {{- end }}
 
+    // Lock since copyFrom requires a delete first before being executed.  If multiple processes are updating
+    // same subset of rows, both deletes could occur before the copyFrom resulting in unique constraint
+    // violations
     s.mutex.Lock()
     defer s.mutex.Unlock()
 
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -97,6 +97,10 @@ type Store interface {
 
 type storeImpl struct {
     db *pgxpool.Pool
+
+    // Lock since copyFrom requires a delete first before being executed we can get in odd states if
+    // multiple processes are trying to work on the same subsets of rows.
+    mutex sync.Mutex
 }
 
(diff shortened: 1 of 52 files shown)


tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -105,6 +106,10 @@ type Store interface {
 
 type storeImpl struct {
     db *pgxpool.Pool
+
+    // Lock since copyFrom requires a delete first before being executed we can get in odd states if
+    // multiple processes are trying to work on the same subsets of rows.
```
c-du, line 111: The mutex is available globally. You can move this specific comment to UpsertMany before locking.

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -105,6 +106,10 @@ type Store interface {
 
 type storeImpl struct {
     db *pgxpool.Pool
+
+    // Lock since copyFrom requires a delete first before being executed we can get in odd states if
```
c-du, line 110: Can you please explain what kind of issue did you see? Like the sequence of operations? Does it only happen to UpsertMany conflicts?

dashrews78: well @md2119 saw it.  But in generally if separate threads are updating the same rows we could get into a variety of weird states.  Since the copyFrom does not deal with on conflict it requires a manual delete first before using the copyFrom the flow is:

delete rows
insert rows with copyFrom

if we have multiple threads processing subsets of rows we can process things in a weird order like
delete - thread1
delete - thread2
copyFrom -thread2
copyFrom -thread1
in that example it really doesn't matter what order the copyFrom occurs in, since both threads have deleted, we can get into a situation where we are trying to overwrite with copyFrom and that would hit us with a unique constraint violation.  I saw this was actually happening in the bespoke Image store and wrote ROX-11050 to account for that.

Hopefully all those words clears that up a little bit.

connorgorman (approved):
This makes sense though probably a little coarse from the locking perspective. We'll have to see if this shows up in perf tests

PR #1597, opened 2022-05-06
ROX-9358: Add auth providers, groups, roles, notifiers and config to diagnostic bundle

mtodor:
## Description

Added new files to the diag bundle:
- export auth providers
- export groups (related to auth providers)
- export roles (with permission sets and scopes)
- export notifiers
- export system configurations

Small refactorings:
- added a helper function to add JSON file to zip (updated also functions not related to this task)
- moved export of logs to end of zip generation function to catch also logs during the creation of diag bundle

*Considerations*
- remove `permissions.ResolvedRole` and introduce protobuf struct. This change is quite big, because `permissions.ResolvedRole` is used in many places (> 30 files). I would suggest doing this change in some 

*Open questions*
- should we add an entry to CHANGELOG?
- should we add anything to "openshift-docs" (user-facing changes)?

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- [x] Evaluated and added CHANGELOG entry if required
- ~[ ] Determined and documented upgrade steps~ - not needed
- ~[ ] Documented user facing changes (create PR based on [stackrox/openshift-docs](https://github.com/stackrox/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~ - user facing stays same and there is no need to update this documentation

## Testing Performed

- as admin - go to `/main/system-health` and create Diagnostic Bundle (upper left)
- as a user with only `DebugLogs` permission - go to `/main/system-health` and create Diagnostic Bundle (upper left)
- compare results for newly added files -> *they should be the same*

PR #1778, opened 2022-05-19
Use Role instead of Access when creating diag bundle

janisz:
## Description

New resources types introduced in #1384 are not supported in Postgres SAC. This lead some e2e tests to fail.
This PR uses deprecated resources in diagnostic bundle in order to make code work and not panic.

Refs: https://github.com/stackrox/stackrox/pull/1597/files#r876980951

## Testing Performed

CI


janisz:
Duplicaates: https://github.com/stackrox/stackrox/pull/1780

PR #1781, opened 2022-05-19
Update action text in Access Control to match UX mocks.

dvail:
## Description

Lost follow up "low hanging fruit" item from my first ACS PR. This updates the "Add" wording on Access Control pages to read "Create" instead, as shown in the mocks: https://www.sketch.com/s/9d082d82-3515-40f2-875a-1b322a02bace/p/F106DCE0-CC2F-4816-8567-767F886FB8E5

## TODO
- [ ] Check with Docs team to coordinate changes

## Checklist

## Testing Performed

Updated e2e tests that checked for the existence of the "Add" text and replaced with "Create".

Manual verification that the button on each of the main Access Control tabs (Auth Providers, Roles, Permission Sets, Access Scopes) instead reads **Create <entity_type>**.

Manual verification that the descriptive text on the main (Roles, Perm Sets, Access Scopes) tabs starts with **Create** instead of **Add**.

Manual verification that the first word in the secondary page heading, as well as the last breadcrumb, is **Create** instead of **Add** when visiting the creation page for each entity.

### DIFF
--- a/ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProviders.tsx
+++ b/ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProviders.tsx
@@ -118,6 +118,11 @@ function AuthProviders(): ReactElement {
         history.push(getEntityPath(entityType, entityId, { ...queryObject, action: undefined }));
     }
 
+    function getProviderLabel(): string {
+        const provider = availableProviderTypes.find(({ value }) => value === type) ?? {};
+        return (provider.label as string) ?? 'auth';
+    }
+
     const selectedAuthProvider = authProviders.find(({ id }) => id === entityId);
     const hasAction = Boolean(action);
     const isList = typeof entityId !== 'string' && !hasAction;
@@ -158,7 +163,7 @@ function AuthProviders(): ReactElement {
                                             toggleIndicator={CaretDownIcon}
                                             isPrimary
                                         >
-                                            Add auth provider
+                                            Create auth provider
                                         </DropdownToggle>
                                     }
                                     isOpen={isCreateMenuOpen}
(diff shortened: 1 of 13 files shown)


ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProviders.tsx
```diff
@@ -118,6 +118,11 @@ function AuthProviders(): ReactElement {
         history.push(getEntityPath(entityType, entityId, { ...queryObject, action: undefined }));
     }
 
+    function getProviderLabel(): string {
+        const provider = availableProviderTypes.find(({ value }) => value === type) ?? {};
+        return (provider.label as string) ?? 'auth';
```
pedrottimark, line 123: Out of curiosity, what does TypeScript think about `provider?.label ?? 'auth'`

dvail: It was not a fan. `provider` comes from a value stored in Redux and the selector is still .js, so it has an `any` type.

That change gives an `unsafe return of 'any' type` error. This seemed like a less invasive way to get this value to typecheck, versus defining types for the source of the data.

pedrottimark (approved):
One comment for your consideration.

ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProviders.tsx
pedrottimark: Good call.

Out of scope for this PR, there are 4 occurrences of type parameter for `createStructuredSelector` call.

vjwilson (approved):

PR #1783, opened 2022-05-19
Dashrews/ROX-11004 busted nightly

dashrews78:
## Description

The timing of events in some of these larger environments was such that things were taking longer to work their way through the system.  I have adjusted the timing in the tests and the tests of the osd-gcp runs have passed the last few runs.  So hopefully this will stabilize this tests for the nightly runs as those were the only runs from which it would fail from time to time.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
@@ -234,7 +234,9 @@ class ProcessBaselinesTest extends BaseSpecification {
 
         println "Locked Process Baseline: ${lockProcessBaselines}"
 
+        // Give the reprocessing some time.
         sleep 30000
+
         // check for process baseline violation
         assert waitForViolation(containerName, "Unauthorized Process Execution", 120)
         List<AlertOuterClass.ListAlert> alertList = AlertService.getViolations(AlertServiceOuterClass.ListAlertsRequest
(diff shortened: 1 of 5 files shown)


qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
```diff
@@ -228,12 +228,17 @@ class ProcessBaselinesTest extends BaseSpecification {
...
-        // sleep 5 seconds to allow for propagation to sensor
-        sleep 5000
+        // sleep 30 seconds to allow for propagation to sensor
+        sleep 30000
         orchestrator.execInContainer(deployment, "pwd")
 
+        log.info "Locked Process Baseline: ${lockProcessBaselines}"
+
+        // Give the reprocessing some time.
+        sleep 30000
```
janisz, line 238: Do we need this? In next line we wait so we can increase timeout

janisz (approved):
LGTM but the question is why it gets slower then before previously we have 6s sleep and now 2×30s

qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
dashrews78: @janisz yeah I hear what you are saying and I get it.  I wrote ROX-11143 so that we can pass the nightly and work on it at the same time.  It takes forever to run those jobs and half the time the provision jobs fail.  So right or wrong, my theory was get it passing now and then work on some debug and tracing so I can optimize it in a more leisurely fashion.  Make sense?

PR #1788, opened 2022-05-20
ROX-10995: Rename PostgresDB tables

c-du:
Rename Postgres tables to unify its style.


viswajithiii (approved):
Pls fix the compile errors, but this looks fine to me!

PR #1790, opened 2022-05-20
Jv/update docs content 3.69.2

JoukoVirtanen:
## Description

Docs update 

## Checklist

## Testing Performed

See above

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,5 @@
 	url = https://github.com/stackrox/docs-tools.git
 	branch = main
 	shallow = true
+[submodules "docs/content"]
+	branch = rhacs-docs 3.69.2


.gitmodules
```diff
@@ -7,3 +7,5 @@
 	url = https://github.com/stackrox/docs-tools.git
 	branch = main
 	shallow = true
+[submodules "docs/content"]
+	branch = rhacs-docs 3.69.2
```
vjwilson, line 11: Why was this present to update in the PR for 3.68.2, but no longer present and had to be added back in for this patch, 3.69.2 ?

JoukoVirtanen: Good catch. The docs/content submodule appears twice. I think something went wrong. I may close this pr. Create a new branch for this and start over to make sure this is done correctly.

PR #1793, opened 2022-05-23
ROX-11019: Only run postgres tests on master.

dhaus67:
## Description

`gke-postgres-api-e2e-tests` depend on feature flags. Nightly tests run on tagged builds ( release builds), which do not include features flags, hence the tests are always failing.


gavin-stackrox (approved):
LGTM

gavin-stackrox:
Oooops. I missed it on the review, this will need to be applied to `provision-gke-postgres-api-e2e-tests` as well.

PR #1794, opened 2022-05-23
ROX-10993: Respect replaced resources within allow fixed scope checker.

dhaus67:
## Description

Continuation of a series of PRs to adapt `ScopeCheckerCore` implementations to respect replacing resources so that we can seamlessly deprecate resources now and in the future. For more general context, you can see the description of
#1756 .

This PR targets the `AllowedFixedScopes` implementation, which is usually used throughout our codebase like this:
```go
ctx := sac.WithGlobalAccessScopeChecker(context.Background(),
		sac.AllowFixedScopes(
			sac.ResourceScopeKeys(resources.APIToken)))
``` 

We expect the returned context `ctx` to have access to `resources.APIToken` as well as the replacing resource, `resources.Integration`. This has been achieved within this PR by:
- Adopting the interface `ResourceHandle` to include `GetReplacingResource`.
- Within `ResourceScopeKeys`, ensure we resolve replacing resources and conditionally add their scope keys as well.
- Adopt the `EffectiveAccessScope`, specifically the `getResourceEffectiveAccessScope` to conditionally verify replacing resources.

## Testing Performed
- Unit tests (see CI).
- Additionally running more QA tests (i.e. postgres) to ensure we don't have unexpected issues.

### DIFF
--- a/pkg/sac/allow_fixed_scopes_checker_core.go
+++ b/pkg/sac/allow_fixed_scopes_checker_core.go
@@ -87,7 +87,13 @@ func (c allowFixedScopesCheckerCore) getResourceEffectiveAccessScope(resource pe
 	}
 	_, resourceAllowed := c.topLevelKeys()[ResourceScopeKey(resource.Resource.GetResource())]
 	if !resourceAllowed {
-		return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
+		if resource.Resource.GetReplacingResource() == nil {
+			return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
+		}
+		_, resourceAllowed = c.topLevelKeys()[ResourceScopeKey(*resource.Resource.GetReplacingResource())]
+		if !resourceAllowed {
+			return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
+		}
 	}
 	return c.next().getClusterEffectiveAccessScope()
 }
--- a/pkg/sac/allow_fixed_scopes_checker_core.go
+++ b/pkg/sac/allow_fixed_scopes_checker_core.go
@@ -90,8 +90,8 @@ func (c allowFixedScopesCheckerCore) getResourceEffectiveAccessScope(resource pe
 		if resource.Resource.GetReplacingResource() == nil {
 			return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
 		}
-		_, resourceAllowed = c.topLevelKeys()[ResourceScopeKey(*resource.Resource.GetReplacingResource())]
-		if !resourceAllowed {
+		_, replacingResourceAllowed := c.topLevelKeys()[ResourceScopeKey(*resource.Resource.GetReplacingResource())]
+		if !replacingResourceAllowed {
 			return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
 		}
 	}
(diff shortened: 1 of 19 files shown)


janisz (approved):

pkg/sac/allow_fixed_scopes_checker_core.go
```diff
@@ -87,7 +87,13 @@ func (c allowFixedScopesCheckerCore) getResourceEffectiveAccessScope(resource pe
 	}
 	_, resourceAllowed := c.topLevelKeys()[ResourceScopeKey(resource.Resource.GetResource())]
 	if !resourceAllowed {
-		return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
+		if resource.Resource.GetReplacingResource() == nil {
+			return effectiveaccessscope.DenyAllEffectiveAccessScope(), nil
+		}
+		_, resourceAllowed = c.topLevelKeys()[ResourceScopeKey(*resource.Resource.GetReplacingResource())]
```
rukletsov, line 93: Please don't do variable shadowing, instead say `replacingResourceAllowed`.

rukletsov (approved):
Looks good overall, thanks!

pkg/sac/scope_key.go
```diff
@@ -86,9 +86,14 @@ func (k ResourceScopeKey) String() string {
 
 // ResourceScopeKeys wraps the given resources in a scope key slice.
 func ResourceScopeKeys(resources ...permissions.ResourceHandle) []ScopeKey {
-	keys := make([]ScopeKey, len(resources))
-	for i, resource := range resources {
-		keys[i] = ResourceScopeKey(resource.GetResource())
+	// keys := make([]ScopeKey, len(resources))
+	// TODO(dhaus): Not sure this semantic change will bear issues.
```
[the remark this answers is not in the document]
dhaus67: Added the additional comment. I've checked throughout the code base, `ResourceScopeKeys` is at the moment only used in combination with `AllowFixedScopes`.

pkg/sac/scope_key.go
```diff
@@ -85,10 +85,16 @@ func (k ResourceScopeKey) String() string {
 }
 
 // ResourceScopeKeys wraps the given resources in a scope key slice.
+// Note: The returned scope keys _may_ be greater than the given resources,
+// since replacing resources will be taken into account and conditionally added
+// to the returned scope keys.
 func ResourceScopeKeys(resources ...permissions.ResourceHandle) []ScopeKey {
 	keys := make([]ScopeKey, len(resources))
```
rukletsov, line 92: This creates a slice of length `n`, what you need is capacity `n` because you only append. So `keys := make([]ScopeKey, 0, len(resources))`

pkg/sac/scope_key.go
```diff
@@ -85,10 +85,16 @@ func (k ResourceScopeKey) String() string {
 }
 
 // ResourceScopeKeys wraps the given resources in a scope key slice.
+// Note: The returned scope keys _may_ be greater than the given resources,
+// since replacing resources will be taken into account and conditionally added
+// to the returned scope keys.
```
rukletsov, line 90: I'd also explicitly say that you think that the broadening is fine because of the contexts this function is used in.

rukletsov (changes requested):

rukletsov (approved):

PR #1803, opened 2022-05-23
ROX-10845: Refactor notifier scrubbing

mtodor:
## Description

We have moved logic for scrubbing of notifiers to the datastore and introduced pair interface functions with and without scrubbing.

**NOTE:** for more information, please check the related ticket.

## Checklist
- ~[ ] Evaluated and added CHANGELOG entry if required~ - not relevant
- ~[ ] Determined and documented upgrade steps~ - not relevant
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~ - not relevant

## Testing Performed

1. Login into UI as `admin`
2. Create a `test` PagerDuty notifier
3. Get Bearer token from Browser
```
export TOKEN='<my-bearer-token>'
```
5. Run the `curl` command to get data over REST API
```
curl --request GET --insecure --url https://localhost:8000/v1/notifiers --header "Authorization: Bearer ${TOKEN}" --silent --output -
```
6. Run the `curl` command to get data over GraphQL
```
curl --request POST --insecure --url https://localhost:8000/api/graphql --header "Authorization: Bearer ${TOKEN}" --header 'Content-Type: application/json' --data '{"query":"{notifiers{pagerduty{apiKey}}}"}' --silent --output -
```

### DIFF
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -44,7 +44,6 @@ import (
 	"github.com/stackrox/rox/pkg/prometheusutil"
 	"github.com/stackrox/rox/pkg/sac"
 	"github.com/stackrox/rox/pkg/sac/observe"
-	"github.com/stackrox/rox/pkg/secrets"
 	"github.com/stackrox/rox/pkg/telemetry/data"
 	"github.com/stackrox/rox/pkg/version"
 	"google.golang.org/grpc"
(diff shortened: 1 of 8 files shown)


viswajithiii (approved):
Makes sense, thanks for doing this!

Only suggestion is to update the PR title (and, therefore, commit message) to be more specific about the change being done (something like "Move notifier scrubbing to datastore layer") rather than just saying "refactor".

PR #1532, opened 2022-05-03
Update Network Graph to use URL parameters for core state management

dvail:
## Description

Note: Using the URL for search filters only has been moved to PR https://github.com/stackrox/stackrox/pull/1804

This change builds on #1804 to track additional items in the Network Graph URL state for better usability and easier linking from other areas in the app. The goal is to be able to use the URL as the source of truth for the selected cluster and selected namespaces in addition to search filters.

### Still TODO

- [ ] Better comments and tests
- [ ] See if there is a cleaner way to integrate URL state with Redux/Redux-sagas (using the URL as the source of truth)
- [ ] Come up with cleaner solution to setting cluster + namespaces from global search results
- [ ] Make the sidebar panel state sync with URL state when navigating via the back/forward buttons

### Code Oddities:
- Due to the Network Graph page being so intwined with Redux/sagas, the "source of truth" for search parameters is a little murky. Any time the URL is updated the new search filters are dispatched to Redux and stored there as well. **Is this reasonable for now, or should this page wait until the true next phase of Network Graph to work with URL search parameters?**

## Other Considerations:

1. For a better UX it might make sense for the cluster and namespace parameters to be available in the URL as well, for consistency and the ability to "preset" the network graph state via links. There are a couple of design concerns in doing this:

If we remove the integration with Redux entirely, network graph state will only persist through the browser's history. Users will have to click the back button until they return to the configuration they wanted. Currently the cluster and namespace values persist through all navigation actions, as that state is stored globally, but search filters disappear when you leave the page.

- Would this change be acceptable? i.e. We no longer persist global state for the Network Graph?
- Do we abandon URL parameters entirely for this page, and keep everything in Redux?
- Is there a hybrid URL parameter + Redux state approach that won't be a horrible mess, in code and for UX?

2. If we move the namespace state into the URL, should it be part of the search `s` parameter, or separate? Currently the namespace filter gets mixed in with the rest of the search filters to be used in the `query` part of a REST or GraphQL call. However, until that point in the code, namespaces are part of a separate widget and sent as a separate parameter from the search filters in function calls. Should we consider the fact that namespace is sent to the backend as part of the same query an implementation detail?

e.g. 
```
https://localhost:3000/main/network?ns=stackrox&s[Deployment]=scanner
https://localhost:3000/main/network?ns[0]=stackrox&ns[1]=kube-system&s[Deployment]=scanner
```
vs
```
https://localhost:3000/main/network?s[Namespace]=stackrox&s[Deployment]=scanner
[...]

PR #1804, opened 2022-05-23
Update Network Graph to use URL parameters for search.

dvail:
## Description

This updates the Network Graph page to use URL parameters for search filters to match the functionality and conventions used on other pages. This does not store other Network Graph state in the URL; in order to keep the scope of changes small this has been extracted from the larger WIP PR https://github.com/stackrox/stackrox/pull/1532.

## Checklist

## Testing Performed

Test that applying a search filter correctly filters the graph display and tracks search filters in the URL.
image

Test that applying multiple search filters correctly filters the graphs and tracks entries in the URL:
image

Test that the back button correctly removes one filter at a time:
image

Test that global search correctly applies the search filters to the graph and URL when visiting via the "Filter On" link:
(Note that the selected cluster and namespaces _will not_ change with these searches, and functions identically to `master`. Work to get context aware switching of clusters/namespaces is part of https://github.com/stackrox/stackrox/pull/1532) 
image
image

Test that clicking a namespace or deployment within the graph correctly updates the URL and page state without clearing search parameters.
image

Test that opening the CIDR block side panel does not clear search results:
image

Test the opening the Net Policy Simulator does not clear search results:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Network/Header/NetworkSearch.js
+++ b/ui/apps/platform/src/Containers/Network/Header/NetworkSearch.js
@@ -1,29 +1,69 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 
 import { selectors } from 'reducers';
 import { actions as pageActions } from 'reducers/network/page';
 import { actions as searchActions } from 'reducers/network/search';
+import useURLSearch from 'hooks/useURLSearch';
+import searchOptionsToQuery from 'services/searchOptionsToQuery';
+import { getSearchOptionsForCategory } from 'services/SearchService';
+import { isCompleteSearchFilter } from 'utils/searchUtils';
+import SearchFilterInput from 'Components/SearchFilterInput';
 import {
     ORCHESTRATOR_COMPONENT_KEY,
     orchestratorComponentOption,
 } from 'Containers/Navigation/OrchestratorComponentsToggle';
-import ReduxSearchInput from 'Containers/Search/ReduxSearchInput';
 
 import './NetworkSearch.css';
 
+function searchFilterToSearchEntries(searchFilter) {
+    const entries = [];
+    Object.entries(searchFilter).forEach(([key, value]) => {
+        entries.push({ label: `${key}:`, value: `${key}:`, type: 'categoryOption' });
+        if (value !== '') {
+            const values = Array.isArray(value) ? value : [value];
+            const valueOptions = values.map((v) => ({ label: v, value: v }));
+            entries.push(...valueOptions);
+        }
+    });
+    return entries;
+}
+
+const searchCategory = 'DEPLOYMENTS';
+const searchOptionExclusions = ['Cluster', 'Namespace', 'Namespace ID', 'Orchestrator Component'];
+
 function NetworkSearch({
-    searchOptions,
-    searchModifiers,
     selectedNamespaceFilters,
-    setSearchOptions,
-    setSearchSuggestions,
+    dispatchSearchFilter,
     closeSidePanel,
     isDisabled,
 }) {
+    const [searchOptions, setSearchOptions] = useState([]);
+    const { searchFilter, setSearchFilter } = useURLSearch();
+
+    useEffect(() => {
+        const { request, cancel } = getSearchOptionsForCategory(searchCategory);
+        request
+            .then((options) => {
+                const filteredOptions = options.filter((o) => !searchOptionExclusions.includes(o));
+                setSearchOptions(filteredOptions);
+            })
+            .catch(() => {
+                // A request error will disable the search filter.
+            });
+
+        return cancel;
+    }, [setSearchOptions]);
+
+    // Keep the Redux store in sync with the URL Search Filter
+    useEffect(() => {
+        dispatchSearchFilter(searchFilterToSearchEntries(searchFilter));
+    }, [searchFilter, dispatchSearchFilter]);
+
     function onSearch(options) {
-        if (options.length && !options[options.length - 1].type) {
+        setSearchFilter(options);
+        if (isCompleteSearchFilter(options)) {
             closeSidePanel();
         }
     }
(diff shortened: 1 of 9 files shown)


pedrottimark:
Here is an existing issue, found while testing, which might be outside the scope of this pull request.

If `onSearch` callback in `NetworkSearch` component closes search panel because search filter might exclude the currently selected deployment, then it seems more consistent to remove deployment id from page url instead.

Both master and branch have the following behavior:

1. Visit **Network Graph**
2. Select Namespaces: **stackrox**
3. Add **Deployment:** and **scanner-db** to search filter
4. Click **scanner** deployment in graph: side panel opens and url now has deployment id
5. Delete **scanner-db** from search filter
6. Add **collector** to search filter: side panel closes **but url still has deployment id**

    Branch has the following additional behavior:

7. Go directly to original **Network Graph** in browser history: side panel is closed and search filter is empty (expected)
8. Go directly to last **Network Graph** in browser history: side panel is open for **scanner** because deployment id and search filter has **Deployment:** **collector** (expected according to code but inconsistent with step 6)

ui/apps/platform/src/utils/searchUtils.ts
```diff
@@ -107,6 +107,16 @@ export function searchOptionsToSearchFilter(searchOptions: SearchEntry[]): Searc
...
 
+/**
+ * Determines whether or not a SearchFilter contains a valid value for
+ * all keys. A valid value is either a non-empty string or non-empty array.
+ */
+export function isCompleteSearchFilter(searchFilter: SearchFilter) {
+    return Object.values(searchFilter).every(
+        (o) => Boolean(o) && (!Array.isArray(o) || o.length > 0)
+    );
+}
```
vjwilson, line 118: I think there is similar functionality in the WorkflowState class. 
(Unfortunately, that is written in classic OO style, and has become something of a "god-class", so I'm not sure it's worth trying to reconcile what's in it with helper functions like this one.)

vjwilson (approved):
/lgtm

Just a couple of observations

dvail:
> If onSearch callback in NetworkSearch component closes search panel because search filter might exclude the currently selected deployment, then it seems more consistent to remove deployment id from page url instead.

I agree, and I found that as I was working on #1532 that adding URL parameters to track state and the current URL structure could be a little confusing. With the deployment id as part of the URL, it seems the full URL path (without query string) should point to a definitive resource within the graph and open side panels, etc. that are appropriate. Right now from what I can tell, only a selected deployment is supported with a structure of
`https://staging.demo.stackrox.com/main/network/<deployment_id>`

Since users can also select a full namespace or external entities it might make sense in the future to migrate the base URL structure to something like:
`https://staging.demo.stackrox.com/main/network/deployment/<deployment_id>`
`https://staging.demo.stackrox.com/main/network/namespace/<namespace_id>`
`https://staging.demo.stackrox.com/main/network/external/<external_entity>`
...and possibly more if there are other use cases I missed. That way we could track the "selected" state and render appropriately along side whatever query string parameters are applied.

For this:
> Go directly to last Network Graph in browser history: side panel is open for scanner because deployment id and search filter has Deployment: collector (expected according to code but inconsistent with step 6)

how exactly did you reproduce this issue?

If I go directly to the following URL on my branch: https://localhost:3000/main/network/966f30dd-a451-45bf-9a7e-7a5f505db7f0?s[Deployment]=collector, I get the following view:
image

If I go to the first Network Graph page in my browser's history though, and step through with the forward button to get to the above URL, I get the behavior you describe:
image

This makes sense (somewhat) because the side panel open state isn't really a declarative, derived state. From what I can tell, other state changes may trigger an action that sets the open/closed state on the panel but the panel state and the rest of the page state can get out of sync.

A possible enhancement for this in the future could be to have the panel state tied to a URL scheme like I suggested above, since it seems that is what the intended behavior is. Additionally we could easily tie in the selected panel tab with something like:
`https://staging.demo.stackrox.com/main/network/deployment/<deployment_id>/flows`
`https://staging.demo.stackrox.com/main/network/deployment/<deployment_id>/network-policies`
`https://staging.demo.stackrox.com/main/network/deployment/<deployment_id>/details`

Your suggested fix of clearing the selected deployment from the URL `onSearch` would be a good solution to the first issue, but I'm not sure that it will resolve the second.

ui/apps/platform/src/Containers/Network/Header/NetworkSearch.js
```diff
@@ -1,29 +1,69 @@
...
+    Object.entries(searchFilter).forEach(([key, value]) => {
+        entries.push({ label: `${key}:`, value: `${key}:`, type: 'categoryOption' });
+        if (value !== '') {
+            const values = Array.isArray(value) ? value : [value];
+            const valueOptions = values.map((v) => ({ label: v, value: v }));
+            entries.push(...valueOptions);
+        }
+    });
+    return entries;
+}
```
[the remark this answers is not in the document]
dvail: Hmm good call, and each one has a very slight variation in functionality that prevents a simple one-for-one replacement. I'll make a note to follow up and clean these up as an independent change.

ui/apps/platform/src/utils/searchUtils.ts
pedrottimark: Yeah, maybe whenever attention turns back to workflow, we might factor out some workflow methods to become available to call as pure functions.

pedrottimark:
> Since users can also select a full namespace or external entities it might make sense in the future to migrate the base URL structure to something like

Yes, possible smaller and larger steps toward the goal for page url to become source of truth:

1. Include namespace in search query params. It seems worthwhile to explore cost and benefit of usual format `s[Namespace]=namespaceId` to hide implementation detail whether interaction is via combined or separate widgets (which might even change in the future).

2. Use nouns in url to represent selected item or interaction panel:
    * /main/network/deployments/deploymentId
    * /main/network/namespaces/namespaceId
    * /main/network/entities/TODO needs investigation what is the "id" for external entities
    * /main/network/blocks
    * /main/network/simulator
    * example of possible future: /main/network/policies/policyId

What are pro and con of singular versus plural nouns, which follows example of usual REST API endpoints?

Do any future concepts have **tables** of deployments, namespaces, or entities, corresponding to url without id?

pedrottimark (approved):
Super step forward. Keep up the good work.

dvail:
@vjwilson @pedrottimark , for awareness since you already approved:

I found another issue after some exploration based on Mark's comments above that clears search parameters when the user clicks the CIDR Blocks or Net Policy buttons. Since those buttons push to the history stack we need to manually retain the `search` part of the URL.

The last commit I pushed addresses this and I added two screenshots to the test section on the PR.

ui/apps/platform/src/Containers/Network/SidePanel/useNavigateToEntity.ts
```diff
@@ -9,9 +9,9 @@ function useNavigateToEntity(): NavigateToEntityHook {
     const history = useHistory();
     return function onNavigateToEntityById(entityId: string, type: EntityType): void {
         if (type === nodeTypes.CIDR_BLOCK || type === nodeTypes.EXTERNAL_ENTITIES) {
-            history.push(`/main/network/${entityId}/${type}`);
+            history.push(`/main/network/${entityId}/${type}${history.location.search as string}`);
```
pedrottimark, line 12: Interesting investigation for the future whether this type of url is possible (pardon pun).

PR #1810, opened 2022-05-23
Move kernel-api tests to nightly only

connorgorman:
## Description

Not a lot of value in the kernel tests and all it does is just run collector in kernel module mode. ebpf is now the default so that's more important to check anyways

## Checklist

## Testing Performed

Straightforward


gavin-stackrox (approved):
LGTM!

PR #1816, opened 2022-05-24
ROX-10736: Add PostgresSearchHelper to process baseline searcher.

dhaus67:
## Description

Conditionally add `PostgresSearchHelper` within the `searcher` for `process baseline` objects.

## Testing Performed
- Integration tests will be added in a follow-up PR in the scope of [ROX-10736](https://issues.redhat.com/browse/ROX-10736).

### DIFF
--- a/central/processbaseline/search/searcher_impl.go
+++ b/central/processbaseline/search/searcher_impl.go
@@ -3,6 +3,7 @@ package search
 import (
 	"context"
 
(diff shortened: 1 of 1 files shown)


janisz (approved):

janisz:
Why did you disable all tests? I think we should run them and pg tests too

rhybrillou (approved):

PR #1829, opened 2022-05-24
Fix missing log method

janisz:
## Description

There is no `warnf` method in log. Let's use `warn` and GString.

## Testing Performed
CI


gavin-stackrox (approved):
Thanks! LGTM pending CI.

PR #1830, opened 2022-05-24
ROX-10701:  Generate gorm models with tests

c-du:
## Description

Generate gorm models and test the models.
We are going to use these models to:
1) create tables in central
2) create tables and migrate data from legacy databases in migrator.

## Checklist
- ~[ ] Documented user facing changes (create PR based on [stackrox/openshift-docs]~(https://github.com/stackrox/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

Add a schema_test to verify the table generated through gorm model and the table generated through sql commands are the same.


pkg/postgres/pgtest/conn.go
```diff
@@ -19,3 +19,14 @@ func GetConnectionString(_ *testing.T) string {
 	host := env.GetString("POSTGRES_HOST", "localhost")
 	return fmt.Sprintf("host=%s port=5432 database=%s user=%s password=%s sslmode=disable statement_timeout=600000 pool_min_conns=1 pool_max_conns=90", host, database, user, pass)
 }
+
+// GetGormConnectionString returns a connection string for Gorm DB
+func GetGormConnectionString(_ *testing.T) string {
```
viswajithiii, line 24: Why not use the same one?

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,395 @@
+//go:build sql_integration
```
viswajithiii, line 1: So this test is useful for validation purposes, but just to confirm, you are eventually planning to remove the create table statements altogether, right?

c-du: Yes, I will remove it later.

pkg/postgres/pgtest/conn.go
c-du: pool_min_conns=1 pool_max_conns=90 are the pool connection config, it is not supported by gorm.

c-du:
The test failed because:
"""
Running: /usr/bin/pg_dump --schema-only -d postgres -h localhost -U postgres -p 5432 -t alerts
pg_dump: error: server version: 14.2; pg_dump version: 12.10 (Ubuntu 12.10-1.pgdg20.04+1)
pg_dump: error: aborting because of server version mismatc
"""
I need to look at upgrade pg_dump in testing env.

.circleci/config.yml
```diff
@@ -2285,7 +2285,7 @@ jobs:
           path: go-unit-release-junit-reports
   go-postgres-tests:
     docker:
-      - image: quay.io/rhacs-eng/apollo-ci:0.3.31
+      - image: quay.io/rhacs-eng/apollo-ci:stackrox-test-0.3.36-4-g645f6152d7
```
viswajithiii, line 2288: Update to a tagged image?

viswajithiii (approved):

.circleci/config.yml
c-du: The PR is still in review. We may need to separate the scanner test and stackrox test image. I will update it when ready.

connorgorman (approved):

pkg/postgres/pgtest/conn.go
janisz: It's also not supported by pg 14 I think it's safe to remove it

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+	suite.Run(t, new(SchemaTestSuite))
+}
+
+func (s *SchemaTestSuite) SetupTest() {
+	s.envIsolator = envisolator.NewEnvIsolator(s.T())
+	s.envIsolator.Setenv(features.PostgresDatastore.EnvVar(), "true")
+
+	if !features.PostgresDatastore.Enabled() {
+		s.T().Skip("Skip postgres store tests")
+		s.T().SkipNow()
```
janisz, line 57: > Skip is equivalent to Log followed by SkipNow.

https://pkg.go.dev/testing#T.Skip

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+)
+
+type SchemaTestSuite struct {
+	suite.Suite
+	envIsolator *envisolator.EnvIsolator
+	connConfig  *pgx.ConnConfig
+	pool        *pgxpool.Pool
+	gorm        *gorm.DB
+	ctx         context.Context
+	tmpDir      string
```
janisz, line 44: Unused

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+	s.pool = pool
+	s.tmpDir, err = os.MkdirTemp("", "schema_test")
+	s.Require().NoError(err)
+	source = pgtest.GetGormConnectionString(s.T())
+	fmt.Println("source:", source)
+	s.gorm, err = gorm.Open(postgres.Open(source), &gorm.Config{})
+	s.Require().NoError(err)
+}
+
+func (s *SchemaTestSuite) TearDownTest() {
```
janisz, line 80: If `s.pool` is nill we will get panic.
There is a check in 84 but not here

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+	gorm        *gorm.DB
+	ctx         context.Context
+	tmpDir      string
+}
+
+func TestSchema(t *testing.T) {
+	suite.Run(t, new(SchemaTestSuite))
+}
+
+func (s *SchemaTestSuite) SetupTest() {
```
janisz, line 51: I think this could be SetupSuite

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+				sqlSchema := s.dumpSchema(table)
+				s.Require().Equal(sqlSchema, gormSchema)
+			}
+			// Check if the table name is reversible.
+			// Gorm may have wrong behavior if the table name is not reversible.
+			schemaName := pgutils.NamingStrategy.SchemaName(testCase.name)
+			s.Require().Equal(testCase.name, pgutils.NamingStrategy.TableName(schemaName))
+		})
+	}
+	s.Require().Len(allTestCases, 0)
```
janisz, line 329: With this if you run just single test case the test will fail

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+	tableMap := make(map[string]string, len(tables))
+	for _, tbl := range tables {
+		tableMap[tbl] = s.dumpSchema(tbl)
+	}
+	return tableMap
+}
+
+func (s *SchemaTestSuite) dumpSchema(table string) string {
+	// Dump Postgres schema
+	cmd := exec.Command(`pg_dump`, `--schema-only`,
```
janisz, line 361: I think we should document requirement for `pg_dump` and add it to https://github.com/stackrox/stackrox-env/blob/master/flake.nix#L25

central/postgres/schema/schema_test.go
```diff
@@ -0,0 +1,389 @@
...
+		"-d", s.connConfig.Database,
+		"-h", s.connConfig.Host,
+		"-U", s.connConfig.User,
```
janisz, line 364: Missing password

pkg/postgres/pgtest/conn.go
```diff
@@ -19,3 +19,14 @@ func GetConnectionString(_ *testing.T) string {
...
+	user := os.Getenv("USER")
+	if _, ok := os.LookupEnv("CI"); ok {
+		user = "postgres"
+	}
+	database := env.GetString("POSTGRES_DB", "postgres")
+	host := env.GetString("POSTGRES_HOST", "localhost")
+	return fmt.Sprintf("host=%s port=5432 database=%s user=%s sslmode=disable statement_timeout=600000", host, database, user)
```
janisz, line 31: Missing password

janisz:
https://github.com/stackrox/stackrox/pull/1862

PR #1832, opened 2022-05-25
Set GOCACHE in Makefile

stehessel:
Currently, a Docker volume for the golang cache is mounted into the
build container, however the GOCACHE environment variable is unset. The
cache volume is therefore unused. By setting GOCACHE to the mounted
volume, the total time for a repeated `make image` decreased from 12
minutes to 5 minutes 30 seconds.

In general I think we should refactor the build pipeline to make use of Docker multi stage and layer caching. This way we could also cache the UI and other non-golang parts of the pipeline. But that is a more complex refactoring for another day.

## Checklist

## Testing Performed

Ran `make image` and timed execution.


connorgorman (approved):
:)

dhaus67 (approved):
LGTM 🚀 

As a time comparison on my machine with Rancher Desktop (14 GB RAM & 8 CPUs assigned):
`pre change (second run): time make image: = 9:37.76 total`
`post change (second run): time make image = 3:25.46 total`

PR #1839, opened 2022-05-25
local-sensor: cleanup + fix small issues

fredrb:
## Description

There were two _tiny_ bug in the `local-sensor` that caused it to panic:

1. If you don't provide `-verbose` flag it crashes since the message callback function is `nil`. This adds a noop callback in the constructor. 
2. When connecting to infra clusters, we require a `gcp` provider to be imported in our `main.go` file. 

## Checklist

This change only affects the local sensor. 

## Testing Performed

- Run `local-sensor` without verbose with a `KUBECONFIG` file pointing to a GCP cluster:
```
go run ./tools/local-sensor -output results.json
```


lvalerom (approved):
LGTM

vikin91 (approved):

PR #1844, opened 2022-05-25
Dashrews/ROX-11126 fix flaky network baseline test

dashrews78:
## Description

Updated the test `Verify user get for non-existent baseline` to more explicitly create network communication between 2 deployments.  Additionally added some logging to help us track some of these things when they do flake out and fail.  The focus of this test is on the test at hand.  Also made some changes where all the tests are run together now as that was causing some issues.  A later refactor may want to look at breaking them up individually instead of strung together, but that is a problem for another day, IMO.

The nongroovy test failure is unrelated to this PR as the whole purpose of this PR is a groovy test.  I used the `ci-no-api-tests` to skip the nongroovy tests, but I'm not sure that works currently in openshift ci.
The openshift test failure is ROX-9864 it appears.

## Checklist

## Testing Performed
 The issue at hand is a flaky test so CI tests are sufficient.

### DIFF
--- a/qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
+++ b/qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
@@ -322,7 +322,10 @@ class NetworkBaselineTest extends BaseSpecification {
         // Get the server baseline to simulate a user asking for a baseline prior to
         // observation ending.  This will generate a baseline at the time of request
         // instead of after observation.
-        assert NetworkBaselineService.getNetworkBaseline(userReqBaselineServerDeploymentID)
+        def userReqServerBaseline = NetworkBaselineService.getNetworkBaseline(userReqBaselineServerDeploymentID)
+        // Ensure the baseline is STILL in the observation window
+        def now = System.currentTimeSeconds()
+        assert userReqServerBaseline.getObservationPeriodEnd().getSeconds() > now
 
         // Add a client deployment
         def beforeClientDeploymentCreate = System.currentTimeSeconds()
@@ -346,7 +349,7 @@ class NetworkBaselineTest extends BaseSpecification {
         assert userReqBaselinedClientBaseline
 
         // Grab a fresh copy of the userReqServerBaseline after the client connection has been added.
-        def userReqServerBaseline = NetworkBaselineService.getNetworkBaseline(userReqBaselineServerDeploymentID)
+        userReqServerBaseline = NetworkBaselineService.getNetworkBaseline(userReqBaselineServerDeploymentID)
 
         log.info "Server Baseline: ${userReqServerBaseline}"
         log.info "Client Baseline: ${userReqBaselinedClientBaseline}"
--- a/qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
+++ b/qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
@@ -57,7 +57,8 @@ class NetworkBaselineTest extends BaseSpecification {
                 .setImage(NGINX_IMAGE)
                 .addLabel("app", BASELINED_USER_CLIENT_DEP_NAME)
                 .setCommand(["/bin/sh", "-c",])
-                .setArgs(["for i in \$(seq 1 10); do wget -S http://${USER_DEP_NAME}; sleep 1; done; sleep 1000" as String])
+                .setArgs(["for i in \$(seq 1 10); do wget -S http://${USER_DEP_NAME};" +
+                    "sleep 1; done; sleep 1000" as String])
 
     static final private ANOMALOUS_CLIENT_DEP = createAndRegisterDeployment()
         .setName(ANOMALOUS_CLIENT_DEP_NAME)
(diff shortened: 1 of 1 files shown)


dashrews78:
ROX-11126

dashrews78:
Moving out of draft to make prow life easier

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
```diff
@@ -277,57 +288,61 @@ class NetworkBaselineTest extends BaseSpecification {
...
-
-        log.info "Deployment IDs Server: ${serverDeploymentID}, " +
-                    "Baselined client: ${baselinedClientDeploymentID}"
+        log.info "Deployment IDs Server: ${serverDeploymentID}"
 
+        // Get the server baseline to simulate a user asking for a baseline prior to
+        // observation ending.
         def serverBaseline = NetworkBaselineService.getNetworkBaseline(serverDeploymentID)
         log.info "Requested Baseline: ${serverBaseline}"
         assert serverBaseline
```
msugakov, line 308: [not your code, but]
1. Could you please extend comment about _why_ is this needed: what would go wrong if server baseline is not requested at this point?
2. Since you overwrite `serverBaseline` variable later anyways, I suggest that you don't save anything here and just do

```groovy
assert NetworkBaselineService.getNetworkBaseline(serverDeploymentID).tap { log.info "Requested Baseline: ${it}" }
```

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
```diff
@@ -277,57 +288,61 @@ class NetworkBaselineTest extends BaseSpecification {
...
+        log.info "Client Baseline: ${baselinedClientBaseline}"
 
         then:
         "Validate user requested server baseline"
         // The client->server connection should be baselined since the client as the
         // connection occurred during the observation window.
         validateBaseline(serverBaseline, beforeDeploymentCreate, justAfterDeploymentCreate,
-            [new Tuple2<String, Boolean>(baselinedClientDeploymentID, true)])
-        validateBaseline(baselinedClientBaseline, beforeDeploymentCreate, justAfterDeploymentCreate,
-            [new Tuple2<String, Boolean>(serverDeploymentID, false)]
```
msugakov, line 330: Why not validating client baseline any more?

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
```diff
@@ -277,57 +288,61 @@ class NetworkBaselineTest extends BaseSpecification {
...
+        assert baselinedClientDeploymentID != null
+        log.info "Client deployment: ${baselinedClientDeploymentID}"
+
+        def clientDeployment = DEPLOYMENTS.find { it.name == BASELINED_USER_CLIENT_DEP_NAME }
+
+        assert retryUntilTrue({
+            // Trigger http traffic to tests service deployment to ensure
+            // StackRox would see it and we can later include it in the baseline.
+            orchestrator.execInContainer(clientDeployment,
+                    "curl --insecure --head --request GET --max-time 5 http://${USER_DEP_NAME}")
```
msugakov, line 325: Why do `curl` here if the client calls `wget` anyway? What would go wrong if you remove this line?

msugakov (commented):
Sharing some initial thoughts after glancing at the change.
Please note that I'll be on PTO from Wednesday. Therefore don't treat my review as blocking. Please check with @fredrb as he's looking at the similar https://issues.redhat.com/browse/ROX-11142

fredrb:
FYI I've opened a PR for the other test in the same file: #2459. The symptom there seems to be a bit different -- there are external connections coming to the running pod during the baseline capture period.

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
```diff
@@ -275,58 +286,57 @@ class NetworkBaselineTest extends BaseSpecification {
...
-        def beforeDeploymentCreate = System.currentTimeSeconds()
-        batchCreate([USER_DEP, BASELINED_USER_CLIENT_DEP])
-        def justAfterDeploymentCreate = System.currentTimeSeconds()
+        "Verify user get for non-existent baseline"
+        beforeDeploymentCreate = System.currentTimeSeconds()
+        batchCreate([USER_DEP])
+        justAfterDeploymentCreate = System.currentTimeSeconds()
 
-        def serverDeploymentID = USER_DEP.deploymentUid
+        serverDeploymentID = USER_DEP.deploymentUid
```
fredrb, line 296: Now that both tests are merged, shouldn't we use a different variable name for this deployment? Maybe `userDeploymentID`? When reading the assertions below I was a bit confused because I thought `serverDeploymentID` was still referencing `SERVER_DEP` deployment other than `USER_DEP`.

fredrb (commented):
I wonder also if `USER_DEP` will face the same issue as `SERVER_DEP` receiving external requests on port `9537`.

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
dashrews78: I will address that.  A case of being in a hurry and probably a little lazy.

dashrews78:
> I wonder also if USER_DEP will face the same issue as SERVER_DEP receiving external requests on port 9537.

@fredrb I suspect that it will on these openshift clusters.  After looking at your fix, I suspect it would make sense for this PR to go after yours.

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
```diff
@@ -297,59 +307,58 @@ class NetworkBaselineTest extends BaseSpecification {
...
-        def baselinedClientDeploymentID = BASELINED_USER_CLIENT_DEP.deploymentUid
-        assert baselinedClientDeploymentID != null
+        log.info "Deployment IDs Server: ${userReqBaselineServerDeploymentID}"
 
-        log.info "Deployment IDs Server: ${serverDeploymentID}, " +
-                    "Baselined client: ${baselinedClientDeploymentID}"
+        // Get the server baseline to simulate a user asking for a baseline prior to
+        // observation ending.  This will generate a baseline at the time of request
+        // instead of after observation.
+        assert NetworkBaselineService.getNetworkBaseline(userReqBaselineServerDeploymentID)
```
fredrb, line 325: Here we're just asserting that _something_ returns from `getNetworkBaseline`. Shouldn't we check some property of the baseline? E.g. you mention in the comment above about the observation ending, isn't it worth asserting on it?

dashrews78: yeah that makes sense.

qa-tests-backend/src/test/groovy/NetworkBaselineTest.groovy
dashrews78: updated

fredrb (approved):

PR #1846, opened 2022-05-25
Dashrews/rox-10086 ROX-11144 add postgres restore capability

dashrews78:
## Description

Provides a postgres restore capability that builds upon the backup capability delivered with rox-10085.  This uses pg_restore in order to take a dump generated by pg_dump to restore to a database.  This restores to a separate database and if that is successful, it will then drop the active database and rename the restore database to be the active database.

## Checklist

## Testing Performed

TODO(replace-me)
Updated the posgrest CI tests to run a backup and restore from postgres vs rocksdb.  Additionally performed many iterations of running backups and restores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - Use `/v1/platformcves/suppress` and `/v1/platformcves/unsuppress` to snooze and unsnooze platform (k8s, istio, and openshift) vulnerabilities.
 - /v1/compliance/results was never implemented and will be removed in this release
 - In release 73.0, the /v1/compliance/runresults endpoint will contain a slimmed down version of the ComplianceDomain object. This allows for greater scalability and reduced memory usage.
+- /db/restore will not be updated to support the database change to Postgres.  This API will become unsupported and deprecated at that time.
 
 ## [70.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - Use `/v1/platformcves/suppress` and `/v1/platformcves/unsuppress` to snooze and unsnooze platform (k8s, istio, and openshift) vulnerabilities.
 - /v1/compliance/results was never implemented and will be removed in this release
 - In release 73.0, the /v1/compliance/runresults endpoint will contain a slimmed down version of the ComplianceDomain object. This allows for greater scalability and reduced memory usage.
-- /db/restore will not be updated to support the database change to Postgres.  This API will become unsupported and deprecated at that time.
+- `/db/restore` will not be updated to support the database change to Postgres.  This API will become unsupported and deprecated at that time.
+  - `roxctl` will be the supported mechanism for performing database restores with the change to Postgres
 
 ## [70.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - Use `/v1/platformcves/suppress` and `/v1/platformcves/unsuppress` to snooze and unsnooze platform (k8s, istio, and openshift) vulnerabilities.
 - /v1/compliance/results was never implemented and will be removed in this release
 - In release 73.0, the /v1/compliance/runresults endpoint will contain a slimmed down version of the ComplianceDomain object. This allows for greater scalability and reduced memory usage.
-- `/db/restore` will not be updated to support the database change to Postgres.  This API will become unsupported and deprecated at that time.
-  - `roxctl` will be the supported mechanism for performing database restores with the change to Postgres
+- When the underlying database changes to Postgres the api `/db/restore` will no longer be a supported means for database restores.  At that time using `roxctl` will be the supported mechanism for database restores.
 
 ## [70.0]
 
--- a/central/globaldb/handlers/http_handler.go
+++ b/central/globaldb/handlers/http_handler.go
@@ -57,6 +57,12 @@ func deferredRestart(ctx context.Context) {
 func RestoreDB(boltDB *bolt.DB, rocksDB *rocksdb.RocksDB) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		log.Info("Starting DB restore ...")
+		// This is the old v1 API.  No need to support that for Postgres
+		if features.PostgresDatastore.Enabled() {
+			logAndWriteErrorMsg(w, http.StatusInternalServerError, "api is deprecated and does not support Postgres.")
+			return
+		}
+
 		filename := filepath.Join(os.TempDir(), time.Now().Format(restoreFileFormat))
 
 		f, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0600)
--- a/central/globaldb/postgres.go
+++ b/central/globaldb/postgres.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stackrox/rox/central/globaldb/metrics"
 	"github.com/stackrox/rox/pkg/config"
 	"github.com/stackrox/rox/pkg/retry"
+	"github.com/stackrox/rox/pkg/stringutils"
 	"github.com/stackrox/rox/pkg/sync"
 )
 
@@ -64,21 +67,14 @@ var (
 	postgresDB                 *pgxpool.Pool
 	pgSync                     sync.Once
 
-	postgresQueryTimeout = 10 * time.Second
+	// PostgresQueryTimeout - Postgres query timeout value
+	PostgresQueryTimeout = 10 * time.Second
 )
 
 // GetPostgres returns a global database instance
 func GetPostgres() *pgxpool.Pool {
 	pgSync.Do(func() {
-		centralConfig := config.GetConfig()
-		password, err := os.ReadFile(dbPasswordFile)
-		if err != nil {
-			log.Fatalf("pgsql: could not load password file %q: %v", dbPasswordFile, err)
-			return
-		}
-		source := fmt.Sprintf("%s password=%s", centralConfig.CentralDB.Source, password)
-
-		config, err := pgxpool.ParseConfig(source)
+		_, config, err := GetPostgresConfig()
 		if err != nil {
 			log.Fatalf("Could not parse postgres config: %v", err)
 		}
@@ -104,8 +100,49 @@ func GetPostgres() *pgxpool.Pool {
 	return postgresDB
 }
 
+// GetPostgresConfig - gets the configuration used to connect to Postgres
+func GetPostgresConfig() (map[string]string, *pgxpool.Config, error) {
+	centralConfig := config.GetConfig()
+	password, err := os.ReadFile(dbPasswordFile)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "pgsql: could not load password file %q", dbPasswordFile)
+	}
+	source := fmt.Sprintf("%s password=%s", centralConfig.CentralDB.Source, password)
+
+	config, err := pgxpool.ParseConfig(source)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Could not parse postgres config")
+	}
+
+	sourceMap, err := ParseSource(source)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Could not parse postgres source")
+	}
+
+	return sourceMap, config, nil
+}
+
+// ParseSource - parses source string into a map for easier use
+func ParseSource(source string) (map[string]string, error) {
+	if source == "" {
+		return nil, errors.New("source string is empty")
+	}
+
+	sourceSlice := strings.Fields(source)
+	sourceMap := make(map[string]string)
+	for _, pair := range sourceSlice {
+		// Due to the possibility that the password could potentially have an = we
+		// need to ensure that we get the entire password
+		key, value := stringutils.Split2(pair, "=")
+
+		sourceMap[key] = strings.TrimSpace(value)
+	}
+
+	return sourceMap, nil
+}
+
 func collectPostgresStats(db *pgxpool.Pool) {
-	ctx, cancel := context.WithTimeout(context.Background(), postgresQueryTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), PostgresQueryTimeout)
 	defer cancel()
 	row, err := db.Query(ctx, tableQuery)
 	if err != nil {
--- a/central/globaldb/v2backuprestore/backup/generators/dbs/postgres.go
+++ b/central/globaldb/v2backuprestore/backup/generators/dbs/postgres.go
@@ -2,20 +2,13 @@ package dbs
 
 import (
 	"context"
-	"fmt"
-	"os"
+	"io"
 	"os/exec"
-	"strconv"
-	"strings"
 
 	"github.com/jackc/pgx/v4/pgxpool"
-	"github.com/stackrox/rox/pkg/config"
+	"github.com/stackrox/rox/central/globaldb"
 	"github.com/stackrox/rox/pkg/logging"
-)
-
-const (
-	dumpTmpPath    = "pg_backup"
-	dbPasswordFile = "/run/secrets/stackrox.io/db-password/password"
+	"github.com/stackrox/rox/pkg/postgres/pgadmin"
 )
 
 var (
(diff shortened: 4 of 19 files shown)


central/globaldb/handlers/http_handler.go
```diff
@@ -57,6 +57,13 @@ func deferredRestart(ctx context.Context) {
 func RestoreDB(boltDB *bolt.DB, rocksDB *rocksdb.RocksDB) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		log.Info("Starting DB restore ...")
+		
+		// This is the old v1 API.  No need to support that for Postgres
```
dashrews78, line 61: This is an assumption on my part that we don't need to support this API with Postgres.  With roxctl, it doesn't even look like there is a way to get in here without going through lots of hoops.  Though I may be missing something.

janisz: @connorgorman What do you think? Do we need to support a case when user has old backup and want to restore it to newer version?

central/globaldb/postgres.go
```diff
@@ -104,6 +97,37 @@ func GetPostgres() *pgxpool.Pool {
 	return postgresDB
 }
 
+// GetPostgresConfig - gets the configuration used to connect to Postgres
+func GetPostgresConfig() (map[string]string, *pgxpool.Config, error) {
+	centralConfig := config.GetConfig()
+	password, err := os.ReadFile(dbPasswordFile)
+	if err != nil {
+		log.Fatalf("pgsql: could not load password file %q: %v", dbPasswordFile, err)
```
janisz, line 105: I know it was extracted but this function returns error so instead of fatal we should return an error and exit in the caller

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,237 @@
...
+		log.Errorf("open database: %v", err)
+	})); err != nil {
+		log.Fatalf("Timed out trying to open database: %v", err)
+	}
+
+	return postgresDB
+}
+
+// CheckIfRestoreDBExists - checks to see if a restore database exists
+func CheckIfRestoreDBExists() bool {
```
janisz, line 213: Could you add a test for it?

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,237 @@
...
+	log.Info("CheckIfRestoreDBExists")
+	_, config, err := globaldb.GetPostgresConfig()
+	if err != nil {
+		log.Fatalf("Could not parse postgres config: %v", err)
+		return false
+	}
+	// connect on template1
```
janisz, line 220: what is `template1`?

central/globaldb/handlers/http_handler.go
dashrews78: Based on my limited knowledge I would think something like that would have to go through a migration vs a restore.  But I may not have the whole picture.

central/globaldb/v2backuprestore/restore/postgres.go
dashrews78: template1 is a database that automatically gets started with postgres.  Since I need to rename the restore database to the postgres database name we are using, I had to connect to something.  I did write [ROX-11272](https://issues.redhat.com/browse/ROX-11272) to reserve the postgres database for admin type connections such as these and use a different database for general processing.   I will capture and clarify that with a comment and a TODO.

roxctl/central/db/db.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stackrox/rox/roxctl/common/flags"
 )
 
-// Command controls all of the functions being applied to a sensor
+// Command controls all of the functions being applied to a central
```
c-du, line 13: Thanks! This needs to be "... to a central db"

central/main.go
```diff
@@ -235,6 +236,22 @@ func main() {
...
+	if features.PostgresDatastore.Enabled() {
+		sourceMap, config, err := globaldb.GetPostgresConfig()
+		if err != nil {
+			log.Errorf("Unable to get Postgres DB config: %v", err)
+			return
+		}
+		// Check to see if a restore DB exists, if so use it.
+		if restore.CheckIfRestoreDBExists(config) {
+			// Now flip the restore DB to be the primary DB
+			err := restore.SwitchToRestoredDB(sourceMap, config)
```
c-du, line 248: The restored db needs to go through migration process. If the restored version is earlier than the current version, we need to upgrade it.

central/globaldb/handlers/http_handler.go
dashrews78: well there really isn't a path in there from our code base.  roxctl uses the v2 api.  Even the stuff within `proto/api/v1/db_service.proto` goes through the v2 api via the v2_restorer and such.  The v2 restore implementation seems much more feature rich, as well.  So I figured that api was already on the way out and a major upgrade such as this would make a good time to pull the plug.  

But of course I called it out here as I was not 100% certain of that desire.  I can support it if we need to easy enough, but I think one path for Postgres restore code is better, if possible.

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,237 @@
...
+	cmd := exec.Command("createdb", options...)
+
+	common.SetPostgresCmdEnv(cmd, sourceMap, config)
+
+	return common.ExecutePostgresCmd(cmd)
+}
+
+func runRestore(dumpPath string, sourceMap map[string]string, config *pgxpool.Config) error {
+	// Set the options for pg_dump from the connection config
+	options := []string{
```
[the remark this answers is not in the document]
dashrews78: done.

central/globaldb/handlers/http_handler.go
connorgorman: Yeah I think we can support the v2 api only moving forward and using roxctl as the only supported way to restore

central/globaldb/postgres.go
```diff
@@ -104,6 +98,52 @@ func GetPostgres() *pgxpool.Pool {
...
+	}
+	source := fmt.Sprintf("%s password=%s", centralConfig.CentralDB.Source, password)
+
+	config, err := pgxpool.ParseConfig(source)
+	if err != nil {
+		log.Fatalf("Could not parse postgres config: %v", err)
+		return nil, nil, err
+	}
+
+	sourceMap, err := ParseSource(source)
```
connorgorman, line 117: What fields are we making use of that require us to also parse the source instead of using fields that are within the config?

dashrews78: @connorgorman The SSL mode and cert location.  I chose to set that stuff in an env var and there was no real easy way to get it out of the connection.
See `central/globaldb/v2backuprestore/common/postgres.go`. 
There were equally clumsy ways around it where I could pass the source in a quoted strings, but that made feeding the options to cmd more difficult.  At least it did for me.

central/globaldb/postgres.go
```diff
@@ -104,6 +98,52 @@ func GetPostgres() *pgxpool.Pool {
...
+	if source == "" {
+		return nil, errors.New("source string is empty")
+	}
+
+	sourceSlice := strings.Fields(source)
+	sourceMap := make(map[string]string)
+	for _, pair := range sourceSlice {
+		// Due to the possibility that the password could potentially have an = we
+		// need to ensure that we get the entire password
+		configSetting := strings.SplitN(pair, "=", 2)
```
connorgorman, line 137: We've got a Split2 function in stringutils that can probably just alleviate some of these. Not sure we need to check for empty field values

central/globaldb/v2backuprestore/backup/generators/dbs/postgres.go
```diff
@@ -30,102 +23,41 @@ func NewPostgresBackup(db *pgxpool.Pool) *PostgresBackup {
...
-		"-p",
-		strconv.FormatUint(uint64(config.ConnConfig.Port), 10),
 		"-d",
 		config.ConnConfig.Database,
-		"-Fd",
-		"-f",
-		backupPath,
-		"-j",
-		"5",
+		"-Fc", // Custom format, compressed hopefully supports stdin to restore
```
c-du, line 43: We cannot easily check the data from database dump. Can you create a jira to create/find a tool to convert db dump to human readable format.

central/globaldb/v2backuprestore/common/postgres.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
+
+// ExecutePostgresCmd -- executes a command
+func ExecutePostgresCmd(cmd *exec.Cmd) error {
+	log.Info(cmd)
```
c-du, line 19: We can change it to Debug level.

central/globaldb/v2backuprestore/common/postgres.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+	err = cmd.Wait()
+
+	if exitError, ok := err.(*exec.ExitError); ok {
+		log.Error(exitError)
+		return err
+	}
+
+	log.Info("Exiting execution of command")
```
c-du, line 35: log.Debug

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,222 @@
...
+// SwitchToRestoredDB - switches the restore DB to be the active DB
+func SwitchToRestoredDB(sourceMap map[string]string, config *pgxpool.Config) error {
+	log.Info("Switching to restored database")
+
+	// Connect to different database for admin functions
+	connectPool := adminPool(config)
+
+	// Restore succeeded to the separate DB, so we need to drop the original in order to rename
+	// the new one.
+	err := dropDB(sourceMap, config, config.ConnConfig.Database)
```
c-du, line 158: We should keep original db in case we need to switch back.

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,222 @@
...
+var (
+	log = logging.LoggerForModule()
+
+	postgresOpenRetries        = 10
+	postgresTimeBetweenRetries = 10 * time.Second
+)
+
+// dropDB - drops a database.  This is so we can restore to a new database then flip the name
+// of the restored database to the original for consistency
+func dropDB(sourceMap map[string]string, config *pgxpool.Config, databaseName string) error {
```
c-du, line 39: Can you move postgres databases management like dropDB, createDB, renameDB and some form of admtool under pkg?  And  we need to separate that from the pgxpoo.Config.

central/globaldb/handlers/http_handler.go
dashrews78: How are we handling that with Postgres?  It was essentially deprecated with version 27.0.  But as far as these changes are concerned, we won't support this API for Postgres.  So how would we do the CHANGE.log for that since while the code will be in the next release, it wouldn't impact anyone until Postgres was enabled.  Are we capturing such things elsewhere so we can dump them all in when Postgres goes?  I don't want to confuse the CHANGE.log with things the users won't see or know nothing about.

central/globaldb/v2backuprestore/backup/generators/dbs/postgres.go
dashrews78: Looks like that is as simple as running pg_restore without a database name.

https://stackoverflow.com/questions/21429335/how-do-i-convert-a-binary-pgdump-compressed-to-a-plain-sql-file

That dumps the SQL, I'm not terribly sure that we will get much more readable than that.  So I'm not sure a jira ticket is necessary.

central/globaldb/v2backuprestore/backup/generators/dbs/postgres.go
```diff
@@ -30,102 +23,41 @@ func NewPostgresBackup(db *pgxpool.Pool) *PostgresBackup {
...
+	// Set the stdout of the command to be the output writer.
+	cmd.Stdout = out
 
-	return backupPath, nil
-}
+	common.SetPostgresCmdEnv(cmd, sourceMap, config)
 
-func (bu *PostgresBackup) findScratchPath(ctx context.Context) (string, error) {
-	dbSize, err := bu.getPostgresSize(ctx)
+	err = common.ExecutePostgresCmd(cmd)
```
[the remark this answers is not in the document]
dashrews78: done

central/globaldb/v2backuprestore/restore/postgres.go
dashrews78: I think I would like to defer the work with the admin pool to ROX-11272.  I think that would be cleaner to do then we can define an admin DB and an admin source in the central-db config if we desire in addition to a default admin source to use to create an admin pool.  The rest of the stuff I have moved.

central/globaldb/v2backuprestore/restore/postgres.go
dashrews78: We can do that as part of restore and always keep 2 OR we can deal with that as part of https://issues.redhat.com/browse/ROX-9884 when we have a better plan on how we would want to do rollbacks.  I think I would prefer to do it under that ticket than this one.

central/globaldb/handlers/http_handler.go
connorgorman: I would say in the CHANGELOG that when Postgres is released, the v1 API will become unsupported

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,167 @@
...
+// CheckIfRestoreDBExists - checks to see if a restore database exists
+func CheckIfRestoreDBExists(config *pgxpool.Config) bool {
+	log.Info("CheckIfRestoreDBExists")
+
+	// Connect to different database for admin functions
+	connectPool := adminPool(config)
+
+	existsStmt := "SELECT EXISTS(SELECT 1 FROM pg_catalog.pg_database WHERE datname = $1)"
+
+	row := connectPool.QueryRow(context.Background(), existsStmt, restoreDB)
```
connorgorman, line 156: We should create a timeout here with something reasonable, otherwise this could hang for a long time if there is a network issue

central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -0,0 +1,167 @@
...
+	var exists bool
+	if err := row.Scan(&exists); err != nil {
+		return false
+	}
+
+	// Close the admin connection pool
+	connectPool.Close()
```
connorgorman, line 163: We should defer this from above right? We always want to close this even if we return false?

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return ExecutePostgresCmd(cmd)
+}
+
+// RenameDB - renames a database
+func RenameDB(adminPool *pgxpool.Pool, originalDB, newDB string) error {
+	log.Infof("Renaming database %q to %q", originalDB, newDB)
+	sqlStmt := fmt.Sprintf("ALTER DATABASE %s RENAME TO %s", originalDB, newDB)
+
+	_, err := adminPool.Exec(context.TODO(), sqlStmt)
```
connorgorman, line 64: in the future, what would we change this context to? Does having one with a timeout make sense?

pkg/postgres/pgadmin/postgres_command_utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Run the command
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+	err = cmd.Wait()
+
+	if exitError, ok := err.(*exec.ExitError); ok {
+		log.Error(exitError)
```
connorgorman, line 26: context on this exit error?

central/globaldb/v2backuprestore/restore/postgres.go
dashrews78: ah, yeah I need to get in the habit of using that go feature.

central/globaldb/v2backuprestore/restore/postgres.go
dashrews78: I just used a context.WithTimeout like you suggested for the other one.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - Use `/v1/platformcves/suppress` and `/v1/platformcves/unsuppress` to snooze and unsnooze platform (k8s, istio, and openshift) vulnerabilities.
 - /v1/compliance/results was never implemented and will be removed in this release
 - In release 73.0, the /v1/compliance/runresults endpoint will contain a slimmed down version of the ComplianceDomain object. This allows for greater scalability and reduced memory usage.
+- /db/restore will not be updated to support the database change to Postgres.  This API will become unsupported and deprecated at that time.
```
connorgorman, line 31: Can you add a note that the supported way is to use roxctl?

connorgorman: This reads like restore is deprecated

connorgorman (approved):

dashrews78:
Just nothing that the failures in 
`ci/circleci: gke-postgres-api-e2e-tests `
relate to a collector version change that is being reverted.

PR #1849, opened 2022-05-25
Skip UI make targets

stehessel:
The `touch build` seems to be required to properly skip UI make targets
if no change has been detected. Without it, the `make build` target in
`ui/Makefile` is always executed, regardless of actual changes to source
files.

As a result, this change decreases the execution time of `make image` in
most cases.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


viswajithiii (approved):

pedrottimark (approved):
Thank you for investigating the problem and finding a solution.

PR #1863, opened 2022-05-27
Scan main image with roxctl

janisz:
## Description

We can use our own scanner to scan our images.

## Testing Performed

CI

https://github.com/stackrox/stackrox/runs/6625002213?check_suite_focus=true


.github/workflows/roxctl.yml
```diff
@@ -0,0 +1,33 @@
...
+    - cron: '0 2 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  roxctl:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
```
stehessel, line 14: Is this necessary? It seems like there is no `codeql-action`.

.github/workflows/roxctl.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Image scan
+      env:
+        ROX_API_TOKEN: ${{ secrets.STACKROX_CI_INSTANCE_API_KEY }}
+        ROX_CENTRAL_ADDRESS: ${{ secrets.STACKROX_CI_INSTANCE_CENTRAL_HOST }}
+      run: ./roxctl image scan -e $ROX_CENTRAL_ADDRESS:443 -i quay.io/stackrox-io/main -o table
+    - name: Image check
+      env:
+        ROX_API_TOKEN: ${{ secrets.STACKROX_CI_INSTANCE_API_KEY }}
+        ROX_CENTRAL_ADDRESS: ${{ secrets.STACKROX_CI_INSTANCE_CENTRAL_HOST }}
+      run: ./roxctl image check -e $ROX_CENTRAL_ADDRESS:443 -i quay.io/stackrox-io/main -o junit > check.xml && ./roxctl image check -e $ROX_CENTRAL_ADDRESS:443 -i quay.io/stackrox-io/main -o table
```
stehessel, line 28: Why scan twice? Do we need the table output if we have the test summary?

.github/workflows/roxctl.yml
```diff
@@ -0,0 +1,33 @@
...
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download roxctl
+      run: curl -sSLO https://mirror.openshift.com/pub/rhacs/assets/latest/bin/Linux/roxctl && chmod +x roxctl
+    - name: Image scan
+      env:
+        ROX_API_TOKEN: ${{ secrets.STACKROX_CI_INSTANCE_API_KEY }}
+        ROX_CENTRAL_ADDRESS: ${{ secrets.STACKROX_CI_INSTANCE_CENTRAL_HOST }}
```
stehessel, line 22: Where is the Central hosted?

stehessel (approved):
Looks good other than what I left in comments.

.github/workflows/roxctl.yml
janisz: 🤫  It's a secret

.github/workflows/roxctl.yml
janisz: Good point I'm not sure what will be better
Test summary do not show skipped tests (maybe we need find something different) and table output is OK. Maybe we should just put this table in test summary it may be already in markdown format?

.github/workflows/roxctl.yml
janisz: Oh, no it's not needed. I though we have sarif format support

.github/workflows/roxctl.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: ./roxctl image scan -e $ROX_CENTRAL_ADDRESS:443 -i quay.io/stackrox-io/main -o table
```
gavin-stackrox, line 23: whatabout getting roxctl from the CI central so versions match like: https://github.com/stackrox/rox-ci-image/blob/8825725583407f5cc1bb185c0989154a86712a05/.circleci/config.yml#L167

.github/workflows/roxctl.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  roxctl:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download roxctl
+      run: curl -sSLO https://mirror.openshift.com/pub/rhacs/assets/latest/bin/Linux/roxctl && chmod +x roxctl
```
gavin-stackrox, line 18: whatabout getting roxctl from the CI central so versions match like: https://github.com/stackrox/rox-ci-image/blob/8825725583407f5cc1bb185c0989154a86712a05/.circleci/config.yml#L167

.github/workflows/roxctl.yml
gavin-stackrox, line 28: Will you do this for `quay.io/stackrox-io/roxctl` and `quay.io/stackrox-io/central-db` also?

janisz:
Without sarif support in `roxctl` this PR has no sense

PR #1865, opened 2022-05-27
Remove licensing and telemetry

connorgorman:
## Description

Removing licensing and telemetry as they are no longer relevant potentially. This will need a migration to drop the buckets from Bolt before merging

## Checklist

## Testing Performed

CI


central/license/service/service_impl.go
```diff
@@ -3,53 +3,15 @@ package service
 import (
```
viswajithiii, line 3: Should we put a deprecation notice for these APIs so we can remove them eventually?

central/telemetry/service/service_impl.go
```diff
@@ -5,7 +5,6 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
```
viswajithiii, line 6: Same

viswajithiii (approved):
image

central/license/service/service_impl.go
connorgorman: Yes, will do this

PR #1866, opened 2022-05-27
ROX-11145: Replace SQL schema with Gorm schema

c-du:
## Description
Replace SQL schema with Gorm schema.

## Checklist

## Testing Performed
existing CI tests


dashrews78:
Does Gorm have access to a low level pool?  I'm curious if it would be nice to only have to create a single connection and use that.

viswajithiii (commented):
I'm not a huge fan of having two different handles passed to every single function. I think we should hide this detail from all the functions outside pgutils, and just have a single composite handle we create (something like: a `struct` that includes both DB handles and proxies to the appropriate one)?

c-du:
> Does Gorm have access to a low level pool? I'm curious if it would be nice to only have to create a single connection and use that.
> I thought about it too. But did not find a good way to do it. Th

I talked with Gorman and we can offload the schema management completely to migrator. Good for code aesthetics and cleanness.

PR #1870, opened 2022-05-27
Dashrews/rox-10085 postgres backup capability

dashrews78:
## Description

Added the ability to backup a postgres DB via the existing Backup APIs. These changes creates a Postgres backup by utilizing pg_dump.

Note:  This is take 2 on these changes.  First time I missed hiding a call to globaldb.GetPostgres() behind the feature flag.  That caused all sorts of problems with deployment so those changes had to be reverted.

## Checklist

## Testing Performed

Ran roxctl backups to produce the backups files. Visually looked at a few to see how they looked. This also helped me narrow down some of the options like writing it to a tar in compressed format and then exporting it as a zip. Further testing of the backup will be performed when [ROX-10086](https://issues.redhat.com/browse/ROX-10086) is worked which will restore from a backup.

### DIFF
--- a/scripts/grab-data-from-central.sh
+++ b/scripts/grab-data-from-central.sh
@@ -40,13 +40,16 @@ main() {
 
     roxctl -e "${api_endpoint}" -p "${ROX_PASSWORD}" central backup --output "${dest}"
 
-    if ! [ -x "$(command -v rocksdbdump)" ]; then
-        go install ./tools/rocksdbdump
+    # With Postgres we no longer take RocksDB dumps
+    if [ -z "${ROX_POSTGRES_DATASTORE}" ] || [ "${ROX_POSTGRES_DATASTORE}" = false ]; then
+      if ! [ -x "$(command -v rocksdbdump)" ]; then
+          go install ./tools/rocksdbdump
+      fi
+
+      rocksdbdump -b "${dest}"/*.zip -o "${dest}"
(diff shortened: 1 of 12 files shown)


c-du (approved):

scripts/grab-data-from-central.sh
```diff
@@ -40,13 +40,16 @@ main() {
 
     roxctl -e "${api_endpoint}" -p "${ROX_PASSWORD}" central backup --output "${dest}"
 
-    if ! [ -x "$(command -v rocksdbdump)" ]; then
-        go install ./tools/rocksdbdump
-    fi
+    # With Postgres we no longer take RocksDB dumps
+    if [ -z "${ROX_POSTGRES_DATASTORE}" ] || [ "${ROX_POSTGRES_DATASTORE}" = false ]; then
```
c-du, line 44: Quote "false"

scripts/grab-data-from-central.sh
```diff
@@ -40,13 +40,16 @@ main() {
...
+      if ! [ -x "$(command -v rocksdbdump)" ]; then
```
c-du, line 45: We still have rocksdb even with Postgres DB enabled. For the purpose of debugging, we'd better still collect the remaining rocksdb dump.

c-du (approved):

PR #1871, opened 2022-05-27
Fix race condition tests by setting MAIN_IMAGE_TAG in qa-test portion

connorgorman:
## Description

Race condition on the workspace persisting so work around that 

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1962,7 +1962,7 @@ commands:
             fi
 
             if [[ "${CI_JOBNAME}" == "race-condition-tests" ]]; then
-              MAIN_IMAGE_TAG="$(make --quiet tag)-rcd"
+              export MAIN_IMAGE_TAG="$(make --quiet tag)-rcd"
             fi
 
             deploy_dir="deploy/<< parameters.orchestrator-flavor >>"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1961,6 +1961,10 @@ commands:
               fi
             fi
 
+            if [[ "${CI_JOBNAME}" == "race-condition-tests" ]]; then
+              MAIN_IMAGE_TAG="$(make --quiet tag)-rcd"
```
gavin-stackrox, line 1965: I think you'll need to permafix the tag for other dependencies on `make tag` but if setting MAIN_IMAGE_TAG is enough then maybe not :)
```suggestion
              TAG="$(make --quiet tag)"
              echo "${TAG}-rcd" > CI_TAG
              make --quiet tag
```

gavin-stackrox (approved):
LGTM pending CI

.circleci/config.yml
connorgorman: Let's see cause nothing is "made" in this step and we are just using the premade things to deploy

PR #1875, opened 2022-05-29
Remove `clean-image` from `all-builds` target

stehessel:
Currently the `clean-image` target is called on every invocation of `make image`. This causes a rebuild of the rhel and postgres bundles, which delays `make image` execution. To my understanding, the bundles do not require frequent rebuilding.

In the future we could look to remove the bundles altogether to make the build process more efficient and benefit more from Docker caching layers. However, dropping `clean-image` already speeds up `make image` by ~1-2 minutes on my machine.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -505,13 +505,13 @@ generate-junit-reports: $(GO_JUNIT_REPORT_BIN)
 image: main-image
 
 .PHONY: all-builds
-all-builds: cli main-build clean-image $(MERGED_API_SWAGGER_SPEC) ui-build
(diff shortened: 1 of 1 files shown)


stehessel:
Not sure about this change. Maybe somebody more familiar with the bundle workflow could confirm that his makes sense 😅 .

connorgorman (changes requested):
If we add makefile dependencies to bundle.tar.gz on the proper directories, then it should only rebuild once files have changed underneath those directories?

connorgorman (approved):

PR #1876, opened 2022-05-30
Improve duration of image signature verification tests.

dhaus67:
## Description

Currently, ImageSignatureVerification tests within `race-condition-tests` take up to 47 minutes.
Although reprocessing takes a bit longer within the race build, this is not expected.

The test itself will deploy four deployments with four different image, three images will have a signature associated with them. Four policies will be created to test the alert creation based on the different images and their respective signature.
The exact number of violations will not be tested, but rather whether there are violations or not.

When looking at the test output, the following can be seen:
```
ImageSignatureVerificationTest > Check violations of policy 'Distroless+Tekton' for deployment 'with-signature-verified-by-distroless' STANDARD_OUT
    07:54:07 | INFO  | ImageSignatureVerificationTest | Starting testcase
    08:01:43 | INFO  | Services                  | Failed to trigger Distroless+Tekton after waiting 456 seconds
    08:01:43 | INFO  | ImageSignatureVerificationTest | Ending testcase
```

When no violations are expected, we _still_ wait the full amount of time. This has been changed now to increase the speed, as there are multiple cases where no violations are expected (to be exact: 6). This would lead to an overhead of `6 x 456 seconds = 2736 seconds = 45.7 minutes`. This would explain the long duration of those tests within the `race-condition-tests`, since for `race-condition-tests` the timeout is conditionally increased:
```groovy
    static final private Integer WAIT_FOR_VIOLATION_TIMEOUT =
            isRaceBuild() ? 450 : ((Env.mustGetOrchestratorType() == OrchestratorTypes.OPENSHIFT) ? 100 : 30)
```

It would also explain why in the `gke-api-e2e-tests` / `openshift-4-tests` this overhead wasn't as painfully obvious as it was for race condition tests.

I've added a simple check to use `checkForNoViolations` in case no violations are expected.

## Testing Performed
- ImageSignatureVerificationTests should take significantly less time within `race-condition-tests` (the test should at most take 5-8 mins).

### DIFF
--- a/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
@@ -1,4 +1,6 @@
+import static Services.checkForNoViolations
 import static Services.waitForViolation
+
 import groups.BAT
 import groups.Integration
 import io.stackrox.proto.storage.PolicyOuterClass
@@ -206,7 +208,11 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
     def "Check violations of policy '#policyName' for deployment '#deployment.name'"() {
         expect:
         "Verify deployment has expected violations"
-        assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
+        if (expectViolations) {
+            assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
+        } else {
+            assert checkForNoViolations(deployment.name, policyName)
+        }
 
         where:
         policyName                                 | deployment                   | expectViolations
--- a/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
@@ -9,7 +9,6 @@ import io.stackrox.proto.storage.ScopeOuterClass
 import io.stackrox.proto.storage.SignatureIntegrationOuterClass.CosignPublicKeyVerification
 import io.stackrox.proto.storage.SignatureIntegrationOuterClass.SignatureIntegration
 import objects.Deployment
-import orchestratormanager.OrchestratorTypes
 import org.junit.experimental.categories.Category
 import services.PolicyService
 import services.SignatureIntegrationService
@@ -21,9 +20,6 @@ import util.Env
 // Do not run tests on crio due to an issue with crio trying to pull images referenced by digest from gcr.io.
 @IgnoreIf({ Env.CI_JOBNAME.contains("crio") })
 class ImageSignatureVerificationTest extends BaseSpecification {
-    // https://issues.redhat.com/browse/ROX-6891
-    static final private Integer WAIT_FOR_VIOLATION_TIMEOUT =
-            isRaceBuild() ? 450 : ((Env.mustGetOrchestratorType() == OrchestratorTypes.OPENSHIFT) ? 100 : 30)
 
     static final private String SIGNATURE_TESTING_NAMESPACE = "qa-signature-tests"
 
@@ -209,9 +205,9 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
         expect:
         "Verify deployment has expected violations"
         if (expectViolations) {
-            assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
+            assert waitForViolation(deployment.name, policyName, 15) == expectViolations
         } else {
-            assert checkForNoViolations(deployment.name, policyName)
+            assert checkForNoViolations(deployment.name, policyName, 15)
         }
 
         where:
--- a/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
@@ -205,7 +205,7 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
         expect:
         "Verify deployment has expected violations"
         if (expectViolations) {
-            assert waitForViolation(deployment.name, policyName, 15) == expectViolations
+            assert waitForViolation(deployment.name, policyName, 5)
         } else {
             assert checkForNoViolations(deployment.name, policyName, 15)
         }


dhaus67:
@msugakov / other reviewers: `race-condition-tests` should be running once more, so we should observe the expected result now.

qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
```diff
@@ -206,7 +208,11 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
     def "Check violations of policy '#policyName' for deployment '#deployment.name'"() {
         expect:
         "Verify deployment has expected violations"
-        assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
+        if (expectViolations) {
+            assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
```
janisz, line 212: How about sleeping in setup spec once and here just query for the result without or with minimal timeout?

dhaus67:
Looking at the execution time:
```
ImageSignatureVerificationTest STANDARD_OUT
    08:55:11 START

ImageSignatureVerificationTest STANDARD_OUT
    08:56:44 STOP
```

qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
dhaus67: I was thinking about that, but didn't want to introduce this due to potential flakes when i.e. reprocessing of one particular image takes longer (for whatever reason) than the other.
I'd propose to stay defensive right now and not introduce a potential to flake and potentially change this later on if we feel the need for it.

WDYT?

janisz (approved):

qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
```diff
@@ -206,7 +208,11 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
...
+        } else {
+            assert checkForNoViolations(deployment.name, policyName)
```
msugakov, line 214: Well, the old code waited for
* 450 seconds in race build,
* otherwise 100 seconds on OpenShift,
* otherwise 30 seconds.

The new code would wait for just 5 seconds.
This may give a false confidence of succeeding test while violations actually appear but after 5 seconds.
Maybe a way to speed things up is by decreasing 450 seconds timeout in race tests? How long does it take for violation to appear when `expectViolations==true` in race tests?

msugakov:
Also, related Slack thread https://srox.slack.com/archives/C02TPMGCBQW/p1652961853693739

qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
msugakov: Hm, looking again at the test output https://circleci.com/api/v1.1/project/github/stackrox/stackrox/575517/output/134/0?file=true&allocation-id=6285e4d6387d7e6780308a30-0-build%2F38D39C05

`expectViolations==true` cases find violations... after `0` seconds. Checks for no-violations are using up full timeout.

```
ImageSignatureVerificationTest > Check violations of policy 'Distroless' for deployment 'with-signature-verified-by-tekton' STANDARD_OUT
    07:31:16 | INFO  | ImageSignatureVerificationTest | Starting testcase
    07:31:17 | INFO  | Services                  | violation size is: 1
    07:31:17 | INFO  | Services                  | Distroless triggered after waiting 0 seconds
    07:31:17 | INFO  | ImageSignatureVerificationTest | Ending testcase

ImageSignatureVerificationTest > Check violations of policy 'Distroless' for deployment 'with-signature-unverifiable' STANDARD_OUT
    07:31:17 | INFO  | ImageSignatureVerificationTest | Starting testcase
    07:31:18 | INFO  | Services                  | violation size is: 1
    07:31:18 | INFO  | Services                  | Distroless triggered after waiting 0 seconds
    07:31:18 | INFO  | ImageSignatureVerificationTest | Ending testcase

ImageSignatureVerificationTest > Check violations of policy 'Distroless' for deployment 'without-signature' STANDARD_OUT
    07:31:18 | INFO  | ImageSignatureVerificationTest | Starting testcase
    07:31:18 | INFO  | Services                  | violation size is: 1
    07:31:18 | INFO  | Services                  | Distroless triggered after waiting 0 seconds
    07:31:18 | INFO  | ImageSignatureVerificationTest | Ending testcase

ImageSignatureVerificationTest > Check violations of policy 'Distroless' for deployment 'with-signature-verified-by-distroless' STANDARD_OUT
    07:31:18 | INFO  | ImageSignatureVerificationTest | Starting testcase
    07:38:54 | INFO  | Services                  | Failed to trigger Distroless after waiting 456 seconds
    07:38:54 | INFO  | ImageSignatureVerificationTest | Ending testcase
```

I suggest to tighten up timeouts the following way:
1. Use explicitly small timeout in the call to `waitForViolation()`.
2. Use timeout three times as big in the call to `checkForNoViolations()`.

Both items together give a confidence that when the test that expects violation passes and they appear in time `T` (or less), the succeeding test that expects no violations to appear would also legitimately pass because `3*T` should be more than enough for any violations to appear and fail it.

dhaus67: > Checks for no-violations are using up full timeout.

I'd say this is expected, as the code behind `checkForNoViolations()` will always wait the full specified timeout.

I adjusted the timeout values to now use 15 seconds for both, which will add an overhead of 90 seconds. Have to carefully observe whether this will introduce any unwanted flakes in the future.

qa-tests-backend/src/test/groovy/ImageSignatureVerificationTest.groovy
```diff
@@ -206,7 +204,11 @@ w9e2Azq1OYIh/pbeBMHARDrBaqqmuMR9+BfAaPAYdkNTU6f58M2zBbuL0A==
     def "Check violations of policy '#policyName' for deployment '#deployment.name'"() {
         expect:
         "Verify deployment has expected violations"
-        assert waitForViolation(deployment.name, policyName, WAIT_FOR_VIOLATION_TIMEOUT) == expectViolations
+        if (expectViolations) {
+            assert waitForViolation(deployment.name, policyName, 15) == expectViolations
+        } else {
+            assert checkForNoViolations(deployment.name, policyName, 15)
```
msugakov, line 210: This makes me happier but not 100% happy. As I was suggesting previously, let's have timeout for `checkForNoViolations` N (2,3,...) times bigger than the timeout for `waitForViolation`. This will tell the reader: _Look, we waited **more than enough** for violation to appear but it had not appeared._
```suggestion
            assert waitForViolation(deployment.name, policyName, 5) == expectViolations
        } else {
            assert checkForNoViolations(deployment.name, policyName, 15)
```

PR #1879, opened 2022-05-30
Fred/rox 11155 add resync sensor options

fredrb:
## Description

- Extracts `CreateSensor` parameters into a new options property called `CreateOptions`. 
- Adds a new create option for sensor `resyncPeriod` together with `-resync` flag for `local-sensor`.

Configuring resync period allows us two things:
1. Create tests that use a variable resync period, making them faster if the tests requires to test resync specific behavior. (i.e. it's better to wait 1 second than 1 minute to test resync behavior). 
2. Gracefully shutdown resync without removing the code that controls it. If resync is set to `0`, the informer is created without resync. This enables us to either have a higher resync period or set it to `0` with an env variable without removing the code entirely on the first iteration. 

## Checklist

## Testing Performed

- Run local sensor with `-resync 1s` parameter to get events resynced every minute.

### DIFF
--- a/sensor/kubernetes/listener/listener.go
+++ b/sensor/kubernetes/listener/listener.go
@@ -2,6 +2,7 @@ package listener
 
 import (
 	"context"
+	"time"
 
 	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/pkg/concurrency"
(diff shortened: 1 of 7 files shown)


lvalerom (approved):
LGTM. [nit] would it make sense to add the `client.Interface` (`ClientSet`) to the `CreateOptions` struct?

fredrb:
> would it make sense to add the client.Interface (ClientSet) to the CreateOptions struct?

Sure, added. At first, I thought more of it as an "optional" parameter list, but we're adding the connection factory anyway, so I guess it fits.

PR #1884, opened 2022-05-31
Fred/rox 11023 resync test

fredrb:
## Description

This is on top of #1879.

## Checklist

## Testing Performed

- Running the newly added tests


sensor/tests/resync_test.go
```diff
@@ -0,0 +1,187 @@
...
+				// This function will create the fake k8s event and wait for one second.
+				// This way we guarantee some level of order for the events. Since sensor
+				// processes events in different goroutines, some events take more time to
+				// be processed than others. Therefore, even if we send two different events
+				// at a particular order, e.g. Deployment and role respectively, it might be
+				// that Role finishes first and is sent to Central before deployment gets to
+				// the point of checking for permission levels.
+				// To force what would happen in conditions where deployment is fully processed
+				// before a Role is received, a timer was introduced here as well.
+				// TODO: This should be done by checking the received event rather than arbitrarily waiting.
```
lvalerom, line 139: Should we create an issue for this?

lvalerom (approved):
Looks really good.

sensor/tests/resync_test.go
fredrb: Sure. If we're okay not having this for now and working on it in parallel then I'll create an issue for it.

fredrb: Yea, I tried a couple of times yesterday and wasn't able to get this logic working. I tried it now and works like a charm. Maybe I messed something up in my original implementations. 🤷 
Pushing the new version now. Thanks!

PR #1885, opened 2022-05-31
Dashrews/ROX-11143 optimize timing in flaky process baseline test for nightly tests

dashrews78:
## Description

Looked like a large part of the problem is that in these nightly environments the deployments would sometimes take a while to come up.  The test would continue while the deployment was not necessarily running its containers yet.  So I made an adjustment to wait on a process to show up and be part of the baseline before proceeding.  I've run this successfully all week except for the environments that won't provision.  So I think it is pretty stable.  If not, we can always ignore it and go back to the drawing board.

## Checklist

## Testing Performed

Set up labels to run the tests for the larger environments that the nightly tests use.

### DIFF
--- a/central/detection/lifecycle/manager_impl.go
+++ b/central/detection/lifecycle/manager_impl.go
@@ -183,7 +183,14 @@ func (m *managerImpl) flushIndicatorQueue() {
 	// Map copiedQueue to slice
 	indicatorSlice := make([]*storage.ProcessIndicator, 0, len(copiedQueue))
 	for _, indicator := range copiedQueue {
+		if indicator.ContainerName == "pb-deploymentnginx-violation-resolve" {
+			log.Info("SHREWS -- we have the pb-deploymentnginx-violation-resolve add indicator")
+			log.Infof("SHREWS -- %s", indicator.Signal.Name)
+		}
 		if deleted, _ := m.deletedDeploymentsCache.Get(indicator.GetDeploymentId()).(bool); deleted {
+			if indicator.ContainerName == "pb-deploymentnginx-violation-resolve" {
+				log.Info("SHREWS -- we have the pb-deploymentnginx-violation-resolve but the deployment is deletd already")
+			}
 			continue
 		}
 		indicatorSlice = append(indicatorSlice, indicator)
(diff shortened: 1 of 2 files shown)


gavin-stackrox (approved):
Hi @dashrews78, I rudely removed some of the cluster labels from your PR. CI resources for OSD are maxed out and PR runs tends to exacerbate failures on release CI and nightly runs.
To compensate I approved this PR based on your changes and the successful runs.

central/detection/lifecycle/manager_impl.go
```diff
@@ -240,6 +240,7 @@ func (m *managerImpl) buildMapAndCheckBaseline(indicatorSlice []*storage.Process
 }
 
 func (m *managerImpl) checkAndUpdateBaseline(baselineKey processBaselineKey, indicators []*storage.ProcessIndicator) (bool, error) {
+
```
vikin91, line 243: This empty line could annoy `golanglint-ci`

qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
```diff
@@ -218,24 +218,33 @@ class ProcessBaselinesTest extends BaseSpecification {
...
-        sleep 30000
-        orchestrator.execInContainer(deployment, "pwd")
 
-        log.info "Locked Process Baseline: ${lockProcessBaselines}"
+        // sleep 5 seconds to allow for propagation to sensor
+        sleep 5000
+        orchestrator.execInContainer(deployment, "pwd")
 
-        // Give the reprocessing some time.
-        sleep 30000
```
vikin91, line 238: It looks like we removed two sleeps here. Maybe in face of flakes, it would be better to change one thing at a time? or are you confident about these removals?

vikin91 (commented):
Not an expert in this corner of code, so I left just two comments. It looks good in general, but lets make sure, we do not remove too many sleeps ;)

qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
dashrews78: well I did do it one at a time just within the PR.  Once multiple runs were successful after the first one, I removed the second and had multiple successful runs.  Now of course that doesn't mean once it is merged that we won't find some nightly environment where it fails.  In that event I would just recommend we write a ticket and ignore the test while we are working on it.

PR #1891, opened 2022-05-31
Fix job test by pinning perl version

connorgorman:
## Description

Test using job was failing because perl upgrade was causing job to go to error instead of job going to completed, which is what we are testing

## Checklist

## Testing Performed

CI


gavin-stackrox (approved):
Thanks!

theencee (approved):

PR #1892, opened 2022-05-31
Add image sha to vuln management image details page.

dvail:
## Description

As described in PR title. For https://issues.redhat.com/browse/ROX-8054

## Checklist

## Testing Performed

Test visiting the details page for a specific image in Vuln Management and see that the SHA is present in the UI
image

Test that the SHA is present when visiting an image details page from a different top level entity (e.g. CVEs):
image

Note that a side effect of this is that the new wrapping behavior in the Metadata.js component will cause long lines for other detail entries to wrap to a new line. The previous behavior is that the text would be cut off for very long single words. See "Vector" in the below screenshots. The first is the behavior in this PR, the second is the old behavior on `master`.


vjwilson (approved):
New wrapping behavior seems like the behavior we've wanted to begin with.

PR #1901, opened 2022-06-01
ROX-11166: Create fixtures for SAC indirectly scoped resources.

rhybrillou:
## Description

Generate object stubs for dackbox graph structure.

## Checklist

## Testing Performed

CI run should be enough. Objects will be used later on in tests


janisz:
@rhybrillou How did you generated that? Do you think there is a way to visualize it? It'll be great to see the graph when writing a test?

rhybrillou:
The generated dataset is too complex to be realistically useable in tests.
Will rely on https://github.com/stackrox/stackrox/pull/2115 instead.

PR #1911, opened 2022-06-01
Port singleton stores over to Postgres

connorgorman:
## Description

Create a new singleton store pattern in Postgres and port over the 3 singleton stores from Bolt over

## Checklist

## Testing Performed

CI, plus unit tests, plus manual


viswajithiii (approved):

viswajithiii:
> No, they are just copied from the singleton store, but I guess I could actually just modify the generator now that I think about it (facepalm). Will do that


These are Bolt stores, I don't think it's a big deal tbh

PR #1915, opened 2022-06-01
WIP

clickboo:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


clickboo:
> The current pattern we've been using is to move this logic up to the datastore layer for now instead of modifying the generic template:
> 
> ```
>     RenamePolicyCategory(request *v1.RenamePolicyCategoryRequest) error
>     DeletePolicyCategory(request *v1.DeletePolicyCategoryRequest) error
>     ```
> ```

Too much work for something that is being deprecated and totally redesigned to be brought back. I'd strongly recommend an exception for this.

central/policy/service/service_impl.go
```diff
@@ -742,29 +742,30 @@ func checkIdentityFromMetadata(ctx context.Context, metadata map[string]interfac
...
-				Error: &v1.PolicyError{
-					Error: policyErrors[i].Error(),
-				},
-			})
-		}
+	errDetails := &v1.ExportPoliciesErrorList{}
+	for _, missingIndex := range missingIndices {
+		policyID := request.PolicyIds[missingIndex]
+		errDetails.Errors = append(errDetails.Errors, &v1.ExportPolicyError{
+			PolicyId: request.PolicyIds[missingIndex],
```
connorgorman, line 753: policyID

central/policy/service/service_impl.go
```diff
@@ -742,29 +742,30 @@ func checkIdentityFromMetadata(ctx context.Context, metadata map[string]interfac
...
+			Error: &v1.PolicyError{
+				Error: "not found",
+			},
+		})
+		log.Warnf("A policy error ocurred for id %s: not found", policyID)
```
connorgorman, line 758: occurred

central/policy/store/boltdb/store.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+var (
+	policyBucket = []byte("policies")
+	// As of 66.0, any default policies that were added from 65.0 onwards cannot be deleted.
+	// Any policies added prior to 65.0 can be deleted because the criteria fields are not locked.
+	// Locked policy criteria guarantees that the criteria remains unchanged as is as it was shipped.
+	removedDefaultPolicyBucket = []byte("removed_default_policies")
+
+	policyCtx = sac.WithGlobalAccessScopeChecker(context.Background(),
```
connorgorman, line 27: policyCtx = context.Background()

We don't need sac inside this store because there is no evaluation for it within bolt

central/policy/store/boltdb/store.go
```diff
@@ -0,0 +1,150 @@
...
+// New returns a new Store instance using the provided bolt DB instance.
+func newWithoutDefaults(db *bolt.DB) Store {
+	bolthelper.RegisterBucketOrPanic(db, policyBucket)
+	bolthelper.RegisterBucketOrPanic(db, removedDefaultPolicyBucket)
+	return &storeImpl{
+		DB: db,
+	}
+}
+
+func (s *storeImpl) addDefaults() {
```
connorgorman, line 106: What is the equivalent for this in Postgres currently?

tools/generate-helpers/pg-table-bindings/index.go.tpl
```diff
@@ -50,17 +50,29 @@ func (b *indexerImpl) Add{{.TrimmedType}}(deployment *{{.Type}}) error {
 	return nil
 }
 
+{{if eq .TrimmedType "Policy" }}
+func (b *indexerImpl) Add{{trimSuffix "y" .TrimmedType}}ies(_ []*{{.Type}}) error {
```
connorgorman, line 54: Can we just add logic that will do this automatically? So a function that checks if it ends with a y then will change the name to ies? etc?

connorgorman (changes requested):
See comments, not sure how we will do the default policies checks currently in postgres?

central/policy/store/boltdb/store.go
clickboo: None as of now, but can add a datastore layer init function that runs to add these if they don't exist.

central/policy/store/boltdb/store.go
clickboo: ok didn't change this bit (just moved the location of the file) but changing to context.Background

central/policy/store/boltdb/store.go
connorgorman: Right, but if we do a migration, we lose the knowledge that the defaults were removed right?

clickboo:
> See comments, not sure how we will do the default policies checks currently in postgres?

A datastore init method to call the upsert on the default policies? Since default policies are now not editable, we can skip if policy id exists.

central/policy/store/boltdb/store.go
clickboo: We don't allow removal of default policies 66 onwards.

tools/generate-helpers/pg-table-bindings/index.go.tpl
clickboo: you mean add a new func to the template funcs?

clickboo: Expect a lot more changes - there a bunch of types that end in 'y' which had their plural naming messed up.

PR #1916, opened 2022-06-02
ROX-11145: Change to Gorm schema

c-du:
## Description

## Checklist

## Testing Performed

CI tests and update schema_test to go through the new code path.


central/alert/datastore/internal/store/postgres/store.go
```diff
@@ -672,6 +671,12 @@ func Destroy(ctx context.Context, db *pgxpool.Pool) {
 	dropTableAlerts(ctx, db)
 }
 
+// NewTestStore returns a new Store instance for testing
+func NewTestStore(ctx context.Context, db *pgxpool.Pool, gormDB *gorm.DB) Store {
```
dashrews78, line 675: Is this how we usually handle cases like this where we need the class set up slightly differently for test?  I ask because when I see things like this I always concerned about something that isn't related to testing using it.  I would probably prefer to see the call to ApplySchemaForTable in the test setup itself if that were possible.

c-du: Ideally if we could have another for test lib in this package. But since destroy is already here, I would add it here. I have the first version with ApplySchemaForTable in tests but this way it is much cleaner.

dashrews78: I see what you are saying about the Destroy being there as a precedent, but along those lines, couldn't we make a CreateGormTestSchema function to match the Destroy instead of creating a separate instantiation?  All New does is set the pool.  So couldn't the test code do something like 
```
s.store = New(ctx, s.pool)
gormDB := pgtest.OpenGormDB(s.T(), source)
s.CreateGormTestSchema(ctx, gormDB) // or something like that
```

c-du: See the store are created at many different tests in different pkgs, not only the local store tests. And the table name needs to come from different sources, and many of them are not auto generated. I then kept asking myself why I would ask for the table name from others if the store already had it. I wont try to figure out one by one that is a disaster I wont go back.

dashrews78: fair enough.

viswajithiii (approved):
LGTM once you add the `*testing.T` parameter.

dashrews78 (approved):

c-du:
> LGTM once you add the `*testing.T` parameter.

Per our chat offline, I will rename it to CreateTableAndNewStore. We wont accidently use it because central wont have a gormDB.

md2119:
Saw the `*testing.T` conversation after the page reload.

PR #1917, opened 2022-06-02
ROX-11053: Use v1.Query for for multikey resources

janisz:
## Description

Use v1.Query for multikey resources instead of SQL string. This will unify generated code to use v1.Query for every operation (except methods generated for tests `DROP TABLE...`).

## Testing Performed

CI

### DIFF
--- a/central/componentcveedge/datastore/internal/postgres/store.go
+++ b/central/componentcveedge/datastore/internal/postgres/store.go
@@ -26,10 +26,6 @@ import (
 const (
 	baseTable = "image_component_cve_edges"
 
-	existsStmt = "SELECT EXISTS(SELECT 1 FROM image_component_cve_edges WHERE Id = $1 AND ImageComponentId = $2 AND ImageCveId = $3)"
-	getStmt    = "SELECT serialized FROM image_component_cve_edges WHERE Id = $1 AND ImageComponentId = $2 AND ImageCveId = $3"
-	deleteStmt = "DELETE FROM image_component_cve_edges WHERE Id = $1 AND ImageComponentId = $2 AND ImageCveId = $3"
-
 	batchAfter = 100
 
 	// using copyFrom, we may not even want to batch.  It would probably be simpler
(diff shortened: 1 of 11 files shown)


janisz:
* tj/add_search_tag_for_pk
  * **PR #1917** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/1917" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/1917?utm_source=stack-comment).

md2119:
We could remove the extra PKs if we mark the needed columns searchable so that it shows up as top-level. I remember seeing a PR that marked those cols as searchable. If that PR has been closed, let me know, and I can TAL.

janisz:
@md2119 it's this PR :) 
So we need to decide which way we should go FK or search tags as we can't have both

md2119:
> @md2119 it's this PR :) So we need to decide which way we should go FK or search tags as we can't have both

Decision: 
Keep multiple pks and not add search tags to those.

Background:
Upon further investigation, I realized that adding search tags on the graph paths breaks current searching in dackbox.

central/componentcveedge/datastore/internal/postgres/store.go
```diff
@@ -81,27 +77,34 @@ func (s *storeImpl) Count(ctx context.Context) (int, error) {
...
-	row := s.db.QueryRow(ctx, existsStmt, id, imageComponentId, imageCveId)
-	var exists bool
-	if err := row.Scan(&exists); err != nil {
-		return false, pgutils.ErrNilIfNoRows(err)
-	}
-	return exists, nil
+	var sacQueryFilter *v1.Query
+
+	q := search.ConjunctionQuery(
+		sacQueryFilter,
```
md2119, line 83: umm, why are we adding `sacQueryFilter` when we know it is nil?

janisz: Good question. I think the name of this PR was wrong, I'll fixt that. This change unify the generated code to not use SQL but `v1.Query` instead (look at removed lines at the top of the file). Previously we cannot have it (v1.Query) for multikey resources so I skipped them. SAC filter  is nil because we have not implemented SAC for indirectly scoped resources. It will be fixed in another PR [ROX-11159](https://issues.redhat.com/browse/ROX-11159)

md2119 (approved):

central/componentcveedge/datastore/internal/postgres/store.go
connorgorman: What is the current functionality if sacQueryFilter is nil in a conjunction query? Does this currently just make it not work?

connorgorman (approved):
same question about the sac filter being nil and the impact of that

rhybrillou (approved):

PR #1919, opened 2022-06-02
ROX-11107: Test docs branch create access

0x656b694d:
As a step of upstream release automation, the workflow will create a release branch in the `openshift/openshift-docs` repository. The PR adds a manually triggered GitHub workflow to test if the `ORG_TOKEN_FOR_GITHUB` secret has enough access to do that.


0x656b694d:
Indeed. Let's try both.

PR #1923, opened 2022-06-02
ROX-7834: Increase timeout for roxctl commands in flaky test

mtodor:
## Description

I wasn't able to reproduce this problem locally. I had 0/100 fails. The assumption is that we had timeouts in past and the only reasonable improvement is to increase the timeout for roxctl calls.

This change does not affect the correctness of tests.

## Considerations

We are currently running:
```
test_roxctl_cmd central whoami
test_roxctl_cmd central db backup
```

It could be that `db backup` is expensive in some cases. Would it be better to replace it with some less havier command, but with the same required auth rights?

## Checklist
- ~[ ] Unit test and regression tests added~ - not relevant
- ~[ ] Evaluated and added CHANGELOG entry if required~ - not relevant
- ~[ ] Determined and documented upgrade steps~ - not relevant
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~ - not relevant

## Testing Performed

Executed test in loop locally:
```
for _ in {1..2}; do tests/roxctl/token-file.sh; done | rg Passed | wc -l
```

### DIFF
--- a/tests/roxctl/token-file.sh
+++ b/tests/roxctl/token-file.sh
@@ -15,7 +15,7 @@ eecho() {
 
 # Retrieve API token
 TOKEN_FILE=$(mktemp)
-curl -k -f \
+curl -k -f --silent \
   -u "admin:$ROX_PASSWORD" \
   -d '{"name": "test", "role": "Admin"}' \
   "https://$API_ENDPOINT/v1/apitokens/generate" \
@@ -29,6 +29,7 @@ test_roxctl_cmd() {
 
   # Verify that specifying a token file works.
   if OUTPUT=$(roxctl --insecure-skip-tls-verify --insecure -e "$API_ENDPOINT" \
+    --timeout 2m \
     --token-file "$TOKEN_FILE" \
     "$@" \
     2>&1); then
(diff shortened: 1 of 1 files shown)


tests/roxctl/token-file.sh
```diff
@@ -29,6 +29,7 @@ test_roxctl_cmd() {
 
   # Verify that specifying a token file works.
   if OUTPUT=$(roxctl --insecure-skip-tls-verify --insecure -e "$API_ENDPOINT" \
+    --timeout 2m \
```
stehessel, line 32: The default timeout for `roxctl central db` is `1h`. Did you mean to set this to `2h` instead of `2m`?

stehessel:
A one hour timeout seems already very long for a CI run. I think it would be good to understand in which scenarios this command could run longer than that.

stehessel:
Maybe the issue is a missing retry rather than the timeout? The timeouts seem sufficient to me, I think something else is going on if they are hit.

stehessel:
I think option 2 is also reasonable, fixing transient issues without logs is not really possible.

mtodor:
I'll close PR. In case this fails again, we at least have some references to get back to them.

PR #1924, opened 2022-06-02
ROX-9652: Version port to Postgres

connorgorman:
## Description

Initial move of Version to Postgres.  Plan is for it to work mostly the same as it does now so a straight port is the first step.

## Checklist

## Testing Performed

CI failures are unrelated and already have tickets.


connorgorman:
Okay this one is unmergeable until we figure out the migrator

dashrews78:
@connorgorman I checked with @c-du and she currently anticipates version working basically the same as it does before postgres where we simply use a sequence number.  Obviously that may evolve, but I think a simple port, will suffice initially.

central/main.go
```diff
@@ -132,6 +132,7 @@ import (
 	"github.com/stackrox/rox/central/ui"
 	userService "github.com/stackrox/rox/central/user/service"
 	"github.com/stackrox/rox/central/version"
+	"github.com/stackrox/rox/central/version/store"
```
c-du, line 135: Since this is central/main, it would be better to call it like vStore.

c-du (approved):
With a minor comment.

PR #1927, opened 2022-06-02
[ROX-11295] Postgres store for nodes

md2119:
## Description

- Added Postgres store for nodes
- Updated node merge code to account for suppressed state
- Updated split/merge code with deduplication
- Removed newly added `storage.NodeComponentCVEEdge` proto

## Checklist

## Testing Performed
Unit


pkg/postgres/schema/node_components_cves_edges.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+var (
+	// CreateTableNodeComponentsCvesEdgesStmt holds the create statement for table `node_components_cves_edges`.
+	CreateTableNodeComponentsCvesEdgesStmt = &postgres.CreateStmts{
+		Table: `
+               create table if not exists node_components_cves_edges (
+                   Id varchar,
+                   IsFixable bool,
+                   FixedBy varchar,
```
connorgorman, line 23: Can we fix this redundancy? FixedBy != null is IsFixable

central/nodecomponentcveedge/datastore/store/postgres/gen.go
```diff
@@ -0,0 +1,3 @@
+package postgres
+
+//go:generate pg-table-bindings-wrapper --type=storage.ComponentCVEEdge --table=node_components_cves_edges --search-category NODE_COMPONENT_CVE_EDGE --references=node_components:storage.ImageComponent,node_cves:storage.CVE --join-table true
```
connorgorman, line 3: Why still storage.ImageComponent, and storage.CVE? I was under the impression we were going to split these out?

pkg/postgres/schema/node_components_cves_edges.go
md2119: Such tasks should not go under postgres umbrella. It is more of a deprecation that requires us to visit UI and service layer, wrapper code, etc.

proto/storage/cve.proto
```diff
@@ -88,6 +88,44 @@ message CVE {
     VulnerabilitySeverity       severity            = 20 [(gogoproto.moretags) = "search:\"Severity\""];
 }
 
+message CVEInfo {
+    message Reference {
```
connorgorman, line 92: I guess we should keep this, but I think we can just deprecate it. It's never been widely used in our platform

connorgorman (approved):

PR #1929, opened 2022-06-02
Port LogImbue to Postgres

connorgorman:
## Description

Ports log imbue over. I have created a ticket for the fact that these can grow continuously without ever being pruning and added the ts field so that this is an easy query in the future. There is also another ticket to reduce the size of the payloads

## Checklist

## Testing Performed

CI, plus I guess I can try to break the UI


central/logimbue/store/singleton.go
```diff
@@ -13,7 +18,11 @@ var (
 // Singleton returns the singleton instance for the sensor connection manager.
 func Singleton() Store {
 	storeInstanceInit.Do(func() {
-		storeInstance = newStore(globaldb.GetGlobalDB())
+		if features.PostgresDatastore.Enabled() {
+			storeInstance = postgres.New(context.TODO(), globaldb.GetPostgres())
```
connorgorman, line 22:
```suggestion:
storeInstance = postgres.New(globaldb.GetPostgres())
```

pkg/postgres/schema/log_imbues.go
```diff
@@ -0,0 +1,48 @@
...
+
+var (
+	// CreateTableLogImbuesStmt holds the create statement for table `log_imbues`.
+	CreateTableLogImbuesStmt = &postgres.CreateStmts{
+		Table: `
+               create table if not exists log_imbues (
+                   Id varchar,
+                   serialized bytea,
+                   PRIMARY KEY(Id)
+               )
```
viswajithiii, line 21: Why is the timestamp field missing here? 🤔

connorgorman: I didn't mark it with search currently. Will only do that once we actually use those endpoints

viswajithiii: Ah, of course.

viswajithiii (approved):

PR #1930, opened 2022-06-03
ROX-11250: Scan images during admission request for image signature policy field

dhaus67:
## Description

When creating a policy for image signatures and enabling enforcement via the admission controller, the policy will currently fail although the image _does_ have a valid signature.

With the following image, `gcr.io/distroless:debug` and its [respective key](https://github.com/GoogleContainerTools/distroless/blob/main/cosign.pub), a policy with the image signature field will not create any alerts. When trying to create a deployment with that image and enforcement enabled, the following will be returned:
```bash
Error from server (Failed currently enforced policies from StackRox): error when creating "dep-distroless.yaml": admission webhook "policyeval.stackrox.io" denied the request:
The attempted operation violated 1 enforced policy, described below:

Policy: Image Signature Policy
- Description:
    ↳
- Rationale:
    ↳
- Remediation:
    ↳
- Violations:
    - Container 'base' image signature is unverified
```
_Note: As a caveat, when deploying the image consecutive and using SHA reference instead of tag reference, a cached image, if it exists, will return. When the enforcement was manually tested, within demo instances that were used the `unscanned image` policy was enabled, thus scanning and upserting the image in the cache, which lead to the cached image being returned during admission controller enforcement testing 😞 ._
 
The reason for this is due to the admission controller only waiting for the scan result if one of the alert's policy has the fieldname `Unscanned Image`:
```go
for !hasNonNoScanAlerts(alerts) && err == nil {
	// Wait for the result and re-evaluate the deployment
}
```

Since this is not the case, the admission controller will return the generated alerts and immediately fail.

To fix this behavior, the following has been done in the PR:
- Rename `ContainsUnscannedImageField` to `ContainsScanRelatedFields`.
- Within ^, add an additional check for the `ImageSignatureVerifiedBy` field. 

What will be done in a follow-up PR:
- Add more testcases to `ImageSignatureVerificationTests` to also cover the admission controller blocking / not blocking deployments, re-using the existing test cases done for alert matching. 

## Checklist

## Testing Performed
- Integration tests extended (`AdmissionControllerNoImageScanTest`).
- Manual testing
  - Create a signature integration + policy with enforcement mode usign [the distroless public key](https://github.com/GoogleContainerTools/distroless/blob/main/cosign.pub)
  - Two delpoyments for testing, one with an unsigned image, the other one with the distroless image:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: unsigned-image
  name: unsigned-image
  namespace: signature-testing
spec:
  replicas: 1
  selector:
    matchLabels:
      app: unsigned-image
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: unsigned-image
    spec:
      containers:
      - image: nginx:latest
        name: nginx
        resources: {}
status: {}
```
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: signed-image-distroless
  name: signed-image-distroless
  namespace: signature-testing
spec:
  replicas: 1
  selector:
    matchLabels:
      app: signed-image-distroless
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: signed-image-distroless
    spec:
      containers:
      - image: gcr.io/distroless/base:debug
        name: base
        command: ["sleep", "600000"]
        resources: {}
status: {}
```
  - Deploy the image without signature and observe the following output:
```bash
k create -f dep-unsigned.yaml
Error from server (Failed currently enforced policies from StackRox): error when creating "dep-unsigned.yaml": admission webhook "policyeval.stackrox.io" denied the request:
The attempted operation violated 1 enforced policy, described below:

Policy: SigPolicy
- Description:
    ↳
- Rationale:
    ↳
- Remediation:
    ↳
- Violations:
    - Container 'nginx' image signature is unverified


In case of emergency, add the annotation {"admission.stackrox.io/break-glass": "ticket-1234"} to your deployment with an updated ticket number
``` 
  - Deploy the distroless image and observe the following output:
```bash
k create -f dep-distroless.yaml
deployment.apps/signed-image-distroless created
```

### DIFF
--- a/pkg/booleanpolicy/policyfields/fields.go
+++ b/pkg/booleanpolicy/policyfields/fields.go
@@ -106,9 +106,10 @@ func ContainsMemResourceLimit(p *storage.Policy) bool {
 	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.ContainerMemLimit)
 }
 
-// ContainsScanRelatedFields returns whether the policy contains fields related to image scanning,
-// i.e. fieldnames.UnscannedImage or fieldnames.ImageSignatureVerifiedBy.
-func ContainsScanRelatedFields(p *storage.Policy) bool {
+// ContainsScanRequiredFields returns whether the policy contains fields related to image scanning,
+// which require a scan result and may otherwise fail, i.e. fieldnames.UnscannedImage or
+// fieldnames.ImageSignatureVerifiedBy.
+func ContainsScanRequiredFields(p *storage.Policy) bool {
 	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.UnscannedImage) ||
 		booleanpolicy.ContainsValueWithFieldName(p, fieldnames.ImageSignatureVerifiedBy)
 }
--- a/pkg/booleanpolicy/policyfields/fields.go
+++ b/pkg/booleanpolicy/policyfields/fields.go
@@ -106,7 +106,9 @@ func ContainsMemResourceLimit(p *storage.Policy) bool {
 	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.ContainerMemLimit)
 }
 
-// ContainsUnscannedImageField returns whether the policy contains the unscanned image field.
-func ContainsUnscannedImageField(p *storage.Policy) bool {
-	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.UnscannedImage)
+// ContainsScanRelatedFields returns whether the policy contains fields related to image scanning,
+// i.e. fieldnames.UnscannedImage or fieldnames.ImageSignatureVerifiedBy.
+func ContainsScanRelatedFields(p *storage.Policy) bool {
+	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.UnscannedImage) ||
+		booleanpolicy.ContainsValueWithFieldName(p, fieldnames.ImageSignatureVerifiedBy)
 }
--- a/qa-tests-backend/src/test/groovy/AdmissionControllerNoImageScanTest.groovy
+++ b/qa-tests-backend/src/test/groovy/AdmissionControllerNoImageScanTest.groovy
@@ -1,13 +1,16 @@
 import groups.BAT
 import io.stackrox.proto.storage.ClusterOuterClass.AdmissionControllerConfig
 import io.stackrox.proto.storage.PolicyOuterClass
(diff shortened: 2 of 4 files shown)


pkg/booleanpolicy/policyfields/fields.go
```diff
@@ -106,7 +106,9 @@ func ContainsMemResourceLimit(p *storage.Policy) bool {
...
 }
 
-// ContainsUnscannedImageField returns whether the policy contains the unscanned image field.
-func ContainsUnscannedImageField(p *storage.Policy) bool {
-	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.UnscannedImage)
+// ContainsScanRelatedFields returns whether the policy contains fields related to image scanning,
+// i.e. fieldnames.UnscannedImage or fieldnames.ImageSignatureVerifiedBy.
+func ContainsScanRelatedFields(p *storage.Policy) bool {
+	return booleanpolicy.ContainsValueWithFieldName(p, fieldnames.UnscannedImage) ||
+		booleanpolicy.ContainsValueWithFieldName(p, fieldnames.ImageSignatureVerifiedBy)
```
dhaus67, line 113: This list isn't really complete, right now, only `fieldnames.UnscannedImage, fieldnames.ImageSignatureVerifiedBy` are used, there's also `fieldnames.ImageScanAge`.
Since I'm anyways touching this, would it make sense to also add the scan age one here?

connorgorman: I would not add the other ones, but it does make sense for ImageSignatureVerifiedBy. Just note that if the scan/signature cannot be pulled then it will be fail open based on the current code. This is actually how it is for the other ones as well if the scan is unsuccessful, just by the nature of the scan being nil so things like image scan age, etc cannot be checked

connorgorman (approved):

pkg/booleanpolicy/policyfields/fields.go
viswajithiii: Yeah, we should rename the function to be more clear. These are not just scan related fields, they are a specific kind of field where the policy might fail _without_ a scan but pass _after_ a scan is received. All the other scan related fields (CVE, scan age) are such that getting a scan makes the policy strictly more likely to fail.

viswajithiii (approved):

connorgorman (approved):

PR #1933, opened 2022-06-03
ROX-11059: Finish environment.Environment migration

dhaus67:
## Description

This PR finishes the open migration for environment.Environment for all commands.

Since the "common" package is pretty much intertwined, I decided to move the environment's implementation from `roxctl/common/environment` to `roxctl/common`, but left the interface definitions. Otherwise, we would have cyclic imports.

What will be done in a follow-up:
- Re-organize the `common` pkg and create independent packages, such that the environment can live in its own package and the `flag` package can also be migrated to the `environment` without introducing cyclic imports.

## Testing Performed
- see CI (`style-checks` should pass).


.golangci.yml
```diff
@@ -69,6 +74,14 @@ linters-settings:
   gocritic:
     enabled-checks:
       - commentFormatting
+  forbidigo:
```
janisz, line 77: With this it should be safe to remove [`tools/roxvet/analyzers/invalidoutputroxctl`](https://github.com/stackrox/stackrox/blob/b49ecfbdfb1924a9c55ae0c6675d79a7b457794f/tools/roxvet/analyzers/invalidoutputroxctl/analyzer.go) but it can be done as a follow up

roxctl/helm/derivelocalvalues/derivelocalvalues.go
```diff
@@ -345,13 +346,14 @@ func retrieveCustomEnvVars(envVars map[string]interface{}) map[string]interface{
...
+func printWarnings(env common2.Environment, warnings []string) {
 	if len(warnings) == 0 {
 		return
 	}
-	fmt.Fprintln(os.Stderr, "The following warnings occured:")
+
+	env.Logger().WarnfLn("The following warnings occured:")
 	for _, msg := range warnings {
-		fmt.Fprintf(os.Stderr, "  WARNING: %s\n", msg)
+		env.Logger().WarnfLn("  WARNING: %s", msg)
```
janisz, line 356:
```suggestion
		env.Logger().WarnfLn("  - %s", msg)
```

roxctl/common/tls.go
```diff
@@ -88,7 +87,7 @@ func tlsConfigOptsForCentral() (*clientconn.TLSConfigOptions, error) {
 			return nil, errors.Errorf("CA certificates file %s contains no certificates!", flags.CAFile())
 		}
 		if flags.SkipTLSValidation() != nil && *flags.SkipTLSValidation() {
-			fmt.Fprintln(os.Stderr, "Warning: --insecure-skip-tls-verify has no effect when --ca is set")
+			CLIEnvironmentSingleton().Logger().WarnfLn("Warning: --insecure-skip-tls-verify has no effect when --ca is set")
```
janisz, line 90:
```suggestion
			CLIEnvironmentSingleton().Logger().WarnfLn("--insecure-skip-tls-verify has no effect when --ca is set")
```
Also  `insecure-skip-tls-verify` and `ca` should be extracted to const

roxctl/common/zipdownload/download_zip.go
```diff
@@ -118,16 +115,19 @@ func GetZip(opts GetZipOptions) error {
 	zipFileName, err := download.ParseFilenameFromHeader(resp.Header)
 	if err != nil {
 		zipFileName = fmt.Sprintf("%s.zip", opts.BundleType)
-		printf("Warning: could not obtain output file name from HTTP response: %v.", err)
-		printf("Defaulting to filename %q", zipFileName)
+		common.CLIEnvironmentSingleton().Logger().WarnfLn(
+			"Warning: could not obtain output file name from HTTP response: %v", err)
```
janisz, line 119:
```suggestion
			"could not obtain output file name from HTTP response: %v", err)
```

janisz: Just an idea for a separated task if we have debug mode it will be great to use `%+v` to print stacktrace

roxctl/helm/derivelocalvalues/derivelocalvalues.go
```diff
@@ -345,13 +346,14 @@ func retrieveCustomEnvVars(envVars map[string]interface{}) map[string]interface{
...
 	}
-	fmt.Fprintln(os.Stderr)
+	env.Logger().WarnfLn("")
```
janisz, line 358: I think we should remove this line it will generate empty warning `WARN: `

janisz (changes requested):
Could you please break this down into smaller changes? I do not have a lot of time for code reviews and this one is just too large/complex to review in one pass. :)

I think we could have a separated PR for 
- creating `common.Environment`
- Using logger instead of fmt.Print
- Using errox
- adding goforbid and removing custom one

roxctl/helm/derivelocalvalues/derivelocalvalues.go
```diff
@@ -2,14 +2,15 @@ package derivelocalvalues
...
-	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/pkg/maputil"
+	common2 "github.com/stackrox/rox/roxctl/common"
```
janisz, line 13:
```suggestion
	roxctlcommon "github.com/stackrox/rox/roxctl/common"
```
or rename helm common

dhaus67:
@janisz I've split up the changes into four PRs as suggested, PTAL once more (comments should be addressed within the other PRs).

janisz (approved):

PR #1936, opened 2022-06-03
Revert "Remove `clean-image` from `all-builds` target"

msugakov:
## Description

This reverts commit 3149711ad41787b732c6afb674e5384848af6c9b. https://github.com/stackrox/stackrox/pull/1875

The reason is that `create-bundle.sh` looks not only in `image/rhel` directory, but also takes files from `image/bin`, `image/static-bin` and others.

As a result of the change, Makefile did not detect that bundle needs to be regenerated when, for example, new `roxctl` binaries
were built, and old binaries were packaged in the bundle tarball and then in image.

The bug can be reproduced by the following sequence:
1. Run `make image`
2. Make any changes in Go code and commit them
3. Run `make image` again
4. Compare results of `roxctl version --json` and `docker run --rm -it quay.io/stackrox-io/main:$(make tag) version --json`, notice reported commits are different.

See also https://srox.slack.com/archives/CELUQKESC/p1654264051403859

## Checklist

The following aren't needed

## Testing Performed

* Followed reproduction scenario above and saw that in-container and out-of-container commits are the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -510,15 +510,13 @@ generate-junit-reports: $(GO_JUNIT_REPORT_BIN)
 image: main-image
 
 .PHONY: all-builds
-all-builds: cli main-build $(MERGED_API_SWAGGER_SPEC) ui-build
+all-builds: cli main-build clean-image $(MERGED_API_SWAGGER_SPEC) ui-build
 
 .PHONY: main-image
 main-image: all-builds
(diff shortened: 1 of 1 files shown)


stehessel (approved):
Alright roll it back. This was more of a quick fix, I think long term we should remove the bundles all together. That will enable proper caching.

PR #1937, opened 2022-06-03
ROX-11252: Fix `deploy-local.sh` in development context

msugakov:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


gavin-stackrox (commented):
*.sh looks good to me.
I'll defer *.go etc to others.

pkg/helm/charts/tests/centralservices/flavor/flavor_test.go
```diff
@@ -52,30 +53,24 @@ func TestOverriddenTagsAreRenderedInTheChart(t *testing.T) {
...
-			testutils.SetVersion(t, testutils.GetExampleVersionUnified(t))
-			return defaults.RHACSReleaseImageFlavor()
-		},
-		"custom": func() defaults.ImageFlavor {
-			return customFlavor(t)
-		},
+	testutils.SetVersion(t, testutils.GetExampleVersion(t))
+	imageFlavorCases := map[string]defaults.ImageFlavor{
+		"development": defaults.DevelopmentBuildImageFlavor(),
+		"stackrox":    defaults.StackRoxIOReleaseImageFlavor(),
```
fredrb, line 59: The reason we used callbacks is that `stackrox` and `rhacs` need to call `testutils.GetExampleVersionUnified(t)` rather than `testutils.GetExampleVersion(t)`. But I see that tests are passing and as long as the flavor is calling `version.GetAllVersionsDevelopment()` or `version.GetAllVersionsUnified()` we should be ok. I'm just puzzled why this was there in the first place. Maybe we don't need it after all.

pkg/images/defaults/flavor_test.go
```diff
@@ -84,6 +84,26 @@ func (s *imageFlavorTestSuite) TestGetImageFlavorFromEnv() {
 	}
 }
 
+func (s *imageFlavorTestSuite) TestOpenSourceImageFlavorDevReleaseTags() {
+	f := OpenSourceImageFlavor()
```
fredrb, line 88: I think these tests would be more readable if you assert on the example versions set in `testutils.SetExampleVersion(s.T())`. 

```go
if buildinfo.ReleaseBuild {
     s.Equal("3.0.99.0", f.MainImageTag)
     s.Equal("3.0.99.0", f.CollectorImageTag)
     // ...
}
```
WDYT?

pkg/helm/charts/tests/centralservices/flavor/flavor_test.go
```diff
@@ -52,30 +53,24 @@ func TestOverriddenTagsAreRenderedInTheChart(t *testing.T) {
...
+		imageFlavorCases["opensource-development"] = defaults.OpenSourceImageFlavor()
 	}
 
 	for name, f := range imageFlavorCases {
-		imageFlavor := f()
 		t.Run(name, func(t *testing.T) {
+			f := f
 			helmChartTestUtils.RunHelmTestSuite(t, testDir, image.CentralServicesChartPrefix, helmChartTestUtils.RunHelmTestSuiteOpts{
-				Flavor:       &imageFlavor,
+				Flavor:       &f,
```
fredrb, line 73: Could we keep the variable name as `imageFlavor`? Change it in the range variable as well just to be a bit more clear.

pkg/helm/charts/tests/securedclusterservices/flavor/flavor_test.go
```diff
@@ -32,27 +33,23 @@ func TestOverriddenTagsAreRenderedInTheChart(t *testing.T) {
...
+		imageFlavorCases["opensource-development"] = defaults.OpenSourceImageFlavor()
 	}
 
 	for name, f := range imageFlavorCases {
-		imageFlavor := f()
 		t.Run(name, func(t *testing.T) {
+			f := f
 			helmChartTestUtils.RunHelmTestSuite(t, testDir, image.SecuredClusterServicesChartPrefix, helmChartTestUtils.RunHelmTestSuiteOpts{
-				Flavor: &imageFlavor,
+				Flavor: &f,
```
fredrb, line 52: Same as above

pkg/helm/charts/tests/securedclusterservices/flavor/flavor_test.go
```diff
@@ -32,27 +33,23 @@ func TestOverriddenTagsAreRenderedInTheChart(t *testing.T) {
...
+	testutils.SetVersion(t, testutils.GetExampleVersion(t))
+	imageFlavorCases := map[string]defaults.ImageFlavor{
+		"development": defaults.DevelopmentBuildImageFlavor(),
+		"stackrox":    defaults.StackRoxIOReleaseImageFlavor(),
+		"rhacs":       defaults.RHACSReleaseImageFlavor(),
+	}
+	if buildinfo.ReleaseBuild {
+		imageFlavorCases["opensource-release"] = defaults.OpenSourceImageFlavor()
+	} else {
+		imageFlavorCases["opensource-development"] = defaults.OpenSourceImageFlavor()
```
fredrb, line 45: Nice! I like how you split to run the respective `yaml` file based on the build flag.

fredrb (approved):
Looks good. Non-blocking comments.

pkg/helm/charts/tests/centralservices/flavor/flavor_test.go
msugakov: Flavors indeed call `version.GetAllVersionsDevelopment()` or `version.GetAllVersionsUnified()` and the latter one does this:
https://github.com/stackrox/stackrox/blob/c9392345af59a237ae52ec4b56fee9bfed156d99/pkg/version/version.go#L66-L71

Therefore we don't need to use `testutils.GetExampleVersionUnified(t)`.
I suppose the old setup was a result of either over-editing or back then things were working differently.

pkg/helm/charts/tests/securedclusterservices/flavor/flavor_test.go
msugakov: This was a workaround to me not seeing any elegant way to express the condition within one file :-P

pkg/helm/charts/tests/centralservices/flavor/flavor_test.go
msugakov: Done.

pkg/helm/charts/tests/securedclusterservices/flavor/flavor_test.go
msugakov: Done.

pkg/images/defaults/flavor_test.go
msugakov: Sure, both ways work for me.

PR #1939, opened 2022-06-03
ROX-10966 local-sensor: record and replay k8s events

lvalerom:
## Description

This PR aims to enable a way to record k8s events handled by sensor as well as a way to replay those events.

## Checklist

## Testing Performed

Local testing performed with `local-sensor`

```bash
go run tools/local-sensor/main.go -record -record-out=trace.jsonl
go run tools/local-sensor/main.go -replay -replay-in=trace.jsonl -delay=0s -resync=5s
```


sensor/debugger/k8s/trace.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Init initializes the reader
+func (tw *TraceReader) Init() error {
+	if !tw.Enabled || path.Dir(tw.Source) == "" {
+		return nil
+	}
+	tw.mu.Lock()
+	defer tw.mu.Unlock()
+	tw.lineNo = 1
+	return os.MkdirAll(path.Dir(tw.Source), os.ModePerm)
```
fredrb, line 68: Why are we creating a folder on _reading_ the trace file? If you just want to check if the file exists you can do `os.Stat`

sensor/debugger/k8s/trace.go
```diff
@@ -0,0 +1,145 @@
...
+			done <- 0
+			readerError <- err
+			return
+		}
+		handle(buf[:n], mode)
+	}
+}
+
+// Read reads one line from the trace file. This line corresponds to a single K8s event
+func (tw *TraceReader) Read(p []byte) (n int, err error) {
```
fredrb, line 89: Why do we want to read the file line-by-line? Isn't it easier to read the whole thing? `os.ReadFile` is a utility function that handles the reading buffer for you. Even if the file is MBs large, we won't read more than one, right? 
It would simplify a lot the logic here.

sensor/kubernetes/listener/resources/dispatcher.go
```diff
@@ -146,69 +196,69 @@ func (m metricDispatcher) ProcessEvent(obj, oldObj interface{}, action central.R
 }
 
 func (d *registryImpl) ForDeployments(deploymentType string) Dispatcher {
-	return wrapWithMetricDispatcher(newDeploymentDispatcher(deploymentType, d.deploymentHandler))
+	return wrapWithMetricDispatcher(wrapWithDumpingDispatcher(newDeploymentDispatcher(deploymentType, d.deploymentHandler), d.traceWriter))
```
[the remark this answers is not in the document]
fredrb: If this is going to be a thing, I mean. Adding more of these wrappers would be annoying.

tools/local-sensor/main.go
```diff
@@ -57,16 +58,67 @@ func createConnectionAndStartServer(fakeCentral *centralDebug.FakeService) (*grp
...
+	ShallReplayK8sTrace bool
+	ReplayK8sTraceFile  string
+	Verbose             bool
+	ResyncPeriod        time.Duration
+	CreateMode          k8s.CreateMode
+	Delay               time.Duration
+}
+
+func mustGetCommandLineArgs() localSensorConfig {
+	fsc := localSensorConfig{}
```
fredrb, line 75: nit: `s/fsc/sensorConfig`

tools/local-sensor/main.go
```diff
@@ -98,23 +149,47 @@ func main() {
...
 
 	conn, spyCentral, shutdownFakeServer := createConnectionAndStartServer(fakeCentral)
 	defer shutdownFakeServer()
 	fakeConnectionFactory := centralDebug.MakeFakeConnectionFactory(conn)
 
+	traceRec := &k8s.TraceWriter{
+		Destination: path.Clean(fsc.RecordK8sFile),
+		Enabled:     fsc.ShallRecordK8sInput,
+	}
+	_ = traceRec.Init()
```
fredrb, line 168: Could you please handle the error here?

tools/local-sensor/main.go
```diff
@@ -98,23 +149,47 @@ func main() {
...
-	registerHostKillSignals(startTime, fakeCentral, *outputFileFlag)
+	go registerHostKillSignals(startTime, fakeCentral, fsc.CentralOutput)
 
 	conn, spyCentral, shutdownFakeServer := createConnectionAndStartServer(fakeCentral)
 	defer shutdownFakeServer()
 	fakeConnectionFactory := centralDebug.MakeFakeConnectionFactory(conn)
 
+	traceRec := &k8s.TraceWriter{
+		Destination: path.Clean(fsc.RecordK8sFile),
+		Enabled:     fsc.ShallRecordK8sInput,
```
fredrb, line 166: Why are you guarding this by an `Enabled` flag and not an if statement here? Wouldn't it be simpler to:
```golang
var traceRec k8s.TraceWriter
if fsc.ShallRecordK8sInput {
        traceRec = &k8s.TraceWriter{
		Destination: path.Clean(fsc.RecordK8sFile),
        }
        _ = traceRec.Init()
}
```

tools/local-sensor/main.go
```diff
@@ -57,16 +58,67 @@ func createConnectionAndStartServer(fakeCentral *centralDebug.FakeService) (*grp
 	return conn, fakeCentral, closeF
 }
 
+type localSensorConfig struct {
+	Duration            time.Duration
+	CentralOutput       string
+	ShallRecordK8sInput bool
+	RecordK8sFile       string
+	ShallReplayK8sTrace bool
```
fredrb, line 66: nit: not a huge fan of prefixing these with `Shall*`. Maybe simply `ReplayK8sEnabled`?

fredrb (changes requested):
First pass done. Didn't test it myself yet. Let me know what you think about the suggestions, mainly the bigger one about using reflection and I'll continue reviewing later. I think it has some potential to simplify things here.

Conceptually, I think this is great, my comments are more from the code design perspective. This is a great step toward our brand new sensor test framework. 🙂

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,458 @@
...
+	objType := strings.Split(msg.ObjectType, ".")
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&msg.Payload)
+	if err != nil {
+		log.Printf("error constructing unstructured: %s", err)
+		return err
+	}
+	Kind := objType[1]
+	obj.Object = u
+
+	switch Kind {
```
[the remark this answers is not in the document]
lvalerom: Great suggestion! I've changed the code to use reflection. Let me know what you think.

sensor/debugger/k8s/trace.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	}
+	tw.mu.Lock()
+	defer tw.mu.Unlock()
+	tw.lineNo = 1
+	return os.MkdirAll(path.Dir(tw.Source), os.ModePerm)
+}
+
+// ReadFile reads the file line by line and executes the handle function
+func (tw *TraceReader) ReadFile(mode CreateMode, done chan int, readerError chan error, handle func([]byte, CreateMode)) {
```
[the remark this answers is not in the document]
lvalerom: Done!

sensor/debugger/k8s/trace.go
lvalerom: oops.. I think this was a mistake from copy-pasting from the writer 😅

sensor/debugger/k8s/trace.go
lvalerom: The initial idea was to have a generic reader that implemented `io.Reader` and pass it to `sensor` but it morphed into this. I simplified the code with your suggestion. Let me know what you think.

tools/local-sensor/main.go
```diff
@@ -76,14 +128,13 @@ func registerHostKillSignals(startTime time.Time, fakeCentral *centralDebug.Fake
...
 func main() {
-	durationFlag := flag.Duration("duration", 0, "duration that the scenario should run (leave it empty to run it without timeout)")
-	outputFileFlag := flag.String("output", "results.json", "output all messages received to file")
-	verboseFlag := flag.Bool("verbose", false, "prints all messages to stdout as well as to the output file")
-	resyncPeriod := flag.Duration("resync", 1*time.Minute, "resync period")
-
-	flag.Parse()
-
+	// TODO(do-not-merge): remove test.jsonl file from git
+	fsc := mustGetCommandLineArgs()
```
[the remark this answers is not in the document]
lvalerom: Done

tools/local-sensor/main.go
lvalerom: oops! Done

sensor/kubernetes/listener/resources/dispatcher.go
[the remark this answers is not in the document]
lvalerom: Yeah. I think it's much more convenient your way.

tools/local-sensor/main.go
lvalerom: I got rid of the Enable field in both `TraceWriter` and `TraceReader` I think we don't need them (at least for the time being)

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,271 @@
...
+func getNamespace(resource reflect.Value) string {
+	values := resource.MethodByName("GetNamespace").Call([]reflect.Value{})
+	if len(values) != 1 {
+		return ""
+	}
+	return values[0].String()
+}
+
+// handleRunOp handles the execution of runOp
+func (f *FakeEventsManager) handleRunOp(action, kind string, client, object reflect.Value, ch chan string) error {
```
vikin91, line 213: Following on my previous comment, this function is a writer so it should create the channel inside and return it to the caller.

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,271 @@
...
+		default:
+		}
+		f.waitOnMode()
+		return nil
+	}
+	return errInt.(error)
+}
+
+// createEvent creates a single k8s event
+func (f *FakeEventsManager) createEvent(msg resources.InformerK8sMsg, ch chan string) error {
```
vikin91, line 231: Going further here, this function should also create a channel, multiplex the channels returned by `handleRunOp` and return the channel containing all messages to the caller

sensor/debugger/k8s/trace.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Init initializes the reader
+func (tw *TraceReader) Init() error {
+	_, err := os.Stat(path.Dir(tw.Source))
+	return err
+}
+
+// ReadFile reads the entire file and returns a slice of objects
+func (tw *TraceReader) ReadFile() ([][]byte, error) {
```
vikin91, line 58: This is fine for now but we need to make sure to produce a meaningful error in case the file is too big.

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+		case <-doneC:
+			return errors.New("the events file did not contain the minimum resources required to start sensor")
+		}
+	}
+}
+
+// CreateEvents creates the k8s events from a given jsonl file
+func (f *FakeEventsManager) CreateEvents() error {
+	ch := make(chan string)
```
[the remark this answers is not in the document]
vikin91: (read this comment after reading my other comments) 

Here we have the following possibilities:
1. Close the channel as suggested and see whether there are any issues. If any panics appear we need to then implement the concurrency patterns as I suggested in other comments.
2. Not close the channel and leave it as it is. In this particular case there should be no severe consequences of this. I think it might be fine leaving it like this as this coat will not be run in the production.

vikin91 (commented):
This looks really nice thanks for finishing it. I still haven't run the code but I will do this on Monday. I will review green or red after I play with the binary.

Please have a look at the issue with concurrency that I mentioned and tell me what you think about it.

vikin91:
I think that the https://github.com/stackrox/stackrox/pull/1799 is now obsolete and can be closed, right?

sensor/debugger/k8s/fake.go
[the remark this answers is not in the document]
lvalerom: I think we can just close the channel in the `CreateEvents` function. 

```golang
ch := make(chan string)
f.done = concurrency.NewSignal()
objs, err := f.Reader.ReadFile()
if err != nil {
    return err
}
f.Init()
go func() {
    defer close(ch)
```

lvalerom:
> I think that the #1799 is now obsolete and can be closed, right?

Yes, you can close #1799.

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,296 @@
...
+			if _, ok := minimumResources[obj]; ok {
+				minimumResources[obj]--
+				if minimumResources[obj] == 0 {
+					count++
+				}
+				if len(minimumResources) == count {
+					return nil
+				}
+			}
+		case <-doneC:
```
fredrb, line 164: Wouldn't `case <-done.WaitC()` work where?

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,296 @@
...
+			d = true
+		}
+	}
+}
+
+// CreateEvents creates the k8s events from a given jsonl file
+func (f *FakeEventsManager) CreateEvents() error {
+	f.done = concurrency.NewSignal()
+	objs, err := f.Reader.ReadFile()
+	ch := make(chan string, len(objs))
```
fredrb, line 188: The `length` of a channel isn't the same thing as the length of a slice. What you're defining here is a `buffered` channel that allows for sending data into a channel in a non-blocking way up until the buffer size. See [this as an example](https://go.dev/tour/concurrency/3).

I think it's fine to have a buffered channel and even more optimal here, but you have to consider this when handling the `done` signal. See my next comment.

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,296 @@
...
+				if minimumResources[obj] == 0 {
+					count++
+				}
+				if len(minimumResources) == count {
+					return nil
+				}
+			}
+		case <-doneC:
+			// We make sure there are not more messages in the channel
+			obj, more := <-ch
```
fredrb, line 166: I think this might lead to a race condition. 

If you're using a buffered channel, it means that the emitter of events (goroutine under `CreateEvents`) is allowed to write in a non-blocking way. Therefore, it could be that once the done signal is called, `ch` has more than one record waiting in the queue. So if we're only reading one here is not sufficient. You'd still have to loop until the channel is empty (or closed).

fredrb: I think it might make more sense to control the flow by simply iterating on the channel until closed rather than using two constructs (a channel and a signal).

fredrb (changes requested):
Thanks for making this reflection based. I think it's much cleaner now. I think there still might be some concurrency problems after you implemented @vikin91's suggestion.

sensor/debugger/k8s/fake.go
lvalerom: Yeah, it was a work in progress. After my chat with @vikin91 we changed this part a lot.

sensor/debugger/k8s/fake.go
lvalerom: You're right, it could lead to a race condition. We changed the code in this part too after the chat.

sensor/debugger/k8s/fake.go
lvalerom: I believe so. I've changed that in the last commit.

sensor/debugger/k8s/fake.go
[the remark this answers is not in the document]
lvalerom: Just to keep record here: After our chat we reworked all the concurrency and it is significantly different now.

sensor/debugger/k8s/trace.go
lvalerom: I think we can leave it like this for now. If at some point we need really large files we can always rework this part.

sensor/debugger/k8s/fake.go
fredrb: Ok. Let me know when it's ready to reviewed again.

lvalerom:
> I am getting
> 
> ```
> # github.com/stackrox/rox/sensor/debugger/k8s
> sensor/debugger/k8s/fake.go:14:2: invalid import "github.com/stackrox/rox/sensor/kubernetes/listener/resources": sensor/debugger cannot import from sensor/kubernetes/listener/resources; only allowed roots are [sensor/debugger generated image pkg]
> ```
> 
> when running `make style`. Could you check please?

That's weird, that should be fixed with the changes I added in the `analyzer`. Did you fetched the last changes? I've just tried `make style` in a fresh cloned repository and it seems to be working.

vikin91 (approved):
I am unblocking this, as we worked on this together, but please double-check it before merging.

I have run this, but was unable to see sensor stop on its own after all events were replayed - is this intended?

vikin91:
> > I am getting
> > ```
> > # github.com/stackrox/rox/sensor/debugger/k8s
> > sensor/debugger/k8s/fake.go:14:2: invalid import "github.com/stackrox/rox/sensor/kubernetes/listener/resources": sensor/debugger cannot import from sensor/kubernetes/listener/resources; only allowed roots are [sensor/debugger generated image pkg]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when running `make style`. Could you check please?
> 
> That's weird, that should be fixed with the changes I added in the `analyzer`. Did you fetch the last changes? I've just tried `make style` in a fresh cloned repository and it seems to be working.

If it works for you, then all is good :)

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,330 @@
...
+		return nil
+	case <-f.done.WaitC():
+		return errors.New("all events processed without reaching the minimumResourcesMap resources needed")
+	}
+}
+
+// WaitForDone waits for all the events to be processed
+func (f *FakeEventsManager) WaitForDone() {
+	f.done.Wait()
+}
```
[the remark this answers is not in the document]
lvalerom: This was supposed to be used in the tests.

The idea behind `WaitForDone` and `WaitForMinimumResources` was to encapsulate the handling of reading the channel in the `FakeEventsManager` but after a little bit of thinking I think I'm going with the approach we took this morning and just handle the channels outside.

sensor/debugger/k8s/fake.go
```diff
@@ -0,0 +1,330 @@
...
+
+// CreateEvents creates the k8s events from a given jsonl file
+func (f *FakeEventsManager) CreateEvents(ctx context.Context) <-chan error {
+	errorCh := make(chan error)
+	min, done, errCh := f.handleEventsCreation(ctx)
+	f.done = done
+	f.minimumResources = min
```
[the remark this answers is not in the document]
lvalerom: Yeah. I think it's simpler to just return them. I've changed the code accordingly.

tools/local-sensor/main.go
```diff
@@ -124,10 +228,11 @@ func main() {
 
 	spyCentral.ConnectionStarted.Wait()
 
-	<-time.Tick(*durationFlag)
+	log.Printf("Running scenario for %f minutes\n", localConfig.Duration.Minutes())
+	<-time.Tick(localConfig.Duration)
```
fredrb, line 232: If you decide to implement the suggestion above, you could handle the error signal here other than in a separate `goroutine` like it's being done now.

```golang
select {
   case <-time.Tick(localConfig.Duration):
      // ran scenario for long enough, continue
   case <- errSig.Done():
      // something went wrong 
}
```

fredrb (approved):
Thanks for working on the changes, Luis. I'm unblocking this as well, but I still think there are some potential improvements in the error handling. Consider using the constructs in our `concurrency` package to make the code more maintainable. 

If you agree with the suggestion but don't want to do this now, then could you please add a tech debt JIRA task in our local-sensor epic for this?

tools/local-sensor/main.go
lvalerom: Makes sense. I will do this in the new task.

lvalerom:
> Thanks for working on the changes, Luis. I'm unblocking this as well, but I still think there are some potential improvements in the error handling. Consider using the constructs in our `concurrency` package to make the code more maintainable.
> 
> If you agree with the suggestion but don't want to do this now, then could you please add a tech debt JIRA task in our local-sensor epic for this?

I agree with your suggestion. I think we can do that in a different PR so we can unblock this. There are a couple enhancements that need to be done either way for testing the re-sync. I will open follow up PRs for those as well.

lvalerom:
> @lvalerom create a ticket in local sensor epic also, please :-)

Here [ROX-11396](https://issues.redhat.com/browse/ROX-11396) :)

PR #1943, opened 2022-06-03
Remove extra primary keys from edge objects

md2119:
## Checklist


connorgorman:
Failing unit tests

md2119:
Adding redundant search tags to edges breaks searching in dackbox because the dackbox relies on resolving options map in  order provided in compound searcher.

PR #1948, opened 2022-06-03
ROX-11111: Fix inability to scan internal images with SHA specified in name

RTann:
## Description

Sensor did not identify images with explicitly set SHAs as cluster-local. For example, any of the sample OpenShift applications. When deployed to a namespace `testdev`, an example image is: `image-registry.openshift-image-registry.svc:5000/testdev/nodejs-basic@sha256:31734e0a1e52996cde63a67c18eafebeb99149d0b3d0b56e1c5f31b0583ec44b`. In `sensor/kubernetes/listener/resources/convert.go`, we did not bother setting the image's `ID` nor `NotPullable` fields if the image's `ID` was already set (happens in a case like this where it is explicit). Therefore, setting `IsClusterLocal` was also skipped. This PR sets both `NotPullable` and `IsClusterLocal` even when the `ID` is already known and adds unit tests to check this.

## Checklist

## Testing Performed

Unit tests as well as running an OpenShift 4.x cluster and deploying a sample application (these all specify the tag explicitly in the image name)


viswajithiii:
@RTann I must confess I'm confused: what exactly was the root cause of the issue described in the ticket, and how does this code solve it? Would be great if you could update the PR description with a more detailed explanation. 🙂

RTann:
@viswajithiii I updated the description. Let me know if it still isn't clear

sensor/kubernetes/listener/resources/convert.go
```diff
@@ -334,7 +352,9 @@ func (w *deploymentWrap) populateImageIDs(pods ...*v1.Pod) {
 				if features.LocalImageScanning.Enabled() {
 					// imageutil.IsInternalImage requires Sensor to already know about the OpenShift internal registries,
 					// which is ok because Sensor listens for Secrets before it starts listening for Deployment-like resources.
-					image.IsClusterLocal = imageutil.IsInternalImage(image.GetName())
+					image.IsClusterLocal = imageutil.IsInternalImage(image.GetName(), &imageutil.IsInternalImageOptions{
```
viswajithiii, line 355: What is the difference between `w.registryStore` and the singleton? 🤔

sensor/common/imageutil/util.go
```diff
@@ -5,13 +5,24 @@ import (
...
+// IsInternalImageOptions provides options for determining if an image is "internal".
+type IsInternalImageOptions struct {
+	RegistryStore *registry.Store
+}
+
 // IsInternalImage determines if the image represented by the given name
 // is an "internal" image. An internal image is one which is hosted by an internal registry.
 // An internal registry is on which is only accessible from within the cluster in which it lives.
-func IsInternalImage(image *storage.ImageName) bool {
+func IsInternalImage(image *storage.ImageName, opts *IsInternalImageOptions) bool {
```
connorgorman, line 16: Just pass the registry into this function? It's not clear why we would want to use an options pattern here

sensor/kubernetes/listener/resources/convert.go
```diff
@@ -58,6 +59,9 @@ type deploymentWrap struct {
 	original         interface{}
 	portConfigs      map[portRef]*storage.PortConfig
 	pods             []*v1.Pod
+	// registryStore is the Sensor's registry store to use.
+	// This is typically only used for testing purposes.
```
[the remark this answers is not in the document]
connorgorman: I think it'd be better if for testing purposes, we just passed the registry options to the calls to populateImageMetadata instead of embedding the store in the deployment wrap

sensor/common/imageutil/util.go
RTann: I was originally trying to avoid passing `nil` to use the default, but that's exactly what I'm doing here, so I'll just pass the registry

sensor/kubernetes/listener/resources/convert.go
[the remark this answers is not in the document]
RTann: I meant this is only non-`nil` for testing. I thought it made more sense to make this an attribute of the deployment wrapper as we are testing a deployment wrapper method but as a parameter to `populateMetadata` makes sense too

sensor/kubernetes/listener/resources/convert.go
RTann: the singleton is meant to be used for all non-unit-testing purposes. Otherwise, a non-singleton registry store should be used for more control over how it is populated

sensor/common/imageutil/util.go
RTann: It was also meant to somewhat hide the registry from the users

sensor/common/imageutil/util.go
```diff
@@ -8,10 +8,18 @@ import (
...
 // An internal registry is on which is only accessible from within the cluster in which it lives.
-func IsInternalImage(image *storage.ImageName) bool {
+// IsInternalImage uses the data from the given store to determine this.
+// If registryStore is nil, registry.Singleton() is used as the store.
+func IsInternalImage(image *storage.ImageName, registryStore *registry.Store) bool {
 	// If the Sensor knows about the registry in which the image is hosted,
 	// then the image must be "internal" to the cluster, as Sensor only tracks
 	// "internal" registries.
-	reg, err := registry.Singleton().GetRegistryForImage(image)
+	var store *registry.Store
```
connorgorman, line 17: I would modify this and instead just ensure that registryStore is always set to a valid value?

connorgorman: I think having a pure function that then has a Singleton inside of it is a weird side effect of the function if you pass the wrong parameters. I agree that `(*registry.Store).HasRegistryForImage` would be cleaner

sensor/common/image/service_impl.go
```diff
@@ -55,7 +55,9 @@ func (s *serviceImpl) GetImage(ctx context.Context, req *sensor.GetImageRequest)
 
 	// Note: The Admission Controller does NOT know if the image is cluster-local,
 	// so we determine it here.
-	req.Image.IsClusterLocal = imageutil.IsInternalImage(req.GetImage().GetName())
+	// If Sensor's registry store has an entry for the given image's registry,
+	// it is considered cluster-local.
+	req.Image.IsClusterLocal = registry.Singleton().HasRegistryForImage(req.GetImage().GetName())
```
viswajithiii, line 60: Ideally, the registry should be a field in the `serviceImpl` rather than us invoking `registry.Singleton` here.

sensor/kubernetes/listener/resources/convert.go
```diff
@@ -273,10 +273,16 @@ func (w *deploymentWrap) getPods(hierarchy references.ParentHierarchy, labelSele
 
 func (w *deploymentWrap) populateDataFromPods(pods ...*v1.Pod) {
 	w.pods = pods
-	w.populateImageIDs(pods...)
+	// Sensor must already know about the OpenShift internal registries to determine if an image is cluster-local,
+	// which is ok because Sensor listens for Secrets before it starts listening for Deployment-like resources.
+	w.populateImageMetadata(registry.Singleton(), pods...)
```
viswajithiii, line 278: This dependency should ideally be injected too, but if you decide it's too painful, that's fine (since you're not making it worse than it was before in this respect)

viswajithiii (approved):

sensor/common/image/service_impl.go
RTann: done

sensor/kubernetes/listener/resources/convert.go
RTann: done

viswajithiii (approved):

PR #1779, opened 2022-05-19
ROX-10947: Empty name under spec.customresourcedefinitions.owned.resources

cchen666:
## Description

Change [CustomResource]_types.go to empty the "name" value spec.customresourcedefinitions.owned.resources otherwise this will cause https://bugzilla.redhat.com/show_bug.cgi?id=2082491

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

PR #1954, opened 2022-06-06
ROX-10947: Empty name under spec.customresourcedefinitions.owned.resources

porridge:
## Description

Change [CustomResource]_types.go to empty the "name" value spec.customresourcedefinitions.owned.resources otherwise this will cause https://bugzilla.redhat.com/show_bug.cgi?id=2082491

A copy of #1779 by @cchen666 to let CI pass

## Checklist
- [x] Check that the related resources links now work in the console

## Testing Performed

Relying on CI.

Also installed the operator, provisioned a central and made sure at least a few of the following links work OK:

Screenshot from 2022-06-06 13-37-55

### DIFF
--- a/operator/apis/platform/v1alpha1/central_types.go
+++ b/operator/apis/platform/v1alpha1/central_types.go
@@ -342,7 +342,7 @@ type CentralComponentStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
-//+operator-sdk:csv:customresourcedefinitions:resources={{Deployment,v1,central},{Deployment,v1,scanner},{Deployment,v1,scanner-db},{Secret,v1,central-htpasswd},{Service,v1,central-loadbalancer},{Route,v1,central}}
+//+operator-sdk:csv:customresourcedefinitions:resources={{Deployment,v1,""},{Secret,v1,""},{Service,v1,""},{Route,v1,""}}
 //+genclient
 
 // Central is the configuration template for the central services. This includes the API server, persistent storage,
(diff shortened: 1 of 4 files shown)


porridge:
FTR, the test job failures seem unrelated.

msugakov (approved):
I think it is worth to make a record in CHANGELOG.md saying something like:
"Fixed operator link to managed resources so that navigating to e.g. Deployments from the operator Resources page in OpenShift works correctly (earlier resulted in error page)."

msugakov:
[not-blocking-this-PR]
Question to @cchen666 and @camilamacedo86 what is the point/benefit of providing k8s kinds in `resources:` at all? What happens if we don't provide that? What happens if information is incorrect?

Docs [describe](https://docs.openshift.com/container-platform/4.10/operators/operator_sdk/osdk-generating-csvs.html#osdk-crds-owned_osdk-generating-csvs) `Resources` as (emphasis mine):

> Your CRDs own one or more types of Kubernetes objects. These are listed in the resources section to inform your users of the objects they might need to troubleshoot or how to connect to the application, such as the service or ingress rule that exposes a database.
> It is recommended to only list out the objects that are important to a human, **not an exhaustive list of everything you orchestrate**. For example, do not list config maps that store internal state that are not meant to be modified by a user.

This doesn't bring much clarity to me.

porridge:
I don't think this change meets the bar for inclusion in the changelog according to [the rules](https://github.com/stackrox/stackrox/blob/master/CHANGELOG.md#changelog).

msugakov:
When suggesting the change I was thinking of _obscure side effects_ in case old operator that mentions manged Deployments by the name gets upgraded to the new operator that mentions managing _some_ Deployments. The most obvious thing should be verified by the upgrade test which hopefully checks that operand deployments are really upgraded. However, there are more usage scenarios after upgrade and I was concerned if all will keep functioning correctly.

porridge:
@msugakov the thing changed here is purely UI related - these references are not used for anything other than creating links on "related resources" page in openshift console.

PR #1958, opened 2022-06-06
Add initial dashboard chart for Policy Violations by Category

dvail:
## Description

This adds the initial widget for [5: "Policy violations by category" ](https://docs.google.com/document/d/1z8kQOGRkZaN_RnaG4_kKjZHMC2O5Z-hSfWM4sqhyzuI/edit#heading=h.a0smrcssp8v6).  This is a feature incomplete widget but I wanted to get a PR out since this also contains code and patterns that will be used in other widgets.

This contains the chart and connection to back end data source using the default options, without filtering or Cypress tests. 

## Checklist

## Testing Performed

Visit the dashboard page at `main/dashboard-pf`, when the `ROX_SECURITY_METRICS_PHASE_ONE` flag is enabled.

By default the chart will render in "severity" order, with the most Critical violations at the top. Ties are broken by Important, Medium, and Low severity, in order.

image

Note that the top two bars have one critical severity, placing them at the top of the chart:
image

Test that the chart renders at max-width without distortion at various screen sizes:
image

Test that a skeleton is rendered when the widget is in a loading state:

Test that an error state is rendered when the data fetching promise rejects:

Test that clicking on a label in the chart axis links to the Violations page with the correct category applied, with Violations sorted by Severity.
image

Test that clicking on a bar in the chart links to the Violations page with the correct category applied, with Violations sorted by Severity.
image


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        return linkForViolationsCategory(category);
+                    }),
+                ]}
+            />
+        );
+    });
+
+    let cardContent: ReactNode;
+
+    if (error) {
```
pedrottimark, line 93: 1. It feels unexpected to compute `bars` above the conditional rendering. What do you think about factoring out `ViolationsByPolicyCategoryChart` to encapsulate computation of bars and rendering of `Chart` below with the assumed precondition not loading and not error?

2. A minor style point is rendering seems more common in order: loading, error, data.
    https://www.apollographql.com/docs/react/data/queries/#executing-a-query

pedrottimark: If an error is possible while loading, then error before loading with a comment why is clear enough.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        // TODO Auto-adjust padding based on screen size and/or max text length, if possible
+                        left: 180, // left padding is dependent on the length of the text on the left axis
+                        bottom: 75, // Adjusted to accommodate legend
+                    }}
+                    theme={patternflySeverityTheme}
+                >
+                    <ChartAxis
+                        tickLabelComponent={
+                            <LinkableChartLabel
+                                linkWith={({ text }) => linkForViolationsCategory(String(text))}
```
pedrottimark, line 129: If `linkWith` prop is a pure function, what do you think about factoring it out to module scope?

Integration tests have sometimes failed because element detached from the DOM, if inline callback function has different value on every render.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const widgetContainerResizeEntry = useResizeObserver(widgetContainer);
+    const { violationCounts, loading, error } = useViolationCounts('CATEGORY', ''); // TODO Implement query filtering
+
+    const countsBySeverity = getCountsBySeverity(violationCounts);
+    const bars = Object.entries(countsBySeverity).map(([severity, counts]) => {
+        const data = Object.entries(counts).map(([group, count]) => ({
+            name: severity,
+            x: group,
+            y: count,
+            label: `${severity}: ${count}`,
```
pedrottimark, line 72: Can you ask for recommendations about accessibility in charts at Developer Office Hours?

In this case, data is conceptually a table. How a chart can provide as close a user experience as possible.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        { name: 'Low' },
+                        { name: 'Medium' },
+                        { name: 'High' },
+                        { name: 'Critical' },
+                    ]}
+                    legendPosition="bottom"
+                    height={chartHeight}
+                    width={widgetContainerResizeEntry?.contentRect.width} // Victory defaults to 450
+                    padding={{
+                        // TODO Auto-adjust padding based on screen size and/or max text length, if possible
```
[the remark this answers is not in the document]
pedrottimark: This seems like another good question for PatternFly Developer office hours.

ui/apps/platform/src/Containers/Dashboard/PatternFly/WidgetErrorEmptyState.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export default function WidgetErrorEmptyState({
+    children,
+    title,
+    height,
+}: WidgetErrorEmptyStateProps) {
+    return (
+        <>
+            <Divider component="div" />
```
pedrottimark, line 30: Can you take a picture of an error state?

Because content is centered, I am trying to imagine whether the divider is necessary.

ui/apps/platform/src/Containers/Dashboard/PatternFly/hooks/useViolationCounts.ts
```diff
@@ -0,0 +1,64 @@
...
+                    pluckSeverityCount('CRITICAL_SEVERITY'),
+                    pluckSeverityCount('HIGH_SEVERITY'),
+                    pluckSeverityCount('MEDIUM_SEVERITY'),
+                    pluckSeverityCount('LOW_SEVERITY'),
+                ])
+                    // TODO Note: the backend does not appear to support the documented pagination queries, so we
+                    // need to limit the returned number of results client side
+                    .slice(0, limit)
+                    // We reverse here, because PF/Victory charts stack the bars from bottom->up
+                    .reverse();
```
pedrottimark, line 50: Two questions:
* Is `limit` for the number of policy categories? Looking from the outside in, it is not apparent that the widget has filtered the data in this way.
* To call `reverse` for Victory chart seems to couple data fetching with presentation implementation.

Maybe a question to discuss in team meeting how to balance convenience with coupling.

Not requesting a change, but suggesting to evaluate pro and con of an alternative:
* use specific hook to reduce boilerplate of `useEffect` with fetch function and 3 `useState` hooks
* call pure function to transform response in component that renders the chart (question where to declare)
* optional: reverse the return value also to make it visible in the component that renders the chart

ui/apps/platform/src/Containers/Dashboard/PatternFly/WidgetErrorEmptyState.tsx
dvail: Ack, I had these ready to go and never attached them to the PR. I've uploaded some screenshots of loading and error states.

I was torn for both cases on some style nits, that I can ping the UX team on:

For loading - whether or not to have padding around the grey skeleton component.
For error - whether or not to have the divider between the title.

ui/apps/platform/src/Containers/Dashboard/PatternFly/hooks/useViolationCounts.ts
dvail: Initially I was aiming to use `limit` to pass to the pagination part of the API since we only want the top 5 results for the widget. However there were two problems with this: the server doesn't appear to support the pagination API, and more importantly, we need to do the sorting client side since the sort method is a bit more complex that what the backend can do.

I think it definitely makes sense to move the `limit` and `reverse` calls out of here since they are coupled to the presentation. Actually now that I think of it the sorting functionality should be moved out as well. The next iteration is going to allow the user to change the way sorting is applied, so I'll retain the server response in the hook state and run pure functions in the component for presentation changes.

dvail: For a boilerplate reducing hook that wraps our fetch calls, I'm all for that idea. I know there was some discussion a while back regarding bringing in react-query, or pulling [this code](https://github.com/stackrox/infra/blob/master/ui/src/client/useApiQuery.ts) from stackrox/infra.

I was thinking I would refactor this as a follow up, since we'll want to have some team discussions on what approach works best for everyone and possibly draft up a design doc sometime after Saif gets back.

ui/apps/platform/src/Containers/Dashboard/PatternFly/WidgetErrorEmptyState.tsx
pedrottimark: One other thought about error state. Unlike a table or form which displays nothing and offers few if any actions when request fails, for dashboard with several widgets which have only view actions (links, filtering, sorting) can you ask about pro and con of danger versus warning variant?

pedrottimark (approved):
Early progress with one widget to illustrate pattern and reusable parts is wise way to work.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/WidgetCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        );
+    } else {
+        cardContent = children;
+    }
+
+    return (
+        <Card>
+            <Title headingLevel="h2" className="pf-u-p-md">
+                {title}
+            </Title>
```
vjwilson, line 35: I can see the value in having this wrapper for the loading/skeleton and errors states, but I don't think we should assume that every widget will only have a title. There could be subtitles, calls-to-action, etc. (I already know of a few places where we have already done this, or will do it when we migrate some of the *-Management sections to PF.

Let's leave it to the caller to set all of the "happy-path" content, including the header.

ui/apps/platform/src/utils/chartUtils.ts
```diff
@@ -0,0 +1,56 @@
+import { History } from 'react-router-dom';
+import { getTheme, ChartThemeColor } from '@patternfly/react-charts';
+import { EventCallbackInterface, EventPropTypeInterface } from 'victory-core';
+
+const severityColorScale = [
+    'var(--pf-chart-color-black-300)',
+    'var(--pf-chart-color-gold-300)',
+    'var(--pf-chart-color-orange-300)',
+    'var(--pf-chart-color-red-100)',
+];
```
vjwilson, line 10: Let's add a TODO to lift these colors into a global constant list in our app.

cc: @alanonthegit

vjwilson (approved):
Implementation looks pretty good. Left a few high-level comments.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/WidgetCard.tsx
dvail: Good idea, definitely easier to move the <Title> out and pass as a child element now instead of refactoring _n_ widgets later. I'll push this fix up.

alanonthegit:
@dvail @vjwilson Looking at this widget with real data (and this will vary per customer obviously), I'm noticing the click targets for network tools are way too small. It makes me think that clicking on the legend boxes to filter by severity is definitely necessary to help normalize the values. That alone may be sufficient.

Or maybe we decide to focus only on critical and high, changing the title to "Policy categories with most severe violations". Just something the team could consider.

ui/apps/platform/src/utils/chartUtils.ts
vjwilson: > @vjwilson Related to your comment about the `@patternfly/patternfly` dependency. How do you feel about dropping the dependency and using the colors that are part of PF's global palette instead? I can see us reusing these colors in the future in contexts that aren't chart specific, and it would save us some kbs in the bundle.
> 
> The global palette matches the chart colors exactly, except for the gray, which is -1 red and +1 blue component compared to the charts version (pictured on the right below).
> 
> image

Ok with me. What do you think @alanonthegit  ?

PR #1959, opened 2022-06-06
OSCI: collect policies-diff before exit!

davdhacs:
Fix policy check to include artifacts if failure.

This is to fix my mistake putting the artifacts collection at the end matching the yaml layout. I noticed it when making sure I had the artifacts for the mitre check.


gavin-stackrox (approved):
LGTM

PR #1963, opened 2022-06-07
Apply schemas after upgrade

c-du:
## Description
This is the fix for the apply schema not working in case of bolt version does not exist.

## Checklist

## Testing Performed

CI and manual check the log

### DIFF
--- a/migrator/main.go
+++ b/migrator/main.go
@@ -74,7 +74,7 @@ func run() error {
 		var gormDB *gorm.DB
 		gormDB, err = postgreshelper.Load(conf)
 		if err != nil {
-			return errors.Wrapf(err, "failed to connect to postgres DB")
+			return errors.Wrap(err, "failed to connect to postgres DB")
 		}
 		pkgSchema.ApplyAllSchemas(context.Background(), gormDB)
 	}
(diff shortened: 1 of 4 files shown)


migrator/postgreshelper/load.go
```diff
@@ -22,31 +23,39 @@ const (
...
+		// Waits for central-db ready with retries
+		err = retry.WithRetry(func() error {
+			gormDB, err = gorm.Open(postgres.Open(source), &gorm.Config{
+				NamingStrategy:  pgutils.NamingStrategy,
+				CreateBatchSize: 1000})
+			return err
+		}, retry.Tries(postgresConnectionRetries), retry.BetweenAttempts(func(attempt int) {
+			time.Sleep(postgresConnectRetryInterval)
+		}), retry.OnFailedAttempts(func(err error) {
+			log.WriteToStderrf("fail to connect to central database: %v", err)
```
connorgorman, line 53: fail -> failed

migrator/main.go
```diff
@@ -108,7 +117,7 @@ func upgrade(conf *config.Config) error {
 	if features.PostgresDatastore.Enabled() {
 		gormDB, err = postgreshelper.Load(conf)
 		if err != nil {
-			return errors.Wrapf(err, "Failed to connect to postgres DB with Gorm %v", err)
+			return errors.Wrapf(err, "Failed to connect to postgres DB %v", err)
```
connorgorman, line 120: same comment as above, no need to add the err as %v when using wrap

connorgorman (approved):
Small nits and changes, then ship. Thanks!

PR #1964, opened 2022-06-07
ROX-11059: Use environment.Logger instead of fmt.*Print within roxctl.

dhaus67:
## Description

This is the second PR in a series of PRs to finish the migration of roxctl to use `Environment`.

For more context of the overall series, see #1933.

This PR will refactor all usages of `fmt.*Print*` to use `common.Logger` instead. This will ensure that the correct output stream will be used as well as the correct colorful output, ensuring we have a consistent printing behavior throughout roxctl.

Which PRs will follow after this:
- Move remaining errors from `errros.New*/fmt.Errorf` to `errox sentinel errors`.
- Enabled `forbidigo` to verify neither `fmt.*Print/os.Stdout/os.Stderr` is used within roxctl.

### DIFF
--- a/roxctl/common/environment-impl.go
+++ b/roxctl/common/environment-impl.go
@@ -2,9 +2,12 @@ package common
 
 import (
 	"io"
+	"os"
+	"testing"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/sync"
 	"github.com/stackrox/rox/roxctl/common/flags"
 	"github.com/stackrox/rox/roxctl/common/printer"
 	"google.golang.org/grpc"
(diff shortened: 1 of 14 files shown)


roxctl/common/environment-impl.go
```diff
@@ -16,15 +19,42 @@ type cliEnvironmentImpl struct {
...
+func NewTestCLIEnvironment(_ *testing.T, io IO, c printer.ColorfulPrinter) Environment {
 	return &cliEnvironmentImpl{
 		io:              io,
 		colorfulPrinter: c,
 		logger:          NewLogger(io, c),
 	}
 }
 
+// CLIEnvironment creates a new default CLI environment.
+func CLIEnvironment() Environment {
```
janisz, line 38: I don't like this approach as it makes our code not testable :(
We should have a single instance of env and pass it everywhere where needed

roxctl/common/zipdownload/download_zip.go
```diff
@@ -23,6 +23,10 @@ const (
 	inMemFileSizeThreshold = 1 << 20 // 1MB
 )
 
+var (
+	log = common.CLIEnvironment().Logger()
```
janisz, line 27: This should be injected

janisz (approved):
We discussed online and decided to merge it as is and fix it in next PR after we untangle environment, io and log

PR #1969, opened 2022-06-07
ROX-10891: GitHub Actions for upstream release automation

0x656b694d:
## Description

The initial PR on upstream automation. Covered some GitHub related activities, such as creating the release branch, milestones, cherry-picking and such.

Happy-path:
1. Start Release (manual action on GitHub);
2. Close RC milestone (the first milestone is created automatically);
3. Check CI runs (not automated);
4. Edit the published pre-release (RC) to be the final release with the suggested command.

The idea is to automate the steps which can be automated, suggest what to do next and notify about taken steps.
Workflow steps have to be restartable with no unexpected effects.

Steps summary is printed under the workflow run on GitHub for the release managers and the notifications for wider audience are sent to Slack.

*See ROX-10891 ticket for the link to the more detailed documentation*.

## Notes

The Slack channel ID points to my personal chat.
The long-running cluster creation is requested in the workflow, but no further action is taken.
Further work is required on the workflows to improve user communication and test cluster creation.

## Testing Performed

The Dry-run option implemented with which the jobs would only do local changes and print the progress without affecting the repos or GitHub objects.
Manual tests performed on https://github.com/stackrox/test-gh-actions repository.

### DIFF
--- a/.github/workflows/create-cluster.yml
+++ b/.github/workflows/create-cluster.yml
@@ -1,0 +1,60 @@
+name: Create Test Cluster
+on:
+  workflow_call:
+    inputs:
+      flavor:
+        description: Flavor
+        type: string
+        required: true
+      name:
+        description: Cluster name
+        type: string
+        required: true
+      lifespan:
+        description: Cluster lifespan
+        type: string
+        default: 48h
+        required: false
+      args:
+        description: Comma separated flavor arguments. Ex. nodes=5,main-image=main:tag
+        type: string
+        required: false
+        # Spaces in arguments are not supported.
+        default: ""
+      wait:
+        description: Wait or not for the cluster readiness
+        type: boolean
+        required: false
+        default: false
+jobs:
+  infra:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download infractl
+        run: |
+          curl -sL https://infra.rox.systems/v1/cli/linux/amd64/upgrade \
+          | jq -r ".result.fileChunk" \
+          | base64 -d \
+          > infractl
+          chmod +x infractl
+      - name: Create Cluster
+        env:
+          INFRA_TOKEN: ${{secrets.INFRA_TOKEN}}
+        run: |
+          NAME="${{inputs.name}}"
+          WAIT=""
+          if [ "${{inputs.wait}}" = "true" ]
+          then
+            WAIT="--wait";
+            echo "::warning::The job will wait for the cluster creation."
+          fi
+
+          eval ./infractl create "${{inputs.flavor}}" "${NAME//./-}" \
+          --lifespan "${{inputs.lifespan}}" \
+          $WAIT \
+          $(python -c "import sys; args=sys.argv
+          args='' if len(args) <= 1 else ' --arg '.join(args[1].split(','))
+          if args: print('--arg', args)" \
+          "${{inputs.args}}")
+
+          ./infractl get "${NAME//./-}" >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -1,0 +1,446 @@
+name: Cut RC
+on:
+  milestone:
+    types:
+      - closed
+
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Full RC number (A.B.C-rc.D)
+        required: true
+        default: 0.0.0-rc.1
+        type: string
+      dry-run:
+        description: Dry-run
+        required: false
+        default: true
+        type: boolean
+
+env:
+  main_branch: master
+  docs_repository: openshift/openshift-docs
+  jira_projects: ROX, RS, RTOOLS
+  slack_channel: U02MPRJDANM
+
+jobs:
+  check-jira:
+    name: Check Jira tickets for release ${{inputs.version}}
+    runs-on: ubuntu-latest
+    env:
+      JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
+    steps:
+      - name: Query JIRA
+        env:
+          JQL: >-
+            project IN (${{env.jira_projects}})
+            AND fixVersion = \"${{inputs.version}}\"
+            AND status != CLOSED
+            AND Component != Documentation AND type != Epic
+            order by assignee
+
+        run: |
+          GH_MD_FORMAT_LINE=$(cat << EOF
+          "* [\(.key)](https://issues.redhat.com/browse/\(.key)): **"+(.fields.assignee.displayName // "unassigned")+"** (\(.fields.status.name)) — _"+(.fields.summary | gsub (" +$";""))+"_"
+          EOF
+          )
+
+          ISSUES=$(curl -f -sSL --get --data-urlencode "jql=$JQL" \
+            -H "Authorization: Bearer $JIRA_TOKEN" \
+            -H "Accept: application/json" \
+            "https://issues.redhat.com/rest/api/2/search" \
+          | jq -r ".issues[] | $GH_MD_FORMAT_LINE" \
+          | sort)
+
+          if [ -n "$ISSUES" ]
+          then
+            cat << EOF >> $GITHUB_STEP_SUMMARY
+          The following Jira issues are still open for release ${{inputs.version}}:
+
+          $ISSUES
+
+          Please contact the assignees to clarify the status.
+          EOF
+            echo "::error::There are non-closed Jira issues for version ${{inputs.version}}."
+
+            curl -f -sSL --get --data-urlencode "jql=$JQL" \
+            -H "Authorization: Bearer $JIRA_TOKEN" \
+            -H "Accept: application/json" \
+            "https://issues.redhat.com/rest/api/2/search" \
+            | jq -r ".issues[] | .key" | while read -r KEY
+            do
+              curl -f -sSL -X POST \
+                -H "Authorization: Bearer $JIRA_TOKEN" \
+                --data '{"body": "Release ${{inputs.version}} is ongoing. Please update the status."}' \
+                -H "Content-Type: application/json" \
+                "https://issues.redhat.com/rest/api/2/issue/$KEY/comment"
+            done
+
+            exit 1
+          fi
+          echo "All issues for Jira release ${{inputs.version}} are closed." >> $GITHUB_STEP_SUMMARY
+
+  variables:
+    name: Setup variables
+    uses: stackrox/stackrox/.github/workflows/variables.yml@master
+    with:
+      version: ${{format('{0}{1}', github.event.milestone.title, github.event.inputs.version)}}
+
+  postpone-prs:
+    name: Postpone open PRs
+    needs: variables
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Check open PRs
+        id: check
+        run: |
+          PRs=$(gh pr list -s open \
+          --repo "${{github.repository}}" \
+          --search "milestone:${{needs.variables.outputs.milestone}}" \
+          --json number \
+          --jq length)
+          if [ "$PRs" -gt 0 ]
+          then
+            echo "::set-output name=open-issues::$PRs"
+          fi
+
+      - name: Create next milestone
+        if: github.event.inputs.dry-run == 'false' && steps.check.outputs.open-issues != ''
+        run: |
+          if ! http_code=$(gh api --silent -X POST \
+            "repos/${{github.repository}}/milestones" \
+            -f "title"="${{needs.variables.outputs.next-milestone}}" \
+            2>&1)
+          then
+            if grep "HTTP 422" <<< "$http_code"
+            then
+              echo "Milestone ${{needs.variables.outputs.next-milestone}} already exists." >> $GITHUB_STEP_SUMMARY
+            else
+              echo "::error::Couldn't create milestone ${{needs.variables.outputs.next-milestone}}: $http_code"
+              exit 1
+            fi
+          else
+            echo "Milestone ${{needs.variables.outputs.next-milestone}} has been created." >> $GITHUB_STEP_SUMMARY
+          fi
+
+      - name: Move open PRs
+        run: |
+          for PR in $(gh pr list -s open \
+            --repo "${{github.repository}}" \
+            --search "milestone:${{needs.variables.outputs.milestone}}" \
+            --json number \
+            --jq ".[] | .number")
+          do
+            [ "${{github.event.inputs.dry-run}}" = "false" ] && \
+            gh pr edit $PR \
+              --milestone "${{needs.variables.outputs.next-milestone}}" \
+              --repo "${{github.repository}}"
+            echo "PR $PR has been moved to milestone ${{needs.variables.outputs.next-milestone}}." >> $GITHUB_STEP_SUMMARY
+          done
+
+  check-merged-prs:
+    name: Cherry-pick from PRs
+    needs: [variables, postpone-prs]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{needs.variables.outputs.branch}}
+      - name: Initialize mandatory git config
+        run: |
+          git config user.name "${{github.event.sender.login}}"
+          git config user.email noreply@github.com
+      - run: |
+          # Could be optimized: done only if there are closed PRs to cherry-pick
+          git fetch origin ${{env.main_branch}}:${{env.main_branch}} --unshallow
+      - env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        name: Cherry-pick commits from the main branch
+        id: cherry-pick
+        run: |
+          SLACK_MESSAGE_FILE=$(mktemp)
+          EMPTY=""
+          gh pr list -s closed \
+            --search "milestone:${{needs.variables.outputs.milestone}}" \
+            --base "${{env.main_branch}}" \
+            --json number,url,mergeCommit,author,title \
+            --jq '.[] | "\(.number)\t\(.url)\t\(.mergeCommit.oid)\t\(.author.login)\t\(.title)"' \
+          | while read -r PR_COMMIT
+          do
+            IFS=$'\t' read -r PR URL COMMIT AUTHOR TITLE <<< $PR_COMMIT
+            # Skip commits merged before branching.
+            if git merge-base --is-ancestor "$COMMIT" HEAD
+            then
+              continue
+            fi
+
+            FORK_POINT=$(diff -u \
+              <(git rev-list --first-parent ${{needs.variables.outputs.branch}}) \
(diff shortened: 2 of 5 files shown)


.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+      - name: Check open PRs
+        id: check
+        run: |
+          PRs=$(gh pr list -s open \
+          --repo "${{github.repository}}" \
+          --search "milestone:${{needs.variables.outputs.milestone}}" \
+          --json number \
+          --jq length)
+          if [ "$PRs" -gt 0 ]
+          then
```
vikin91, line 105: I usually prefer the condensed version of the if:

```suggestion
          if [ "$PRs" -gt 0 ]; then
```

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+            echo "Milestone ${{needs.variables.outputs.next-milestone}} has been created." >> $GITHUB_STEP_SUMMARY
+          fi
+
+      - name: Move open PRs
+        run: |
+          for PR in $(gh pr list -s open \
+            --repo "${{github.repository}}" \
+            --search "milestone:${{needs.variables.outputs.milestone}}" \
+            --json number \
+            --jq ".[] | .number")
```
vikin91, line 134: This may work fine, but it has some drawbacks.
I would suggest to capture the command output into an array and then iterate over that array in the for loop. This would allow us catch potential mistakes and handle properly edge-cases (e.g., empty values).

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+          do
+            [ "${{github.event.inputs.dry-run}}" != "true" ] && \
```
vikin91, line 136: Why don't we move this check to the top of this block - outside of the loop?

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+        id: cherry-pick
+        run: |
+          SLACK_MESSAGE_FILE=$(mktemp)
+          EMPTY=""
+          gh pr list -s closed \
+            --search "milestone:${{needs.variables.outputs.milestone}}" \
+            --base "${{env.main_branch}}" \
+            --json number,url,mergeCommit,author,title \
+            --jq '.[] | "\(.number)\t\(.url)\t\(.mergeCommit.oid)\t\(.author.login)\t\(.title)"' \
+          | while read -r PR_COMMIT
```
vikin91, line 171: Again separating the command from the loop would allow us to do some debugging in case the command returns something unexpected.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+          git push
+          if [ -n "$EMPTY" ]; then
+            echo "Cherry-picked commits have been pushed to the release branch." >> $GITHUB_STEP_SUMMARY
+          fi
+
+          # Replace % with %25 and join lines with %0A:
+          FAILED="$(sed ':a; N; $!ba; s/%/%25/g; s/\n/%0A/g' $SLACK_MESSAGE_FILE)"
+          rm -f "$SLACK_MESSAGE_FILE"
+
+          if [ -n "$FAILED" ]
```
vikin91, line 217: I do not understand how the failed sed on a slack-message file is used to determine whether cherry-picking has succeeded. Could we think of providing some more robust check?

vikin91 (changes requested):
This is a great piece of great work - thank you once again for starting this!

So far I have just went though the code (I still want to try it out), but I am sharing my feedback already now.

I think the code in the current shape needs some cleaning and reorganization. I suggest the following as the minimum:

1. Extract all bash code into functions and place then in a separate file that can be sourced.
2. Apply a dose of defensive programing (verify each input and output for correctness) with chatty logging, so that the debugging is easy
3. Source the bash file in each action and use the functions - this will keep the yaml size minimal
4. If possible and reasonable, cover the library with bats tests.

As this may be a relatively large effort, I am happy to help you with that :)

.github/workflows/cut-rc.yml
0x656b694d: SLACK_MESSAGE_FILE file is basically a list of failed PRs to be reported to SLACK. /* I could probably use bash arrays, but I avoid advanced bash features as you could have noticed */
So if it is empty, there were no merge failures. I considered it to be a good check.

0x656b694d:
> 1. Extract all bash code into functions and place then in a separate file that can be sourced.

This would require `git clone` of the repository, which is often an overkill for a couple of commands. I'd suggest to move only complex code to scripts (like deploying long-running cluster) and keep fewliners inlined.

> 2. Apply a dose of defensive programing (verify each input and output for correctness) with chatty logging, so that the debugging is easy

Agree.

> 3. Source the bash file in each action and use the functions - this will keep the yaml size minimal

Don't fully agree.

> 4. If possible and reasonable, cover the library with bats tests.

All the actions have side effects, not sure we want to mock all the external services to test how we call their APIs. But I'm open for reasonable testing suggestions.

.github/workflows/create-cluster.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Create Cluster
+        env:
+          INFRA_TOKEN: ${{secrets.INFRA_TOKEN}}
+        run: |
+          NAME="${{inputs.name}}"
+          WAIT=""
+          if [ "${{inputs.wait}}" = "true" ]
+          then
+            WAIT="--wait";
+            echo "::warning::The job will wait for the cluster creation."
```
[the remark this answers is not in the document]
porridge: I'd also suggest leaving at least the ` to finish` part for clarity.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,547 @@
...
+        description: Dry-run
+        required: false
+        default: true
+        type: boolean
+
+env:
+  main_branch: master
+  docs_repository: openshift/openshift-docs
+  jira_projects: ROX, RS, RTOOLS
+  slack_channel: C03D78P4RFF
```
porridge, line 24: How about a comment which channel this is?
Also, should we change this to something visible to others now?

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,547 @@
...
+
+              # Create PR to the release branch:
+              gh create pr --base "${{needs.variables.outputs.branch}}" --fill \\
+                --milestone "${{needs.variables.outputs.milestone}}"
+
+          EOF
+            exit 1
+          fi
+      - name: Post to Slack
+        if: always() && steps.cherry-pick.outputs.bad-cherries != ''
```
porridge, line 262: `if: always()`?
What does it mean 🤯 ?

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,547 @@
...
+      - name: Initialize mandatory git config
+        run: |
+          git config user.name "${{github.event.sender.login}}"
+          git config user.email noreply@github.com
+      - name: Update docs submodule
+        run: |
+          git submodule update --init --remote -- docs/content
+          git add docs/content
+          if ! git diff-index --quiet HEAD; then
+            git commit -m "Update docs submodule"
```
porridge, line 294:
```suggestion
            git commit -m "Update docs submodule for milestone ${{needs.variables.outputs.milestone}}"
```
? (no idea whether that would actually work)

.github/workflows/create-cluster.yml
```diff
@@ -0,0 +1,60 @@
...
+      lifespan:
+        description: Cluster lifespan
+        type: string
+        default: 48h
+        required: false
+      args:
+        description: Comma separated flavor arguments. Ex. nodes=5,main-image=main:tag
+        type: string
+        required: false
+        # Spaces in arguments are not supported.
```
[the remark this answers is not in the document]
0x656b694d: Now supported. Comment removed.

.github/workflows/create-cluster.yml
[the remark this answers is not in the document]
0x656b694d: Ok!

.github/workflows/create-cluster.yml
```diff
@@ -0,0 +1,60 @@
...
+        default: false
+jobs:
+  infra:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download infractl
+        run: |
+          curl -sL https://infra.rox.systems/v1/cli/linux/amd64/upgrade \
+          | jq -r ".result.fileChunk" \
+          | base64 -d \
```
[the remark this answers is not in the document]
0x656b694d: Unfortunately in this case `jq` fails to decode the output.

.github/workflows/create-cluster.yml
```diff
@@ -0,0 +1,60 @@
...
+        type: boolean
+        required: false
+        default: false
+jobs:
+  infra:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download infractl
+        run: |
+          curl -sL https://infra.rox.systems/v1/cli/linux/amd64/upgrade \
```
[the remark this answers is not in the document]
0x656b694d: `set -e` is implicitly set, I'll see if `ou` would help.
Will add the missing `--fail` here, thanks.

.github/workflows/cut-rc.yml
0x656b694d: This is now the `epic-release-improvements` channel. We'll change it to `eng-release` eventually. Well, I could add that in a comment.

.github/workflows/cut-rc.yml
0x656b694d: github actions magic, this is required to not skip the step after previous step failure. Could probably be changed to `failure()`.

0x656b694d:
@porridge, thanks for the review!
With GitHub actions we've got the following possibilities for code reuse:
* Call external action (ex. in the PR: checkout code, slack, install kubectl);
* Call another workflow (ex. in the PR: create-clusters.yml, variables.yml);
* Call local script (ex. in the PR: ./deploy.sh)

Please highlight the code which could have been extracted for reuse so we could move it to an appropriate module.

porridge:
> Please highlight the code which could have been extracted for reuse so we could move it to an appropriate module.

I think anything longer than, say, 5 lines is a candidate for extraction 🤔

0x656b694d:
> I think anything longer than, say, 5 lines is a candidate for extraction 

When extracting the code, we'll need to consider the (time) cost of repository cloning for calling a script from the repo, and/or instantiating another job runner. Working on a branch we'll probably want to call scripts from the main branch (as from where the workflow is running), which may also add delays and complexity.
That may add seconds to the procedure that a person would monitor, which may not always be beneficial for user experience.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,550 @@
...
+
+  create-k8s-cluster:
+    name: Create k8s cluster
+    needs: [variables, cut-rc]
+    if: github.event.inputs.dry-run != 'true'
+    uses: ./github/workflows/create-cluster.yml
+    with:
+      flavor: qa-demo
+      name: qa-k8s-${{needs.variables.outputs.milestone}}
+      args: main-image=main:${{needs.variables.outputs.milestone}}
```
0x656b694d, line 375:
```suggestion
      args: main-image=docker.io/stackrox/main:${{needs.variables.outputs.milestone}}
```

0x656b694d (commented):
Set the full main image name.

vikin91:
Sorry, I didn't make to give it another look :( I will do it as soon as I feel a bit better

sbostick (commented):
### Overall impression and cursory analysis

* Nicely captures the steps and dependencies
* Clear separation of inputs, outputs, secrets
* Readable and maintainable
* Deps (cli): curl, gh, infractl, git
* Deps (gha): actions/checkout@v3, slackapi/slack-github-action@v1.19.0, azure/setup-kubectl@v2.0
* Runtime: ubuntu-latest, bash

sbostick:
### Suggestions

Add a README.md to the effect:

Implements ACS upstream release branching model including:
* targeted issue state verification
* pr creation
* test env provisioning

References:
* [Release Mangement - upstream](https://docs.engineering.redhat.com/display/StackRox/How+to+manage+a+release)
* [Release Management - downstream](https://docs.engineering.redhat.com/display/StackRox/Managing+a+Release+on+Red+Hat+Infrastructure)
* [Release Process](https://docs.engineering.redhat.com/display/StackRox/Managing+a+Release+on+Red+Hat+Infrastructure)
* [Release Tracker](https://docs.google.com/spreadsheets/d/19HaplE_vNOUFN2wrkQRf-hNXNspajaOA2mgXMdjyRPs/edit#gid=769572822)

sbostick:
### Possible improvements

I agree with Piotr on script reuse if possible, and generally, factoring out shell code from yaml. However in this case I found it quite readable and self-documenting as is, so no strong preference from me. I'd say the current implementation is a good evolution of the long-form instructions.

So factoring out into scripts could improve portability and testability, but I'm not sure I could justify the ROI. And based on what you captured into code already, if one was motivated to reimplement, it would be a fairly straightfoward translation. For example, I'm most comfortable in Python and would prefer being able to mock out external API calls and add tests for the actual logic and corner cases related to branching/pr/milestone/issue manipulation, empty search results, etc. Another option might be making an e2e test of the release model, essentially treating it as a strategy that can be reused. But designing and testing a more generalized, reusable solution is sort of outside the scope. At least could defer larger refactoring and/or generalization efforts to separate initiatives with their own clear motivations and goals.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,598 @@
...
+        description: Dry-run
+        required: false
+        default: true
+        type: boolean
+
+env:
+  main_branch: master
+  docs_repository: openshift/openshift-docs
+  jira_projects: ROX, RS, RTOOLS
+  slack_channel: C03KSV3N6N8 #test-release-automation
```
JoukoVirtanen, line 24: I think we already have this in CircleCI config in plain text. Is it possible to put this in a secret?

0x656b694d: It is possible. Benefit: single source for all workflows. Disadvantage: only admins have access. As the ID is not necessarily a sensitive information, we might want to keep during testing phase.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,598 @@
...
+
+          ./deploy/k8s/sensor.sh
+
+          kubectl -n stackrox set env deploy/sensor MUTEX_WATCHDOG_TIMEOUT_SECS=0
+          kubectl -n stackrox set env deploy/sensor ROX_FAKE_KUBERNETES_WORKLOAD=long-running
+          kubectl -n stackrox patch deploy/sensor -p '{"spec":{"template":{"spec":{"containers":[{"name":"sensor","resources":{"requests":{"memory":"3Gi","cpu":"2"},"limits":{"memory":"12Gi","cpu":"4"}}}]}}}}'
+
+          kubectl -n stackrox set env deploy/central MUTEX_WATCHDOG_TIMEOUT_SECS=0
+          kubectl -n stackrox patch deploy/central -p '{"spec":{"template":{"spec":{"containers":[{"name":"central","resources":{"requests":{"memory":"3Gi","cpu":"2"},"limits":{"memory":"12Gi","cpu":"4"}}}]}}}}'
+          echo "Long-running GKE cluster has been patched" >> $GITHUB_STEP_SUMMARY
```
JoukoVirtanen, line 556: When I created my script to create the long-running cluster I had to make some changes to have it wait for things. When copying and pasting these commands from the documentation it works, because there is time for things to finish, but if you put everything in a single scripts you need to put checks to make sure that the previous step completed. I think you will eventually replace this with my script, so it is fine.

JoukoVirtanen:
This is great work! Thanks for doing this! I thought that automating everything that you have automated here would take a long time and be done in many PRs, but you are doing it in one. My main criticism is that you have shell in the config files for things that we already have script files for. This is an issue that I think can be addressed in future PRs.

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,598 @@
...
+      JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
+    steps:
+      - name: Query JIRA
+        env:
+          JQL: >-
+            project IN (${{env.jira_projects}})
+            AND fixVersion = \"${{needs.variables.outputs.release-patch}}\"
+            AND status != CLOSED
+            AND Component != Documentation AND type != Epic
+            order by assignee
```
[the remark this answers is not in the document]
0x656b694d: Using script would require code checkout here, which is acceptable as there will be no confusion with the release branch. This code also post comments in the found Jira tickets to grab attention of the assignees.

rukletsov (approved):
I think the suggestion to pull some part of the workflows into scripts is a reasonable one. There is an interleaved dependency on CircleCI, ideally, we should simplify them to the following: run desired CI checks and trigger a release after. I believe this (i.e. simplification) will affect the workflows here.

So let's maybe start with what we have, polish it during the upcoming release, and factor out _after_ we shuffle steps between CI and release for simplicity.

sbostick (approved):

.github/workflows/cut-rc.yml
vikin91: I do not understand your answer. If the `"${{github.event.inputs.dry-run}}"` is set to `true`, then nothing will be run from the lines 144-150. What is the point in checking `[ "${{github.event.inputs.dry-run}}" != "true" ]` multiple times when once would be enough?

.github/workflows/cut-rc.yml
```diff
@@ -0,0 +1,503 @@
...
+            --jq ".[] | .number")
+          do
+            [ "${{github.event.inputs.dry-run}}" != "true" ] && \
+            gh pr edit $PR \
+              --milestone "${{needs.variables.outputs.next-milestone}}" \
+              --repo "${{github.repository}}"
+            echo "PR $PR has been moved to milestone ${{needs.variables.outputs.next-milestone}}." >> $GITHUB_STEP_SUMMARY
+          done
+
+  check-merged-prs:
```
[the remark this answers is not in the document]
vikin91: I think, we should extract it on the next occasion. This should apply to all the bash code used in the yamls.

vikin91 (approved):
It looks good, so I am happy to accept it the way it is.

However, I find it important to make the bash code easier to manage. I think we should follow up on the idea of an external script library as one of the next things to work on.

.github/workflows/cut-rc.yml
0x656b694d: I don't understand your question :). The current check disables only the `gh pr edit` command. The following `echo` need to be always printed in the loop. So I need this check inside the loop.

.github/workflows/create-cluster.yml
```diff
@@ -0,0 +1,71 @@
...
+        type: boolean
+        required: false
+        default: false
+jobs:
+  infra:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download infractl
+        run: |
+          curl --fail -sL https://infra.rox.systems/v1/cli/linux/amd64/upgrade \
```
janisz, line 34: How about adding `-S`?

>-S, --show-error
 >      When used with -s it makes curl show an error message if it fails.

PR #1961, opened 2022-06-06
Delete fixture from Configuration Management dashboard integration test

pedrottimark:
### Problem

logimbue-data.json file always contains at least one exception even when tests pass, which slows down investigation of integration test failures.

Tests added in https://github.com/stackrox/rox/pull/2929 and most recently edited in https://github.com/stackrox/rox/pull/9371

### Analysis

```json
"url": "https://central-lb/main/configmanagement/subjects/MDZkMWMzYjktOGQ5Ny00OTZkLTg0ZGItMGM1Nzg5MTBiNzFi%3Ac3lzdGVtOm1hc3RlcnM%3D"
```

```json
          "type": "TypeError",
          "value": "Cannot read properties of undefined (reading 'subject')",
```

```json
      "lastAction": { "type": "serverError/RECORD_SUCCESS" },
      "state": {
        "router": {
          "location": {
            "pathname": "/main/configmanagement/subjects/MDZkMWMzYjktOGQ5Ny00OTZkLTg0ZGItMGM1Nzg5MTBiNzFi%3Ac3lzdGVtOm1hc3RlcnM%3D",
            "search": "",
            "hash": "",
            "key": "y2nbu4",
            "query": {}
          },
          "action": "PUSH"
        },
```

Only one test related to Users & Groups (also known as subjects) has a click in svg.

It has **an unneeded** fixture for /api/graphql?opname=usersWithClusterAdminRoles

It does not have corresponding fixtures
* for /api/graphql?opname=getSubjectName
* for /api/graphql?opname=getSubject

```json
        {
          "timestamp": 1653326296.904,
          "category": "ui.click",
          "message": "div \u003e div.rv-xy-plot. \u003e svg \u003e g \u003e rect"
        },
        {
          "timestamp": 1653326296.904,
          "category": "navigation",
          "data": {
            "to": "/main/configmanagement/subjects/MDZkMWMzYjktOGQ5Ny00OTZkLTg0ZGItMGM1Nzg5MTBiNzFi%3Ac3lzdGVtOm1hc3RlcnM%3D",
            "from": "/main/configmanagement"
          }
        },

        {
          "timestamp": 1653326297.026,
          "type": "http",
          "category": "xhr",
          "data": {
            "method": "POST",
            "url": "/api/graphql?opname=getSubject",
            "status_code": 200
          }
        },
        {
          "timestamp": 1653326297.026,
          "category": "redux-action",
          "message": "serverError/RECORD_SUCCESS"
        },
        {
[...]

PR #1971, opened 2022-06-07
ROX-11246: Remove redux state from Raven logimbue POST calls

pedrottimark:
### Problem

Need to reduce size of payload in /api/logimbue request.

### Solution

https://www.npmjs.com/package/raven-for-redux#statetransformer-function

> In some cases your state may be extremely large, or contain sensitive data. In those cases, you may want to transform your state before sending it to Sentry. This function allows you to do so. It is passed the current state object, and should return a serializable value.

> Be careful not to mutate your state within this function.

```js
createRavenMiddleware(Raven, { stateTransformer: () => null });
```

## Checklist

## Testing Performed

* 1 exception: same error as in #1961:
    Visit /main/configmanagement/subjects/id-does-not-exist
    Copy payload of /api/logimbue request without (after change) and with (before change) state.

* 3 exceptions: 3.68.2-rc.3 upstream build
    https://app.circleci.com/pipelines/github/stackrox/stackrox/12424/workflows/54cfd89f-1a51-45f4-891d-97c313c6a860/jobs/586272
    To simulate without state, run small script which has replacer `"state:{…}"` with `"state":null` in `JSON.stringify` call.

Unformatted file size

| exceptions | without |   with |   delta | delta / exceptions |
|       ---: |    ---: |   ---: |    ---: |               ---: |
|          1 |    7595 |  37429 |  -29834 |             -29834 |
|          3 |   40010 | 168989 | -128979 |             -42993 |

Formatted file lines (because I format the file when I investigate integration tests failures)

| exceptions | without |   with |   delta | delta / exceptions |
|       ---: |    ---: |   ---: |    ---: |               ---: |
|          1 |     434 |   2019 |   -1585 |              -1585 |
|          3 |    1864 |   8011 |   -6147 |              -2049 |

### DIFF
--- a/ui/apps/platform/src/store/configureStore.js
+++ b/ui/apps/platform/src/store/configureStore.js
@@ -15,7 +15,8 @@ const sagaMiddleware = createSagaMiddleware({
     onError: (error) => Raven.captureException(error),
 });
 
-const ravenMiddleware = createRavenMiddleware(Raven);
+// Omit Redux state to reduce size of payload in /api/logimbue request.
+const ravenMiddleware = createRavenMiddleware(Raven, { stateTransformer: () => null });
 
 export default function configureStore(initialState = {}, history) {
     const middlewares = [sagaMiddleware, routerMiddleware(history), ravenMiddleware];


vjwilson (approved):
/lgtm

PR #1975, opened 2022-06-07
[ROX-11011] Add new service for suppressing node vulns

md2119:
## Description

- New datastore for node vulns
- New service for node vulns
- Refactor and fixes to image vulns APIs for backward compatibility (when cve name and ID were the same)
- Changelog

## Checklist

## Testing Performed
unit


CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - These endpoints are deprecated as license files are not required to run the platform
 - `firstNodeOccurrence` field of `storage.Node` object, which is in the response of Node endpoints, has been removed.
 - `vulns` fields of `storage.Node` object, which is in the response payload of `v1/nodes` is deprecated and will be removed in 72.0.
```
connorgorman, line 24: i know it was last PR, but I think we should probably just say removed in the future unless we can orient them without postgres?

md2119: Makes sense. It is better to not bind postgres related deprecation to a release RN.

central/cve/image/datastore/datastore_impl.go
```diff
@@ -116,71 +112,53 @@ func (ds *datastoreImpl) Suppress(ctx context.Context, start *types.Timestamp, d
...
-	// to determine if the cve is image cve was not added.
+func (ds *datastoreImpl) Unsuppress(ctx context.Context, cves ...string) error {
 	if ok, err := vulnRequesterOrApproverSAC.WriteAllowedToAll(ctx); err != nil {
 		return err
 	} else if !ok {
 		return sac.ErrResourceAccessDenied
 	}
 
-	cves, missing, err := ds.storage.GetMany(ctx, ids)
+	vulns, err := ds.searcher.SearchRawCVEs(ctx, pkgSearch.NewQueryBuilder().AddExactMatches(pkgSearch.CVE, cves...).ProtoQuery())
```
connorgorman, line 140: Can we just run a GetMany instead of a search?

central/cve/image/datastore/datastore_impl.go
```diff
@@ -116,71 +112,53 @@ func (ds *datastoreImpl) Suppress(ctx context.Context, start *types.Timestamp, d
 		return err
 	}
 
-	cves, missing, err := ds.storage.GetMany(ctx, ids)
+	vulns, err := ds.searcher.SearchRawCVEs(ctx, pkgSearch.NewQueryBuilder().AddExactMatches(pkgSearch.CVE, cves...).ProtoQuery())
```
connorgorman, line 115: Same here, let's just run a GetMany and if some are missing then some are missing?

connorgorman: Or does is the CVE not the ID anymore...

central/cve/node/datastore/datastore_impl.go
```diff
@@ -0,0 +1,201 @@
...
+	query := pkgSearch.NewQueryBuilder().AddBools(pkgSearch.CVESuppressed, true).ProtoQuery()
+	suppressedCVEs, err := ds.searcher.SearchRawCVEs(accessAllCtx, query)
+	if err != nil {
+		return errors.Wrap(err, "searching suppress CVEs")
+	}
+
+	ds.cveSuppressionLock.Lock()
+	defer ds.cveSuppressionLock.Unlock()
+	for _, cve := range suppressedCVEs {
+		ds.cveSuppressionCache[cve.GetId()] = common.SuppressionCacheEntry{
```
connorgorman, line 48: Ooo, are we having this cache by id and the other cache by name for backwards compatibility reasons?

central/cve/node/datastore/datastore_impl.go
```diff
@@ -0,0 +1,201 @@
...
+			SuppressActivation: cve.GetSnoozeStart(),
+			SuppressExpiry:     cve.GetSnoozeExpiry(),
+		}
+	}
+	return nil
+}
+
+func (ds *datastoreImpl) Search(ctx context.Context, q *v1.Query) ([]pkgSearch.Result, error) {
```
connorgorman, line 56: wdyt if we just embed the search into the datastore like:

type datastoreImpl struct {
  search.Searcher
}

then we don't even need these functions which are just passed through

central/cve/node/datastore/datastore_impl.go
connorgorman: Actually, below we are doing it by CVE name. Let's standardize?

connorgorman (changes requested):
See confusion over sometimes the node suppression cache has id and other times it has cve name, which will be different in PG right? Due to cve:os namespacing

central/cve/image/datastore/datastore_impl.go
md2119: we have cves, not the ids. Therefore, we cannot.

central/cve/image/datastore/datastore_impl.go
md2119: we did not have cve before. We have it now. ID always existed.

central/cve/node/datastore/datastore_impl.go
md2119: This is a miss, good catch.

central/cve/node/datastore/datastore_impl.go
md2119: I prefer not to do any specialization for dackbox-related stores. It makes auto-generation tedious. If this is the pattern that can be followed everywhere, then it can very well come as an independent central-wide task.

md2119:
> See confusion over sometimes the node suppression cache has id and other times it has cve name, which will be different in PG right? Due to cve:os namespacing

Clarified. See comments. We never stored cve name as a field. I added a new field to distinguish between the two. Up until now, id was same as the public cve id. But this is not the case anymore.

PR #1977, opened 2022-06-07
ROX-11245 New graphql endpoint to get a list of sorted images, sorted by CVE severity distribution.

clickboo:
## Description
New endpoint for the new actionable widget in the main dashboard redesign.

## Checklist

## Testing Performed
Manual on gke cluster using postman

### DIFF
--- a/central/graphql/resolvers/images.go
+++ b/central/graphql/resolvers/images.go
@@ -2,11 +2,14 @@ package resolvers
 
 import (
 	"context"
+	"math"
+	"sort"
 	"time"
 
 	"github.com/graph-gophers/graphql-go"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/graphql/resolvers/distroctx"
+	"github.com/stackrox/rox/central/graphql/resolvers/inputtypes"
 	"github.com/stackrox/rox/central/graphql/resolvers/loaders"
 	"github.com/stackrox/rox/central/metrics"
 	"github.com/stackrox/rox/central/role/resources"
@@ -15,6 +18,7 @@ import (
 	"github.com/stackrox/rox/pkg/features"
 	pkgMetrics "github.com/stackrox/rox/pkg/metrics"
 	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/search/paginated"
 	"github.com/stackrox/rox/pkg/search/scoped"
 	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/rox/pkg/utils"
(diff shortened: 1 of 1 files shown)


md2119 (commented):
Boo, this seems like a good opportunity to change our search fields to address use cases like this. Currently, we have search field tags on proto fields that mark them searchable as well as sortable. Fields not coming directly are derived fields. What if we add a derived field option to `search/options.go` and tackle this field at the internal level instead of graphQL layer. That opens doors to use this field to be used elsewhere—image endpoints, images csv export, etc. Further, we can control if the field is searchable or sortable by augmenting `FieldLabel` with new properties.

clickboo:
> Boo, this seems like a good opportunity to change our search fields to address use cases like this. Currently, we have search field tags on proto fields that mark them searchable as well as sortable. Fields not coming directly are derived fields. What if we add a derived field option to `search/options.go` and tackle this field at the internal level instead of graphQL layer. That opens doors to use this field to be used elsewhere—image endpoints, images csv export, etc. Furthermore, we can control if the field is searchable or sortable by augmenting `FieldLabel` with new properties.

Good idea though I believe all that is changing with postgres and Viswa is currently working on it (derived fields especially).

connorgorman (approved):
If my read is correct, this will read *all* images from the database then run the image vulnerability counter over them. This is really expensive from both a CPU and Memory perspective, and I can't imagine the latency being good at scale. I think we need to rethink the approach

connorgorman (changes requested):
See previous comment, meant to request changes

clickboo:
> If my read is correct, this will read _all_ images from the database then run the image vulnerability counter over them. This is really expensive from both a CPU and Memory perspective, and I can't imagine the latency being good at scale. I think we need to rethink the approach

This has already been brought up to PM but the widget needs it :). The images are going to be filtered down by cluster and/or images from a widget perspective.  In the postgres world, if the derived fields work by Viswa can help this great but the need for the sorting by vuln severity distribution is not going away from the dashboard.

viswajithiii:
> Boo, this seems like a good opportunity to change our search fields to address use cases like this. Currently, we have search field tags on proto fields that mark them searchable as well as sortable. Fields not coming directly are derived fields. What if we add a derived field option to `search/options.go` and tackle this field at the internal level instead of graphQL layer. That opens doors to use this field to be used elsewhere—image endpoints, images csv export, etc. Furthermore, we can control if the field is searchable or sortable by augmenting `FieldLabel` with new properties.

I'm doing this, but it will only work for postgres. @clickboo, do we need this functionality before postgres?

clickboo:
> > Boo, this seems like a good opportunity to change our search fields to address use cases like this. Currently, we have search field tags on proto fields that mark them searchable as well as sortable. Fields not coming directly are derived fields. What if we add a derived field option to `search/options.go` and tackle this field at the internal level instead of graphQL layer. That opens doors to use this field to be used elsewhere—image endpoints, images csv export, etc. Furthermore, we can control if the field is searchable or sortable by augmenting `FieldLabel` with new properties.
> 
> I'm doing this, but it will only work for postgres. @clickboo, do we need this functionality before postgres?

Yes, needed before postgres given release timelines of this feature.

md2119:
@clickboo, maybe we can incorporate the count fields required for the widget in question as top-level fields on image proto. Currently, we store total cve count and total fixable cve count. The total count fields are populated at upserts anyway. We can expand on the current proto to store counts by severity. 

This will also give the data readily for several other use cases.

PR #1980, opened 2022-06-07
[ROX-11301] Add new datastore for node components

md2119:
## Checklist


central/nodecomponent/datastore/datastore.go
```diff
@@ -0,0 +1,41 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	searchPkg "github.com/stackrox/rox/pkg/search"
+)
+
+// DataStore is an intermediary to NodeComponent storage.
+//go:generate mockgen-wrapper
+type DataStore interface {
+	Search(ctx context.Context, q *v1.Query) ([]searchPkg.Result, error)
+	SearchNodeComponents(ctx context.Context, q *v1.Query) ([]*v1.SearchResult, error)
```
connorgorman, line 20: Do we need this one?

central/nodecomponent/datastore/index/index.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Indexer provides funtionality to index node components.
+type Indexer interface {
+	AddNodeComponent(components *storage.NodeComponent) error
+	AddNodeComponents(components []*storage.NodeComponent) error
+	Count(q *v1.Query, opts ...blevesearch.SearchOption) (int, error)
+	DeleteNodeComponent(id string) error
+	DeleteNodeComponents(ids []string) error
+	MarkInitialIndexingComplete() error
+	NeedsInitialIndexing() (bool, error)
```
connorgorman, line 18: I think it's okay to omit the ones that are not relevant in Postgres

central/nodecomponent/datastore/search/searcher.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stackrox/rox/pkg/search/blevesearch"
+	pkgPostgres "github.com/stackrox/rox/pkg/search/scoped/postgres"
+)
+
+// Searcher provides search functionality on existing image components.
+//go:generate mockgen-wrapper
+type Searcher interface {
+	Search(ctx context.Context, query *v1.Query) ([]search.Result, error)
+	Count(ctx context.Context, query *v1.Query) (int, error)
+	SearchNodeComponents(context.Context, *v1.Query) ([]*v1.SearchResult, error)
```
connorgorman, line 20: Don't think this will be used so can be omitted

connorgorman (changes requested):
Small changes. I don't think we should carry forward the v1.SearchResult as we don't actually care about registering this with global search right now. In fact, not sure if we should be registering the other components either?

central/nodecomponent/datastore/datastore.go
md2119: Yea, graphQL

central/nodecomponent/datastore/search/searcher.go
md2119: we need it for graphQL :)

md2119:
> Small changes. I don't think we should carry forward the v1.SearchResult as we don't actually care about registering this with global search right now. In fact, not sure if we should be registering the other components either?

@connorgorman, do we know for sure that our global search is not used directly from APIs? If not, we will have to release a notice about the change. But that also means that we cannot remove it RN.

central/nodecomponent/datastore/index/index.go
md2119: Fair enough. I seemed have to have copied the common interface pattern between dackbox and postgres.

connorgorman (approved):

central/nodecomponent/datastore/search/searcher.go
md2119: discussed offline

PR #1995, opened 2022-06-08
[ROX-11280] : Add decommissioned cluster retention config and default value

charmik-redhat:
## Description

- Added `decommissioned_cluster_retention_duration_days` config to config proto
- Added 90 days default value for retention config

## Checklist

- No upgrade steps needed for this change
- Autogenerated API Reference doc shows new retention config under `ConfigService`
- Change log and additional documentation will be added after adding the cluster auto removal mechanism

## Testing Performed

Manually called the ConfigService API to see if `decommissionedClusterRetentionDurationDays` field exists and is set to expected default value

D50F03EC-742B-4BA8-819C-E5E56D3FB1C0_4_5005_c


proto/storage/config.proto
```diff
@@ -43,15 +43,16 @@ message AlertRetentionConfig {
...
-        int32                DEPRECATED_alert_retention_duration_days      = 1;
-        AlertRetentionConfig alert_config                                  = 3;
+        int32                DEPRECATED_alert_retention_duration_days       = 1;
+        AlertRetentionConfig alert_config                                   = 3;
     }
-    int32                    image_retention_duration_days                 = 2;
-    int32                    expired_vuln_req_retention_duration_days      = 4;
+    int32                    image_retention_duration_days                  = 2;
+    int32                    expired_vuln_req_retention_duration_days       = 4;
+    int32                    decommissioned_cluster_retention_duration_days = 5;
```
md2119, line 55: Let's see if we get something that beats this var name length

clickboo (approved):

proto/storage/config.proto
pedrottimark: What is plus and minus of uncontacted versus decomissioned, from viewpoint of some who reads API Reference without the context that we have from discussion of feature?

pedrottimark: Or does that imply never contacted aka uninitialized?

clickboo: To me that: uncontacted is not as intentional. Could be inferred as uninitialized.

PR #1998, opened 2022-06-08
Prototype CRD API generation

SimonBaeumer:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


rukletsov:
> Why is declarative configuration an advantage?

What are the disadvantages? I can think of "backward reconciliation": imagine someone makes changes to the auth provider configured via the CRD, e.g. adds more rules. Will that change be reflected in the CRD? If not, if someone adds another, potentially conflicting, change to the CRD, how the reconciler shall reconcile?

SimonBaeumer:
>  imagine someone makes changes to the auth provider configured via the CRD, e.g. adds more rules. Will that change be reflected in the CRD?

This is depended on the implementer. We can use the CRD as a storage backend (in fact it is), thus removing postgres/rocksdb for AuthProvider config altogether.
If a hybrid approach is chosen either disabling the configuration from the UI or Central modifies the CRD directly on UI updates. This must be implemented in the Central's data layer.
The reconcile function is only responsible for setting up the desired state from the CRD to match the server/application state.

> If not, if someone adds another, potentially conflicting, change to the CRD, how the reconciler shall reconcile?

Overwriting UI changes, treating etcd like any other database. If a change is made directly to rocksdb/postgres these are overwritten directly too.

PR #2004, opened 2022-06-09
Jv/rox 10930 automate creation of long running script

JoukoVirtanen:
## Description

Automates the creation of long-running clusters for releases.

## Checklist
- [ ] Ran the script

## Testing Performed

See above


scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
...
+  
+  export ROX_ADMIN_PASSWORD="$(cat deploy/k8s/central-deploy/password)"
+  export ROX_PASSWORD="$ROX_ADMIN_PASSWORD"
+
+  "$toplevel_dir/deploy/k8s/sensor.sh"
+  wait_for_pods_to_be_ready
+
+  kubectl -n stackrox set env deploy/sensor MUTEX_WATCHDOG_TIMEOUT_SECS=0
+  kubectl -n stackrox set env deploy/sensor ROX_FAKE_KUBERNETES_WORKLOAD=long-running
+  wait_for_pods_to_be_ready
```
0x656b694d, line 280: Do we need to wait for pods twice here?

0x656b694d (changes requested):

scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
...
+      infractl create gke-default $CLUSTER_NAME --lifespan 168h --arg nodes=5 --wait --slack-me
+  fi
+  
+  # Set your local kubectl context to the remote cluster once the above completes successfully.
+  infractl get $CLUSTER_NAME --json | jq '.Connect' -r | bash
+
+  kubectl -n stackrox get pod
+  
+  
+  export MAIN_IMAGE_TAG=$(git describe --tags --abbrev=0) # Release version, e.g. 3.63.0-rc.2.
```
vikin91, line 253: Can't we use the variables `RELEASE PATCH_NUMBER RC_NUMBER` here?

scripts/release-tools/clusters.sh
vikin91: Yes, I think this is necessary, as deploying the long-running workload changes several things in the cluster - including restarts of some pods.

scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
   kubectl -n stackrox get pods
 }
 
+create_long_running_cluster() {
+
```
vikin91, line 233: I suggest adding here sanity checks:

```shell
  [[ -n "$RC_NUMBER" ]] || die "RC_NUMBER undefined"
  [[ -n "$RELEASE" ]] || die "RELEASE undefined"
  [[ -n "$PATCH_NUMBER" ]] || die "PATCH_NUMBER undefined"
```

scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
...
+  export API_ENDPOINT="localhost:8000"
+  
+  export STORAGE=pvc # Backing storage
+  export STORAGE_CLASS=faster # Runs on an SSD type
+  export STORAGE_SIZE=100 # 100G
+  export MONITORING_SUPPORT=true # Runs monitoring
+  export LOAD_BALANCER=lb
+  
+  toplevel_dir="$(git rev-parse --show-toplevel)"
+  "$toplevel_dir/deploy/k8s/central.sh" # Launches central
```
vikin91, line 263: This step failed for me, because docker daemon was not running

```
docker: Cannot connect to the Docker daemon at unix:///Users/prygiels/.colima/default/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
```

Could we add a check for the daemon to be running at the beginning of this function?

(big plus for the current possibility to start docker, rerun the script, and not waste the cluster)

scripts/release-tools/clusters.sh
vikin91: Now I am sure this will not work, because  I have tried it :)

```
    Image:      quay.io/stackrox-io/main:3.70.x
```

vikin91 (dismissed):
Looks really good, there are just few rough edges to polish 😉

scripts/release-tools/clusters.sh
```diff
@@ -114,6 +119,17 @@ ensure_cluster_exists() {
   infractl get "${cluster_name}" || die "cluster '${cluster_name}' not found"
 }
 
+wait_for_pods_to_be_ready() {
+  while true; do
+    sleep 5
```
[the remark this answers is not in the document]
JoukoVirtanen: My thought here was that it is only checking if existing pods are ready. My worry is that maybe we are waiting on pods that don't even show up with kubectl get pods and we have to wait a bit for them to show up. Maybe this concern doesn't make sense. Probably there is a better way to check for this.

scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
...
+  export MONITORING_SUPPORT=true # Runs monitoring
+  export LOAD_BALANCER=lb
+  
+  toplevel_dir="$(git rev-parse --show-toplevel)"
+  "$toplevel_dir/deploy/k8s/central.sh" # Launches central
+  wait_for_pods_to_be_ready
+  
+  # Open port-forward to central, e.g. with
+  kubectl -n stackrox port-forward deploy/central 8000:8443 > /dev/null 2>&1 &
+  sleep 60
```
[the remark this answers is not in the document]
JoukoVirtanen: This was not working without the sleep. I don't remember the error that I had without it.

scripts/release-tools/clusters.sh
JoukoVirtanen: Done

scripts/release-tools/clusters.sh
JoukoVirtanen: I added a function check_docker_login. I was not sure about the best implementation, but it seems to work.

scripts/release-tools/clusters.sh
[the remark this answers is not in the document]
JoukoVirtanen: I moved the sleep to the bottom of the loop and added a check to make sure that pods with the expected names exist.

scripts/release-tools/clusters.sh
```diff
@@ -212,6 +229,62 @@ create_qa_gke_cluster() {
...
+
+  CLUSTER_NAME="$(get_cluster_name long-running)"
+  export CLUSTER_NAME
+
+  echo "cluster_name= $CLUSTER_NAME"
+  
+  status="$(check_cluster_status "$CLUSTER_NAME")"
+  
+  if does_cluster_exist "$CLUSTER_NAME"; then
+      echo "Unable to create cluster"
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

scripts/release-tools/clusters.sh
JoukoVirtanen: I now use the environment variables.

scripts/release-tools/clusters.sh
JoukoVirtanen: Done

scripts/release-tools/clusters.sh
0x656b694d: same effect as "set -u"?

scripts/release-tools/clusters.sh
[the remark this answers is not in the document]
0x656b694d: As this opens a tunnel to the central, we just need to wait for the central pod to be up and running. But you've added the wait few lines up already.

0x656b694d (approved):
LGTM

PR #1878, opened 2022-05-30
ROX-11138: Skip splunk test

janisz:
## Description

This test if failing often. Let's ignore it for now and reenable when it's fixed.

## Testing Performed
CI

PR #2005, opened 2022-06-09
ROX-11138: fix splunk e2e test

ivan-degtiarenko:
## Description

In May 2022, IntegrationsSplunkViolationsTest stopped working as expected with an error

```
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: no baseline with given deployment ID found: invalid arguments
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:235)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:216)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:141)
	at io.stackrox.proto.api.v1.NetworkBaselineServiceGrpc$NetworkBaselineServiceBlockingStub.lockNetworkBaseline(NetworkBaselineServiceGrpc.java:394)
	at services.NetworkBaselineService.lockNetworkBaseline(NetworkBaselineService.groovy:17)
	at IntegrationsSplunkViolationsTest.triggerNetworkFlowViolation(IntegrationsSplunkViolationsTest.groovy:303)
	at IntegrationsSplunkViolationsTest.Verify Splunk violations: StackRox violations reach Splunk TA(IntegrationsSplunkViolationsTest.groovy:98)
```

The root cause of that was change in a network baseline behaviour implemented here: https://github.com/stackrox/stackrox/pull/1489

```
We are moving to a model where we do not build network baselines by default as flows come in. We will be a network baseline under 2 circumstances:

* The observation window for a deployment expires
* The user requests to view a network baseline
```

This PR fixes an error by adding the `NetworkBaselineService.getNetworkBaseline` call before calling `NetworkBaselineService.lockNetworkBaseline`.

There are also two other small improvements made within this PR:
* when triggering process violation, time for `curl` operation is now limited since endpoint we are calling is unreachable - we are only using `curl` because it's a command triggering violation
* to be more aligned with `NetworkBaselineTest`, we switched from `curl` to `wget`

## Checklist

## Testing Performed

1. Ran locally 3 times
2. CI should be sufficient - if we see an error returned, we will disable the test again

### DIFF
--- a/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
@@ -304,7 +304,7 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
         NetworkBaselineService.getNetworkBaseline(uid)
         NetworkBaselineService.lockNetworkBaseline(uid)
         orchestrator.execInContainer(splunkDeployment.deployment,
-                "for i in `seq 100`; do curl http://${centralHost}:443 --max-time 1; sleep 1; done")
+                "for i in `seq 100`; do  wget -S http://${centralHost}:443; sleep 1; done")
 
         // TODO: this code is flaky; see https://stack-rox.atlassian.net/browse/ROX-7772
         assert waitForAlertWithPolicyId("1b74ffdd-8e67-444c-9814-1c23863c8ccb")
--- a/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
@@ -11,18 +11,15 @@ import com.jayway.restassured.response.Response
 import io.stackrox.proto.api.v1.AlertServiceOuterClass
 
 import groups.Integration
-import objects.Deployment
 import services.AlertService
 import services.ApiTokenService
 import services.NetworkBaselineService
-import util.NetworkGraphUtil
 import util.SplunkUtil
 import util.Timer
 
 import org.junit.Rule
 import org.junit.experimental.categories.Category
 import org.junit.rules.Timeout
-import spock.lang.Ignore
 
 class IntegrationsSplunkViolationsTest extends BaseSpecification {
     @Rule
@@ -221,7 +218,7 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
     private static String extractNestedString(JsonPath jsonPath, String path) {
         try {
             return jsonPath.getString(path)
-        } catch (IllegalArgumentException e) {
+        } catch (IllegalArgumentException ignored) {
             return null
         }
     }
--- a/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
@@ -11,9 +11,11 @@ import com.jayway.restassured.response.Response
 import io.stackrox.proto.api.v1.AlertServiceOuterClass
 
 import groups.Integration
+import objects.Deployment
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
```diff
@@ -83,7 +82,6 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
     }
 
     @Category(Integration)
-    @Ignore("ROX-11138")
```
janisz, line 86: Previously it was[ `@IgnoreIf({ Env.CI_JOBNAME.contains("postgres") })`](https://github.com/stackrox/stackrox/pull/1878/files) but I'm fine with enabling this test on postgres if it passes

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
```diff
@@ -294,14 +292,16 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
...
     }
 
     def triggerNetworkFlowViolation(SplunkUtil.SplunkDeployment splunkDeployment, String centralHost) {
-        NetworkBaselineService.lockNetworkBaseline(splunkDeployment.getDeployment().deploymentUid)
+        def uid = splunkDeployment.getDeployment().getDeploymentUid()
+        NetworkBaselineService.getNetworkBaseline(uid)
+        NetworkBaselineService.lockNetworkBaseline(uid)
         orchestrator.execInContainer(splunkDeployment.deployment,
-                "for i in `seq 25`; do curl http://${centralHost}:443; sleep 1; done")
+                "for i in `seq 100`; do  wget -S http://${centralHost}:443; sleep 1; done")
```
janisz, line 304: Why `wget` instead of `curl`?

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
dashrews78: glancing at it, I think it should pass with postgres on.  I recommend adding the postgres ci label to find out

janisz (commented):
@ivan-degtiarenko Could you run postgres e2e?

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
```diff
@@ -220,7 +218,7 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
     private static String extractNestedString(JsonPath jsonPath, String path) {
         try {
             return jsonPath.getString(path)
-        } catch (IllegalArgumentException e) {
+        } catch (IllegalArgumentException ignored) {
```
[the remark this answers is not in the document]
ivan-degtiarenko: Remembering what we tried to achieve here, this function is used to simulate expected CIM mappings behaviour. CIM mappings are basically how the violation is mapped into violation in Splunk.

`IllegalArgumentException` is thrown here when there is no value with such a path. Thus, it is **expected** for `jsonPath.getString(path)` in some cases to throw `IllegalArgumentException`.  Moreover, if we log this error, we potentially will log it [several times](https://github.com/stackrox/stackrox/blob/6f94f90eb5e9109118785693181ac6bdc3e55bd8/qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy#L143-L202) for every violation. That's a lot of logs.

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
dashrews78: I wish I could tell you why that one uses wget, but it was like that before I showed up.

dashrews78 (approved):

PR #2012, opened 2022-06-09
Update SearchFilter to increase type saftey against undefined keys

dvail:
## Description

This is something I ran into during dashboard work. The type
` type SearchFilter = Record<string, string | string[]>` asserts that _every_ string key will return a `string | string[]` on a given SearchFilter object, which is not the case. Almost all uses of the SearchFilter type access the keys/values via `Object.entries` which avoids the issue as a side effect, but this change will ensure that direct key access does not type check by default.

The trade off is that before this change you could not assign a value of `undefined` to a search filter object, but you could incorrectly read an `undefined` value.

```typescript
    const oldSearch: SearchFilter = {};
    const test: string | string[] = oldSearch.doesNotExist; // `test` is actually undefined at this point leading to runtime errors
    oldSearch.b = undefined; // Type 'undefined' is not assignable to type 'string | string[]'
```

After this change, the behavior is reversed:
```typescript
    const newSearch: SearchFilter = {};
    const test: string | string[] = newSearch.doesNotExist; // Error - Type 'undefined' is not assignable to type 'string | string[]'.
    newSearch.b = undefined; // This is fine
```

This makes the writing of values to a `SearchFilter` more permissive, but makes reading values more restrictive (also correct), which IMO is a better trade.

## Checklist

## Testing Performed

Automated testing only.

The only non-type level change here is in a function that constructs a query string from a `SearchFilter`. In the previous behavior this would result in backend queries that looked like `"Cluster:undefined+Namespace:stackrox"`, while the new behavior would result in `"Namespace:stackrox"`, as the `undefined` value would be filtered out.

### DIFF
--- a/ui/apps/platform/src/Components/SearchFilterInput/SearchFilterInput.tsx
+++ b/ui/apps/platform/src/Components/SearchFilterInput/SearchFilterInput.tsx
@@ -115,7 +115,7 @@ function getSearchEntriesForFilter(
                         value,
                     });
                 });
-            } else if (valueOrValues.length !== 0) {
+            } else if (valueOrValues && valueOrValues.length !== 0) {
                 searchEntries.push({
                     label: valueOrValues,
                     value: valueOrValues,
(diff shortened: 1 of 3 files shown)


pedrottimark (approved):
Thank you for this example how to write more effective TypeScript!

dvail:
I was doing some reading on some of the unsoundness in TypeScript's type system and it looks like you can enabled this behavior by default for object and array access by turning on `"noUncheckedIndexedAccess" : true`.

`yarn tsc` in our repo gives: `Found 168 errors in 59 files.` though so probably not something we want to do at this point. 😅

PR #2016, opened 2022-06-09
Add tracing when Postgres is enabled for legacy storage

connorgorman:
## Description

Logs when on Postgres if we access Bolt, RocksDB or Bleve

## Checklist

## Testing Performed

Will run postgres tests and see what log lines exist


md2119 (approved):
Nice. Add one for dackbox.

PR #2029, opened 2022-06-10
Add cluster/namespace scope bar to action widget dashboard.

dvail:
## Description

Adds a "scope" bar to the new dashboard page allowing the user to filter data in the page widgets by multiple clusters and namespaces.

## Notes/Follow ups

1. These components are driven by a single GQL request that pulls all clusters (id/name) and their associated namespaces (id/name). Depending on scale we might want to optimize this by only pulling a subset of the data (either by pagination, by search filter values...?). This will have the following impacts:
- More latency as the user browses the dropdown lists in exchange for a quicker initial load of the data
- A change in the UX of how the lists work (do we lazy load data, have a "View More" link, only load more through search?)
- An increase in code complexity as we'll have to carefully cache data in order to keep the two dropdowns in sync

2. The ClusterSelect and Namespace dropdowns are _very_ similar in implementation, but since there are only two implementations and they are different enough I held off on pulling out an abstraction. We might want this in the future though if we adopt the pattern more.

3. Currently data for all charts is reloaded each time an option changes. We might want to consider only reloading the data when the dropdowns are closed for performance reasons if the queries are intensive.

## Checklist

## Testing Performed

Test that the initial page load populates the Cluster dropdown with all available clusters, and that both dropdowns are in an "All Selected" state. The NS dropdown should be disabled.
image

Test that selecting a cluster enables the namespace dropdown, updates the page URL, and updates the data in the chart.
image

Test that selecting a namespace updates the page URL and updates the data in the chart.
image

Test that selecting another cluster automatically selects all namespace in that cluster when the namespace dropdown is not in an "All Selected" state.
image
image

Test that deselecting a cluster deselects all namespaces for that cluster and removes those cluster options from the namespace dropdown.
image
image

Test that selecting "All Clusters" clears the URL and sets both dropdowns to an "All Selected" state.
image

Test that selecting a single cluster, and then deselecting that cluster reverts to an "All Selected" state instead of resulting in an empty dropdown.

Test that the scope bar wraps to a vertical layout and remains usable on smaller screens.
image


ui/apps/platform/src/Components/PatternFly/ResourceIcon/ResourceIcon.css
```diff
@@ -0,0 +1,14 @@
+.resource-icon {
+  color: var(--pf-global--Color--light-100);
+  border-radius: 30em;
+  padding: 0 4px;
+  font-size: var(--pf-global--FontSize--xs);
+}
+
+.resource-icon-cluster {
+  background-color: #6752ae;
```
dvail, line 9: These colors are pulled from the mocks and should be similar to the ones in the OpenShift console code. I didn't think we have a use for these other than localized to this component, so I didn't pull them into a global "colors" file or CSS vars.

ui/apps/platform/src/Containers/Dashboard/PatternFly/ScopeBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+`;
+
+function ScopeBar() {
+    const { searchFilter, setSearchFilter } = useURLSearch();
+    const { data, loading, error } = useQuery<NamespacesResponse>(namespacesQuery, {
+        variables: { query: '' },
+    });
+    const selectedClusterData = useMemo(() => {
+        return data?.clusters.filter(({ name }) => searchFilter.Cluster?.includes(name)) ?? [];
```
dvail, line 38: This searching an array for values in another array isn't great, but I'm not sure that number of items we are comparing here calls for making this more complicated/performant.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -126,6 +128,7 @@ function ViolationsByPolicyCategoryChart({ alertGroups }: ViolationsByPolicyCate
             <Chart
                 ariaDesc="Number of violation by policy category, grouped by severity"
                 ariaTitle="Policy Violations by Category"
+                animate={{ duration: 300 }}
```
dvail, line 131: I think this looks good, and makes it more clear when data is changing, but I can run it by UX during our Monday meeting.

ui/apps/platform/src/Components/PatternFly/ResourceIcon/ResourceIcon.css
vjwilson: @alanonthegit 
It looks like there are some colors used in different places in OpenShift Admin console. (see example in screenshot below)

Could we check with the other OpenShift UX teams to see if there are recommended colors for k8s entities, and use those?

Or are the colors in this mock already the standard colors to represent Cluster and Namespace entities in OpenShift?

Regardless of whether we adhere to an existing OpenShift UX standard, or create our own, we should document those colors and make them app-wide for our app. We should not be defining them over and over again in different components.

Screen Shot 2022-06-10 at 12 04 29 PM

alanonthegit: @vjwilson @dvail I suggest using the OpenShift defaults, whatever they may be. Our mocks weren't going for accuracy there. My guess is that the component you're working with expects a color value/variable as a prop?

I don't know if OpenShift already has fixed colors per resource kind defined in some HOC map, or if that HOC does some kind of auto-color generation (since there are so many more resources in OCP, including customer-defined CRDs).

But in either case, we should aim for consistency with how OCP does them and avoid creating our own color variables if possible. I don't believe many designers will know how the colors are selected/picked/generated, so it may be worthwhile for engineering to follow up instead and get the source of truth from the component code instead.

alanonthegit: @dvail @vjwilson also as a general comment, maybe we should be careful and look to avoid defining/hardcoding color values in our UI whenever possible and use OCP variables instead so we can ensure we don't have any weird quirks when we implement functionality like OCP's dark mode.

From what I'm seeing in the current OCP UI, the resource chips remain the same but that may not always be the case for every component. Maybe as a stop-gap, we can still copy over their color variable mappings into our UI and sort out a more integrated / single source of truth solution down the road. 

image

image

alanonthegit: @dvail @vjwilson (sorry, last message) - Some elements certainly looks hardcoded in OCP 
image

However, I just realized you won't find "Cluster" since OCP is single cluster. Maybe with an ACM integration build it could be there. The rest of the CSS variables seem to be defined in a `resource.scss`  which isn't helpful to us since we don't compile SCSS in our build. I think they may be using SCSS variables and outputting CSS variables.

Here are some mappings I pulled from the compiled CSS file. Not sure if they're complete but it's a start.

```
.co-m-resource-icon {
  background-color: #2b9af3;
  border-radius: 20px;
  color: var(--pf-global--palette--white);
  display: inline-block;
  flex-shrink: 0;
  font-size: 14px;
  line-height: 16px;
  margin-right: 4px;
  min-width: 18px;
  padding: 1px 4px;
  text-align: center;
  white-space: nowrap;
}

.co-m-resource-icon--lg {
  font-size: 17px;
  line-height: 20px;
  margin-right: 7px;
  min-width: 24px;
  padding: 0 7px;
}

.co-resource-icon--fixed-width {
  display: flex;
  justify-content: center;
  text-align: center;
  width: 50px;
}

.co-m-resource-application {
  background-color: #3e8635;
}

.co-m-resource-clusterrole,
.co-m-resource-role {
  background-color: #795600;
}

.co-m-resource-clusterrolebinding,
.co-m-resource-rolebinding {
  background-color: #008bad;
}

.co-m-resource-daemonset,
.co-m-resource-deployment,
.co-m-resource-deploymentconfig,
.co-m-resource-job,
.co-m-resource-petset,
.co-m-resource-replicaset,
.co-m-resource-replicationcontroller {
  background-color: #004080;
}

.co-m-resource-machine,
.co-m-resource-machineautoscaler,
.co-m-resource-machineclass,
.co-m-resource-machineconfig,
.co-m-resource-machineconfigpool,
.co-m-resource-machinedeployment,
.co-m-resource-machinehealthcheck,
.co-m-resource-machineset,
.co-m-resource-node,
.co-m-resource-policy {
  background-color: #8476d1;
}

.co-m-resource-alertrule,
.co-m-resource-configmap {
  background-color: #40199a;
}

.co-m-resource-alert,
.co-m-resource-container {
  background-color: #2b9af3;
}

.co-m-resource-pod {
  background-color: #009596;
}

.co-m-resource-namespace,
.co-m-resource-project {
  background-color: #1e4f18;
}

.co-m-resource-secret {
  background-color: #c46100;
}

.co-m-resource-alertmanager {
  background-color: #773d00;
}

.co-m-resource-service {
  background-color: #6ca100;
}

.co-m-resource-serviceaccount {
  background-color: #40199a;
}

.co-m-resource-ingress {
  background-color: #1f0066;
}
```

vjwilson: @dvail We should definitely extract it, but let's not wait to get the actual code out of OpenShift code.

I like your intermediate idea of "do something similar on our end with CSS vars".

For validation on cluster colors, you could ask James Talton, the UI lead with ACM. He is on CoreOS Slack.

dvail: Ok I've updated to use CSS vars and matched the colors to openshift console defined colors. James has his slack status set to 💤 🌴  so I've made a note to myself to follow up with him on Monday.

ui/apps/platform/src/Containers/Dashboard/PatternFly/ClusterSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+    }
+
+    function onFilter(e: ChangeEvent<HTMLInputElement> | null, filterValue: string) {
+        return createOptions(clusters, filterValue);
+    }
+
+    return (
+        <Select
```
pedrottimark, line 87: Two optional properties to consider for accessibility:
* `aria-label="Selected clusters"` here: you might improve this or ask on forum-acs-ui-team channel
    Similar naming convention for `Select` element in `NamespaceSelect` component
* `aria-hidden="true"` for `span` element in `ResourceIcon` component

PR #2035, opened 2022-06-10
CI: build CLI regardless of context

gavin-stackrox:
## Description

#1945 broke roxctl tests in merge CI. This PR fixes them by re-enabling the cli build regardless of context.

## Checklist
- [x] Investigated and inspected CI test results

## Testing Performed

CI is sufficient

PR #2037, opened 2022-06-10
Revert "CI: build CLI regardless of context"

gavin-stackrox:
Reverts stackrox/stackrox#2035


gavin-stackrox:
This does not need to be reverted to undo #1945

PR #2039, opened 2022-06-10
ROX-11335: Move Scanner DB secret out of main Scanner DB container

RTann:
## Description

If someone can access the main Scanner DB container, then they can read the password mounted in the container. This PR removes the k8s secret volume mount from the main container and only uses it in the init container.

See https://github.com/stackrox/scanner/pull/774 and https://github.com/stackrox/scanner/pull/793 for related Scanner-side changes.

Also did a little bit of minor cleanup while I'm at it.

## Checklist

## Testing Performed

CI


SimonBaeumer (approved):

image/templates/helm/shared/templates/02-scanner-06-deployment.yaml.htpl
```diff
@@ -273,26 +276,26 @@ spec:
         [< else ->]
         image: {{ ._rox.scanner.dbImage.fullRef | quote }}
         [< end ->]
+        env:
+          {{- include "srox.envVars" (list . "deployment" "scanner-db" "db") | nindent 10 }}
         ports:
-        - name: https-db
+        - name: postgresql
```
jvdm, line 282: Nitpick: All the other resources a nicely prefixed with `scanner-`, though this could also be called `scanner-db-port`? Do we reference this name in the service definition?

RTann: actually, looking at the ports in Scanner, they are called `https` and `grpc` so we seem to name them in the deployment definitions based on the protocol, so i'm going to keep this as `postgres`

connorgorman: I think we want to do tcp-postgresql to make Istio happy if it is used

RTann: good point about istio i didnt know that. i'll switch it to `tcp-postgres` just to be safe. Looks like Istio https://istio.io/latest/docs/reference/config/analysis/ist0118/ wants `<protocol>[-<suffix>]`

image/templates/helm/shared/templates/02-scanner-06-deployment.yaml.htpl
```diff
@@ -273,26 +276,26 @@ spec:
...
-          runAsGroup: 70
         volumeMounts:
-          - name: db-data
+          - name: scanner-db-data
             mountPath: /var/lib/postgresql/data
           - name: scanner-db-tls-volume
             mountPath: /run/secrets/stackrox.io/certs
-          - name: scanner-db-password
-            mountPath: /run/secrets/stackrox.io/secrets
+            readOnly: true
```
RTann, line 292: Does this being read-only affect us updating the certs for local-scanner, for example? I don't think so because this is really just mounting a k8s secret, but just checking. @SimonBaeumer

SimonBaeumer (approved):
/approve

PR #2046, opened 2022-06-13
Clean up compliance datastore and service

connorgorman:
## Description

There is an extra endpoint in the datastore that is the equivalent of another one. Cleaning that up

## Checklist

## Testing Performed

Trivial, these were basically the same functions


proto/api/v1/compliance_service.proto
```diff
@@ -102,12 +98,6 @@ service ComplianceService {
         };
     }
 
-    rpc GetComplianceControlResults(RawQuery) returns (ComplianceControlResultsResponse) {
-        option (google.api.http) = {
-            get: "/v1/compliance/results"
```
connorgorman, line 107: Will add a changelog entry for this, but this does not need to be deprecated as it was never implemented so no customers are relying on it

theencee (approved):

central/compliance/datastore/datastore_impl.go
```diff
@@ -35,11 +31,6 @@ type datastoreImpl struct {
 	aggregationSequenceNumber uint64
 }
 
-func (ds *datastoreImpl) QueryControlResults(ctx context.Context, query *v1.Query) ([]*storage.ComplianceControlResult, error) {
-	// TODO(ROX-2575): this might need implementing.
```
viswajithiii, line 39: Lol

viswajithiii (approved):

PR #2049, opened 2022-06-13
ROX-9640: Port Compliance over to Postgres

connorgorman:
## Description

Work in progress PR for porting compliance over to Postgres. This will initially be a naive port

## Checklist

## Testing Performed

Just trying to build it

### DIFF
--- a/central/compliance/datastore/internal/store/postgres/store.go
+++ b/central/compliance/datastore/internal/store/postgres/store.go
@@ -160,7 +160,7 @@ func (s *storeImpl) GetLatestRunResultsBatch(ctx context.Context, clusterIDs, st
 			if err != nil {
 				return nil, err
 			}
-			pairsToResults[compliance.ClusterStandardPair{ClusterID: clusterID, StandardID: standardID}] = results
+			pairsToResults[compliance.NewPair(clusterID, standardID)] = results
 		}
 	}
 	return pairsToResults, nil
--- a/central/compliance/datastore/internal/store/postgres/store.go
+++ b/central/compliance/datastore/internal/store/postgres/store.go
@@ -135,7 +135,7 @@ func (s *storeImpl) GetSpecificRunResults(ctx context.Context, _, _, runID strin
 func (s *storeImpl) GetLatestRunResults(ctx context.Context, clusterID, standardID string, flags types.GetFlags) (types.ResultsWithStatus, error) {
 	metadataBatch, err := s.getLatestRunMetadata(ctx, clusterID, standardID)
 	if err != nil {
-		return types.ResultsWithStatus{}, nil
+		return types.ResultsWithStatus{}, err
 	}
 	if metadataBatch.LastSuccessfulRunMetadata == nil {
 		return types.ResultsWithStatus{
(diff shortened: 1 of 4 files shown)


central/compliance/datastore/internal/store/postgres/store.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	return types.ResultsWithStatus{
+		LastSuccessfulResults: results,
+	}, nil
+}
+
+func (s *storeImpl) GetLatestRunResults(ctx context.Context, clusterID, standardID string, flags types.GetFlags) (types.ResultsWithStatus, error) {
+	metadataBatch, err := s.getLatestRunMetadata(ctx, clusterID, standardID)
+	if err != nil {
+		return types.ResultsWithStatus{}, nil
```
dashrews78, line 138: Shouldn't we return err here?  Later when this thing is called it checks if it returns err.

central/compliance/datastore/internal/store/postgres/store.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if metadataBatch.LastSuccessfulRunMetadata == nil {
+		return types.ResultsWithStatus{
+			FailedRuns: metadataBatch.FailedRunsMetadata,
+		}, nil
+	}
+	results, err := s.getResultsFromMetadata(ctx, metadataBatch.LastSuccessfulRunMetadata, flags)
+	if err != nil {
+		return types.ResultsWithStatus{}, nil
```
dashrews78, line 147: Shouldn't we return err here?  Later when this thing is called it checks if it returns err.

central/compliance/datastore/internal/store/postgres/store.go
connorgorman: yikes, thanks

central/compliance/datastore/internal/store/postgres/store.go
connorgorman: yikes, thanks

central/compliance/datastore/internal/store/postgres/store.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (s *storeImpl) GetLatestRunResultsBatch(ctx context.Context, clusterIDs, standardIDs []string, flags types.GetFlags) (map[compliance.ClusterStandardPair]types.ResultsWithStatus, error) {
+	pairsToResults := make(map[compliance.ClusterStandardPair]types.ResultsWithStatus)
+	for _, clusterID := range clusterIDs {
+		for _, standardID := range standardIDs {
+			results, err := s.GetLatestRunResults(ctx, clusterID, standardID, flags)
+			if err != nil {
+				return nil, err
+			}
+			pairsToResults[compliance.ClusterStandardPair{ClusterID: clusterID, StandardID: standardID}] = results
```
dashrews78, line 163: This is either a nit OR an opportunity for me to learn something.  Why do you set the map like this here but down below use the NewPair down below?
`results[compliance.NewPair(clusterID, standardID)] = metadata`

connorgorman: because I messed up :)

dashrews78 (approved):
LGTM, only comment outstanding as I type is that tiny nit.

connorgorman:
RocksDB:
Screen Shot 2022-06-14 at 2 50 45 PM

Postgres:
Screen Shot 2022-06-14 at 3 35 20 PM

Aggregation storage optimizations were turned off due to SAC implications so I have not implemented them unnecessarily. If we want to turn them back on, we will have to reconsider. In any case, we would probably want to optimize aggregations and calculations in Postgres anyways

PR #2020, opened 2022-06-10
ROX-11314: add more info to reconciliation tests flake

fredrb:
## Description

In order to try to get more information into why is this test failing, I've added additional logs to the tests, which might be a bit verbose for now, but it could help us understand what's going on. 

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

- Ran groovy tests locally

Here's an example of how the output looks like before Sensor get's removed:
```
10:59:59 | INFO  | ReconciliationTest        | Pods in cluster before deleting:
10:59:59 | INFO  | ReconciliationTest        | kube-system:calico-node-m8b5g
10:59:59 | INFO  | ReconciliationTest        | stackrox:collector-rmk7m
10:59:59 | INFO  | ReconciliationTest        | kube-system:calico-typha-69f8dcc469-rs7pt
10:59:59 | INFO  | ReconciliationTest        | kube-system:calico-typha-horizontal-autoscaler-79f4fdc8d5-ghpdn
10:59:59 | INFO  | ReconciliationTest        | kube-system:fluentbit-gke-v2x7s
10:59:59 | INFO  | ReconciliationTest        | kube-system:event-exporter-gke-5479fd58c8-6gdx2
10:59:59 | INFO  | ReconciliationTest        | kube-system:gke-metrics-agent-th525
10:59:59 | INFO  | ReconciliationTest        | stackrox:scanner-988498f65-mrqgj
10:59:59 | INFO  | ReconciliationTest        | stackrox:sensor-5f67dcf669-wd4bq
10:59:59 | INFO  | ReconciliationTest        | kube-system:kube-dns-697dc8fc8b-sq8dz
10:59:59 | INFO  | ReconciliationTest        | stackrox:collector-k8z4v
10:59:59 | INFO  | ReconciliationTest        | kube-system:kube-proxy-gke-fred-0610-9564-fred--default-pool-5d439f63-zz1p
10:59:59 | INFO  | ReconciliationTest        | kube-system:calico-typha-vertical-autoscaler-b54586f76-ch5c7
[...]

PR #2051, opened 2022-06-14
ROX-11314: Remove reconciliation assertion in the beginning of the test

fredrb:
## Description

After adding the additional logs in #2020 I realized that this failure is happening much before the test is executed. There was an assertion for maximum number of reconciliations done before the test case executed. Here's an example in prow: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-gke-qa-e2e-tests/1536481601444122624

Maybe something changed in the environment, and the number of pods removed while the sensor was down last time is different now. But since this is handled outside the scope of this test, I don't see the value of this assertion here. 

Maybe @viswajithiii can provide more context on why this might be needed? 

## Checklist
- [x] ~~Unit test and regression tests added~~
- [x] ~~Evaluated and added CHANGELOG entry if required~~
- [x] ~~Determined and documented upgrade steps~~
- [x] ~~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed

- CI

### DIFF
--- a/qa-tests-backend/src/test/groovy/ReconciliationTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ReconciliationTest.groovy
@@ -108,10 +108,6 @@ class ReconciliationTest extends BaseSpecification {
         when:
         "Get Sensor and counts"
 
-        // Verify initial reconciliation stats (from the reconciliation that must have happened
-        // whenever the sensor first connected).
-        verifyReconciliationStats(false)
-
         OrchestratorDeployment sensorOrchestratorDeployment =
                 orchestrator.getOrchestratorDeployment("stackrox", "sensor")
         Deployment sensorDeployment = new Deployment().setNamespace("stackrox").setName("sensor")


viswajithiii:
> Maybe @viswajithiii can provide more context on why this might be needed?

This wasn't needed, but this assertion was always true because we didn't really delete anything the rest of the time when sensor bounced. In any case, nothing wrong in removing it, it wasn't really adding anything.

qa-tests-backend/src/test/groovy/ReconciliationTest.groovy
```diff
@@ -108,10 +108,6 @@ class ReconciliationTest extends BaseSpecification {
         when:
         "Get Sensor and counts"
 
-        // Verify initial reconciliation stats (from the reconciliation that must have happened
-        // whenever the sensor first connected).
-        verifyReconciliationStats(false)
```
viswajithiii, line 113: We can now simplify that function by removing the bool parameter altogether, since we only call it once.

fredrb: Thanks. Updated

gavin-stackrox (approved):

PR #2056, opened 2022-06-14
Dashrews/ROX-10903 ROX-11357 flaky process baseline deletion test

dashrews78:
## Description

A couple of process baseline tests were a little flaky on the timing of the processes being added to the baseline or processed as risks.  Turns out we were having issues with slow environments such that if an environment was being slow and an exec process would cause the test to fail, the subsequent retries would try to exec a process on the deployment of the failed attempt instead of the deployment of the retry attempt.  So I switched the cleanup to use deleteAndWaitForDeploymentDeletion and I have had successful runs since.

The test for `Processes come in after baseline deleted by API for #deploymentName` needed some adjustments to make sure the baseline was processing and completing observation before we delete it and again after to ensure the behavior is what we are looking for.

This PR only makes changes to Process Baseline Tests so any CI issues outside of that test are related to other issues that appear to have tickets associated with them already.  Since making the change to wait for deployment deletion, I've had 6 runs without the tests failing, particularly the openshift test as these flakes were more often seen there.

## Checklist
-~~ [ ] Evaluated and added CHANGELOG entry if required~~

## Testing Performed

Tested some locally, but mostly the CI tests are sufficient here as that is where it most often failed.

### DIFF
--- a/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
@@ -473,8 +473,8 @@ class ProcessBaselinesTest extends BaseSpecification {
 
         log.info "Process Baseline before pwd: ${baselineAfterDelete}"
 
-        // sleep 5 seconds to allow for propagation to sensor
-        sleep 5000
+        // sleep 10 seconds to allow for propagation to sensor
+        sleep 10000
         orchestrator.execInContainer(deployment, "pwd")
 
         then:
--- a/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
@@ -153,6 +153,7 @@ class ProcessBaselinesTest extends BaseSpecification {
 
         cleanup:
         "Remove deployment"
+        log.info "Cleaning up deployment: ${deployment}"
         orchestrator.deleteDeployment(deployment)
 
         where:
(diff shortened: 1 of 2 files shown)


qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
```diff
@@ -397,6 +398,7 @@ class ProcessBaselinesTest extends BaseSpecification {
         DEPLOYMENTNGINX_REMOVEPROCESS           |   "nginx"
     }
 
+    @Unroll
```
gavin-stackrox, line 401: This `@Unroll` is not needed because there is no `where:` data set.

dashrews78:
Moved it back to draft as it needs some adjustments and I'm going to be OOO for a stretch.

dashrews78:
moving back to ready as it makes running prow jobs easier.

qa-tests-backend/src/test/groovy/ProcessBaselinesTest.groovy
```diff
@@ -456,21 +470,37 @@ class ProcessBaselinesTest extends BaseSpecification {
         // Baseline should still exist but have no elements associated.  Essentially cleared out.
         assert  ( baselineAfterDelete.elementsList == [] )
 
-        // Give the baseline time to lock again keep in mind process indicators flush every 60 seconds
-        sleep 60000
+        // Give the baseline time to come back out of observation
+        baselineAfterDelete = evaluateWithRetry(30, 3) {
```
gavin-stackrox, line 474:
```suggestion
        baselineAfterDelete = evaluateWithRetry(10, 10) {
```

PR #2062, opened 2022-06-14
ROX-11070: Add ID to group proto.

dhaus67:
## Description

This is the first PR in a series of PRs regarding removing the composite key from the groups datastore, allowing users to specify different roles for the same group properties.

For full context and overview, you can navigate across the PRs within this series via the generated comment.

This will add the field `id` to the `GroupProperties` proto.

The reason why the ID field was added to the GroupProperties message was due to the fact that the API itself requires the `GroupProperties` as input for `GetGroup / RemoveGroup` endpoints.

I'm open for other suggestions, i.e. refactoring the API completely, but I do not see a way of doing this without proper announcement beforehand, thus delaying the fix for this.

## Checklist


dhaus67:
* master
  * **PR #2062** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2062" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2062?utm_source=stack-comment).

theencee (approved):

rukletsov (approved):

PR #2063, opened 2022-06-14
ROX-11070: Update group service and datastore to use ID instead of composite key.

dhaus67:
## Description

This is the second PR in a series of PRs regarding removing the composite key from the groups datastore, allowing users to specify different roles for the same group properties.

For full context, you can navigate to the previous PRs by using the comment created below.

This PR will cover the following:
- Refactoring of the `store` interface and its implementations to use IDs instead of `GroupProperties`, as well as changing the key that is used within the bucket from a composite key to the ID.
- Refactoring the `datastore` interface and its implementations to use IDs instead of `GroupProperties`, as well as moving the validation within the datastore instead of the service layer (similar to how other datastores behave currently).
- Refactoring the `service` layer to use IDs instead of a composite key created out of the `GroupProperties`.

## Checklist


central/group/service/service_impl.go
```diff
@@ -87,7 +93,7 @@ func (s *serviceImpl) GetGroups(ctx context.Context, req *v1.GetGroupsRequest) (
 }
 
 func (s *serviceImpl) GetGroup(ctx context.Context, props *storage.GroupProperties) (*storage.Group, error) {
-	group, err := s.groups.Get(ctx, props)
+	group, err := s.groups.Get(ctx, props.GetId())
```
rukletsov, line 96: This might break current clients who don't know about ID yet (same for `DeleteGroup()`). To avoid breaking the API, let's:
* iff `id` is absent, search the store using `props` and log a warning (maybe one day we'll be adding [deprecation headers](https://tools.ietf.org/id/draft-dalal-deprecation-header-00.html#rfc.section.2))
* iff there is at most one object found, return it
* else, i.e., more than one group is associated with `props`, return `400`
* leave a note to remove that code after the deprecation cycle.

central/group/datastore/validate.go
```diff
@@ -0,0 +1,54 @@
...
+
+// ValidateGroup validates the given group for conformity.
+// A group must fulfill the following:
+//	- have valid properties (validated via ValidateProps).
+//	- have a role name set.
+func ValidateGroup(group *storage.Group) error {
+	if group.GetProps() == nil {
+		return errox.InvalidArgs.New("group properties must be set")
+	}
+	if err := ValidateProps(group.GetProps()); err != nil {
```
rukletsov, line 24: This might be problematic when `ValidateGroup` is called in e.g. `Update(), Upsert(), Mutate()`

rukletsov (changes requested):

central/group/datastore/internal/store/store_impl.go
```diff
@@ -61,16 +63,23 @@ func (s *storeImpl) Walk(authProviderID string, attributes map[string][]string)
...
 	// Search for groups in the list.
 	err = s.db.View(func(tx *bolt.Tx) error {
-		buc := tx.Bucket(groupsBucket)
 		for _, check := range toSearch {
-			serializedKey := serializePkg.PropsKey(check)
-			if serializedVal := buc.Get(serializedKey); serializedVal != nil {
-				grp, err := deserialize(serializedKey, serializedVal)
-				if err != nil {
-					return err
+			grpss, err := s.GetFiltered(func(props *storage.GroupProperties) bool {
```
rukletsov, line 67: This changes the time complexity of `Walk()` and hence of `storeBasedMapperImpl.getRoles()` from `O(k * log n)` to `O(k * n)`. It's hard to believe that this would be a drastic change but it feels that the whole composite key jazz with serializing and deserializing was done for that logarithmic search. Maybe it's worth verifying with @viswajithiii?

I think once we migrate to postgres, we might be able to create arbitrary indices and so fix this back to logarithm. Could you please check with @connorgorman and @rhybrillou?

viswajithiii: This is totally fine, I don't think we expect there to be enough groups for this to be a big deal. And you're right that postgres will make this kind of moot. Fine with me to go ahead with this. (Note: commenting based on the description of the change, haven't looked at the code itself)

central/group/service/service_impl.go
dhaus67: I've addressed this (to be precise, not only for `Get, Delete` but also for `Update, Mutate`). 
This is handled in the store layer instead of the datastore layer, providing the same behavior that you have previously mentioned.

central/group/datastore/validate.go
dhaus67: I've removed the ID check within `ValidateGroup`, as we currently do not want to more lax, supporting the previous way of retrieving groups via their properties. I've added a comment to later on add the ID within the validation.

central/group/datastore/internal/store/store_impl.go
```diff
@@ -15,16 +15,17 @@ type storeImpl struct {
 }
 
 // Get returns a group matching the given properties if it exists from the store.
-func (s *storeImpl) Get(props *storage.GroupProperties) (grp *storage.Group, err error) {
+func (s *storeImpl) Get(id string) (grp *storage.Group, err error) {
 	err = s.db.View(func(tx *bolt.Tx) error {
 		buc := tx.Bucket(groupsBucket)
-		k := serializePkg.PropsKey(props)
```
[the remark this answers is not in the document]
dhaus67: The only logic that remains is the `PropsKey`, as it is still used within `diffGroups` in the service to determine which groups are added / updated / removed. It's used together with the role name of the group as key within a map.

I've created a [new PR ](https://github.com/stackrox/stackrox/pull/2160)where the unused code is removed to not make this one larger than it already is.

central/group/datastore/internal/store/store_impl.go
dhaus67: @rukletsov I've talked to Yann, doing things like that is planned based on the performance investigation results. Right now this would be later this year, where this place would be an opportunity for improvement (albeit, as discussed, the impact may not be as big as other places). Potentially @connorgorman  will have more info, but I don't think it will block us in this PR.

connorgorman: If reading groups is a hot path, we can just cache all of them in memory. This seems like a store with a small cardinality so walk is probably fine for now

rukletsov (changes requested):
Looks good to me modulo some questions for `diffGroups()`. Left a couple of comments on the cosmetics.

central/group/datastore/validate.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/gogo/protobuf/proto"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/errox"
+	"github.com/stackrox/rox/pkg/uuid"
+)
+
+// groupIDPrefix should be prepended to every human-hostile ID of a
+// signature integration for readability, e.g.,
```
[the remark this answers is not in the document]
dhaus67: nothing_to_see

central/group/datastore/validate_test.go
```diff
@@ -32,6 +32,18 @@ func TestValidate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "Group.props.id must be set",
```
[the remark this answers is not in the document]
dhaus67: Turns out it didn't, I've fixed the test. Issue is that prow doesn't run every job when the target branch is not master, which is _meh_ (there's already [a ticket filed to allow this](https://issues.redhat.com/browse/DPTP-2943)).

central/group/datastore/internal/store/store_impl.go
```diff
@@ -146,28 +162,124 @@ func (s *storeImpl) Mutate(toRemove, toUpdate, toAdd []*storage.Group) error {
...
-		return errox.NotFound.Newf("group config for %q does not exist", key)
+
+	// TODO: Once the deprecation of retrieving groups by their properties is fully deprecated, this condition
+	// can be removed and groups shall only be retrievable via their id.
+	if id != "" {
+		if buc.Get([]byte(id)) == nil {
+			return errox.NotFound.Newf("group config for %q does not exist", id)
+		}
+	} else {
+		grps, err := filterInTransaction(tx, func(props *storage.GroupProperties) bool {
```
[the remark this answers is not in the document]
dhaus67: I've used now `getByPropsInTransaction`, which is also used within `store.getByProps`.

central/group/service/service_impl.go
```diff
@@ -98,8 +104,8 @@ func (s *serviceImpl) GetGroup(ctx context.Context, props *storage.GroupProperti
 }
 
 func (s *serviceImpl) BatchUpdate(ctx context.Context, req *v1.GroupBatchUpdateRequest) (*v1.Empty, error) {
-	for _, group := range req.GetRequiredGroups() {
-		if err := validate(group); err != nil {
+	for _, group := range req.GetPreviousGroups() {
```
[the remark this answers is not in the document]
dhaus67: Since we are using the groups within `diffGroups`, we should make sure that they are valid to begin with.
I've added validation for both.

central/group/service/service_impl.go
```diff
@@ -143,28 +146,61 @@ func (s *serviceImpl) DeleteGroup(ctx context.Context, props *storage.GroupPrope
...
+
 	for _, group := range required {
-		requiredByProps[string(serialize.PropsKey(group.GetProps()))] = group
+		// Need to deal with groups that do not have an ID yet, since they are potentially new groups.
+		if group.GetProps().GetId() == "" {
+			requiredWithoutID = append(requiredWithoutID, group)
+			continue
+		}
+		requiredByID[group.GetProps().GetId()] = group
+		groupsByPropsAndRole[string(serialize.PropsKey(group.GetProps()))+group.GetRoleName()] = struct{}{}
```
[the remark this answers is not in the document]
dhaus67: So, we have three types of groups we have to deal with in `diffGroups`. The `groupsByPropsAndRole` will ensure that we dedupe newly added groups if they have the same properties + rolename as existing ones.
We need to add the required groups which are _not_ new groups to the map, ensuring that we also cover updates of existing groups that may turn out to dedupe newly added ones.

Taking the following example:
```
required groups:
- {"role": a", "props": {"id": "something1", "authProviderID": "something", "key": "email"}}
- {"role": a", "props": {"authProviderID": "something", "key": "email"}}
```
By adding the existing required group to `groupsByPropsAndRole` we ensure we don't create the second group, which would be a duplicate of the first one.

Since we are talking about it, I've added a comment explaining the thoughts behind this + a test that will handle this specifically.

central/group/datastore/validate_test.go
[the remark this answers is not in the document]
rukletsov: wat, not even cheap go unit tests?!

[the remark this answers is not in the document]
dhaus67: No sadly not. Currently branches are configured in a file statically and we can not use regex for the names.This is one major pitfall of prow.

central/group/datastore/internal/store/store_impl.go
```diff
@@ -16,36 +16,49 @@ type storeImpl struct {
...
-			}
-			if filter == nil || filter(grp.GetProps()) {
-				grps = append(grps, grp)
-			}
+// getByProps returns a group matching the given properties if it exists from the store.
+// If more than one group is found matching the properties, an error will be returned.
+// TODO(ROX-11592): This can be removed once retrieving the group by its properties is fully deprecated.
+func (s *storeImpl) getByProps(props *storage.GroupProperties) (grp *storage.Group, err error) {
+	err = s.db.View(func(tx *bolt.Tx) error {
+		groups, err := getByPropsInTransaction(tx, props)
```
rukletsov, line 42:
```suggestion
		grp, err := getByPropsInTransaction(tx, props)
```

rukletsov (approved):
One last change and we're good to go. Thank you, Daniel!

central/group/datastore/validate_test.go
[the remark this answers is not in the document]
rukletsov: 🙊 
Looks like using graphite until this is fixed is like playing russian roulette. Or we can delete all our tests and enjoy fast CI runs!

central/group/datastore/internal/store/store_impl.go
```diff
@@ -146,28 +135,119 @@ func (s *storeImpl) Mutate(toRemove, toUpdate, toAdd []*storage.Group) error {
...
+	}
+	return grps, nil
+}
+
+func getByPropsInTransaction(tx *bolt.Tx, props *storage.GroupProperties) (*storage.Group, error) {
+	grps, err := filterByPropsInTransaction(tx, props)
+	if err != nil {
+		return nil, err
+	}
+	if len(grps) == 0 {
```
rukletsov, line 246: You previously had an error case if more than one group has been found. I think you want to put it back here.
```
	if len(grps) > 1 {
		return nil, errox.InvalidArgs.Newf("multiple groups found for properties (auth provider id=%s, key=%s, "+
			"value=%s), provide an ID to retrieve a group unambiguously",
			props.GetAuthProviderId(), props.GetKey(), props.GetValue())
```

dhaus67:
Will merge this now, the UI related tests are also sadly failing within [nightlies](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-stackrox-stackrox-master-nightly-gke-ui-e2e-tests/1543822023950602240).

PR #2067, opened 2022-06-14
Add Cypress test suite for Policy Violations by Category widget

dvail:
## Description

Adds Cypress tests to verify the core sorting functionality of the widget.

## Testing notes, follow ups

- A fixture is used for testing instead of hitting the backend API because by default a fresh stackrox installation will only report three categories of policy violations, and those categories each only have a single severity type. The impact of this is that none of the filters or sorting functionality has any effect on the chart display.
- There currently are no tests that use the scope bar cluster/ns filters nor the "Policy lifecycle" option. This is because these options only impact the query sent to the server and won't effectively test functionality with mocked data.
- It would be nice for the tests to validate the display of the bars in the chart, but the Victory library renders plain `<path>` SVG elements with very little metadata that can be used for validation. I'd like to revisit this post MVP and see if we can make these tests more robust.

## Checklist

## Testing Performed

Run Cypress tests locally.

### DIFF
--- a/ui/apps/platform/cypress/integration/dashboard/dashboardSecurityMetrics.test.js
+++ b/ui/apps/platform/cypress/integration/dashboard/dashboardSecurityMetrics.test.js
@@ -1,0 +1,19 @@
+import withAuth from '../../helpers/basicAuth';
+import { hasFeatureFlag } from '../../helpers/features';
+import { visitMainDashboardPF } from '../../helpers/main';
+
+describe('Dashboard security metrics phase one action widgets', () => {
+    withAuth();
+
+    before(function beforeHook() {
+        if (!hasFeatureFlag('ROX_SECURITY_METRICS_PHASE_ONE')) {
+            this.skip();
+        }
+    });
+
+    it('should visit patternfly dashboard', () => {
+        visitMainDashboardPF();
+
+        cy.get("h1:contains('Dashboard')");
+    });
+});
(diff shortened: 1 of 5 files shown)


ui/apps/platform/cypress/integration/dashboard/dashboardSecurityMetrics.test.js
```diff
@@ -0,0 +1,97 @@
...
+    it('should visit patternfly dashboard', () => {
+        visitMainDashboardPF();
+
+        cy.get(selectors.pageHeader);
+    });
+
+    it('should sort a policy violations by category widget by severity and volume of violations', () => {
+        visitMainDashboardPF();
+
+        cy.intercept('GET', api.alerts.countsByCategory, {
```
pedrottimark, line 46: Aha, your thoughts are welcome for a clear pattern to distinguish:
* Tests of other containers that call `visitMainDashboard` need the updated function to wait on relevant requests before interactive action.
* Tests of dashboard need to mock relevant requests. If widgets are responsible for own data fetching, other requests do not matter for DOM assertions. However, they might matter to prevent next test from starting before requests have completed for current test. For example, compliance tests did not fail, but cypress side panel suggested several false starts.

pedrottimark:
Good timing. It looks like the naming convention for the support file might change in Cypress 10.

PR #2066, opened 2022-06-14
ROX-11070: Add migration for groups bucket.

dhaus67:
## Description

This is the fourth PR in a series of PRs regarding removing the composite key from the groups datastore, allowing users to specify different roles for the same group properties.

For full context, you can navigate to the previous PRs by using the comment created below.

This PR will create a migration for the `groups` bucket, adding a ID to all groups which don't have one associated with them as well as using the ID as the new key within the bucket instead of the composite key.

## Testing Performed
- see CI.

Manual testing:
- create some auth providers (number as you like), create a couple of groups per auth provider:
```bash
curl -X POST https://<endpoint>/v1/groupsbatch -H "Authorization: Bearer <token>" -d ‘{"previous_groups":[{"props":{"authProviderId”:”<auth-provider-id>”},”roleName":"Admin"}],"required_groups":[{"props":{"authProviderId":"<auth-provider-id>","key":"name","value":"someday"},"roleName":"None"},{"props":{"authProviderId":"<auth-provider-id>","key":"name","value":"someway"},"roleName":"Scope Manager"},{"props":{"authProviderId":"<auth-provider-id>","key":"email","value":"someone"},"roleName":"Vulnerability Management Requester"},{"props":{"authProviderId":"<auth-provider-id>","key":"email","value":"somewhen"},"roleName":"Admin"},{"props":{"authProviderId":"<auth-provider-id>","key":"email","value":"something"},"roleName":"Vulnerability Report Creator"},{"props":{"authProviderId":"<auth-provider-id>","key":"groups","value":"somehow"},"roleName":"Vulnerability Management Approver"},{"props":{"authProviderId":"<auth-provider-id>","key":"groups","value":"somewhere"},"roleName":"Sensor Creator"},{"props":{"authProviderId":"<auth-provider-id>","key":"userid","value":"someday"},"roleName":"None"},{"props":{"authProviderId":"<auth-provider-id>","key":"userid","value":"sometime"},"roleName":"Continuous Integration"},{"props":{"authProviderId":"<auth-provider-id>","key":"userid","value":"somewhat"},"roleName":"Analyst"},{"props":{"authProviderId":"<auth-provider-id>"},"roleName":"Admin"}]}’
```
- Ensure groups are added correctly within UI as well as returned via API:
```bash
curl https://dh0701wildlimittype.demo.stackrox.com/v1/groups  -H "Authorization: Bearer <token>" 
```
[...]

PR #2075, opened 2022-06-14
ROX-11070: Handle ID of groups within UI.

dhaus67:
## Description

This is the third PR in a series of PRs regarding removing the composite key from the groups datastore, allowing users to specify different roles for the same group properties.

For full context, you can navigate to the previous PRs by using the comment created below.

This PR will update the UI to include the ID of the default role when creating auth providers, as well as update the group object to include the ID within the properties.

## Testing Performed
- see CI (existing tests regarding auth providers should still pass).

Manual testing
1. Create an auth provider, use minimum access role `None`.
Screenshot 2022-06-15 at 01 45 30
1. Configure multiple rules with the same attribute key and values, but different role mappings.
Screenshot 2022-06-15 at 01 55 22
1. Login with the configured user and observe multiple roles being assigned.
Screenshot 2022-06-15 at 01 56 13


dhaus67:
* master
  * **PR #2062** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2062" target="_blank">Graphite</a> 
      * **PR #2075** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2075" target="_blank">Graphite</a>  👈
        * **PR #2066** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2066" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2075?utm_source=stack-comment).

ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
```diff
@@ -191,7 +191,6 @@ export function mergeGroupsWithAuthProviders(
             }
         });
     }
-
```
vjwilson, line 194: nit: Try not to let your editor take out single blank lines. This one give some visual separation of the return statement from the logic of the function, which aid in "scanning" the code late.

ui/apps/platform/src/services/GroupsService.js
```diff
@@ -23,6 +23,19 @@ export function updateOrAddGroup({ oldGroups, newGroups }) {
...
 }
 
+/**
+ * Fetches the default rule.
+ *
+ * @returns {Promise<Object, Error>} fulfilled with array of groups
+ */
+export function getDefaultGroup({ authProviderId, roleName }) {
+    return axios
+        .get(`${url}?authProviderId=${authProviderId}&key=&value=&roleName=${roleName}`)
```
vjwilson, line 33: Does the API require us to spend a `key` and `value` with blank values specifically?
Seems unnecessary--is it a limitation of the proto system?

vjwilson (commented):
Seems fine to me (just minor nit and question about API usage)

I'd like @pedrottimark to check it, too.

vjwilson (approved):
/lgtm

ui/apps/platform/src/services/GroupsService.js
rukletsov: I thought that 'empty' and 'absence' will map to the same logical situation: don't filter by this field.

ui/apps/platform/src/services/GroupsService.js
```diff
@@ -23,6 +23,23 @@ export function updateOrAddGroup({ oldGroups, newGroups }) {
...
+ */
+export function getDefaultGroup({ authProviderId, roleName }) {
+    // The default group is characterized by the following:
+    // - Only authProviderID is set, key and value are empty.
+    // - The role name of the group matches the given role name.
+    // We need _explicitly_ ask for empty key and value fields to receive the actual default role.
+    return axios
+        .get(`${url}?authProviderId=${authProviderId}&key=&value=&roleName=${roleName}`)
+        .then((response) => ({
+            response: response.data?.groups[0],
```
rukletsov, line 39: Does this mean that if _multiple_ default groups have been created (should be possible now from the backend perspective, e.g., via `curl`), a batch group request will delete all except the one that we stored here? Looks like a bug.

vjwilson: @rukletsov Can you elaborate on what you now expect for multiple default groups?

Just because it is technically possible through the API, is it a requirement, or even an option, that we need to expose to customers?

The current UI has one form input, labelled "Minimum access role" that only allows you select one role from a list. Has this requirement changed?

ui/apps/platform/src/services/GroupsService.js
dhaus67: No, surprisingly not.

Say you have two groups:
```json
{"props":{"authProviderId":"0ceed22f-026f-41a2-9d91-96410e4c4846","key":"","value":""},"roleName":"Scope Manager"}
{"props":{"authProviderId":"0ceed22f-026f-41a2-9d91-96410e4c4846","key":"email","value":"something"},"roleName":"Scope Manager"}
```

both groups are mapping to the role name `Scope Manager`.

Executing a request, i.e. via `curl`, you will retrieve the following **without explicit empty fields**:
```
curl <endpoint>/v1/groups?authProviderId=0ceed22f-026f-41a2-9d91-96410e4c4846&roleName="Scope Manager"

response:
{"groups":[{"props":{"authProviderId":"0ceed22f-026f-41a2-9d91-96410e4c4846","key":"","value":""},"roleName":"Scope Manager"},{"props":{"authProviderId":"0ceed22f-026f-41a2-9d91-96410e4c4846","key":"email","value":"something"},"roleName":"Scope Manager"}]}
```

Instead, if you execute the request with **explicitly setting key and value to empty strings**, you will receive the following:
```
curl <endpoint>/v1/groups?authProviderId=0ceed22f-026f-41a2-9d91-96410e4c4846&roleName="Scope Manager"&key=&value=

response:
{"groups":[{"props":{"authProviderId":"0ceed22f-026f-41a2-9d91-96410e4c4846","key":"","value":""},"roleName":"Scope Manager"}]}
```

TL;DR: we do need to explicitly set those, otherwise we return other groups that are not the default created one.

ui/apps/platform/src/services/GroupsService.js
dhaus67: @rukletsov 
While I agree it's technically possible to do so, I feel like we should be more strict on the datastore regarding this ambiguity.

I don't think we are ready to change requirements in the UI for this, i.e. allowing multiple "Minimum access roles" to be set (which would turn out to be multiple default groups) nor do I personally want this, from a user point of view it would be confusing.

Instead, if we decide to do anything, I feel we should restrict the API. **Iff** the client creates a group with only the auth provider ID set, it's not possible to do so when another group with only the auth provider ID set exists. This way, we wouldn't have to deal with the ambiguity in the UI, can keep it as-is, and don't have to worry about potentially removing default groups here. However, this would mean we restrict the behavior from the API.

Otherwise, if we decide to change anything in the UI, how would you even know which default group to change? Let's say you have multiple groups with the same role name (which is what you would yield here): which group is the one originally created, and which is the one not and shouldn't be changed?

pedrottimark: If you will forgive a question out of an abundance of ignorance, is there any situation in which `groups` might be an empty array? Or more relevant, any situation which a user can cause to happen?

Apparently yes, because `getGroupsWithDefault` function has conditional logic for the situation:

```js
// The default group is not yet created, meaning we have to make sure we create it here.
// Set the auth provider ID and role name.
// Use the default group if we receive a value here.
```

pedrottimark: This is one of the most important files in services folder that is still JavaScript. Adding types/group.proto.ts and rewriting this file in TypeScript might be a good collaborative follow up for UI and Merlin in 72.

dhaus67: @pedrottimark Yes, this may happen.
It could be, as you rightly mentioned, the auth provider is currently created and the default group has not been created yet _or_ the default group has been deleted via API (which is unlikely, but still possible).

I agree on the collaborative effort to migrate this to typescript, it would certainly make life better. I've created [ROX-11600](https://issues.redhat.com/browse/ROX-11600) for that with the aim to complete it within 3.72.

rukletsov: @vjwilson @dhaus67 Let's restrict the API to at most one default rule / group per auth provider. @dhaus67 could you please send a new PR in this chain for this?

rukletsov (approved):

PR #2081, opened 2022-06-15
ROX-11341 Dump sensor's messages in binary format

lvalerom:
## Description

This PR aims to make the local-sensor output file easier to unmarshal in the automatic tests.

## Checklist

## Testing Performed

* The file is create in binary format with:

```
go run tools/local-sensor/main.go -binary
```

* Comparing that the results of `central-out.json` and `central-out.bin` are the same.

### DIFF
--- a/tools/local-sensor/main.go
+++ b/tools/local-sensor/main.go
@@ -2,14 +2,18 @@ package main
 
 import (
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
 	"log"
 	"net"
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -61,6 +65,7 @@ func createConnectionAndStartServer(fakeCentral *centralDebug.FakeService) (*grp
 
 type localSensorConfig struct {
 	Duration           time.Duration
+	BinaryOutput       bool
 	CentralOutput      string
 	RecordK8sEnabled   bool
 	RecordK8sFile      string
@@ -76,6 +81,7 @@ func mustGetCommandLineArgs() localSensorConfig {
 	sensorConfig := localSensorConfig{
 		Verbose:            false,
 		Duration:           0,
+		BinaryOutput:       false,
 		CentralOutput:      "central-out.json",
 		RecordK8sEnabled:   false,
 		RecordK8sFile:      "k8s-trace.jsonl",
@@ -88,6 +94,7 @@ func mustGetCommandLineArgs() localSensorConfig {
 	flag.BoolVar(&sensorConfig.Verbose, "verbose", sensorConfig.Verbose, "prints all messages to stdout as well as to the output file")
 	flag.DurationVar(&sensorConfig.Duration, "duration", sensorConfig.Duration, "duration that the scenario should run (leave it empty to run it without timeout)")
 	flag.StringVar(&sensorConfig.CentralOutput, "central-out", sensorConfig.CentralOutput, "file to store the events that would be sent to central")
+	flag.BoolVar(&sensorConfig.BinaryOutput, "binary", sensorConfig.BinaryOutput, "store all the events that would be sent to central in binary format")
 	flag.BoolVar(&sensorConfig.RecordK8sEnabled, "record", sensorConfig.RecordK8sEnabled, "whether to record a trace with k8s events")
 	flag.StringVar(&sensorConfig.RecordK8sFile, "record-out", sensorConfig.RecordK8sFile, "a file where recorded trace would be stored")
 	flag.BoolVar(&sensorConfig.ReplayK8sEnabled, "replay", sensorConfig.ReplayK8sEnabled, "whether to reply recorded a trace with k8s events")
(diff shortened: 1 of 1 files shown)


tools/local-sensor/main.go
```diff
@@ -88,6 +94,7 @@ func mustGetCommandLineArgs() localSensorConfig {
 	flag.BoolVar(&sensorConfig.Verbose, "verbose", sensorConfig.Verbose, "prints all messages to stdout as well as to the output file")
 	flag.DurationVar(&sensorConfig.Duration, "duration", sensorConfig.Duration, "duration that the scenario should run (leave it empty to run it without timeout)")
 	flag.StringVar(&sensorConfig.CentralOutput, "central-out", sensorConfig.CentralOutput, "file to store the events that would be sent to central")
+	flag.BoolVar(&sensorConfig.BinaryOutput, "binary", sensorConfig.BinaryOutput, "store all the events that would be sent to central in binary format")
```
fredrb, line 97: Maybe a better UI would be to allow a `-format` (or `-output-format`) param. The default would always be `json`, but you could provide `raw` or `bytes` and get the binary encoding.

tools/local-sensor/main.go
```diff
@@ -243,9 +250,24 @@ type sensorMessagesJSONOutput struct {
 	MessagesFromSensor []*central.MsgFromSensor `json:"messages_from_sensor"`
 }
 
-func dumpMessages(messages []*central.MsgFromSensor, start, end time.Time, outfile string) {
+func dumpMessages(messages []*central.MsgFromSensor, start, end time.Time, outfile string, binaryOutput bool) {
 	dateFormat := "02.01.15 11:06:39"
 	log.Printf("Dumping all sensor messages to file: %s\n", outfile)
+	if binaryOutput {
+		fname := strings.TrimSuffix(outfile, filepath.Ext(outfile))
+		file, err := os.OpenFile(fmt.Sprintf("%s.bin", fname), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
fredrb, line 258: If `-central-out` is `result.bin` the two files would overwrite.

tools/local-sensor/main.go
```diff
@@ -243,9 +250,24 @@ type sensorMessagesJSONOutput struct {
 	MessagesFromSensor []*central.MsgFromSensor `json:"messages_from_sensor"`
 }
 
-func dumpMessages(messages []*central.MsgFromSensor, start, end time.Time, outfile string) {
+func dumpMessages(messages []*central.MsgFromSensor, start, end time.Time, outfile string, binaryOutput bool) {
 	dateFormat := "02.01.15 11:06:39"
 	log.Printf("Dumping all sensor messages to file: %s\n", outfile)
+	if binaryOutput {
```
fredrb, line 256: I think it's better to do only one at a time with the `-format` param as I mentioned before. Unless there is a need to output both files for the same scenario.

lvalerom: The only reason for having both at the same time (that I can think of) is in case you want to use that file in a tests but you also want to check that the file contains something. In this scenario it would be nice to have both. What do you think? Should we maybe default in both and if the user specifies `-format` then just write the specified format?

tools/local-sensor/main.go
lvalerom: Honestly, I didn't think someone would name a `json` file with a `.bin` extension 😅

tools/local-sensor/main.go
lvalerom: I like this. But check my other comment and let me know what you think about having a way of writing both formats at the same time.

tools/local-sensor/main.go
fredrb: I wouldn't be surprised 😛

tools/local-sensor/main.go
fredrb: > Should we maybe default in both and if the user specifies -format then just write the specified format?

I don't get this suggestion.

fredrb: Well, since you're writing the assertions for this test, I trust your judgment if you think it's going to be useful to have a sibling JSON file with human-readeble content. I don't have a strong opinion tbh.

fredrb (approved):

tools/local-sensor/main.go
lvalerom: I think you're right. I've changed the code to just write the file in the specified format. Let me know what you think.

tools/local-sensor/main.go
```diff
@@ -77,11 +74,25 @@ type localSensorConfig struct {
...
 }
 
+const (
+	jsonFormat string = "json"
+	rawFormat  string = "raw"
+)
+
+func isValidOutputFormat(format string) bool {
+	var validFormats = map[string]struct{}{
+		jsonFormat: {},
```
fredrb, line 84: A ridiculous nitpick: this map could hold the "formatter" function that receives the outfile path. You can still use it to check if the format is valid and use the static function to do the work at the end. I.e. remove the `switch` case.

fredrb (approved):

tools/local-sensor/main.go
lvalerom: Done!

fredrb: Amazing. Thanks!

PR #2085, opened 2022-06-15
ROX-11169: Remove external plugin usages in e2e tests

janisz:
## Description

Remove usages of external plugin in e2e tests as it's deprecated.

## Testing Performed

CI


gavin-stackrox (approved):

rhybrillou (approved):

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
```diff
@@ -37,13 +37,6 @@ class IntegrationsSplunkViolationsTest extends BaseSpecification {
     private static final String TEST_NAMESPACE = "qa-splunk-violation"
     private static final String SPLUNK_INPUT_NAME = "stackrox-violations-input"
 
-    def setupSpec() {
-        // when using "Analyst" api token to access central Splunk violations endpoint
-        // authorisation plugin prevents violations from being returned
-        // this leads to no violations being propagated to Splunk
-        disableAuthzPlugin()
-    }
```
msugakov, line 45: This removal may break Splunk test. I added a missing label. Please re-run the pipeline.

msugakov (dismissed):

qa-tests-backend/src/test/groovy/IntegrationsSplunkViolationsTest.groovy
janisz: I don't think it will break anything as plugin is not enabled so there is no need to disable it. Let's wait for tests.

janisz: OK, this test passed but there are other failures in tests that do require auth plugin. I'll move them to builtin plugin

msugakov: Thanks for checking the Splunk test! I removed my blocking review.
I don't know the background of this change, what plugins are and how they work. I'd defer to other reviewers. Feel free to ship it once you're confident.

janisz: Ip updated the description to give more context

PR #2086, opened 2022-06-15
Dashrews/ROX-11272 postgres admin database name

dashrews78:
## Description

As I was working through the backup/restore of Postgres, it became apparent that we needed an admin pool and an admin database.  It made sense to reserve the "postgres" database this and name the primary database with the contents something else.  So I changed that so it will default to central_active and a restore will use central_restore.  This will set us up for dealing with the upgrade and rollback replicas that will come in ROX-9884. 

## Checklist

## Testing Performed

Ran through all the postgres unit tests.  Ran through backup/restore scenarios to ensure that the database name was what it needed to be.  Logged into the DB to ensure that the postgres database did not have tables associated with it.

### DIFF
--- a/central/globaldb/v2backuprestore/restore/postgres.go
+++ b/central/globaldb/v2backuprestore/restore/postgres.go
@@ -118,5 +118,5 @@ func CheckIfRestoreDBExists(dbConfig *pgxpool.Config) bool {
 
 func getRestoreDBName() string {
 	// Build the active database name for the connection
-	return fmt.Sprintf("%s%s", config.GetConfig().CentralDB.RootDatabaseName, restoreSuffix)
+	return fmt.Sprintf("%s%s", config.GetConfig().CentralDB.DatabaseName, restoreSuffix)
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -18,7 +18,7 @@ var (
 	defaultBucketFillFraction = 0.5
 	defaultCompactionState    = true
 	defaultDBSource           = "host=central-db.stackrox port=5432 user=postgres sslmode=verify-full statement_timeout=600000 pool_min_conns=1 pool_max_conns=90"
-	defaultRootDatabase       = "central"
+	defaultDatabase           = "central"
 
 	once   sync.Once
 	config *Config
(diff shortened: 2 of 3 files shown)


image/templates/helm/stackrox-central/templates/01-central-08-configmap.yaml
```diff
@@ -14,10 +14,10 @@ data:
     {{- ._rox.central._config | nindent 4 }}
     {{- if ._rox.central.enableCentralDB }}
     centralDB:
+      rootDatabaseName: central
```
c-du, line 17: When do you plan to change this config?

Why call it root database? 
It implies the databases have a tree-like structure but I think it does not. How about centralDatabaseName?

central/globaldb/postgres.go
```diff
@@ -75,13 +68,27 @@ var (
...
 		if err != nil {
 			log.Fatalf("Could not parse postgres config: %v", err)
 		}
 
+		// Get the active database name for the connection
+		activeDB := pgconfig.GetActiveDB()
+
+		// Create the active database if necessary
+		if !pgadmin.CheckIfDBExists(dbConfig, activeDB) {
+			err = pgadmin.CreateDB(sourceMap, dbConfig, pgadmin.AdminDB, activeDB)
```
c-du, line 81: We do not need to try to create the database if it does not exists. We let the migrator ensure that the current active database presents.

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -71,3 +81,52 @@ func RenameDB(adminPool *pgxpool.Pool, originalDB, newDB string) error {
...
+		return false
+	}
+
+	log.Infof("%q database exists => %t", dbName, exists)
+	return exists
+}
+
+// AdminPool - returns a pool to connect to the admin database.
+// This is useful for renaming databases such as a restore to active.
+func AdminPool(config *pgxpool.Config) *pgxpool.Pool {
```
c-du, line 110: nit: GetAdminPool

c-du (commented):
Looks good to my overall. With some comments.

image/templates/helm/stackrox-central/templates/01-central-08-configmap.yaml
dashrews78: removed the database name from the config

c-du (approved):
LGTM. Thanks!

PR #2092, opened 2022-06-15
[ROX-11046] : Add label selector and last updated timestamp fields to decommissioned clusters retention config

charmik-redhat:
## Description

- Added type `ignore_label` field to specify key and value strings for the cluster label to ignore from auto removal.
- Added `last_updated_time` field to store the last time decommissioned cluster retention config was updated. This will be used to provide a 24 hour grace period after a change to config, so that new clusters that became eligible for deletion after the config change do not get deleted without sufficient warning.

## Checklist

#### Checked if the `/v1/config` API returns the config with decommissioned clusters retention values set to defaults
   
74270A80-D64C-4CF3-A407-C3DAFFA80C58
  
#### Checked if last updated timestamp is not updated if there is no change to cluster retention config
  
C085163A-68EC-4933-ABA7-C0CCE4FE7213_1_105_c

#### Checked if updating the cluster retention config updates the last updated timestamp as well

D3EE6310-F1D0-4035-B86B-3F3DCFEDA109_1_105_c

1598A4C8-8856-438C-8122-E2AA9DF7F9BD_1_105_c


proto/storage/config.proto
```diff
@@ -52,7 +66,7 @@ message PrivateConfig {
     }
     int32                    image_retention_duration_days                  = 2;
     int32                    expired_vuln_req_retention_duration_days       = 4;
-    int32                    decommissioned_cluster_retention_duration_days = 5;
+    DecommissionedClusterRetentionConfig decommissioned_cluster_retention   = 5;
```
charmik-redhat, line 69: Would this change need a separate changeling entry?

proto/storage/config.proto
```diff
@@ -43,6 +45,18 @@ message AlertRetentionConfig {
...
 
+// next available tag: 4
+message DecommissionedClusterRetentionConfig {
+    // next available tag: 3
+    message IgnoreClusterLabel {
+        string key    = 1;
+        string value  = 2;
+    }
+    int32 retention_duration_days           = 1;
+    IgnoreClusterLabel ignore_label         = 2;
```
md2119, line 56: nit: `ClusterLabel ignore_cluster_label`. Why are not multiple labels supported?

central/config/datastore/datastore.go
```diff
@@ -50,6 +51,60 @@ func (d *datastoreImpl) UpsertConfig(ctx context.Context, config *storage.Config
...
+
+func (d *datastoreImpl) getClusterRetentionConfig(ctx context.Context) (*storage.DecommissionedClusterRetentionConfig, error) {
+	conf, err := d.GetConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if privateConf := conf.GetPrivateConfig(); privateConf != nil {
+		return privateConf.GetDecommissionedClusterRetention(), nil
+	}
+	return nil, nil
```
md2119, line 86: Getters handle nil check. 
nit: `return conf.GetPrivateConfig().GetDecommissionedClusterRetention(), nil`

central/config/datastore/datastore.go
```diff
@@ -50,6 +51,60 @@ func (d *datastoreImpl) UpsertConfig(ctx context.Context, config *storage.Config
...
+		return true
+	}
+	if c1 == nil || c2 == nil {
+		return false
+	}
+	return c1.GetRetentionDurationDays() == c2.GetRetentionDurationDays() &&
+		ignoreLabelsEqual(c1.GetIgnoreLabel(), c2.GetIgnoreLabel())
+}
+
+func ignoreLabelsEqual(l1 *storage.DecommissionedClusterRetentionConfig_IgnoreClusterLabel,
```
md2119, line 101: nit: `LabelsEqual`

proto/storage/config.proto
clickboo: +1, should be an array of labels. The UI will need to have some kind of delimiter in the textbox to separate multiple pairs. Assuming setting this label is available in the Add/Edit cluster workflow and the UI/UX on that is TBD like you alluded to in the standup today.

central/config/datastore/datastore.go
clickboo: That would have 💣 if `GetPrivateConfig` can return nil but it cannot be:
https://github.com/stackrox/stackrox/blob/8864b67f8b7f5c64601dcdabb88cd33e122ee9e0/central/config/datastore/singleton.go#L77

Hence the chaining will work.

proto/storage/config.proto
pedrottimark: It seems like a question for solution architects to answer how realistic the need, then for product to confirm a change in feature scope if needed.

Strictly from UI perspective, if it goes forward, for consistency with cluster labels, please imitate:
https://github.com/stackrox/stackrox/blob/master/proto/storage/cluster.proto#L100

Therefore, in response JSON:
* no label similar to current design:
    ```json
    "ignoreLabels": null
    ```
* one label:
    ```json
    "ignoreLabels": {"key1": "value1"}
    ```
* multiple labels:
    ```json
    "ignoreLabels": {"key1": "value1","key2":"value2"}
    ```

pedrottimark:
Last updated time will be added as supporting information to System Configuration view and form.

charmik-redhat:
> Last updated time will be added as supporting information to System Configuration view and form.

It can be added to the view (not to the form) but it is not a must have for UI. The purpose of last updated time is to provide a minimum 24 hr grace period after a recent change to config before removing any clusters.

central/config/datastore/singleton.go
```diff
@@ -79,6 +85,16 @@ func initialize() {
 			PublicConfig:  config.GetPublicConfig(),
 			PrivateConfig: &defaultPrivateConfig,
 		}))
+		return
+	}
+
+	if config.GetPrivateConfig().GetDecommissionedClusterRetention() == nil {
```
charmik-redhat, line 91: @clickboo Should this change go to migrations, or is it fine here?

charmik-redhat: I mean, should this population logic for retention config be added to migrator/migrations somewhere instead of here? 

New installs would be taken care of by the block above this.

clickboo: There should be a migration for existing clusters with the defaults added. And for new installs, like you mention, this block should remain.

charmik-redhat: Thanks, I will add that in a separate task and get this in so that UI development can move forward.

proto/storage/config.proto
charmik-redhat: Yeah thinking more about it, I think that makes sense to allow multiple labels in the system config rather than asking a user to add a new label to the cluster/s. I will make the change and push it.

central/config/datastore/singleton.go
md2119: I don't think migration is needed. In this init function you are pulling out the config from the DB and writing a non-empty value if not set. We always expect a non-zero value for retention. Therefore, once you have written it once, this condition will not be true again. 

I recommend refactoring the code to not write twice `ln84` and `ln94`.

charmik-redhat: Done

proto/storage/config.proto
charmik-redhat: Done

central/config/datastore/singleton.go
```diff
@@ -50,8 +56,7 @@ var (
 				AttemptedRuntimeRetentionDurationDays: DefaultAttemptedRuntimeAlertRetention,
 			},
 		},
-		ExpiredVulnReqRetentionDurationDays:        DefaultExpiredVulnReqRetention,
-		DecommissionedClusterRetentionDurationDays: DefaultDecommissionedClusterRetention,
+		ExpiredVulnReqRetentionDurationDays: DefaultExpiredVulnReqRetention,
```
md2119, line 59:
```suggestion
		ExpiredVulnReqRetentionDurationDays: DefaultExpiredVulnReqRetention,
		DecommissionedClusterRetention: &storage.DecommissionedClusterRetentionConfig{
		RetentionDurationDays: DefaultDecommissionedClusterRetention,
		LastUpdated:           types.TimestampNow(),
	}
```

central/config/datastore/singleton.go
md2119:
```suggestion
		privateConfig := config.GetPrivateConfig()
		if privateConfig == nil {
			privateConfig = &defaultPrivateConfig
		} else if config.GetPrivateConfig().GetDecommissionedClusterRetention() == nil {
		        privateConfig.DecommissionedClusterRetention = defaultClusterRetentionConfig.GetDecommissionedClusterRetention()
		}
		
```

md2119 (approved):

PR #2094, opened 2022-06-15
Fix blevebindings script to not use GOPATH

viswajithiii:
## Description

It was writing files to `$GOPATH/src/$packageName`, which is not what we want. It just happened to work in CI.

## Checklist

## Testing Performed

Local testing


connorgorman (commented):
while you're here, can you change the wrapper to use python3 :)

viswajithiii:
> while you're here, can you change the wrapper to use python3 :)

I took a look, we just call `python` and the code is written in a way that's compatible with either version. I don't know if we should just blindly change that to `python3`, since I'm not sure if the standard is to always have a `python3` binary available.

connorgorman:
> I took a look, we just call python and the code is written in a way that's compatible with either version. I don't know if we should just blindly change that to python3, since I'm not sure if the standard is to always have a python3 binary available.

Python2 is completely sunset and I've already changes a bunch of the other bindings to python3 and no one has hollered. Python binary isn't installed on any macs anymore :(

viswajithiii:
> > I took a look, we just call python and the code is written in a way that's compatible with either version. I don't know if we should just blindly change that to python3, since I'm not sure if the standard is to always have a python3 binary available.
> 
> Python2 is completely sunset and I've already changes a bunch of the other bindings to python3 and no one has hollered. Python binary isn't installed on any macs anymore :(

Oh, I always thought they would make `python` point to `python3` eventually? Do they not?

connorgorman:
```
 python
-bash: python: command not found
```

viswajithiii:
> > > I took a look, we just call python and the code is written in a way that's compatible with either version. I don't know if we should just blindly change that to python3, since I'm not sure if the standard is to always have a python3 binary available.
> > 
> > 
> > Python2 is completely sunset and I've already changes a bunch of the other bindings to python3 and no one has hollered. Python binary isn't installed on any macs anymore :(
> 
> Oh, I always thought they would make `python` point to `python3` eventually? Do they not?

In any case, I've updated it. Didn't realize you had change some of the others already.

clickboo (approved):

PR #2096, opened 2022-06-15
ROX-11405: Add functionality to Postgres for search after

connorgorman:
## Description

This should fix the Splunk TA checks

## Checklist

## Testing Performed

Unit tests, plus e2e for splunk

### DIFF
--- a/pkg/postgres/walker/walker.go
+++ b/pkg/postgres/walker/walker.go
@@ -278,9 +278,11 @@ func handleStruct(ctx context, schema *Schema, original reflect.Type) {
 			continue
 		}
 		opts := getPostgresOptions(structField.Tag.Get("sql"), schema.Parent == nil, ctx.ignorePK, ctx.ignoreUnique, ctx.ignoreFKs)
-
 		if opts.Ignored {
 			continue
+		}
+		if opts.PrimaryKey && structField.Type.Kind() != reflect.String {
+			panic("only string primary keys are supported due to DocID queries")
 		}
 
 		searchOpts, derivedFields := getSearchOptions(ctx, structField.Tag.Get("search"))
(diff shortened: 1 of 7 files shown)


pkg/search/postgres/common.go
```diff
@@ -194,24 +195,55 @@ func parsePaginationQuery(pagination *v1.QueryPagination, schema *walker.Schema,
...
+			table, column = dbField.Schema.Table, dbField.ColumnName
+		}
+		if idx != 0 && so.GetSearchAfter() != "" {
+			return parsedPaginationQuery{}, errors.New("search after for pagination must be defined for only the first sort option")
 		}
+
 		out.OrderBys = append(out.OrderBys, orderByEntry{
 			Field: pgsearch.SelectQueryField{
-				SelectPath: qualifyColumn(dbField.Schema.Table, dbField.ColumnName),
-				FieldType:  dbField.DataType,
```
viswajithiii, line 205: Hmm, I think the `FieldType` is required for the deserialization to work

connorgorman: Ah yeah, I guess it needs to be contained in the output because we deserialize it then potentially throw it away? The issue is that the ID doesn't have a walker.DataType which is why I did that

connorgorman: I just assume ID is a string. This is a rare code path and can be removed once we deprecate bleve as the cost of search on a "DocID" is no different in postgres since the column must be top level

viswajithiii: Yeah we can also enforce that IDs are strings at the walker level

viswajithiii (approved):
Seems great! Can we disallow id fields that are not strings in the generator, to make sure that assumption always holds?

pkg/search/postgres/common.go
connorgorman: done

pkg/postgres/walker/walker.go
```diff
@@ -278,10 +278,12 @@ func handleStruct(ctx context, schema *Schema, original reflect.Type) {
 			continue
 		}
 		opts := getPostgresOptions(structField.Tag.Get("sql"), schema.Parent == nil, ctx.ignorePK, ctx.ignoreUnique, ctx.ignoreFKs)
-
 		if opts.Ignored {
 			continue
 		}
+		if opts.PrimaryKey && structField.Type.Kind() != reflect.String {
```
md2119, line 284: All children have idx (int) which is a primary key

connorgorman: These aren't in the walker

PR #2099, opened 2022-06-16
ROX-11308: Minimum access role incorrectly displayed when auth provider is created via API

parametalol:
## Description

UI set a hardcoded value 'Admin' for when there is no role group defined for a given auth provider, which looked misleading, as the provider still couldn't be used to login because of the missing role error.

The PR changes the value to 'Undefined, assuming None', so the behavior of such provider is functionally identical to one with the minimum access role set to 'None'.

For a new provider being created via UI the value will be set to 'Admin' as before.

Provider without minimum role will still show empty value in the list of providers in the UI, which seems logical for such providers.

## Checklist

## Testing Performed

Manual testing showed same login error for a provider without minimum access role and with 'None':
* Normal login;
* Test login on the provider editor page.

An example of a provider created via API without minimum access role defined:
image

### DIFF
--- a/ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
+++ b/ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
@@ -196,7 +196,7 @@ export function mergeGroupsWithAuthProviders(
 }
 
 export function getDefaultRoleByAuthProviderId(groups: Group[], id: string): string {
-    let defaultRoleGroups = groups.filter(
+    const defaultRoleGroups = groups.filter(
         (group) =>
             group.props &&
             group.props.authProviderId &&


ivan-degtiarenko (commented):
Can you add screenshots of "before" vs "after" + add @vjwilson as a reviewer?

0x656b694d:
Screenshots "before" could be found in the Jira ticket. Well, the diff is only in the "Minimum access role" field value.

vjwilson (approved):
/lgtm

ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
```diff
@@ -212,5 +212,5 @@ export function getDefaultRoleByAuthProviderId(groups: Group[], id: string): str
     if (defaultRoleGroups.length) {
```
rukletsov, line 212: Global default is not a thing any more. If there is no default role, we should use `Undefined, assuming None`.

ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
```diff
@@ -212,5 +212,5 @@ export function getDefaultRoleByAuthProviderId(groups: Group[], id: string): str
...
         return defaultRoleGroups[0].roleName;
     }
-    return 'Admin';
+    return id ? 'Undefined, assuming None' : 'Admin';
```
rukletsov, line 215: What is the use case for the `: 'Admin'` case?

0x656b694d: It is selected when creating a new provider from the UI.

0x656b694d:
Unfortunately, looks like it's too complicated to hack the list with virtual values without breaking the other form logic.

rukletsov (approved):

vjwilson (approved):
Assuming Mark helped you through the bug

PR #2103, opened 2022-06-16
Disable Splunk TA for Postgres until search after is implemented

connorgorman:
## Description

SearchAfter is required for the Splunk TA, but it has not been implemented yet. https://github.com/stackrox/stackrox/pull/2096 will fix this, but there are going to be some challenges so better to stop the noise 

## Checklist

## Testing Performed

e2e in postgres-tests


dashrews78 (approved):
LGTM

PR #2109, opened 2022-06-16
ROX-10980: Use CVE and not CVE ID in vulnerability requests

sachaudh:
## Description

JIRA + Context: https://issues.redhat.com/browse/ROX-10980

* Made a change in the GraphQL query to not create the `id` by aliasing the `cve` field.
* Using the non-deprecated `imageVulnerabilities` and `imageComponents` GraphQL fields instead of `vulns` and `components`. Was not aware they were deprecated 
* Apparently the row actions used the `cve` while the bulk actions used the `id` field. It wasn't a problem since they were the same value, but this should make it consistent when the Backend change happens

## Checklist

## Testing Performed

* No visual changes
* Can't really test until backend makes changes. I suggest the backend uses the changes in this PR when they make the `cve` and `id` different, or we can push this in and backend can check it separately


dvail (approved):
LGTM based on the context I have from that ticket.

Good call on the deprecated fields too, looks like I have something to update in some of my recent code.

PR #2115, opened 2022-06-17
ROX-11166: composite data for indirectly scoped resources

rhybrillou:
## Description

This change aims at generating pseudo-realistic dackbox datasets for SAC integration testing.
Generating realistic datasets is time consuming and not necessarily helpful when it comes to SAC testing, as it in not easy to visualize how much data is present in the storage, nor how the items are related between themselves.

Having a simple, low-cardinality data set allow to more easily focus on what items should be reachable and which not when it comes to scoped access control and scoped data filtering testing.

## Checklist

## Testing Performed

This change aims at creating a dataset for testing purposes. Actual tests will come in a later pull request.
CI run should be enough.

### DIFF
--- a/pkg/fixtures/dackbox.go
+++ b/pkg/fixtures/dackbox.go
@@ -1,0 +1,722 @@
+package fixtures
+
+import (
+	"github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+///////////////////////////////
+// Dackbox synthetic dataset //
+///////////////////////////////
+
+/////////////////////
+// Image with CVEs //
+
+// Data relationships
+//
+//     Namespace (scope information holder)
+//         ^ 1
+//         |
+//         v *
+//     Deployment
+//         ^ *
+//         |
+//         v *
+//       Image <--------------+ *
+//         ^ *                |
+//         |                  |
+// ImageComponentEdge         |
+//         |                  |
+//         v *                |
+//   ImageComponent      ImageCVEEdge
+//         ^ *                |
+//         |                  |
+//  ComponentCVEEdge          |
+//         |                  |
+//         v *                |
+//        CVE <---------------+ *
+//
+// Three types of data have to be injected here:
+// - NamespaceMetadata
+// - Deployment
+// - Image
+//
+// * NamespaceMetadata is used to provide the scope information
+//
+// * The Deployment links namespace (scope) with deployed images (possibly using multiple containers
+// with one image per container. The reference to the image contains the image ID (sha image hash),
+// as well as image name information.
+//
+// * The image information is injected in the form of an image with scan information.
+// ** The Image Scan field contains the data that will be used to fill the dackbox storage
+// ** The Image Scan contains EmbeddedImageScanComponent objects, which are used to populate
+// ImageComponent as well as ImageComponentEdge storage entities.
+// ** EmbeddedImageScanComponent contains in turn EmbeddedVulnerability objects which contain CVE data
+// and are used to populate the CVE, ComponentCVEEdge and ImageCVEEdge storage entities.
+//
+//
+// For testing purposes, a graph of objects like the one below could be used.
+//
+// Cluster1 -- NamespaceA -- Deployment1 -- Image1 --+--> ImageComponent1 --+--> ImageCVE1
+//                                                   |                      |
+//                                                   |                      +--------+
+//                                                   |                               |
+//                                                   |                               v
+//                                                   |                           ImageCVE2
+//                                                   |                               ^
+//                                                   |                               |
+//                                                   |                               +--------+
+//                                                   |                                        |
+//                                                   +--> ImageComponent2 -----> ImageCVE3    |
+//                                                   |                                        |
+//                                                   +-----------+                            |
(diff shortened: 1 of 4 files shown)


pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1128 @@
...
+//                                                               |          +--> ImageCVE5    |
+//                                                   +-----------+                            |
+//                                                   |                                        |
+//                                                   +--> ImageComponent4                     |
+//                                                   |                                        |
+//                                                   |                      +-----------------+
+//                                                   |                      |
+// Cluster2 -- NamespaceB -- Deployment2 -- Image2 --+--> ImageComponent5 --+--> ImageCVE6
+//                                                                          |
+//                                                                          +--> ImageCVE7
```
janisz, line 86: ❤️

pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1128 @@
...
+// ImageComponentEdge         |
+//         |                  |
+//         v *                |
+//   ImageComponent      ImageCVEEdge
+//         ^ *                |
+//         |                  |
+//  ComponentCVEEdge          |
+//         |                  |
+//         v *                |
+//        CVE <---------------+ *
```
janisz, line 37: Maybe we should have a file with fill schema diagram generated from source? What do you think about it?

pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1162 @@
...
+)
+
+///////////////////////////////
+// Dackbox synthetic dataset //
+///////////////////////////////
+
+/////////////////////
+// Image with CVEs //
+
+// Data relationships
```
dhaus67, line 17: not related at all: appreciate the diagram, helps better in understanding things 😸

pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1162 @@
...
+			Score:               5.9,
+			Severity:            storage.CVSSV3_MEDIUM,
+		},
+		PublishedOn:       &types.Timestamp{Seconds: 1234567890},
+		LastModified:      &types.Timestamp{Seconds: 1235467890},
+		VulnerabilityType: storage.EmbeddedVulnerability_IMAGE_VULNERABILITY,
+		VulnerabilityTypes: []storage.EmbeddedVulnerability_VulnerabilityType{
+			storage.EmbeddedVulnerability_IMAGE_VULNERABILITY,
+		},
+		Suppressed:            false,
```
dhaus67, line 437: <strike>Would it be worthwhile to add suppressed CVE's as well?</strike>

I see we haven't implemented that yet.

pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1162 @@
...
+				Id:   "169b0d3f-8277-4900-bbce-1127077defae",
+				Name: "Stackrox Scanner",
+			},
+			Notes: []storage.ImageScan_Note{},
+		},
+		SignatureVerificationData: nil,
+		Signature:                 nil,
+		SetComponents:             &storage.Image_Components{Components: 3},
+		SetCves:                   &storage.Image_Cves{Cves: 5},
+		SetFixable:                &storage.Image_FixableCves{FixableCves: 2},
```
dhaus67, line 629: Not sure I'm reading things right, but shouldn't it be logically 5 fixable CVE's?

All three embedded image components have the `FixedBy` field set, all three components are different and `1_1 & 1s2_3` have each respectively 2 CVE's and `1_2` has 1 CVE, or are my assumptions wrong that they are not somehow fixed?

pkg/fixtures/dackbox.go
```diff
@@ -0,0 +1,1162 @@
...
+			Components: []*storage.EmbeddedNodeScanComponent{
+				GetEmbeddedNodeComponent_1_1(),
+				GetEmbeddedNodeComponent_1_2(),
+				GetEmbeddedNodeComponent_1s2_3(),
+			},
+			Notes: nil,
+		},
+		SetComponents: &storage.Node_Components{Components: 3},
+		SetCves:       &storage.Node_Cves{Cves: 5},
+		SetFixable:    &storage.Node_FixableCves{FixableCves: 2},
```
dhaus67, line 1102: I guess same question here as above: in the current mock data, how is a node CVE classified as fixable or not fixable? I'd expect the `storage.EmbeddedVulnerability.SetFixedBy` to be it?

pkg/fixtures/dackbox.go
rhybrillou: Good point.
I think I'd use the datastore Suppress/Unsuppress functions to test the behaviour, though.

pkg/fixtures/dackbox.go
rhybrillou: To be honest, I did not dig that deep in the effort to provide pseudo-realistic test data.
My guess would be that the `SetFixedBy` field in the CVEs from the graph is what is actually used to consider a CVE as fixed. I think the `FixableCves` value is the count of visible CVEs that are flagged as fixable.
Although the values are not relevant in the tests I intend to run, I adjusted the dataset to be consistent.

dhaus67 (approved):

PR #2117, opened 2022-06-17
ROX-9843: Add expose monitoring flag to operator

mtodor:
## Description

With this PR we are adding an option to operator configuration to enable monitoring for Central. This is required for proper integration with the observability operator installed together with ACS Fleetshard. Related PR on ASC Fleet manager: https://github.com/stackrox/acs-fleet-manager/pull/112

The primary function of the operator is to provide values for the Helm template, and everything will be installed/updated via helm. Our helm template already provides `central.exposeMonitoring` boolean value, which is used to add the monitoring port to the central pod: https://github.com/stackrox/stackrox/blob/210afd36677ab3e6b66ca6dc512d3d075b2450ff/image/templates/helm/stackrox-central/templates/_central_endpoints.tpl#L47-L50  and also to deploy relevant network policy for it https://github.com/stackrox/stackrox/blob/210afd36677ab3e6b66ca6dc512d3d075b2450ff/image/templates/helm/stackrox-central/templates/01-central-10-networkpolicy.yaml#L21-L26 This PR provides a possibility to set that value (flag) via operator CRD.

**Made changes**

- new flag added to the operator configuration
- tests when the flag is enabled and disabled

**Open question**

- Should we add information about the new flag to Changelog?

## Checklist
- ~[ ] Determined and documented upgrade steps~ - no need for upgrade steps
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~ - also not relevant

## Testing Performed 

1. Letting CI run the KUTTL tests
2. Start OS4 cluster with `infractl`
3. Deploy operator to the new cluster (using images built by CI pipeline). i.e.
```
VERSION=3.70.0-414-gd49bd107a4 make deploy
```
**NOTE** It can happen that we need to adjust pull image secrets if the cluster is not able to pull images
4. Create a Central CRD resource (without monitoring first). First, we need a new project and then create CRD inside that project. An example can be taken from the test: `operator/tests/central/basic/10-central-cr.yaml`
5. Check that everything is deployed properly
6. Set `exposeMonitoring` flag on the previously created Central resource
7. Validate that the central pod has an additional port and that network policy is created

**Test full data plane observability flow**
1. Deploy observability helm provided by PR: https://github.com/stackrox/acs-fleet-manager/pull/112
2. Validate that the observability operator is installed properly and that metrics are collected. For that, we need to open OS4 UI and navigate to Routes of the project where the observability stack is deployed
3. in Prometheus UI - `/graph`, check if `rox_*` metrics are visible
4. checking other parts like Grafana, Observatorium, and Alert Manager are not relevant for this PR. It's important that Prometheus is able to collect metrics, everything else relies on that.


stehessel:
Looks good, but maybe I'm missing something. What code actually sets the port? I only see it mentioned in the operator tests.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -109,6 +109,12 @@ type CentralComponentSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=4
 	Persistence *Persistence `json:"persistence,omitempty"`
 
+	// Enable necessary components for provisioning of monitoring metrics for
+	// Central and other components of ACS stack. This is required if metrics
+	// are collected by Observability Operator.
```
dhaus67, line 114: Since we also will have this in upstream operator, does it make sense to mention the `Observability operator` here? 
From the customer's perspective, this will expose the prometheus port, using the observability operator is something specific to managed services, I'd say we are better off removing this and mentioning that it will expose a metrics port at 9090? Same applies to all occurrences.

dhaus67 (commented):
Where do we do the templating for the network policy / exposed port? For the sake of understanding it, would be nice if you could give some pointers where that's done.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -109,6 +109,12 @@ type CentralComponentSpec struct {
...
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=5
+	ExposeMonitoring *bool `json:"exposeMonitoring,omitempty"`
```
porridge, line 116: Please [avoid booleans](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types), instead use a "typed string" idiom, see `ScannerComponentPolicy` below for an example.

It might also make sense to consider introducing a `Monitoring` sub-struct and place the knob there, in case we plan to introduce more related knobs in the future (you'd probably know better if there would be anything else).

Another thing is exposing monitoring for `scanner`. AFAICT there is functionality for a metrics endpoint in scanner at the moment, but if we also plan to add it in the future, then perhaps it makes more sense to add this `Monitoring` sub-struct at the `spec` level?

porridge (changes requested):
I agree the description in the CRD should be a bit more specific about what it's actually doing.

porridge:
> Looks good, but maybe I'm missing something. What code actually sets the port? I only see it mentioned in the operator tests.

@stehessel [this line](https://github.com/stackrox/stackrox/pull/2117/files#diff-dc2128503d982cb402ef432ceb81160a1597d012a51b791f3673c11b8345702cR109) in `translation.go` does it.

operator/apis/platform/v1alpha1/central_types.go
mtodor: @dhaus67 @stehessel I have changed the structure a bit, based on the comment from @porridge. Now the structure is `central.monitoring.exposeEndpoint` - with string value `Enabled|Disabled`.

Please check the descriptions in the comments again. Only comments are relevant, everything else is automatically generated.

mtodor:
@porridge @dhaus67 @stehessel It would be great if you could provide any improvements for descriptions.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -104,9 +104,14 @@ type CentralComponentSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="User-facing TLS certificate secret",order=3
 	DefaultTLSSecret *LocalSecretReference `json:"defaultTLSSecret,omitempty"`
 
+	// Configures monitoring endpoint for central. Monitoring endpoint allows
+	// other services, like Prometheus, to collect metrics provided by central.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=4
```
dhaus67, line 109: Just as a suggestion from my side, in which direction we should move with the comment, let's see what other's are suggesting:

```suggestion
	// Configures monitoring for Central. 
	// Central provides metrics in prometheus compatible format.
	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=4
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -160,6 +165,36 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
 	return pointer.BoolPtrDerefOr(c.AdminPasswordGenerationDisabled, false)
 }
 
+// Monitoring defines settings for monitoring endpoint.
+type Monitoring struct {
+	// Exposes monitoring endpoint by adding port 9090 with the name
+	// "monitoring" to the deployment, creating service with the same port,
+	// and adding the required network policy.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
```
dhaus67, line 173: Just as a suggestion from my side, in which direction we should move with the comment, let's see what other's are suggesting:

```suggestion
	// Expose Central's monitoring endpoint. A new service, "monitoring", with port
	// 9090 will be created as well as a network policy allowing inbound connections
	// to the port.
	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -160,6 +165,36 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
+	ExposeEndpoint *ExposeEndpoint `json:"exposeEndpoint,omitempty"`
```
dhaus67, line 174: Looking at @porridge's comment regarding the extensibility of i.e. exposing monitoring for central / scanner  / sensor in the future, would it make sense to do `spec.monitoring.exposeCentralMonitoring` instead, allowing us to later on add other options such as `spec.monitoring.expose(Sensor|Scanner)Monitoring` ? Not sure how we typically deal with this within our spec, whether `spec.monitoring.central.exposeMonitoring` is preferred.

stehessel:
> @stehessel [this line](https://github.com/stackrox/stackrox/pull/2117/files#diff-dc2128503d982cb402ef432ceb81160a1597d012a51b791f3673c11b8345702cR109) in translation.go does it.

I still don't get it, this line does not contain the port number. How does the operator which prometheus port to expose?

porridge:
> > @stehessel [this line](https://github.com/stackrox/stackrox/pull/2117/files#diff-dc2128503d982cb402ef432ceb81160a1597d012a51b791f3673c11b8345702cR109) in translation.go does it.
> 
> I still don't get it, this line does not contain the port number. How does the operator know which prometheus port to expose?

It's hardcoded: https://github.com/stackrox/stackrox/blob/f753db3cfc622d397c4a1768ffd602f3b25c0750/image/templates/helm/stackrox-central/templates/_central_endpoints.tpl#L47-L50

operator/apis/platform/v1alpha1/central_types.go
porridge: I actually like the version which mentions "endpoint" more. Seeing just "configures monitoring", some might think that this provides a full solution, including a graphana dashboard ;-)

operator/apis/platform/v1alpha1/central_types.go
mtodor: I've thought about that. The main reason why I went with this nesting is to keep relevant information in the related scope. If we would go with `spec.monitoring.central.*`, we would have a configuration like this:
```
central: {},
monitoring:
  central: {}
```
with this, we have two sections that describe behaviour of one component. It's not localized.

In the future, if we decide to extend the Monitoring struct with options like port number/name, we can easily do that.

And if we also expose scanner monitoring at some point, we can simply use the same Monitoring struct and benefit from available options. (in theory)

operator/apis/platform/v1alpha1/central_types.go
mtodor: I have tweaked this a bit. I like the statement `Prometheus compatible format`, so I have integrated that part.

stehessel (approved):

dhaus67 (approved):
LGTM, sorry for taking a bit longer!

PR #2129, opened 2022-06-20
Create auth provider + groups via ConfigMap+Secret

ivan-degtiarenko:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/main.go
+++ b/central/main.go
@@ -282,7 +282,7 @@ func servicesToRegister(registry authproviders.Registry, authzTraceSink observe.
 		alertService.Singleton(),
 		apiTokenService.Singleton(),
 		authService.New(),
-		authProviderService.New(registry, groupDataStore.Singleton()),
+		authProviderService.NewWithDefaultProvider(registry, groupDataStore.Singleton()),
 		backupRestoreService.Singleton(),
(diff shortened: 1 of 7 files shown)


central/rhsso/load_secret.go
```diff
@@ -0,0 +1,18 @@
+package rhsso
+
+import (
+	"io"
+	"os"
+)
+
+func LoadRhSsoSecret() string {
```
rukletsov, line 8: I don't think we should have such functions in the code. Instead, we need to reference a secret in the auth provider config and load the secret dynamically.

central/main.go
```diff
@@ -282,7 +282,7 @@ func servicesToRegister(registry authproviders.Registry, authzTraceSink observe.
 		alertService.Singleton(),
 		apiTokenService.Singleton(),
 		authService.New(),
-		authProviderService.New(registry, groupDataStore.Singleton()),
+		authProviderService.NewWithDefaultProvider(registry, groupDataStore.Singleton()),
```
rukletsov, line 285: I imagine it'd be cleaner to call into the service (or even datastore) to set up initial auth provider and groups config.

PR #2133, opened 2022-06-21
ROX-11491: Upgrade base images to ubi9

0x656b694d:
The latest tag for `ubi9` is 9.0.0.

### DIFF
--- a/pkg/images/defaults/flavor.go
+++ b/pkg/images/defaults/flavor.go
@@ -167,23 +167,23 @@ func RHACSReleaseImageFlavor() ImageFlavor {
 	v := version.GetAllVersionsUnified()
 	return ImageFlavor{
 		MainRegistry:  "registry.redhat.io/advanced-cluster-security",
-		MainImageName: "rhacs-main-rhel8",
+		MainImageName: "rhacs-main-rhel9",
 		MainImageTag:  v.MainVersion,
-		/* TODO(ROX-9858): Create repo rhacs-central-db-rhel8 when starting building rhacs */
+		/* TODO(ROX-9858): Create repo rhacs-central-db-rhel9 when starting building rhacs */
 		CentralDBImageTag:  v.MainVersion,
-		CentralDBImageName: "rhacs-central-db-rhel8",
+		CentralDBImageName: "rhacs-central-db-rhel9",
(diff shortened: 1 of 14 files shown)


0x656b694d:
adduser warning: postgres's uid 70 outside of the UID_MIN 1000 and UID_MAX 60000 range.

pkg/images/defaults/flavor.go
```diff
@@ -167,23 +167,23 @@ func RHACSReleaseImageFlavor() ImageFlavor {
 	v := version.GetAllVersionsUnified()
 	return ImageFlavor{
 		MainRegistry:  "registry.redhat.io/advanced-cluster-security",
-		MainImageName: "rhacs-main-rhel8",
+		MainImageName: "rhacs-main-rhel9",
```
msugakov, line 170: Appreciate your willingness to get this going! I think migrating to RHEL9/UBI9 is going to be more than just one PR. Specifically, we need to revisit several things downstream so that images really appear in `registry.redhat.io/advanced-cluster-security/rhacs-*-rhel9` repositories. Right now, there are no `-rhel9` images in the existence, therefore this change will break the moment someone tries to deploy `rhacs` flavor (i.e. release 3.71.0). I'm blocking the PR due to this.

We must migrate to 9. Ross brought it up earlier too. Back then I said that this needs to be planned. The migration need to be synchronized between the upstream and the downstream.
https://srox.slack.com/archives/CELUQKESC/p1652905303812889?thread_ts=1652898099.639479&cid=CELUQKESC
Maybe the safest approach is to migrate container after container.
I still think the same way. The portion downstream risks to be bigger than the upstream.

I don't want to commit myself to "help" with the downstream because I have other priorities and commitments at the moment. Someone else needs to lead that part, and I can consult.

I think, the best way is to create a JIRA epic, raise its importance and get at least 1-2 engineers time commitment to work on this.

msugakov (changes requested):

pkg/images/defaults/flavor.go
0x656b694d: Right, thanks. I just wanted to see if that'll be an easy sed, but it's not apparently.

PR #2138, opened 2022-06-21
Add Aging Images widget to main dashboard

dvail:
## Description

Adds an "Aging Images" widget to the dashboard. This widget displays the number of images that have an age range corresponding to one of four age range "buckets". Note that these buckets do not overlap, which is in contrast with the data that is returned from the server API, hence the amount of processing that needs to happen client side before display.

Users are able to customize the date ranges (maximum value of one year) and disable individual ranges.

## Additional notes

- [ ] TODO - Announce on team channel when this is merged so team can run `yarn` in repo root, otherwise unit tests will fail.
- Viswa is working [an extension of the query language](https://github.com/stackrox/stackrox/pull/2159) that will make the outbounds links to the Vuln Mgmt image page work correct as well as simplify the querying and client side processing that is done to group the counts into buckets. I'll put out a follow up PR when both of these are merged that uses the new API.
- I've used Jest/React Testing Library in this PR for a couple of reasons. IMO it is quicker to iterate and run tests and less prone to flakes when the browser/network combo is removed. The trade off being that we need to mock API requests which adds a little bit of verbosity. In think in the case of these dashboard widgets it is worthwhile since we aren't interacting with the server other than to pull data, and would otherwise be at the mercy of whatever data the CI system happens to have at the time tests run. 

## Checklist

## Testing Performed

On page load, the aging images data should be grouped into buckets. The total count displayed in the title should match the sum of all bars in the graph.
image

Open the option menu and disabled the first option. The grey bar should disappear and the widget title should update to reflect the new totals.
image

Disabled the third option in the menu. The orange bar should disappear and the image counts for the yellow and orange bars should be summed in the yellow bar.
image

Update the value of one of the inputs. The total image counts should update to reflect the new date ranges.
image

Update the value of one of the inputs so that the values in the options menu are no longer in ascending order. The graph should display an error and there should be an inline error state for the individual inputs.
image

Update the values of the text inputs. The x-axis links should update to reflect the new date ranges.
image

Click one of the links in the x-axis labels. You should be take to the Vuln Mgmt image page with a pre-applied image age filter. The resulting table should also be sorted with oldest images first.
image

Repeat this process by clicking on one of the bars in the graph.

Test that the "View All" link button takes you to the Vuln Mgmt image page with no filters applied, with images sorted by oldest age first.
image

Test that links from the widget also apply any selected cluster and namespace filters.
image

### DIFF
--- a/ui/apps/platform/package.json
+++ b/ui/apps/platform/package.json
@@ -124,6 +124,7 @@
         "@testing-library/jest-dom": "^5.15.0",
         "@testing-library/react": "^12.1.2",
         "@testing-library/react-hooks": "^7.0.2",
+        "@testing-library/user-event": "^14.2.1",
         "@types/jest": "^27.4.0",
         "@types/lodash": "^4.14.176",
         "@types/node": "^17.0.16",
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
@@ -1,0 +1,252 @@
+import React, { useReducer } from 'react';
+import {
+    Flex,
+    FlexItem,
+    Title,
+    Button,
+    Dropdown,
+    DropdownToggle,
+    Form,
+    FormGroup,
+    Checkbox,
+    TextInput,
+    ValidatedOptions,
+} from '@patternfly/react-core';
+import { useQuery, gql } from '@apollo/client';
+import cloneDeep from 'lodash/cloneDeep';
+import pluralize from 'pluralize';
+
+import LinkShim from 'Components/PatternFly/LinkShim';
+import useURLSearch from 'hooks/useURLSearch';
+import { getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import { SearchFilter } from 'types/search';
+import { vulnManagementImagesPath } from 'routePaths';
+import { getQueryString } from 'utils/queryStringUtils';
+import WidgetCard from './WidgetCard';
+import AgingImagesChart, {
+    TimeRangeCounts,
+    TimeRangeTupleIndex,
+    TimeRangeTuple,
+    timeRangeTupleIndices,
+} from './AgingImagesChart';
+
+export const imageCountQuery = gql`
+    query agingImagesQuery($query0: String, $query1: String, $query2: String, $query3: String) {
+        timeRange0: imageCount(query: $query0)
+        timeRange1: imageCount(query: $query1)
+        timeRange2: imageCount(query: $query2)
+        timeRange3: imageCount(query: $query3)
+    }
+`;
+
+function queryStringFor(timeRangeValue: number, searchFilter: SearchFilter) {
+    return getRequestQueryStringForSearchFilter({
+        ...searchFilter,
+        'Image Created Time': `>${timeRangeValue}d`,
+    });
+}
+
+type QueryVariables = Record<`query${TimeRangeTupleIndex}`, string>;
+
+function getQueryVariables(timeRanges: TimeRangeTuple, searchFilter: SearchFilter): QueryVariables {
+    return {
+        query0: queryStringFor(timeRanges[0].value, searchFilter),
+        query1: queryStringFor(timeRanges[1].value, searchFilter),
+        query2: queryStringFor(timeRanges[2].value, searchFilter),
+        query3: queryStringFor(timeRanges[3].value, searchFilter),
+    };
+}
+
+// Gets the header string title for the widget based on applied filters and resulting counts
+function getWidgetTitle(
+    searchFilter: SearchFilter,
+    selectedTimeRanges: TimeRangeTuple,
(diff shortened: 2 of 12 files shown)


dvail:
* master
  * **PR #2138** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2138" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2138?utm_source=stack-comment).

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -0,0 +1,258 @@
...
+                                                aria-label="Toggle image time range"
+                                                id={`${fieldIdPrefix}-time-range-${index}`}
+                                                name={`${fieldIdPrefix}-time-range-${index}`}
+                                                className="pf-u-mb-sm pf-u-display-flex pf-u-align-items-center"
+                                                isChecked={timeRanges[index].enabled}
+                                                onChange={() => dispatch({ type: 'toggle', index })}
+                                                label={
+                                                    <TextInput
+                                                        aria-label="Image age in days"
+                                                        style={{ minWidth: '100px' }}
```
dvail, line 209: The checkboxes and number inputs alone are small enough to cause the menu label to wrap. This ensures we have a large enough minimum size to not break the layout of the dropdown.

ui/apps/platform/package.json
```diff
@@ -124,6 +124,7 @@
         "@testing-library/jest-dom": "^5.15.0",
         "@testing-library/react": "^12.1.2",
         "@testing-library/react-hooks": "^7.0.2",
+        "@testing-library/user-event": "^14.2.1",
```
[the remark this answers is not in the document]
pedrottimark: Yes, please update any shared dependencies in package.json files:
* apps/platform
* packages/tailwind-config
* packages/ui-components

Keep it simple for yarn because we have 2-level tree monorepo instead of 1-level flat monorepo.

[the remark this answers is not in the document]
pedrottimark: Thinking about your question more carefully, **ordinary** dependencies need to be consistent in all package.json files, not just in yarn.lock file, because infra project sees package.json of the 2 reusable packages only, therefore cannot inherit a newer version from apps/platform package.json file, as has happened here for the dev dependency. For example, React 18 upgrade. Having said that, updates to infra have slowed down in 2022.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -0,0 +1,252 @@
...
+    if (!timeRangeCounts) {
+        return 'Aging images';
+    }
+
+    const totalImages =
+        Object.values(timeRangeCounts).find((range, index) => {
+            return selectedTimeRanges[index].enabled;
+        }) ?? 0;
+
+    const isActiveImages = Boolean(searchFilter.Cluster) || Boolean(searchFilter['Namespace ID']);
```
pedrottimark, line 76: What do you think about pure helper function as future refactoring after initial release?

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -0,0 +1,252 @@
...
+    timeRangeCounts?: TimeRangeCounts
+): string {
+    if (!timeRangeCounts) {
+        return 'Aging images';
+    }
+
+    const totalImages =
+        Object.values(timeRangeCounts).find((range, index) => {
+            return selectedTimeRanges[index].enabled;
+        }) ?? 0;
```
pedrottimark, line 74: A short comment here might help. It looks like the count value of the first enabled range index.

From pictures, it looks like it works. Therefore, are the following assumptions valid:
* counts are cumulative? Somewhere else you explain that frontend computation is needed for bars.
* counts are in descending order? Therefore, count of first enabled range is total of enabled? Can you double check if it still works correctly if you disable 365 (scenario not included in the pictures).

A short comment here might help for now, and maybe future refactoring to see all pure computational helpers for aging images together with related types?

pedrottimark (approved):
Thank you again for careful coding and writing unit tests as you go.

Here is an observation from a neutral perspective:
* The amount of frontend computation to transform the response is yellow caution for GraphQL query (because not considered public for customers) and would be orange warning for REST API (because customer would need to reverse engineer the code to do similarly in their own automation).
* As a thought experiment, a table which has columns for both accumulating subtotals and differences (which correspond to bars) would be more transparent, and also answer plausible questions like “how many images older than 90 days” directly with 31 instead of indirectly with 1 + 13 + 18 (see first picture). A question that the query is designed to answer.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
dvail: I think this is already handled in a downstream PR, as the pattern had emerged quite a few times across widgets.

PR #2143, opened 2022-06-22
ROX-10273: Protos and service, datastore, searcher, indexer and store layers for Policy categories

clickboo:
## Description

1. New storage and API layer protos for Policy categories and the related service.
2. Datastore and store layers.
3. Indexer and searcher.
4. Service registered under a feature flag.
5. Unit tests where necessary.
6. e2e tests TBD - need to unblock Linda's UI work.
7.  Added default category jsons
8.  Migration TBD in follow on PRs (in 72).
9. Fixed a bug in populating default policies in rocksdb and in postgres on startup.

## Checklist

## Testing Performed
manual TBD


clickboo:
Ping @theencee @viswajithiii

central/policycategory/service/service_impl.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if query.GetQuery() == "" {
+		categories, err := s.policyCategoriesDatastore.GetAllPolicyCategories(ctx)
+		if err != nil {
+			return nil, err
+		}
+		resp.Categories = s.convertCategoriesToV1Categories(categories)
+
+	} else {
+		parsedQuery, err := search.ParseQuery(query.GetQuery())
```
theencee, line 65: Does the searcher not work if the query is empty? If it does, this could be simplified.

central/policycategory/service/convertor.go
```diff
@@ -0,0 +1,24 @@
+package service
+
+import (
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// ToStorageProto converts the v1 representation to its corresponding storage representation
+func ToStorageProto(category *v1.PolicyCategory) *storage.PolicyCategory {
```
theencee, line 9: yay! Happy to see presentation and storage split

central/policy/datastore/singleton.go
```diff
@@ -31,11 +31,12 @@ func initialize() {
 	if features.PostgresDatastore.Enabled() {
 		storage = policyPostgres.New(globaldb.GetPostgres())
 		indexer = policyPostgres.NewIndexer(globaldb.GetPostgres())
-		addDefaults(storage)
 	} else {
 		storage = boltdb.New(globaldb.GetGlobalDB())
 		indexer = index.New(globalindex.GetGlobalTmpIndex())
 	}
+
+	addDefaults(storage)
```
theencee, line 39: Why is this moved out of the postgres if block?

clickboo: Related to point 9. in the description: "Fixed a bug in populating default policies in rocksdb and in postgres on startup.
". Basically fixed a bug since default policies must be added regardless of postgres or rockdb

central/policycategory/service/service_impl.go
clickboo: Ack, yes! Done.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, errorsPkg.Wrap(err, "getting all policies")
+	}
+	categoryNameIDMap := make(map[string]string, len(categories))
+	for _, category := range categories {
+		categoryNameIDMap[category.GetName()] = category.GetId()
+	}
+
+	category.Name = strings.Title(category.GetName())
+	if _, isUnique := ds.isCategoryNameUnique(categoryNameIDMap, category.GetName()); !isUnique {
```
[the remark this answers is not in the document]
clickboo: Already done in the proto with sql tags for uniqueness, but rocksdb needs uniqueness checking done in datastore layer.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,222 @@
...
+	return ds.searcher.SearchRawCategories(ctx, q)
+}
+
+func (ds *datastoreImpl) GetAllPolicyCategories(ctx context.Context) ([]*storage.PolicyCategory, error) {
+	if ok, err := policyCategorySAC.ReadAllowed(ctx); err != nil || !ok {
+		return nil, err
+	}
+
+	var categories []*storage.PolicyCategory
+	err := ds.storage.Walk(policyCategoryCtx, func(category *storage.PolicyCategory) error {
```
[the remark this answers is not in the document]
clickboo: Hmm, given the scale of this datastore I think we should be ok with a Walk. @connorgorman? Also - you mean not have GetAll at all @theencee ? The UI list views will need this and so will list APIs. Not sure I understand.

pkg/defaults/categories/files/anomalous_activity.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
clickboo: Kept it consistent with policies, tbh. Only reason why.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,222 @@
...
+	categories, err := ds.GetAllPolicyCategories(ctx)
+	if err != nil {
+		return errorsPkg.Wrap(err, "getting all policies")
+	}
+	categoryNameIDMap := make(map[string]string, len(categories))
+	for _, category := range categories {
+		categoryNameIDMap[category.GetName()] = category.GetId()
+	}
+
+	existingID, isUnique := ds.isCategoryNameUnique(categoryNameIDMap, newName)
```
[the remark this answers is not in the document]
clickboo: Same as above (rocksdb). I've added the `uniq-key-func` to the `gen.go` but I don't believe that helps with uniqueness constraint checking on the way into the db ... the datastore needs to do that.

qa-tests-backend/src/test/groovy/SACTest.groovy
```diff
@@ -424,7 +424,8 @@ class SACTest extends BaseSpecification {
         "Verify something was returned for every search category"
         for (SSOC.SearchResponse.Count numResults : result.countsList) {
             // Policies are globally scoped so our cluster-scoped query won't return any
-            if (numResults.category == SSOC.SearchCategory.POLICIES) {
+            if (numResults.category == SSOC.SearchCategory.POLICIES ||
```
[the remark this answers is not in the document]
clickboo: I think it pulls from the proto, since the test failed without the if check.

central/policycategory/datastore/datastore_impl.go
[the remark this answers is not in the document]
theencee: Ah I forgot about that use case. Makes sense.

theencee (approved):
lgtm

central/policy/datastore/singleton.go
connorgorman: This is in the bolt store currently so now its duplicated

clickboo: sorry @theencee I meant bolt. I need to add defaults regardless of the store.

central/policycategory/search/searcher_impl.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/search/blevesearch"
+	"github.com/stackrox/rox/pkg/search/paginated"
+	"github.com/stackrox/rox/pkg/search/sortfields"
+)
+
+var (
+	defaultSortOption = &v1.QuerySortOption{
```
connorgorman, line 20: Default sort options should be moved to the service layer and not in the searcher layer

central/policycategory/search/searcher_impl.go
```diff
@@ -0,0 +1,107 @@
...
+func (s *searcherImpl) searchCategories(ctx context.Context, q *v1.Query) ([]*storage.PolicyCategory, []search.Result, error) {
+	results, err := s.Search(ctx, q)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var categories []*storage.PolicyCategory
+	var newResults []search.Result
+	for _, result := range results {
+		category, exists, err := s.storage.Get(ctx, result.ID)
```
connorgorman, line 76: This seems like it would be more efficient to get many then filter out by the missing indices

central/policycategory/service/service.go
```diff
@@ -0,0 +1,25 @@
...
+
+	AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error)
+
+	v1.PolicyCategoryServiceServer
+}
+
+// New returns a new Service instance using the given DataStore.
+func New(policyCategoriesDatastore datastore.DataStore) Service {
+	return &serviceImpl{
+		policyCategoriesDatastore: datastore.Singleton(),
```
connorgorman, line 23: This should use the passed in value and not the Singleton()

central/policycategory/service/service_impl.go
```diff
@@ -0,0 +1,123 @@
...
+func (s *serviceImpl) PostPolicyCategory(ctx context.Context, request *v1.PostPolicyCategoryRequest) (*v1.PolicyCategory, error) {
+	category, err := s.policyCategoriesDatastore.AddPolicyCategory(ctx, ToStorageProto(request.GetPolicyCategory()))
+	if err != nil {
+		return nil, err
+	}
+	return ToV1Proto(category), nil
+}
+
+func (s *serviceImpl) RenamePolicyCategory(ctx context.Context, request *v1.NewRenamePolicyCategoryRequest) (*v1.Empty, error) {
+	return &v1.Empty{}, s.policyCategoriesDatastore.RenamePolicyCategory(ctx, request.GetId(), request.GetNewCategoryName())
```
connorgorman, line 79: Should we validate that the new name isn't empty?

central/policycategory/service/service_impl.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, err
+	}
+	resp.Categories = s.convertCategoriesToV1Categories(categories)
+	
+	return resp, nil
+}
+
+func (s *serviceImpl) PostPolicyCategory(ctx context.Context, request *v1.PostPolicyCategoryRequest) (*v1.PolicyCategory, error) {
+	category, err := s.policyCategoriesDatastore.AddPolicyCategory(ctx, ToStorageProto(request.GetPolicyCategory()))
```
connorgorman, line 71: validate non empty name

proto/api/v1/policy_category_service.proto
```diff
@@ -0,0 +1,78 @@
...
+import "api/v1/empty.proto";
+import "api/v1/search_service.proto";
+
+package v1;
+
+
+message PolicyCategory {
+    string id       = 1 ;
+    string name     = 2 ;
+    bool isDefault  = 3 ;
```
connorgorman, line 17: Nit: align?

proto/api/v1/policy_category_service.proto
```diff
@@ -0,0 +1,78 @@
...
+    string id = 1;
+}
+
+// PolicyCategoryService APIs can be used to manage policy categories.
+service PolicyCategoryService {
+
+    // GetPolicyCategory returns the requested policy category by ID.
+    rpc GetPolicyCategory (ResourceByID) returns (v1.PolicyCategory) {
+        option (google.api.http) = {
+            get: "/v1/policyCategories/{id}"
```
connorgorman, line 44: Can we remove the capitalization? `policycategories` seems cleaner and more standard

connorgorman (changes requested):
See comments

central/policy/datastore/singleton.go
connorgorman: That's correct. The comment on the addDefaults function remarks that the migration will need to take care of the fact we removed the removed_default_policies bucket so the migration will need to account for it

central/policycategory/datastore/datastore_impl.go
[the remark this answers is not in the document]
connorgorman: the rocksdb code does check that the value being upsert is unique

central/policycategory/search/searcher_impl.go
clickboo: A quick search in the code base shows default options in searcher in almost all entities. Also makes sense since it is used in the searcher. Prefer leaving this as is, unless there is a strong reason to not do so.

connorgorman: Correct, this is the beginning of a change that we should start making. Sorting unnecessarily makes the queries more expensive and many internal use cases don't care that the data is sorted. We should rely on the caller to request that the data be sorted and so the use case of a default sort should be applied in the service layer as it is handling the response to the client

pkg/search/sortfields/fields_map.go
```diff
@@ -19,6 +19,14 @@ var (
 				},
 			}
 		},
+		search.PolicyCategoryName: func(option *v1.QuerySortOption) []*v1.QuerySortOption {
+			return []*v1.QuerySortOption{
+				{
+					Field:    search.SORTPolicyName.String(),
```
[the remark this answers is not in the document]
clickboo: Drat yes, fixed.

central/policycategory/search/searcher_impl.go
clickboo: Ok then let's do a separate pass on this. In this particular case, I wouldn't worry since scale isn't an issue.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,183 @@
...
+	if category.Id == "" {
+		category.Id = uuid.NewV4().String()
+	}
+
+	ds.categoryMutex.Lock()
+	defer ds.categoryMutex.Unlock()
+
+	category.Name = strings.Title(category.GetName())
+	err := ds.storage.Upsert(ctx, category)
+
```
connorgorman, line 129: nit: remove line

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,183 @@
...
+	ds.categoryMutex.Lock()
+	defer ds.categoryMutex.Unlock()
+
+	category, exists, err := ds.storage.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return fmt.Errorf(" policy category '%s' not found", id)
```
connorgorman, line 152: '%s' -> %q

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	category.Name = strings.Title(newName)
+	err = ds.storage.Upsert(ctx, category)
+	if err != nil {
+		return errorsPkg.Wrap(err, fmt.Sprintf("failed to rename category '%s' to '%s'", id, newName))
```
connorgorman, line 158: same '%s' to %q

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,183 @@
...
+		return sac.ErrResourceAccessDenied
+	}
+
+	ds.categoryMutex.Lock()
+	defer ds.categoryMutex.Unlock()
+
+	return ds.deleteCategoryNoLock(ctx, id)
+}
+
+func (ds *datastoreImpl) deleteCategoryNoLock(ctx context.Context, id string) error {
```
connorgorman, line 178: No need for a new function here

central/policycategory/datastore/datastore_impl.go
```diff
@@ -0,0 +1,183 @@
...
+type datastoreImpl struct {
+	storage  store.Store
+	indexer  index.Indexer
+	searcher search.Searcher
+
+	categoryMutex sync.Mutex
+}
+
+func (ds *datastoreImpl) buildIndex() error {
+	var categories []*storage.PolicyCategory
```
connorgorman, line 42: Add if postgres enabled return nil

central/policycategory/search/searcher_impl.go
connorgorman: fair enough

central/policycategory/service/service_impl.go
```diff
@@ -0,0 +1,136 @@
...
+func (s *serviceImpl) getPolicyCategory(ctx context.Context, id string) (*v1.PolicyCategory, error) {
+	if id == "" {
+		return nil, errors.Wrap(errox.InvalidArgs, "Policy category ID must be provided")
+	}
+	category, exists, err := s.policyCategoriesDatastore.GetPolicyCategory(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, errors.Wrapf(errox.NotFound, "policy with ID '%s' does not exist", id)
```
connorgorman, line 124: %q

central/policycategory/service/service_impl.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return ToV1Proto(category), nil
+
+}
+
+func (s *serviceImpl) convertCategoriesToV1Categories(categories []*storage.PolicyCategory) []*v1.PolicyCategory {
+	v1Categories := make([]*v1.PolicyCategory, len(categories))
```
connorgorman, line 131: this should be 0, len(categories). Otherwise, this has len(categories) nil values in the front

connorgorman (approved):
Looks good! Added some minor comments

pkg/search/sortfields/fields_map.go
```diff
@@ -19,6 +19,14 @@ var (
 				},
 			}
 		},
+		search.PolicyCategoryName: func(option *v1.QuerySortOption) []*v1.QuerySortOption {
```
md2119, line 22: This is a noop `PolicyCategoryName` to `PolicyCategoryName`

tools/generate-helpers/pg-table-bindings/list.go
```diff
@@ -43,6 +43,8 @@ func init() {
 		&storage.NodeComponentEdge{}:                            resources.Node,
 		&storage.PermissionSet{}:                                resources.Role,
 		&storage.Pod{}:                                          resources.Deployment,
+		&storage.Policy{}:                                       resources.Policy,
```
md2119, line 46: Policy is not indirectly scoped. Policy SAC -> policy SAC. Policy entry should not be here.

clickboo: https://github.com/stackrox/stackrox/pull/2289

pkg/search/sortfields/fields_map.go
clickboo: https://github.com/stackrox/stackrox/pull/2289

PR #2148, opened 2022-06-22
Adds -rcX support for collector probe versions.

Stringy:
## Description

To support new MODULE_VERSION strategy in collector, this adds support for release candidate (`rcX`) versions in sensor.

## Checklist

## Testing Performed

Added new unit test cases for release candidate versions, and tested locally end-to-end with a COLLECTOR_VERSION that contains a release candidate version. Note that until an rc version is merged to collector master, that version is not available for download (i.e. this process does not work with in-progress PRs)

### DIFF
--- a/pkg/probeupload/names_test.go
+++ b/pkg/probeupload/names_test.go
@@ -19,6 +19,8 @@ func TestIsValidModuleVersion_Valid(t *testing.T) {
 		"0.0.1",
 		"1.2.3",
 		"10.5.155",
+		"1.0.0-rc1",
+		"2.3.4-rc123",
 	}
 
 	for _, v := range values {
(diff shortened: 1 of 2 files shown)


pkg/probeupload/names_test.go
```diff
@@ -35,6 +37,8 @@ func TestIsValidModuleVersion_Invalid(t *testing.T) {
 		".2.3",
 		"2.3.",
 		"10.5.155.123",
+		"1.0.0-rc",
+		"2.0.0-invalid",
```
Molter73, line 41: Pretty silly nitpick, but could we add a few cases like `1.0.0-rc1-rc4` and `2.0.0-rc1invalid`?

Molter73 (approved):
Looking good! Thanks for addressing this @Stringy!

PR #2151, opened 2022-06-22
ROX-11442: Increase default roxctl timeout

stehessel:
The timeout has been hit in this [CI flake](https://issues.redhat.com/browse/ROX-11442).
The root cause of the timeout is hard to find because the issue is not
reproduced in subsequent CI runs. To decrease the probability of
further issues, the default timeout is increased to 30 seconds. I
believe this is still a reasonable value for roxctl commands.

## Checklist

## Testing Performed

* CI


stehessel:
\test all

PR #2157, opened 2022-06-22
Use cursors in Postgres Walk instead of select all

connorgorman:
## Description

The select all is pretty suboptimal in terms of memory pressure as seen on the scale run I had
spike in pruning

## Checklist

## Testing Performed

Will evaluate this on a scale run to see the memory usage and also evaluate start up time

### DIFF
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -14,6 +14,7 @@ import (
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/pointers"
 	"github.com/stackrox/rox/pkg/postgres/walker"
+	"github.com/stackrox/rox/pkg/random"
 	searchPkg "github.com/stackrox/rox/pkg/search"
 	"github.com/stackrox/rox/pkg/search/postgres/mapping"
 	pgsearch "github.com/stackrox/rox/pkg/search/postgres/query"
@@ -24,6 +25,8 @@ import (
 
 var (
 	log = logging.LoggerForModule()
+
+	emptyQueryErr = errox.InvalidArgs.New("empty query")
 )
 
 // QueryType describe what type of query to execute
@@ -593,7 +596,7 @@ func RunGetQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Quer
 		return nil, err
 	}
 	if query == nil {
-		return nil, errox.InvalidArgs.New("empty query")
+		return nil, emptyQueryErr
 	}
 
 	queryStr := query.AsSQL()
@@ -611,7 +614,7 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
 		return nil, err
 	}
 	if query == nil {
-		return nil, errox.InvalidArgs.New("empty query")
+		return nil, emptyQueryErr
 	}
 
 	queryStr := query.AsSQL()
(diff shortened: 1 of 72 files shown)


viswajithiii (approved):

theencee (approved):

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+
+	cursor := random.GenerateString(16, random.CaseInsensitiveAlpha)
+	_, err = tx.Exec(ctx, fmt.Sprintf("DECLARE %s CURSOR FOR %s", cursor, queryStr), query.Data...)
+	if err != nil {
+		closer()
+		return nil, nil, err
+	}
+
+	return func(n int) ([][]byte, error) {
+		rows, err := tx.Query(ctx, fmt.Sprintf("FETCH %d FROM %s", n, cursor))
```
janisz, line 640: nit: should use sprintf here or pg template to handle escaping?

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+		return nil, nil, errox.InvalidArgs.New("empty query")
+	}
+
+	queryStr := query.AsSQL()
+
+	conn, err := db.Acquire(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	closer = func() {
```
janisz, line 616: this could be inlined

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
 }
 
+// RunCursorQueryForSchema creates a cursor against the database
+func RunCursorQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (fetcher func(n int) ([][]byte, error), closer func(), err error) {
+	query, err := standardizeQueryAndPopulatePath(q, schema, GET)
+	if err != nil {
+		return nil, nil, err
+	}
+	if query == nil {
+		return nil, nil, errox.InvalidArgs.New("empty query")
```
janisz, line 607: How about moving err to package. It's in 2 places

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+		conn.Release()
+	}
+
+	tx, err := conn.Begin(ctx)
+	if err != nil {
+		closer()
+		return nil, nil, err
+	}
+	closer = func() {
+		if err := tx.Rollback(ctx); err != nil {
```
janisz, line 626: Why not commit? Why do we need rollback?

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
 	return results, nil
 }
 
+// RunCursorQueryForSchema creates a cursor against the database
+func RunCursorQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (fetcher func(n int) ([][]byte, error), closer func(), err error) {
+	query, err := standardizeQueryAndPopulatePath(q, schema, GET)
+	if err != nil {
+		return nil, nil, err
```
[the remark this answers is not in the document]
connorgorman: yeah can wrap them

pkg/search/postgres/common.go
```diff
@@ -13,18 +13,26 @@ import (
...
 	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/rox/pkg/stringutils"
 	"github.com/stackrox/rox/pkg/ternary"
+	"go.uber.org/atomic"
 )
 
 var (
 	log = logging.LoggerForModule()
+
+	cursorNum atomic.Int32
```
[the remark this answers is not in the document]
connorgorman: ah will remove, thx

pkg/search/postgres/common.go
connorgorman: yeah changed to commit

pkg/search/postgres/common.go
connorgorman: actually only can template in the where

rhybrillou (approved):

pkg/search/postgres/common.go
janisz: Now the log statement is wrong

pkg/search/postgres/common.go
```diff
@@ -589,6 +597,64 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+	}
+
+	tx, err := conn.Begin(ctx)
+	if err != nil {
+		closer()
+		return nil, nil, err
+	}
+	closer = func() {
+		if err := tx.Rollback(ctx); err != nil {
+			log.Errorf("error rolling back cursor transaction: %v", err)
```
janisz, line 614:
```suggestion
			log.Errorf("error committing cursor transaction: %v", err)
```

pkg/search/postgres/common.go
```diff
@@ -589,6 +592,55 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+		closer()
+		return nil, nil, errors.Wrap(err, "creating transaction")
+	}
+	closer = func() {
+		if err := tx.Commit(ctx); err != nil {
+			log.Errorf("error rolling back cursor transaction: %v", err)
+		}
+	}
+
+	cursor := random.GenerateString(16, random.CaseInsensitiveAlpha)
```
erthalion, line 618: Since `GenerateString` uses crypto/rand, it may fail. Do I see it correctly that in such situation the opened transaction will not be closed?

connorgorman: It panics as that should never happen so the whole application will fail

pkg/search/postgres/common.go
```diff
@@ -589,6 +592,55 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+	_, err = tx.Exec(ctx, fmt.Sprintf("DECLARE %s CURSOR FOR %s", cursor, queryStr), query.Data...)
```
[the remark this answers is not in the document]
connorgorman: Yeah, this is definitely going to trade off memory and realistically, we avoid walking all of the objects in a table, but we do it in some places and this makes us less likely to crash

pkg/search/postgres/common.go
erthalion: I guess it could happen if there is not enough entropy for generating the random value. The whole application will fail, right, but an orphaned connection will be left and connection leaking normally is a problem.

pkg/search/postgres/common.go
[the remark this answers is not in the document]
connorgorman: Let me run an empirical test

pkg/search/postgres/common.go
erthalion: Thanks!

connorgorman:
@erthalion We'll trade this off for now. Iterating over 264117 took 17s. We can optimize this and hone this in as we perform more performance testing. The main culprit in time is network flow pruning which doesn't use the cursor so that's the bigger fish to fry

Screen Shot 2022-06-23 at 4 17 28 PM

PR #2159, opened 2022-06-22
Support range queries for numeric fields

viswajithiii:
## Description

Add support for numeric queries which are ranges. This is done in a backward compatible way. You can say something like "1-2" for values between 1 and 2 (non-inclusive).

## Checklist

## Testing Performed

Unit tests

### DIFF
--- a/pkg/search/postgres/query/numeric_query.go
+++ b/pkg/search/postgres/query/numeric_query.go
@@ -2,12 +2,12 @@ package pgsearch
 
 import (
 	"fmt"
-	"math"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/readable"
 )
 
 type prefixAndInversion struct {
@@ -27,7 +27,7 @@ var (
 			validPrefixes = append(validPrefixes, pAndI.prefix)
 			validPrefixes = append(validPrefixes, pAndI.inversion)
 		}
-		validPrefixes = append(validPrefixes, "==")
+		validPrefixes = append(validPrefixes, "=", "==")
 		sort.Slice(validPrefixes, func(i, j int) bool {
 			return len(validPrefixes[i]) > len(validPrefixes[j])
 		})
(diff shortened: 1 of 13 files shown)


pkg/search/postgres/query/numeric_query.go
```diff
@@ -110,13 +110,24 @@ func getEquivalentGoFuncForNumericQuery(prefix string, value float64) func(found
...
-	var valueStr string
-	if _, fraction := math.Modf(value); fraction > 0 {
-		valueStr = fmt.Sprintf("%0.2f", value)
-	} else {
-		valueStr = fmt.Sprintf("%0.0f", value)
+func createNumericRangeQuery(root string, lower float64, upper float64) WhereClause {
+	lowerStr := readable.Float(lower, 2)
+	upperStr := readable.Float(upper, 2)
+	return WhereClause{
+		Query:  fmt.Sprintf("(%s > $$) AND (%s < $$)", root, root),
```
connorgorman, line 117: Why are these exclusive? That seems unintuitive

clickboo: My thoughts on this: This is really a preference thing and there is no clear answer. Having said that I am for keeping the range exclusive. Since otherwise I have to specify 31-59 if it is inclusive and I want >30 < 60. Which is rather unintuitive IMO. If someone wants it inclusive, = 30, >30 < 60, = 60 as a union should give them everything that they need!

clickboo:
@charmik-redhat looping you in because I think you might find this query ability useful in the context of the decommissioning clusters feature.

pkg/search/postgres/query/numeric_query.go
connorgorman: So I think either we just make the format capable of understanding `[]` `()` or we just allow multiple values in the same chip so we can do >=30 <=60 if you want inclusivity.

viswajithiii: The problem is multiple values in the same chip mean "OR" right now. So if you do `>=30,<=60`, it becomes ">=30 OR <= 60" which is... everything.

In any case, we don't have that many numeric fields, this is mostly a concern for days. There, the inclusivity versus exclusivity thing is kind of moot anyway because the query goes down to a specific number of seconds.

connorgorman: It would have to be in the same chip and be parsed out

viswajithiii: Hmm, so do you think we should replace the whole syntax from `a-b` to `<a>b` and have that implicitly mean `AND`?

viswajithiii: It's a bit ugly but that would be unambiguous

connorgorman: Follow up question, should we even expose this to users in this form? Like wouldn't this just be for internal use? Where did this use case come from?

viswajithiii: This is an internal use-case (for the dashboards and such). I don't see a reason to go out of our way to make it impossible for users to use, but we also don't have to advertise it or anything

connorgorman: Another follow up, if it's for dashboards, then we should try and figure out how to expose this as part of a query language aggregation?

viswajithiii: I think those are all good for follow-up work, but those will need more careful design and thought. This was meant to just be a quicker fix to unblock this use case within the existing query langauge.

connorgorman: Perf impact seems negligible, can you create some tickets for the follow ups?

viswajithiii: Done: https://issues.redhat.com/browse/ROX-11528

connorgorman (approved):

PR #2163, opened 2022-06-23
ROX-10987: Splitting Components into Node and Image Components

sachaudh:
## Description

* Added two new entity types: `IMAGE_COMPONENT` and `NODE_COMPONENT`
* It's difficult to feature flag the `entityRelationships.ts` because that module doesn't live within the React lifecycle so I'm open to suggestions on how to go about feature flagging it. For now, I commented out the changes so the entity relationships won't be updated right now

## Checklist

## Testing Performed

Screen Shot 2022-06-24 at 4 28 44 PM
Screen Shot 2022-06-24 at 4 29 02 PM
Screen Shot 2022-06-24 at 4 29 20 PM
Screen Shot 2022-06-24 at 4 29 30 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
@@ -27,7 +27,8 @@ const entityMenuTypes = [
     entityTypes.CLUSTER,
     entityTypes.NAMESPACE,
     entityTypes.DEPLOYMENT,
-    entityTypes.COMPONENT,
+    entityTypes.NODE_COMPONENT,
+    entityTypes.IMAGE_COMPONENT,
 ];
 
 const VulnDashboardPage = ({ history }) => {
--- a/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
@@ -24,18 +24,21 @@ import MostCommonVulnerabilities from '../widgets/MostCommonVulnerabilities';
 import DeploymentsWithMostSeverePolicyViolations from '../widgets/DeploymentsWithMostSeverePolicyViolations';
 import ClustersWithMostOrchestratorIstioVulnerabilities from '../widgets/ClustersWithMostOrchestratorIstioVulnerabilities';
 
-const entityMenuTypes = [
-    entityTypes.CLUSTER,
-    entityTypes.NAMESPACE,
-    entityTypes.DEPLOYMENT,
-    entityTypes.NODE_COMPONENT,
-    entityTypes.IMAGE_COMPONENT,
-];
+const baseEntityMenuTypes = [entityTypes.CLUSTER, entityTypes.NAMESPACE, entityTypes.DEPLOYMENT];
+const componentMenuType = [entityTypes.COMPONENT];
+const splitComponentMenuTypes = [entityTypes.NODE_COMPONENT, entityTypes.IMAGE_COMPONENT];
 
 const VulnDashboardPage = ({ history }) => {
     const workflowState = useContext(workflowStateContext);
     const searchState = workflowState.getCurrentSearchState();
     const { isFeatureFlagEnabled } = useFeatureFlags();
+
+    let entityMenuTypes = [...baseEntityMenuTypes];
+    if (isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES')) {
+        entityMenuTypes = [...baseEntityMenuTypes, ...splitComponentMenuTypes];
+    } else {
+        entityMenuTypes = [...baseEntityMenuTypes, ...componentMenuType];
+    }
 
     const cveFilterButtons = [
         {
--- a/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
@@ -11,6 +11,7 @@ import RadioButtonGroup from 'Components/RadioButtonGroup';
 import workflowStateContext from 'Containers/workflowStateContext';
 import { DASHBOARD_LIMIT } from 'constants/workflowPages.constants';
 import DashboardMenu from 'Components/DashboardMenu';
+import useFeatureFlags from 'hooks/useFeatureFlags';
 import PoliciesCountTile from '../Components/PoliciesCountTile';
 import CvesCountTile from '../Components/CvesCountTile';
 import ImagesCountTile from '../Components/ImagesCountTile';
@@ -34,6 +35,7 @@ const entityMenuTypes = [
 const VulnDashboardPage = ({ history }) => {
     const workflowState = useContext(workflowStateContext);
     const searchState = workflowState.getCurrentSearchState();
+    const { isFeatureFlagEnabled } = useFeatureFlags();
 
     const cveFilterButtons = [
         {
--- a/ui/apps/platform/src/Containers/VulnMgmt/Entity/Component/VulnMgmtComponentOverview.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Entity/Component/VulnMgmtComponentOverview.js
@@ -7,7 +7,7 @@ import TopCvssLabel from 'Components/TopCvssLabel';
 import RiskScore from 'Components/RiskScore';
 import Metadata from 'Components/Metadata';
 import CvesByCvssScore from 'Containers/VulnMgmt/widgets/CvesByCvssScore';
-import { entityGridContainerClassName } from 'Containers/Workflow/WorkflowEntityPage';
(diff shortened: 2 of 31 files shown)


ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
```diff
@@ -23,16 +24,21 @@ import MostCommonVulnerabilities from '../widgets/MostCommonVulnerabilities';
...
     const workflowState = useContext(workflowStateContext);
     const searchState = workflowState.getCurrentSearchState();
+    const { isFeatureFlagEnabled } = useFeatureFlags();
+
+    let entityMenuTypes = [...baseEntityMenuTypes];
+    if (isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES')) {
+        entityMenuTypes = [...baseEntityMenuTypes, ...splitComponentMenuTypes];
+    } else {
+        entityMenuTypes = [...baseEntityMenuTypes, ...componentMenuType];
+    }
```
sachaudh, line 41: Feature flagged the entities we show in the VM dashboard menu dropdown

Screen Shot 2022-06-24 at 4 11 36 PM
Screen Shot 2022-06-24 at 4 11 53 PM

ui/apps/platform/src/Containers/VulnMgmt/List/Images/VulnMgmtListImages.js
```diff
@@ -34,7 +35,9 @@ export const defaultImageSort = [
     },
 ];
 
-export function getCurriedImageTableColumns(watchedImagesTrigger) {
+export function getCurriedImageTableColumns(watchedImagesTrigger, isFeatureFlagEnabled) {
+    const isFrontendVMUpdatesEnabled = isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES');
```
sachaudh, line 39: Feature flagging the count link for components

ui/apps/platform/src/utils/entityRelationships.ts
```diff
@@ -86,6 +88,8 @@ const entityRelationshipMap: Record<string, EntityRelationshipData> = {
         // extendedMatches: [entityTypes.POLICY]
     },
     [entityTypes.NODE]: {
+        // @TODO: Uncomment this once we're using the new entity
+        // children: [entityTypes.NODE_COMPONENT],
```
sachaudh, line 92: Open to suggestions on how to feature flag the following in `entityRelationships`. Since it's a separate module and outside the React lifecycle, I can't use the feature flag hook here

ui/apps/platform/src/Containers/VulnMgmt/List/NodeComponents/VulnMgmtListNodeComponents.js
```diff
@@ -0,0 +1,228 @@
...
+const VulnMgmtNodeComponents = ({ selectedRowId, search, sort, page, data, totalResults }) => {
+    const query = gql`
+        query getComponents($query: String, $pagination: Pagination) {
+            results: nodeComponents(query: $query, pagination: $pagination) {
+                ...nodeComponentFields
+            }
+            count: nodeComponentCount(query: $query)
+        }
+        ${VULN_NODE_COMPONENT_LIST_FRAGMENT}
+    `;
```
vjwilson, line 198: I wonder if it is worth doing 2 different sets of files at this "container" level. We should compare the complexity of the single CVE list file with conditionals for the three new types of CVEs, and decide which way is better. Then we made them consistent one way or the other.

ui/apps/platform/src/Containers/VulnMgmt/VulnMgmt.fragments.js
```diff
@@ -243,7 +243,7 @@ export const VULN_CVE_LIST_FRAGMENT = gql`
 `;
 
 export const IMAGE_CVE_LIST_FRAGMENT = gql`
-    fragment cveFields on ImageVulnerability {
+    fragment imageCVEFields on ImageVulnerability {
```
vjwilson, line 246: I think we need to use this new name in the file `ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js`  above, at line 321.

ui/apps/platform/src/utils/entityRelationships.ts
vjwilson: As long as we feature flag the links, it may not be worth feature-flagging the direct "sub-routes" built from this map. (Re-factoring this map everywhere it is used would be a risky change.)

vjwilson (approved):
A good foundation.
A couple of places I think we should revisit as we integrate the split CVE types, and one stylistic suggestion for tile sidebar width.

Regardless, let's merge once tests pass.

ui/apps/platform/src/Containers/VulnMgmt/Entity/Component/VulnMgmtComponentOverview.js
```diff
@@ -73,9 +75,11 @@ function VulnMgmtComponentOverview({ data, entityContext }) {
         );
     }
 
-    const currentEntity = { [entityTypes.COMPONENT]: id };
+    const currentEntity = { [currentEntityType]: id };
     const newEntityContext = { ...entityContext, ...currentEntity };
 
+    const entityGridContainerClassName = `${entityGridContainerBaseClassName} grid-columns-1 md:grid-columns-2 lg:grid-columns-2`;
```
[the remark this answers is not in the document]
sachaudh: So this is the before and after of this change

Before:
Screen Shot 2022-06-29 at 10 59 12 PM

After:
Screen Shot 2022-06-29 at 10 59 15 PM

I noticed that we were only showing two columns in the component overview so it differed from the class we were using in general, which was for 3 columns. There is a base class which I use and then append classes specific for this view

ui/apps/platform/src/Containers/VulnMgmt/Entity/RelatedEntitiesSideList.js
```diff
@@ -46,7 +48,7 @@ const RelatedEntitiesSideList = ({ entityType, data, altCountKeyMap, entityConte
     }
     return (
         <div
-            className={` h-full relative border-base-100 border-l w-32 ${
+            className={` h-full relative border-base-100 border-l w-43 ${
```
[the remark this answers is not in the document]
sachaudh: Hmm, yea that's true. This probably wasn't the best solution. I could do two things:

1. Remove the width restriction and add a max-width and wrapping
2. Keep the width restriction and just add wrapping

I'll try #2 which is what you're suggesting.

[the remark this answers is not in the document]
sachaudh: Seems like text wrapping does look a bit odd (at least to me)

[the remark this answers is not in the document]
sachaudh: I ended up doing #1 instead. So I'll remove the `w-32`, add `max-w-43`, and add a text wrap class in case the text might be longer than that

Screen Shot 2022-06-29 at 11 28 57 PM
Screen Shot 2022-06-29 at 11 29 01 PM
Screen Shot 2022-06-29 at 11 29 37 PM

ui/apps/platform/src/Containers/VulnMgmt/List/NodeComponents/VulnMgmtListNodeComponents.js
sachaudh: We can try to leave this as-is for now, and come back tomorrow and make refactors so it's more consistent with your changes

ui/apps/platform/src/Containers/VulnMgmt/VulnMgmt.fragments.js
sachaudh: Oh, good catch

ui/apps/platform/src/Containers/VulnMgmt/Entity/Component/VulnMgmtComponentOverview.js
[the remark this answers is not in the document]
vjwilson: ^ That's probably a side-effect of widening the tiles right sidebar.

I'm fine with your fix for it, but I'm afraid this may turn into whack-a-mole. Let's get UX suggestion on this. cc: @alanonthegit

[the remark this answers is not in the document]
sachaudh: Oh no, that wasn't the issue. We're using grid columns for the overview and it was using a 3 column grid before. If you look at the first image you'll see it looks like you can add another third element to the right. It was on master as well. It's not really related to splitting the components but I just did it anyway.

sachaudh:
@vjwilson I think the failed ui-unit-tests aren't related to this PR. Maybe @dvail you can let me know in case it is, since that component might be related to your recent changes

dvail:
@sachaudh You're right, this is related to my stuff on not this PR. Investigating why this passed on my PR branch and not master now... :/

sachaudh:
> @sachaudh You're right, this is related to my stuff on not this PR. Investigating why this passed on my PR branch and not master now... :/

Ah, yea let me know. I don't think I can push my changes in until it's fixed. I'll try re-running

PR #2165, opened 2022-06-23
Add resource replacements within UI

dhaus67:
## Description

Within the PR of [ROX-8520](https://github.com/stackrox/stackrox/pull/1384) some resources have been grouped and replaced. This will land within 3.71 release.
What's currently not done properly is to respect the newly created resources as replacements within the UI, i.e. allowing access to existing pages by
- either having access to the old, to-be-deprecated resource.
- having access to the newly introduced resource.

This PR adds the support for this, by maintaining a map of old, to-be-deprecated resources as well as their newly created counterparts.

Within [ROX-11453](https://issues.redhat.com/browse/ROX-11453), the plan is to fully deprecate the old resource once the grace period is over, and simplify existing authorization checks.

### DIFF
--- a/ui/apps/platform/src/hooks/usePermissions.ts
+++ b/ui/apps/platform/src/hooks/usePermissions.ts
@@ -24,22 +24,88 @@ const stateSelector = createStructuredSelector<{
     isLoadingPermissions: selectors.getIsLoadingUserRolePermissions,
 });
 
+// TODO(ROX-11453): Remove this mapping once the old resources are fully deprecated.
+const replacedResourceMapping = new Map([
+    ['AllComments', 'Administration'],
+    ['APIToken', 'Integration'],
+    ['AuthPlugin', 'Access'],
+    ['AuthProvider', 'Access'],
+    ['BackupPlugins', 'Integration'],
+    ['ComplianceRuns', 'Compliance'],
+    ['ComplianceRunSchedule', 'Administration'],
+    ['Config', 'Administration'],
+    ['DebugLogs', 'Administration'],
+    ['Group', 'Access'],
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #2165** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2165" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2165?utm_source=stack-comment).

vjwilson:
@dhaus67 I'm fine with the code, but I'd like @pedrottimark to review it as well.

I just kicked off another test run, because the failure looked like an infrastructure problem. If it still doesn't pass, you may need to check with Gavin. (Could just require a rebase for CI fixes)

dhaus67:
@vjwilson Rebase did its magic, the CI is passing now. 
This would be essential to get into the 3.71 release, as otherwise we could have potential bug reports where the access to pages is not given with the new resources. Thus review would be highly appreciated 😅

pedrottimark (approved):
Thank you for making the best of the situation where usePermissions did not yet completely replace the functions in roles reducer. That looks like a follow up for UI team in 72.

pedrottimark:
Can you say here in comment something certainly said elsewhere: the new second `if` statements assume during the transition period that each newly created resource is initialized with the minimum access level of the corresponding replacement resources? For example,
* READ_WRITE_ACCESS only if all corresponding have READ_WRITE_ACCESS
* otherwise READ_ACCESS only if all corresponding have at least READ_ACCESS
* otherwise NO_ACCESS if any corresponding have NO_ACCESS

PR #2168, opened 2022-06-23
Add last updated time display at top of dashboard

dvail:
## Description

As described in title. The timestamp will only update when the page is refreshed.

Note that UX has decided to not use the circular arrow icon shown in [the mocks](https://www.sketch.com/s/9d082d82-3515-40f2-875a-1b322a02bace/a/1KQPyRp).

## Checklist

## Testing Performed

Load the new dashboard and view the time the summary count request completed in the top right of the page.
image

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
@@ -54,8 +54,9 @@ const tileLinks: Record<TileEntity, string> = {
     Secret: `${configManagementPath}/${urlEntityListTypes[resourceTypes.SECRET]}`,
 };
 
-const dateFormatter = new Intl.DateTimeFormat('en-US');
-const timeFormatter = new Intl.DateTimeFormat('en-US', { hour: 'numeric', minute: 'numeric' });
+const locale = window.navigator.language ?? 'en-US';
+const dateFormatter = new Intl.DateTimeFormat(locale);
+const timeFormatter = new Intl.DateTimeFormat(locale, { hour: 'numeric', minute: 'numeric' });
 
 function SummaryCounts() {
     const [lastUpdate, setLastUpdate] = useState<Date>(new Date());
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
@@ -2,7 +2,6 @@ import React, { useState } from 'react';
 import { gql, useQuery } from '@apollo/client';
 import Raven from 'raven-js';
 import pluralize from 'pluralize';
-import { format } from 'date-fns';
 import {
     Alert,
     Button,
@@ -54,6 +53,9 @@ const tileLinks: Record<TileEntity, string> = {
     Image: vulnManagementImagesPath,
     Secret: `${configManagementPath}/${urlEntityListTypes[resourceTypes.SECRET]}`,
 };
+
+const dateFormatter = new Intl.DateTimeFormat('en-US');
+const timeFormatter = new Intl.DateTimeFormat('en-US', { hour: 'numeric', minute: 'numeric' });
 
 function SummaryCounts() {
     const [lastUpdate, setLastUpdate] = useState<Date>(new Date());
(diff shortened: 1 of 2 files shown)


dvail:
* master
  * **PR #2168** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2168" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2168?utm_source=stack-comment).

ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
```diff
@@ -80,24 +95,36 @@ function SummaryCounts() {
...
+                        </Button>
+                    ))}
+                </Split>
+            </SplitItem>
+            <div
+                style={{ fontStyle: 'italic' }}
+                className="pf-u-color-200 pf-u-font-size-sm pf-u-mr-md pf-u-mr-lg-on-lg"
+            >
+                {`Last updated ${dateFormatter.format(lastUpdate)} at ${timeFormatter.format(
+                    lastUpdate
```
pedrottimark, line 125: For your info, here is a similar example with current dateUtils:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Clusters/Components/SensorStatus.tsx#L54

How high a priority was this specific date format? which differs from everywhere else in UI, as far as I know.

@vjwilson do you have any thought about pro and con of diverging from reusable functions?

pedrottimark (approved):
One concern about pro and con of unique date format, but it is not a blocker to merge.

ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
pedrottimark: Here is another precedent where the format coulda shoulda been extracted to dateUtils:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Network/Graph/Overlays/NodesUpdateSection.tsx#L46-L49

Last_Updated_Network_Graph

dvail: Good insight, I'll bring this up with UX as well. Gut feeling is we should use the same date/time format for Clusters and this dashboard, and it makes sense to use just the time for network graph. Even if we leave the date off though I don't see a compelling reason for the respective time formats to be different.

vjwilson: Late to the discussion, but just a  couple of thoughts
1. Eventually, we need to converge on what the rest of PatternFly/OpenShift uses, and if they don't have a standard, we need to tell them that our format is the standard and they should converge on it. ;) 
2. Even after that, we should code date/times in a way that they are internationalized, based on your local settings.

pedrottimark: @vjwilson PatternFly element for date and time that we have heard about in Design Share is still future.

@dvail Can you include in the ticket the issue of when (if ever) local time zone (as opposed to locale) which UI displays now is appropriate instead of universal time zone:
* which supports clear communication by global teams
* which usually correlates with log files

pedrottimark: If universal time is most appropriate for security, then the geekier format is tends to communicate it clearly.

pedrottimark:
Here is a delayed question after reviewing the next component: how does its response affect the time. Answer: It doesn’t.

Was there any discussion whether the time might be interpreted as applying to more than the summary counts?

ui/apps/platform/src/Containers/Dashboard/PatternFly/SummaryCounts.tsx
dvail: https://issues.redhat.com/browse/ROX-11527

dvail:
@pedrottimark  I don't think there was a ton of discussion specific to that possible confusion, but I believe that was part of the reasoning that the time was placed up alongside the system wide counts. Similarly to how the scope filters do not change the values of those counts, it was placed in the top location less so because it relates to the new dashboard functionality directly and more because that is the place that customers are accustomed to seeing the system overview.

PR #2171, opened 2022-06-23
Optionally unsilence some make commands.

porridge:
## Description

This is useful when debugging makefile issues.
No change to default behaviour.

## Checklist

## Testing Performed

- Ran `make proto-generated-srcs SILENT=` and saw the whole command as expected. 
- Ran `make proto-generated-srcs` and command was **not** echoed as expected.


Makefile
```diff
@@ -4,6 +4,8 @@ ROX_PROJECT=apollo
 TESTFLAGS=-race -p 4
 BASE_DIR=$(CURDIR)
 
+SILENT ?= @
```
janisz, line 7: How about adding doc for this?

PR #2173, opened 2022-06-23
ROX-11515: Support custom names in release versions

0x656b694d:
## Description

Support `A.B.C-N` versions in the upstream release workflow, where A, B, C are numbers and N - alphanumeric:

* The release branch will be formatted as `A.B.x-N`;
* The documentation submodule will still reference `rhacs-docs-A.B.C`;
* The milestones will look like `A.B.C-N-rc.D`.

*This PR does not include any changes in the produced image tags triggered by CI builds*. This will need to be done separately.

## Testing Performed

Manual testing in `stackrox/test-gh-actions` repository.

### DIFF
--- a/.github/workflows/variables.yml
+++ b/.github/workflows/variables.yml
@@ -3,7 +3,7 @@ on:
   workflow_call:
     inputs:
       version:
-        description: Version (A.B.C or A.B.C-rc.D)
+        description: Version (A.B.C[-N][-rc.D])
         type: string
         required: true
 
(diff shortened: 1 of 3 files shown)


.github/workflows/variables.yml
```diff
@@ -41,13 +44,15 @@ jobs:
...
+      dash-name: ${{steps.parse.outputs.dash-name}}
+      next-rc: ${{steps.parse.outputs.next-milestone}}
     steps:
       - id: parse
         run: |
-          read RELEASE PATCH_NUMBER RC_NUMBER <<< \
-          $(sed -rn 's/^([0-9]+\.[0-9]+)\.([0-9]+)(-rc\.([0-9]+))?$/\1 \2 \4/p' <<< ${{inputs.version}})
+          set -u
+          read RELEASE PATCH_NUMBER RELEASE_NAME RC_NUMBER <<< \
+          $(sed -rn 's/^([0-9]+\.[0-9]+)\.([0-9]+)(-([a-zA-Z0-9-]+))?(-rc\.([0-9]+))?$/\1 \2 \4 \6/p' <<< ${{inputs.version}})
```
vikin91, line 55: I am not fully sure about this group: `(-([a-zA-Z0-9-]+))?`

It would allow such things as: `x.y.z---rc.0`

How about we allow only valid kebab-case strings (that start with `-`) using: `(-([a-z][a-z0-9]*)(-[a-z0-9]+)*)`? 
(capitals can be optional - I think, we do not need them.)

vikin91 (approved):
Looks good, but one regex group could be improved.

.github/workflows/variables.yml
0x656b694d: Well, that's possible, why not. Although the "official" semver regexp permits such weird strings.

PR #2174, opened 2022-06-23
Rename github.event.inputs to just inputs, which is the same

0x656b694d:
```
sed -i "s/github.event.inputs/inputs/g" .github/workflows/*.yml
```

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -7,9 +7,9 @@ on:
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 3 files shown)


vikin91 (approved):
LGTM. 
Does it work only for `inputs`, or can we drop also other `github.event` occurrences?

0x656b694d:
> LGTM. Does it work only for `inputs`, or can we drop also other `github.event` occurrences?

Not only, but not all other occurrences. I don't think there are other shortcuts that are not used.

PR #2177, opened 2022-06-23
Add PID and args to mock gRPC output

Stringy:
## Description

Adds PID and process args to mock gRPC output to improve test coverage on collector. 

## Checklist

## Testing Performed

Tested manually with collector's integration tests specifically, with various versions of collector that have a bug in which process args were not reported. Tested alongside collector as part of https://github.com/stackrox/collector/pull/740

### DIFF
--- a/integration-tests/mock-grpc-server/main.go
+++ b/integration-tests/mock-grpc-server/main.go
@@ -49,7 +49,7 @@ func (s *signalServer) PushSignals(stream sensorAPI.SignalService_PushSignalsSer
 			processSignal = signal.GetSignal().GetProcessSignal()
 		}
 
-		processInfo := fmt.Sprintf("%s:%s:%d:%d", processSignal.GetName(), processSignal.GetExecFilePath(), processSignal.GetUid(), processSignal.GetGid())
+		processInfo := fmt.Sprintf("%s:%s:%d:%d:%d:%s", processSignal.GetName(), processSignal.GetExecFilePath(), processSignal.GetUid(), processSignal.GetGid(), processSignal.GetPid(), processSignal.GetArgs())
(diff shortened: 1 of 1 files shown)


Molter73 (approved):
LGTM! I think it might be a good idea to use the tag created in the CI run for the gRPC server in stackrox/collector#740 before we merge anything, just to make double sure everything works. You can find the tag in the `build-scale-monitoring-and-mock-server` step.

Molter73 (approved):

integration-tests/mock-grpc-server/main.go
```diff
@@ -49,9 +50,9 @@ func (s *signalServer) PushSignals(stream sensorAPI.SignalService_PushSignalsSer
 			processSignal = signal.GetSignal().GetProcessSignal()
 		}
 
-		processInfo := fmt.Sprintf("%s:%s:%d:%d", processSignal.GetName(), processSignal.GetExecFilePath(), processSignal.GetUid(), processSignal.GetGid())
+		processInfo := fmt.Sprintf("%s:%s:%d:%d:%d:%s", processSignal.GetName(), processSignal.GetExecFilePath(), processSignal.GetUid(), processSignal.GetGid(), processSignal.GetPid(), processSignal.GetArgs())
```
Molter73, line 53: I just realized while looking at [this check](https://github.com/stackrox/collector/blob/f5964cbc35478f957aa800476b66aa53fd3df5cb/integration-tests/process.go#L19-L21) that if any of the fields has a `:` this simple serialization would completely break. For now, keeping the arguments as the last part should work fine, but we might want to look into using a proper serializer or even encoding the string parts of the data as base64.

PR #2179, opened 2022-06-23
ROX-11283 re-sync tests with recorded events (final state of deployments)

lvalerom:
## Description

This PR aims to provided testing for sensor's re-sync behaviour using pre-recorded k8s events.

The events were recorded using the `local-sensor` tool.

* Create the `trace.jsonl` file containing the k8s events.

```
go run tools/local-sensor/main.go -record -record-out=trace.jsonl -resync=0s
```

* Create the `central-out.bin` file containing sensor's output events.

```
go run tools/local-sensor/main.go -replay -replay-in=trace.jsonl -resync=10s -delay=0s -format=raw -central-out=central-out.bin
```

## Checklist

## Testing Performed

Manually testing with:

```
LOGLEVEL=panic go test github.com/stackrox/rox/sensor/tests/replay -count=1
```


sensor/debugger/k8s/fake.go
```diff
@@ -318,9 +400,48 @@ func (f *FakeEventsManager) createEvent(msg resources.InformerK8sMsg, ch chan<-
...
+		receivedEvents := 0
+		for {
+			timeout := time.After(5 * time.Second)
+			select {
+			case <-timeout:
+				return errors.New("timeout reached waiting for event")
+			case event, more := <-f.AckChannel:
+				if !more {
+					return errors.New("channel closed")
+				}
```
vikin91, line 421: Is it really an error? In which situations the channel gets closed prematurely?

sensor/debugger/k8s/fake.go
```diff
@@ -318,9 +400,48 @@ func (f *FakeEventsManager) createEvent(msg resources.InformerK8sMsg, ch chan<-
...
+						return fmt.Errorf("compare function for resource '%s' not found", resource.Type().String())
+					}
+					if compareFunc(sensorEvent.GetResource(), event) {
+						receivedEvents++
+						break
+					}
+				}
+				if receivedEvents == len(events) {
+					return nil
+				}
```
vikin91, line 442: It looks like a good candidate for refactoring into a separate func.

tools/allowed-large-files
```diff
@@ -30,6 +30,8 @@ qa-tests-backend/test-images/registry-image/tomcat-files/web.xml
 roxctl/helm/derivelocalvalues/testdata/k8s-resources/*.yaml
 scale/image/fixtures/scrape.json
 scale/mocksensor/image_names.go
+sensor/tests/replay/data/central-out.bin
+sensor/tests/replay/data/trace.jsonl
```
vikin91, line 34: Naming: do we focus on testing anything particular here? If not, then maybe we could name this file using the term `safety net`?

sensor/tests/replay/replay_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+type TraceWriterWithChannel struct {
+	destinationChannel chan *central.SensorEvent
+	// mu mutex to avoid multiple goroutines writing at the same time
+	mu sync.Mutex
+}
+
+func (tw *TraceWriterWithChannel) Write(b []byte) (int, error) {
+	tw.mu.Lock()
+	defer tw.mu.Unlock()
```
vikin91, line 104: What is the shared resource that we protect with the mutex? The channel?

If yes, then I am not sure that we need it.

sensor/tests/replay/replay_test.go
```diff
@@ -0,0 +1,300 @@
...
+	spyCentral.ConnectionStarted.Wait()
+
+	cases := map[string]struct {
+		k8sEventsFile    string
+		sensorOutputFile string
+	}{
+		"Safety net test": {
+			k8sEventsFile:    "data/trace.jsonl",
+			sensorOutputFile: "data/central-out.bin",
+		},
```
vikin91, line 156: I guessed it with the safety net ;)

sensor/tests/replay/replay_test.go
```diff
@@ -0,0 +1,300 @@
...
+					assert.Equal(t, exp.GetPod().GetName(), e.GetPod().GetName())
+					assert.Equal(t, exp.GetPod().GetNamespace(), e.GetPod().GetNamespace())
+				}
+			}
+			cancelFunc()
+		})
+	}
+}
+
+func readSensorOutputFile(fname string) ([]*central.MsgFromSensor, error) {
```
vikin91, line 236: Pure dark magic here :) 

I guess this code is vulnerable to changes to the proto definition of the `MsgFromSensor`. How can we prevent it going out of sync?

sensor/debugger/k8s/fake.go
vikin91: Found it - we close the channel when the tests are done. So this is not an error - I suggest to: (1) ignore this case, (2) return nil. I think, that option (2) maybe cleaner.

vikin91 (changes requested):
Looks really good! Just please check the handling of the closing of the ack-chan and the comment about the mutex.

sensor/debugger/k8s/fake.go
lvalerom: I decided to ignore as you suggested. If we return nil we can land in a situation where the channel is closed and we keep processing events without waiting thus making the results of the test unpredictable.

sensor/debugger/k8s/fake.go
lvalerom: Done

tools/allowed-large-files
lvalerom: I've added the prefix `safety-net` 🙂

sensor/tests/replay/replay_test.go
lvalerom: 👍

sensor/tests/replay/replay_test.go
lvalerom: It is vulnerable to changes. I don't know how to work around that. I think the entire suit of tests is sensitive to changes in the proto definitions. If we change something there, we can expect the tests might stop working.

sensor/debugger/k8s/fake.go
```diff
@@ -318,9 +400,56 @@ func (f *FakeEventsManager) createEvent(msg resources.InformerK8sMsg, ch chan<-
...
+			return nil
+		}
+		receivedEvents := 0
+		for {
+			timeout := time.After(5 * time.Second)
+			select {
+			case <-timeout:
+				return errors.New("timeout reached waiting for event")
+			case event := <-f.AckChannel:
+				eventFound, err := isEventInSlice(event, events)
```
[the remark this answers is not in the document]
lvalerom: `isEventInSlice` will return false if the `event` is empty. I think we don't need to change anything there.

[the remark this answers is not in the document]
vikin91: Yes, it will return false but it also returns an error and this may be not what we want. Maybe you could check whether the object parsed from json is not 0-valued?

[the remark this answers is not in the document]
lvalerom: I'm not sure if I'm understanding what you're saying 🤔 We are not parsing the event received in the channel. We parse the list of events read from the trace file. The event received is passed to the `compareFunc` where we read the specific resource (e.g. `event.GetNamespace()`) This function will return nil if the event is empty or nil. So, at the end we return `false, nil` if the event is empty. We're only returning an error here if the trace file contains events that cannot be parsed.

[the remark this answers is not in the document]
vikin91: Oh, many thanks for the explanation, now this is clear :)

Now I wonder if there is a way to not parse the entire array on each message but to do it once somewhere earlier - do you think this would be possible? (separate PR is okay)

vikin91 (approved):
Thank you for improving!

sensor/debugger/k8s/fake.go
```diff
@@ -318,9 +400,56 @@ func (f *FakeEventsManager) createEvent(msg resources.InformerK8sMsg, ch chan<-
...
+func isEventInSlice(event *central.SensorEvent, events []string) (bool, error) {
+	for _, e := range events {
+		sensorEvent := &central.SensorEvent{}
+		if err := jsonpb.UnmarshalString(e, sensorEvent); err != nil {
+			return false, fmt.Errorf("error unmarshaling '%s'", e)
+		}
+		if sensorEvent.GetResource() == nil {
+			return false, fmt.Errorf("resource not found in sensor event '%s'", e)
+		}
+		resource := reflect.ValueOf(sensorEvent.GetResource())
```
fredrb, line 445: If you don't need the reflected value (`reflect.Value` object), it's better to use `reflect.TypeOf(...)` here. It will only return the type spec. It's the same thing returned by `resource.Type()`.

sensor/tests/replay/replay_test.go
```diff
@@ -0,0 +1,295 @@
...
+		if event.GetDeployment() != nil {
+			if event.GetDeployment().GetId() != "" {
+				events[event.GetDeployment().GetId()] = event
+			}
+		}
+	}
+	return events
+}
+
+func getAllPods(messages []*central.MsgFromSensor) map[string]*central.SensorEvent {
```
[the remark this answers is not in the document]
lvalerom: Done

sensor/debugger/k8s/fake.go
lvalerom: Done

sensor/tests/replay/replay_test.go
lvalerom: I think it makes sense, I also don't remember why this wouldn't be a good idea. Maybe we can open a follow-up task and investigate. What do you think?

lvalerom:
> LGTM. The only thing I'm not too happy about is how `k8s/fake.go` (this also applies to other `debugger/*` packages) is becoming a file with utility functions. Do you think it makes sense to extract all the record/replay functionalities into its own package?

Yeah, I think so too. Maybe we can create a utils package at `sensor/utils`. Or maybe `sensor/utils/replay` to separate it from possible future utilities. What do you think?

sensor/tests/replay/replay_test.go
fredrb: Sounds good

sensor/debugger/k8s/fake.go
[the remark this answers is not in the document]
vikin91: Thanks, it looks good :) It could be even a separate function ;)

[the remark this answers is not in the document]
lvalerom: Done 🙂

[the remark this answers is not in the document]
vikin91: Thanks!

sensor/tests/replay/replay_test.go
lvalerom: Created task for this [ROX-11566](https://issues.redhat.com/browse/ROX-11566)

PR #2182, opened 2022-06-23
[ROX-11282 + ROX-11517] : Update ClusterResponse and ClusterList response to include cluster retention metadata; Make clusters searchable by last contact time.

charmik-redhat:
## Description

- Created type `DecommissionedClusterRetentionInfo` to include retention info about a cluster such as days remaining until deletion and whether cluster is protected from deletion.
- Updated `ClusterResponse` to include `DecommissionedClusterRetentionInfo`
- Updated `ClusterList` response to include `DecommissionedClusterRetentionInfo` for unhealthy clusters
- Made `last_contact_time` searchable in ClusterHealthStatus

## Checklist

## Testing Performed

- Added unit tests to test cluster search using last_contact_time
- Added unit tests to test if cluster service sends correct retention info

TODO(replace-me)
screenshots.


central/cluster/service/service_impl_test.go
```diff
@@ -71,3 +80,158 @@ func (suite *ClusterServiceTestSuite) TestGetClusterDefaults() {
...
+
+	clusters := []*storage.Cluster{
+		{
+			Id: "HEALTHY cluster",
+			HealthStatus: &storage.ClusterHealthStatus{
+				SensorHealthStatus: storage.ClusterHealthStatus_HEALTHY,
+			},
+		},
+		{
+			Id:     "UNHEALTHY cluster with a label matching an ignored label",
```
clickboo, line 169: matching a label to ignore the cluster

central/cluster/service/service_impl_test.go
```diff
@@ -71,3 +80,158 @@ func (suite *ClusterServiceTestSuite) TestGetClusterDefaults() {
...
+			Labels: map[string]string{"k1": "v2"},
+			HealthStatus: &storage.ClusterHealthStatus{
+				SensorHealthStatus: storage.ClusterHealthStatus_UNHEALTHY,
+				LastContact:        suite.timeBeforeDays(80),
+			},
+		},
+	}
+
+	expectedIds := []string{
+		"UNHEALTHY cluster with a label matching an ignored label",
```
clickboo, line 194: same text change as above

proto/api/v1/cluster_service.proto
```diff
@@ -23,8 +23,19 @@ enum LoadBalancerType {
     ROUTE         = 3;
 }
 
+// next available tag: 3
+message DecommissionedClusterRetentionInfo {
+    oneof RetentionInfo {
+        // indicates whether a cluster is protected from deletion
+        bool is_protected_cluster    = 1;
```
clickboo, line 30: naming suggestion earlier

proto/storage/cluster.proto
```diff
@@ -296,7 +296,7 @@ message ClusterHealthStatus {
 
     // For sensors not having health capability, this will be filled with gRPC connection poll. Otherwise,
     // this timestamp will be updated by central pipeline when message is processed
-    google.protobuf.Timestamp last_contact                      = 5;
+    google.protobuf.Timestamp last_contact                      = 5 [(gogoproto.moretags) = "search:\"Last Contact,store\""];;
```
clickboo, line 299: What is the reason to make this searchable?

proto/api/v1/cluster_service.proto
pedrottimark: UI code convention strongly prefers `is` or `has` prefix for boolean state values, so also prefer something like `is_excluded_from_decommissioning` which becomes `isExcludedFromDecommissioning` in JSON

central/cluster/service/service_impl.go
```diff
@@ -102,8 +108,70 @@ func (s *serviceImpl) getCluster(ctx context.Context, id string) (*v1.ClusterRes
...
+
+	clusterRetentionConfig := sysConfig.GetPrivateConfig().GetDecommissionedClusterRetention()
+	if clusterRetentionConfig == nil {
+		return nil, nil
+	}
+
+	if mapsIntersect(clusterRetentionConfig.GetIgnoreClusterLabels(), cluster.GetLabels()) {
+		return &v1.DecommissionedClusterRetentionInfo{
+			RetentionInfo: &v1.DecommissionedClusterRetentionInfo_IsProtectedCluster{
+				IsProtectedCluster: true,
```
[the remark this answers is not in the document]
pedrottimark: Agree, y'all can decide short or long here, but note double mm and double ss if you choose long name.

proto/api/v1/cluster_service.proto
```diff
@@ -23,8 +23,19 @@ enum LoadBalancerType {
...
+        // days after which cluster will be deleted if sensor health remains UNHEALTHY
+        int32 days_until_deletion  = 2;
```
pedrottimark, line 32: Forgive me for not reading code to answer this question: what is value if chluser

proto/storage/cluster.proto
charmik-redhat: Made that searchable so that the garbage collection task (coming soon) is able to search clusters with no contact since <x> days. The other way would be to search for all clusters with `SensorHealthStatus : UNHEALTHY` and then filter out those that have last_contact more than <x> days old. I can use the second approach if making searchable would significantly affect db storage size.

proto/api/v1/cluster_service.proto
charmik-redhat: @pedrottimark This will be embedded into `ClusterResponse`, and cluster response will already have the cluster. 
```
message ClusterResponse {
     storage.Cluster cluster                                                                 = 1;
     DecommissionedClusterRetentionInfo cluster_retention_info  = 2;
 }
```

In case of `CluserList` , the ClusterList will have a map<clusterID -> DecommissionedClusterRetentionInfo > that will have entries for all `SensorStatus UNHEALTHY` clusters.
```
message ClustersList {
     repeated storage.Cluster clusters                                                                                 = 1;
     // Maps 'UNHEALTHY' clusters' IDs to their retention info
     map<string, DecommissionedClusterRetentionInfo> cluster_id_to_retention_info = 2;
 }
```

proto/storage/cluster.proto
clickboo: @connorgorman We've been mindful about adding search fields and stuff to index hence my question ^. I remember a concern that Conor had in the past.

connorgorman: This should be fine. Really the nested one is the main concern

clickboo: If we don't have a UI/UX requirement to make clusters searchable by this, I would avoid adding to index. But maybe at the scale of clusters this really doesn't matter.

connorgorman: It's actually not nested because we save it separately then reconstruct when grabbing the cluster

central/cluster/service/service_impl.go
```diff
@@ -159,3 +259,14 @@ func (s *serviceImpl) GetClusterDefaultValues(ctx context.Context, _ *v1.Empty)
 	}
 	return defaults, nil
 }
+
+func mapsIntersect(m1 map[string]string, m2 map[string]string) bool {
```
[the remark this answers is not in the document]
charmik-redhat: Yes, that is correct, I will refactor it out in another PR.

clickboo (approved):

md2119 (commented):
I feel we could really benefit from having health/status endpoints for clusters. That way we can move health out of the cluster proto and make it a top-level.

PR #2183, opened 2022-06-23
RS-443: Trigger nightly tests in prow also

gavin-stackrox:
## Description

It is unclear to me why, but this change (`--follow-tags`) is required to get OpenShift CI postsubmit jobs to run for nightly tags. From all the jobs that my test tags kicked off in Circle CI, it looks like it works there also.

## Checklist

## Testing Performed

CI is sufficient


sbostick (approved):
LGTM

PR #2188, opened 2022-06-24
Ensure no parallel reprocessing runs are triggered

dhaus67:
## Description

In case the reprocessing loop takes longer than the ticker (i.e. in environments with a high scale), we shouldn't run two reprocessing runs in parallel, as this would slow the system down further.

Instead, check before triggering the reprocess whether a reprocessing run is currently in progress.

### DIFF
--- a/central/reprocessor/reprocessor.go
+++ b/central/reprocessor/reprocessor.go
@@ -126,9 +126,7 @@ func newLoopWithDuration(connManager connection.Manager, imageEnricher imageEnri
 
 		signatureVerificationSig: concurrency.NewSignal(),
 
-		// Used for testing purposes
-		reprocessingStarted:  concurrency.NewSignal(),
-		reprocessingComplete: concurrency.NewSignal(),
+		reprocessingInProgress: concurrency.NewSignal(),
 
 		connManager: connManager,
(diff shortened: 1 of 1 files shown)


dhaus67:
* master
  * **PR #2188** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2188" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2188?utm_source=stack-comment).

connorgorman (approved):

viswajithiii:
Instead of this, wouldn't it be better to just start the countdown for the next reprocessing when the current one is done?

connorgorman:
> Instead of this, wouldn't it be better to just start the countdown for the next reprocessing when the current one is done?


@viswajithiii At that point, we lose our try to do this every 4 hours right?

viswajithiii:
> @viswajithiii At that point, we lose our try to do this every 4 hours right?

Yeah, fair enough. In that case, though, shouldn't we ensure that we trigger a new reprocessing run right away after the old one is done, _if_ we skip a reprocessing run via the code introduced in this PR?

PR #2189, opened 2022-06-24
Short-circuit in image enricher if an error occurred fetching image metadata.

dhaus67:
## Description

Currently, if the image's metadata cannot be fetched, we will still attempt to scan / fetch signatures for it.
However, these actions should not succeed, since no layer information will be present for the scanner as well as registry connection / access is not available for signature fetching.

If there's any specific case where that would not be the case (i.e. scanner somehow having a cache for the image based on its name instead of its layer SHAs (?)), then we would have to add it within the current proposed changes, 
as the assumption is currently: no image metadata == failing image scan + fetching signature.

## Testing Performed
- see CI / Unit tests.

### DIFF
--- a/pkg/images/enricher/enricher_impl.go
+++ b/pkg/images/enricher/enricher_impl.go
@@ -136,16 +136,27 @@ func (e *enricherImpl) EnrichImage(ctx context.Context, enrichContext Enrichment
 		imageNoteSet[note] = struct{}{}
 	}
 
+	// Ensure we set the correct image notes when returning, also during short-circuting.
+	defer setImageNotes(image, imageNoteSet)
+
 	// Signals whether any updates to the image were made throughout the enrichment flow.
 	var updated bool
 
 	didUpdateMetadata, err := e.enrichWithMetadata(ctx, enrichContext, image)
-	errorList.AddError(err)
 	if image.GetMetadata() == nil {
 		imageNoteSet[storage.Image_MISSING_METADATA] = struct{}{}
 	} else {
 		delete(imageNoteSet, storage.Image_MISSING_METADATA)
 	}
+
+	// Short-circuit if image metadata could not be retrieved. This indicates that connection or authentication to the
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #2189** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2189" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2189?utm_source=stack-comment).

connorgorman:
Ah so there may be a case when this is possible if we are using a scanner that does not have a registry, then we may have a api token that works for the scanner, but not for the metadata?

dhaus67:
@connorgorman  Hm, I guess ? I'm confused by the expected behavior (specifically `TestZeroRegistryIntegrations` / `TestRegistryMissingFromImage`), how would the registries be different used by scanner compared to fetching metadata? [Looks like the same registry set is used](https://github.com/stackrox/stackrox/blob/1ec040affd21d0927804ae2bebe0dc4304aa18af/pkg/images/integration/set.go#L27-L30). Does this mean the test was just not aligning with the actual behavior?  I'm not really aware of the way things work within scanner so might be just some wrong assumption on my side.

pkg/images/enricher/enricher_impl.go
```diff
@@ -195,6 +200,13 @@ func (e *enricherImpl) EnrichImage(ctx context.Context, enrichContext Enrichment
 	}, errorList.ToError()
 }
 
+func setImageNotes(image *storage.Image, imageNoteSet map[storage.Image_Note]struct{}) {
+	image.Notes = image.Notes[:0]
+	for note := range imageNoteSet {
```
connorgorman, line 205: Can we actually sort this?

connorgorman (approved):
idk why we didn't do it before because we should deterministically sort the notes

PR #2190, opened 2022-06-24
Add "Policy violations by severity" widget

dvail:
## Description

Adds a Policy violations by severity widget. This widget contains counters for each severity class of violations based on the user's selected scope. The bottom of the widget also displays up to three of the most recent _critical_ violations the occurred within the scope. If there are less than three critical violations, less severe violations will not be shown.

## Checklist

## Testing Performed

Test that on initial load all system violation counts are displayed. Note that the total in the title matches the total of all four severities as well as the total in the summary counts at the top of the page.
image

Test that applying a cluster and namespace filter reduces the total counts to match violations in the chosen scope. The title of the widget still matches the total in the individual severity counts.
image

Test that an empty state is displayed when a scope is selected that results in no recent critical violations.
image

Test that the View All link takes the user to the violation page with the correct scope applied. (Note that Namespace ID filters are not present in the search bar due to [ROX-11234](https://issues.redhat.com/browse/ROX-11234))
image

Test that the individual tile links take the user to the violation page with the correct scope and severity filter applied.
image

Test that the policy name link in the most recent violations section takes the user to the violation page for that policy and the deployment in violation.
image


dvail:
* master
  * **PR #2190** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2190" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2190?utm_source=stack-comment).

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/MostRecentViolations.tsx
```diff
@@ -0,0 +1,75 @@
...
+                            </Td>
+                            <Td dataLabel="Deployment in violation">
+                                <Truncate content={deployment?.name ?? ''} />
+                            </Td>
+                            <Td
+                                width={35}
+                                className="pf-u-pr-0 pf-u-text-align-right-on-md"
+                                dataLabel="Time of last violation occurrence"
+                            >
+                                {getDateTime(time)}
```
dvail, line 65: This deviates from the mocks in favor of using our consistent dateTime helper functions. I have a pending follow up with UX to get direction on how we want to handle the various formats throughout the app.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicySeverity.tsx
```diff
@@ -0,0 +1,124 @@
...
+        data: alertCountData,
+        loading: alertCountLoading,
+        error: alertCountError,
+    } = useAlertGroups(countsFilterQuery);
+    const {
+        data: currentRecentAlertsData,
+        previousData: previousRecentAlertsData,
+        loading: recentAlertsLoading,
+        error: recentAlertsError,
+    } = useQuery<{ alerts: DeploymentAlert[] }>(mostRecentAlertsQuery, {
```
dvail, line 68: Using `DeploymentAlert` as the type here makes me a bit uneasy, as it isn't exactly what is being returned by the gql query even if all of the required fields do exist at runtime. `Partial<DeploymentAlert>` is probably more technically correct but required extensive null checks for properties that we never null.

pedrottimark:
Off the topic of the review, layout of this component brings to mind comment during sprint demo about order of severity:
* less severe to more severe
* more severe to less severe

policy_violations_by_severity

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicySeverity.tsx
pedrottimark: Any reason (asking, not recommending) not to use REST request like:

 /v1/alerts?query=Severity%3ACRITICAL_SEVERITY&pagination.offset=0&pagination.limit=3&pagination.sortOption.field=Violation%20Time&pagination.sortOption.reversed=true

whose response is `ListAlert` which has
* `ListAlertDeployment` https://github.com/stackrox/stackrox/blob/master/proto/storage/alert.proto#L202-L215
* `ListAlertPolicy` https://github.com/stackrox/stackrox/blob/master/proto/storage/alert.proto#L187-L200

pedrottimark (approved):
Thank you for writing tests as you go. Unit tests is a good call for these components.

PR #2192, opened 2022-06-24
ROX-11836: Network Policy Recommendation 

Maddosaurus:
## Description

This PR introduces a new subcommand to roxctl: `roxctl generate netpol <folder-path> --offline`.
It uses the [Cluster Topology Analyzer from NP-Guard](https://pkg.go.dev/github.com/np-guard/cluster-topology-analyzer@v1.2.0/pkg/controller) for this functionality.
Currently, it is in a preliminary (Tech Preview) state, but ROX-11623 introduces unit tests, robust error handling, and additional improvements like more file output options.

We don't want to announce the new capability just yet, as it is very limited in its first iteration.
An entry to the changelog and docs will follow once we're able to recommend Network Policies that adhere to all System Policies defined in a connected ACS Central. This is a follow-up feature that is scheduled for 3.73.

## Checklist
- [ ] ~~Unit test and regression tests added~~ -> Follow up in ROX-11774
- [ ] ~~Evaluated and added CHANGELOG entry if required~~ ->Skipping for now, until we have full online capability

## Testing Performed
To test the new functionality, point `roxctl generate netpol` to any folder containing any pod-creating YAML resources.
It will recursively scan all YAMLs and generate a list of recommended Network Policies to stdout.
```
git clone https://github.com/nadgowdas/microservices-demo /tmp/microservices-demo
./roxctl netpol generate /tmp/microservices-demo
```
Observe the recommended Network Policies being generated and printed to stdout.


Maddosaurus:
Rebasing to latest master to benefit from longer image retention on quay

roxctl/netpol/generate/command.go
```diff
@@ -0,0 +1,53 @@
...
+				return err
+			}
+			if err := netpolGenerateCmd.construct(args); err != nil {
+				return err
+			}
+			return netpolGenerateCmd.generateNetpol()
+		},
+	}
+
+	c.Flags().BoolVarP(&netpolGenerateCmd.offline, "offline", "", false, "whether to connect to a central instace for additional information")
```
fredrb: Why `BoolVarP` if you're not adding a shorthand? Either add a `o` shorthand or use `BoolVar`

roxctl/netpol/generate/command.go
```diff
@@ -0,0 +1,53 @@
...
+	// Properties that are bound to cobra flags.
+	offline    bool
+	folderPath string
+
+	// injected or constructed values
+	env     environment.Environment
+	printer printer.ObjectPrinter
+}
+
+// Command defines the central command tree
```
fredrb: Central?

roxctl/netpol/generate/generate.go
```diff
@@ -0,0 +1,23 @@
...
+
+func (cmd *netpolGenerateCommand) generateNetpol() error {
+	recommendedNetpols, err := controller.PoliciesFromFolderPath(cmd.folderPath)
+	if err != nil {
+		return errors.Errorf("Error synthesizing policies from folder: %v", err)
+	}
+
+	for _, netpol := range recommendedNetpols {
+		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
+		if err != nil {
```
fredrb: Question: could we `log` the error and skip adding it to the output? Or the plan is to abort the process entirely if a single NetworkPolicy is incorrect?

roxctl/netpol/generate/generate.go
```diff
@@ -0,0 +1,23 @@
...
+			return errors.Errorf("Error converting YAML into Network Policies: %v", err)
```
fredrb: Maybe singular? Since the error message will be about one policy failing.
```suggestion
			return errors.Errorf("Error converting YAML into Network Policy: %v", err)
```

roxctl/netpol/generate/generate.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/np-guard/cluster-topology-analyzer/pkg/controller"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/protoconv/networkpolicy"
+)
+
+func (cmd *netpolGenerateCommand) generateNetpol() error {
+	recommendedNetpols, err := controller.PoliciesFromFolderPath(cmd.folderPath)
+	if err != nil {
+		return errors.Errorf("Error synthesizing policies from folder: %v", err)
```
fredrb: You should wrap errors to preserve the context.
```suggestion
		return errors.Wrap(err, "error synthesizing policies from folder")
```

If you use `Errorf` you're creating a new error with a string that contains the other error message. By wrapping the errors you preserve the error chain which can be unwrapped one by one. Plus it allows for convenience functions like `error.Is` to check for the presence of an error in the chain.

roxctl/netpol/generate/generate.go
fredrb: Same thing about wrapping errors as above.

fredrb (changes requested):

roxctl/generate/netpol/netpol.go
```diff
@@ -0,0 +1,53 @@
...
+	// injected or constructed values
+	env     environment.Environment
+	printer printer.ObjectPrinter
+}
+
+// Command defines the central command tree
+func Command(cliEnvironment environment.Environment) *cobra.Command {
+	generateNetpolCmd := &generateNetpolCommand{env: cliEnvironment}
+	c := &cobra.Command{
+		Use: "netpol <folder-path>",
```
janisz, line 24:
```suggestion
		Use: "netpol <folder-path>",
		Args: cobra.ExactArgs(1),
```

roxctl/generate/netpol/netpol.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+			return generateNetpolCmd.generateNetpol()
+		},
+	}
+
+	c.Flags().BoolVarP(&generateNetpolCmd.offline, "offline", "", false, "whether to connect to a central instace for additional information")
+	return c
+}
+
+func validate(args []string) error {
```
janisz, line 40: This could be replaces with `Args:      cobra.ExactArgs(1),`

roxctl/generate/netpol/cta.go
```diff
@@ -0,0 +1,23 @@
+package netpol
+
+import (
+	"github.com/np-guard/cluster-topology-analyzer/pkg/controller"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/protoconv/networkpolicy"
+)
+
+func (cmd *generateNetpolCommand) generateNetpol() error {
```
janisz, line 9: Could you add a test?

Maddosaurus: Yes, absolutely. We will add tests in the follow up [ROX-11774](https://issues.redhat.com/browse/ROX-11774).
This branch started as a PoC and I treated the attached Epic as an opportunity to slice the work into multiple smaller issues.
This is an exception and won't be the norm, therefore I am asking for a one-time allowance to introduce tests in a second PR :)

roxctl/netpol/command.go
```diff
@@ -0,0 +1,19 @@
+package netpol
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stackrox/rox/roxctl/common/environment"
+	"github.com/stackrox/rox/roxctl/netpol/generate"
+)
+
+// Command defines the collector command tree
```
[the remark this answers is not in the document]
Maddosaurus: Textbook copy/paste error. I found another one one deeper 😄

roxctl/netpol/generate/command.go
Maddosaurus: I originally thought we might add a shorthand, but we can switch that over if we need it. Done!

roxctl/netpol/generate/command.go
Maddosaurus: Done!

roxctl/netpol/generate/generate.go
Maddosaurus: Right now, I lean towards breaking the whole process to not mask that we might be missing Network Policies.
If this is run by CI, no one might ever see the warning.

roxctl/netpol/generate/generate.go
Maddosaurus: Thanks for the explanation, changed to Wrap!

roxctl/generate/netpol/netpol.go
Maddosaurus: Thanks for the hint with `ExactArgs`!
For now, that is correct, but to increase robustness, there are plans to introduce more validation checks (e.g. checking if all YAMLs in the provided folder are valid) in a follow up ticket in the same Epic, so I'd lean towards keeping this for now. What do you think?

roxctl/netpol/generate/generate.go
fredrb: Got it

fredrb (approved):
Thanks for addressing the review comments.

vikin91 (approved):
Looks actually pretty good to me! I have run it locally and it works correctly, even the errors look reasonable out of the box.

lvalerom (approved):
LGTM 🚀  4min late 😢

vikin91:
Interesting run `bin/darwin/roxctl generate netpol /` 😂  - we need to think about a `--depth` parameter

Maddosaurus:
> LGTM 🚀 4min late 😢

@lvalerom Here, have a pupper for your troubles. It's at least something 😅

PR #2196, opened 2022-06-24
Dackbox sac tests

rhybrillou:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -0,0 +1,1239 @@
...
+			contextKey:     sacTestUtils.Cluster1NamespacesABReadWriteCtx,
+			expectedExists: true,
+		},
+		"Wrong cluster full read-write can NOT see single scope CVE": {
+			contextKey: sacTestUtils.Cluster2ReadWriteCtx,
+		},
+		"Wrong cluster partial read-write to the matching namespace can NOT see single scope CVE": {
+			contextKey: sacTestUtils.Cluster2NamespaceAReadWriteCtx,
+		},
+		"Wrong cluster partial read-write to any not-matching namespace can NOT see single scope CVE": {
```
dhaus67, line 314: Potentially room for one more test case:
`Wrong cluster partial read-write to at least the right namespace can NOT see single scope CVE`

central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -0,0 +1,1239 @@
...
+			contextKey: sacTestUtils.Cluster2NamespacesBCReadWriteCtx,
+		},
+	}
+	cve := "CVE-1234-0001"
+	cveId := s.getCVEID(cve)
+	cvss := float32(5.8)
```
dhaus67, line 511: Would also be better to use the value here provided from the fixture, in case we ever change it.
Same applies for few other places here.

central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -0,0 +1,1239 @@
...
+		})
+	}
+}
+
+func (s *cveDataStoreSACTestSuite) TestSACImageCVEExistsSharedAcrossComponents() {
+	// Inject the fixture graph, and test exists for CVE-4567-0002
+	s.dackboxTestStore.PushImageToVulnerabilitiesGraph()
+	defer s.cleanImageToVulnerabilitiesGraph()
+	cveId := s.getCVEID("CVE-4567-0002")
+	testCases := map[string]existsTestCase{
```
dhaus67, line 336: For those, do you think it would be worthwhile is to extend the `MixedClusterNamespace` test context, allowing us to also check when both are matching or its not strictly required?

central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -0,0 +1,1239 @@
...
+			contextKey: sacTestUtils.Cluster3ReadWriteCtx,
+		},
+	}
+	for name, c := range testCases {
+		s.Run(name, func() {
+			if features.PostgresDatastore.Enabled() {
+				s.T().Skip("Skipping CVE tests for postgres for now.")
+			}
+			testCtx := s.imageTestContexts[c.contextKey]
+			nodeCVEs, err := s.imageCVEStore.GetBatch(testCtx, cveIDs)
```
dhaus67, line 774: imageCVEs, right?

central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -0,0 +1,1239 @@
...
+
+func (s *cveDataStoreSACTestSuite) cleanImageToVulnerabilitiesGraph() {
+	s.Require().NoError(s.dackboxTestStore.CleanImageToVulnerabilitiesGraph())
+}
+
+func (s *cveDataStoreSACTestSuite) cleanNodeToVulnerabilitiesGraph() {
+	s.Require().NoError(s.dackboxTestStore.CleanNodeToVulnerabilitiesGraph())
+}
+
+type existsTestCase struct {
```
dhaus67, line 259: I think we have quite some duplicated code, specifically regarding the test case definitions.

Would it make sense to combine them based on the type we are testing for (where type == single scope CVE, non-shared, shared)?

What I'm suggestion would be looking like the following:
```go
type imageCVETestcase struct {
  contextKey string
  expectExists map[string]bool
  expectFound map[string]bool
  expectedCount map[string]int
}

func createTestCases() map[string]imageCVETestcase {
  return map[string]imageCVETestcase{
    "some test case description": {
       contextKey: <some-context-key>,
       expectExists: map[string]bool{
         "CVE-ID1": true,
         "CVE-ID2": false,
       },
       // and so on....
    }
  }
}

// Within the tests, usage would look like the following:
for name, c := range createTestCases {
		s.Run(name, func() {
			testCtx := s.imageTestContexts[c.contextKey]
			exists, err := s.imageCVEStore.Exists(testCtx, cveId)
			s.NoError(err)
                        // Using the CVE-ID that's used for the test.
			s.Equal(c.expectedExists[CVE-ID], exists)
		})
}
```
 It's a bit more complex to initially create the test cases, but at the very least we won't have them duplicated for the three different scopes (single, not shared, shared) for `Exists, Get, Count`. Would make sure we use the same test cases across the different scopes, potentially making maintenance of them also easier.

WDYT about making this adjustment?

central/cve/datastoretest/datastore_sac_test.go
rhybrillou: Good catch 👍

dhaus67 (approved):

PR #2197, opened 2022-06-24
Adding valid values to SecuredCluster CRD description

lbac-redhat:
Amending description of securedcluster.spec.admissonControl.bypass and securedcluster.spec.admissionControl.contactImageScanners keys with valid enum values for use with `kubectl explain securedcluster.spec.admissionControl`

## Description

The description of the admissionControl.bypass and admissionControl.contactImageScanners keys are missing valid values, so the only way of finding out what the valid values are is to read the code.

End users should be able to see and understand the valid values without reading the code, so this PR adds the valid values to the CRD description.

### DIFF
--- a/operator/apis/platform/v1alpha1/securedcluster_types.go
+++ b/operator/apis/platform/v1alpha1/securedcluster_types.go
@@ -114,6 +114,10 @@ type AdmissionControlComponentSpec struct {
 	ListenOnEvents *bool `json:"listenOnEvents,omitempty"`
 
 	// Should inline scanning be performed on previously unscanned images during a deployments admission review.
+    //
+    // Valid values are:
+    // - "DoNotScanInline" (default): to skip scanning the image when processing the admission request
+    // - "ScanIfMissing": if the scan results for the image are missing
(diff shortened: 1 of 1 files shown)


operator/apis/platform/v1alpha1/securedcluster_types.go
```diff
@@ -114,6 +114,10 @@ type AdmissionControlComponentSpec struct {
 	ListenOnEvents *bool `json:"listenOnEvents,omitempty"`
 
 	// Should inline scanning be performed on previously unscanned images during a deployments admission review.
+    //
```
lbac-redhat, line 117: Personally, as an end-user, I would actually prefer the description with proper newlines like this:
```
    // Should inline scanning be performed on previously unscanned images during a deployments admission review.
    //
    //
    // Valid values are:
    //
    // - "DoNotScanInline" (default): to skip scanning the image when processing the admission request
    //
    // - "ScanIfMissing": if the scan results for the image are missing
```

But I completely understand that it makes the actual code less readable. (This is where I really wish controller-gen/kubebuilder would make newlines easier or fix it in a better way.)

PR #2206, opened 2022-06-24
[ROX-11050] Add locks around image and node ops

md2119:
## Description

Add keyed locking mechanism around image and node upserts. Found more places where SAC needed to be guarded behind the feature flag. Also, I removed the sac filter stuff that I had added initially, will let Yann close that loop.

## Checklist
- ~Determined and documented upgrade steps

## Testing Performed
CI

### DIFF
--- a/central/image/datastore/datastore_impl.go
+++ b/central/image/datastore/datastore_impl.go
@@ -17,6 +17,7 @@ import (
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/pkg/errorhelpers"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/images/enricher"
 	imageTypes "github.com/stackrox/rox/pkg/images/types"
 	"github.com/stackrox/rox/pkg/logging"
@@ -125,16 +126,22 @@ func (ds *datastoreImpl) ListImage(ctx context.Context, sha string) (*storage.Li
 
 // CountImages delegates to the underlying store.
 func (ds *datastoreImpl) CountImages(ctx context.Context) (int, error) {
-	if ok, err := imagesSAC.ReadAllowed(ctx); err != nil {
-		return 0, err
-	} else if ok {
-		return ds.storage.Count(ctx)
+	if !features.PostgresDatastore.Enabled() {
+		if ok, err := imagesSAC.ReadAllowed(ctx); err != nil {
+			return 0, err
+		} else if ok {
+			return ds.storage.Count(ctx)
+		}
 	}
 
 	return ds.Count(ctx, pkgSearch.EmptyQuery())
 }
 
 func (ds *datastoreImpl) canReadImage(ctx context.Context, sha string) (bool, error) {
+	if features.PostgresDatastore.Enabled() {
+		return true, nil
+	}
+
 	if ok, err := imagesSAC.ReadAllowed(ctx); err != nil {
 		return false, err
 	} else if ok {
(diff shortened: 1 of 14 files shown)


md2119:
image

central/image/datastore/datastore_impl.go
```diff
@@ -125,16 +126,22 @@ func (ds *datastoreImpl) ListImage(ctx context.Context, sha string) (*storage.Li
 
 // CountImages delegates to the underlying store.
 func (ds *datastoreImpl) CountImages(ctx context.Context) (int, error) {
-	if ok, err := imagesSAC.ReadAllowed(ctx); err != nil {
-		return 0, err
-	} else if ok {
-		return ds.storage.Count(ctx)
+	if !features.PostgresDatastore.Enabled() {
```
connorgorman, line 129: We shouldn't do this, instead we should fix the images store to do what all the other stores do for sac scopes

central/image/datastore/datastore_impl.go
```diff
@@ -183,9 +190,15 @@ func (ds *datastoreImpl) GetImage(ctx context.Context, sha string) (*storage.Ima
 // GetImagesBatch delegates to the underlying store.
 func (ds *datastoreImpl) GetImagesBatch(ctx context.Context, shas []string) ([]*storage.Image, error) {
 	var imgs []*storage.Image
-	if ok, err := imagesSAC.ReadAllowed(ctx); err != nil {
-		return nil, err
-	} else if ok {
+	var err error
+	ok := true
+	if !features.PostgresDatastore.Enabled() {
+		if ok, err = imagesSAC.ReadAllowed(ctx); err != nil {
```
connorgorman, line 196: Not sure I see the need to remove these for now, these are almost free checks

central/image/datastore/store/postgres/store.go
```diff
@@ -857,35 +862,33 @@ func (s *storeImpl) Delete(ctx context.Context, id string) error {
...
 	// Delete orphaned image components.
-	if _, err := conn.Exec(ctx, "delete from "+imageComponentsTable+" where id = ANY($1::text[])", componentIDs); err != nil {
+	_, err := tx.Exec(ctx, "delete from "+imageComponentsTable+" where not exists (select "+imageComponentsTable+".id FROM "+imageComponentsTable+", "+imageComponentEdgesTable+" WHERE "+imageComponentsTable+".id = "+imageComponentEdgesTable+".imagecomponentid)")
+	if err != nil {
 		return err
 	}
+
 	// Delete orphaned cves.
-	if _, err := conn.Exec(ctx, "delete from "+imageCVEsTable+" where not exists (select "+imageCVEsTable+".id FROM "+imageCVEsTable+", "+componentCVEEdgesTable+" WHERE "+imageCVEsTable+".id = "+componentCVEEdgesTable+".imagecveid)"); err != nil {
+	if _, err := tx.Exec(ctx, "delete from "+imageCVEsTable+" where not exists (select "+imageCVEsTable+".id FROM "+imageCVEsTable+", "+componentCVEEdgesTable+" WHERE "+imageCVEsTable+".id = "+componentCVEEdgesTable+".imagecveid)"); err != nil {
```
connorgorman, line 891: This one confuses me, shouldn't this just be 

`delete from image_cves where not exists (select imagecveid from imagecomponentcveedges)`

central/image/datastore/datastore_impl.go
viswajithiii: Still seems cleaner to remove them. More than the performance, it's about readability -- someone in the future will see this and be confused as to whether they are necessary or not

viswajithiii: @md2119 I think this might be better accomplished by making `imagesSAC` something that always returns `true, nil` if the postgres flag is enabled, rather than branching everywhere

PR #2211, opened 2022-06-24
feat: adding feature flag for frontend vm updates

sachaudh:
## Description

Added a new feature flag for the frontend-only VM updates

## Checklist


.circleci/config.yml
```diff
@@ -3355,6 +3355,7 @@ jobs:
       - ROX_POLICIES_PATTERNFLY: true
       - ROX_NEW_POLICY_CATEGORIES: true
       - ROX_SECURITY_METRICS_PHASE_ONE: true
+      - ROX_FRONTEND_VM_UDPATES: true
```
vjwilson, line 3358: @gavin-stackrox 

How are we turning on feature flags for dev CI in the OpenShift world?

vjwilson (approved):
/lgtm

One question for Gavin.

.circleci/config.yml
gavin-stackrox: To set a flag for all e2e tests: https://github.com/stackrox/stackrox/blob/99a8eb435bcf31ba471e4e47242d7b868f11697f/tests/e2e/lib.sh#L32

To override one of those or for a setting that is just for UI e2e tests:
https://github.com/stackrox/stackrox/blob/99a8eb435bcf31ba471e4e47242d7b868f11697f/tests/e2e/run-ui-e2e.sh#L20

(and I just merged #2217 which decommissioned Circle CI UI e2e tests 🚢 🚢 )

sachaudh: Cool cool. I undid my change in the `config.yml` and made the necessary change in `lib.sh`

PR #2214, opened 2022-06-25
[ROX-11258] : Add node vulnerability resolver for postgres and wire it in cluster and node graphQL; Update PlottedNodeVulnerabilities resolvers to use Postgres

charmik-redhat:
## Description

- Added NodeVulnerabilitiesResolver implementation that uses Postgres
- Wired up node vulnerability sub resolvers in cluster and node to use the new implementation when Postgres is enabled
- Added NodeCVELoader
- Updated plotted node vulnerabilities resolvers to use NodeCVEDatastore when Postgres is enabled 

## Checklist
- ~[ ] Determined and documented upgrade steps

## Testing Performed

- Manually tested by querying the resolvers to see if they return node vulnerabilities (and no other vulns) and that they contain expected fields, when Postgres is enabled.
- e2e/unit tests will be added in a separate task.

### DIFF
--- a/central/graphql/resolvers/clusters.go
+++ b/central/graphql/resolvers/clusters.go
@@ -649,8 +649,9 @@ func (resolver *clusterResolver) NodeVulnerabilityCount(ctx context.Context, arg
 func (resolver *clusterResolver) NodeVulnerabilityCounter(ctx context.Context, args RawQuery) (*VulnerabilityCounterResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "NodeVulnerabilityCounter")
 
+	// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
 	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-	return resolver.root.NodeVulnCounter(ctx, RawQuery{Query: &query})
+	return resolver.root.NodeVulnerabilityCounter(ctx, RawQuery{Query: &query})
 }
 
 func (resolver *clusterResolver) withClusterScope(ctx context.Context) context.Context {
@@ -1117,6 +1118,7 @@ func (resolver *clusterResolver) LatestViolation(ctx context.Context, args RawQu
 func (resolver *clusterResolver) PlottedNodeVulnerabilities(ctx context.Context, args RawQuery) (*PlottedNodeVulnerabilitiesResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "PlottedNodeVulnerabilities")
 
+	// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
 	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
 	return newPlottedNodeVulnerabilitiesResolver(ctx, resolver.root, RawQuery{Query: &query})
 }
--- a/central/graphql/resolvers/clusters.go
+++ b/central/graphql/resolvers/clusters.go
@@ -1117,12 +1117,8 @@ func (resolver *clusterResolver) LatestViolation(ctx context.Context, args RawQu
 func (resolver *clusterResolver) PlottedNodeVulnerabilities(ctx context.Context, args RawQuery) (*PlottedNodeVulnerabilitiesResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "PlottedNodeVulnerabilities")
 
-	if !features.PostgresDatastore.Enabled() {
-		query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-		return newPlottedNodeVulnerabilitiesResolver(ctx, resolver.root, RawQuery{Query: &query})
-	}
-	// TODO : Add postgres support
-	return nil, errors.New("Sub-resolver PlottedNodeVulnerabilities in Cluster does not support postgres yet")
+	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
+	return newPlottedNodeVulnerabilitiesResolver(ctx, resolver.root, RawQuery{Query: &query})
 }
 
 // PlottedImageVulnerabilities returns the data required by top risky entity scatter-plot on vuln mgmt dashboard
--- a/central/graphql/resolvers/clusters.go
+++ b/central/graphql/resolvers/clusters.go
@@ -631,38 +631,26 @@ func (resolver *clusterResolver) VulnCounter(ctx context.Context, args RawQuery)
 func (resolver *clusterResolver) NodeVulnerabilities(ctx context.Context, args PaginatedQuery) ([]NodeVulnerabilityResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "NodeVulnerabilities")
 
-	if !features.PostgresDatastore.Enabled() {
-		// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
-		query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-		return resolver.root.NodeVulnerabilities(ctx, PaginatedQuery{Query: &query, Pagination: args.Pagination})
-	}
-	// TODO : Add postgres support
-	return nil, errors.New("Sub-resolver NodeVulnerabilities in Cluster does not support postgres yet")
+	// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
+	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
+	return resolver.root.NodeVulnerabilities(ctx, PaginatedQuery{Query: &query, Pagination: args.Pagination})
 }
 
 // NodeVulnerabilityCount returns the number of node vulnerabilities in the cluster.
 func (resolver *clusterResolver) NodeVulnerabilityCount(ctx context.Context, args RawQuery) (int32, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "NodeVulnerabilityCount")
 
-	if !features.PostgresDatastore.Enabled() {
-		// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
-		query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-		return resolver.root.NodeVulnerabilityCount(ctx, RawQuery{Query: &query})
-	}
-	// TODO : Add postgres support
-	return 0, errors.New("Sub-resolver NodeVulnerabilityCount in Cluster does not support postgres yet")
+	// (ROX-10911) Cluster scoping the context is not able to resolve node vulns when combined with 'Fixable:true/false' query
+	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
+	return resolver.root.NodeVulnerabilityCount(ctx, RawQuery{Query: &query})
 }
 
 // NodeVulnerabilityCounter resolves the number of different types of node vulnerabilities contained in the cluster.
 func (resolver *clusterResolver) NodeVulnerabilityCounter(ctx context.Context, args RawQuery) (*VulnerabilityCounterResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "NodeVulnerabilityCounter")
 
-	if !features.PostgresDatastore.Enabled() {
-		query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-		return resolver.root.NodeVulnCounter(ctx, RawQuery{Query: &query})
-	}
-	// TODO : Add postgres support
-	return nil, errors.New("Sub-resolver NodeVulnerabilityCounter in Cluster does not support postgres yet")
+	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
+	return resolver.root.NodeVulnCounter(ctx, RawQuery{Query: &query})
 }
 
 func (resolver *clusterResolver) withClusterScope(ctx context.Context) context.Context {
--- a/central/graphql/resolvers/common_vulnerabilities.go
+++ b/central/graphql/resolvers/common_vulnerabilities.go
@@ -33,7 +33,6 @@ var commonVulnerabilitySubResolvers = []string{
 	"suppressed: Boolean!",
 	"unusedVarSink(query: String): Int",
 	"vectors: EmbeddedVulnerabilityVectors",
-	"vulnerabilityState: String!",
 }
 
 // CommonVulnerabilityResolver represents the supported API on all vulnerabilities
@@ -59,5 +58,4 @@ type CommonVulnerabilityResolver interface {
 	Suppressed(ctx context.Context) bool
 	UnusedVarSink(ctx context.Context, args RawQuery) *int32
 	Vectors() *EmbeddedVulnerabilityVectorsResolver
-	VulnerabilityState(ctx context.Context) string
 }
--- a/central/graphql/resolvers/loaders/images.go
+++ b/central/graphql/resolvers/loaders/images.go
@@ -135,11 +135,3 @@ func (idl *imageLoaderImpl) readAll(ids []string) (images []*storage.Image, miss
 	}
 	return
 }
-
-func collectMissing(ids []string, missing []int) []string {
-	missingIds := make([]string, 0, len(missing))
-	for _, missingIdx := range missing {
-		missingIds = append(missingIds, ids[missingIdx])
-	}
-	return missingIds
-}
--- a/central/graphql/resolvers/loaders/node_cves.go
+++ b/central/graphql/resolvers/loaders/node_cves.go
@@ -9,7 +9,6 @@ import (
 	nodeCVEDataStore "github.com/stackrox/rox/central/cve/node/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
-	"github.com/stackrox/rox/pkg/cvss"
 	"github.com/stackrox/rox/pkg/search"
 	"github.com/stackrox/rox/pkg/sync"
 )
@@ -56,10 +55,6 @@ type nodeCVELoaderImpl struct {
 
 	ds nodeCVEDataStore.DataStore
 }
-
-//func enrichNodeCVESeverity(value *storage.NodeCVE) {
-//	value.Severity = cvss.VulnToSeverity(cvss.NewFromNodeCVE(value))
-//}
 
 // FromIDs loads a set of nodeCVEs from a set of ids.
 func (idl *nodeCVELoaderImpl) FromIDs(ctx context.Context, ids []string) ([]*storage.NodeCVE, error) {
@@ -109,9 +104,6 @@ func (idl *nodeCVELoaderImpl) load(ctx context.Context, ids []string) ([]*storag
 		if err != nil {
 			return nil, err
 		}
-		for _, cve := range nodeCves {
-			cve.Severity = cvss.VulnToSeverity(cvss.NewFromNodeCVE(cve))
-		}
 		idl.setAll(nodeCves)
 		nodeCves, missing = idl.readAll(ids)
 	}
@@ -120,7 +112,7 @@ func (idl *nodeCVELoaderImpl) load(ctx context.Context, ids []string) ([]*storag
 		for _, m := range missing {
 			missingIDs = append(missingIDs, ids[m])
 		}
-		return nil, errors.Errorf("not all cves could be found: %s", strings.Join(missingIDs, ","))
+		return nil, errors.Errorf("not all node cves could be found: %s", strings.Join(missingIDs, ","))
 	}
 	return nodeCves, nil
 }
--- a/central/graphql/resolvers/loaders/node_cves.go
+++ b/central/graphql/resolvers/loaders/node_cves.go
@@ -9,6 +9,7 @@ import (
 	nodeCVEDataStore "github.com/stackrox/rox/central/cve/node/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/search"
 	"github.com/stackrox/rox/pkg/sync"
 )
@@ -16,9 +17,11 @@ import (
 var nodeCveLoaderType = reflect.TypeOf(storage.NodeCVE{})
 
 func init() {
-	RegisterTypeFactory(reflect.TypeOf(storage.NodeCVE{}), func() interface{} {
-		return NewNodeCVELoader(nodeCVEDataStore.Singleton())
-	})
+	if features.PostgresDatastore.Enabled() {
+		RegisterTypeFactory(reflect.TypeOf(storage.NodeCVE{}), func() interface{} {
+			return NewNodeCVELoader(nodeCVEDataStore.Singleton())
+		})
+	}
 }
 
 // NewNodeCVELoader creates a new loader for nodeCVE data.
--- a/central/graphql/resolvers/loaders/node_cves.go
+++ b/central/graphql/resolvers/loaders/node_cves.go
@@ -1,0 +1,150 @@
+package loaders
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	"github.com/pkg/errors"
+	nodeCVEDataStore "github.com/stackrox/rox/central/cve/node/datastore"
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/cvss"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var nodeCveLoaderType = reflect.TypeOf(storage.NodeCVE{})
+
+func init() {
+	RegisterTypeFactory(reflect.TypeOf(storage.NodeCVE{}), func() interface{} {
+		return NewNodeCVELoader(nodeCVEDataStore.Singleton())
+	})
+}
+
+// NewNodeCVELoader creates a new loader for nodeCVE data.
+func NewNodeCVELoader(ds nodeCVEDataStore.DataStore) NodeCVELoader {
+	return &nodeCVELoaderImpl{
+		loaded: make(map[string]*storage.NodeCVE),
+		ds:     ds,
+	}
+}
+
+// GetNodeCVELoader returns the NodeCVELoader from the context if it exists.
+func GetNodeCVELoader(ctx context.Context) (NodeCVELoader, error) {
+	loader, err := GetLoader(ctx, nodeCveLoaderType)
+	if err != nil {
+		return nil, err
+	}
+	return loader.(NodeCVELoader), nil
+}
+
+// NodeCVELoader loads nodeCVE data, and stores already loaded nodeCVEs for other ops in the same context to use.
+type NodeCVELoader interface {
+	FromIDs(ctx context.Context, ids []string) ([]*storage.NodeCVE, error)
+	FromID(ctx context.Context, id string) (*storage.NodeCVE, error)
+	FromQuery(ctx context.Context, query *v1.Query) ([]*storage.NodeCVE, error)
+
+	CountFromQuery(ctx context.Context, query *v1.Query) (int32, error)
+	CountAll(ctx context.Context) (int32, error)
+}
+
+// nodeCVELoaderImpl implements the NodeCVELoader interface.
+type nodeCVELoaderImpl struct {
+	lock   sync.RWMutex
+	loaded map[string]*storage.NodeCVE
+
+	ds nodeCVEDataStore.DataStore
+}
+
+//func enrichNodeCVESeverity(value *storage.NodeCVE) {
+//	value.Severity = cvss.VulnToSeverity(cvss.NewFromNodeCVE(value))
+//}
+
+// FromIDs loads a set of nodeCVEs from a set of ids.
+func (idl *nodeCVELoaderImpl) FromIDs(ctx context.Context, ids []string) ([]*storage.NodeCVE, error) {
+	nodeCves, err := idl.load(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+	return nodeCves, nil
+}
+
+// FromID loads a nodeCVE from an ID.
+func (idl *nodeCVELoaderImpl) FromID(ctx context.Context, id string) (*storage.NodeCVE, error) {
+	nodeCves, err := idl.load(ctx, []string{id})
+	if err != nil {
+		return nil, err
+	}
+	return nodeCves[0], nil
+}
+
+// FromQuery loads a set of nodeCVEs that match a query.
+func (idl *nodeCVELoaderImpl) FromQuery(ctx context.Context, query *v1.Query) ([]*storage.NodeCVE, error) {
+	results, err := idl.ds.Search(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	return idl.FromIDs(ctx, search.ResultsToIDs(results))
+}
+
+func (idl *nodeCVELoaderImpl) CountFromQuery(ctx context.Context, query *v1.Query) (int32, error) {
+	count, err := idl.ds.Count(ctx, query)
+	if err != nil {
+		return 0, err
+	}
+	return int32(count), nil
+}
+
+func (idl *nodeCVELoaderImpl) CountAll(ctx context.Context) (int32, error) {
+	count, err := idl.ds.Count(ctx, search.EmptyQuery())
+	return int32(count), err
+}
+
+func (idl *nodeCVELoaderImpl) load(ctx context.Context, ids []string) ([]*storage.NodeCVE, error) {
+	nodeCves, missing := idl.readAll(ids)
+	if len(missing) > 0 {
+		var err error
+		nodeCves, err = idl.ds.GetBatch(ctx, collectMissing(ids, missing))
+		if err != nil {
+			return nil, err
+		}
+		for _, cve := range nodeCves {
+			cve.Severity = cvss.VulnToSeverity(cvss.NewFromNodeCVE(cve))
+		}
+		idl.setAll(nodeCves)
+		nodeCves, missing = idl.readAll(ids)
+	}
+	if len(missing) > 0 {
+		missingIDs := make([]string, 0, len(missing))
+		for _, m := range missing {
+			missingIDs = append(missingIDs, ids[m])
+		}
+		return nil, errors.Errorf("not all cves could be found: %s", strings.Join(missingIDs, ","))
+	}
+	return nodeCves, nil
+}
+
+func (idl *nodeCVELoaderImpl) setAll(nodeCves []*storage.NodeCVE) {
+	idl.lock.Lock()
+	defer idl.lock.Unlock()
+
+	for _, cve := range nodeCves {
+		idl.loaded[cve.GetId()] = cve
+	}
+}
+
+func (idl *nodeCVELoaderImpl) readAll(ids []string) (nodeCves []*storage.NodeCVE, missing []int) {
+	idl.lock.RLock()
+	defer idl.lock.RUnlock()
+
+	for idx, id := range ids {
+		cve, isLoaded := idl.loaded[id]
+		if !isLoaded {
+			missing = append(missing, idx)
+		} else {
+			nodeCves = append(nodeCves, cve)
+		}
+	}
+	return
+}
--- a/central/graphql/resolvers/node_vulnerabilities.go
+++ b/central/graphql/resolvers/node_vulnerabilities.go
@@ -56,12 +56,14 @@ func (resolver *Resolver) NodeVulnerability(ctx context.Context, args IDQuery) (
 	if err := readNodes(ctx); err != nil {
 		return nil, err
 	}
-	vuln, exists, err := resolver.NodeCVEDataStore.Get(ctx, string(*args.ID))
+
+	vulnLoader, err := loaders.GetNodeCVELoader(ctx)
 	if err != nil {
 		return nil, err
 	}
-	if !exists {
-		return nil, errors.Errorf("node cve not found: %s", string(*args.ID))
+	vuln, err := vulnLoader.FromID(ctx, string(*args.ID))
+	if err != nil {
+		return nil, err
 	}
 	vulnResolver, err := resolver.wrapNodeCVE(vuln, true, nil)
 	if err != nil {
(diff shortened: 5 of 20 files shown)


central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+
+// ID of the node CVE
+func (resolver *nodeCVEResolver) ID(ctx context.Context) graphql.ID {
+	value := resolver.data.GetId()
+	return graphql.ID(value)
+}
+
+// IsFixable returns whether node CVE is fixable by any component
+func (resolver *nodeCVEResolver) IsFixable(ctx context.Context, args RawQuery) (bool, error) {
+	// TODO : Why do we remove this field query and then add it back again in addScopeContextOrNodeCVEQuery ?
```
charmik-redhat, line 270: @md2119

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return &EmbeddedVulnerabilityVectorsResolver{
+			resolver: &cVSSV2Resolver{resolver.ctx, resolver.root, val},
+		}
+	}
+	return nil
+}
+
+// VulnerabilityState returns the effective state of the node CVE (observed, deferred or marked as false positive).
+func (resolver *nodeCVEResolver) VulnerabilityState(ctx context.Context) string {
+	//TODO Should this be removed from nodeVulnerabilities graphQL ?
```
charmik-redhat, line 419: @md2119

central/graphql/resolvers/loaders/images.go
```diff
@@ -135,11 +135,3 @@ func (idl *imageLoaderImpl) readAll(ids []string) (images []*storage.Image, miss
...
 	return
 }
-
-func collectMissing(ids []string, missing []int) []string {
-	missingIds := make([]string, 0, len(missing))
-	for _, missingIdx := range missing {
-		missingIds = append(missingIds, ids[missingIdx])
-	}
-	return missingIds
-}
```
keyallis, line 145: +1 on the utils file

central/graphql/resolvers/loaders/node_cves.go
```diff
@@ -0,0 +1,150 @@
...
+type nodeCVELoaderImpl struct {
+	lock   sync.RWMutex
+	loaded map[string]*storage.NodeCVE
+
+	ds nodeCVEDataStore.DataStore
+}
+
+//func enrichNodeCVESeverity(value *storage.NodeCVE) {
+//	value.Severity = cvss.VulnToSeverity(cvss.NewFromNodeCVE(value))
+//}
```
keyallis, line 62: Looks like it should be cleaned up

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+	}
+
+	unFixableVulnsQuery := search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, false).ProtoQuery())
+	unFixableCVEs, err := vulnLoader.FromQuery(ctx, unFixableVulnsQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	return mapNodeCVEsToVulnerabilityCounter(fixableVulns, unFixableCVEs), nil
+}
```
keyallis, line 190: It feels like the logic of this function could be collapsed into logic higher up, similar to previous comment

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+	nodeLoader, err := loaders.GetNodeLoader(ctx)
+	if err != nil {
+		return 0, 0, err
+	}
+	withThisCVECount, err := nodeLoader.CountFromQuery(resolver.withNodeVulnerabilityScope(ctx), search.EmptyQuery())
+	if err != nil {
+		return 0, 0, err
+	}
+	return int(withThisCVECount), allNodesCount, nil
+}
```
keyallis, line 232: Logic could be pushed up here as well

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return 0, err
+	}
+	if d == 0 {
+		return 0, nil
+	}
+	return float64(n) / float64(d), nil
+}
+
+func (resolver *nodeCVEResolver) getEnvImpactComponentsForNodeCVE(ctx context.Context) (numerator, denominator int, err error) {
+	allNodesCount, err := resolver.root.NodeGlobalDataStore.CountAllNodes(ctx)
```
keyallis, line 216: We can query the all count through the loader here, not sure if that's really necessary anymore with postgres since a count operation should be relatively lightweight, but worth considering

central/graphql/resolvers/node_vulnerabilities.go
keyallis: This is I think because if this subresolver is invoked with scoping from a component/node/cluster, we define the `IsFixable` as if it is fixable in any context, not just the context that it's being queried in

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+}
+
+func (resolver *nodeCVEResolver) getNodeCVERawQuery() string {
+	return search.NewQueryBuilder().AddExactMatches(search.CVE, resolver.data.GetId()).Query()
+}
+
+// LastModified is the time this node CVE was last modified in the system
+func (resolver *nodeCVEResolver) LastModified(ctx context.Context) (*graphql.Time, error) {
+	value := resolver.data.GetCveBaseInfo().GetLastModified()
+	return timestamp(value)
```
keyallis, line 326: I noticed this in a couple of other places as well, but I don't think you need to cast here, you should be able to just return the type you get. Also I think we need to check whether the value returned from the base info function is null before we call on it

central/graphql/resolvers/node_vulnerabilities.go
keyallis: IIRC supression is only for image vulns, I didn't think node vulns would have snooze functions either but they ended up there after data model transition

central/graphql/resolvers/vulnerability_counter.go
```diff
@@ -123,6 +123,25 @@ func mapCVEsToVulnerabilityCounter(fixable, unFixable []*storage.CVE) *Vulnerabi
 	return counter
 }
 
+func mapNodeCVEsToVulnerabilityCounter(fixable, unFixable []*storage.NodeCVE) *VulnerabilityCounterResolver {
```
keyallis, line 126: I think this is fine for now, but I'm pretty sure with postgres there's some way of being able to do this query easily with less overhead, something to look at in the future

central/graphql/resolvers/loaders/node_cves.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (idl *nodeCVELoaderImpl) load(ctx context.Context, ids []string) ([]*storage.NodeCVE, error) {
+	nodeCves, missing := idl.readAll(ids)
+	if len(missing) > 0 {
+		var err error
+		nodeCves, err = idl.ds.GetBatch(ctx, collectMissing(ids, missing))
+		if err != nil {
+			return nil, err
+		}
+		for _, cve := range nodeCves {
```
md2119, line 112: This is not required in Postgres because each instance of node vulns is unique to only one distro and therefore the stored top-level cvss and severity fields hold the final required values.

central/graphql/resolvers/loaders/node_cves.go
```diff
@@ -0,0 +1,150 @@
...
+
+// nodeCVELoaderImpl implements the NodeCVELoader interface.
+type nodeCVELoaderImpl struct {
+	lock   sync.RWMutex
+	loaded map[string]*storage.NodeCVE
+
+	ds nodeCVEDataStore.DataStore
+}
+
+//func enrichNodeCVESeverity(value *storage.NodeCVE) {
```
md2119, line 60: remove

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -47,8 +52,7 @@ func (resolver *Resolver) NodeVulnerability(ctx context.Context, args IDQuery) (
 	if !features.PostgresDatastore.Enabled() {
 		return resolver.nodeVulnerabilityV2(ctx, args)
 	}
-	// TODO : Add postgres support
-	return nil, errors.New("Resolver NodeVulnerability does not support postgres yet")
+	return resolver.nodeCveV2(ctx, args)
```
md2119, line 55: This indirection makes it readable. We do not expect logic in `nodeCveV2` to be accessible from anywhere else except this resolver. Therefore, no need to add a func.

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -58,8 +62,7 @@ func (resolver *Resolver) NodeVulnerabilities(ctx context.Context, q PaginatedQu
 		query := withNodeCveTypeFiltering(q.String())
 		return resolver.nodeVulnerabilitiesV2(ctx, PaginatedQuery{Query: &query, Pagination: q.Pagination})
 	}
-	// TODO : Add postgres support
-	return nil, errors.New("Resolver NodeVulnerabilities does not support postgres yet")
+	return resolver.nodeCvesV2(ctx, q)
```
md2119, line 65: same as above  for all such instances in the new resolver

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
 		query := withNodeCveTypeFiltering(args.String())
 		return resolver.vulnCounterV2(ctx, RawQuery{Query: &query})
 	}
-	// TODO : Add postgres support
-	return nil, errors.New("Resolver NodeVulnCounter does not support postgres yet")
+	return resolver.nodeCveCounterV2(ctx, args)
+}
+
+func (resolver *Resolver) nodeCveV2(ctx context.Context, args IDQuery) (NodeVulnerabilityResolver, error) {
+	if err := readCVEs(ctx); err != nil {
```
md2119, line 89: Check node permission. All parts of nodes should rely on users having permission to node (similarly for images @keyallis)

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+	vulnResolver, err := resolver.wrapNodeCVE(vuln, true, nil)
+	if err != nil {
+		return nil, err
+	}
+	vulnResolver.ctx = ctx
+	return vulnResolver, nil
+}
+
+func (resolver *Resolver) nodeCvesV2(ctx context.Context, args PaginatedQuery) ([]NodeVulnerabilityResolver, error) {
+	if err := readCVEs(ctx); err != nil {
```
md2119, line 108: same as above

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return 0, 0, err
+	}
+	if allNodesCount == 0 {
+		return 0, 0, nil
+	}
+	nodeLoader, err := loaders.GetNodeLoader(ctx)
+	if err != nil {
+		return 0, 0, err
+	}
+	withThisCVECount, err := nodeLoader.CountFromQuery(resolver.withNodeVulnerabilityScope(ctx), search.EmptyQuery())
```
md2119, line 227: It is more readable with `CountFromQuery` receiving node query instead of scoped ctx. Furthermore, this is not a redirect to another resolver.

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return "", nil
+	}
+	if scope.Level != v1.SearchCategory_NODE_COMPONENTS {
+		return "", nil
+	}
+	edgeID := edges.EdgeID{ParentID: scope.ID, ChildID: resolver.data.GetId()}.ToString()
+	edge, found, err := resolver.root.NodeComponentCVEEdgeDataStore.Get(resolver.ctx, edgeID)
+	if err != nil || !found {
+		return "", err
+	}
```
md2119, line 258: `FixedByVersion` is a sub-resolver of `nodeCVEResolver`. That is, we only expect calls like,
```
nodeVulnerabilities {
  fixedByVersion
}
```
On `ln251` we ensure that ctx should contain component scope. Therefore, when node vulnerability already pulled out of db, it already has the fixedby field value correctly scoped, and hence it can it returned here.

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+	ctx := resolver.ctx
+	scope, ok := scoped.GetScope(ctx)
+	if !ok {
+		return resolver.withNodeVulnerabilityScope(ctx), query
+	}
+	// If the scope is not set to node vulnerabilities then
+	// we need to add a query to scope the search to the current nodeCVE
+	if scope.Level != v1.SearchCategory_NODE_VULNERABILITIES {
+		return ctx, search.AddRawQueriesAsConjunction(query, resolver.getNodeCVERawQuery())
+	}
```
md2119, line 310: similar to above.

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+
+// NodeComponents are the node components that contain the node CVE.
+func (resolver *nodeCVEResolver) NodeComponents(ctx context.Context, args PaginatedQuery) ([]NodeComponentResolver, error) {
+	//TODO implement me (ROX-11299)
+	panic("implement me")
+}
+
+// NodeCount is the number of nodes that contain the node CVE
+func (resolver *nodeCVEResolver) NodeCount(ctx context.Context, args RawQuery) (int32, error) {
+	ctx, query := resolver.addScopeContextOrNodeCVERawQuery(args.String())
```
md2119, line 437: No need to update the scope. Pass in conjunction containing node cve id.

central/graphql/resolvers/node_vulnerabilities.go
charmik-redhat: Yes the indirection would make it readable from outside. I had abstracted that code out to avoid function from getting too long.

central/graphql/resolvers/node_vulnerabilities.go
charmik-redhat: So we do not need to check for `found` ?

central/graphql/resolvers/node_vulnerabilities.go
md2119: sorry...I really meant `less readable`. 😬

central/graphql/resolvers/node_vulnerabilities.go
md2119: Yes, node components and node vulns should use node permission.

central/graphql/resolvers/node_vulnerabilities.go
md2119: If `scope.Level == v1.SearchCategory_NODE_COMPONENTS` we know that component scope was incorporated when node vulnerability was fetched from the store (and the scope is an inner join— `select * from node_cves inner join componentcveedge on node_cves.id == componentcveedge.nodecveid ....`. The node vuln, with the component scope, is fetched only because the edge exists. Hence, the `found` is known to be true when you are here.

if `scope.Level != v1.SearchCategory_NODE_COMPONENTS`, the code returns empty string.

central/graphql/resolvers/node_vulnerabilities.go
charmik-redhat: Done

central/graphql/resolvers/loaders/node_cves.go
charmik-redhat: Good catch, removed it

central/graphql/resolvers/node_vulnerabilities.go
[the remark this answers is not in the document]
charmik-redhat: Scoping wasn't needed at the usages of this function, so removed it and replaced the calls with nodeCVE conjunction queries.

central/graphql/resolvers/vulnerability_counter.go
charmik-redhat: +1

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +82,366 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return false, err
+	}
+	return count != 0, nil
+}
+
+func (resolver *nodeCVEResolver) addScopeContextOrNodeCVEQuery(query *v1.Query) (context.Context, *v1.Query) {
+	ctx := resolver.ctx
+	scope, ok := scoped.GetScope(ctx)
+	if !ok {
+		return resolver.withNodeVulnerabilityScope(ctx), query
```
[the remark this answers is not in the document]
charmik-redhat: yeah, removed scoping and moved conjunction query logic to `IsFixable`

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +141,239 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+	}
+	if scope.Level != v1.SearchCategory_NODE_COMPONENTS {
+		return "", nil
+	}
+	edgeID := edges.EdgeID{ParentID: scope.ID, ChildID: resolver.data.GetId()}.ToString()
+	edge, _, err := resolver.root.NodeComponentCVEEdgeDataStore.Get(resolver.ctx, edgeID)
+	if err != nil {
+		return "", err
+	}
+	return edge.GetFixedBy(), nil
```
md2119, line 216:
```suggestion
	resolver.data.GetFixedBy(), nil
```

central/graphql/resolvers/plotted_vulnerabilities_common.go
```diff
@@ -8,34 +8,42 @@ import (
 )
 
 func getPlottedVulnsIdsAndFixableCount(ctx context.Context, root *Resolver, args RawQuery) ([]string, int, error) {
```
md2119, line 10:
```
  return nil, errors.New("Unexpected access to legacy CVE datastore")
}
```

central/graphql/resolvers/vulnerability_counter.go
md2119: +1 Connor

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +141,239 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return false, err
+	}
+
+	scope, ok := scoped.GetScope(resolver.ctx)
+	if !ok || scope.Level != v1.SearchCategory_NODE_VULNERABILITIES {
+		query = search.ConjunctionQuery(query, resolver.getNodeCVEQuery())
+	}
+
+	query = search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery())
+	count, err := resolver.root.NodeComponentCVEEdgeDataStore.Count(ctx, query)
```
md2119, line 238: You can fire the query on node CVE datastore since the result of inner join would be the same. That way we can remove dependency from nodecomponent-cve edge datastore in graphQL.

md2119 (commented):
couple of comments, rest lgtm

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +141,239 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+// FixedByVersion returns the version of the parent component that removes this CVE
+func (resolver *nodeCVEResolver) FixedByVersion(_ context.Context) (string, error) {
+	scope, hasScope := scoped.GetScope(resolver.ctx)
+	if !hasScope {
+		return "", nil
+	}
+	if scope.Level != v1.SearchCategory_NODE_COMPONENTS {
+		return "", nil
+	}
+	edgeID := edges.EdgeID{ParentID: scope.ID, ChildID: resolver.data.GetId()}.ToString()
```
md2119, line 211: `postgres.IDFromPKs([]string{scope.ID, resolver.data.GetId()})`

central/graphql/resolvers/loaders/node_cves.go
```diff
@@ -0,0 +1,142 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var nodeCveLoaderType = reflect.TypeOf(storage.NodeCVE{})
+
+func init() {
+	RegisterTypeFactory(reflect.TypeOf(storage.NodeCVE{}), func() interface{} {
```
md2119, line 19: Move the registration behind feature flag

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -47,19 +52,61 @@ func (resolver *Resolver) NodeVulnerability(ctx context.Context, args IDQuery) (
 	if !features.PostgresDatastore.Enabled() {
 		return resolver.nodeVulnerabilityV2(ctx, args)
 	}
-	// TODO : Add postgres support
-	return nil, errors.New("Resolver NodeVulnerability does not support postgres yet")
+
+	if err := readNodes(ctx); err != nil {
+		return nil, err
+	}
+	vuln, exists, err := resolver.NodeCVEDataStore.Get(ctx, string(*args.ID))
```
md2119, line 59: use loader

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -24,8 +26,32 @@ type PlottedNodeVulnerabilitiesResolver struct {
...
+			all:     allCveIds,
+			fixable: fixableCount,
+		}, nil
+	}
+
+	query, err := getPlottedVulnsV1Query(args)
+	if err != nil {
+		return nil, err
+	}
+	all, err := root.NodeCVEDataStore.Search(ctx, query)
```
md2119, line 47: use loader

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -24,8 +26,32 @@ type PlottedNodeVulnerabilitiesResolver struct {
...
+	if err != nil {
+		return nil, err
+	}
+	allCveIds := search.ResultsToIDs(all)
+
+	fixableCount, err := root.NodeCVEDataStore.Count(ctx,
+		search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery()))
```
md2119, line 54: Ensure that the incoming query does not have `Fixable` in it else the result could be unexpected.

central/graphql/resolvers/nodes.go
```diff
@@ -510,12 +524,7 @@ func (resolver *nodeResolver) NodeVulnerabilityCounter(ctx context.Context, args
 	if err := readNodes(ctx); err != nil {
 		return nil, err
 	}
-
-	if !features.PostgresDatastore.Enabled() {
-		return resolver.root.NodeVulnCounter(resolver.nodeScopeContext(ctx), args)
-	}
-	// TODO : Add postgres support
-	return nil, errors.New("Sub-resolver NodeVulnerabilityCounter in Node does not support postgres yet")
+	return resolver.root.NodeVulnCounter(resolver.nodeScopeContext(ctx), args)
```
md2119, line 527: `withNodeScopeContext`

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -49,7 +75,41 @@ func (pvr *PlottedNodeVulnerabilitiesResolver) BasicNodeVulnerabilityCounter(_ c
 
 // NodeVulnerabilities returns the node vulnerabilities for top risky nodes scatter-plot
 func (pvr *PlottedNodeVulnerabilitiesResolver) NodeVulnerabilities(ctx context.Context, args PaginatedQuery) ([]NodeVulnerabilityResolver, error) {
```
md2119, line 77: The graphQL field is `"nodeVulnerabilities(pagination: Pagination): [NodeVulnerability]!",` but the resolver is accepting full query—query string and pagination. Update this resolver sig to only accept the pagination.

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -49,7 +75,41 @@ func (pvr *PlottedNodeVulnerabilitiesResolver) BasicNodeVulnerabilityCounter(_ c
...
+	if err != nil {
+		return nil, err
+	}
+
+	vulns, err := pvr.root.NodeCVEDataStore.GetBatch(ctx, cvesInterface.([]string))
+	if err != nil {
+		return nil, err
+	}
+
+	vulnResolvers, err := pvr.root.wrapNodeCVEs(vulns, err)
```
md2119, line 112: Create a raw query for cve ids with pagination and redirect to top-level node vulnerability resolver.

central/graphql/resolvers/plotted_node_vulnerabilities.go
charmik-redhat: Hmm, I see a few other places where a conjunction query is added without ensuring the input query does not have the fieldLabel. I would add `ExcludeFieldLabel` options to other places as well.

central/graphql/resolvers/clusters.go
```diff
@@ -628,38 +628,26 @@ func (resolver *clusterResolver) VulnCounter(ctx context.Context, args RawQuery)
...
 func (resolver *clusterResolver) NodeVulnerabilityCounter(ctx context.Context, args RawQuery) (*VulnerabilityCounterResolver, error) {
 	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.Cluster, "NodeVulnerabilityCounter")
 
-	if !features.PostgresDatastore.Enabled() {
-		query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
-		return resolver.root.NodeVulnCounter(ctx, RawQuery{Query: &query})
-	}
-	// TODO : Add postgres support
-	return nil, errors.New("Sub-resolver NodeVulnerabilityCounter in Cluster does not support postgres yet")
+	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getClusterRawQuery())
```
keyallis, line 649: missing comment for rox-10911

central/graphql/resolvers/common_vulnerabilities.go
```diff
@@ -33,7 +33,6 @@ var commonVulnerabilitySubResolvers = []string{
 	"suppressed: Boolean!",
 	"unusedVarSink(query: String): Int",
 	"vectors: EmbeddedVulnerabilityVectors",
-	"vulnerabilityState: String!",
```
keyallis, line 36: Why is this being pulled out and added to just node and image vulns? Is it not appropriate for cluster?

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+		return nil, err
+	}
+
+	unFixableVulnsQuery := search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, false).ProtoQuery())
+	unFixableCVEs, err := vulnLoader.FromQuery(ctx, unFixableVulnsQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	return mapNodeCVEsToVulnerabilityCounter(fixableVulns, unFixableCVEs), nil
```
keyallis, line 171: Check out the new map function and code flows as added in #2253

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+}
+
+// CreatedAt is the time a node CVE first seen in the system
+func (resolver *nodeCVEResolver) CreatedAt(ctx context.Context) (*graphql.Time, error) {
+	value := resolver.data.GetCveBaseInfo().GetCreatedAt()
+	return timestamp(value)
```
keyallis, line 177: nit: can be one lined

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+// FixedByVersion returns the version of the parent component that removes this CVE
+func (resolver *nodeCVEResolver) FixedByVersion(_ context.Context) (string, error) {
+	scope, hasScope := scoped.GetScope(resolver.ctx)
+	if !hasScope {
+		return "", nil
+	}
+	if scope.Level != v1.SearchCategory_NODE_COMPONENTS {
+		return "", nil
+	}
+	edgeID := postgres.IDFromPks([]string{scope.ID, resolver.data.GetId()})
```
keyallis, line 213: @mandar mentioned that we can and should remove the dependency on this function, check out #2253 to see how I handled this function

central/graphql/resolvers/node_vulnerabilities.go
keyallis: nit: can be one lined

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+
+// Link to the node CVE
+func (resolver *nodeCVEResolver) Link(ctx context.Context) string {
+	return resolver.data.GetCveBaseInfo().GetLink()
+}
+
+// PublishedOn is date and time when this node CVE was first published in the cve feeds
+func (resolver *nodeCVEResolver) PublishedOn(ctx context.Context) (*graphql.Time, error) {
+	value := resolver.data.GetCveBaseInfo().GetPublishedOn()
+	return timestamp(value)
```
keyallis, line 302: nit: can be one lined

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+
+// Summary of the node CVE
+func (resolver *nodeCVEResolver) Summary(ctx context.Context) string {
+	return resolver.data.GetCveBaseInfo().GetSummary()
+}
+
+// SuppressActivation returns the snooze start timestamp of the node CVE
+func (resolver *nodeCVEResolver) SuppressActivation(ctx context.Context) (*graphql.Time, error) {
+	value := resolver.data.GetSnoozeStart()
+	return timestamp(value)
```
keyallis, line 319: nit: can be one lined

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+			resolver: &cVSSV2Resolver{resolver.ctx, resolver.root, val},
+		}
+	}
+	return nil
+}
+
+// NodeComponentCount is the number of node components that contain the node CVE.
+func (resolver *nodeCVEResolver) NodeComponentCount(ctx context.Context, args RawQuery) (int32, error) {
+	// TODO implement me (ROX-11299)
+	panic("implement me")
```
keyallis, line 356: Better to return `0, errors.New("...")` here for the time being

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+}
+
+// NodeComponents are the node components that contain the node CVE.
+func (resolver *nodeCVEResolver) NodeComponents(ctx context.Context, args PaginatedQuery) ([]NodeComponentResolver, error) {
+	// TODO implement me (ROX-11299)
+	panic("implement me")
```
keyallis, line 362: Better to return `nil, errors.New("...")` here for the time being

central/graphql/resolvers/plotted_node_vulnerabilities.go
md2119: go for it

central/graphql/resolvers/common_vulnerabilities.go
charmik-redhat: I think it is only appropriate for image vulns and is not needed for cluster and node vulns

central/graphql/resolvers/node_vulnerabilities.go
charmik-redhat: Updated the code flow

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -80,8 +143,243 @@ func (resolver *Resolver) NodeVulnCounter(ctx context.Context, args RawQuery) (*
...
+
+// ID of the node CVE
+func (resolver *nodeCVEResolver) ID(ctx context.Context) graphql.ID {
+	value := resolver.data.GetId()
+	return graphql.ID(value)
+}
+
+// IsFixable returns whether node CVE is fixable by any component
+func (resolver *nodeCVEResolver) IsFixable(ctx context.Context, args RawQuery) (bool, error) {
+	query, err := args.AsV1QueryOrEmpty(search.ExcludeFieldLabel(search.CVEID), search.ExcludeFieldLabel(search.Fixable))
```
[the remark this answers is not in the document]
charmik-redhat: Updated it

central/graphql/resolvers/common_vulnerabilities.go
md2119: Yes, `vulnerabilityState` is only supported for image vulns.

central/graphql/resolvers/nodes.go
```diff
@@ -383,18 +387,35 @@ func (resolver *nodeResolver) TopNodeVulnerability(ctx context.Context, args Raw
...
+
+	vulnLoader, err := loaders.GetNodeCVELoader(ctx)
+	if err != nil {
+		return nil, err
+	}
+	vulns, err := vulnLoader.FromQuery(ctx, query)
+	if err != nil {
+		return nil, err
+	} else if len(vulns) == 0 {
+		return nil, err
```
md2119, line 411:
```suggestion
	if err != nil || len(vulns) == 0 {
		return nil, err
```

central/graphql/resolvers/utils.go
```diff
@@ -357,3 +357,13 @@ func V1RawQueryAsResolverQuery(rQ *v1.RawQuery) (RawQuery, PaginatedQuery) {
 		},
 	}
 }
+
+// ErrorOnQueryContainingField logs error if the query contains the given field label
+func ErrorOnQueryContainingField(query *v1.Query, label search.FieldLabel, errMsg string) {
```
md2119, line 362: nit : `LogError...`
Also, doesn't need to be exported

central/graphql/resolvers/utils.go
```diff
@@ -357,3 +357,13 @@ func V1RawQueryAsResolverQuery(rQ *v1.RawQuery) (RawQuery, PaginatedQuery) {
...
+	search.ApplyFnToAllBaseQueries(query, func(bq *v1.BaseQuery) {
+		mfQ, ok := bq.GetQuery().(*v1.BaseQuery_MatchFieldQuery)
+		if ok && mfQ.MatchFieldQuery.GetField() == label.String() {
+			log.Errorf(errMsg)
```
md2119, line 366: Consider adding a suffix `Response maybe unexpected` to the incoming message. Since we are not erroring out, the execution continues on the caller side, and may lead to undesired results.

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -48,15 +85,23 @@ func (pvr *PlottedNodeVulnerabilitiesResolver) BasicNodeVulnerabilityCounter(_ c
...
 
-	ret := make([]NodeVulnerabilityResolver, 0, len(vulnResolvers))
-	for _, resolver := range vulnResolvers {
-		ret = append(ret, resolver)
+	if len(pvr.all) == 0 {
+		return nil, nil
 	}
-	return ret, nil
+	q := search.NewQueryBuilder().AddExactMatches(search.CVEID, pvr.all...).Query()
+	return pvr.root.NodeVulnerabilities(ctx, PaginatedQuery{Query: &q, Pagination: args.Pagination})
```
md2119, line 106: you can remove the postgres check from here since the top-level node vulnerabilities resolver handles the fork.

```
	q := search.NewQueryBuilder().AddExactMatches(search.CVEID, pvr.all...).Query()
	return pvr.root.NodeVulnerabilities(ctx, PaginatedQuery{Query: &q, Pagination: args.Pagination})
```

md2119 (approved):
Some minor comments, but otherwise lgtm

central/graphql/resolvers/nodes.go
keyallis: +1

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -47,19 +51,63 @@ func (resolver *Resolver) NodeVulnerability(ctx context.Context, args IDQuery) (
...
+	if err != nil {
+		return nil, err
+	}
+	vulnResolver.ctx = ctx
+	return vulnResolver, nil
 }
 
 // NodeVulnerabilities resolves a set of vulnerabilities based on a query.
-func (resolver *Resolver) NodeVulnerabilities(ctx context.Context, q PaginatedQuery) ([]NodeVulnerabilityResolver, error) {
+func (resolver *Resolver) NodeVulnerabilities(ctx context.Context, args PaginatedQuery) ([]NodeVulnerabilityResolver, error) {
```
keyallis, line 76: This file should have metrics added to the subresolver functions, can be addressed in a future PR just to get this one merged in

central/graphql/resolvers/plotted_node_vulnerabilities.go
```diff
@@ -24,16 +27,50 @@ type PlottedNodeVulnerabilitiesResolver struct {
...
+	allCves, err := vulnLoader.FromQuery(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	allCveIds := make([]string, 0, len(allCves))
+	for _, cve := range allCves {
+		allCveIds = append(allCveIds, cve.GetId())
+	}
+
+	ErrorOnQueryContainingField(query, search.Fixable, "Unexpected `Fixable` field in PlottedNodeVulnerabilities resolver")
```
keyallis, line 61: I'd say move this function to right after you create the query object and check for error on that. Otherwise there's a chance this doesn't get logged

central/graphql/resolvers/utils.go
keyallis: My 2c:
`log.Errorf("Unexpected field (%s) found in query on subresolver (%s)", label.String(), resolver)`
Then all you have to pass in is the field to check for and the name of the subresolver we're checking in

keyallis (approved):
Couple comments, otherwise LGTM

PR #2216, opened 2022-06-27
ROX-11533: Fix `nodeAffinity` to `Exists`/`DoesNotExist`

msugakov:
### The context

@fredrb pointed out a possibility for Sensor to get scheduled onto `master` node when its memory requirements are bumped too high. See https://github.com/stackrox/architecture-decision-records/pull/20#discussion_r905806227

I believe this will not happen in practical cases because master nodes should be tainted by default. For pod to be scheduled there either master nodes need to be un-tainted which is an administrator action, or pod has to have toleration configured (which is not the case in our manifests).

Nevertheless, I looked at our `nodeAffinit`-ies and found their expressions suspicious in that they compare label value with `true`.

Another part of the context is that [we were asked](https://srox.slack.com/archives/C02HTD51SMD/p1649174140940459) to favor OpenShift infra nodes more for running ACS. While this was mostly implemented in https://issues.redhat.com/browse/ROX-10014, I spotted that configuration was missing for Sensor.

### The problem

Node affinity `matchExpression`-s like
```yaml
                matchExpressions:
                  - key: node-role.kubernetes.io/master
                    operator: NotIn
                    values:
                    - "true"
```
and
```yaml
                matchExpressions:
                  - key: node-role.kubernetes.io/compute
                    operator: In
                    values:
                    - "true"
```
have no chance of working correctly because `node-role.kubernetes.io/*` labels don't have value equal to `true`. Their value is always an empty string `""`. Labels are present on corresponding type of nodes. Absence of the label means that the node does not correspond to the type. I confirmed this by checking both Kubernetes and OpenShift clusters.

### Steps to verify

I confirmed the problem with an artificial experiment. Here are steps I took.
1. Create an OpenShift cluster with infra.
2. Deploy master branch with `./deploy/openshift/deploy.sh`. It's important to use OpenShift because default config will skip `matchExpressions` on Kubernetes.
3. Modify Central deployment
```sh
$ kubectl -n stackrox edit deployment central
```
Insert
```yaml
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: Exists
```
just before `preferredDuringSchedulingIgnoredDuringExecution`.
4. Check Central pod status
```sh
$ kubectl -n stackrox describe po central-5f9b758d7d-rdrrn
...
  Warning  FailedScheduling  24s   default-scheduler  0/6 nodes are available: 3 node(s) didn't match Pod's node affinity, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```
This means that the Central pod matches `master` nodes but they don't let it get deployed because they have taints.
5. Untaint some master node
```sh
kubectl taint nodes ms-06-28-pastoral-nig-258bh-master-2 node-role.kubernetes.io/master=:NoSchedule-
```
6. Check Central pod status again
```sh
$ kubectl -n stackrox describe po central-5f9b758d7d-rdrrn
...
  Normal   Scheduled         62s                   default-scheduler  Successfully assigned stackrox/central-5f9b758d7d-rdrrn to ms-06-28-pastoral-nig-258bh-master-2
  Normal   AddedInterface    58s                   multus             Add eth0 [10.128.0.55/23]
  Normal   Pulled            58s                   kubelet            Container image "quay.io/stackrox-io/main:3.70.x-500-g1ec040affd" already present on machine
  Normal   Created           58s                   kubelet            Created container central
  Normal   Started           58s                   kubelet            Started container central
...
```
Central pod got launched onto the node we just untainted. This confirms that the expression with the `operator: Exists` works as expected.
Now, let's check that `operator: In`+`values: ["true"]` doesn't work.
7. Modify Central deployment again
```sh
$ kubectl -n stackrox edit deployment central
```
Find and change `requiredDuringSchedulingIgnoredDuringExecution` to be
```yaml
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: In
                values:
                  - "true"
```
Add the following toleration at the same level as `affinity:`
```yaml
      tolerations:
      - key: node-role.kubernetes.io/master
        operator: Exists             
        effect: NoSchedule
```
This allows Central pods to be **scheduled** onto master nodes. See [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/). Yes, the effect is the opposite to what it looks like.
8. Get Central pod status
```sh
$ kubectl -n stackrox describe po central-566fd8fdc5-qw6hr
...
  Warning  FailedScheduling  9s    default-scheduler  0/6 nodes are available: 6 node(s) didn't match Pod's node affinity.
```
This tells that no pods match affinity expression that we configured.
If you'd like to, change again `operator:` to `Exists` and remove `values:` to see that the pod gets scheduled successfully.

## For reviewers

Here you'll find multiple changes loosely related. It is best to check this PR by commits. I provided description with each commit so that should help.

## Checklist
- ~~[ ] Unit test and regression tests added~~ - mostly modified. I could have added a helm-test for Sensor but don't feel that the same changes will behave differently there. Will leave it for other time.
- [ ] Determined and documented upgrade steps - this is a bit of open question. Checking with Draco folks here.
- ~~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~ - requested release notes on the JIRA ticket.

## Testing Performed

* [x] Deployed successfully locally on minikube (single master node).
* [x] Deployed successfully locally on CRC (single master node).
* [x] Deployed successfully to infra GKE (just worker nodes).
* [x] Deployed successfully to infra OCP (it has both master and worker nodes).


msugakov:
Hey @SimonBaeumer , @mtesseract and @porridge 

Here's a question about writing upgrade steps for this change.
This change modifies helm charts and theoretically should be manually reflected for folks who deployed with `roxctl`+manifests.
However, I'm looking to cut this corner and not write upgrade instructions at all. My reasons are the following:
* We all hope that `roxctl`+manifests based deployment will be deprecated and so we should maintain upgrade instructions for significant changes which this one isn't, in my opinion.
* The change here doesn't make the existing deployment "better" if it is already running. It just addresses some corner case scenarios.
* Chances that someone untainted master nodes and our things get scheduled there are rather low. If that happens due to low resources on other nodes, our expression does not prohibit that because we use `preferredDuringSchedulingIgnoredDuringExecution` not `required...`.
* We've been having incorrect node affinities with `operator: In`/`NotIn` and `values: ["true"]` since 2019 and nobody complained so far that these don't work. I take that this issue was not noticed since is not critical for customers.

WDYT, any objections that I skip manual upgrade instructions?

vikin91:
Have you seen also the second annotation `node-role.kubernetes.io/control-plane`?

According to https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#mark-the-node-as-control-plane and https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#no-really-you-must-read-this-before-you-upgrade

> The node-role.kubernetes.io/master taint is deprecated and will be removed in kubeadm version 1.25

Not sure how widely it applies, but I think we should also support it.

vikin91 (approved):
LGTM - nice to read with organized commits.

I suggest to tackle the `node-role.kubernetes.io/control-plane` annotation - may be in a separate PR if you like.

Regarding the upgrade instructions - they would be needed only for yaml-based installations, right? Helm and operator should handle it automatically.

msugakov:
@vikin91 thanks for pointing out `node-role.kubernetes.io/control-plane` label! I included that in the change.

> Regarding the upgrade instructions - they would be needed only for yaml-based installations, right? Helm and operator should handle it automatically.

Correct on both points.

vikin91 (approved):
Thanks for improving!

SimonBaeumer:
@msugakov 
> WDYT, any objections that I skip manual upgrade instructions?

Sounds good! Also auto-upgrader takes care if necessary in roxctl based deployments.

PR #2219, opened 2022-06-27
Postgres test suites create separate DBs

connorgorman:
## Description

Isolate tests from one another and allow the potential for them to be run in parallel. Currently, this is still not possible as there are one off cases that still use the old code path. To not convolute this PR, it will only contain the autogenerated stores. This will also allow us to more easily test higher level code

## Checklist

## Testing Performed

CI tests, and will evaluate how long they take

### DIFF
--- a/pkg/postgres/pgtest/postgres.go
+++ b/pkg/postgres/pgtest/postgres.go
@@ -3,17 +3,14 @@ package pgtest
 import (
 	"context"
 	"database/sql"
-	"fmt"
-	"os"
 	"strings"
 	"testing"
 
 	"github.com/jackc/pgx/v4/pgxpool"
-	"github.com/stackrox/rox/pkg/postgres/pgutils"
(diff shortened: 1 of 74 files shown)


pkg/postgres/pgtest/postgres.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/stackrox/rox/pkg/postgres/pgtest/conn"
+	pkgSchema "github.com/stackrox/rox/pkg/postgres/schema"
+	"github.com/stackrox/rox/pkg/random"
+	"github.com/stretchr/testify/require"
+	"gorm.io/gorm"
+	"k8s.io/utils/env"
+
+	_ "github.com/lib/pq"
```
connorgorman, line 17: I'll tweak this to use gorm instead now that can open the postgres db

PR #2222, opened 2022-06-27
Orient ActiveComponents so they can be converted to Postgres easily

connorgorman:
## Description

The shape of ActiveComponents made it really hard to put into postgres correctly. This removes the active context map and makes it a slice which can be indexed by both bleve and postgres nicely

https://github.com/stackrox/stackrox/pull/2223 is based on this one, and puts active components in PG

## Checklist

## Testing Performed

CI should do it


viswajithiii (approved):
🚀

proto/storage/active_component.proto
```diff
@@ -10,14 +10,18 @@ package storage;
...
     message ActiveContext {
-        string container_name = 1;
-        string image_id       = 2 [(gogoproto.moretags) = "search:\"Image Sha,hidden,store\""];
+        string container_name = 1 [(gogoproto.moretags) = "search:\"Container Name,hidden\""];
+        string image_id       = 2 [(gogoproto.moretags) = "search:\"Image Sha,hidden\""];
     }
 
     // Map from container name to the active context of an edge.
-    map<string, ActiveContext> active_contexts = 2;
+    map<string, ActiveContext> DEPRECATED_active_contexts   = 2  [(gogoproto.moretags) = 'search:"-"'];
```
md2119, line 24: Is this exposed on external APIs? Changelog?

md2119 (approved):

proto/storage/active_component.proto
connorgorman: This is not exposed directly, and is only exposed as ActiveState

PR #2226, opened 2022-06-28
Add the concept of directional graph nodes

connorgorman:
## Description

Add a node that would otherwise short circuit the graph, but by marking the edges it generates as directional, makes it be unable to be traversed unless it is the source of the queries.

Note: this may not be ideal long term of we want to be able to search things like `Active`, but this will allow us to have feature parity and iterate from there

## Checklist

## Testing Performed

Unit testing on the graph


viswajithiii (approved):
Looks fine. We may need something a little more tuneable than just "directional" -- like, if you want an _in_ edge but not an _out_ edge -- but we can YAGNI it

PR #2229, opened 2022-06-28
ROX-10920: Fix ComplianceTest aggregation

connorgorman:
## Description

Let's do this again. Fix the groovy. This would fail on select orchestrators because most of the checks for docker are skipped and then if its passes for ssh, then there are no failed and no errored

## Checklist

## Testing Performed

CI


mtodor (approved):
Nice! LGTM!

PR #2230, opened 2022-06-28
Upstream release automation: further scripts extraction and refactoring

0x656b694d:
## Description

A follow-up PR with more bash scripts extracted from the upstream release workflow. No new flows or functionalities are introduced.

* A `shared.sh` script has been created and is sourced by the workflow before running main scripts;
* The scripts are refactored and a few bugs have been fixed;
* `github.event.inputs.version` is back in a couple of places, as apparently there's a condition when it is different from `inputs.version` <- to be understood and tested further;
* `DRY_RUN` is now also a global variable accessible to all scripts.

## Checklist

## Testing Performed

Manually on `stackrox/test-gh-actions` repository.


.github/workflows/scripts/common.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# The output appears in the GitHub step summary box.
+# Multiple lines are supported.
+# Markdown is supported.
+gh_summary() {
+    if [ "$#" -eq 0 ]; then
+        cat
+    else
+        echo -e "$@"
+    fi >>"$GITHUB_STEP_SUMMARY"
```
vikin91, line 43: Should:

Could we simplify this into

```suggestion
    echo -e "${@}" >>"$GITHUB_STEP_SUMMARY"
```

If `${@}` is empty then this function should add nothing to the file, right? (I am wondering why we need this `cat` in line 40 🤔

.github/workflows/scripts/patch-changelog.sh
```diff
@@ -4,46 +4,43 @@
...
 BRANCH="$3"
-DRY_RUN="$4"
 
 check_not_empty \
-    GITHUB_STEP_SUMMARY GITHUB_SERVER_URL GITHUB_REPOSITORY GITHUB_ACTOR \
-    main_branch \
-    VERSION REF BRANCH DRY_RUN
+    GITHUB_SERVER_URL GITHUB_REPOSITORY GITHUB_ACTOR \
+    DRY_RUN main_branch \
+    VERSION REF BRANCH
```
vikin91, line 14: Should: split onto 1 var per line

.github/workflows/scripts/test-env.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
vikin91, line 1: Could: This is better portable

```suggestion
#!/usr/bin/env bash
```

vikin91 (changes requested):
I will follow the [MoSCoV](https://allthecode.co/blog/post/how-to-get-better-at-code-reviews) method for this review.

There are some rough edges here, but in general it is a piece of good work 👍

.github/workflows/scripts/check-jira-issues.sh
```diff
@@ -0,0 +1,81 @@
...
+echo "Open issues:"
+echo "$OPEN_ISSUES" | jq -r '.issues[] | "\(.key) - \(.fields.assignee.displayName)"'
+
+if [ "$DRY_RUN" != "true" ]; then
+    echo "$OPEN_ISSUES" | jq -r ".issues[] | .key" | while read -r KEY; do
+        comment_issue "$KEY"
+    done
+fi
+
+exit 1
```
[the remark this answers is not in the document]
0x656b694d: Not always, but only if there are open issues. We exit 0 earlier otherwise.

.github/workflows/scripts/common.sh
0x656b694d: Cat is there to pass data from the pipe. This function can be called either with arguments:
```
gh_summary "Some" "summary"
```
or on a pipe:
```
gh_summary <<EOF
multiline
summary
EOF
```

.github/workflows/scripts/patch-changelog.sh
0x656b694d: okay

.github/workflows/start-release.yml
```diff
@@ -162,9 +171,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -uo pipefail
+          SCRIPT="common.sh"
+          source /dev/stdin <<<"$(gh api -H "$ACCEPT_RAW" "${{env.script_url}}")"
+
           SCRIPT="patch-changelog.sh"
-          gh api -H 'Accept: application/vnd.github.v3.raw' "${{env.script_url}}" | bash -s -- \
-            "${{inputs.version}}" "${{inputs.ref}}" "${{needs.variables.outputs.branch}}" "${{inputs.dry-run}}"
+          gh api -H "$ACCEPT_RAW" "${{env.script_url}}" | bash -s -- \
```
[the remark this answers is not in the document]
0x656b694d: There's no place where we could define this function. That could be a GitHub Action, but we'd need to publish it somewhere and that would be even more repetitive code for this one-liner.

.github/workflows/scripts/cherry-pick.sh
```diff
@@ -0,0 +1,131 @@
...
+
+FORK_POINT=$(find_fork_point "$BRANCH" "$main_branch")
+
+gh_log debug "Fork point: $FORK_POINT"
+
+if [ -n "$PR_COMMITS" ]; then
+    PICKED=0
+    while read -r PR_COMMIT; do
+        cherry_pick "$PR_COMMIT"
+        PICKED=$((PICKED + $?))
```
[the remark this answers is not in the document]
vikin91: we can simply mute that other function `work 2>&1 > /dev/null` or redirect it to non-stdout `2>&1 work > 3`

.github/workflows/cut-rc.yml
```diff
@@ -22,75 +22,37 @@ env:
...
 
-            exit 1
-          fi
-          echo "All issues for Jira release ${{needs.variables.outputs.release-patch}} are closed." >> $GITHUB_STEP_SUMMARY
+          SCRIPT="check-jira-issues.sh"
+          gh api -H "$ACCEPT_RAW" "${{env.script_url}}" | bash -s -- \
+            "${{env.jira_projects}}" \
+            "${{needs.variables.outputs.release}}" \
+            "${{needs.variables.outputs.patch}}" \
+            "${{needs.variables.outputs.release-patch}}"
```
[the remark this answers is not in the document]
0x656b694d: Please check the new approach. It makes the `common.sh` call the requested script.

.github/workflows/scripts/cherry-pick.sh
```diff
@@ -0,0 +1,133 @@
...
+}
+
+find_fork_point() {
+    set +o pipefail
+    diff -u \
+        <(git rev-list --first-parent "$1") \
+        <(git rev-list --first-parent "$2") |
+        sed -ne 's/^ //p' |
+        head -1
+    set -o pipefail
```
vikin91, line 71: Would something like this also work here? (I am trying to get rid of the `set +o pipefail`)

```suggestion
    local parent_branch 
    parent_branch="$(git rev-list --first-parent "$1")" || true
    local main_branch
    main_branch="$(git rev-list --first-parent "$2")" || true
    diff -u "$parent_branch" "$main_branch" | 
        sed -ne 's/^ //p' |
        head -1
```

The main question here is actually which pipe do we allow to fail and which not.

vikin91: Understanding question: The `sed -ne 's/^ //p'` does: filter lines starting with a space and print them without that space. Is this correct?

vikin91: Actually, I was thinking entire time that such function should already exist. How about using `git merge-base branch2 branch3`?

.github/workflows/scripts/common.sh
vikin91: I think this can be simplified, but there is no need to do this in this PR. Would you only add a comment in the code why we have this `cat`? It was not clear for me when reading the code.

.github/workflows/start-release.yml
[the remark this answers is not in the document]
vikin91: We could simply add it in the `common` as:

```shell
function ghapi() {
    gh api -H "Accept: application/vnd.github.v3.raw" "${@}"
}
```

or even as crazy as

```shell
gh api -H  "Accept: application/vnd.github.v3.raw" "/repos/${{github.repository}}/contents/.github/workflows/scripts/common.sh?ref=${{github.event.repository.default_branch}}" "${@}"
```

as we practically call the other scripts from the common :)

.github/workflows/scripts/common.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ "$#" -gt 0 ]; then
+    SCRIPT="$1"
+    check_not_empty \
+        SCRIPT \
+        GITHUB_REPOSITORY \
+        main_branch
+    URL="/repos/$GITHUB_REPOSITORY/contents/.github/workflows/scripts/$SCRIPT.sh?ref=$main_branch"
+    shift
+    gh api -H "Accept: application/vnd.github.v3.raw" "$URL" | bash -s -- "$@"
+fi
```
vikin91, line 56: I see now :) 
[This construct](https://github.com/stackrox/stackrox/blob/master/release/scripts/scan-images-with-roxctl.sh#L60) allows to control the behavior of the script when running it and when sourcing it. 

Just incase you want to add extra protection here (or run a command that accepts no parameters).

vikin91 (approved):
Looks much better now, thank you for improving!

My comments are not blocking, but please give them a look before merging.

.github/workflows/scripts/check-jira-issues.sh
```diff
@@ -0,0 +1,81 @@
...
+
+GH_MD_FORMAT_LINE=$(
+    cat <<EOF
+"* [\(.key)](https://issues.redhat.com/browse/\(.key)): **" \
++ (.fields.assignee.displayName // "unassigned") \
++ "** (\(.fields.status.name)) — _" \
++ (.fields.summary | gsub (" +$";"")) \
++ "_"
+EOF
+)
```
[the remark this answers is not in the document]
0x656b694d: '+' here is a `jq` operator, used to join strings. But indeed there is a slightly shorter syntax:
```
* [\(.key)](https://issues.redhat.com/browse/\(.key)): **\(.fields.assignee.displayName // "unassigned")** (\(.fields.status.name)) — _\(.fields.summary | gsub (" +$";""))_
```
The expression must be a single line, so `\` is required.

.github/workflows/start-release.yml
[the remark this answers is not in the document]
0x656b694d: So far only the `common.sh` may call other scripts, and so it will be the only user of this function. But why not.

.github/workflows/scripts/cherry-pick.sh
```diff
@@ -0,0 +1,133 @@
...
+    fi
+
+    # Find commits with the specific commit message after the fork point
+    # to not cherry-pick twice the same commit.
+    ALREADY_PICKED_CHERRIES=$(git log \
+        --grep "^(cherry picked from commit $COMMIT)\$" --format='%H' \
+        HEAD..."$FORK_POINT")
+
+    if [ -n "$ALREADY_PICKED_CHERRIES" ]; then
+        gh_summary "Already picked cherries for commit $COMMIT:\n\`\`\`\n$ALREADY_PICKED_CHERRIES\n\`\`\`"
```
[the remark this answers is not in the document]
0x656b694d: Hm, thanks, but I'm not sure how to use `printf` here to make the line look nicer.

.github/workflows/scripts/cherry-pick.sh
0x656b694d: Your suggested code still needs `set +o pipefail` as it is required to ignore the exit code of `diff`. I guess it could also be done via `if diff ... ; then; fi`.
`git merge-base` didn't always work correctly in my experiments. Teh internets suggested the solution with diff.

.github/workflows/scripts/common.sh
0x656b694d: Thanks. I don't want extra protection here :)

PR #2237, opened 2022-06-28
Update aging images query to use new date range queries.

dvail:
## Description

Updates the Aging Images widget to use new date range query functionality in central.

This allow us to send queries like `Image Updated Time:30d-90d` to group date ranges into buckets. This also enables us to pass these values via URL parameters to the VulnMgmt images page to filter that table to the selected range.

## Checklist

## Testing Performed

The test process for the widget itself is unchanged from the original test steps described in this PR:
https://github.com/stackrox/stackrox/pull/2138

The new test procedure for the x-axis links on the chart is as follows.

Load the dashboard and take note of the number of images in the 90-180 column, then click the link below the graph. The link should take you to the VulnMgmt images page with a `90d-180d` filter applied, resulting in the same number of images.
image
image

Repeat this process, but selecting the >1 year column. You should be taken to the VulnMgmt page with a `>365d` filter applied, resulting in the same number of images.
image
image

On the dashboard, deselect one of the time ranges to collapse the bucket with the one before it. Repeat the process above and ensure the VulnMgmt page displays the correct filter and number of images.
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
@@ -29,6 +29,7 @@ import AgingImagesChart, {
     TimeRangeTupleIndex,
     TimeRangeTuple,
     timeRangeTupleIndices,
+    getTimeFilterOption,
 } from './AgingImagesChart';
 import isResourceScoped from '../utils';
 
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -69,10 +72,17 @@ function getWidgetTitle(
...
-        Object.values(timeRangeCounts).find((range, index) => {
-            return selectedTimeRanges[index].enabled;
-        }) ?? 0;
+    // The total number of images is obtained by finding the first enabled bucket
+    // and getting the sum of that bucket and all following buckets.
+    let totalImages = 0;
+    const firstEnabledIndex = selectedTimeRanges.findIndex(({ enabled }) => enabled);
+    const rangesToCount = Object.values(timeRangeCounts).slice(
+        firstEnabledIndex,
+        selectedTimeRanges.length
```
pedrottimark, line 81: Not important to merge, but for my understanding: can omit length argument, or not?

> If end is omitted, slice extracts through the end of the sequence (arr.length).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

pedrottimark (approved):
Another small step toward clear and correct code.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
dvail: Oh, good catch!

PR #2204, opened 2022-06-24
Add compliance levels by standard widget

dvail:
## Description

Adds a Compliance levels by standard widget. This displays the total compliance percentage across standards at the _cluster_ aggregation level. This is very similar to the existing dashboard widget, with the exception that the data can now be filtered by Cluster and Namespace.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added

## Testing Performed

Test that compliance percentages are displayed in ascending order by default:
image

Test that the ascending sort option reverses the display of the chart:
image

Test that Cluster and Namespace filters adjust the data in the chart. Note that as the size of the selected scope decreases, the compliance percentages should always _increase_ or stay the same. This is because aggregation counts a failing standard in one resource (cluster/ns) as a failure for that standard in the entire scope.
image

Test the links from the axis labels and bars take the user to the compliance page to that specific standard:
image

Test that the 'View all' button takes the user to the main compliance page:
image

PR #2166, opened 2022-06-23
Add "Images at most risk" widget

dvail:
## Description

Adds the ["Images at Most Risk"](https://docs.google.com/document/d/1z8kQOGRkZaN_RnaG4_kKjZHMC2O5Z-hSfWM4sqhyzuI/edit#heading=h.gnwde7z6qg1o) widget.

This widget displays the top six images based on risk score along with the number of important and critical CVEs in those images. Users can toggle options to show all images or only active images, as well as toggle the display between all CVEs or _fixable_ CVEs only.

Each image name links to the VulnMgmt page, scrolled to the "Image Findings" CVE section.

## Follow Ups

- Add the blue information icon that indicates a non-default option has been applied to the table. (this may apply to other widgets as well)
- Fix the [Namespace ID filter bug](https://issues.redhat.com/browse/ROX-11510) when linking to the Vuln Mgmt page
- Sort the image list in the widget alphabetically to break ties when multiple images have the same risk priority
- General layout fixes that break widget headers in very specific screen sizes (multiple widgets)

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added

## Testing Performed

Test that the default view of the widget shows "All images" in the table.
image

Test that applying a resource filter updates the widget to only show "Active images".
image

Test that the "Image status: All images" option is disabled when a resource filter is applied.
image
image

Test that the "Image vulnerabilities: All CVEs" option updates the table to show _total_ CVE counts instead of _fixable_ CVE counts.
image

Test that the "View All" link takes the user to the Vuln Mgmt images page, with any selected filters applied:
image

Test that the individual image links in the table take the user to the Vuln Mgmt page for the individual image, focused on the "Image Findings" CVE table in the details page.
image

PR #2238, opened 2022-06-28
Swap from PF Grid component to custom grid to better fit widget content.

dvail:
## Description

~Updates the layout of the dashboard to use a custom css grid (thanks @alanonthegit) instead of the built in PatternFly Grid component. The PF Grid component relies on responsive breakpoints and classes applied via React props in order to determine the size of columns and how many columns a widget should span. This isn't a great fit for the dashboard, as a fixed column span at specific breakpoints either did not make optimal use of space or cause widget headers to overflow.~

Updates the dashboard grid to use the `<Gallery>` component, which is almost identical to the original custom CSS.

## Checklist

## Testing Performed

Test view at default 13" Macbook resolution (~1500px)
image

Shrink the browser ensuring no titles overflow. The grid will convert to single column when the browser is ~1400px wide.
image
image

Dropping down to ~1200px, the PatternFly sidebar will no longer take up horizontal space and will be set to overlay on the content behind it. This gives the widgets room to switch back to two columns until about ~1090px.
image
image

At around 540px, the single column layout will have a horizontal overflow instead of breaking widget layout.
image

Very large screens will adjust to a 3 column layout.
image


dvail:
* master
  * **PR #2138** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2138" target="_blank">Graphite</a> 
    * **PR #2166** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2166" target="_blank">Graphite</a> 
    * **PR #2190** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2190" target="_blank">Graphite</a> 
    * **PR #2204** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2204" target="_blank">Graphite</a> 
      * **PR #2238** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2238" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2238?utm_source=stack-comment).

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.tsx
```diff
@@ -129,7 +129,7 @@ function ComplianceLevelsByStandard() {
             header={
                 <Flex direction={{ default: 'row' }}>
                     <FlexItem grow={{ default: 'grow' }}>
-                        <Title headingLevel="h2">Compliance coverage by standard</Title>
+                        <Title headingLevel="h2">Compliance by standard</Title>
```
dvail, line 132: This buys us 70px on the minimum widget size, which translates into 140px of extra space before we drop to a single column layout.

ui/apps/platform/src/Containers/Dashboard/PatternFly/DashboardPage.css
```diff
@@ -0,0 +1,13 @@
+.widget-grid {
+  --min-widget-width: 510px;
```
dvail, line 2: This should prevent headers from overflowing on widgets until the "# Active aging images widget" exceeds 10,000 images total in the title.

vjwilson (approved):
I'm not a fan of going off-script for styles, so I think that after this ships, we should present this use case to the PatternFly team, so that they can adopt it somehow as part of their offering.

World domination, one CSS class at a time.

alanonthegit:
@dvail this looks great! Thanks for your hard work. I'm exploring a way to condense the options button, which will let us shrink cards a bit more and make the headers more future-proof. That should translate to minor tweaks to the min-width and max-width values. Will follow up with a mock.

dvail:
@vjwilson I took a closer look at this, and PatternFly is way ahead of us. Customizing the <Grid> component was what I was looking at originally, but they offer a <Gallery> component which is essentially this exact use case. World domination will have to wait.

@alanonthegit The only thing I didn't include (to avoid custom CSS) was a max-width on the grid, so someone with an enormous screen will see 4, 5, or more widgets side-by-side. Is this OK or do we still want to enforce a limit here?

alanonthegit:
@dvail Oh great! I still think we should enforce a max limit of 3 per row or else the page will become unbalanced.

PR #2239, opened 2022-06-29
Fix table row duplication bug in risky deployments widget

dvail:
## Description

Fixes a bug that resulted in duplicate table rows if deployments across clusters had the same name. This change uses a more sensible field (id) as the React element key...

## Checklist

## Testing Performed

Against a demo with multiple clusters, this change was easily reproducible by filtering the display to one cluster at a time. Each time a duplicate namespace appeared in the list it would be duplicated in the table.

Testing this now always results in 6 or less deployments in the table.


dvail:
* master
  * **PR #2239** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2239" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2239?utm_source=stack-comment).

pedrottimark (approved):
Disregard the integration test failure which is unrelated. Will fix it as soon as possible.

PR #2249, opened 2022-06-29
Enable feature flag for security metrics dashboard phase one

dvail:
## Description

Enables the new dashboard as the default.

## Checklist

## Testing Performed

TODO(replace-me)

### DIFF
--- a/ui/apps/platform/cypress/helpers/main.js
+++ b/ui/apps/platform/cypress/helpers/main.js
@@ -1,5 +1,5 @@
 import * as api from '../constants/apiEndpoints';
-import { pfUrl, url } from '../constants/DashboardPage';
+import { url } from '../constants/DashboardPage';
 import navSelectors from '../selectors/navigation';
 
 import { visit } from './visit';
@@ -26,7 +26,7 @@ export function visitMainDashboard(requestConfig, staticResponseMap) {
 
 // TODO Make this the default once phase one of the PF Dashboard is enabled
 export function visitMainDashboardPF() {
-    visit(pfUrl);
+    visit(url);
 }
 
 export function visitMainDashboardViaRedirectFromUrl(redirectFromUrl) {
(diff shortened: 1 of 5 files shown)


dvail:
* master
  * **PR #2249** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2249" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2249?utm_source=stack-comment).

ui/apps/platform/cypress/helpers/main.js
```diff
@@ -26,7 +26,7 @@ export function visitMainDashboard(requestConfig, staticResponseMap) {
 
 // TODO Make this the default once phase one of the PF Dashboard is enabled
 export function visitMainDashboardPF() {
-    visit(pfUrl);
+    visit(url);
```
pedrottimark, line 29: What are pro and con instead to delete this function and replace it with `visitMainDashboard` in cypress/integration/dashboard/dashboardSecurityMetrics.test.js to ensure single source for all helpers and tests which also call it?

pedrottimark (approved):
One question about main test helper for your consideration.

ui/apps/platform/cypress/helpers/main.js
dvail: That was what I went with first, but the wait in `visitMainDashboard` was conflicting with a wait in the individual e2e tests for the new dashboard. The request we are awaiting in the individual tests would fire and resolve before the `deploymentsWithProcessInfo` one and cause a failure.

I went with this as a less-intrusive way to enable the feature flag without (hopefully) causing new flakes before the release. My plan is to replace this test with jest/testing-library tests once the dust settles. The original awaited network call should work fine for other tests that rely on `visitMainDashboard` since both the old dashboard and the new one make that request.

pedrottimark: Aha, that makes sense.

PR #2250, opened 2022-06-29
Disable monitoring by default

stehessel:
If `MONITORING_SUPPORT=true`, then a Prometheus and Grafana instance are
deployed. The Grafana instance contains debugging dashboards for
Central and Sensor. It is a debug option which should be disabled by
default.

Note that metrics are exposed by Central and Sensor regardless of
`MONITORING_SUPPORT`.

## Checklist

## Testing Performed

Tested that `./deploy/openshift/deploy.sh` defaults to `MONITORING_SUPPORT=false`.


connorgorman:
why

stehessel:
It makes the deployment for complicated and seems mostly unused and unnecessary to me. The concrete trigger is this issue https://issues.redhat.com/browse/ROX-11546 - which I'll look at separately.

stehessel:
I'll leave this as a judgement call to you @connorgorman .

PR #2253, opened 2022-06-29
ROX-11257: GraphQL Support for Postgres in Image Vulnerability Resolvers

keyallis:
## Description

Support for Postgres in the image vulnerability resolver

## Checklist


central/graphql/resolvers/gen/main.go
```diff
@@ -20,14 +20,17 @@ var (
 			reflect.TypeOf((*storage.ActiveComponent_ActiveContext)(nil)),
 			reflect.TypeOf((*storage.Alert)(nil)),
 			reflect.TypeOf((*storage.Cluster)(nil)),
+			//reflect.TypeOf((*storage.ClusterCVE)(nil)), TODO rox-11259
 			reflect.TypeOf((*storage.ComplianceAggregation_Response)(nil)),
 			reflect.TypeOf((*storage.ComplianceControlResult)(nil)),
 			reflect.TypeOf((*storage.CVE)(nil)),
+			reflect.TypeOf((*storage.CVEInfo)(nil)),
```
md2119, line 27: You do not need this. Since you are generating `ImageCVE` in which `CVEInfo` is embedded, a separate resolver is generated for `CVEInfo`

central/graphql/resolvers/loaders/image_cves.go
```diff
@@ -0,0 +1,142 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var imageCveLoaderType = reflect.TypeOf(storage.ImageCVE{})
+
+func init() {
+	RegisterTypeFactory(reflect.TypeOf(storage.ImageCVE{}), func() interface{} {
```
md2119, line 19: feature flag

central/graphql/resolvers/vulnerability_counter.go
```diff
@@ -104,6 +104,25 @@ func mapVulnsToVulnerabilityCounter(vulns []*storage.EmbeddedVulnerability) *Vul
 	return counter
 }
 
+func mapImageCVEsToVulnerabilityCounter(fixable, unFixable []*storage.ImageCVE) *VulnerabilityCounterResolver {
```
md2119, line 107: Implement an interface so that we don't have to implement such func for each vuln type.
Something like,
```
type CVEWithSev interface {
GetId()
GetSeverity()
}
```

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +199,323 @@ func withImageCveTypeFiltering(q string) string {
...
+
+func (resolver *imageCVEResolver) getImageCVEQuery() *v1.Query {
+	return search.NewQueryBuilder().AddExactMatches(search.CVEID, resolver.data.GetId()).ProtoQuery()
+}
+
+func (resolver *imageCVEResolver) getImageCVERawQuery() string {
+	return search.NewQueryBuilder().AddExactMatches(search.CVEID, resolver.data.GetId()).Query()
+}
+
+func (resolver *imageCVEResolver) EnvImpact(ctx context.Context) (float64, error) {
```
md2119, line 218: urgh...this field

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +199,323 @@ func withImageCveTypeFiltering(q string) string {
...
+	if scope.Level != v1.SearchCategory_IMAGE_COMPONENTS {
+		return "", nil
+	}
+
+	edgeID := edges.EdgeID{ParentID: scope.ID, ChildID: resolver.data.GetId()}.ToString()
+	edge, found, err := resolver.root.ComponentCVEEdgeDataStore.Get(ctx, edgeID)
+	if err != nil || !found {
+		return "", err
+	}
+	return edge.GetFixedBy(), nil
```
md2119, line 244: see comments on #2214

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +199,323 @@ func withImageCveTypeFiltering(q string) string {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "could not determine vulnerability discovered time in image")
+		}
+	}
+
+	if imageID == "" {
+		return nil, nil
+	}
+
+	edgeID := edges.EdgeID{ParentID: imageID, ChildID: resolver.data.GetId()}.ToString()
```
md2119, line 447: - This will not work on postgres
- We should avoid constructing IDs at service layer. Instead, use query.

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +212,323 @@ func withImageCveTypeFiltering(q string) string {
...
+		return false, err
+	}
+
+	// check scoping, add as conjunction if needed
+	scope, ok := scoped.GetScope(ctx)
+	if !ok || scope.Level != v1.SearchCategory_IMAGE_VULNERABILITIES {
+		query = search.ConjunctionQuery(query, resolver.getImageCVEQuery())
+	}
+
+	query = search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery())
```
md2119, line 272: To avoid deep unwrapping because multiple levels of conjunctions, do this way:

```
conjuncts := []*v1.Query{query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery()}
if scope, ok := scoped.GetScope(ctx); !ok || scope.Level != v1.SearchCategory_IMAGE_VULNERABILITIES {
  conjuncts = append(conjuncts, resolver.getImageCVEQuery())
}
query = search.ConjunctionQuery(conjuncts...)
```

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +212,323 @@ func withImageCveTypeFiltering(q string) string {
...
+	scopeQuery, err := args.AsV1QueryOrEmpty()
+	if err != nil {
+		return nil, err
+	}
+	deploymentID := getDeploymentScope(scopeQuery, ctx, resolver.ctx)
+	if deploymentID == "" {
+		return nil, nil
+	}
+	// We only support OS level component. The active state is not determined if there is no OS level component associate with this vuln.
+	query := search.NewQueryBuilder().AddExactMatches(search.CVEID, resolver.data.GetId()).AddStrings(search.ComponentSource, storage.SourceType_OS.String()).ProtoQuery()
```
md2119, line 375: We can avoid this query by checking if the operating system field of this image vuln is not set. @connorgorman confirm.

central/graphql/resolvers/image_vulnerabilities.go
md2119: `postgres.IDFromPKs`

central/graphql/resolvers/image_vulnerabilities.go
connorgorman: No, we still need this query because this checks that the components causing this vuln are from the OS and not languages. Just having the OS will also exist for language vulns because they are coming from the result of that OS

md2119: @connorgorman This is an important point. There seems to be a big gap that existed in dackbox since `Source Type` field actually did not exist on any top-level protos—storage.ImageComponent and edges. So the queries never would have worked.

md2119: Found another field `executables` that never existed in the top-level proto.

md2119: Also, for any given image, the component may come from lang as well as OS. Therefore, seems like the current data model as well as scanner interpretation seems to not fully address the issue.

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -91,8 +159,69 @@ func (resolver *Resolver) ImageVulnerabilityCounter(ctx context.Context, args Ra
...
+
+	// get loader
+	loader, err := loaders.GetImageCVELoader(ctx)
+	if err != nil {
+		return nil, err
+	}
+	query = tryUnsuppressedQuery(query)
+
+	// get fixable vulns
+	fixableQuery := search.ConjunctionQuery(query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery())
```
md2119, line 182: Make sure `query` does not have `search.Fixable`. If the query has `Fixable=false` then the resolver will return an unexpected response. 
Ideally, we should throw an error to the user. Someone querying this field and passing a fixable query is misuse. But for now log an error because I feel UI could be passing incorrect queries and we do not want to fail.

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +230,313 @@ func withImageCveTypeFiltering(q string) string {
...
+	}
+	if scope.Level != v1.SearchCategory_IMAGE_COMPONENTS {
+		return "", nil
+	}
+
+	edgeID := postgres.IDFromPks([]string{scope.ID, resolver.data.GetId()})
+	edge, found, err := resolver.root.ComponentCVEEdgeDataStore.Get(ctx, edgeID)
+	if err != nil || !found {
+		return "", err
+	}
```
md2119, line 263: Thought about this a little more. We can (should) remove dependency on `postgres.IDFromPks`. Instead search `ComponentCVEEdgeDataStore` for `Component ID=scope.ID AND CVE ID = resolver.data.GetId()`

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +230,313 @@ func withImageCveTypeFiltering(q string) string {
...
+	return edge.GetFixedBy(), nil
+}
+
+func (resolver *imageCVEResolver) IsFixable(ctx context.Context, args RawQuery) (bool, error) {
+	query, err := args.AsV1QueryOrEmpty(search.ExcludeFieldLabel(search.CVEID))
+	if err != nil {
+		return false, err
+	}
+
+	conjuncts := []*v1.Query{query, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery()}
```
md2119, line 273: Make sure `query` does not have `Fixable` at this point. Ideally, we should throw an error to the user. Someone querying this field and passing a fixable query is misuse. For now, log an error because I feel UI could be passing incorrect queries and we do not want to fail.

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -101,3 +230,313 @@ func withImageCveTypeFiltering(q string) string {
...
+		}
+	}
+
+	if imageID == "" {
+		return nil, nil
+	}
+
+	edgeID := postgres.IDFromPks([]string{scope.ID, resolver.data.GetId()})
+
+	edge, found, err := resolver.root.ImageCVEEdgeDataStore.Get(ctx, edgeID)
```
md2119, line 470: change to query

md2119 (approved):
some minor comments, otherwise lgtm

central/graphql/resolvers/image_vulnerabilities.go
keyallis: How do you tell if there's an already included scoping on fixable? I feel like the easiest way to do this would be to just remove filtering on fixable for this resolver since we expressly pass back fixable and unfixable counts

md2119 (approved):

PR #2256, opened 2022-06-29
ROX-10299: Adding PolicyCategory tab to Policy Management

alwayshooin:
## Description

As title states above. Nothing too crazy, just bare bones adding the Policy Category tab in the Policy Management page behind `ROX_NEW_POLICY_CATEGORIES` and moving some files so that `PolicyManagment` and `PoilcyCategories` are their own directories. 

image

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


ui/apps/platform/src/Containers/MainPage/Body.tsx
```diff
@@ -114,7 +117,7 @@ function Body({ hasReadAccess, isFeatureFlagEnabled }: BodyProps): ReactElement
                     <Route path={violationsPath} component={AsyncViolationsPage} />
                     <Route path={compliancePath} component={AsyncCompliancePage} />
                     <Route path={integrationsPath} component={AsyncIntegrationsPage} />
-                    <Route path={policiesPath} component={AsyncPolicyManagementPage} />
+                    <Route path={policyManagementBasePath} component={AsyncPolicyManagementPage} />
                     <Redirect exact from={deprecatedPoliciesPath} to={policiesPath} />
```
pedrottimark, line 121: This reminds me to ask if you know some sources of inbound links? I might work on that in 72, if it helps.

For example, I know global search will take some effort to change to new path.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { SimpleList, SimpleListItem } from '@patternfly/react-core';
+
+type PolicyCategoriesListProps = {
+    // policyCategories: {
+    //     id: string;
+    //     name: string;
+    //     isDefault: boolean;
+    // }[];
```
pedrottimark, line 9: This reminds me, let’s collaborate on writing new endpoints and especially frontend types for backend protos when they become available, as a way to improve the information about APIs in ui/apps/platform/README.md

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type PolicyCategoriesListSectionProps = {
+    policyCategories: string[];
+};
+
+function PolicyCategoriesListSection({ policyCategories }: PolicyCategoriesListSectionProps) {
+    // const [selectedCategory, setSelectedCategory] = useState('');
+    return (
+        <PageSection isFilled id="policy-categories-list-section">
+            <PageSection isFilled variant="light">
```
pedrottimark, line 14: Let’s research an alternative to nested `section` HTML elements to achieve the needed presentation.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+            .then((categories) => {
+                setPolicyCategories(categories);
+                setErrorMessage('');
+            })
+            .catch((error) => {
+                setPolicyCategories([]);
+                setErrorMessage(getAxiosErrorMessage(error));
+            })
+            .finally(() => setIsLoading(false));
+    }, []);
```
pedrottimark, line 60: Not a blocker to merge, it seems like `useEffect` is often closer to `useState` hooks and `pageContent` is closer to `return` statement.

ui/apps/platform/src/Containers/PolicyManagement/PolicyManagementPage.tsx
```diff
@@ -12,6 +18,9 @@ function PolicyManagementPage() {
             <Route path={policiesPath}>
                 <PoliciesPage />
             </Route>
+            <Route path={policyCategoriesPath}>
+                <PolicyCategoriesPage />
+            </Route>
```
pedrottimark, line 23: Does policy categories route need to have conditional rendering according to feature flag?

Comment about preceding unchanged line 17 can you verify that it is **not** needed here, because I think the `Redirect` element in Body.tsx is enough.

ui/apps/platform/src/routePaths.js
```diff
@@ -131,7 +132,9 @@ export const basePathToLabelMap = {
     [apidocsPath]: 'API Reference',
     [productDocsPath]: 'Help Center',
     [clustersBasePath]: 'Clusters',
+    [policyManagementBasePath]: 'Policy Management',
     [policiesBasePath]: 'Policy Management',
+    [policyCategoriesPath]: 'Policy Categories',
```
pedrottimark, line 137: For your consideration, left nav uses these strings, but Access Control takes responsibility for page titles:
* Access Control - Auth providers
* Access Control - Auth provider
* Access Control - Roles
* Access Control - Role
* Access Control - Permission sets
* Access Control - Permission set
* Access Control - Access scopes
* Access Control - Access scope

Here is an application of that pattern:
* Policy management - Policies
* Policy management - Policy
* Policy management - Policy categories

pedrottimark (approved):
Super step forward.
* One two-part comment about routes in PolicyManagementPage.tsx to consider now.
* A few comments to consider for future follow up.
* A few comments about possible collaboration on future follow up in 72.

ui/apps/platform/src/Containers/MainPage/Body.tsx
alwayshooin: hm i'm not sure what you mean by sources of inbound links, could you clarify?

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
alwayshooin: yeah it's not the best but it was the quickest way to get the look I was looking for. open to other ideas to achieve the same effect that we can adopt across the rest of the UI!

ui/apps/platform/src/Containers/PolicyManagement/PolicyManagementPage.tsx
alwayshooin: hmm good question. i can check without the flag just in case, but i'm under the impression that it's okay

alwayshooin: yeah it is, i'll remove it

ui/apps/platform/src/routePaths.js
alwayshooin: i'll address in the next PR, thanks

ui/apps/platform/src/Containers/MainPage/Body.tsx
pedrottimark: Links using the classic /main/policies path that need to be replaced with /main/policy-management/policies as a prerequisite to delete the redirect.

PR #2258, opened 2022-06-29
ROX-11202: Add links to split CVE types

vjwilson:
## Description

This PR displays links to the new sub-categories of CVEs, when the ROX_FRONTEND_VM_UDPATES feature flag is on, but keeps the existing links to generic CVE lists, when the flag is off.

This only affects the following links:
* Header tiles on the VM Dashboard
* Entities "shortcut" menu on the Workflow List layout
* Entities "shortcut" menu on the Workflow Entity (single) layout

Note: the 2nd and 3rd menus look the same, but had to be changed in 2 different layout files because they need access to the feature flag.

I though about abstracting the repeated function to its own file, but soon realized I would need a curried function to pass the flag in, and since this check will only survive one release, I decided the repeated code was actually simpler to reason about.

## Checklist
- [x] Investigated and inspected CI test results (failures are unrelated flakes)

## Testing Performed

Manual testing with the flag ON

Dashboard:
with menu closed
Screen Shot 2022-07-01 at 7 16 01 PM

with menu open
Screen Shot 2022-07-01 at 7 16 07 PM

Inside list page:
Screen Shot 2022-06-29 at 5 45 53 PM

Inside single page:
Screen Shot 2022-06-29 at 5 46 06 PM

Manual testing with the flag OFF

Dashboard:
Screen Shot 2022-06-29 at 6 28 32 PM

Inside page:
Screen Shot 2022-06-29 at 6 28 45 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
@@ -82,11 +82,6 @@ const VulnDashboardPage = ({ history }) => {
                     {showVmUpdates && (
                         <div className="flex mr-2">
                             <CvesMenu />
-                            {/*
-                            <CvesCountTile entityType={entityTypes.IMAGE_CVE} />
-                            <CvesCountTile entityType={entityTypes.NODE_CVE} />
-                            <CvesCountTile entityType={entityTypes.CLUSTER_CVE} />
-                            */}
                         </div>
                     )}
                     <PoliciesCountTile />
--- a/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/CvesMenu.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Dashboard/CvesMenu.js
@@ -11,37 +11,24 @@ function getURL(workflowState, entityType) {
     return url;
 }
 
-const cveCountQueriesMap = {
-    [entityTypes.IMAGE_CVE]: gql`
-        query imageCvesCount($query: String) {
-            vulnerabilityCount: imageVulnerabilityCount
-            fixableCveCount: imageVulnerabilityCount(query: $query)
-        }
-    `,
-    [entityTypes.NODE_CVE]: gql`
-        query nodeCvesCount($query: String) {
-            vulnerabilityCount: nodeVulnerabilityCount
-            fixableCveCount: nodeVulnerabilityCount(query: $query)
-        }
-    `,
-    [entityTypes.CLUSTER_CVE]: gql`
-        query clusterCvesCount($query: String) {
-            vulnerabilityCount: clusterVulnerabilityCount
-            fixableCveCount: clusterVulnerabilityCount(query: $query)
-        }
-    `,
-};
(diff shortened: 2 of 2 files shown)


ui/apps/platform/src/Containers/Workflow/WorkflowEntityPageLayout.js
```diff
@@ -88,6 +91,23 @@ const WorkflowEntityPageLayout = ({ location }) => {
...
+    const showVmUpdates = isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES');
+    const useCaseOptions = useCaseEntityMap[useCase].filter((option) => {
+        if (showVmUpdates) {
+            if (option === entityTypes.CVE) {
+                return false;
+            }
+        } else if (
+            option === entityTypes.IMAGE_CVE ||
+            option === entityTypes.NODE_CVE ||
+            option === entityTypes.CLUSTER_CVE
```
sachaudh, line 104: I think I made changes here as well. You can go with your solution since I think it's cleaner. The only changes you'll need to make is adding a check for `entityTypes.NODE_COMPONENT` and `entityTypes.IMAGE_COMPONENT` here and `entityTypes.COMPONENT` above. Also, adding the `NODE_COMPONENT` and `IMAGE_COMPONENT` to the that entity map

ui/apps/platform/src/Containers/Workflow/WorkflowListPageLayout.js
```diff
@@ -66,6 +69,23 @@ const WorkflowListPageLayout = ({ location }) => {
...
+            }
+        } else if (
+            option === entityTypes.IMAGE_CVE ||
+            option === entityTypes.NODE_CVE ||
+            option === entityTypes.CLUSTER_CVE
+        ) {
+            return false;
+        }
+        return true;
+    });
```
sachaudh, line 87: Same here as above

ui/apps/platform/src/Containers/VulnMgmt/Dashboard/VulnMgmtDashboardPage.js
```diff
@@ -72,7 +76,15 @@ const VulnDashboardPage = ({ history }) => {
             <div className="flex items-center">
                 <div className="flex h-full mr-3 pr-3 border-r-2 border-base-400">
                     <PoliciesCountTile />
-                    <CvesCountTile />
+                    {showVmUpdates ? (
+                        <>
+                            <CvesCountTile entityType={entityTypes.IMAGE_CVE} />
+                            <CvesCountTile entityType={entityTypes.NODE_CVE} />
+                            <CvesCountTile entityType={entityTypes.CLUSTER_CVE} />
+                        </>
```
sachaudh, line 84: The only issue with having multiple tiles is that it'll push the export button offscreen and look a bit odd

Screen Shot 2022-06-30 at 12 15 54 PM

cc: @alanonthegit

vjwilson: @alanonthegit I had an idea for handling the links to the three new types of CVE lists, without having to re-engineer the entire top bar.

How about I make the CVE links a dropdown list of links, like the *Applications & Infrastructure" list of links?

sachaudh (approved):
Other than the comments I made, looks good to me

PR #2260, opened 2022-06-29
CI: handle test-bin issue with nightlies

gavin-stackrox:
## Description

OpenShift CI periodics do not set a JOB_NAME_SAFE env in the test_binary_build_commands context and so the roxctl build for stackrox nightlies is not built with the nightly tag. And so the test-bin provided to e2e and system tests has an incorrectly versioned roxctl and deploy.sh will try to docker run a matching one which is not supported in OpenShift CI. 

This PR is a workaround that recompiles roxctl for e2e and system tests to fix the version for nightly test runs.

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -480,7 +480,7 @@ is_tagged() {
 }
 
 is_nightly_run() {
-    [[ "${CIRCLE_TAG:-}" =~ ^nightly- ]]
+    [[ "${CIRCLE_TAG:-}" =~ nightly ]]
 }
 
 is_in_PR_context() {
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -859,6 +859,27 @@ handle_nightly_runs() {
     fi
 }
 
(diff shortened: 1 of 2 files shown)


scripts/ci/lib.sh
```diff
@@ -859,6 +859,27 @@ handle_nightly_runs() {
     fi
 }
 
+handle_nightly_roxctl_mismatch() {
+    if ! is_OPENSHIFT_CI; then
+        die "Only for OpenShift CI"
+    fi
+
+    if is_in_PR_context || ! [[ "${JOB_NAME_SAFE:-}" =~ ^nightly- ]]; then
```
RTann, line 867: how come this one is `^nightly-` but we only check for `nightly` in `is_nightly_run`?

gavin-stackrox: JOB_NAME_SAFE comes from the test name in openshift/release.
CIRCLE_TAG is how we 'tag' nightlies e.g. 3.70.x-nightly-20220530

PR #2261, opened 2022-06-30
ClusterCVE store

md2119:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


central/cluster/datastore/datastore_impl.go
```diff
@@ -545,6 +545,8 @@ func (ds *datastoreImpl) postRemoveCluster(ctx context.Context, cluster *storage
 	ds.removeK8SRoles(ctx, cluster)
 	ds.removeRoleBindings(ctx, cluster)
 
+	// TODO: Apparently, cluster cves are only cleaned up at next cve fetch cycle.
```
connorgorman, line 548: This seems like a good example of when we should use a hard constraint with cascade?

central/cve/cluster/datastore/store/postgres/full_store.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+	defer release()
+
+	tx, err := conn.Begin(ctx)
+	if err != nil {
+		return err
+	}
+
+	cves := make([]*storage.ClusterCVE, 0, len(cvePartsArr))
```
connorgorman, line 72: take this out of the connection acquisition?

connorgorman (approved):
LG, should we namespace cluster vulns in the future?

central/cluster/datastore/datastore_impl.go
md2119: Unfortunately no. Unless we move away from patterns like copyFrom we cannot.

md2119:
> LG, should we namespace cluster vulns in the future?

Cluster CVEs are namespaced even now. The CVE ID is CVE+CVE Type because,
- It is just easier to deal with it, APIs, presentation, etc.
- Hard requirement from the fetcher code because fetcher does passes by type

The platform cves code needs big-time refactor. The mapping should be done is Sensor returning cves (k8s,istio, etc.) for cluster, the bundle download in Scanner. etc.

central/cve/cluster/datastore/store/postgres/full_store.go
md2119: Good point

central/cluster/datastore/datastore_impl.go
connorgorman: Yeah, this is fine, but we should refactor and reconsider a bunch of the stuff related to copy/etc

PR #2262, opened 2022-06-30
fix: feature flag name for vm updates

sachaudh:
## Description

Made a little typo with the feature flag 🤦🏼 

## Checklist


dvail (approved):
I was very confused by this until I looked closer 😄

vjwilson (approved):
I did not notice before, and it really would have been OK to leave.

ui/apps/platform/src/types/featureFlag.ts
```diff
@@ -6,4 +6,4 @@ export type FeatureFlagEnvVar =
     | 'ROX_SECURITY_METRICS_PHASE_ONE'
     | 'ROX_SYSTEM_HEALTH_PF'
     | 'ROX_VERIFY_IMAGE_SIGNATURE'
-    | 'ROX_FRONTEND_VM_UDPATES';
+    | 'ROX_FRONTEND_VM_UPDATES';
```
pedrottimark, line 9: Unlike list.go file in which pattern is to add at the end, we might form habit to add in alpha order to somewhat reduce the chance of merge conflicts when velocity increases (especially to reduce friction when backend engineers need to add feature flags).

PR #2265, opened 2022-06-30
Upstream release automation: fix version parsing

0x656b694d:
## Description

When parsing a given version string with `sed` there is a problem to store the output in `bash` variables.
This PR suggests replacing the `bash` parsing with a Python script.

## Testing Performed

```
$ cat .github/workflows/scripts/parse-version.py | python - 0.3.4-name-rc.99

::set-output name=release::0.3
::set-output name=patch::4
::set-output name=name::name
::set-output name=rc::99
::set-output name=next-rc::100
```
```
$ cat .github/workflows/scripts/parse-version.py | python - 0.3.4-rc.99     

::set-output name=release::0.3
::set-output name=patch::4
::set-output name=name::
::set-output name=rc::99
::set-output name=next-rc::100
```
```
$ cat .github/workflows/scripts/parse-version.py | python - 0.3.4-name 

::set-output name=release::0.3
::set-output name=patch::4
::set-output name=name::name
::set-output name=rc::1
::set-output name=next-rc::2
```
```
$ cat .github/workflows/scripts/parse-version.py | python - 0.3.4     

::set-output name=release::0.3
::set-output name=patch::4
::set-output name=name::
::set-output name=rc::1
::set-output name=next-rc::2
```
```
$ cat .github/workflows/scripts/parse-version.py | python - a.1.2
$ echo $?
1
```

### DIFF
--- a/.github/workflows/variables.yml
+++ b/.github/workflows/variables.yml
@@ -50,7 +50,25 @@ jobs:
     steps:
       - id: parse
         env:
-          REGEX: 's/^([0-9]+\.[0-9]+)\.([0-9]+)(-([[:alnum:]]+(-[[:alnum:]]+)*))?(-rc\.([0-9]+))?$/\1 \2 \4 \7/p'
+          SCRIPT: |
+            import re
+            import sys
+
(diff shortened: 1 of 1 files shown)


vikin91 (approved):
Oh, so you just used named groups :) It is okay. Thanks for encapsulating it in the separate step and not mixing with the rest.

PR #2267, opened 2022-06-30
Cleanup and some communications improvements

0x656b694d:
## Description

* New job that prints the arguments with which the workflow has been started;
* Fix for the dry-run input parameter handling;
* Workflow-wide `GITHUB_TOKEN` variable to reduce noise.

## Testing Performed

Manual testing in `stackrox/test-gh-actions` repository.

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -23,10 +23,26 @@ env:
   jira_projects: ROX, RS, RTOOLS
   slack_channel: C03KSV3N6N8 #test-release-automation
   script_url: /repos/${{github.repository}}/contents/.github/workflows/scripts/common.sh?ref=${{github.event.repository.default_branch}}
-  DRY_RUN: ${{ fromJSON('["true", "false"]')[inputs.dry-run != 'true'] }}
+  DRY_RUN: ${{ fromJSON('["true", "false"]')[github.event.inputs.dry-run != 'true'] }}
   ACCEPT_RAW: "Accept: application/vnd.github.v3.raw"
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
+  run-parameters:
+    name: Run parameters
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          [ "$DRY_RUN" = "true" ] && echo "::warning::This is a dry run"
+          echo "Event: ${{github.event_name}}" >>"$GITHUB_STEP_SUMMARY"
+          if [ "${{github.event_name}}" = "workflow_dispatch" ]; then
+            cat <<EOF >>"$GITHUB_STEP_SUMMARY"
+          \`\`\`
+          ${{toJSON(inputs)}}
+          \`\`\`
+          EOF
+          fi
+
   variables:
     name: Setup variables
     uses: ./.github/workflows/variables.yml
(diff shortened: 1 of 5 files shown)


.github/workflows/cut-rc.yml
```diff
@@ -23,10 +23,26 @@ env:
   jira_projects: ROX, RS, RTOOLS
   slack_channel: C03KSV3N6N8 #test-release-automation
   script_url: /repos/${{github.repository}}/contents/.github/workflows/scripts/common.sh?ref=${{github.event.repository.default_branch}}
-  DRY_RUN: ${{ fromJSON('["true", "false"]')[inputs.dry-run != 'true'] }}
+  DRY_RUN: ${{ fromJSON('["true", "false"]')[github.event.inputs.dry-run != 'true'] }}
```
vikin91, line 26: Could you maybe explain this a bit? Do you store the negative value of `inputs.dry-run` in `DRY_RUN`?

Could we think of a positive logic here - i.e., using `=` instead of `!=`?

For example, would this be possible?

```suggestion
  DRY_RUN: ${{ fromJSON('["true", "false"]')[github.event.inputs.dry-run = 'true'] }}
```

.github/workflows/cut-rc.yml
```diff
@@ -23,10 +23,26 @@ env:
...
   ACCEPT_RAW: "Accept: application/vnd.github.v3.raw"
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
+  run-parameters:
+    name: Run parameters
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          [ "$DRY_RUN" = "true" ] && echo "::warning::This is a dry run"
```
vikin91, line 36: `[ "$DRY_RUN" = "true" ]` reads nice!

.github/workflows/cut-rc.yml
```diff
@@ -102,7 +115,7 @@ jobs:
           echo "The following PRs are still open: $PRS"
 
           for PR in $PRS; do
-            [ "${{inputs.dry-run}}" != "true" ] && \
+            [ "$DRY_RUN" != "true" ] && \
```
vikin91, line 118: Here, `[ "$DRY_RUN" = "false" ]` would read nicer. Would it be possible?

.github/workflows/cut-rc.yml
0x656b694d: `dry-run` is passed as a string with 3 possible values: 'true', 'false' or ''.
'false' or 'empty' are functionally equivalent, meaning it is a wet run, not a dry one.
Since we don't have much freedom in ${{ }} expressions, the trick is to use JSON list indexes (0 or 1) to evaluate conditions.

The logic here is to have either `'true'` or `'false'` in `DRY_RUN`, and never have it empty.

`dry-run != 'true'` evaluates to 0 for a dry and to 1 otherwise. Therefore `DRY_RUN` holds `'true'` for 0 and `'false'` for 1, essentially copying the value from `dry-run` and evaluating `'false'` if `dry-run` is empty.

Oh, hope this helps!

.github/workflows/cut-rc.yml
0x656b694d: Ah, indeed! Thanks to the trick above, `DRY_RUN` must not be empty!

PR #2268, opened 2022-06-30
Force chart axis format to string to avoid floating point interpolation

dvail:
## Description

This forces the `tickFormat` axis on charts that expect an integer to be a string. Allowing a JS `number` type to be passed causes the tick label to display as a float due to the chart library's interpolation method (but only when the axis contains certain values).

This does not impact the Compliance widget as those tick values are statically defined.

## Checklist

## Testing Performed

This was difficult to pinpoint, but the only way I could reliably get the incorrect format to appear is when the maximum tick value on the axis was 6 or 7. (Hypothesis is that this only happens when the tick range is small enough that every tick value is one more than the last.) Notice the axis on the 2nd and 3rd widgets below.

Before:
image

After:
image


pedrottimark (approved):
Super.

PR #2274, opened 2022-06-30
CI: Fix nightlies III

gavin-stackrox:
## Description

Yesterdays fix did not take because `make cli` attempts to `chmod u+w $(GOPATH)/bin/roxctl` which is [not permitted](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-stackrox-stackrox-master-nightly-gke-qa-e2e-tests/1542372629149650944) in the test context (though it is permitted in the image context and `cp bin/$(HOST_OS)/roxctl $(GOPATH)/bin/roxctl` is fine so /shrug).

Also noted in last nights CI runs is that the roxctl generated in `test-bin` does in fact have the proper nightly tag. My working theory is that the CCI trigger has created the tag before OSCI runs and that when they run that tag is at HEAD. (It would be tempting to rely on this and avoid the hacking in prior fixes but relying on different cron jobs to behave in any sort of order notwithstanding external changes is a 💩 IMHO). And it appears as if JOB_NAME_SPEC is defined in the `test-bin` context. So this PR also adds a bunch of defensive checks when building roxctl and some general debug.

## Checklist

## Testing Performed

CI is sufficient - Forced the code path with commit 0f381d6. Ideally I would also trigger a nightly run but there is no sane way to do that with prow.

### DIFF
--- a/scripts/ci/jobs/test-binary-build-commands.sh
+++ b/scripts/ci/jobs/test-binary-build-commands.sh
@@ -8,7 +8,7 @@ set -euo pipefail
 make_test_bin() {
     info "Making test-bin"
 
-    make cli upgrader
+    make cli-build upgrader
     install_built_roxctl_in_gopath
 }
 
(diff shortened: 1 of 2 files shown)


sbostick (approved):
/lgtm

PR #2275, opened 2022-06-30
Fix deletion of active components when deployments are deleted

connorgorman:
## Description

Use Postgres delete on cascade to handle clean up of active components. We can use it because the deployment exists for this logic specifically. 

Also, move the actual deletion of the deployment to the bottom of the datastore logic to ensure crash replay

## Checklist

## Testing Performed

CI and will stage in my scale cluster to ensure that active components stop going up


viswajithiii (approved):
Seems fine. Just to confirm, it is literally impossible for us to try to insert an active component before inserting the deployment, right? Or is that race a possibility somehow?

central/deployment/datastore/datastore_impl.go
```diff
@@ -283,16 +283,6 @@ func (ds *datastoreImpl) RemoveDeployment(ctx context.Context, clusterID, id str
...
-			return err
-		}
-		return nil
-	})
-	if err != nil {
-		return err
-	}
-
 	errorList := errorhelpers.NewErrorList("deleting related objects of deployments")
 	deleteRelatedCtx := sac.WithGlobalAccessScopeChecker(ctx,
```
md2119, line 287: Not related to your PR, but instance like these should just be allowallaccess

md2119 (approved):

PR #2276, opened 2022-06-30
Evaluate performance of using pgx.Batch

connorgorman:
## Description

Batch uses a transaction under the hood due to the Postgres protocol so let's try batching all of the requests together instead of explicit transaction begin and multiple round trips

## Checklist

## Testing Performed

CI will run testing checks and also scale

### DIFF
--- a/central/activecomponent/datastore/internal/store/postgres/store.go
+++ b/central/activecomponent/datastore/internal/store/postgres/store.go
@@ -292,6 +292,7 @@ func (s *storeImpl) upsert(ctx context.Context, objs ...*storage.ActiveComponent
 			_, err := batchResults.Exec()
 			result = multierror.Append(result, err)
 		}
+		batchResults.Close()
 		if err := result.ErrorOrNil(); err != nil {
 			return err
 		}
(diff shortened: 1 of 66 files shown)


connorgorman:
```
prometheus-metric-parser compare --old-file pg-without-client-compression/metrics-1 --new-file batched-inserts/metrics-1 | grep postgres_op_duration | grep -i upsert | grep -i "pod\|deployment\|binding"
postgres_op_duration Operation=Upsert Type=K8SRoleBinding (old: (25852.45/6939) 3.73, new (17562.39/8548) 2.05): change: -44.8540%
postgres_op_duration Operation=Upsert Type=Pod (old: (371373.78/139305) 2.67, new (178109.97/138900) 1.28): change: -51.9004%
postgres_op_duration Operation=Upsert Type=Deployment (old: (670387.98/42483) 15.78, new (132656.25/41784) 3.17): change: -79.8810%
```
This is Postgres before (actually with an optimization that this does not have for deployments) and this PR

viswajithiii (approved):
Nice

PR #2280, opened 2022-07-01
Upstream release automation: use DRY_RUN when possible (dry-run bugfix)

0x656b694d:
## Description

This PR fixes an issue with the `inputs` context not being _always_ available (not sure why). Using `github.event.inputs` instead seems to fix the issue.

The problem has been discovered during a "Start Release" dry run on `stackrox/stackrox`: the RC1 milestone [was created](https://github.com/stackrox/stackrox/runs/7140127856?check_suite_focus=true) despite the dry-run option being set.

Due to another GitHub context issue, the DRY_RUN environment variable cannot be used at the `job.<job_id>.if` level, so reverting back the use of `github.event.input` for those cases.

A little out-of-scope fix in `variables.yml`: bringing back the forgotten `dash-name` variable, which stores the release name following a dash (or empty if no name). This variable simplifies output formatting.

## Testing Performed

Manual testing on `stackrox/test-gh-actions` repository.

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -23,10 +23,26 @@ env:
   jira_projects: ROX, RS, RTOOLS
   slack_channel: C03KSV3N6N8 #test-release-automation
   script_url: /repos/${{github.repository}}/contents/.github/workflows/scripts/common.sh?ref=${{github.event.repository.default_branch}}
-  DRY_RUN: ${{ fromJSON('["true", "false"]')[inputs.dry-run != 'true'] }}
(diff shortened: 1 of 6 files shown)


.github/workflows/cut-rc.yml
```diff
@@ -273,7 +281,8 @@ jobs:
   create-os4-cluster:
     name: Create OS4 cluster
     needs: [variables, cut-rc]
-    if: inputs.dry-run != 'true'
+    # Cannot use env.DRY_RUN here. `dry-run` can be 'true', 'false' or empty.
```
vikin91, line 284: Cool explanation, thank you!

vikin91 (approved):
Looks good!

PR #2266, opened 2022-06-30
Dackbox imagecomponent and nodecomponent read tests

rhybrillou:
## Description

Adding tests for scope checks on image components and node components.
The goal is to ensure the datastore behaviour is the same before and after the migration to postgres.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Adding unit tests for dackbox imagecomponent and nodecomponent

PR #2235, opened 2022-06-28
Postgres Dackbox CVE SAC read tests

rhybrillou:
## Description

This change introduces the needed SAC changes to tests the image CVE and node CVE reads with scope checks.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
~~- [ ] Evaluated and added CHANGELOG entry if required~~
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

If any of these don't apply, please comment below.

## Testing Performed

Added tests from the old dackbox, and ensured the can run against the postgres version

PR #2281, opened 2022-07-01
ROX-11284: Map dackbox object types to existing resources

rhybrillou:
## Description

In order to generate postgres store code without introducing new resources, data types linked to image and node objects will be associated with the Image and Node resources.

## Checklist

## Testing Performed

CI run is enough for now.
Additional tests will come with PRs https://github.com/stackrox/stackrox/pull/2235/files and https://github.com/stackrox/stackrox/pull/2266


central/role/resources/list.go
```diff
@@ -38,26 +38,18 @@ var (
 	DeploymentExtension = newResourceMetadata("DeploymentExtension", permissions.NamespaceScope)
 	Detection           = newResourceMetadata("Detection", permissions.GlobalScope)
 	Image               = newResourceMetadata("Image", permissions.NamespaceScope)
-	// SAC check is not performed directly on ImageComponent resource. It exists here for postgres sac generation to pass.
-	ImageComponent = newResourceMetadata("ImageComponent", permissions.NamespaceScope)
```
dhaus67, line 42: I don't think we can remove those.
Some (`ImageComponent`) are currently defined within [roleResources.ts](https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/types/roleResources.ts) and exposed to clients in the UI within [ResourceDescription.ts](https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/AccessControl/PermissionSets/ResourceDescription.tsx).
So, technically, clients are capable of choosing these for their permission sets, simply removing them would break them and we require a migration for that.

The second thing, even though they are not defined within UI, it's still possible to use those via API, so we would break existing clients here as well.

For this, we should define them as deprecated resource (`newDeprecatedResourceMetadata`), using `resource.Image/resource.Node` as replacing resource.

Please also mention the deprecation announcement in the CHANGELOG, luckily we still can manage with the 3.71 release. Additionally, would be great if you could create a follow-up task in JIRA that will cover the creation of a migration for permission sets that contain the old resources.

tools/generate-helpers/pg-table-bindings/list.go
```diff
@@ -30,7 +30,9 @@ func init() {
 		&storage.ComplianceRunMetadata{}:                        resources.Compliance,
 		&storage.ComplianceRunResults{}:                         resources.Compliance,
 		&storage.ExternalBackup{}:                               resources.BackupPlugins,
+		&storage.ImageComponent{}:                               resources.Image,
```
dhaus67, line 33: This should be still the old, to-be-deprecated resources.

This will allow existing clients to use the old resource as well as the new resource. Only specifying `resource.Image/resource.Node` here will lead to the old resource as not being valid, since `resources.Image/Node` doesn't know about resource it replaces.

dhaus67 (changes requested):

tools/generate-helpers/pg-table-bindings/list.go
rhybrillou: The postgres stores were added after 3.70. IMHO, they should directly be plugged on the new resource.

dhaus67: Fine by me!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - New permission `DeploymentExtension` will deprecate the permissions `Indicator, NetworkBaseline, ProcessWhitelist, Risk`.
   - New permission `Integration` will deprecate the permissions `APIToken, BackupPlugins, ImageIntegration, Notifier, SignatureIntegration`.
   Each deprecated permission will be removed in a future release.
+- Permission `ImageComponent` is deprecated and will be replaced by permission `Image`. Similar to the permission changes introduced with ROX-8520, `ImageComponent` will be removed in a future release.
```
dhaus67, line 25:
```suggestion
- Permission `ImageComponent` is deprecated and will be superseded by the existing permission `Image`. Similar to the permission changes introduced with ROX-8520, `ImageComponent` will be removed in a future release.
```

dhaus67 (approved):
Minor comment regarding the changelog entry, otherwise LGTM 🚀

rukletsov (approved):

janisz (approved):

PR #2286, opened 2022-07-01
Wrap exact query values individually in dbl quotes instead of as a single string

dvail:
## Description

The function used to create query strings in VulnMgmt will wrap queries for any search key that ends in ` id` with double quotes. However if the search query is an array, instead of wrapping each individual term in quotes it will create a single comma delimited string and wrap the entire thing in a single set of double quotes. This was introduced [quite a while ago](https://github.com/stackrox/rox/pull/3192/commits/678091d14f0b7a2052947cfd088337896fca0fb2#diff-6e11ac39ca0f153e15b0e72560c6bd563c7a3889fca060d5c059415316384c52R9) but I can't really tell what the context is. Most of the UI in the [screenshots on that PR](https://github.com/stackrox/rox/pull/3192) is long gone. This PR changes the behavior to wrap the array of items individually in double quotes:

```
Namespace:"payments,frontend,stackrox"
> becomes
Namespace:"payments","frontend","stackrox"
```

My hunch is that this was just an oversight that didn't account for queries with multiple "ID" values, as there was a test case added (also gone in current code) that only [checked for the single case](https://github.com/stackrox/rox/pull/3192/files#diff-a9de3a6e284d727d68acc071ae38fc935cb2d596e79077ca81fc13a6ffa87552).

This should fix https://issues.redhat.com/browse/ROX-11510 and the links on the new dashboard that end up on VulnMgmt pages.

The weird thing is that this quoted query string does appear to return results. It gives results as if it is checking each item in the array _except_ for the first and last item in quotes. e.g. It would give query results for `"frontend"` only using the examples above. Background discussion on Slack https://srox.slack.com/archives/C01DQH7UP8C/p1656702347490589.

I think this is a bug on the frontend and a ~separate bug on the backend~ that are rare enough that no-one has noticed until this specific use case. _Not a backend bug, this is intended behavior as per the Slack convo above._

## Checklist

## Testing Performed

Visit the dashboard and apply multiple namespace filters:

image
image

Click the `>1 year` link and go to the VulnMgmt images page. This page should contain the same number of images as the bar in the chart.
image

The old behavior would not display 73 images, and show a reduced number (47 images):
image

If we go back to the dashboard on this branch, and deselect the first and last namespaces we picked in the dropdown:
image
image

### DIFF
--- a/ui/apps/platform/src/utils/queryService.js
+++ b/ui/apps/platform/src/utils/queryService.js
@@ -42,10 +42,14 @@ function objectToWhereClause(query, delimiter = '+') {
             if (typeof value === 'undefined' || value === '') {
                 return acc;
             }
-            const flatValue = Array.isArray(value) ? value.join() : value;
-            const needsExactMatch =
-                key.toLowerCase().indexOf(' id') !== -1 && value.indexOf(',') === -1;
-            const queryValue = needsExactMatch ? `"${flatValue}"` : flatValue;
+            const valueArray = Array.isArray(value) ? value : [value];
+            const queryValue = valueArray
+                .map((val) => {
+                    const needsExactMatch =
+                        key.toLowerCase().indexOf(' id') !== -1 && val.indexOf(',') === -1;
+                    return needsExactMatch ? `"${val}"` : val;
+                })
+                .join();
             return `${acc}${key}:${queryValue}${delimiter}`;
         }, '')
         .slice(0, -delimiter.length);
--- a/ui/apps/platform/src/utils/queryService.test.js
+++ b/ui/apps/platform/src/utils/queryService.test.js
@@ -20,6 +20,32 @@ describe('queryService.', () => {
 
             expect(queryStr).toEqual('cve:CVE-2005-2541,CVE-2017-12424,CVE-2018-16402');
         });
+
+        it('correctly quote wraps exact search term values', () => {
+            const singleValueQueryObject = {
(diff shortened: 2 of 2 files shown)


dvail:
* master
  * **PR #2286** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2286" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2286?utm_source=stack-comment).

ui/apps/platform/src/utils/queryService.js
```diff
@@ -42,10 +42,14 @@ function objectToWhereClause(query, delimiter = '+') {
...
-                key.toLowerCase().indexOf(' id') !== -1 && value.indexOf(',') === -1;
-            const queryValue = needsExactMatch ? `"${flatValue}"` : flatValue;
+            const valueArray = Array.isArray(value) ? value : [value];
+            const queryValue = valueArray
+                .map((val) => {
+                    const needsExactMatch =
+                        key.toLowerCase().indexOf(' id') !== -1 && val.indexOf(',') === -1;
+                    return needsExactMatch ? `"${val}"` : val;
+                })
+                .join();
```
vjwilson, line 52: Worth adding a test case for this, but we can do that after the release is cut. OK to merge as-is.

vjwilson (approved):

PR #2288, opened 2022-07-01
ROX-11281: Update pruning to remove clusters that have been unhealthy for a configurable period of time

theencee:
## Description

- Adds a new garbage collection task to auto remove decommissioned clusters (for a configurable period of time)
- The task should run every 24 hours
- Retention value of 0 days should mean that auto remove is disabled
- Do not delete a cluster if it's the last one around
- If all clusters are unhealthy, do not delete; warn the user instead. There's something seriously wrong with their setup.

For the last item, there are a caveat:
- Finding out if a cluster has central in it, is apparently not easy. There is a todo to tag such clusters in the future. But for now, we will search for clusters with a deployment in it called "central" and then we'll validate using a label (app=central) and annotation (owner=stackrox). This may have some FPs and catch other clusters, but that's ok. We'd rather err on not deleting something than to delete cluster with central in it. 

## Checklist
- [] Investigated and inspected CI test results

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/pruning/pruning.go
+++ b/central/pruning/pruning.go
@@ -504,9 +504,11 @@ func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
 	clusterRetention := config.GetDecommissionedClusterRetention()
 	retentionDays := int64(clusterRetention.GetRetentionDurationDays())
 	if retentionDays == 0 {
-		log.Info("[Cluster Pruning] pruning is disabled")
+		// TODO: Nakul - rename from pruning in logs?
+		log.Info("[Cluster Pruning] pruning is disabled.")
 		return
 	}
+
 	lastUpdateTime, err := types.TimestampFromProto(clusterRetention.GetLastUpdated())
 	if err != nil {
 		log.Error(err)
@@ -514,7 +516,7 @@ func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
 	}
 	if timeutil.TimeDiffDays(time.Now(), lastUpdateTime) < 1 {
 		// Allow 24 hr grace period after a change in config
-		log.Info("[Cluster Pruning] skipping pruning to allow 24 hr grace period after a recent change in cluster retention config")
+		log.Info("[Cluster Pruning] skipping pruning to allow 24 hr grace period after a recent change in cluster retention config.")
 		return
 	}
 
@@ -526,6 +528,7 @@ func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
 	if timeutil.TimeDiffDays(time.Now(), configCreationTime) < int(retentionDays) {
 		// In this case, the deployments that became unhealthy after config creation would also be unhealthy for fewer than retention days
 		// and pruning can be skipped
+		log.Info("[Cluster Pruning] skipping pruning to as retention period only starts after upgrade to version 3.71.0.")
 		return
 	}
 
@@ -534,7 +537,8 @@ func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
 		log.Errorf("[Cluster Pruning] error counting clusters: %v", err)
 	}
 	if allClusterCount <= 1 {
-		log.Info("[Cluster Pruning] skipping pruning because there aren't enough clusters")
+		log.Info("[Cluster Pruning] skipping pruning because there is only one cluster which cannot be removed.")
+		return
 	}
 
 	query := search.NewQueryBuilder().
@@ -546,13 +550,26 @@ func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
 		return
 	}
 
+	log.Infof("[Cluster Pruning] found %d clusters that haven't been active in over %d days", len(clusters), retentionDays)
+
 	clustersToPrune := make([]string, 0)
 	for _, cluster := range clusters {
 		if sliceutils.MapsIntersect(clusterRetention.GetIgnoreClusterLabels(), cluster.GetLabels()) {
 			log.Infof("[Cluster Pruning] skipping excluded cluster with id %s", cluster.GetId())
 			continue
 		}
-		// TODO : Skip if the cluster is running central
+
+		// Don't delete if the cluster contains central
+		hasCentral, err := g.checkIfClusterContainsCentral(cluster)
+		if err != nil {
+			log.Errorf("[Cluster Pruning] error searching for deployements in cluster: %v", err)
+			return
+		}
+		if hasCentral {
+			// Warning because it's important to know that your central cluster is unhealthy
+			log.Warnf("[Cluster Pruning] skipping pruning cluster with id %s because this cluster contains the central deployment.", cluster.GetId())
+			continue
+		}
 		clustersToPrune = append(clustersToPrune, cluster.GetId())
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/sliceutils/map.go
```diff
@@ -51,3 +51,17 @@ func Map(slice, mapFunc interface{}) interface{} {
 	}
 	return outSlice.Interface()
 }
+
+// MapsIntersect returns true there is at least one key-value pair that is present in both maps
+// If both, or either maps are empty, it returns false
+// TODO : Implement this so that it can take map[interface{}]interface{}
```
viswajithiii, line 57: `map[interface{}]interface{}` won't help, since we almost never use them. This will be a good use case for generics though :)

viswajithiii (approved):

central/pruning/pruning.go
```diff
@@ -485,6 +494,127 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+		// Only cluster pruning if it's been at least clusterGCFreq since last run
+		return
+	}
+	defer func() {
+		lastClusterPruneTime = time.Now()
+	}()
+
+	clusterRetention := config.GetDecommissionedClusterRetention()
+	retentionDays := int64(clusterRetention.GetRetentionDurationDays())
+	if retentionDays == 0 {
```
md2119, line 512: I realized that our retention configuration is treated inconsistently. Zero value for the images means no retention (because of AddDays query)  and immediate GC whereas for alerts pruning means no GC. (@viswajithiii ?)

We should just make the retention consistent for zero value (meaning disabled).

central/pruning/pruning.go
```diff
@@ -485,6 +494,127 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+		// and pruning can be skipped
+		log.Info("[Cluster Pruning] skipping pruning as retention period only starts after upgrade to version 3.71.0.")
+		return
+	}
+
+	allClusterCount, err := g.clusters.CountClusters(pruningCtx)
+	if err != nil {
+		log.Errorf("[Cluster Pruning] error counting clusters: %v", err)
+	}
+
```
md2119, line 544: check `allClusterCount == 0`

central/pruning/pruning.go
```diff
@@ -485,6 +494,127 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+		}
+		clustersToPrune = append(clustersToPrune, cluster.GetId())
+	}
+
+	if allClusterCount == len(clustersToPrune) {
+		log.Warnf("[Cluster Pruning] skipping pruning because all %d cluster(s) are unhealthy and this is an abnormal state. Please remove any manually if desired.", allClusterCount)
+		return
+	}
+
+	if len(clustersToPrune) > 0 {
```
md2119, line 582:
```
if len(clustersToPrune) == 0 {
  return 0
}

```

central/pruning/pruning.go
```diff
@@ -485,6 +494,127 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+func (g *garbageCollectorImpl) collectClusters(config *storage.PrivateConfig) {
+	if !features.DecommissionedClusterRetention.Enabled() {
+		return
+	}
+
+	if lastClusterPruneTime.Add(clusterGCFreq).After(time.Now()) {
+		// Only cluster pruning if it's been at least clusterGCFreq since last run
+		return
+	}
+	defer func() {
```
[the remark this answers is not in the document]
theencee: It won't actually matter because if it was disabled and then you enabled it you would get 24hrs flex before the clusters get deleted anyway (See line 522).

But yea it does make sense so moving it down

central/pruning/pruning.go
theencee: Agreed on that. However it's a backwards incompatible change, so let's tackle is separately from this project?

central/pruning/pruning.go
viswajithiii: No, you don't need to have a secured cluster running in the cluster where central is. (And this will always be the case in a managed services setting, for example.)

pkg/sliceutils/map.go
viswajithiii: Starting in Go 1.18

central/pruning/pruning.go
```diff
@@ -485,6 +494,127 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+
+	clustersToPrune := make([]string, 0)
+	for _, cluster := range clusters {
+		if sliceutils.MapsIntersect(clusterRetention.GetIgnoreClusterLabels(), cluster.GetLabels()) {
+			log.Infof("[Cluster Pruning] skipping excluded cluster with id %s", cluster.GetId())
+			continue
+		}
+
+		// Don't delete if the cluster contains central
+		hasCentral, err := g.checkIfClusterContainsCentral(cluster)
```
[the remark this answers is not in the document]
theencee: It's part of the spec. 

Basically if a customer spins up ACS, something isn't working and then reaches out to support we want to make sure the cluster doesn't go down before we get a chance to help them. In general it's possible with any cluster for this to be an issue, but usually if it's not their first cluster they probably know how to fix it and it's less urgent than their main (or first) one.

central/pruning/pruning.go
theencee: Ah gotcha. I added the check in.

pkg/sliceutils/map.go
theencee: Updated comment

central/pruning/pruning.go
```diff
@@ -485,6 +494,140 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+
+	if allClusterCount == 0 {
+		log.Info("[Cluster Pruning] skipping pruning because no clusters were found.")
+		return
+	}
+
+	query := search.NewQueryBuilder().
+		AddDays(search.LastContactTime, retentionDays).
+		AddExactMatches(search.SensorStatus, storage.ClusterHealthStatus_UNHEALTHY.String()).ProtoQuery()
+	clusters, err := g.clusters.SearchRawClusters(pruningCtx, query)
```
md2119, line 557: You do not need to pull the full cluster object for ID. You can use `Search` which returns `search.Result` with ID field.

md2119 (approved):
one perf comment, otherwise lgtm

central/pruning/pruning.go
```diff
@@ -485,6 +494,140 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+
+	// As long as we find at least one deployment that matches the criteria
+	// This is meant to be tolerant of false positives while extremely intolerant of false negatives. Rather not delete a cluster than to delete one with central in it.
+	for _, d := range deploys {
+		// While this could be done via the searcher, this was moved out to reduce complexity of the search.
+		// The search might run more often, but this manual grep only happens if the cluster is unhealthy AND has >= 1 deployment with name central.
+		if d.GetLabels()["app"] == "central" && d.GetAnnotations()["owner"] == "stackrox" {
+			return true, nil
+		}
+	}
```
md2119, line 624: You can use a map query and remove the need to pull out the full deployment object.

central/pruning/pruning.go
```diff
@@ -485,6 +494,140 @@ func (g *garbageCollectorImpl) collectImages(config *storage.PrivateConfig) {
...
+		return false, err
+	}
+
+	// As long as we find at least one deployment that matches the criteria
+	// This is meant to be tolerant of false positives while extremely intolerant of false negatives. Rather not delete a cluster than to delete one with central in it.
+	for _, d := range deploys {
+		// While this could be done via the searcher, this was moved out to reduce complexity of the search.
+		// The search might run more often, but this manual grep only happens if the cluster is unhealthy AND has >= 1 deployment with name central.
+		if d.GetLabels()["app"] == "central" && d.GetAnnotations()["owner"] == "stackrox" {
+			return true, nil
```
charmik-redhat, line 622: Are deployment `name = "central" && label "app=central" exists &&  annotation "owner=stackrox" exists` necessary conditions for StackRox central deployment?

central/pruning/pruning.go
charmik-redhat: Need the cluster object here as I need to go though the cluster labels to find an intersection with ignoreLables.

PR #2292, opened 2022-07-01
Honor isDefault while creating, renaming and deleting policy categories

clickboo:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/policycategory/datastore/datastore_impl.go
+++ b/central/policycategory/datastore/datastore_impl.go
@@ -127,6 +127,8 @@ func (ds *datastoreImpl) AddPolicyCategory(ctx context.Context, category *storag
 	if category.Id == "" {
 		category.Id = uuid.NewV4().String()
 	}
+	// Any category added after startup must be marked custom category.
+	category.IsDefault = false
 
 	ds.categoryMutex.Lock()
 	defer ds.categoryMutex.Unlock()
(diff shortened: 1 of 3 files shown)


central/policycategory/datastore/datastore_impl.go
```diff
@@ -127,6 +127,8 @@ func (ds *datastoreImpl) AddPolicyCategory(ctx context.Context, category *storag
 	if category.Id == "" {
 		category.Id = uuid.NewV4().String()
 	}
+	// Any category added after startup must be marked custom category.
+	category.IsDefault = false
```
connorgorman, line 131: I think it would be better to just reject their request rather than mutating it?

central/policycategory/datastore/datastore_impl.go
```diff
@@ -158,6 +160,10 @@ func (ds *datastoreImpl) RenamePolicyCategory(ctx context.Context, id, newName s
 		return errorsPkg.Wrapf(errox.NotFound, "policy category with id '%s' does not exist", id)
 	}
 
+	if category.IsDefault {
```
connorgorman, line 163: GetIsDefault()

connorgorman (approved):
LG, see note about mutating the object

central/policycategory/datastore/datastore_impl.go
clickboo: As discussed, we do this in policies - I went back and forth whether to throw error or mutate but then consistency won :)

PR #2296, opened 2022-07-04
ROX-11070: Only allow a single default group per auth provider

dhaus67:
## Description

This PR is a follow-up from the [discussion here](https://github.com/stackrox/stackrox/pull/2075#discussion_r911410235).

The issue is that, with the changes of introducing a unique identifier not based on a composite key of a group's properties, we potentially allow clients to create an arbitrary amount of default groups.

Previously, this was not possible due to the `authProviderID`'s value being used as a key to store the group, uniquely identifying it, creating a 1:1 mapping between default group and auth provider.

This complicates things for the UI, as we would either change the current flow within the UI and allow multiple "default roles" to be assigned, as well as posing a challenge to distinguish between the different default groups.

Since this is not at all intended, we should restrict the datastore and only allow a single default group (with no key / value field set) _per auth provider ID_. This way, we achieve the same behavior related to default groups as 
beforehand and do not add increased complexity to the front end.

## Testing Performed

- Unit tests added.

### DIFF
--- a/central/group/datastore/internal/store/store_impl.go
+++ b/central/group/datastore/internal/store/store_impl.go
@@ -5,13 +5,15 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/errox"
+	"github.com/stackrox/rox/pkg/expiringcache"
 	bolt "go.etcd.io/bbolt"
 )
 
 // We use custom serialization for speed since this store will need to be 'Walked'
 // to find all of the roles that apply to a given user.
 type storeImpl struct {
-	db *bolt.DB
+	db                *bolt.DB
+	defaultGroupCache expiringcache.Cache
 }
 
 // Get returns a group matching the given properties if it exists from the store.
@@ -89,7 +91,7 @@ func (s *storeImpl) Walk(authProviderID string, attributes map[string][]string) 
 // Returns an error if a group with the same properties already exists.
 func (s *storeImpl) Add(group *storage.Group) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
-		return addInTransaction(tx, group)
+		return addInTransaction(tx, group, s.defaultGroupCache)
 	})
 }
 
@@ -97,7 +99,7 @@ func (s *storeImpl) Add(group *storage.Group) error {
 // Returns an error if a group with the same properties does not already exist.
 func (s *storeImpl) Update(group *storage.Group) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
-		return updateInTransaction(tx, group)
+		return updateInTransaction(tx, group, s.defaultGroupCache)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


dhaus67:
* master
  * **PR #2062** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2062" target="_blank">Graphite</a> 
      * **PR #2075** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2075" target="_blank">Graphite</a> 
      * **PR #2160** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2160" target="_blank">Graphite</a> 
        * **PR #2066** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2066" target="_blank">Graphite</a> 
        * **PR #2296** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2296" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2296?utm_source=stack-comment).

central/group/datastore/internal/store/store_impl.go
```diff
@@ -265,3 +310,37 @@ func getPossibleGroupProperties(authProviderID string, attributes map[string][]s
...
+// Otherwise, it will return true.
+func checkDefaultGroupForProps(tx *bolt.Tx, props *storage.GroupProperties,
+	defaultGroupCache expiringcache.Cache) (bool, error) {
+	// 1. Short-circuit if the props do not indicate a default group. A default group only has the auth provider ID
+	// field set.
+	if !isDefaultGroup(props) {
+		return false, nil
+	}
+	// 2. Short-circuit if the auth providerID is in the cache.
+	if defaultGroupCache.Get(props.GetAuthProviderId()) != nil {
```
rukletsov, line 326: Is this step for performance reasons? If so, I'd not bother and simplify it away, here is why:
* complexity of a normal check is linear;
* _default_ groups are added and modified rarely, much rarely than groups are mapped to identities;
* linear check should be fine given for group mapping we now have `O(k * n)`.

central/group/datastore/internal/store/store_impl.go
```diff
@@ -148,13 +159,28 @@ func addInTransaction(tx *bolt.Tx, group *storage.Group) error {
 		return errox.InvariantViolation.CausedBy(err)
 	}
 
-	return buc.Put([]byte(id), bytes)
+	if err := buc.Put([]byte(id), bytes); err != nil {
+		return err
+	}
+
+	return nil
```
rukletsov, line 166:
```suggestion
	return buc.Put([]byte(id), bytes)
```
: ). Same for update and delete.

central/group/datastore/internal/store/store_test.go
```diff
@@ -509,3 +510,46 @@ func (s *GroupStoreTestSuite) TestGetFiltered() {
...
+	s.NoError(s.sto.Add(defaultGroup))
+
+	// 2. Adding the group a second time should not work and yield a "AlreadyExists" error.
+	defaultGroup.GetProps().Id = "some-id-2"
+	err := s.sto.Add(defaultGroup)
+	s.Error(err)
+	s.ErrorIs(err, errox.AlreadyExists)
+
+	// 3. Updating the initially existing group to make it a default group should fail.
+	//		Fetch the group by its properties.
```
rukletsov, line 545: That's an interesting styling approach : ).

rukletsov (approved):

central/group/datastore/internal/store/store_test.go
dhaus67: Learnt it from the best 😜

PR #2301, opened 2022-07-05
Remove 'hidden' restriction on NamespaceID search field for alert protos

dvail:
## Description

This removes the `hidden` restriction on the Namespace ID search field for alerts in order to fix https://issues.redhat.com/browse/ROX-11234. Before this change, links from the new dashboard to the "Violations" page would not populate the search input with "Namespace ID" parameters, even though they were present in the page URL. This caused the initial search to work correctly, but any change to the search parameters via the UI would clear all "Namespace ID" values transparently to the user.

I just removed the "hidden" tag on this field and ran `make proto-generated-srcs`. If there is more (or less) to it please let me know.

## Checklist

## Testing Performed

Visit the new dashboard and apply multiple cluster and namespace filters via the dropdowns at the top of the page.

Click one of the links in the "violations by severity" widget that take you to the VIolation page.

The selected Namespace IDs should be visible in the Violations page, and the "Namespace ID" option should be returned from the search options API.
image
image


md2119:
@dvail Any particular reason why namespace ID is used instead of namespace name?

dvail:
@md2119 We need to use Namespace ID as it is the only way we can uniquely identify a specific cluster/namespace combination when the user has the option to select multiple clusters as well as multiple namespaces within those clusters as their chosen scope.

Initially we wanted to use cluster and namespace names, but the search API doesn't allow us to scope a namespace name to a specific cluster.

e.g. Using the staging environment, suppose the user selects cluster 'production' with namespaces 'frontend' and 'backend', as well as cluster 'security' with namespace 'stackrox'. This would translate into a search query string of `Cluster:production,security+Namespace:frontend,backend,stackrox` which includes the following namespaces:
```
production/frontend   # ok
production/backend   # ok
production/stackrox   # included, but should not be
security/frontend   # does not exist
security/backend   # does not exist
security/stackrox   # ok
```

In the future UI components will use the namespace id behind the scenes to make queries, while displaying the descriptive name to the user like is done in the new dropdowns on this dashboard. We also discussed adding a new search field to the backend like "Namespace Location" which took search values in the form "<cluster name>/<namespace name>", but decided the namespace id solution would be less work and acceptable short term until older UI sections migrated to PatternFly.

md2119:
Thanks @dvail for the splendid clarification! This calls for the need to rethink search (flexibility) in UI (as in ANDs + ORs grouping).

dvail:
@md2119 do you think this is the right approach, either long term or short term?

My thoughts were that we would want to keep search parity between what you get in the UI and what you get via scripting the API. If we consider this an implementation detail though, and keep "Namespace ID" hidden from API users, I can implement a fix in the UI that will satisfy https://issues.redhat.com/browse/ROX-11234 for the dashboard.

md2119 (approved):
This is fine for now although I find this a weird because it is impossible to correlate the namespace ID to the namespace name in the current UX from the UI. 

This is a good opportunity to rethink UI search filters in general. Un-hiding namespace ID is a workaround, but we need a real solution. I propose we start discussing sooner than later.

dvail:
Thanks @md2119 ! I added a task to the follow on epic so that we don't lose track of the discussion regarding search and UI/UX https://issues.redhat.com/browse/ROX-11658

PR #2303, opened 2022-07-05
Update pf charts to use 'cursor: pointer' by default for bars and legends.

dvail:
## Description

Sets the default styles for PatternFly charts to show a pointer on hover for bars and legends. Since all instances of the charts currently have linkable bars and clickable legends this style was made the default, but can be overridden on a case-by-case basis if necessary.

## Checklist

## Testing Performed

Load the new dashboard and hover over the chart bars for the following widgets and ensure that the cursor style changes to a pointer, indicating a clickable object:
- Aging images
- Policy Violations by Category
- Compliance by standard

Additionally, hover over the legend items in the Policy Violations by Category widget and ensure the cursor style changes to a pointer.


vjwilson (approved):
/lgtm

PR #2307, opened 2022-07-05
upgrade-test: use helm for install and upgrade

gavin-stackrox:
## Description

The upgrade test is [failing](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-gke-upgrade-tests/1544783120299986944) because:
- It deploys the latest scanner with an earlier rox.
- This is due to persisted latest scanner envs in `setup_deployment_env()`.
- The test in general does not care about scanner upgrade and so this might have been fine.
- But the earlier version of the UpgradeTest inherits from a BaseSpec that does a scanImage().
- And the latest scanner is not starting due to an incompatibility with the earlier configuration yaml.

This PR:
- removes the persistent scanner version env settings
- uses helm for the earlier rox deploy
- and uses helm for the upgrade so that scanner (and presumably other components) follow a supported upgrade path.

## Checklist

## Testing Performed

CI is sufficient.

/test openshift-4-qa-e2e-tests

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -56,8 +56,6 @@ setup_deployment_env() {
     ci_export MAIN_IMAGE_REPO "quay.io/$REPO/main"
     ci_export CENTRAL_DB_IMAGE_REPO "quay.io/$REPO/central-db"
     ci_export COLLECTOR_IMAGE_REPO "quay.io/$REPO/collector"
-    ci_export SCANNER_IMAGE "quay.io/$REPO/scanner:$(cat "$(git rev-parse --show-toplevel)/SCANNER_VERSION")"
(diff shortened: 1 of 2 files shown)


RTann:
@gavin-stackrox we will need to add `scanner` and `scanner-db` `2.17.4` to quay. Let me do that now and then you should rerun CI

RTann (approved):
LGTM but I also want to confirm with either @connorgorman or @viswajithiii that switching this to Helm does not change the validity of the testing environment

RTann:
Actually, I think the purpose of these tests is to ensure Central still works properly upon upgrade/downgrade, and the way we upgrade does not matter, so this should be good to go

connorgorman (approved):

viswajithiii (approved):

PR #2309, opened 2022-07-05
Support searching by counts, in addition to sorting

viswajithiii:
## Description

Supports searching by counts (via `HAVING` clause). This is not strictly necessary (as this isn't supported in DackBox), but it seemed easy enough to do.
 
## Checklist
- ~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed

Unit tests


pkg/search/postgres/query/query_functions.go
```diff
@@ -49,31 +49,38 @@ var datatypeToQueryFunc = map[walker.DataType]queryFunction{
 	// Map is handled separately.
 }
 
-func matchFieldQuery(dbField *walker.Field, field *pkgSearch.Field, value string, highlight bool, now time.Time) (*QueryEntry, error) {
-	qualifiedColName := dbField.Schema.Table + "." + dbField.ColumnName
-
+func matchFieldQuery(qualifiedColName string, dataType walker.DataType, field *pkgSearch.Field, value string, highlight bool, now time.Time, goesIntoHavingClause bool) (*QueryEntry, error) {
```
md2119, line 52: is `dataType` sql type? Maybe expand the var name?

pkg/search/postgres/query/common.go
```diff
@@ -91,5 +95,26 @@ func MatchFieldQuery(dbField *walker.Field, value string, highlight bool, now ti
 	if !ok {
 		return nil, nil
 	}
-	return matchFieldQuery(dbField, field, value, highlight, now)
+
+	qualifiedColName := dbField.Schema.Table + "." + dbField.ColumnName
```
md2119, line 99: `qualifiedColName := qualifyColumn(dbField.Schema.Table, dbField.ColumnName)`

pkg/search/postgres/query/common.go
```diff
@@ -91,5 +95,26 @@ func MatchFieldQuery(dbField *walker.Field, value string, highlight bool, now ti
...
+	dataType := dbField.DataType
+	var goesIntoHavingClause bool
+	if derivedMetadata != nil {
+		switch derivedMetadata.DerivationType {
+		case search.CountDerivationType:
+			qualifiedColName = fmt.Sprintf("count(%s)", qualifiedColName)
+			dataType = walker.Integer
+			goesIntoHavingClause = true
```
md2119, line 107: All derivation types can go into having because having is where+aggregations.

pkg/search/postgres/query/common.go
```diff
@@ -91,5 +95,26 @@ func MatchFieldQuery(dbField *walker.Field, value string, highlight bool, now ti
...
-	return matchFieldQuery(dbField, field, value, highlight, now)
+
+	qualifiedColName := dbField.Schema.Table + "." + dbField.ColumnName
+	dataType := dbField.DataType
+	var goesIntoHavingClause bool
+	if derivedMetadata != nil {
+		switch derivedMetadata.DerivationType {
+		case search.CountDerivationType:
+			qualifiedColName = fmt.Sprintf("count(%s)", qualifiedColName)
+			dataType = walker.Integer
```
md2119, line 106: make `dataType` property of CountDerivationType?

pkg/search/postgres/query/common.go
```diff
@@ -91,5 +95,26 @@ func MatchFieldQuery(dbField *walker.Field, value string, highlight bool, now ti
...
+			goesIntoHavingClause = true
+		default:
+			return nil, errors.Errorf("unsupported derivation type %s", derivedMetadata.DerivationType)
```
md2119, line 109: Just a note that I added another derivation type after this PR was opened.

pkg/search/postgres/query/common.go
viswajithiii: Unfortunately that function is in a different package.

pkg/search/postgres/query/common.go
viswajithiii: No, I think it belongs here. This is SQL implementation detail-specific, while the fields in `search/options.go` are intended to be descriptive metadata that don't have any implementation information.

pkg/search/postgres/query/common.go
viswajithiii: Hmm, I'm not so sure that's true. We can figure that out when we add support for other derivation types, I guess?

pkg/search/postgres/query/common.go
viswajithiii: Yeah, I saw that. I think it's going to be harder to support priorities for selects... in either case, I'll update this message to say "unsupported derivation type in select"

pkg/search/postgres/query/query_functions.go
viswajithiii: Yup, done.

connorgorman (approved):

md2119 (approved):

PR #2315, opened 2022-07-06
ROX-11632: The “Most Common Vulnerabilities” widget should be renamed to “Most Common Image Vulnerabilities” and scoped to Image CVEs

sachaudh:
## Description

In accordance with the changes to VM post-splitting the CVEs, we need to update the `Most Common Vulnerabilities` widget to be the `Most Common Image Vulnerabilities`. The links within the widget should now send the user to the `Image CVE` entity page rather than the `CVE` entity page when the `ROX_FRONTEND_VM_UDPATES` feature flag is turned on

## Checklist

## Testing Performed

Screen Shot 2022-07-05 at 8 25 12 PM
Screen Shot 2022-07-05 at 8 25 26 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/widgets/MostCommonVulnerabilities.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/widgets/MostCommonVulnerabilities.js
@@ -78,7 +78,7 @@ const MostCommonVulnerabilities = ({ entityContext, search, limit }) => {
 
     const queryObject = { ...entityContextObject, ...search }; // Combine entity context and search
     const query = queryService.objectToWhereClause(queryObject); // get final gql query string
-    const GQL_QUERY = showVMUpdates
+    const queryToUse = showVMUpdates
         ? MOST_COMMON_IMAGE_VULNERABILITIES
         : MOST_COMMON_VULNERABILITIES;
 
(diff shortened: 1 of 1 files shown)


sachaudh:
@vjwilson @md2119 Would the `x-axis` need to change as well or does it stay the same?

ui/apps/platform/src/Containers/VulnMgmt/widgets/MostCommonVulnerabilities.js
```diff
@@ -53,16 +71,22 @@ const processData = (data, workflowState) => {
...
 
 const MostCommonVulnerabilities = ({ entityContext, search, limit }) => {
+    const { isFeatureFlagEnabled } = useFeatureFlags();
+    const showVMUpdates = isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES');
+
     const entityContextObject = queryService.entityContextToQueryObject(entityContext); // deals with BE inconsistency
 
     const queryObject = { ...entityContextObject, ...search }; // Combine entity context and search
     const query = queryService.objectToWhereClause(queryObject); // get final gql query string
+    const GQL_QUERY = showVMUpdates
```
vjwilson, line 81: nit: I think ALL_CAPS should be reserved for true constants.

Not worth changing, since this code should only live for a couple of months, unless there are other changes needed.

```suggestion
    const queryToUse = showVMUpdates
```

md2119 (approved):
Make sure CVE name is used as display name in tool tip and not ID.

PR #2318, opened 2022-07-06
Convert cypress test to RTL for Violations by category widget

dvail:
## Description

Converts flaky Cypress test for dashboard widget to use RTL.

This was the first widget developed and tests were originally done in Cypress, but moved to RTL for the other widgets. Since we are mocking requests in either case, I've found RTL to be less overhead and more reliable to test the core widget functionality in isolation.

## Checklist

## Testing Performed

Automated test change only.


pedrottimark (approved):
Wow, you are writing good examples of unit tests for graphical components.

PR #2320, opened 2022-07-06
Test out mocking PF react-charts to disable animation.

dvail:
## Description

Potential fix for ui-unit-test flakes.

The error from these seems to originate from animation in the charts. Whenever there is one or more `act` errors thrown in prow at least one stack trace ends at `at VictoryChart (/go/src/github.com/stackrox/stackrox/ui/node_modules/victory-chart/lib/victory-chart.js:47:50)`, which contains the line that calls a `useAnimation` hook in the Victory library. This hook sets React state, and it looks like it does so async after a period of time outside of RTL's `act` call, which causes the error. Thanks @pedrottimark for the insight on this!

This is a hypothesis so far, since the tests never fail locally and only seem to fail intermittently in CI.

## Checklist

## Testing Performed

Automated

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
@@ -39,6 +39,15 @@ const mocks = [
     },
 ];
 
+jest.mock('@patternfly/react-charts', () => {
+    const { Chart, ...rest } = jest.requireActual('@patternfly/react-charts');
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return {
+        ...rest,
+        Chart: (props) => <Chart {...props} animate={undefined} />,
+    };
+});
+
 jest.mock('hooks/useResizeObserver', () => ({
     __esModule: true,
     default: jest.fn().mockImplementation(jest.fn),
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
@@ -80,6 +80,15 @@ const mocks = [
     },
 ];
 
+jest.mock('@patternfly/react-charts', () => {
+    const { Chart, ...rest } = jest.requireActual('@patternfly/react-charts');
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return {
+        ...rest,
+        Chart: (props) => <Chart {...props} animate={undefined} />,
+    };
+});
+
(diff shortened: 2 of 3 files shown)


dvail:
There have been two prow/ui-unit-test runs in a row without `act` errors. Hopefully this is an indication of a fix and not good luck on intermittent failures...

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
```diff
@@ -39,6 +39,15 @@ const mocks = [
     },
 ];
 
+jest.mock('@patternfly/react-charts', () => {
```
dvail, line 42: The duplication is obviously not ideal and is here as the quickest path to fixing CI.

Two possible better solutions if this fix is reliable:

1. Centralize these mocks via https://jestjs.io/docs/manual-mocks . The downside here is that it will be fairly undiscoverable that this library is being mocked during tests. I'm also not sure what needs to be done to implement this in our monorepo, as it seems to assume a single package.json.

2. Expose the animation prop on the individual dashboard components, and mock the components instead of PF react-charts. This feels a little bit like making the code fit the tests, but might be worthwhile as a cleaner fix.

pedrottimark: 1. Agree with the downside. Also might prevent testing with actual animation, if there was a need?
2. Agree that even an optional prop opens the door to mistake in application code.
3. What do you think about factoring out `jest.mock('@patternfly/react-charts', () => { … })` as a helper function so unit tests which need it have a visible call.

pedrottimark (approved):
Thank you for following through. Unit tests were a good call.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
dvail: 3 is my ideal approach, but I'm not sure it is possible with `jest.mock`. From what I was reading, jest will automatically pull `jest.mock` calls to the top of the test file so that they run before imports. I'm assuming the magic jest does will make putting these in a helper function not possible, but I'll give it a shot and hope I am wrong!

pedrottimark: Aha, good point. Maybe `jest.mock('@patternfly/react-charts', mockChartWithoutAnimation);`

and then I am not sure whether with or without `jest` as argument?

```js
export function mockChartWithoutAnimation(jest) {
{
    const { Chart, ...rest } = jest.requireActual('@patternfly/react-charts');
    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
    return {
        ...rest,
        Chart: (props) => <Chart {...props} animate={undefined} />,
    };
})}
```

PR #2328, opened 2022-07-06
Update the proto in edge schema

md2119:
## Checklist


proto/storage/cve.proto
```diff
@@ -23,6 +23,9 @@ enum VulnerabilitySeverity {
     CRITICAL_VULNERABILITY_SEVERITY  = 4;
 }
 
+// ******************************
+// This proto is deprecated.
+// ******************************
```
viswajithiii, line 28: I think a comment in this format will help our proto parser tools pick this up
```suggestion
// Deprecated: Use the ClusterCVE, ImageCVE and NodeCVE variants instead.
```

PR #2336, opened 2022-07-07
Dashrews/ROX-11660 dedupe process baseline

dashrews78:
## Description

When building an on demand process baseline, the process list was duplicated.  used a map to build the process list to ensure that we do not have duplicates in this case going forward.  There was not an issue when the baseline was built organically at the end of the observation window.

## Checklist

## Testing Performed

Tested locally with an nginx deployment that I would scale up to multiple replicas.  I set the observation window to 10 minutes.  I requested to look at the nginx baseline before the observation period was over.  Prior to the fix these steps showed the processes duplicated for each replica.  After the fix these steps showed a process list that had no duplicates.

Additionally I added a unit test to drive the condition.

### DIFF
--- a/central/processbaseline/datastore/datastore_impl.go
+++ b/central/processbaseline/datastore/datastore_impl.go
@@ -399,15 +399,15 @@ func (ds *datastoreImpl) CreateUnlockedProcessBaseline(ctx context.Context, key 
 		return nil, err
 	}
 
-	// Build the elements for the baseline
-	elements := make([]*storage.BaselineItem, 0, len(baselineList))
+	// Build the de-duped elements for the baseline
+	elements := make(map[string]*storage.BaselineItem, len(baselineList))
 
 	for _, indicator := range baselineList {
 		baselineItem := processBaselinePkg.BaselineItemFromProcess(indicator)
 
 		insertableElement := &storage.BaselineItem{Item: &storage.BaselineItem_ProcessName{ProcessName: baselineItem}}
 
-		elements = append(elements, insertableElement)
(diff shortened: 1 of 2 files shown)


central/processbaseline/datastore/datastore_impl.go
```diff
@@ -399,15 +399,15 @@ func (ds *datastoreImpl) CreateUnlockedProcessBaseline(ctx context.Context, key
...
+	// Build the de-duped elements for the baseline
+	elements := make(map[string]*storage.BaselineItem, len(baselineList))
 
 	for _, indicator := range baselineList {
 		baselineItem := processBaselinePkg.BaselineItemFromProcess(indicator)
 
 		insertableElement := &storage.BaselineItem{Item: &storage.BaselineItem_ProcessName{ProcessName: baselineItem}}
 
-		elements = append(elements, insertableElement)
+		elements[baselineItem] = insertableElement
```
connorgorman, line 410: Wdyt about sorting alphabetically to provide a consistent output

dashrews78: @connorgorman I guess that depends.  Looks like the UI is sorting them when they are presented, so not a lot of value in doing it for our UI customers.  they weren't previously sorted in the backend so they would be in whatever order they arrived.  We could sort them, but we would probably want to do that right before calling Upsert.  I'm not certain how valuable that is or would be as part of this fix.  Up to you.

connorgorman: Works for me

connorgorman (approved):

PR #2341, opened 2022-07-07
Write frontend policyCategory service for UI

pedrottimark:
### Goals

* Increase type safety with compile-time checks.
* Simplify maintenance of consistent backend and frontend response types.

### Overview

1. Compare to https://github.com/stackrox/stackrox/blob/master/proto/api/v1/policy_category_service.proto
    * Use naming convention camelCase in TypeScript for underscore case in proto for **file names** and **properties**.
    * Use `getPolicyCategory` instead of change to `fetchPolicyCategory` function name to reduce unneeded barriers for contributions from backend engineers.
2. Compare to https://github.com/stackrox/stackrox/blob/master/proto/storage/group.proto
    * Backend has started to separate response type from storage type, do not create types/policyCategory.proto.ts

### Questions

1. Backend request for `getPolicyCategories` can have a query string. ~If UI needs to filter categories, then it is possible to add it.~ Added comment about unused ability, because UI filters without additional requests.
    * See https://github.com/stackrox/stackrox/blob/master/proto/api/v1/policy_category_service.proto#L49
2. ~Backend request for `renamePolicyCategory` has empty response. What are pro and con from UI viewpoint for response to be updated policy category?~ Rebased with #2370 and replaced `Empty` with `PolicyCategory`
    * See https://github.com/stackrox/stackrox/blob/master/proto/api/v1/policy_category_service.proto#L64

## Checklist

## Testing Performed

Added temporary request calls to PoliciesPage.tsx file:

```js
import {
    deletePolicyCategory,
    getPolicyCategory,
    getPolicyCategories,
    postPolicyCategory,
    renamePolicyCategory,
} from 'services/policyCategoryService';
```

```js
    useEffect(() => {
        postPolicyCategory({ id: '', name: 'policy category name', isDefault: false })
            .then(({ id }) => {
                getPolicyCategory(id).catch(() => {});
                getPolicyCategory(id.slice(2)).catch(() => {});
                renamePolicyCategory(id, 'renamed policy category').catch(() => {});
                getPolicyCategories().catch(() => {});
                deletePolicyCategory(id).catch(() => {});
                getPolicyCategories().catch(() => {});
            })
            .catch(() => {});
    });
```

1. `yarn build` in ui
2. `export ROX_NEW_POLICY_CATEGORIES=true` in ui
3. `yarn deploy-local` in ui
4. `yarn start` in ui

    * POST /v1/policycategories 200 when feature flag enabled and 404 when disabled
    * GET /v1/policycategories/correctId 200
    * GET /v1/policycategories/incorrectId 404
    * PUT /v1/policycategories/correctId 200
    * GET /v1/policycategories 200
    * DELETE /v1/policycategories/correctId 200
    * GET /v1/policycategories 200

### DIFF
--- a/ui/apps/platform/src/services/policyCategoryService.ts
+++ b/ui/apps/platform/src/services/policyCategoryService.ts
@@ -32,14 +32,9 @@ export function postPolicyCategory(policyCategory: PolicyCategory): Promise<Poli
         .then((response) => response.data);
 }
 
-/*
- * Because the response is empty:
- * For success: UI is responsible to update the name property of the renamed policyCategory object.
- * For failure: UI is responsible not to update the name property of the policyCategory object.
- */
-export function renamePolicyCategory(id: string, newCategoryName: string): Promise<Empty> {
+export function renamePolicyCategory(id: string, newCategoryName: string): Promise<PolicyCategory> {
     return axios
-        .put<Empty>(`${policyCategoriesUrl}/${id}`, { newCategoryName })
+        .put<PolicyCategory>(policyCategoriesUrl, { id, newCategoryName })
         .then((response) => response.data);
 }
 
--- a/ui/apps/platform/src/services/policyCategoryService.ts
+++ b/ui/apps/platform/src/services/policyCategoryService.ts
@@ -1,0 +1,50 @@
+import axios from './instance';
+
+const policyCategoriesUrl = '/v1/policycategories';
+
+export type PolicyCategory = {
+    id: string;
+    // central/policycategory/service/service_impl.go
+    // policy category must have a name between 5 and 128 characters long with no new lines or dollar signs
+    name: string;
+    isDefault: boolean;
+};
+
+export function getPolicyCategory(id: string): Promise<PolicyCategory> {
+    return axios
+        .get<PolicyCategory>(`${policyCategoriesUrl}/${id}`)
+        .then((response) => response.data);
+}
+
+export function getPolicyCategories(): Promise<PolicyCategory[]> {
+    return axios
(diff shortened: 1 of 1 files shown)


ui/apps/platform/src/services/policyCategoryService.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/*
+ * Because the response is empty:
+ * For success: UI is responsible to update the name property of the renamed policyCategory object.
+ * For failure: UI is responsible not to update the name property of the policyCategory object.
+ */
+export function renamePolicyCategory(id: string, newCategoryName: string): Promise<Empty> {
+    return axios
+        .put<Empty>(`${policyCategoriesUrl}/${id}`, { newCategoryName })
+        .then((response) => response.data);
```
dvail, line 43: Do you know if the backend plans to implement this so that the response contains the updated `PolicyCategory` object? If so, I wonder if it is assuming too much to return a `PolicyCategory` constructed with the `id` and `newCategoryName` passed to this function so that the caller can rely on the resolved response to update the UI. If the backend makes the change we would only need to update here instead of the call sites.

dvail (approved):
LGTM, I think having the API response type decoupled from the proto types here (no JS pun intended) is going to be a nice change.

One comment that I'm not sure that we should even do, but something worth considering.

ui/apps/platform/src/services/policyCategoryService.ts
pedrottimark: We have time to understand pro and con from backend viewpoint of Empty versus PolicyCategory as response, because the endpoints are not yet visible even with feature flag enabled.

PR #2349, opened 2022-07-07
Delete obsolete functions from policies service in UI

pedrottimark:
## Description

Follow up task that was missed after release of policies in PatternFly.

### Changed files

1. Edit Containers/Policies/Wizard/Step5/ReviewPolicyForm.tsx
    * Replace `{ data: { jobId } }` with `jobId` in response of `startDryRun` function
    * Replace `{ data: { pending, result } }` with `{ pending, result }` in response of `checkDryRun` function

2. Edit services/PoliciesService.ts
    * Delete `fetchPolicy` function
    * Delete `fetchPolicies` function
    * Delete `fetchPolicyCategories` function
    * Replace `Record<string, never>` with `Empty` type for similarity with declarations of backend services
        For example, https://github.com/stackrox/stackrox/blob/master/proto/api/v1/policy_service.proto#L171
    * Simplify return type of `createPolicy` function (note that caller does not use it, but instead goes back to policies table which gets policies)
    * Simplify return type of `startDryRun` function (and make corresponding change to ReviewPolicyForm)
    * Simplify return type of `checkDryRun` function (and make corresponding change to ReviewPolicyForm)

3. Edit services/schemas.js
    * Delete `policy`
    * Delete `deployment` because already unused
    * Delete `secret` because already unused

### Residue

1. Investigate whether it is possible to remove `normalizr` from dependencies
    * Does `fetchImages` function still need `image` schema?
    * Do `fetchClusters`, `fetchCluster`, and `saveCluster` functions still need `cluster` schema?

## Checklist

### manual testing

1. `yarn build` in ui
    * branch - master: 154 = 10390361 - 10390207 slight increase because simplifications did not quite pay for additions of `then((response) => response.data)`
3. `yarn start` in ui

    * Visit /main/policy-management/policies: see GET /v1/policies request
    * For an enabled policy, click kabob, and then click **Disable policy**: see PATCH /v1/policies/id request with `{disabled: true}` payload and `{}` response
    * For the disabled policy, click kabob, and then click **Enable policy**: see PATCH /v1/policies/id request with `{disabled: false}` payload and `{}` response
    * Click existing policy name link: see GET /v1/policies/id request
    * Click back button, click **Create policy** button, continue to click **Next** button at step 4:
        see POST /v1/policies/dryrunjob request with `{jobId}` response
        see GET /v1/policies/dryrunjob/jobId request
    * Click **Save** button: see POST /v1/policies?enableStrictValidation=true request for which `id` property has empty string value in payload and non-empty string value in response
    * Click new policy name link: see GET /v1/policies/id request
    * Click **Actions** button, click **Edit policy**, edit name, click **Review policy**, click **Save** button: see PUT /v1/policies/id request and `{}` response
    * Click **Actions** button, click **Delete policy**, and then click **Delete**: see DELETE /v1/policies/id request and `{}` response

### integration testing

1. `yarn cypress-open` in ui/apps/platform
    * policies/exportPolicy.test.js
    * policies/importPolicy.test.js
    * policies/policyWizardStep3.test.js

### DIFF
--- a/ui/apps/platform/src/services/PoliciesService.ts
+++ b/ui/apps/platform/src/services/PoliciesService.ts
@@ -1,4 +1,3 @@
-import { normalize } from 'normalizr';
 import queryString from 'qs';
 import FileSaver from 'file-saver';
 
@@ -7,44 +6,17 @@ import { addBrandedTimestampToString } from 'utils/dateUtils';
 import { transformPolicyCriteriaValuesToStrings } from 'utils/policyUtils';
 
 import axios from './instance';
-import { policy as policySchema } from './schemas';
 
 const baseUrl = '/v1/policies';
 const policyCategoriesUrl = '/v1/policyCategories';
 
-/*
- * Fetch policy summary for a given policy policyId.
- * TODO delete after policiesSagas.js has been deleted, because superseded by getPolicy
- */
-export function fetchPolicy(
-    policyId: string
-): Promise<{ response: { entities: { policy: Record<string, Policy> } } }> {
-    return axios.get<Policy>(`${baseUrl}/${policyId}`).then((response) => ({
-        response: normalize(response.data, policySchema),
-    }));
-}
+type Empty = Record<string, never>;
 
 /*
  * Get a policy. Policy is a superset of ListPolicy.
  */
 export function getPolicy(policyId: string): Promise<Policy> {
     return axios.get<Policy>(`${baseUrl}/${policyId}`).then((response) => response.data);
-}
-
-/*
- * Fetch a list of policies.
- * TODO delete after policiesSagas.js has been deleted, because superseded by getPolicies
- */
-export function fetchPolicies(filters: { query: string }): Promise<{
-    response: {
-        entities: { policy: Record<string, ListPolicy> };
-        result: { policies: ListPolicy[] };
-    };
-}> {
-    const params = queryString.stringify(filters, { arrayFormat: 'repeat' });
-    return axios.get<{ policies: Policy[] }>(`${baseUrl}?${params}`).then((response) => ({
-        response: normalize(response.data, { policies: [policySchema] }),
-    }));
 }
 
 /*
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/services/PoliciesService.ts
```diff
@@ -7,22 +6,11 @@ import { addBrandedTimestampToString } from 'utils/dateUtils';
...
- * TODO delete after policiesSagas.js has been deleted, because superseded by getPolicy
- */
-export function fetchPolicy(
-    policyId: string
-): Promise<{ response: { entities: { policy: Record<string, Policy> } } }> {
-    return axios.get<Policy>(`${baseUrl}/${policyId}`).then((response) => ({
-        response: normalize(response.data, policySchema),
-    }));
-}
+type Empty = Record<string, never>;
```
alwayshooin, line 13: should we pull this out into a separate utils for service return types? not sure if it is currently used anywhere else but it doesn't seem like it's policy specific enough to stay here.

ui/apps/platform/src/services/PoliciesService.ts
```diff
@@ -132,57 +90,45 @@ export function enableDisableNotificationsForPolicies(
...
-     * function return type: Promise<string>
-     * add method: then((response) => response?.data?.jobId)
-     */
+export function startDryRun(policy: Policy): Promise<string> {
     const transformedPolicy = transformPolicyCriteriaValuesToStrings(policy);
 
-    return axios.post(`${baseUrl}/dryrunjob`, transformedPolicy);
+    return axios
+        .post<{ jobId: string }>(`${baseUrl}/dryrunjob`, transformedPolicy)
+        .then((response) => response.data.jobId);
```
alwayshooin, line 123: should we have the ? operator here for safety? it's mentioned in the comment -- was it intentional to leave it out?

alwayshooin (approved):
looks fine to me, thanks for cleaning this up! just a few questions in the code.

ui/apps/platform/src/services/PoliciesService.ts
pedrottimark: Good question. When I wrote the comment, I had noticed extensive use of optional chaining in service functions, but had second thoughts because an infinite polling loop of `checkDryRun` calls seems possible with empty string or undefined. Therefore, failure might not be an option for the response of this request.

ui/apps/platform/src/services/PoliciesService.ts
pedrottimark: Good eyes. I will merge and we can brainstorm about pro and con of local or shared type definition.

For now, `Empty` is an experimental idiom with two goals:
* Mimic `Empty` which seems to be built into proto notation.
* Reduce prettier wrapping.

There is no `export` because callers will never import it. It tells TypeScript to report an error if a caller destructures the response (for example, a PUT method which returns empty response instead of updated object).

By the way, `.then((response) => response.data)` method call is important even (or especially) for empty responses, otherwise the function leaks the sibling properties of the response object as if they were data!

PR #2358, opened 2022-07-11
ROX-11711: various fixes on Cut RC

0x656b694d:
## Description

* Allow for triggering `create-cluster` workflow manually;
* Allow for not creating cluster on Cut RC manual run;
* Fix `CHANGELOG.md` version patch on master;
* Extracted `create-cluster.sh` script;
* Always create next milestone on Cut RC. This need to be worked further to prevent side effects of milestones closed after release is published.

## Testing Performed

TODO

### DIFF
--- a/.github/workflows/scripts/create-cluster.sh
+++ b/.github/workflows/scripts/create-cluster.sh
@@ -26,7 +26,7 @@ function cluster_status() {
 }
 
 function cluster_destroying() {
-    [ "$(cluster_status "$1")" -eq 2 ]
+    [ "$(cluster_status "$1")" -eq 3 ]
 }
 
 function infra_status_summary() {
--- a/.github/workflows/scripts/create-cluster.sh
+++ b/.github/workflows/scripts/create-cluster.sh
@@ -25,8 +25,8 @@ function cluster_status() {
     cluster_info "$1" | jq -r '.Status'
 }
 
-function cluster_creating() {
-    [ "$(cluster_status "$1")" -eq 1 ]
+function cluster_destroying() {
+    [ "$(cluster_status "$1")" -eq 2 ]
 }
 
 function infra_status_summary() {
@@ -50,6 +50,14 @@ case $(cluster_status "$CNAME") in
     # Cluster exists already.
     infra_status_summary "$CNAME" "Cluster already exists"
     exit 0
+    ;;
+3)
+    # Cluster is being destroyed.
+    infra_status_summary "$CNAME" "Cluster is being destroyed"
+    while cluster_destroying "$CNAME"; do
+        gh_log notice "Waiting 30s for the cluster '$CNAME' to be destroyed"
+        sleep 30
+    done
     ;;
 4)
     # Cluster has already been destroyed. Create it again.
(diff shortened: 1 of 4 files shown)


.github/workflows/scripts/create-cluster.sh
```diff
@@ -0,0 +1,83 @@
...
+    # Cluster exists already.
+    infra_status_summary "$CNAME" "Cluster already exists"
+    exit 0
+    ;;
+4)
+    # Cluster has already been destroyed. Create it again.
+    gh_log notice "Cluster \`$CNAME\` has been destroyed already."
+    infra_status_summary "$CNAME" "Cluster has been destroyed already"
+    ;;
+*)
```
vikin91, line 59: I didn't know that when I wrote this, but status `3` means `destroying` - not sure if it is useful here

.github/workflows/start-release.yml
```diff
@@ -178,7 +178,7 @@ jobs:
           set -uo pipefail
           gh api -H "$ACCEPT_RAW" "${{env.script_url}}" | bash -s -- \
             patch-changelog \
-            "${{inputs.version}}" \
+            "${{needs.variables.outputs.release}}.${{${{needs.variables.outputs.patch}}" \
```
vikin91, line 181: Typo?

```suggestion
            "${{needs.variables.outputs.release}}.${{needs.variables.outputs.patch}}" \
```

.github/workflows/scripts/create-cluster.sh
```diff
@@ -0,0 +1,83 @@
...
+# Creates a cluster on infra.
+#
+
+set -uo pipefail
+
+FLAVOR="$1"
+NAME="$2"
+LIFESPAN="$3"
+WAIT="$4"
+ARGS="$5"
```
vikin91, line 12: Must: You call it with: `name, wait, flavor, lifespan, args` in .github/workflows/create-cluster.yml. Here the order is different

vikin91 (changes requested):
Typo and the script call order must be corrected, the rest looks great!

.github/workflows/scripts/create-cluster.sh
0x656b694d: I'll add it to the case for general awareness.

.github/workflows/start-release.yml
0x656b694d: yes

.github/workflows/scripts/create-cluster.sh
0x656b694d: Oh. We could wait for it to be fully destroyed and then create it again.

vikin91: > Oh. We could wait for it to be fully destroyed and then create it again.

Yes, in some cases it may be useful to wait instead of failing the workflow.

.github/workflows/scripts/create-cluster.sh
```diff
@@ -25,8 +25,8 @@ function cluster_status() {
     cluster_info "$1" | jq -r '.Status'
 }
 
-function cluster_creating() {
-    [ "$(cluster_status "$1")" -eq 1 ]
+function cluster_destroying() {
+    [ "$(cluster_status "$1")" -eq 2 ]
```
vikin91, line 29: It should be `-eq 3`. 2 is status of ready

vikin91 (approved):
Cool!

PR #2361, opened 2022-07-11
Postgres data migration: code generator and fully generated migrations

c-du:
## Description

This looks big but it is not!

This change is part of data migration from legacy databases to Postgres. This PR includes the code generator and all fully generated migrations. It also contains one boltdb migration, one singleton and one dackbox(deployment) to exercise the generator. The full picture is at https://github.com/stackrox/stackrox/pull/new/cong/mig-merge2 just for your reference.

The plan:
1. Migration generator and fully generated migrations.
2. All partially generated migrations from bolt.
3. All partially generated migrations from rocksdb
4. Finish all manual aka backbox components.
5. Database hookup and e2e tests.

The inventory as planned:
[inventory.txt](https://github.com/stackrox/stackrox/files/9104719/inventory.txt)

## Checklist

## Testing Performed

Added a unit test for each migration.


dashrews78:
I assume checking in `migrator/migrations/.i.swp` was not intentional

tools/generate-helpers/pg-table-bindings/main.go
```diff
@@ -94,6 +102,15 @@ type properties struct {
...
 	SingletonStore bool
 
+	// Migration root
+	MigrateRoot string
+
+	// Where the data are migrated from in the format of "database:bucket", eg, \"rocksdb\", \"dackbox\" or \"boltdb\"")
+	MigrateFrom string
+
+	// The unique sequence number to migrate all tables to Postgres
+	MigrateSeq int
```
md2119, line 112: How does this manifest when upgrading from one version to another on postgres?

For example, if I were to update a table schema, do we have to update this seq-num? If yes, only on a concerned table or all tables?

dashrews78: So to clarify, this is only for the initial case of migrating Rocks to Postgres.  Correct?  As such I think I find the n_01_to_n_02_postgres_clusters naming to be confusing as we aren't really going from postgres table version 1 to postgres table version 2, but we are simply working on the first table in the postgres schema hierarchy which is the clusters table.    But I will admit I don't have a great feel for how migrations work now.

md2119: Thanks for the clarification @c-du. I really wanted to understand the post-release process. For example, if I were to add 3 new independent tables, should each table have a unique seq number or can I push all those tables through one sequence number?

md2119:
Can we consider deriving `--migrate-from` arg instead of requiring users to specify it? We can determine this since the buckets are registered through Go code.

tools/generate-helpers/pg-table-bindings/main.go
c-du: If you add new tables after release, you do not need to add new sequence numbers, that is, you do not need to add migration. Once you added the schema, the migrator will pick up and apply the schema before the central starts.

c-du:
>

c-du:
> Can we consider deriving `--migrate-from` arg instead of requiring users to specify it? We can determine this since the buckets are registered through Go code.

It seems to me that it cannot derived from existing parameters. I cannot derive it from the existing parameters. Migrator does not have all the context from the central codes and we can only use the static context from the code-level.

migrator/migrations/n_43_to_n_44_postgres_report_configurations/legacy/rocksdb_plugin.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func keyFunc(msg proto.Message) []byte {
+	return []byte(msg.(*storage.ReportConfiguration).GetId())
+}
+
+// New returns a new Store instance using the provided rocksdb instance.
+func New(db *rocksdb.RocksDB) (Store, error) {
+	baseCRUD := generic.NewCRUD(db, bucket, keyFunc, alloc, false)
+	cacheCRUD, err := mapcache.NewMapCache(baseCRUD, keyFunc)
```
connorgorman, line 38: We don't want to have the mapcache in these instances

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
```diff
@@ -0,0 +1,78 @@
...
+		return nil, err
+	}
+	return &storeImpl{
+		crud: cacheCRUD,
+	}, nil
+	{{- end}}
+}
+
+// UpsertMany batches objects into the DB
+func (b *storeImpl) UpsertMany(_ context.Context, objs []*storage.{{.Type}}) error {
```
connorgorman, line 64: This function can just move into the test for the migration

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
```diff
@@ -0,0 +1,78 @@
...
+func New(db *rocksdb.RocksDB) (Store, error) {
+	{{- if .UniqKeyFunc}}
+	baseCRUD := generic.NewUniqueKeyCRUD(db, bucket, keyFunc, alloc, uniqKeyFunc, {{.TrackIndex}})
+	{{- else}}
+	baseCRUD := generic.NewCRUD(db, bucket, keyFunc, alloc, {{.TrackIndex}})
+	{{- end}}
+	{{- if not .Cache}}
+	return  &storeImpl{crud: baseCRUD}, nil
+	{{- else}}
+	cacheCRUD, err := mapcache.NewMapCache(baseCRUD, keyFunc)
```
connorgorman, line 53: This is not needed in the migration

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
```diff
@@ -0,0 +1,78 @@
+
+package legacy
```
connorgorman, line 2: Add space below and put this at the top of the file

tools/generate-helpers/rocksdb-bindings/main.go
```diff
@@ -246,6 +266,23 @@ func main() {
 		if err := os.WriteFile("store.go", buf.Bytes(), 0644); err != nil {
 			return err
 		}
+		if props.MigrationSeq == 0 {
+			return nil
+		}
+		buf.Truncate(0)
```
connorgorman, line 272: Why do we need to truncate? If we truncate do we need to reallocate below?

connorgorman (approved):
Looks good

connorgorman (commented):
Something that we will want to do in a follow up is allow for partial migrations so maybe we checkpoint after a particular migration has been successful

migrator/migrations/n_43_to_n_44_postgres_report_configurations/legacy/rocksdb_plugin.go
c-du: Fixed.

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
c-du: The migration is generated from postgres autogen. And all boltdbs do not have autogen. So I would keep it in rocksdb side.

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
c-du: Fixed.

tools/generate-helpers/rocksdb-bindings/rocksdb_plugin.go.tpl
c-du: Fixed.

tools/generate-helpers/rocksdb-bindings/main.go
c-du: fixed

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -313,6 +318,7 @@ func (s *storeImpl) upsert(ctx context.Context, objs ...*{{.Type}}) error {
 }
 
 func (s *storeImpl) Upsert(ctx context.Context, obj *{{.Type}}) error {
+    {{- if not $inMigration}}
```
janisz, line 321: Why do we need this? Can't we just pass context with all access?

PR #2365, opened 2022-07-11
ROX-11317, ROX-11316: Adding filter select and default category functionality to Policy Categories

alwayshooin:
## Description

As title suggests -- also adding default category select dropdown as well. Should function as expected and look like below:

image
image
image
image

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesFilterSelect.tsx
+++ b/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesFilterSelect.tsx
@@ -9,13 +9,15 @@ type PolicyCategoriesFilterSelectProps = {
     customPolicyCategories: PolicyCategory[];
 };
 
+type CategoryFilter = 'Default categories' | 'Custom categories';
+
 function PolicyCategoriesFilterSelect({
     setCurrentPolicyCategories,
     defaultPolicyCategories,
     customPolicyCategories,
 }: PolicyCategoriesFilterSelectProps) {
     const [isOpen, setIsOpen] = useState(false);
-    const [selectedFilters, setSelectedFilters] = useState<string[]>([
+    const [selectedFilters, setSelectedFilters] = useState<CategoryFilter[]>([
         'Default categories',
         'Custom categories',
     ]);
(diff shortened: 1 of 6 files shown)


ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
```diff
@@ -2,24 +2,22 @@ import React from 'react';
...
-    //     id: string;
-    //     name: string;
-    //     isDefault: boolean;
-    // }[];
-    policyCategories: string[];
+    policyCategories: {
+        id: string;
+        name: string;
+        isDefault: boolean;
+    }[];
```
pedrottimark, line 9: It will be possible to import `PolicyCategory` type as a follow up after #2341 has been merged.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesFilterSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+    customPolicyCategories: PolicyCategory[];
+};
+
+function PolicyCategoriesFilterSelect({
+    setCurrentPolicyCategories,
+    defaultPolicyCategories,
+    customPolicyCategories,
+}: PolicyCategoriesFilterSelectProps) {
+    const [isOpen, setIsOpen] = useState(false);
+    const [selectedFilters, setSelectedFilters] = useState<string[]>([
```
pedrottimark, line 18: 1. What do you think about reducing responsibility for this component with `selectedFilters` and `setSelectedFilters` props in parallel with its sibling `Input` element, which has:
    * `value={filterTerm}`
    * `onChange={setFilterTerm}`
2. A string union `type Whatever = 'Default categories' | 'Custom categories'` would help TypeScript to help us make sure that all cases are covered.
3. Lifting up the responsibility to compute current policy categories according to this filter and the string filter puts all cause-and-effect in one place.
4.  For extra credit, some special care might minimize re-rendering and make it possible to use assignment to a local `currentPolicyCategories` variable instead of a `useState` hook:
     * if both of these filters are selected, then use `allPolicyCategories` instead of new array `[...defaultPolicyCategories, ...customPolicyCategories]`
     * if no string filter, skip `filter` method instead of new array

pedrottimark (approved):
Another super step forward. Nice sized pull requests.

PR #2375, opened 2022-07-12
Dackbox CVE SAC search tests

rhybrillou:
## Description

This change introduces a basic set of search tests for image and node CVEs.
The goal is to provide a test harness and ensure the functional correctness of scope checks on indirectly scoped resource types

## Checklist

## Testing Performed

New tests added at datastore level.


proto/storage/cluster.proto
```diff
@@ -250,6 +250,22 @@ message UpgradeProgress {
     google.protobuf.Timestamp since    = 3;
 }
 
+message ClusterCVEEdge {
```
md2119, line 253: Can we please move it back to `relations.proto`

rhybrillou: It's done. Along with the move of the proto message in the proto.lock file (which caused `make storage-protos-compatible` or `make style` to fail locally).

pkg/sac/query_helper.go
```diff
@@ -14,6 +14,16 @@ var (
 // BuildClusterLevelSACQueryFilter builds a Scoped Access Control query filter that can be
 // injected in search queries for resource types that have direct cluster scope level.
 func BuildClusterLevelSACQueryFilter(root *effectiveaccessscope.ScopeTree) (*v1.Query, error) {
+	return buildClusterLevelSACQueryFilter(root, true)
+}
+
+// BuildSilentClusterLevelSACQueryFilter builds a Scoped Access Control query filter that can be
+// injected in search queries for resource types that have direct cluster scope level.
+func BuildSilentClusterLevelSACQueryFilter(root *effectiveaccessscope.ScopeTree) (*v1.Query, error) {
```
dhaus67, line 22: nit: `BuildNonVerbose..` instead of `Silent`? I was wondering in the beginning what the name actually stands for.

central/cve/datastoretest/datastore_sac_test.go
```diff
@@ -588,19 +611,114 @@ func (s *cveDataStoreSACTestSuite) TestSACImageCVEGetBatch() {
...
+			testCtx := s.imageTestContexts[c.contextKey]
+			results, err := s.imageCVEStore.Search(testCtx, nil)
+			s.NoError(err)
+			expectedCVENames := make([]string, 0, len(c.expectedCVEFound))
+			for name, visible := range c.expectedCVEFound {
+				if visible {
+					expectedCVENames = append(expectedCVENames, getImageCVEID(name))
+				}
+			}
+			fetchedCVEIDs := make(map[string]bool, 0)
```
dhaus67, line 652: If it's only about storing the key value, the idiomatic thing to use within go would be `map[string]struct{}`. But looking at the code, a simple string array would suffice as well, right?

Applies to more places than just the one here.

dhaus67 (approved):
Just minor comments from my side, otherwise LGTM.

PR #2376, opened 2022-07-12
Improve assertion for risk sort priority integration test in UI

pedrottimark:
## Description

Although I do not have as clear an explanation as I like, there is timing problem between the test and the UI.

Although the screen grab displays the expected state of UI for descending sort when the assertion failed, the test apparently got values during the previous state of ascending sort, just a moment too soon before re-rendering.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2341/pull-ci-stackrox-stackrox-master-gke-ui-e2e-tests/1545393241459789824
cypress-failure

I confused the risk table in which the number of deployments is predictable with the violations table in which the number of rows is not predictable, when I rewrote this test to replace mock response with real response in #1555

Because risk table displays all deployments on only one page in integration testing environment, it is possible to get expected values for descending sort, therefore cypress can retry to overcome timing problems.

## Checklist
- [x] Edited integration test

## Testing Performed

Ran test in local deployment

### DIFF
--- a/ui/apps/platform/cypress/integration/risk/risk.test.js
+++ b/ui/apps/platform/cypress/integration/risk/risk.test.js
@@ -42,7 +42,7 @@ describe('Risk page', () => {
                         'eq',
                         '?sort[id]=Deployment%20Risk%20Priority&sort[desc]=false'
                     );
-                    // No request because response is sorted ascending.
+                    // There is no request because response is sorted ascending.
 
                     // Sort descending by priority.
                     cy.get(priorityColumnHeadingSelector).click();
@@ -50,7 +50,7 @@ describe('Risk page', () => {
                         'eq',
                         '?sort[id]=Deployment%20Risk%20Priority&sort[desc]=true'
                     );
-                    cy.wait('@deploymentswithprocessinfo'); // assume intercept in visitRiskDeployments
+                    // There is a request because of change in sorting.
                     cy.get(priorityColumnHeadingSelector).should('have.class', '-sort-desc');
                     cy.get(
                         `.rt-tr-group:first-child .rt-tr .rt-td:nth-child(5):contains("${priorityLast}")`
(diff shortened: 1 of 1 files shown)


ui/apps/platform/cypress/integration/risk/risk.test.js
```diff
@@ -22,44 +22,57 @@ describe('Risk page', () => {
...
-                    const firstCellPriority = Number($priorityCells.eq(0).text());
-                    const lastCellPriority = Number(
-                        $priorityCells.eq($priorityCells.length - 1).text()
+                    // Sort descending by priority.
+                    cy.get(priorityColumnHeadingSelector).click();
+                    cy.location('search').should(
+                        'eq',
+                        '?sort[id]=Deployment%20Risk%20Priority&sort[desc]=true'
+                    );
+                    cy.wait('@deploymentswithprocessinfo'); // assume intercept in visitRiskDeployments
```
dvail, line 53: Is there still a requirement for the `cy.wait` calls here? It seems that with the more robust selectors you have to grab the `first` and `last` priority table rows Cypress's query retry will make the `wait` redundant.

dvail:
Regarding your comments:

> Although I do not have as clear an explanation as I like, there is timing problem between the test and the UI.
> 
> Although the screen grab displays the expected state of UI for descending sort when the assertion failed, the test apparently got values during the previous state of ascending sort, just a moment too soon before re-rendering.

I had a similar issue with a Cypress dashboard test right before I moved it to Jest/RTL. In my case however there was an assertion that relied on a UI order that passed _right before_ another one failed. Both of the assertions used Cypress's query retry in the same way to get the elements so I was at a loss on why one would pass right before the second failed. The video recording demonstrated that the UI was in the correct order, but the query retry simply never found the element. This was on a test that was working in `master` for a few weeks and I couldn't correlate a change to the breakage.

Might not be relevant, and I didn't spend much time investigating since I was replacing the test anyway, but I thought I would mention it.

ui/apps/platform/cypress/integration/risk/risk.test.js
pedrottimark: Good point. I replaced with comments whether or not there is a request for the sake of future readers.

pedrottimark:
A weakness in the original pattern for this test was to get text from DOM and then assert on the comparison.

For whatever reason why the test got ahead of the UI, cypress cannot retry that pattern.

PR #2377, opened 2022-07-12
Upstream release automation: local scripts run improvements

0x656b694d:
## Description

The `local-env.sh` script prepares environment for running workflow scripts in local.

## Testing Performed

Locally with temporary test scripts and create-cluster.

```
$ ./local-env.sh create-cluster flavor name 2h false key=value
*Unknown status*
Infra status for 'name':
''''''

Will attempt to create the cluster
Error: rpc error: code = NotFound desc = flavor "flavor" not found
```


.github/workflows/scripts/local-env.sh
```diff
@@ -13,4 +16,12 @@ export main_branch=master
...
 export DRY_RUN=true
 
-source "$(git rev-parse --show-toplevel)/.github/workflows/scripts/common.sh"
+SCRIPTS_ROOT="$(git rev-parse --show-toplevel)/.github/workflows/scripts"
+
+# Supress shellcheck false warning:
+# shellcheck source=/dev/null
+source "$SCRIPTS_ROOT/common.sh"
+
+SCRIPT="$SCRIPTS_ROOT/$1.sh"
```
vikin91, line 25: This is quite brave :) Shall we first check if user provided `$1` (can be added later if you like).

.github/workflows/scripts/common.sh
```diff
@@ -49,7 +49,8 @@ gh_summary() {
 }
 export -f gh_summary
 
-if [ "$#" -gt 0 ]; then
+# bash trick to check if the script is sourced.
+if ! (return 0 2>/dev/null); then # called
```
vikin91, line 53: When using dark magic, maybe we could wrap it in a function to make it easier to understand?


```bash
function is_sourced() {
    (return 0 2>/dev/null)
}
```

vikin91 (approved):

PR #2404, opened 2022-07-13
ROX-11767: Client side crash when viewing Image -> CVE

sachaudh:
## Description

We ended up using some of the new fields for the Image Findings section without the feature flag on and it was causing some issues. This PR fixes that and also a few other fixes to get it all working again.

## Checklist

## Testing Performed

1. Navigate to Vuln Mgmt Dashboard -> Images -> a single image -> Click CVE name

https://user-images.githubusercontent.com/4805485/178845679-399711c8-8ed3-4a38-ad8a-903270ac8612.mov

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Entity/Cve/VulnMgmtEntityCve.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Entity/Cve/VulnMgmtEntityCve.js
@@ -13,7 +13,6 @@ import {
     CLUSTER_CVE_DETAIL_FRAGMENT,
 } from 'Containers/VulnMgmt/VulnMgmt.fragments';
 import WorkflowEntityPage from 'Containers/Workflow/WorkflowEntityPage';
-import NoResultsMessage from 'Components/NoResultsMessage';
 import VulnMgmtCveOverview from './VulnMgmtCveOverview';
 import VulnMgmtList from '../../List/VulnMgmtList';
 import {
@@ -58,17 +57,6 @@ const VulmMgmtCve = ({ entityId, entityListType, search, entityContext, sort, pa
     const cveType = getCVETypeFromStack(worklowStateStack) || entityTypes.IMAGE_CVE;
     const vulnQuery = vulnQueryMap[cveType];
     const vulnFields = vulnFieldMap[cveType];
-
-    // When switching between workflow states, the entity state changes before this component dismounts
-    if (!entityListType && (!vulnQuery || !vulnFields)) {
-        return (
-            <NoResultsMessage
-                message={`No vulnerability found of type ${cveType}`}
-                className="p-3"
-                icon="info"
-            />
-        );
-    }
 
     const overviewQuery = gql`
         query getCve($id: ID!, $query: String, $scopeQuery: String) {


ui/apps/platform/src/Containers/VulnMgmt/Entity/Cve/VulnMgmtEntityCve.js
```diff
@@ -36,10 +37,21 @@ const vulnFieldMap = {
 
 const VulmMgmtCve = ({ entityId, entityListType, search, entityContext, sort, page }) => {
     const workflowState = useContext(workflowStateContext);
-    const cveType = workflowState.getBaseEntityType() || entityTypes.IMAGE_CVE;
+    const cveType = workflowState.getCurrentEntityType() || entityTypes.IMAGE_CVE;
```
vjwilson, line 40: I changed this back in the master, after the release cut, because if it is not `getBaseEntityType`, it breaks other things.

If this is what it takes to fix this with the flag off, we can use it, but we'll need to come to consensus on how to handle this for the Postgres release.

vjwilson (approved):

ui/apps/platform/src/Containers/VulnMgmt/Entity/Cve/VulnMgmtEntityCve.js
sachaudh: Hmm, do you know what else it breaks? It was breaking because `getBaseEntityType` ended up returning `IMAGE` as the entity when viewing a CVE single within the Image context

sachaudh: Actually, I see the issue you mentioned. When I go to https://localhost:3000/main/vulnerability-management/cve/CVE-2014-7186 and then click a related entity, it'll break. That's inconvenient. Seems like it either breaks in one or the other so we'll need to find something that works regardless of whether it's nested or not

sachaudh: We can't have a CVE twice in the workflow stack, right? Maybe the best solution is to run through the stack (bottom-up) and find the CVE type?

sachaudh: @vjwilson Ok I have another change that should help

ui/apps/platform/src/Containers/VulnMgmt/Entity/Cve/VulnMgmtEntityCve.js
```diff
@@ -35,9 +42,20 @@ const vulnFieldMap = {
     CLUSTER_CVE: CLUSTER_CVE_DETAIL_FRAGMENT,
 };
 
+function getCVETypeFromStack(worklowStateStack) {
```
vjwilson, line 44: "CVET" looks like the name of a Cold War-era paramilitary group.

Gen X, ftw ;)

ui/apps/platform/src/utils/WorkflowState.js
```diff
@@ -248,6 +248,10 @@ export class WorkflowState {
         return new WorkflowState(useCase, newStateStack, newSearch, newSort, newPaging);
     }
 
+    getStateStack() {
+        return this.stateStack;
+    }
```
vjwilson, line 253: @alwayshooin , For the 72 release, I'd like you to double-check that Saif and I aren't abusing your WorkflowState. 😈

vjwilson (approved):

vikin91:
All good! I think we need to allow merges to the release branch in the future - it makes totally sense. Just this time we were unprepared for that combination 😅

PR #2405, opened 2022-07-13
Split CVE column in Clusters list into 3 types of CVEs

vjwilson:
## Description

Besides the effect in the title, this PR ensures that the links to for each of the complex CVE count widgets in the table columns go to the appropriate sub-page for that type of CVE.

## Checklist

## Testing Performed

Updated the e2e tests to recognize the new column when the flag is turned on.
(TODO: update the counts tests to account for the split counts.)

Manual testing:

Three columns
cluster_list_3_types_of_cves

Clicking the total count for Image CVEs
clicking_on_image_cves_total_link

Clicking the fixable count for Image CVEs
clicking_on_image_cves_fixable_link

Clicking the total count for Node CVEs
clicking_on_node_cves_total_link

Clicking the fixable count for Node CVEs
clicking_on_node_cves_fixable_link

Clicking the total count for Platform CVEs
Screen Shot 2022-07-13 at 5 30 51 PM


ui/apps/platform/src/Containers/Workflow/WorkflowEntityPage.js
```diff
@@ -23,6 +23,12 @@ const useCaseDefaultSorts = {
     [useCases.VULN_MANAGEMENT]: vulnMgmtDefaultSorts,
 };
 
+function removeGraphqlAlias(fieldName) {
+    const parts = fieldName?.split(':');
+
+    return parts ? parts[0] : '';
```
sachaudh, line 29: If `parts` is a falsey value, wouldn't that break things?

sachaudh (approved):
LGTM. Just a question

ui/apps/platform/src/Containers/Workflow/WorkflowEntityPage.js
vjwilson: Falsey will fall through to the empty string option, which I only observed right before the component dismounts before returning to a list page.

PR #2406, opened 2022-07-13
CI: Fix the upgrader version in nightlies

gavin-stackrox:
## Description

Similar to the reason for rebuilding roxctl in nightlies we also need to rebuild upgrader to get the nightly tag.

Note: there is a lot of whack-a-mole around getting 'nightlies' to function in OpenShift CI in the same manner as Circle CI. We may need to revisit the mechanism.

## Checklist

## Testing Performed

CI is sufficient to ensure there is no regression with PR/merge CI.


sbostick (approved):
LGTM

PR #2176, opened 2022-06-23
Add request helpers especially for visit function in integration tests

pedrottimark:
## Description

Provide a reusable declarative convention for the following cypress features:

1. `requestConfig`
    * `routeMatcherMap` see https://docs.cypress.io/api/commands/intercept#Matching-with-RouteMatcher
    * `opnameAliasesMap` see https://docs.cypress.io/api/commands/intercept#Aliasing-individual-requests
    * `waitOptions` see https://docs.cypress.io/api/commands/wait#Timeouts

2. `staticResponseMap` see https://docs.cypress.io/api/commands/intercept#StaticResponse-objects

### Goals

Keep the current inline `cy.intercept(…)`, `visit(…)` or interact, and `cy.wait(…)` sandwich pattern, unless helpers or tests have any of the following goals:

1. Reduce repeated intercept and wait in helper functions which provide mock responses for **different subsets** of requests
    * For example, 6 `visitSystemHealthWithWhatever` functions in helpers/systemHealth.js
    * PatternFly dashboard widgets was another potential example, but might be even better with unit tests instead of integration tests
2. Reuse intercept and wait from helper functions for initial **visit** and also for subsequent **interaction**
    * For example, reuse `requestConfig` in `visitComplianceDashboard` and also `scanCompliance` for intercept-click-wait sandwich
    * For example, reuse `requestConfig` in container page url `visit` function call and also in `visitFromLeftNav` function call for intercept-click-wait sandwich
3. Reuse `requestConfig` in helper functions or tests which use real responses **and also mock responses** via optional `staticResponseMap` arg

By the way, here is one example each of minus and plus to specify most or all requests for initial visit or subsequent interaction:
* Minus: Tests become coupled to implementation of data fetching
* Plus: Eliminate timing problems as a cause of test failure when a sequence of tests assert of DOM elements which do **not** depend on slow responses: for example,becaise most tests in compliance do **not** depend on 6 slow responses to complianceStandards requests, the debug panel in cypress looks like a chain-reaction fender bender with cancelled requests because next test starts before some responses from previous test (although situation in compliance has not caused failures, prevent potential blocker if cypress 13 changes its behavior in this edge case)

Keep current intercept and wait for subset of requests in tests:
* which do not need to factor out common `routeConfig` for real versus mock responses
* which do not have problem with slow responses

### Limitations

1. The `requestHandler` arg of `intercept` is limited in this abstraction to distinguish aliases by payload for multiple GraphQL requests for the same opname
[...]

PR #2407, opened 2022-07-14
Replace wrap with retryable selectors in policies integration tests

pedrottimark:
### Problem

> should have row action to disable policy that has enabled status and then enable it again

> Timed out retrying after 4000ms: Expected to find element: `td[data-label="Status"]:contains("Enabled")`, but never found it.

### Analysis

This problem was caused by my confusion a few months ago, similar to problem recently fixed in #2376

Timing problems are likely if a test gets a DOM element, and then makes a derived assertion, even with `intercept` for request and `wait` for response.

In this case, the table re-renders after the test got a DOM element because of policies request. The test waited on PATCH request, but not GET request.

Although `changePolicyStatusInTable` function does wait for both requests, even more important it gets the expected status with a retryable assertion.

### Changed files

1. Edit cypress/helpers/policiesPatternFly.js
    * Replace repeated imperative `cy.intercept` and `cy.wait` calls with declarative `routeMatcherMap` object added in #2176
    * Replace `cy.visit` with `visit` function call
    * Move heading assertion from individual tests to helper functions
    * Add nav line assertion to helper functions
    * Delete `visitPoliciesCallback` function
    * Add `changePolicyStatusInTable` function
    * Add `deletePolicyInTable` function
    * Replace aliases
    * Edit `editFirstPolicyFromTable` function to add assertions from test
    * Edit `closeFirstPolicyFromTable` function to add assertions from test

2. Edit cypress/integration/policies/policiesTable.test.js
    * Replace `cy.visit` with `visit` function call
    * should navigate using the left nav: Move heading assertion to helper function
    * should have row action to edit policy and then cancel: call `editFirstPolicyFromTable`
    * should have row action to clone policy and then cancel: call `cloneFirstPolicyFromTable`
    * should have row action to disable policy that has enabled status and then enable it again: call `changePolicyStatusInTable`
    * should have disabled row action to delete system default policy: use specific policy name
    * should have enabled row action to delete user generated policy: call `cloneFirstPolicyFromTable`

### Residue

1. Investigate why 2 requests: GET /v1/search/metadata/options?categories=POLICIES
2. Replace imports from classic cypress/constants/PoliciesPage.js
    * cypress/integration/risk/riskSearchToPolicy.test.js
    * cypress/integration/violations/violations.test.js
2. Rename files
    * cypress/constants/PoliciesPagePatternFly.js
    * cypress/helpers/policiesPatternFly.js

## Checklist

## Testing Performed

1. Ran changed tests one at a time in local deployment
    * should have row action to edit policy and then cancel
    * should have row action to clone policy and then cancel
    * should have row action to disable policy that has enabled status and then enable it again
    * should have disabled row action to delete system default policy
2. Ran all policies tests in local deployment
    * policies/exportPolicy.test.js
    * policies/importPolicy.test.js
    * policies/policyWizardStep3.test.js

### DIFF
--- a/ui/apps/platform/cypress/helpers/policiesPatternFly.js
+++ b/ui/apps/platform/cypress/helpers/policiesPatternFly.js
@@ -1,36 +1,54 @@
 import * as api from '../constants/apiEndpoints';
 import { selectors, url as policiesUrl } from '../constants/PoliciesPagePatternFly';
 import { visitFromLeftNavExpandable } from './nav';
+import { visit } from './visit';
 
 // Navigation
 
-export function visitPolicies() {
-    // Include empty search query to distinguish from intercept with search query.
-    cy.intercept('GET', `${api.policies.policies}?query=`).as('getPolicies');
-    cy.visit(policiesUrl);
-    cy.wait('@getPolicies');
-}
+const routeMatcherMap = {
+    notifiers: {
+        method: 'GET',
+        url: api.integrations.notifiers,
+    },
+    'search/metadata/options': {
+        method: 'GET',
+        url: `/v1/search/metadata/options?categories=POLICIES`,
+    },
+    policies: {
+        method: 'GET',
+        // Include empty search query to distinguish from intercept with search query.
+        url: `${api.policies.policies}?query=`,
+    },
+};
 
-export function visitPoliciesCallback(callback) {
-    // Include empty search query to distinguish from intercept with search query.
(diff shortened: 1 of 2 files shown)


pedrottimark:
How convenient that the next integration test failure in the queue occurred for this pull request:

> Network Baseline Flows Navigating to Deployment should navigate to a different deployment when clicking the "Navigate" button

vjwilson (approved):

PR #2411, opened 2022-07-14
ROX-11543: Re-sync pod tests

fredrb:
## Description

Add tests that validate pod event behavior based on deployments being applied. 

## Checklist

## Testing Performed

- Deployment messages are sent with pod information (image ids).
- Multiple pod events are sent per each pod received.
- Pods that don't belong to any deployment are sent as deployment updates as well.


sensor/tests/resource/helper.go
```diff
@@ -287,3 +316,37 @@ func (c *TestContext) waitForResource(timeout time.Duration, fn condition) error
...
+			}
+		}
+	}
+	return uniqueNames.AsSlice()
+}
+
+func GetUniqueDeploymentNames(messages []*central.MsgFromSensor) []string {
+	uniqueNames := set.NewStringSet()
+	for _, msg := range messages {
+		if msg.GetEvent().GetDeployment() != nil {
```
lvalerom, line 347: Same here. I think we should check namespace.

sensor/tests/resource/pod/pod_test.go
```diff
@@ -0,0 +1,93 @@
...
+		s.Fail("failed to setup test context: %s", err)
+	} else {
+		s.testContext = testContext
+	}
+}
+
+func (s *PodHierarchySuite) Test_ContainerSpecOnDeployment() {
+	s.testContext.RunWithResources([]resource.YamlTestFile{
+		resource.NginxDeployment,
+		resource.NginxServiceAccount,
```
lvalerom, line 38: Question, is this only necessary because the `nginx.yaml` has a `serviceAccountName` used in the role tests? Do you think it would make sense to have another `nginx.yaml` without the `serviceAccountName` so the tests don't create an extra resources?

lvalerom (changes requested):
It looks really good. Just a couple of comments/questions. 

I have played a little bit with the tests and if I change the `nginx.yaml` to not have the `serviceAccountName` field the tests pass individually, but if I run both at the same time `Test_ParentlessPodsAreTreatedAsDeployments` fails here:
```golang
s.Require().Len(uniquePodNames, 4,
			"Should have received four different pod events (3 from nginx-deployment and 1 from nginx-rouge")
```

I think it's because the pods created in the previous tests aren't yet deleted (or sensor hasn't sent the deletion message yet) Do you think it would make sense to use different namespaces for each test?

sensor/kubernetes/listener/resources/deployments.go
```diff
@@ -130,6 +130,14 @@ func (d *deploymentHandler) processWithType(obj, oldObj interface{}, action cent
 	// because IF the object is a pod, we want to process the pod event.
 	objAsPod, _ := obj.(*v1.Pod)
 
+	//if objAsPod != nil && objAsPod.Namespace == "sensor-integration" {
+	//	fmt.Printf("== [DEBUG] POD EVENT: %s\n", objAsPod.Name)
+	//}
+	//
+	//if deploymentWrap != nil && deploymentWrap.Namespace == "sensor-integration" {
+	//	fmt.Printf("== [DEBUG] DEPLOYMENT EVENT: %s\n", deploymentWrap.Name)
+	//}
```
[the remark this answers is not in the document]
fredrb: oops

fredrb:
> I have played a little bit with the tests and if I change the nginx.yaml to not have the serviceAccountName field the tests pass individually

The reason there is a service account is simply because `role_test` requires it. The solution is to either add service accounts in the pod test so the pods are created or to simply use a separate `.yaml` file for this test. I'm fine either way.

> I think it's because the pods created in the previous tests aren't yet deleted (or sensor hasn't sent the deletion message yet) Do you think it would make sense to use different namespaces for each test?

I don't think that's the case, because after each test we delete the namespace entirely and even wait for its completion:

```go
func createTestNs(ctx context.Context, r *resources.Resources, name string) (*v1.Namespace, func() error, error) {
	nsObj := v1.Namespace{}
	nsObj.Name = name
	if err := r.Create(ctx, &nsObj); err != nil {
		return nil, nil, err
	}
	return &nsObj, func() error {
		err := r.Delete(ctx, &nsObj)
		if err != nil {
			return err
		}

		// wait for deletion to be finished
		if err := wait.For(conditions.New(r).ResourceDeleted(&nsObj)); err != nil {
			fmt.Printf("failed to wait for namespace deletion")
		}
		return nil
	}, nil
}
```

What could be happening here is that we're filtering all messages received in `fakeCentral`, regardless is they were sent before or after the deletion. Maybe the simplest thing is to clear the messages after each test to make sure we're using an empty message list?

sensor/tests/resource/pod/pod_test.go
fredrb: Replied in the top level comment.

sensor/tests/resource/helper.go
```diff
@@ -287,3 +316,37 @@ func (c *TestContext) waitForResource(timeout time.Duration, fn condition) error
...
+			break
+		}
+	}
+	return lastMessage
+}
+
+func GetUniquePodNamesFromPrefix(messages []*central.MsgFromSensor, prefix string) []string {
+	uniqueNames := set.NewStringSet()
+	for _, msg := range messages {
+		if msg.GetEvent().GetPod() != nil {
```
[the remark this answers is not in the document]
fredrb: Done

sensor/tests/resource/helper.go
fredrb: Done

lvalerom (approved):
Looks good to me! 🚀

PR #2414, opened 2022-07-14
Empty-Commit

sachaudh:
## Description

This PR was created in reference to https://github.com/stackrox/stackrox/pull/2404#issuecomment-1184123548


sachaudh:
@vikin91 let me know if I need to do anything else for this

vikin91:
@sachaudh no worries about that, this is not needed on master. We have already added an empty commit to the release branch and pushed a new tag. All worked well.

The issue was that OSCI does not react to pushing a tag. Instead it requires always a commit to trigger a build. So this was the lesson for us to always push a new tag with at least one commit (empty or not empty). We did it for rc.6 and it worked well.

This PR can be closed now (no need to merge), but I appreciate your help to resolve the situation 🙌

sachaudh:
Sounds good thanks again!

PR #2419, opened 2022-07-14
ROX-10275: Migrating user defined policy categories into new categories bucket

clickboo:
## Description

1. This migration takes care of migrating the user defined categories into rocksdb bucket.
2. Default categories get loaded on datastore init, hence not migrated.
3. Follow on migration to be expected, to read each policy and store it using category ids instead of names. This will be closer to turning on feature flag (since migrations cannot be done under feature flags, and this migration should not be done if we are not releasing the feature - put differenly if this migration is in, we better be able to read policies with category ids and make sense of them :).

## Checklist

## Testing Performed
CI automation only

### DIFF
--- a/migrator/migrations/m_106_to_m_107_policy_categories/migration.go
+++ b/migrator/migrations/m_106_to_m_107_policy_categories/migration.go
@@ -83,10 +83,6 @@ func fetchCategoriesToAdd(db *bolt.DB) (set.FrozenStringSet, error) {
 	categories := set.NewStringSet()
 	err := db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket(policiesBucket)
-		// Pre-req: Migrating a non-existent bucket should not fail.
-		if bucket == nil {
-			return nil
-		}
 		return bucket.ForEach(func(k, v []byte) error {
 			policy := &storage.Policy{}
 			if err := proto.Unmarshal(v, policy); err != nil {
--- a/migrator/migrations/m_106_to_m_107_policy_categories/migration_test.go
+++ b/migrator/migrations/m_106_to_m_107_policy_categories/migration_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stackrox/rox/pkg/rocksdb"
 	"github.com/stackrox/rox/pkg/testutils"
 	"github.com/stackrox/rox/pkg/testutils/rocksdbtest"
+	"github.com/stackrox/rox/pkg/uuid"
 	"github.com/stretchr/testify/suite"
 	"github.com/tecbot/gorocksdb"
 	bolt "go.etcd.io/bbolt"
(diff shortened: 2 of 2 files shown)


migrator/migrations/m_106_to_m_107_policy_categories/migration.go
```diff
@@ -0,0 +1,108 @@
...
+		return errors.Wrap(err, "failed to write new categories to rocksdb")
+	}
+	return nil
+}
+
+func fetchCategoriesToAdd(db *bolt.DB) (set.FrozenStringSet, error) {
+	categories := set.NewStringSet()
+	err := db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(policiesBucket)
+		// Pre-req: Migrating a non-existent bucket should not fail.
```
theencee, line 86: Is this even possible normally? It should fail, because this implies you have no buckets for policies.

migrator/migrations/m_106_to_m_107_policy_categories/migration_test.go
```diff
@@ -0,0 +1,130 @@
...
+	suite.db = rocksDB
+	suite.databases = &dbTypes.Databases{BoltDB: boltdb, RocksDB: rocksDB.DB}
+}
+
+func (suite *categoriesRocksDBMigrationTestSuite) TearDownTest() {
+	testutils.TearDownDB(suite.databases.BoltDB)
+	rocksdbtest.TearDownRocksDB(suite.db)
+}
+
+func (suite *categoriesRocksDBMigrationTestSuite) TestCategoriesMigrationToRocksDB() {
```
theencee, line 98: recommend additional tests:

* Policy with user defined categories AND default -> user defined should be added
* Policy with only default categories -> nothing new should be added

theencee (approved):
lgtm but please add the additional tests

migrator/migrations/m_106_to_m_107_policy_categories/migration.go
clickboo: Nope honestly not possible, will remove

theencee: Instead of removing, I'd just keep it there and return an error (after logging).

migrator/migrations/m_106_to_m_107_policy_categories/migration_test.go
clickboo: Added

migrator/migrations/m_106_to_m_107_policy_categories/migration.go
```diff
@@ -0,0 +1,104 @@
...
+	err := db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(policiesBucket)
+		return bucket.ForEach(func(k, v []byte) error {
+			policy := &storage.Policy{}
+			if err := proto.Unmarshal(v, policy); err != nil {
+				log.WriteToStderrf("Failed to unmarshal policy data for key %s: %v", k, err)
+				return nil
+			}
+			for _, c := range policy.GetCategories() {
+				categories.Add(strings.Title(c))
```
viswajithiii, line 93: Do we really want to do a `strings.Title` on user-entered categories?

clickboo: Yes, to avoid them from having "their own category", "Their own category" and "their own Category" as 3 different categories.

PR #2424, opened 2022-07-15
ROX-11296: set redirect_uri for the OAuth exchange

0x656b694d:
## Description

In the case of multiple routes created for the same service account OAuth client, Openshift needs the `redirect_uri` parameter set for the code exchange call.

The suggested fix makes use of the fact that the backend `callback` is called on the `redirect_uri` passed to the `/authorize`, so the value is available in the request.

## Checklist

## Testing Performed

* Created an instance of Openshift OAuth provider;
* Created 2 additional routes: `alt-central` and `alt2-central`;
* Altered `central` service account to add redirect URIs and routes references;
* Tried to log in via the created provider using the 3 routes;
* Observed successful login and the use of the correct route in the URI.

### DIFF
--- a/pkg/auth/authproviders/openshift/backend_impl.go
+++ b/pkg/auth/authproviders/openshift/backend_impl.go
@@ -103,9 +103,9 @@ func (b *backend) LoginURL(clientState string, ri *requestinfo.RequestInfo) (str
 	state := idputil.MakeState(b.id, clientState)
 
 	// Augment baseRedirectURLPath to a redirect URL with hostname, etc set.
-	redirectURL := dexconnector.MakeRedirectURI(ri, b.baseRedirectURLPath)
+	redirectURI := dexconnector.MakeRedirectURI(ri, b.baseRedirectURLPath)
 
-	return b.openshiftConnector.LoginURL(defaultScopes, redirectURL.String(), state)
+	return b.openshiftConnector.LoginURL(defaultScopes, redirectURI.String(), state)
 }
 
 func (b *backend) RefreshURL() string {
--- a/pkg/auth/authproviders/openshift/backend_impl.go
+++ b/pkg/auth/authproviders/openshift/backend_impl.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/x509"
 	"net/http"
-	"net/url"
 	"os"
 	"time"
 
@@ -17,7 +16,6 @@ import (
 	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/grpc/requestinfo"
 	"github.com/stackrox/rox/pkg/httputil"
-	"github.com/stackrox/rox/pkg/netutil"
 	"github.com/stackrox/rox/pkg/satoken"
 )
 
@@ -56,9 +54,9 @@ type callbackAndRefreshConnector interface {
 }
 
 type backend struct {
-	id                 string
-	baseRedirectURL    url.URL
-	openshiftConnector callbackAndRefreshConnector
+	id                  string
+	baseRedirectURLPath string
+	openshiftConnector  callbackAndRefreshConnector
 }
 
 type openShiftSettings struct {
@@ -75,11 +73,6 @@ func newBackend(id string, callbackURLPath string, _ map[string]string) (authpro
 		return nil, err
 	}
 
-	baseRedirectURL := url.URL{
-		Scheme: "https",
-		Path:   callbackURLPath,
-	}
-
 	dexCfg := dexconnector.Config{
 		Issuer:          openshiftAPIUrl,
 		ClientID:        settings.clientID,
@@ -93,9 +86,9 @@ func newBackend(id string, callbackURLPath string, _ map[string]string) (authpro
 	}
 
 	b := &backend{
-		id:                 id,
-		baseRedirectURL:    baseRedirectURL,
-		openshiftConnector: openshiftConnector,
+		id:                  id,
+		baseRedirectURLPath: callbackURLPath,
+		openshiftConnector:  openshiftConnector,
 	}
 
 	return b, nil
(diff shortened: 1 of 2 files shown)


pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -202,9 +202,7 @@ func validateEndpoint(endpoint string, tlsConfig *tls.Config) error {
 
 // LoginURL returns the URL to redirect the user to login with.
 func (c *openshiftConnector) LoginURL(_ connector.Scopes, callbackURL string, state string) (string, error) {
-	clonedConfig := *c.oauth2Config
```
0x656b694d, line 205: Unnecessary code removed.

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -219,7 +217,20 @@ func (c *openshiftConnector) HandleCallback(s connector.Scopes, r *http.Request)
 		ctx = context.WithValue(r.Context(), oauth2.HTTPClient, c.httpClient)
 	}
 
-	token, err := c.oauth2Config.Exchange(ctx, q.Get("code"))
+	// r.URL might not get the Host part, therefore use r.Host here.
+	redirect_uri := &url.URL{
+		Scheme: "https",
```
rukletsov, line 222: My biggest concern is that the way redirect URI is assembled here is different from the way it is assembled for `LoginURL()`, see e.g. https vs. http part [here](https://github.com/stackrox/stackrox/blob/d6a70fabd2b995dd3063464af8a07b71ca8b599a/pkg/auth/authproviders/openshift/backend_impl.go#L117). Because of this difference, verifying code correctness requires non-local info (openshift/backend_impl.go) and is non-trivial because possible use cases are not explicitly listed everywhere.

What seems cleaner is to _remember_ the redirect URI used in `LoginURL()` and use it in `HandleCallback()`. Maybe we can use something in `r` here to unambiguously identify the matching `LoginURL()` call. We can also extend `state` with some random string and in `LoginURL()` record the cloned config or just the redirect URI in a map keyed by `state`s. Doing so will also allow us to verify `state`, something we actually need to do anyway according to [rfc6749#section-10.12](http://tools.ietf.org/html/rfc6749#section-10.12).

0x656b694d: As discussed, URL reconstruction should be fine, as the callback URL shouldn't differ from the redirect_URI.
Use of `http` scheme for local endpoint has been taken into account in the appended commit.

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -202,9 +204,23 @@ func validateEndpoint(endpoint string, tlsConfig *tls.Config) error {
...
+}
+
+// MakeRedirectURI constructs redirect URI from request info and the path.
+func MakeRedirectURI(ri *requestinfo.RequestInfo, path string) *url.URL {
+	scheme := "https"
+
+	// Allow HTTP only if the client did not use TLS and the host is localhost.
+	if !ri.ClientUsedTLS && netutil.IsLocalEndpoint(ri.Hostname) {
+		scheme = "http"
+	}
```
rukletsov, line 217:
```suggestion
	// Allow HTTP only if the client did not use TLS and the host is localhost.
	scheme := "https"
	if !ri.ClientUsedTLS && netutil.IsLocalEndpoint(ri.Hostname) {
		scheme = "http"
	}
```

I prefer grouping same things together for clarity. WDYT?

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -219,7 +235,13 @@ func (c *openshiftConnector) HandleCallback(s connector.Scopes, r *http.Request)
 		ctx = context.WithValue(r.Context(), oauth2.HTTPClient, c.httpClient)
 	}
 
-	token, err := c.oauth2Config.Exchange(ctx, q.Get("code"))
+	ri := requestinfo.FromContext(ctx)
+	redirect_uri := MakeRedirectURI(&ri, ri.HTTPRequest.URL.Path)
```
rukletsov, line 239:
```suggestion
	redirectURI := MakeRedirectURI(&ri, ri.HTTPRequest.URL.Path)
```

rukletsov: Nit: it makes sense to give this var and the one in `LoginURL()` the same name because they denote the same things. I don't care much whether it would be `redirectURL` or `redirectURI` : )

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -219,7 +235,13 @@ func (c *openshiftConnector) HandleCallback(s connector.Scopes, r *http.Request)
...
+
+	// redirect_uri is set here to support the mulitple redirect URI case.
```
rukletsov, line 241:
```suggestion
	// Our service might be accessible via different routes and hence specify
	// different redirect URIs in the login URL to authorization server. The
	// latter must check that the redirect URI passed with the initial request
	// equals the one passed during the code exchange. Hence we dynamically
	// adjust the redirect URI in the oauth2 config here to the URL deduced
	// from the request to us, which we expect to match the redirect URL we
	// included in login URL earlier in the flow.
```
Let's be more specific here on why we added this. Who knows who and when will have to debug this : )

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
```diff
@@ -17,7 +17,9 @@ import (
 	"github.com/dexidp/dex/connector"
 	"github.com/dexidp/dex/storage/kubernetes/k8sapi"
 	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/grpc/requestinfo"
 	"github.com/stackrox/rox/pkg/httputil/proxy"
+	"github.com/stackrox/rox/pkg/netutil"
 	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/rox/pkg/utils"
 	"golang.org/x/oauth2"
```
rukletsov, line 25: There is a table below saying what has been changed from the original file. Could you please update it with the following:
*  deduce redirect URI's hostname via `MakeRedirectURI()` function

pkg/auth/authproviders/openshift/backend_impl.go
```diff
@@ -109,13 +102,9 @@ func (b *backend) Config() map[string]string {
...
 
-	// baseRedirectURL does not include the hostname, take it from the request.
-	// Allow HTTP only if the client did not use TLS and the host is localhost.
-	redirectURL := b.baseRedirectURL
-	redirectURL.Host = ri.Hostname
-	if !ri.ClientUsedTLS && netutil.IsLocalEndpoint(redirectURL.Host) {
-		redirectURL.Scheme = "http"
-	}
+	// baseRedirectURLPath does not include the hostname, take it from the
+	// request.
```
rukletsov, line 106: This is now a bit misleading, how about

```suggestion
	// Augment baseRedirectURLPath to a redirect URL with hostname, etc set.
```

rukletsov (approved):
The change looks good logically — approving. I left a couple of comments on the form & style though, so please address and commit.

pkg/auth/authproviders/openshift/internal/dexconnector/openshift.go
0x656b694d: Me too, that's why the comment about HTTP exception is close to the related checks. In your suggestion the comment looks belonging to the `scheme := "https"` line, which to me looks confusing.

PR #2428, opened 2022-07-15
ROX-11228: Centralize CRUD SAC test case definitions

rhybrillou:
## Description

Test cases were added for the datastores of directly scoped object types in parallel. Most of the test cases for the Add, Delete, Get, Update, Upsert functions are very similar or identical. Centralizing the test cases in a common location helps defining a single set of pre-defined behaviour patterns and the associated tests.

## Checklist

## Testing Performed

Manual run of the go integration tests in the impacted datastore directories.

CI run should be enough for this test re-structuration change.

### DIFF
--- a/central/alert/datastore/datastore_sac_test.go
+++ b/central/alert/datastore/datastore_sac_test.go
@@ -117,8 +117,7 @@ func (s *alertDatastoreSACTestSuite) TestUpsertAlert() {
 	s.testAlertIDs = append(s.testAlertIDs, alert1.Id)
 	s.testAlertIDs = append(s.testAlertIDs, alert2.Id)
 
-	testedVerb := "upsert"
-	cases := testutils.GenericNamespaceSACUpsertTestCases(s.T(), testedVerb)
+	cases := testutils.GenericNamespaceSACUpsertTestCases(s.T(), testutils.VerbUpsert)
 
 	for name, c := range cases {
 		s.Run(name, func() {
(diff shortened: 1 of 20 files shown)


central/alert/datastore/datastore_sac_test.go
```diff
@@ -141,71 +117,26 @@ func (s *alertDatastoreSACTestSuite) TestUpsertAlert() {
...
-			expectError:   false,
-			expectedError: nil,
-		},
-		"read-write on the right cluster and at least the right namespace can upsert": {
-			scopeKey:      testutils.Cluster2NamespacesABReadWriteCtx,
-			expectError:   false,
-			expectedError: nil,
-		},
-	}
+	testedVerb := "upsert"
```
dhaus67, line 120: Shall we make the `testedVerb | upsert` value as a const of the `testutils` package? Seems like the best solution IMO.

central/processbaseline/datastore/datastore_sac_test.go
```diff
@@ -204,58 +153,17 @@ func (s *processBaselineSACTestSuite) TestUpdateProcessBaselineElements() {
...
-			scopeKey: testutils.Cluster2NamespaceBReadWriteCtx,
-		},
-		"read-write on matching cluster and no namespace should be able to add": {
-			scopeKey: testutils.Cluster2ReadWriteCtx,
-		},
-		"read-write on matching cluster and at least one matching namespace should be able to add": {
-			scopeKey: testutils.Cluster2NamespacesABReadWriteCtx,
-		},
-	}
+	testedVerb := "update"
```
dhaus67, line 156: Same as the `upsert` value, probably best to make this as part of a const within the `testutils` package.

dhaus67 (approved):
Just minor nit's, otherwise great work de-duping the code and introducing the generic test cases! ❤️

PR #2429, opened 2022-07-15
Display Imminent instead of zero or negative days for Cluster Deletion in UI

pedrottimark:
## Description

Here is a screen grab with temporary code change to replace `null` with `{ daysUntilDeletion: 0 }` for `clusterRetentionInfo` prop of `ClusterDeletion` element:
Imminent

Also I corrected my misunderstanding about `isExcluded` property.

### Changed files

1. Edit cypress/integration/clusters/clusterDeletion.test.js
    * Add 2 tests for zero and negative days
    * Delete test for `isExcluded: false` which is not a valid case

2. Edit src/Containers/Clusters/Components/ClusterDeletion.tsx
    * Correct the comments
    * Add **Imminent** for `daysUntilDeletion < 1`
    * Delete **Deletion is turned off** for `isExcluded: false` which is not a valid case

3. Edit src/types/clusterService.proto.ts
    * Correct the comments

### Residue

1. Improve integration tests to mock response in clusters table instead of cluster panel, because **Cluster Deletion** is below the fold on the panel. If the test fails, the snapshot displays the incorrectly rendered result. Also possible to use test to take pictures of UI for different scenarios.

## Checklist

## Testing Performed

1. `export ROX_DECOMMISSIONED_CLUSTER_RETENTION=true` in ui
2. `yarn deploy-local` in ui
3. `yarn start` in ui
4. `export CYPRESS_ROX_DECOMMISSIONED_CLUSTER_RETENTION=true` in ui/apps/platform
5. `yarn cypress-open` in ui/apps/platform
    * clusters/clusterDeletion.test.js see that existing and new tests pass

### DIFF
--- a/ui/apps/platform/cypress/integration/clusters/clusterDeletion.test.js
+++ b/ui/apps/platform/cypress/integration/clusters/clusterDeletion.test.js
@@ -92,17 +92,24 @@ describe('Cluster page clusterRetentionInfo', () => {
         cy.get('div:contains("Cluster Deletion"):contains("in 1 day")'); // FYI red color
     });
 
+    it('should display Imminent for 0 days in Cluster Deletion widget for mock request', () => {
+        const clusterRetentionInfo = { daysUntilDeletion: 0 };
+        visitClusterWithRetentionInfo(clusterRetentionInfo);
+
+        cy.get('div:contains("Cluster Deletion"):contains("Imminent")'); // FYI red color
+    });
+
+    it('should display Imminent for -1 days in Cluster Deletion widget for mock request', () => {
+        const clusterRetentionInfo = { daysUntilDeletion: -1 };
+        visitClusterWithRetentionInfo(clusterRetentionInfo);
+
+        cy.get('div:contains("Cluster Deletion"):contains("Imminent")'); // FYI red color
+    });
+
     it('should display Excuded from deletion in Cluster Deletion widget for mock request', () => {
         const clusterRetentionInfo = { isExcluded: true };
         visitClusterWithRetentionInfo(clusterRetentionInfo);
 
         cy.get('div:contains("Cluster Deletion"):contains("Excluded from deletion")');
-    });
-
-    it('should display Deletion is turned off in Cluster Deletion widget for mock request', () => {
-        const clusterRetentionInfo = { isExcluded: false };
-        visitClusterWithRetentionInfo(clusterRetentionInfo);
-
-        cy.get('div:contains("Cluster Deletion"):contains("Deletion is turned off")');
     });
 });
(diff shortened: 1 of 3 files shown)


central/cluster/service/service_impl.go
```diff
@@ -141,6 +141,11 @@ func (s *serviceImpl) getClusterRetentionInfo(ctx context.Context, cluster *stor
 		return nil, nil
 	}
 
+	// If retention is disabled, there is no "days remaining" calculation to be done.
```
charmik-redhat, line 144: Can we also short circuit the `getClusterIDToRetentionInfoMap` function when retentionDays == 0 ?

theencee: I don't think it makes a massive difference since the beginning of this method doesn't really do much. And it's cleaner than leaking the config outside of this method. Thoughts?

charmik-redhat (approved):

ui/apps/platform/cypress/integration/clusters/clusterDeletion.test.js
```diff
@@ -92,17 +92,24 @@ describe('Cluster page clusterRetentionInfo', () => {
...
+    it('should display Imminent for -1 days in Cluster Deletion widget for mock request', () => {
+        const clusterRetentionInfo = { daysUntilDeletion: -1 };
+        visitClusterWithRetentionInfo(clusterRetentionInfo);
+
+        cy.get('div:contains("Cluster Deletion"):contains("Imminent")'); // FYI red color
     });
 
-    it('should display Deletion is turned off in Cluster Deletion widget for mock request', () => {
-        const clusterRetentionInfo = { isExcluded: false };
+    it('should display Excuded from deletion in Cluster Deletion widget for mock request', () => {
```
vjwilson, line 109: nit: typo
```suggestion
    it('should display Excluded from deletion in Cluster Deletion widget for mock request', () => {
```

ui/apps/platform/src/Containers/Clusters/Components/ClusterDeletion.tsx
```diff
@@ -24,19 +24,23 @@ function ClusterDeletion({ clusterRetentionInfo }: ClusterDeletionProps): ReactE
...
         const healthStatus = getClusterDeletionStatus(daysUntilDeletion);
         const { bgColor, fgColor } = healthStatusStyles[healthStatus];
-        const text = daysUntilDeletion === 1 ? 'in 1 day' : `in ${daysUntilDeletion} days`;
+        /* eslint-disable no-nested-ternary */
+        const text =
+            daysUntilDeletion < 1
+                ? 'Imminent'
+                : daysUntilDeletion === 1
+                ? 'in 1 day'
+                : `in ${daysUntilDeletion} days`;
```
vjwilson, line 33: I personally think a `switch` is easier to read than nested ternary, but 1 extra level, as it is here, it's ok.

vjwilson (approved):

PR #2434, opened 2022-07-15
Add category defaults

clickboo:
## Checklist

## Testing Performed
Manual on cluster after images are built


central/policycategory/datastore/singleton.go
```diff
@@ -41,7 +43,18 @@ func Singleton() DataStore {
...
 }
 
-// TODO: implement addDefaults adds the default categories into the postgres table for policy categories.
 func addDefaults(s policyCategoryStore.Store) {
+	// Preload the default policies.
+	defaultCategories, err := categories.DefaultPolicyCategories()
+	// Hard panic here is okay, since we can always guarantee that we will be able to get the default policies out.
+	utils.CrashOnError(err)
+
+	var count int
```
viswajithiii, line 52: Count will always be 0, as written

central/policycategory/datastore/singleton.go
```diff
@@ -41,7 +43,18 @@ func Singleton() DataStore {
...
+	for _, p := range defaultCategories {
+		if err := s.Upsert(policyCategoryCtx, p); err != nil {
+			panic(err)
```
viswajithiii, line 55:
```suggestion
			utils.CrashOnError(err)
```

PR #2435, opened 2022-07-15
Plotted vuln graphQL resolver enhancements

md2119:
## Description

Performance enhancements and bug fixes to all plotted vulnerability resolvers.

## Checklist

### DIFF
--- a/central/graphql/resolvers/image_vulnerabilities.go
+++ b/central/graphql/resolvers/image_vulnerabilities.go
@@ -263,9 +263,15 @@ func imageCveToVulnerabilityWithSeverity(in []*storage.ImageCVE) []Vulnerability
 }
 
 func (resolver *imageCVEResolver) withImageVulnerabilityScope(ctx context.Context) context.Context {
+	if features.PostgresDatastore.Enabled() {
+		return scoped.Context(ctx, scoped.Scope{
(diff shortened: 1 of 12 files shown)


keyallis (approved):

central/graphql/resolvers/image_vulnerabilities.go
```diff
@@ -263,9 +263,15 @@ func imageCveToVulnerabilityWithSeverity(in []*storage.ImageCVE) []Vulnerability
 }
 
 func (resolver *imageCVEResolver) withImageVulnerabilityScope(ctx context.Context) context.Context {
```
keyallis, line 265: This function should only be used in the postgres code, all of the legacy datastore usage is via conjunction queries I'm pretty sure. I'm not sure this change does anything

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -201,9 +201,15 @@ func withNodeCveTypeFiltering(q string) string {
 }
 
 func (resolver *nodeCVEResolver) withNodeVulnerabilityScope(ctx context.Context) context.Context {
```
keyallis, line 203: Same as image vulnerability scoping function comment

PR #1525, opened 2022-05-03
ROX-10308: Separating ClientPolicy and server Policy types

alwayshooin:
## Description

As it states above -- the code should compile and Policies page/form should work as expected.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [stackrox/openshift-docs](https://github.com/stackrox/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #2444, opened 2022-07-18
Delete unused removeEmptyFieldsDeep from utils in UI

pedrottimark:
## Description

Linda, it looks like the last import was deleted in #1525

Found while searching for some search utils, pardon pun :)

Zero net change in build confirms that it is not used.

## Checklist

## Testing Performed

1. `yarn build` in ui
    * branch - master: 0 = 10416065 - 10416065
3. `yarn test` in ui/apps/platform


alwayshooin (approved):
seems fine to me. i wonder what we used these for. thanks for cleaning up!

PR #2445, opened 2022-07-18
Delete unused pollingReduxRoutines from utils in UI

pedrottimark:
## Description

Saif, it looks like Rob Stark deleted the last import while adding pagination to Violations table

Found while searching for some search utils, pardon pun :)

Zero net change in build confirms that it is not used.
## Checklist

## Testing Performed

1. `yarn build` in ui
    * branch - master: 0 = 10416065 - 10416065


sachaudh (approved):
Nice find 🔍 LGTM let's get rid of it 🚢

PR #2448, opened 2022-07-18
Add transformed sorts field searcher

md2119:
## Checklist

Transformed searcher for nodes needs to be carried over for the short-term since FE cannot accommodate the change, to send full and correct sort fields (component+version), in MVP. Once FE removes the dependency on BE to transform `Component` to `Component+Version` sort, we can remove this wrapper.

## Checklist

## Testing Performed
CI


theencee (approved):

viswajithiii (approved):
Seems ok

PR #2449, opened 2022-07-18
[ROX-11299] : Add node components resolvers and sub-resolvers that use Postgres datastores

charmik-redhat:
## Description

- Added NodeComponentResolver implementation that uses Postgres
- Added NodeComponentLoader
- Refactoring
- Removed `CVE Snoozed : False` query from image component resolvers

## Checklist

## Testing Performed

- Added unit tests for NodeComponentLoader

TODO(replace-me)
screenshots.


central/graphql/resolvers/loaders/node_components.go
```diff
@@ -0,0 +1,139 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var nodeComponentLoaderType = reflect.TypeOf(storage.NodeComponent{})
+
+func init() {
+	RegisterTypeFactory(reflect.TypeOf(storage.NodeComponent{}), func() interface{} {
```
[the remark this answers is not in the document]
charmik-redhat: Moved it behind feature flag

central/graphql/resolvers/node_components.go
```diff
@@ -48,7 +47,7 @@ func init() {
 // NOTE: This list is and should remain alphabetically ordered
 type NodeComponentResolver interface {
 	FixedIn(ctx context.Context) string
-	ID(ctx context.Context) graphql.ID
+	Id(ctx context.Context) graphql.ID
```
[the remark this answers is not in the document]
charmik-redhat: As per schema, FE would call for `id` field. Now, graphQL's reflection mechanism is able to look for the implementing function as long as the first letter is capitalized. So, graphQL should  be able to redirect to `Id` if such a method exists. So, I think `ID` is not needed.

central/graphql/resolvers/node_components.go
```diff
@@ -97,11 +132,151 @@ func (resolver *Resolver) NodeComponentCount(ctx context.Context, args RawQuery)
...
+*/
+
+// FixedIn returns the node component version that fixes all the fixable vulnerabilities in this component.
+func (resolver *nodeComponentResolver) FixedIn(ctx context.Context) string {
+	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.NodeComponents, "FixedIn")
+	return ""
+}
+
+// NodeComponentLastScanned is the last time the node component was scanned in a node.
+func (resolver *nodeComponentResolver) NodeComponentLastScanned(ctx context.Context) (*graphql.Time, error) {
```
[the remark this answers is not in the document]
charmik-redhat: Yeah that's right, it should be renamed. I have kept it same for now since I saw at least one instance in UI that calls for `nodeComponentLastScanned`

central/graphql/resolvers/node_components.go
```diff
@@ -97,11 +132,151 @@ func (resolver *Resolver) NodeComponentCount(ctx context.Context, args RawQuery)
...
+	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.NodeComponents, "NodeComponentLastScanned")
+	if err := readNodes(ctx); err != nil {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
charmik-redhat: Good catch, removed the check.

central/graphql/resolvers/components_v2.go
```diff
@@ -197,15 +201,15 @@ func (eicr *imageComponentResolver) TopVuln(ctx context.Context) (VulnerabilityR
 // TopNodeVulnerability returns the first node component vulnerability with the top CVSS score.
 func (eicr *imageComponentResolver) TopNodeVulnerability(ctx context.Context) (NodeVulnerabilityResolver, error) {
```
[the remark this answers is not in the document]
charmik-redhat: Yeah, that was confusing because in legacy implementation `storage.ImageComponent` was used for both image and node components. I moved the implementation to `nodeComponentResolver` and added conversion methods to convert `storage.ImageComponent` to `storage.NodeComponent` .

central/graphql/resolvers/node_vulnerabilities.go
```diff
@@ -174,6 +174,55 @@ func (resolver *Resolver) NodeVulnerabilityCounter(ctx context.Context, args Raw
...
+	}
+	query = tryUnsuppressedQuery(query)
+	vulns, err := vulnLoader.FromQuery(ctx, query)
+	if err != nil || len(vulns) == 0 {
+		return nil, err
+	} else if len(vulns) > 1 {
+		return nil, errors.New("multiple vulnerabilities matched for top node vulnerability")
+	}
+
+	return resolver.wrapNodeCVE(vulns[0], true, nil)
```
charmik-redhat, line 223: @md2119 TopNodeVulnerability here returns the nodeCVEResolver without embedding any scope. Should it instead embed the context sent to the function TopNodeVulnerability.

md2119: `TopNodeVulnerability` should respect only node scope and no other. If that scope if missing, in ctx or query, then yes, we need to add that piece of information

charmik-redhat: Yes, either node or node_component scope will always be present in ctx when we are returning from `TopNodeVulnerability`, so instead of `return resolver.wrapNodeCVE(vulns[0], true, nil)` should I change it to below?
```
res, err := resolver.wrapNodeCVE(vulns[0], true, nil)
If err != nil {
   return nil, err
}
res.ctx = ctx
return res, nil
```

Also, the same would apply to `TopImageVulnerability` as well since that does not add any ctx to returned imageCVEResolver as well.

central/graphql/resolvers/components_v2.go
```diff
@@ -115,10 +124,11 @@ func (resolver *Resolver) imageComponentDataStoreQuery(ctx context.Context, args
 	} else if !exists {
 		return nil, errors.Errorf("component not found: %s", string(*args.ID))
 	}
-	return resolver.wrapImageComponent(component, true, nil)
+	// return resolver.wrapImageComponent(component, true, nil)
```
md2119, line 127: remove

central/graphql/resolvers/components_v2.go
```diff
@@ -127,7 +137,8 @@ func (resolver *Resolver) imageComponentsLoaderQuery(ctx context.Context, query
 		return nil, err
 	}
 
-	return resolver.wrapImageComponents(componentLoader.FromQuery(ctx, query))
+	return componentLoader.FromQuery(ctx, query)
+	// return resolver.wrapImageComponents(componentLoader.FromQuery(ctx, query))
```
md2119, line 141: remove

central/graphql/resolvers/image_components.go
```diff
@@ -133,7 +133,6 @@ func (resolver *Resolver) ImageComponents(ctx context.Context, q PaginatedQuery)
 	}
 
 	// get values
-	query = tryUnsuppressedQuery(query)
```
md2119, line 136: Nice catch!

central/graphql/resolvers/node_components.go
```diff
@@ -6,36 +6,35 @@ import (
...
+	"github.com/stackrox/rox/pkg/search/scoped"
 	"github.com/stackrox/rox/pkg/utils"
 )
 
 func init() {
 	schema := getBuilder()
 	utils.Must(
 		// NOTE: This list is and should remain alphabetically ordered
-		schema.AddType("NodeComponent", []string{
+		schema.AddExtraResolvers("NodeComponent", []string{
```
md2119, line 24: Add a comment that resolvers for fields of `storage.NodeComponent` are auto-generated and located at `generated.go`. That way readers are not taken by surprise about the missing resolvers.

md2119: Add a comment to imagecomponent resolver as well.

central/graphql/resolvers/node_components.go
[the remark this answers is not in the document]
md2119: +1 Charmik.
As long as the method (resolver) is exported (in golang names starting with capital letter) and matches on spelling, the schema is correctly mapped to appropriate resolvers.

central/graphql/resolvers/node_components.go
```diff
@@ -97,11 +132,151 @@ func (resolver *Resolver) NodeComponentCount(ctx context.Context, args RawQuery)
...
+	return search.NewQueryBuilder().AddExactMatches(search.ComponentID, resolver.data.GetId()).Query()
+}
+
+/*
+Sub Resolver Functions
+*/
+
+// FixedIn returns the node component version that fixes all the fixable vulnerabilities in this component.
+func (resolver *nodeComponentResolver) FixedIn(ctx context.Context) string {
+	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.NodeComponents, "FixedIn")
```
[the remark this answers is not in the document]
md2119: As discussed offline, return empty for now since we never actually supported this field. We do not have support from the scanner. This field indicates the highest version of the component that fixes all the fixable vulnerabilities in node. More importantly, sort the version lexically does not give the highest version since each distro has a specific versioning scheme.

central/graphql/resolvers/node_components.go
```diff
@@ -97,11 +132,151 @@ func (resolver *Resolver) NodeComponentCount(ctx context.Context, args RawQuery)
...
+
+// License of the node component
+func (resolver *nodeComponentResolver) License(ctx context.Context) (*licenseResolver, error) {
+	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.NodeComponents, "License")
+	return nil, nil
+}
+
+// Location of the node component.
+func (resolver *nodeComponentResolver) Location(ctx context.Context, args RawQuery) (string, error) {
+	defer metrics.SetGraphQLOperationDurationTime(time.Now(), pkgMetrics.NodeComponents, "Location")
```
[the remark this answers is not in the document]
md2119: This is not relevant to nodes. We should work closely with Van to remove this fields from FE. For now, keep it to avoid breaking changes but make sure to remove in 72. Add a task to the follow-up epic https://issues.redhat.com/browse/ROX-10992

central/graphql/resolvers/node_components.go
```diff
@@ -97,11 +133,184 @@ func (resolver *Resolver) NodeComponentCount(ctx context.Context, args RawQuery)
...
+		return nil, err
+	} else if len(nodes) > 1 {
+		return nil, errors.New("multiple nodes matched for last scanned component query")
+	}
+
+	return timestamp(nodes[0].GetScan().GetScanTime())
+}
+
+// License of the node component
+func (resolver *nodeComponentResolver) License(_ context.Context) (*licenseResolver, error) {
```
md2119, line 223: This is not relevant to nodes. We should work closely with Van to remove this fields from FE. For now, keep it to avoid breaking changes but make sure to remove in 72. Add a task to the follow-up epic https://issues.redhat.com/browse/ROX-10992

central/graphql/resolvers/node_vulnerabilities.go
md2119: Yea, pass in the context

md2119 (approved):
Address the comments and :shipit:

central/graphql/resolvers/node_components.go
charmik-redhat: Added comments to both.

PR #2451, opened 2022-07-18
Optimize dashboard tests and avoid querying that may cause act errors

dvail:
## Description

This PR optimizes the dashboard tests that use RTL in two main ways:

1. It replaces `*ByRole` queries with a more performant query (`*ByLabelText`, `*ByText`) when possible. I've found that for tests running against components of this size, a `*ByText` query can be used as a replacement easily about 95% of the time. The remaining 5% can usually be targeted by label or by title, with a couple remaining as `*ByRole` when the extra semantic information really makes sense. ([This](https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.test.tsx#L71) is one case where it was difficult to replace `ByRole` without really changing what was being tested.)

With many test runs and some unscientific napkin math, I've found that changing the queries in the above way sped up test execution from anywhere between 20%-70%, even in the case of a single `*ByRole` call in a test.

2. It also ensures that components under test that render data asynchronously (e.g. via the Apollo `MockedProvider`) run at least one query against an element that _does not exist until the data is returned_. I believe this is the other missing piece that is causing occasional `act()` errors in CI.

From what I understand, Apollo's `MockedProvider` is initialized with a state of `loading=true` and then immediately sets state to the provided mocked data, and then rerenders. The place that the tests run into problems is when a user event immediately follows the initial render, causing this state change and the click event overlap `act` calls. (Both `userEvent.click()` and `render()` from RTL call `act` under the hood.) By awaiting a query that looks for an element that only exists when the data is available we can ensure these two calls do not overlap.

I was able to reproduce these failures locally every 1 out of ~15 times before the change, but after implementing this change I have not seen the error a single time.

## Checklist

## Testing Performed

Many, many local unit test runs.

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
@@ -73,9 +73,9 @@ describe('AgingImages dashboard widget', () => {
 
         // When all items are selected, the total should be equal to the total of all buckets
         // returned by the server
-        const cardHeading = await screen.findByRole('heading', {
-            name: `${result0 + result1 + result2 + result3} Aging images`,
-        });
+        const cardHeading = await screen.findByText(
+            `${result0 + result1 + result2 + result3} Aging images`
+        );
         expect(cardHeading).toBeInTheDocument();
 
         // Each bar should display text that is specific to that time bucket, not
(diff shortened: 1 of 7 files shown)


pedrottimark (approved):
Thank you for following up on unit tests. Especially for explaining why you did what you did.

PR #2452, opened 2022-07-18
ROX-11799: Combine entities into one column in VM Images list

vjwilson:
## Description

This combines the Deployments and Components (soon-to-be Image Components) columns into one Entities column on the Images list page.

Notes:
1. I tried reusing the existing multi-entities TableLinksCount component, but would have required making the conditional logic even more serpentine than it already is, so I "forked" it into a new component for this use case.
2. I still had to pass the feature flag into my new component, so that it can show plain "Components" when the flag is off, and "Image Components" when the flag is on.
3. Saif has a fix for the broken link for the Image Components count. I will test this change again, once Saif's fix is merged. However, that doesn't affect this actual change.

## Checklist
- [x] Investigated and inspected CI test results (failures are unrelated and are being fixed in another PR)

## Testing Performed

Manual testing
Screen Shot 2022-07-18 at 5 17 51 PM

### DIFF
--- a/ui/apps/platform/src/Components/workflow/ImageTableCountLinks/ImageTableCountLinks.tsx
+++ b/ui/apps/platform/src/Components/workflow/ImageTableCountLinks/ImageTableCountLinks.tsx
@@ -1,0 +1,56 @@
+import React, { ReactElement, useContext } from 'react';
+
+import { ResourceType, resourceTypes } from 'constants/entityTypes';
+import TableCountLink from 'Components/workflow/TableCountLink';
+import workflowStateContext from 'Containers/workflowStateContext';
+
+type ImageTableCountLinksProps = {
+    row: {
+        deploymentCount: number;
+        componentCount: number;
+        id: string;
+    };
+    textOnly: boolean;
+    isFrontendVMUpdatesEnabled: boolean;
+};
+
+function ImageTableCountLinks({
+    row,
+    textOnly,
+    isFrontendVMUpdatesEnabled = false,
+}: ImageTableCountLinksProps): ReactElement {
+    // const { isFeatureFlagEnabled } = useFeatureFlags();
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Components/workflow/ImageTableCountLinks/ImageTableCountLinks.tsx
```diff
@@ -0,0 +1,56 @@
...
+    isFrontendVMUpdatesEnabled: boolean;
+};
+
+function ImageTableCountLinks({
+    row,
+    textOnly,
+    isFrontendVMUpdatesEnabled = false,
+}: ImageTableCountLinksProps): ReactElement {
+    // const { isFeatureFlagEnabled } = useFeatureFlags();
+    // const isFrontendVMUpdatesEnabled = isFeatureFlagEnabled('ROX_FRONTEND_VM_UDPATES');
```
sachaudh, line 23: You can delete comments

ui/apps/platform/src/Containers/VulnMgmt/List/Images/VulnMgmtListImages.js
```diff
@@ -180,44 +180,19 @@ export function getCurriedImageTableColumns(watchedImagesTrigger, isFeatureFlagE
                 sortable: false,
             },
             {
-                Header: `Deployments`,
+                Header: `Entities`,
```
sachaudh, line 183: Just to confirm, we don't need to feature flag the column changes here, right?

sachaudh (approved):
LGTM

ui/apps/platform/src/Containers/VulnMgmt/List/Images/VulnMgmtListImages.js
vjwilson: We aren't feature-flagging the column itself, just whether it shows Components or Image Components.

PR #2457, opened 2022-07-19
Bump collector version

Molter73:
## Description

The version being used holds the latest changes from the upstream Falco libraries.

## Checklist

## Testing Performed

CI should be enough.


Stringy (approved):

Molter73:
Looking into the failures from CI:
- Some are from registry and scanner tests that should be unrelated to collector.
- There are some more errors that are also showing up on other PRs, so I believe they are also unrelated to this change.

PR #2462, opened 2022-07-19
ROX-11747 Increase default timeout for waitForDeployment in Groovy tests to prevent flakes

viswajithiii:
## Description

Increase the timeout to 2 minutes from the current 1 minute to prevent flakes similar to those in the ticket. This won't make tests slower in the passing case since we still poll every 2 seconds.

## Checklist
- ~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed

CI

### DIFF
--- a/qa-tests-backend/src/test/groovy/Services.groovy
+++ b/qa-tests-backend/src/test/groovy/Services.groovy
@@ -423,7 +423,7 @@ class Services extends BaseService {
         return disappearedFromStackRox
     }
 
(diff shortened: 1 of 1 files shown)


gavin-stackrox (approved):
Worth a shot!

PR #2463, opened 2022-07-19
Dv/jest mock implementation explicit

dvail:
## Description

This is an alternative implementation to https://github.com/stackrox/stackrox/pull/2354 with the same core functionality. The difference in this case is that the mocking of `@patternfly/react-charts` is handled explicitly via an export instead of automagically through Jest's "Manual Mocks" feature. 

The goal here is to provide contrast to the other approach and potentially clearer and more discoverable code after feedback from the last UI meeting.

Note: We should only merge one of the two PRs.

## Checklist

## Testing Performed

Automated tests pass


pedrottimark (approved):
Seeing 0, 1, or 2 `jest.mock(…)` calls is in a test file seems clear and supports search.

PR #2465, opened 2022-07-19
ROX-10935: Bump helm dep to fix a loopclosure mistake

porridge:
## Description

It is likely the bug referred to in the subject is caused by a [classic loopclosure bug](https://go.dev/doc/faq#closures_and_goroutines) that I attempt to [fix here](https://github.com/porridge/helm/commit/59a08440474d707a25b44977e044f17c058840dc).

If this helps with the flake I'll take care of upstreaming it.

## Checklist

## Testing Performed

Relying on CI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -206,4 +206,4 @@ exclude k8s.io/client-go v12.0.0+incompatible
 
 exclude github.com/openshift/client-go v3.9.0+incompatible
 
-replace helm.sh/helm/v3 => github.com/porridge/helm/v3 v3.7.3-0.20220712081439-a4d429d11bd9
+replace helm.sh/helm/v3 => github.com/porridge/helm/v3 v3.7.3-0.20220719213745-59a08440474d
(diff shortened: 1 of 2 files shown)


mtesseract (approved):

viswajithiii:
@porridge @mtesseract Note that this was not a loop closure bug, and the [linked commit](https://github.com/porridge/helm/commit/59a08440474d707a25b44977e044f17c058840dc) is a no-op.

```go
	for _, info := range infos {
		currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
		if kind != currentKind {
			wg.Wait()
			kind = currentKind
		}
		wg.Add(1)
		go func(i *resource.Info) {
			errs <- fn(i)
			wg.Done()
		}(info)
	}
```
This is because we are passing the `info` variable to the function as an argument to the goroutine, and the closure only refers to `info` via the `i` argument that is passed to it. The only outer variable that the closure uses is `wg`, which it should use.

That is to say, the following would have been a classic loopclosure bug, but the original code was not buggy.
```go
		go func() {
			errs <- fn(info)
			wg.Done()
		}()
```

porridge:
> @porridge @mtesseract Note that this was not a loop closure bug, and the [linked commit](https://github.com/porridge/helm/commit/59a08440474d707a25b44977e044f17c058840dc) is a no-op.

💩 🤦🏻

PR #2471, opened 2022-07-20
ROX-9565: support building, pushing and testing operator on prow

porridge:
## Description

Please look at individual commits' descriptions.

These changes should not change anything in the CircleCI-based operator tests.
At the same time, this is the complete set of changes (on the `stackrox/stackrox` repo side) for running operator CI on OpenShift CI.

The [release branch is here](https://github.com/porridge/release/pull/new/ROX-9565-operator-on-prow) and a PR will be created once this part is merged.

## Checklist

## Testing Performed

Roughly the same changes (module some debug output and comments) were tested in https://github.com/openshift/release/pull/30630

### DIFF
--- a/.openshift-ci/build/build-operator-artifacts.sh
+++ b/.openshift-ci/build/build-operator-artifacts.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+# Execute all build steps required to create the operator bundle.tar.gz and
+# scripts used in image/rhel/Dockerfile
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
+source "$ROOT/scripts/ci/lib.sh"
+
+set -euo pipefail
+
+build_operator_bundle_and_binary() {
+    # avoid a -dirty tag
+    info "Reset to remove Dockerfile modification by OpenShift CI"
+    git restore .
+    git status
+
+    info "Current Status:"
+    "$ROOT/status.sh" || true
+
+    openshift_ci_mods
+
+    info "Downloading go modules"
+    # We need to download dependencies before invoking the `manifests` target (depended on by `bundle`).
+    # Otherwise we get the cryptic "Missing value for flag: -I" error when running protoc.
+    go mod download
+
+    info "Preparing bundle sources and smuggled status.sh file"
+    # TODO(ROX-11889): get rid of the SILENT= once we gain some confidence (after release 3.72?)
+    make -C operator bundle bundle-post-process smuggled-status-sh SILENT=
+
+    info "Making a copy of the built bundle sources in a magically named directory that will be used instead of the bundle image."
+    # The hacked opm tool will first see if a directory named as the reference exists, and if so, use its content as if it's an unpacked image of that name.
+    # TODO(ROX-11889): get rid of or upstream this hack in a nicer way
+    # Because the hack needs the directory to be named exactly the same as the image specification, this cannot be placed
+    # in the "build/" directory.
+    bundle_source_parent="operator/$(make --quiet default-image-registry)"
+    mkdir -p "${bundle_source_parent}"
+    cp -a operator/build/bundle "${bundle_source_parent}/stackrox-operator-bundle:v$(make --quiet --no-print-directory -C operator tag)"
+
+    info "Preparing bundle index sources"
+    # TODO(ROX-11889): get rid of the SILENT= once we gain some confidence (after release 3.72?)
+    make -C operator index-build SKIP_INDEX_DOCKER_BUILD=--skip-build SILENT=
+}
+
+build_operator_bundle_and_binary
--- a/.openshift-ci/dispatch.sh
+++ b/.openshift-ci/dispatch.sh
@@ -31,7 +31,7 @@ case "$ci_job" in
         registry_ro_login "quay.io/rhacs-eng"
         # NAMESPACE is injected by OpenShift CI for the cluster that is running the
         # tests but this can have side effects for operator tests due to its use as
-        # the default namespace e.g. in opertor-sdk scorecard.
+        # the default namespace e.g. in operator-sdk scorecard.
         # Cannot use `openshift_ci_e2e_mods` for the following, since it also nukes KUBECONFIG,
         # which we need.
         if [[ -n "${NAMESPACE:-}" ]]; then
--- a/.openshift-ci/dispatch.sh
+++ b/.openshift-ci/dispatch.sh
@@ -27,6 +27,18 @@ case "$ci_job" in
     gke-qa-e2e-tests|gke-nongroovy-e2e-tests|gke-upgrade-tests|gke-ui-e2e-tests)
         openshift_ci_e2e_mods
         ;;
+    openshift-*-operator-e2e-tests)
+        registry_ro_login "quay.io/rhacs-eng"
+        # NAMESPACE is injected by OpenShift CI for the cluster that is running the
+        # tests but this can have side effects for operator tests due to its use as
+        # the default namespace e.g. in opertor-sdk scorecard.
+        # Cannot use `openshift_ci_e2e_mods` for the following, since it also nukes KUBECONFIG,
+        # which we need.
+        if [[ -n "${NAMESPACE:-}" ]]; then
+            export OPENSHIFT_CI_NAMESPACE="$NAMESPACE"
+            unset NAMESPACE
+        fi
+        ;;
 esac
 
 if [[ "$ci_job" =~ e2e|upgrade ]]; then
--- a/.openshift-ci/dispatch.sh
+++ b/.openshift-ci/dispatch.sh
@@ -28,16 +28,7 @@ case "$ci_job" in
         openshift_ci_e2e_mods
         ;;
     openshift-*-operator-e2e-tests)
-        registry_ro_login "quay.io/rhacs-eng"
-        # NAMESPACE is injected by OpenShift CI for the cluster that is running the
-        # tests but this can have side effects for operator tests due to its use as
-        # the default namespace e.g. in operator-sdk scorecard.
-        # Cannot use `openshift_ci_e2e_mods` for the following, since it also nukes KUBECONFIG,
-        # which we need.
-        if [[ -n "${NAMESPACE:-}" ]]; then
-            export OPENSHIFT_CI_NAMESPACE="$NAMESPACE"
-            unset NAMESPACE
-        fi
+        operator_e2e_test_setup
         ;;
 esac
 
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -67,7 +67,11 @@ SCORECARD_STORAGE_IMAGE ?= $(SCORECARD_STORAGE_IMAGE_BASE):$(SCORECARD_STORAGE_I
 SCORECARD_WAIT_TIME_DURATION ?= 5m
 
 # SCORECARD_ARGS should be passed to any invocation of "$(OPERATOR_SDK) scorecard".
-SCORECARD_ARGS ?= --storage-image="$(SCORECARD_STORAGE_IMAGE)" --wait-time="$(SCORECARD_WAIT_TIME_DURATION)" --verbose
+# TODO(ROX-11889): The `--namespace default` is a workaround for failures such as the following when running in a pod:
+# TODO(ROX-11889):   Error: error running tests error creating ConfigMap namespaces "ci-op-05qx07r3" not found
+# TODO(ROX-11889): These seem to happen despite unsetting the $NAMESPACE environment variable.
+# TODO(ROX-11889): Figure out where scorecard gets the namespace name from, elliminate this source, and remove the flag here.
+SCORECARD_ARGS ?= --storage-image="$(SCORECARD_STORAGE_IMAGE)" --wait-time="$(SCORECARD_WAIT_TIME_DURATION)" --verbose --namespace default
 
 # TEST_NAMESPACE is where the operator is installed for e2e tests by CI.
 TEST_NAMESPACE ?= stackrox-operator
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -430,7 +430,7 @@ bundle-test-image: operator-sdk ## Run scorecard tests against bundle image in t
 	$(OPERATOR_SDK) scorecard $(SCORECARD_ARGS) $(BUNDLE_IMG)
 
 .PHONY: index-build
-index-build: bundle-post-process ## Build Index (a.k.a. Catalog) image with the bundle. Bundle image must be pushed beforehand.
+index-build: bundle-post-process ## Build Index (a.k.a. Catalog) image with the bundles. Bundle image must be pushed beforehand.
 # If the bundle has `replaces:` attribute in CSV, then the one it replaces has to already be in the index. Otherwise,
 # validation of the index fails because more than one version is identified as the head of the channel. An example of an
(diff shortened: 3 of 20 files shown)


scripts/ci/jobs/openshift_4_operator_e2e_tests.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env -S python3 -u
+
+"""
+Run operator e2e tests in a openshift 4 cluster provided via a hive cluster_claim.
```
SimonBaeumer, line 4: Do you documentation on this? Honestly I don't understand it.

scripts/ci/lib.sh
```diff
@@ -202,6 +202,103 @@ push_main_image_set() {
...
+        local v
+        for image in "${operator_image_set[@]}"; do
+            if [[ "${image}" != "stackrox-operator" ]]; then
+                # Only the bundle and index image tags have the v prefix.
+                v="v"
+            else
+                v=""
+            fi
+            oc image mirror "${operator_image_srcs[$idx]}" "${registry}/${image}:${v}${tag}"
+            (( idx++ )) || true
```
SimonBaeumer, line 268: Why is this function the only one which has an index in its loop?

scripts/ci/lib.sh
```diff
@@ -202,6 +202,103 @@ push_main_image_set() {
...
+        done
+    }
+
+    if [[ "$brand" == "STACKROX_BRANDING" ]]; then
+        local destination_registries=("quay.io/stackrox-io")
+    elif [[ "$brand" == "RHACS_BRANDING" ]]; then
+        local destination_registries=("quay.io/rhacs-eng")
```
SimonBaeumer, line 275: Does it make sense to have the registry repos configurable by env variables?

scripts/ci/lib.sh
```diff
@@ -202,6 +202,103 @@ push_main_image_set() {
...
+    else
+        die "$brand is not a supported brand"
+    fi
+
+    local tag
+    tag="$(make --quiet -C operator tag)"
```
SimonBaeumer, line 281: Why are two lines needed here?

.openshift-ci/build/build-operator-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+# scripts used in image/rhel/Dockerfile
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
+source "$ROOT/scripts/ci/lib.sh"
+
+set -euo pipefail
+
+build_operator_bundle_and_binary() {
+    # avoid a -dirty tag
+    info "Reset to remove Dockerfile modification by OpenShift CI"
```
SimonBaeumer, line 13: How does the CI adjust the dockerfiles? 🤔 
This seems odd to me tbh

SimonBaeumer (approved):
LGTM, good job @porridge !
Some minor comments, imho you can also address them in another PR to get this one merged.

.openshift-ci/build/build-operator-artifacts.sh
mtesseract: ci-operator has this mechanism of injecting images into the `FROM ...` directives. Seems to be common practice in OSCI world, we also use it in asc-fleet-manager. But, it is also news to me that this works by patching the unpacked source code and leaving it like that.

scripts/ci/lib.sh
mtesseract: To not silently swallow any failures. Shellcheck also warns about that.

scripts/ci/lib.sh
mtesseract: Well, we need credentials for pushing and these are passed in via the CI vault and they are, of course, for a fixed registry. So, in any case, changing the registry would not be simply a matter of changing some registry env variable.

What I attempted the other day was also putting the registry right next to username & password into the vault, so they form a group, like:
```
IMAGE_PUSH_REGISTRY_NAME
IMAGE_PUSH_REGISTRY_USERNAME
IMAGE_PUSH_REGISTRY_PASSWORD
```

But the downside of this is that whenever you would like to output someothing like "Pushing $IMG to $REGISTRY", the registry will be redacted by OSCI. So, this is also not ideal. I would say, under these circumstances it's not an issue to hardcode the registry.

These are just my 2c.

operator/hack/build-index-image.sh
```diff
@@ -2,9 +2,21 @@
...
+# TODO(ROX-11889): named like an image specification. This way we can build the bundle and index sources in the same step
+# TODO(ROX-11889): followed by a parallel build of the bundle and index images, without the need for an intermediate bundle
+# TODO(ROX-11889): image push. This is important in OpenShift CI, where the flow is "all images built first,
+# TODO(ROX-11889): followed by all images pushed together".
+# TODO(ROX-11889): Either upstream the hack or change the flow.
+OPM_VERSION="1.21.0-render-dir"
+# Normally the same as OPM_VERSION. Here we use the upstream version for serving the index,
+# since the render-dir hack only affects the build. This helps us avoid the hassle of hosting the OPM image.
+OPM_IMAGE_VERSION="1.21.0"
+OPM_ORG="porridge"
```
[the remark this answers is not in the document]
mtesseract: This is a dirty hack, because upstream opm didn't work for us.

.openshift-ci/build/build-operator-artifacts.sh
gavin-stackrox: CI updates the `FROM`(s)

scripts/ci/lib.sh
mtesseract: for example:
```
➜  stackrox git:(porridge/ROX-9565-operator-on-prow) ✗ bash -c 'set -e; export foo=$(false); echo "error swallowed"'
error swallowed
➜  stackrox git:(porridge/ROX-9565-operator-on-prow) ✗ bash -c 'set -e; foo=$(false); export foo; echo "error swallowed"'
➜  stackrox git:(porridge/ROX-9565-operator-on-prow) ✗ 
```

Same for `local` instead of `export`.

.openshift-ci/dispatch.sh
```diff
@@ -27,6 +27,18 @@ case "$ci_job" in
     gke-qa-e2e-tests|gke-nongroovy-e2e-tests|gke-upgrade-tests|gke-ui-e2e-tests)
         openshift_ci_e2e_mods
         ;;
+    openshift-*-operator-e2e-tests)
+        registry_ro_login "quay.io/rhacs-eng"
```
gavin-stackrox, line 31: can all this go into a `operator_e2e_test_setup()` (or something) in lib.sh?

operator/Makefile
```diff
@@ -67,7 +70,11 @@ SCORECARD_STORAGE_IMAGE ?= $(SCORECARD_STORAGE_IMAGE_BASE):$(SCORECARD_STORAGE_I
 SCORECARD_WAIT_TIME_DURATION ?= 5m
 
 # SCORECARD_ARGS should be passed to any invocation of "$(OPERATOR_SDK) scorecard".
-SCORECARD_ARGS ?= --storage-image="$(SCORECARD_STORAGE_IMAGE)" --wait-time="$(SCORECARD_WAIT_TIME_DURATION)" --verbose
+# TODO(ROX-11889): The `--namespace default` is a workaround for failures such as the following when running in a pod:
+# TODO(ROX-11889):   Error: error running tests error creating ConfigMap namespaces "ci-op-05qx07r3" not found
+# TODO(ROX-11889): These seem to happen despite unsetting the $NAMESPACE environment variable.
+# TODO(ROX-11889): Figure out where scorecard gets the namespace name from, elliminate this source, and remove the flag here.
+SCORECARD_ARGS ?= --storage-image="$(SCORECARD_STORAGE_IMAGE)" --wait-time="$(SCORECARD_WAIT_TIME_DURATION)" --verbose --namespace default
```
mtesseract, line 77: Shouldn't this be `--storage-image`?

scripts/ci/jobs/push-images.sh
```diff
@@ -44,6 +44,10 @@ push_images() {
         push_context="merge-to-master"
     fi
 
+    # TODO(ROX-11889): make this unconditional once the openshift/release side is ready
```
gavin-stackrox, line 47: nit: shouldn't operator be after main

scripts/ci/lib.sh
gavin-stackrox: thats a `shellcheck` suggestion: https://www.shellcheck.net/wiki/SC2155

gavin-stackrox (changes requested):
Looks good. Some minor nits. I assume that this passed in the rehearsal PR somewhere and a link to a log would be useful.

operator/Makefile
```diff
@@ -67,7 +70,11 @@ SCORECARD_STORAGE_IMAGE ?= $(SCORECARD_STORAGE_IMAGE_BASE):$(SCORECARD_STORAGE_I
 SCORECARD_WAIT_TIME_DURATION ?= 5m
 
 # SCORECARD_ARGS should be passed to any invocation of "$(OPERATOR_SDK) scorecard".
-SCORECARD_ARGS ?= --storage-image="$(SCORECARD_STORAGE_IMAGE)" --wait-time="$(SCORECARD_WAIT_TIME_DURATION)" --verbose
+# TODO(ROX-11889): The `--namespace default` is a workaround for failures such as the following when running in a pod:
+# TODO(ROX-11889):   Error: error running tests error creating ConfigMap namespaces "ci-op-05qx07r3" not found
+# TODO(ROX-11889): These seem to happen despite unsetting the $NAMESPACE environment variable.
+# TODO(ROX-11889): Figure out where scorecard gets the namespace name from, elliminate this source, and remove the flag here.
```
mtesseract, line 76: Could it be that it is set in the default context in the kubeconfig? As in

```
- context:
  [...]
    namespace: ...
  [...]
```

scripts/ci/lib.sh
```diff
@@ -202,6 +202,103 @@ push_main_image_set() {
...
+    if [[ "$#" -ne 2 ]]; then
+        die "missing arg. usage: push_operator_image_set <push_context> <brand>"
+    fi
+
+    local push_context="$1"
+    local brand="$2"
+
+    local operator_image_set=("stackrox-operator" "stackrox-operator-bundle" "stackrox-operator-index")
+    if is_OPENSHIFT_CI; then
+        local operator_image_srcs=("$OPERATOR_IMAGE" "$OPERATOR_BUNDLE_IMAGE" "$OPERATOR_BUNDLE_INDEX_MAGE")
```
mtesseract, line 217: Should this also be initialized somehow in the non-OSCI case?

scripts/ci/lib.sh
mtesseract: Looks like we need to process `operator_image_set` and `operator_image_srcs` in parallel. So we use the `idx` for indexing in the second array.

mtesseract (approved):
Looking good.

scripts/ci/lib.sh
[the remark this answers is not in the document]
gavin-stackrox: > Why the `|| true`?

I think thats from `set -e`. 
```
$ (idx=0; (( idx++ )); echo incremented)
incremented
$ (set -e; idx=0; (( idx++ )); echo incremented)
$
```

operator/hack/build-index-image.sh
```diff
@@ -91,7 +103,7 @@ function fetch_opm() {
   local -r arch=$(go env GOARCH) || true
 
   OPM="${BASE_DIR}/bin/opm-${OPM_VERSION}"
-  "${SCRIPT_DIR}/get-github-release.sh" --to "${OPM}" --from "https://github.com/operator-framework/operator-registry/releases/download/v${OPM_VERSION}/${os_name}-${arch}-opm"
+  "${SCRIPT_DIR}/get-github-release.sh" --to "${OPM}" --from "https://github.com/${OPM_ORG}/operator-registry/releases/download/v${OPM_VERSION}/${os_name}-${arch}-opm"
```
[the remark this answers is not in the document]
porridge: > Does this take a fixed script from your personal github account?

It takes a binary built from the fork.

> WDYTH we should fork the repositories into the stack rocks organization instead of our personal accs?

In general I agree, though there is precedent in `go.mod` for personal orgs.
As mentioned in the TODO above, this is temporary.

scripts/ci/jobs/openshift_4_operator_e2e_tests.py
porridge: I'm just trying to be consistent with the another script in the same directory from which I cargo-culted.

scripts/ci/lib.sh
porridge: What I [alluded (probably not strongly enough) to in the commit description](https://github.com/stackrox/stackrox/pull/2471/commits/5ae1bc4a5ef1a102df992e03b5246bbd8412be43) is that this whole function is an almost exact copy of `push_main_image_set`, so I'm just being consistent with what the other does.

Unfortunately due to bash limitations **and** the inconsistent versioning scheme of our images, it would be a bit arcane to try and reuse code between them.

We can think about improving both of them, but I'd consider this to be out of scope for this PR.

scripts/ci/lib.sh
porridge: Again, I just cargo culted, thanks @mtesseract for the explanation, TIL.

operator/hack/build-index-image.sh
```diff
@@ -2,9 +2,21 @@
 # Builds operator index image.
 
 set -eou pipefail
+# TODO(ROX-11889): Remove the -x once we gain some confidence (after 3.72 release?)
+set -x
```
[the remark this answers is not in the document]
porridge: I double-checked there are none [in the log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/30630/rehearse-30630-pull-ci-stackrox-stackrox-porridge-ROX-9569-operator-prow-upgrade-images/1549746175630905344/artifacts/build-logs/operator-artifacts.log).

scripts/ci/jobs/push-images.sh
porridge: I explain the reason for this order in the commit description.

>  I assume that this passed in the rehearsal PR somewhere and a link to a log would be useful.

As mentioned in the PR description:
> Roughly the same changes (module some debug output and comments) were tested in https://github.com/openshift/release/pull/30630

I avoided doing this for this PR, because it would cause the operator images to be built both by prow and by circleci and clash. OTOH if I made a change here to remove the CircleCI operator jobs in this PR, then we wouldn't know if this PR didn't break them. But I guess that's OK since they are scheduled for disabling today.

operator/Makefile
porridge: The kubeconfig used by scorecard is the kubeconfig for the leased cluster, which cannot possibly mention the CI namespace. I suspect it's trying to be overly clever by looking up the namespace from some kubernetes downward API.

porridge:
Created a [temporary rehearsal PR](https://github.com/openshift/release/pull/30674).
Still not opening a PR for the `openshift/release` [branch](https://github.com/porridge/release/pull/new/ROX-9565-operator-on-prow) yet because it would fail anyway until this PR is merged.

mtesseract (approved):

porridge:
@gavin-stackrox I think I addressed all your comments and [ran a rehearsal](https://github.com/openshift/release/pull/30674) (which revealed I was missing one more `export`), PTAL

gavin-stackrox (approved):

PR #2472, opened 2022-07-20
ROX-11318: Adding side panel to policy categories page

alwayshooin:
## Description

As title suggests -- the height of the side panel form needs to be polished (was running into issues because the form is apparently a grid)

Opening/closing of the side panel should work as expected.

image

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
+++ b/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
@@ -1,13 +1,15 @@
 import React from 'react';
 import { SimpleList, SimpleListItem } from '@patternfly/react-core';
 
+import { PolicyCategory } from 'types/policy.proto';
+
 type PolicyCategoriesListProps = {
     policyCategories: {
         id: string;
         name: string;
         isDefault: boolean;
     }[];
-    setSelectedCategory: (selectedCategory: string) => void;
+    setSelectedCategory: (selectedCategory: PolicyCategory) => void;
 };
 
 function PolicyCategoriesList({
@@ -17,7 +19,7 @@ function PolicyCategoriesList({
     return (
         <SimpleList onSelect={() => {}}>
             {policyCategories.map((category) => {
-                const { id, name, isDefault } = category;
+                const { id, name } = category;
                 return (
                     <SimpleListItem
                         key={id}
--- a/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
+++ b/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
@@ -7,21 +7,30 @@ type PolicyCategoriesListProps = {
         name: string;
         isDefault: boolean;
     }[];
+    setSelectedCategory: (selectedCategory: string) => void;
 };
 
-function PolicyCategoriesList({ policyCategories }: PolicyCategoriesListProps) {
+function PolicyCategoriesList({
+    policyCategories,
+    setSelectedCategory,
+}: PolicyCategoriesListProps) {
     return (
         <SimpleList onSelect={() => {}}>
-            {policyCategories.map(({ id, name, isDefault }) => (
-                <SimpleListItem
-                    key={id}
-                    onClick={() => {}}
-                    isActive={false}
-                    componentProps={{ disabled: isDefault }}
-                >
-                    {name}
-                </SimpleListItem>
-            ))}
+            {policyCategories.map((category) => {
+                const { id, name, isDefault } = category;
+                return (
+                    <SimpleListItem
+                        key={id}
+                        onClick={() => {
+                            setSelectedCategory(category);
+                        }}
+                        isActive={false}
+                        // componentProps={{ disabled: isDefault }}
+                    >
+                        {name}
+                    </SimpleListItem>
+                );
+            })}
         </SimpleList>
     );
 }
--- a/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
+++ b/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
@@ -57,15 +57,15 @@ function PolicyCategoriesListSection({
                 <FlexItem flex={{ default: 'flex_1' }}>
                     <PageSection isFilled variant="light" className="pf-u-h-100">
                         <Flex direction={{ default: 'column' }}>
-                            <Flex
-                                justifyContent={{ default: 'justifyContentSpaceBetween' }}
-                                fullWidth={{ default: 'fullWidth' }}
-                            >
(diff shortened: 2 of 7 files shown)


ui/apps/platform/src/Containers/Policies/Table/PoliciesTablePage.tsx
```diff
@@ -10,10 +10,6 @@ import {
     AlertVariant,
     Divider,
     Button,
-    Flex,
-    Toolbar,
-    ToolbarContent,
-    ToolbarItem,
```
alwayshooin, line 16: not sure how these were passing through before 🤔

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
```diff
@@ -1,29 +1,38 @@
...
+            {policyCategories.map((category) => {
+                const { id, name } = category;
+                return (
+                    <SimpleListItem
+                        key={id}
+                        onClick={() => {
+                            setSelectedCategory(category);
+                        }}
+                        isActive={false}
+                        // componentProps={{ disabled: isDefault }}
```
alwayshooin, line 30: will add this back once the API is working as expected

ui/apps/platform/src/Containers/Policies/Table/PoliciesTablePage.tsx
pedrottimark: Uh, oh. Maybe CI not running lint? There is also `console.log` in PolicyCategorySidePanel.tsx file

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
```diff
@@ -1,27 +1,38 @@
 import React from 'react';
 import { SimpleList, SimpleListItem } from '@patternfly/react-core';
 
+import { PolicyCategory } from 'types/policy.proto';
```
pedrottimark, line 4: If you rebase from master, you can import `PolicyCategory` from services/policyCategoryService.ts instead.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
```diff
@@ -1,33 +1,32 @@
...
-        name: category,
-        isDefault: true,
-    }));
-    const customPolicyCategories = allPolicyCategories.filter(({ isDefault }) => !isDefault);
-    const defaultPolicyCategories = allPolicyCategories.filter(({ isDefault }) => isDefault);
+function PolicyCategoriesListSection({
+    policyCategories,
+    addToast,
+}: PolicyCategoriesListSectionProps) {
+    const [selectedCategory, setSelectedCategory] = useState<PolicyCategory>();
```
pedrottimark, line 22: What do you think about `selectedCategoryId` instead to avoid any problem with object identity?

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategorySidePanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+                            <ActionGroup>
+                                <Button
+                                    variant="primary"
+                                    isDisabled={!dirty}
+                                    onClick={() => handleSubmit()}
+                                >
+                                    Save
+                                </Button>
+                                <Button variant="secondary" onClick={clearSelectedCategory}>
+                                    Cancel
```
pedrottimark, line 97: A thought, although not sure whether it is a good thought, is whether text depends on whether or not edited?

Cancel versus Close? Somehow, it caused me to pause and think twice how to close the panel without changes.

ui/apps/platform/src/Containers/Policies/Table/PoliciesTablePage.tsx
pedrottimark: Until we get to the bottom of it, run `yarn lint` in ui folder before you merge.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
vjwilson: I'm working on this PR while Linda is out, and I'm confused by this. I thought we were moving our type definitions to the `types/<x>.proto.ts` files, where each file mapped to a backend API proto?

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
vjwilson: Since any selected category is passed down to the `<PolicyCategorySidePanel>` below, and this will only be a reference to the object, it seems that only storing the ID field would make more work to get the category to pass down, without much performance benefit. Am I missing some other advantage?

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
pedrottimark: The frontend file corresponds to backend file which declares the the type, so all of us know where to compare. A recent trend is for service to declare it response types.

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesListSection.tsx
pedrottimark: The question is about potential loss of current selection if object identity changes for selected category.

For example, pages often get all entities after any change. Even if not, after update to selected entity.

vjwilson: I think we should defer this until we see how big a problem it would be.

ui/apps/platform/src/services/PolicyCategoriesService.ts
```diff
@@ -1,15 +1,8 @@
...
 const policyCategoriesUrl = '/v1/policycategories';
 
-export type PolicyCategory = {
-    id: string;
-    // central/policycategory/service/service_impl.go
-    // policy category must have a name between 5 and 128 characters long with no new lines or dollar signs
-    name: string;
-    isDefault: boolean;
-};
-
```
pedrottimark, line 12: This is an incorrect change, but we can sort it out in the next step:

* This type (with comment that I found in go implementation code) corresponds to https://github.com/stackrox/stackrox/blob/master/proto/api/v1/policy_category_service.proto#L14-L18
* When a service defines a response type, which is the recent pattern in 2022, it is the backend source of truth, even if there is a corresponding storage type in https://github.com/stackrox/stackrox/blob/master/proto/storage/policy_category.proto#L11-L15
* There is no `PolicyCategory` type in https://github.com/stackrox/stackrox/blob/master/proto/storage/policy.proto

pedrottimark (approved):

PR #2473, opened 2022-07-20
ROX-11564 replay tests with alerts

lvalerom:
## Description

This PR aims to provided testing for sensor's re-sync behaviour using pre-recorded k8s events.

The events were recorded using the local-sensor tool.

Create the policies.json file containing the policies:
```
curl -k -u admin:$ROX_ADMIN_PASSWORD -H "Content-Type:application/json" https://$ROX_CENTRAL_ADDRESS/v1/policies > $tmpf
# and then
# for each id
curl -k -u admin:$ROX_ADMIN_PASSWORD -H "Content-Type:application/json" https://$ROX_CENTRAL_ADDRESS/v1/policies/$id >> $fname 
```

Create the trace.jsonl file containing the k8s events.
```
go run tools/local-sensor/main.go -record -record-out=trace.jsonl -resync=0s
```
Create the central-out.bin file containing sensor's output events.
```
go run tools/local-sensor/main.go -replay -replay-in=trace.jsonl -resync=10s -delay=0s -format=raw -central-out=central-out.bin -with-policies=policies.json
```

## Checklist

## Testing Performed

Manually testing with:
```
LOGLEVEL=panic go test -race -count=1 github.com/stackrox/rox/sensor/tests/replay
```
Disabling the re-sync by changing resyncTime := 1 * time.Second to resyncTime := 0 * time.Second will make the test fail.


lvalerom:
Part of this PR is addressed here #2483

vikin91:
I see some rebase is needed here - please ping me for a review once the conflicts are resolved

lvalerom:
> I see some rebase is needed here - please ping me for a review once the conflicts are resolved

Yeah. The idea is to merge #2483 first and then rebase this one.

sensor/debugger/k8s/fake.go
```diff
@@ -216,6 +219,188 @@ func (f *FakeEventsManager) Init() {
 		cronJobKind:               &batchv1.CronJob{},
 		podKind:                   &corev1.Pod{},
 	}
+	f.resourceListMap = map[string]func(reflect.Value, func(string) error) error{
```
fredrb, line 222: If I understand this code correctly, the only reason why we have this map is to call `GetName()` on each element on the `Items` slice. Since every resource list has the same interface (`.Items` + `.GetName()`), and we're only interested in the `string` value from the resource itself, I suggest that we make this generic. 

[Here's an example from playground](https://go.dev/play/p/90kOgXPFkWU). 

I suggest writing a single reflect-based function that iterates on `.Items` and calls `.GetName()` on each element:
```golang
func getNameGeneric(obj reflect.Value, f func(string) error) error {
	errorList := errorhelpers.NewErrorList("resource list function")
	items := obj.FieldByName("Items")
	if !items.IsValid() {
		errorList.AddError(errors.New("invalid kind: missing `Items` property"))
	} else {
		for i := 0; i < items.Len(); i++ {
			getNameFn := items.Index(i).MethodByName("GetName")
			returnVals := getNameFn.Call([]reflect.Value{})
			if err := f(returnVals[0].String()); err != nil {
				errorList.AddError(err)
			}
		}	
	}
	return errorList.ToError()
}
```

sensor/debugger/k8s/fake.go
```diff
@@ -436,6 +621,68 @@ func (f *FakeEventsManager) waitOnMode(events []string) error {
 	return nil
 }
 
+// execFuncToAllResourcesOfKind executes a given function to all resources of a given kind
```
fredrb, line 624: I understand that maybe the goal was to provide a callback to each resource object . However, only the name seems passed. Which is enough for removing namespaces. I suggest rewording this.
Additionally, you could already provide an `interface{}` object parameter to the `execFunc` callback. 

IMO this would read better:
```suggestion
// forEachResourceName iterates on every resource of given kind and call `callbackFn` for each resource object name
```
Or `forEachResource` if you decide on passing the entire object to the callback.

fredrb: Keep in mind that if you decide to pass the entire object you could remove the`GetName()` reflection above. It would be the responsibility of the caller to cast the `interface{}` into whatever resource is being listed.

sensor/tests/replay/replay_test.go
```diff
@@ -293,6 +323,23 @@ func readSensorOutputFile(fname string) ([]*central.MsgFromSensor, error) {
 	return msgs, nil
 }
 
+func getAlerts(messages []*central.MsgFromSensor) map[string]map[string]*storage.Alert {
+	events := make(map[string]map[string]*storage.Alert)
+	for _, msg := range messages {
+		event := msg.GetEvent()
+		if event.GetAlertResults() != nil {
+			if event.GetAlertResults().GetDeploymentId() != "" {
+				alerts := make(map[string]*storage.Alert)
```
fredrb, line 332: Maybe add the map size if you know it ahead of time:
```golang
alertResults := event.GetAlertResults().GetAlerts()
alerts := make(map[string]*storage.Alert, len(alertResults))
for _, a := range alertResults {
       alerts[a.GetPolicy().GetId()] = a
}
```

fredrb (changes requested):
Requesting changes due to the reflection suggestion. Let me know if I missed something or if there is a reason not to do it.

sensor/debugger/k8s/fake.go
lvalerom: Yeah. I changed your suggestion a little bit. Let me know what you think.

sensor/debugger/k8s/fake.go
lvalerom: Done. I've decided to pass the reflected value of the resource instead of the name. Let me know what you think.

sensor/tests/replay/replay_test.go
lvalerom: 👍

fredrb (approved):

sensor/debugger/k8s/fake.go
```diff
@@ -436,6 +437,123 @@ func (f *FakeEventsManager) waitOnMode(events []string) error {
...
+	}
+	errInt := returnVals[len(returnVals)-1].Interface()
+	if errInt != nil {
+		return errInt.(error)
+	}
+
+	errorList := errorhelpers.NewErrorList("for each resource")
+	items := reflect.Indirect(returnVals[0]).FieldByName("Items")
+	if !items.IsValid() {
+		errorList.AddError(errors.New("invalid kind: missing `Items` property"))
```
vikin91, line 458: Minor: we could return here and save the `else` block

sensor/debugger/k8s/fake.go
```diff
@@ -436,6 +437,123 @@ func (f *FakeEventsManager) waitOnMode(events []string) error {
...
+	return errorList.ToError()
+}
+
+// getNameFromObjectMeta returns the name of the resource
+func getNameFromObjectMeta(obj reflect.Value) (string, error) {
+	objMeta := obj.FieldByName("ObjectMeta")
+	if !objMeta.IsValid() {
+		return "", errors.New("resource does not have the ObjectMeta field")
+	}
+	objMetaPtr := objMeta.Addr()
```
vikin91, line 476: `objMetaPtr` is used only in the next line - maybe a candidate to remove?

sensor/debugger/k8s/fake.go
```diff
@@ -436,6 +437,123 @@ func (f *FakeEventsManager) waitOnMode(events []string) error {
...
+			cl = reflect.ValueOf(clFunc(ns))
+			if err := f.forEachResource(cl, func(resource reflect.Value) error {
+				name, err := getNameFromObjectMeta(resource)
+				if err != nil {
+					return err
+				}
+				retVals := runOp(removeAction, cl, reflect.ValueOf(name))
+				if len(retVals) == 0 {
+					return fmt.Errorf("expected 1 or 2 values from %s. Received: %d", removeAction, len(retVals))
+				}
```
vikin91, line 543: I guess `len(retVals)` cannot be 3 or more? If it can, then maybe we should change it to

```suggestion
				if len(retVals) < 1 || len(retVals) > 2 {
					return fmt.Errorf("expected 1 or 2 values from %s. Received: %d", removeAction, len(retVals))
				}
```

vikin91 (approved):
Looks good, I found just minor and fully optional improvements

sensor/debugger/k8s/fake.go
lvalerom: Done

sensor/debugger/k8s/fake.go
lvalerom: Done

sensor/debugger/k8s/fake.go
lvalerom: Done!

PR #1419, opened 2022-04-25
Update PatternFly 2022.05 dependencies in UI

pedrottimark:
## Description

https://www.patternfly.org/v4/developer-resources/release-notes/#2022.05-release-notes-2022-04-22

https://www.patternfly.org/v4/developer-resources/release-notes/html#2022.05-release-notes-2022-04-20

## Checklist
- [x] Investigated and inspected CI test results
- [x] ~Unit test and regression tests added~

## Testing Performed

PR #2475, opened 2022-07-20
Update PF dependencies to 2022.07

dvail:
## Description

Upgrading directly to 2022.08 is blocked by
https://github.com/patternfly/patternfly-react/issues/7722 (which is fixed in `@patternfly/react-core` version 4.225.11 and will be part of the next PF release.)

Related to https://github.com/stackrox/stackrox/pull/1419

## Checklist

## Testing Performed

Manual spot check by using the app, focusing on PatternFly pages.


vjwilson:
Looks like this PF upgrade might have really broken some things, based on the ui-e2e tests.

dvail:
Yikes! I'll take a look and see how serious it is. Hoping this is just an internal PF change based on some selectors (I see a lot of failures failing to find `td.pf-c-table__action`) and not a more functional issue.

dvail:
Two issues introduced by this update that should be fixed now. (Passing locally, just pushed to CI)

- In menu dropdowns, the `role="menuitem"` attribute is now placed on the _child_ of the `li` instead of the `li` itself. The `li` gets `role="none"`. (Unless the child is a checkbox, then this behavior is reversed). Introduced by PF change https://github.com/patternfly/patternfly-react/pull/7368/files#diff-dba8c84d269ff65dc2f536cdc0fcf812decfd80e52d7f4674c9fe86313422be2

- The need to press 'downarrow' followed by 'enter' when attempting to create a tag is no longer. This reverts [this change](https://github.com/stackrox/stackrox/pull/980), which was the original fix for a PF behavior change that broke e2e tests. I haven't tracked down exactly where they updated this yet to verify, but the behavior is consistent on my machine.

PR #2477, opened 2022-07-20
ROX-11715: Error better when deployment fails

gavin-stackrox:
## Description

This PR:
- Creates a JUNIT test error when StackRox deployment fails [e.g.](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2477/pull-ci-stackrox-stackrox-master-gke-qa-e2e-tests/1549904766564306944)
- Skips Part II QA tests when StackRox deployment fails

## Checklist
  - For commit e37d157 with tag mismatch, the results call out the 'Stackrox_Deployment' failure.
  - Runs fine (except known GCR flake) otherwise.

## Testing Performed

CI is sufficient

### DIFF
--- a/qa-tests-backend/scripts/run-part-1.sh
+++ b/qa-tests-backend/scripts/run-part-1.sh
@@ -32,7 +32,7 @@ test_part_1() {
     remove_existing_stackrox_resources
     setup_default_TLS_certs
 
-    deploy_stackrox 
+    deploy_stackrox
 
     deploy_default_psp
     deploy_webhook_server
--- a/qa-tests-backend/scripts/run-part-1.sh
+++ b/qa-tests-backend/scripts/run-part-1.sh
@@ -24,11 +24,12 @@ test_part_1() {
     remove_existing_stackrox_resources
     setup_default_TLS_certs
 
-    if ! deploy_stackrox; then
+    deploy_stackrox || {
+        local exitstatus="$?"
         echo "deploy_stackrox() failed, will skip part II tests"
         touch "$ROOT/SKIP_PART_II"
-        exit 1
-    fi
+        exit "$exitstatus"
+    }
 
     deploy_default_psp
     deploy_webhook_server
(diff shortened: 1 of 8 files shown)


tests/e2e/lib.sh
```diff
@@ -26,6 +26,14 @@ deploy_stackrox() {
     sensor_wait
 }
 
+deploy_stackrox() {
+    _deploy_stackrox 2>&1 | tee "$TEST_ROOT/deployment_output.txt" || {
```
sbostick, line 30: Fine as is, but wanted to share this snippet in case you haven't seen this construct. Works in bash, zsh, sh.

```
dup_to_log() {
    local transcript="${1:-/tmp/output}"
    exec > >(tee -i "$transcript") 2>&1
}

greeting() {
    dup_to_log "/tmp/output"
    set -x
    echo "hello"
}

echo "foo"
(greeting)
echo "bar"
```

sbostick (approved):
LGTM

qa-tests-backend/scripts/run-part-1.sh
```diff
@@ -24,7 +24,11 @@ test_part_1() {
     remove_existing_stackrox_resources
     setup_default_TLS_certs
 
-    deploy_stackrox
+    if ! deploy_stackrox; then
```
ivan-degtiarenko, line 27:
```suggestion
    if !deploy_stackrox; then
```

scripts/ci/lib.sh
```diff
@@ -1102,6 +1102,29 @@ send_slack_notice_for_failures_on_merge() {
     curl -XPOST -d @- -H 'Content-Type: application/json' "$webhook_url"
 }
 
+save_junit_failure() {
```
ivan-degtiarenko, line 1105: where do we use `save_junit_failure()` other than in `deploy_stackrox()`

gavin-stackrox: nowhere as this is a new function.

tests/e2e/lib.sh
gavin-stackrox: dang. I should have tried this. `_deploy_stackrox 2>&1 | tee "$TEST_ROOT/deployment_output.txt"` clobbers the `exports` in `_deploy_stackrox` because (I guess) it creates a subshell even though not a subprocess.

PR #2479, opened 2022-07-20
ROX-9350 Use fine-grained host paths for compliance mounts

robbycochran:
## Description

The compliance container Is mounting the whole host filesystem, including any PV mounts that the node might be mounted which can prevent the unmounting of PV mounts, as the `compliance` binary will keep the mount points busy. This has been noticed on an OpenShift node when a Pod was mounting a 1TB volume with thousands of files. The compliance container started to scan these files and prevented the unmounting of the PV.

The problem of this behavior has been discussed in detail in https://bugzilla.redhat.com/show_bug.cgi?id=2036055

## Checklist

### DIFF
--- a/image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
@@ -116,8 +116,41 @@ spec:
           name: etc-ssl
         - mountPath: /etc/pki/ca-trust/
           name: etc-pki-volume
-        - mountPath: /host
-          name: host-root-ro
+        - mountPath: /host/etc
+          name: etc-ro
+          readOnly: true
+        - mountPath: /host/proc
+          name: proc-ro
+          readOnly: true
+        - mountPath: /host/opt
+          name: opt-ro
+          readOnly: true
+        - mountPath: /host/run
+          name: run-ro
+          readOnly: true
+        - mountPath: /host/srv
+          name: srv-ro
+          readOnly: true
+        - mountPath: /host/sys
+          name: sys-ro
+          readOnly: true
+        - mountPath: /host/usr
+          name: usr-ro
+          readOnly: true
+        - mountPath: /host/var/lib/kubelet/kubeconfig
+          name: var-lib-kubelet-kubeconfig
+          readOnly: true
+        - mountPath: /host/var/lib/docker
+          name: var-lib-docker
+          readOnly: true
+        - mountPath: /host/var/lib/containers
+          name: var-lib-containers
+          readOnly: true
+        - mountPath: /host/var/log
+          name: var-log
+          readOnly: true
+        - mountPath: /host/var/run
+          name: var-run
           readOnly: true
         - mountPath: /run/secrets/stackrox.io/certs/
           name: certs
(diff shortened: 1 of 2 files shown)


image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
```diff
@@ -116,16 +118,46 @@ spec:
...
+          readOnly: true
+        - mountPath: /host/srv
+          name: srv-ro
+          readOnly: true
+        - mountPath: /host/sys
+          name: sys-ro
+          readOnly: true
+        - mountPath: /host/usr
+          name: usr-ro
+          readOnly: true
```
vikin91, line 141: I am wondering why do we mount now `/usr` when we used to mount only `/usr/lib`. Is this intended?

vikin91 (approved):
Looks good in general - the original issue seems to be addressed.

The PR description does not mention any testing activities - has this been tested?
I also left one question about mounting entire `/usr` instead of `/usr/lib`.

I am marking it green, so to avoid that the TZ difference would block shipping this, but please check my comment.

janisz:
What's the status of this? Is it meant to be included in 3.73

vikin91:
> What's the status of this? Is it meant to be included in 3.73

Friendly ping @robbycochran

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
```diff
@@ -154,9 +210,6 @@ spec:
             path: key.pem
           - key: ca.pem
             path: ca.pem
-      - hostPath:
```
RTann, line 157: note: if we do decide to support language-level vulnerabilities in node-scanning, then will will probably need to mount `/` again, as these packages can, in theory, be anywhere

robbycochran:
@theencee Can you take a look at this PR from the perspective of audit log consumption in the compliance container?

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
robbycochran: Yes that's a concern. The key issue is the fact that the compliance container does long-running mounts. Have we thought about node-scanning in an "on-demand" model, where the node scanning occurs with a short-term mount of the root directory?

theencee:
> @theencee Can you take a look at this PR from the perspective of audit log consumption in the compliance container?

I think it should be good because of this:

```
        - mountPath: /host/var/log
          name: var-log
          readOnly: true
```

Audit log requires [this file path](https://github.com/stackrox/stackrox/blob/master/compliance/collection/auditlog/auditlog.go#L13) mounted: /host/var/log/kube-apiserver/audit.log

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
robbycochran: It is intended. The collector and compliance containers now share volumes. The mount directories for the collector container are not changed, just the volumes used to provide those mounts (previously, collector directly mounted `/usr/lib` via a volume on the host `/usr/lib`, now we mount via `subPath` on the host `/usr` volume mount that is shared with the compliance container

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
RTann: We could consider that. I'm not familiar with this concept at all, but we can discuss more when/if we need to add language-level vulns

PR #2484, opened 2022-07-21
Quickfix: Change `cd` into a pair of `pushd` and `popd` to fix release failures

vikin91:
See https://srox.slack.com/archives/CMH5M8MHN/p1658395416730139 for more
info about why this is a bug

## Description

In the release process, before running `release/scripts/scan-images-with-roxctl.sh` we run `mark_collector_release` and this function does `cd /tmp/collector || exit`. This `cd` "stays" after `mark_collector_release` exits (we do not do `cd -` later), so the parent script is in a wrong directory (collector instead of stackrox) and when it tries to call `make collector-tag` it cannot find the target, because collector's makefile doesn’t have it.

This is a strong argument for using `pushd` and `popd` in the scripts instead of `cd`.

## Checklist

This is a CI-only change.

## Testing Performed

- None, this code is only used when releasing the final upstream tag, so it is also difficult to test int he field. But the changes int the code are so tiny that it should be easy to test it by looking at the code.

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -470,7 +470,7 @@ mark_collector_release() {
     info "Create a branch for the PR"
 
     collector_version="$(cat COLLECTOR_VERSION)"
-    cd /tmp/collector || exit
+    pushd /tmp/collector || exit
     gitbot(){
         git -c "user.name=RoxBot" -c "user.email=roxbot@stackrox.com" \
             -c "url.https://${GITHUB_TOKEN}:x-oauth-basic@github.com/.insteadOf=https://github.com/" \
@@ -502,6 +502,7 @@ mark_collector_release() {
     export CIRCLE_PULL_REQUEST="https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/${JOB_NAME}/${BUILD_ID}"
     export GITHUB_TOKEN_FOR_PRS="${GITHUB_TOKEN}"
     /scripts/create_update_pr.sh "${branch_name}" collector "Update RELEASED_VERSIONS" "Add entry into the RELEASED_VERSIONS file"
+    popd
 }
 
 is_tagged() {


scripts/ci/lib.sh
```diff
@@ -470,7 +470,7 @@ mark_collector_release() {
     info "Create a branch for the PR"
 
     collector_version="$(cat COLLECTOR_VERSION)"
-    cd /tmp/collector || exit
+    pushd /tmp/collector || exit
```
misberner, line 473: If I am not mistaken, this was refactored from a stand-alone script into a bash function; and `cd`s from the script invocation were obviously not propagated back to the caller. Not sure why this refactoring was done, but I think a better way to eliminate the problems introduced by it would be to have the entire function spawn a subshell. I.e., instead of writing `mark_collector_release() { ... }`, do `mark_collector_release() ( ... )`. This ensures the invoking shell will no longer have its state modified by the function, without depending on the user to revert modified state manually on every exit path.

Make sure that the function is properly invoked with `set -e` set or with a `... || die`, because `die` in l.459 and `exit` here will no longer propagate either.

misberner, line 473: Secondary concern, not your change: `|| exit` seems wrong here because it will return silently with a _success_ exit code 0. I think this should be `|| die "Could not cd into tmp directory"` or `|| exit 1` or so.

scripts/ci/lib.sh
```diff
@@ -502,6 +502,7 @@ mark_collector_release() {
     export CIRCLE_PULL_REQUEST="https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/${JOB_NAME}/${BUILD_ID}"
```
misberner, line 502: In the current version, these `export`s get propagated through to the invoking environment, but no longer will after the `(...)` change. This should be perfectly fine, but if after the suggested change you see errors about these vars not being defined, the behavior change might be to blame.

vikin91: I think the exports are only needed for `create_update_pr.sh` cause it was written to run in CircleCI and took those for granted. So the exports are harmless her, but a proper fix would be to eliminate them by making `create_update_pr.sh` CI-independent.

scripts/ci/lib.sh
vikin91: >Not sure why this refactoring was done

I think it was easier to migrate to OSCI that way (provided the time pressure). Gavin would know more about this.

> I think a better way to eliminate the problems introduced by it would be to have the entire function spawn a subshell

We can try this, as this would nicely isolate things. Alternatively, we could also operate with absolute paths without depending on `cd`. I think sub-shell may be cleaner from those two options.

scripts/ci/lib.sh
```diff
@@ -502,6 +502,7 @@ mark_collector_release() {
...
     export GITHUB_TOKEN_FOR_PRS="${GITHUB_TOKEN}"
     /scripts/create_update_pr.sh "${branch_name}" collector "Update RELEASED_VERSIONS" "Add entry into the RELEASED_VERSIONS file"
```
porridge, line 504: How is this supposed to work? Is it meant to be invoked in a container or something?

porridge (approved):
+1 to everything that @misberner says

There are more puzzling things here, like the hardcoded `/tmp/collector` that does not get cleaned up, or the invocation of `/scripts/...` which seems to make strong assumptions about the execution environment, which I'm not aware of.

But as a quick fix without properly refactoring this function, +1

scripts/ci/lib.sh
vikin91: this was supposed to be executed in the CircleCI job. The script assumes that some env variables are given - which is the case only in CircleCI. Here it was patched by defining the vars that it requires.

gavin-stackrox (approved):

scripts/ci/lib.sh
porridge: I meant the fact that it assumes a `scripts` directory right under the filesystem root.

vikin91:
Merging this as it is (a quickfix). I created a followup to refactor the code as discussed in this PR: https://issues.redhat.com/browse/ROX-11988

PR #2486, opened 2022-07-21
[ROX-11908] Cleanup orphaned image dependencies

md2119:
## Checklist

## Testing Performed
Unit

### DIFF
--- a/central/image/datastore/keyfence/singleton.go
+++ b/central/image/datastore/keyfence/singleton.go
@@ -1,0 +1,19 @@
+package keyfence
+
+import (
+	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var (
+	once sync.Once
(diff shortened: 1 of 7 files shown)


central/image/datastore/keyfence/singleton.go
```diff
@@ -0,0 +1,19 @@
+package keyfence
```
md2119, line 1: moved here to break up import cycle

central/image/datastore/store/postgres/store.go
```diff
@@ -108,12 +108,13 @@ func (s *storeImpl) insertIntoImages(
 		cloned.GetCves(),
 		cloned.GetFixableCves(),
 		pgutils.NilOrTime(cloned.GetLastUpdated()),
+		cloned.GetPriority(),
```
md2119, line 111: This column was added recently (#2400) because the field was marked searchable.

theencee (approved):

PR #2495, opened 2022-07-22
ROX-11919: Get rid of the GO_VERSION operator build arg.

porridge:
## Description

Hardcode the go version in the operator `Dockerfile`, rather than require it to be passed as a build argument.

This may seem counter-productive at first, but there is an important detail
outside of this repository. Namely `operator/Dockerfile` is also referenced
from the OpenShift CI configuration, which currently also needs to hardcode
the go version, since the value of the build arg there needs to be provided
literally (there does not seem to be a way to extract it from
`EXPECTED_GO_VERSION`).

So rather than have the Go version be defined in two separate repositories
(here and in openshift/release), we put it in two places in *this* repository,
and add CI- and build-time checks to make sure they are in sync, as well as
cross-references to remember to update them simultaneously.

## Checklist

Not a user-visible change.

## Testing Performed

Manually ran the make target and make sure the output is expected both when values are in and out of sync.
Relying on CI otherwise.

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -1,8 +1,8 @@
 # We have to emulate directory layout as in the repo so that imports in go files work fine.
 ARG roxpath=/workspace/src/github.com/stackrox/rox
-ARG GO_VERSION
 
-FROM golang:${GO_VERSION} as builder
+# Note: the following version needs to be in sync with the ../EXPECTED_GO_VERSION file.
+FROM golang:1.17.2 as builder
 
 # Build the manager binary
 ARG roxpath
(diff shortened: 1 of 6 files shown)


operator/Dockerfile
```diff
@@ -1,8 +1,8 @@
 # We have to emulate directory layout as in the repo so that imports in go files work fine.
 ARG roxpath=/workspace/src/github.com/stackrox/rox
-ARG GO_VERSION
```
porridge, line 3: I vaguely remember `ci-operator` complaining about a `build_arg` which was specified in its configuration but missing in the dockerfile, but I don't know whether it was fatal. But we will soon know if that's the case...

porridge: Looks like it's not an issue.

mtesseract (approved):
looking good to me.

PR #2497, opened 2022-07-22
Dashrews/ROX-11677 ROX-11235 ROX-11922 Initial postgres rollback and upgrade capability

dashrews78:
## Description

The initial cut at the rollback and upgrade code for Postgres.  The idea here is to treat database names in postgres similarly to how we were treating file system directories in rocksdb.  So the initial used database is central_active.  From there we have a few scenarios:

- Restore
    - A restore of a backup is initiated.  
    - The backup bundle will be written to a database called central_restore.  
    - central_active will be renamed to be central_backup in case something goes awry.  
    - central_restore will then be moved to central_active 
    - version checks will be done to see if the version of central supports what was in central_restore
    - If all that was successful and central starts back up successfully then central_backup will be deleted.
    - Any data created after the creation of the backup bundle will be lost

- Upgrade
    - Upgrade is a similar process
    - The user requests to upgrade central from an older version to a newer one
    - central_active will be copied to central_previous in case something goes awry
    - Migrations will be executed on central_active to bring it up to the new version
    - central_previous remains in the event the user experience requires them to downgrade
- Rollback/Downgrade
    - User can only downgrade IF central_previous exists and if its version matches
    - central_previous will be moved to central_active
    - no backup replica will remain
    - data between the upgrade and the rollback will be lost.

It should be noted that migrations are not executing through this process with this PR.  [ROX-11922](https://issues.redhat.com/browse/ROX-11922) will add that.

## Checklist

## Testing Performed

updated replica_test.go to support Postgres.
[ROX-11730](https://issues.redhat.com/browse/ROX-11730) will update the e2e upgrade tests for Postgres

Ran through various scenarios locally and in infra of backups/restores, upgrades, rollbacks, etc.


dashrews78:
moving out of draft so CI can do some work.

pkg/migrations/migration_version.go
```diff
@@ -55,6 +61,29 @@ func Read(dbPath string) (*MigrationVersion, error) {
 	return version, nil
 }
 
+// ReadVersion - reads the version from the postgres database.
+func ReadVersion(pool *pgxpool.Pool) (*MigrationVersion, error) {
```
dashrews78, line 65: I need to be able to read/write the version from migrator and from pkg/migrations.  It didn't make a lot of sense to keep the migration version in a file on central with the database deployed some places else entirely as that caused lots of complications.  So we are reading and writing to the version table which is what I believe its purpose to be.  So to do that I moved the postgres version store to pkg.  That may or may not have been the correct thing to do.  In doing so, I fail style checks here and in the store because of the import for resources.Version.  Do I need to move that to pkg as well?  Should I not move the store?  Thoughts on how best to proceed there?

central/centralhealth/service/service_impl.go
```diff
@@ -46,6 +49,32 @@ func (s *serviceImpl) AuthFuncOverride(ctx context.Context, fullMethodName strin
 
 // GetUpgradeStatus returns the upgrade status for Central.
 func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v1.GetUpgradeStatusResponse, error) {
+	// TODO: Need to figure out if there is anything we can do with size in Postgres as that data
+	// IS NOT accessible from central
+	if features.PostgresDatastore.Enabled() {
+		upgradeStatus := &v1.CentralUpgradeStatus{
+			Version: version.GetMainVersion(),
+			//CanRollbackAfterUpgrade:               int64(freeBytes)+toBeFreedBytes > requiredBytes,
```
dashrews78, line 57: We do not have access to the capacity of postgres.  That is only available from the file system where the database resides.  I'm not sure what to do with this.  Thoughts?

dashrews78:
ROX-11235

pkg/migrations/migration_version.go
c-du: Yes, I got you. You can move the version store to pkg and leave a read-only wrapper in central with all the resource and access control. We can let migrator take care all maintainence of version.

migrator/replica/postgres/db_replica_manager_impl.go
```diff
@@ -0,0 +1,279 @@
...
+	replicas := pgadmin.GetDatabaseReplicas(d.adminConfig)
+
+	// We use replicas to collect all db replicas (directory starting with db- or .restore-) matching upgrade or restore pattern.
+	// We maintain replicas with a known link in replicaMap. All unknown replicas are to be removed.
+	replicasToRemove := set.NewStringSet()
+	for _, replica := range replicas {
+		switch name := replica; {
+		case knownReplicas.Contains(name):
+			// Get a short-lived connection for the purposes of checking the version of the replica.
+			pool := pgadmin.GetReplicaPool(d.adminConfig, name)
```
c-du, line 48: It will be desirable if we can switch between databases.

migrator/main.go
```diff
@@ -56,38 +61,71 @@ func run() error {
...
 		}
-	} else {
-		var gormDB *gorm.DB
-		gormDB, err := postgreshelper.Load(conf)
-		if err != nil {
-			return errors.Wrap(err, "failed to connect to postgres DB")
+
+		// If we have database replicas we need to process them.  Otherwise, we can assume we are
+		// starting from scratch and as such we will need to create the DB and apply schemas with Gorm.
+		if len(dbm.ReplicaMap) > 0 {
```
c-du, line 79: We may need to check both postgres and rocksdb for replicas.
We have a few scenarios here:
1. This is the first time we migrate from Rocksdb -> Postgresdb, we have the flag enabled but we actually need to get all replicas from rocksdb.
2. We had tried to upgrade to Rocksdb -> Postgresdb before but failed somewhere, so we may have most replicas in rocksdb but some residues in Postgres DB.
3. We have upgraded to Postgres before any changes are persisted successfully.
We can tell such conditions by checking if the action is in Postgres or in Rocksdb or both.

central/centralhealth/service/service_impl.go
dashrews78: I can, BUT that will not work if we use a managed service for the database.  In those instances we will have to query the provider.  I think I'm going to write another subtask to cover this and note that in the TODO unless you would prefer to tackle it with this one.

central/globaldb/export/backup.go
```diff
@@ -64,9 +64,5 @@ func BackupPostgres(ctx context.Context, postgresDB *pgxpool.Pool, includeCerts
 		}
 	}
 
-	if err := generators.PutStreamInZip(generators.PutFileInStream(filepath.Join(migrations.CurrentPath(), backup.MigrationVersion)), backup.MigrationVersion).WriteTo(ctx, zipWriter); err != nil {
```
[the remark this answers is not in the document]
dashrews78: No.  With this, it is only available within the postgres database.  If we pull that file down and manage that file, then we still have to manage directories on the file system in addition to databases in Postgres.  That seemed like a great potential for error.  Now we could query the database on backup and write the file out in an information only way.  Thoughts?

migrator/main.go
```diff
@@ -56,38 +61,71 @@ func run() error {
...
+				return err
+			}
+
+			// Save the replica after upgrades have been processed
+			if err = dbm.Persist(replica); err != nil {
+				return err
+			}
+		} else {
+			// No existing DB replica so need to create it.
+			gormDB, err := postgreshelper.Load(conf)
```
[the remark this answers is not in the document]
dashrews78: yeah, probably.  I will take care of that.

migrator/main.go
```diff
@@ -56,38 +61,71 @@ func run() error {
...
+			if err != nil {
+				return errors.Wrap(err, "failed to connect to postgres DB")
+			}
+			pkgSchema.ApplyAllSchemas(context.Background(), gormDB)
```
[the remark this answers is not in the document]
dashrews78: I will take another look at this.  Gorm was failing when it was doing this so I coded around it.  I will investigate further.

migrator/replica/postgres/db_replica_manager_impl.go
dashrews78: I'm not following you.  That pool is to whichever database we are looking at.  central_active, central_backup, central_restore, etc.  Now what I do is move them around such that central itself only ever hits central_active.

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -51,7 +59,13 @@ func DropDB(sourceMap map[string]string, adminConfig *pgxpool.Config, databaseNa
 
 // CreateDB - creates a database from template with the given database name
 func CreateDB(sourceMap map[string]string, adminConfig *pgxpool.Config, dbTemplate, dbName string) error {
-	log.Debugf("CreateDB %s", dbName)
+	log.Infof("CreateDB %q", dbName)
+	pc, _, _, ok := runtime.Caller(1)
+	details := runtime.FuncForPC(pc)
+	if ok && details != nil {
+		fmt.Printf("called from %s\n", details.Name())
```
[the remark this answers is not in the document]
dashrews78: yep

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -105,23 +120,88 @@ func CheckIfDBExists(config *pgxpool.Config, dbName string) bool {
...
+
+	// Connect to different database for admin functions
+	connectPool := GetAdminPool(pgConfig)
+	// Close the admin connection pool
+	defer connectPool.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), postgresQueryTimeout)
+	defer cancel()
+
+	selectStmt := fmt.Sprintf("SELECT datname FROM pg_catalog.pg_database WHERE datname ~ '^%s.*'", config.GetConfig().CentralDB.DatabaseName)
```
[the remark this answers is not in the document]
dashrews78: I think migrator will support other database names but we will need to update Load to take in a database name so the Gorm we create in the Load call and pass to the migrator will be pointing to the correct replica such that the migrations are executed on the correct replica.

central/centralhealth/service/service_impl.go
dashrews78: ROX-12059

migrator/replica/rocksdb/db_replica_manager.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// PreviousReplica - previous rocksdb replica used for rollback
+	PreviousReplica = ".previous"
+
+	// TempReplica - temp rocksdb replica
+	TempReplica = "temp"
+
+	// EmergencyReplica - emergency rocksdb replica, only to be used in the event an
+	// upgrade to Postgres was followed with a rollback
+	EmergencyReplica = ".emergency"
```
c-du, line 33: No, it wont work.
The earlier version wont be able to recognize the new EmergencyReplica. 
We can use the .previous link for rollback. I feel like we do not need to create a new emergency replica here. After migration, we should have .previous and a temp replica in preparing for migration. We can use .previous replica for rollback and make the temp replica as a fake current which can be recognized by Postgres.

c-du (commented):
I recommend you to keep one replica manager with hooks to both rocksdb and postgres. It would be very hard to manage replicas with two managers.

migrator/main.go
```diff
@@ -56,38 +61,100 @@ func run() error {
...
+				return err
+			}
+		}
 	}
 
 	return nil
 }
 
-func upgrade(conf *config.Config) error {
+func upgrade(conf *config.Config, dbReplica string, runLegacy bool) error {
```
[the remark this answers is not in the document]
dashrews78: The idea there is a case where we are doing an upgrade from sometime back and upgrading to Postgres at the same time.  Thought was first run through the migrations on rocks to get it up to date and then migrate the updated rocks to postgres.  So no, I can't just use the flag as I need to update Rocks first.

migrator/replica/rocksdb/db_replica_manager.go
dashrews78: you are right, it won't work.  The permutations here are tricky.  Not having a current is not a legal state, currently.  I have an idea, I'm going to work through.

migrator/main.go
```diff
@@ -56,38 +61,100 @@ func run() error {
...
+		if err != nil {
+			return err
+		}
+
+		// No existing DB replica so need to create it.
+		gormDB, err := postgreshelper.Load(conf, replica)
 		if err != nil {
 			return errors.Wrap(err, "failed to connect to postgres DB")
 		}
 		pkgSchema.ApplyAllSchemas(context.Background(), gormDB)
```
c-du, line 121: Leave this schema apply to after all data migration done.

migrator/main.go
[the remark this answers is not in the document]
c-du: We do not have to seperate the process into two phrases. The DB sequence number should tell us the sequence number to start and end.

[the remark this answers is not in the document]
dashrews78: I was going to change it so update is only called once and the runner figures out what to run.  If the runner recives both a rocks and a postgres it will have to determine what to run and in which order.  That just can't fully happen until we wire in your migrations.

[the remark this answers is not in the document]
dashrews78: We are going to have to push some of this to the various tickets dealing with wiring in the migrations.  I can't handle all that with the replicas.  I can send the right replicas with the right versions to the migrations runner, but I think that is as far as I can go in this PR.

migrator/replica/rocksdb/db_replica_manager.go
dashrews78: The more I think about the more I think we are over-complicating that scenario.  If Postgres has been updated based on Rocks then Postgres will have a version >= Rocks.  When we run the migrations we will have to use that knowledge to determine which set of migrations to run.  I don't think it matters if we look at the Rocks replicas and see a current once the initial move to Postgres occurs.  The only time that might be a problem is if we rollback from Postgres to Rocks.  Then when we upgrade again to the same version as before, then we will think that Postgres is newer.  However, we have ROX-12094 to cleanup Postgres residue in the event of a rollback to Rocks.  This is feeling like an edge case that I'm not seeing a solution to right now that doesn't involve human intervention either to clean things up or require them to upgrade beyond the Postgres version.

migrator/replica/postgres/db_replica_manager_impl.go
dashrews78: there is no concept of an alias of a database if that is what you are getting at

tests/upgrade/run.sh
```diff
@@ -38,8 +38,8 @@ test_upgrade() {
     export OUTPUT_FORMAT="helm"
     export STORAGE="pvc"
     export CLUSTER_TYPE_FOR_TEST=K8S
-    require_environment "LONGTERM_LICENSE"
-    export ROX_LICENSE_KEY="${LONGTERM_LICENSE}"
+#    require_environment "LONGTERM_LICENSE"
```
dashrews78, line 41: yeah I'm going to undo this.

central/centralhealth/service/service_impl.go
connorgorman: 👍 we can also potentially pass the max to Central

dashrews78: @connorgorman are you talking about setting something like an env var for central based off the central-db pvc size config in values?  Or just setting an env var?  Something else?  Lots of permutations on what the max size could be.  With a pvc it is configurable.  It isn't really with a hostPath.  It is something else entirely when we standup a managed service with a managed Postgres, though I assume that value could be a parameter.  And the max is "infinite" if we back it with Aurora Serverless.

connorgorman: Yeah, I think there are definitely some complications with that

central/centralhealth/service/service_impl.go
```diff
@@ -46,6 +49,35 @@ func (s *serviceImpl) AuthFuncOverride(ctx context.Context, fullMethodName strin
...
+		// Get a short-lived connection for the purposes of checking the version of the replica.
+		_, adminConfig, err := pgconfig.GetPostgresConfig()
+		if err != nil {
+			return nil, err
+		}
+		pool := pgadmin.GetReplicaPool(adminConfig, "central_previous")
+		defer pool.Close()
+
+		// Get rollback to version
+		migVer, err := migrations.ReadVersion(pool)
```
connorgorman, line 71: We're not doing anything with this err if it's != nil

migrator/main.go
```diff
@@ -56,76 +62,146 @@ func run() error {
...
 		if err != nil {
-			return err
-		}
-		option.MigratorOptions.DBPathBase = replicaPath
-		if err = upgrade(conf); err != nil {
-			return err
-		}
-		if err = dbm.Persist(replicaName); err != nil {
-			return err
+			return errors.Wrap(err, "unable to get Postgres DB config.")
```
connorgorman, line 70: no need for the period at the end

migrator/main.go
```diff
@@ -56,76 +62,146 @@ func run() error {
...
+	// Set the path to Rocks if it exists.
+	if replicaPath != "" {
+		option.MigratorOptions.DBPathBase = replicaPath
 	}
 
-	boltDB, err := bolthelpers.Load()
+	// If GetReplicaToMigrate returns Rocks and Postgres replicas that means we need
+	// to migrate Rocks->Postgres.  Otherwise, we need to process Rocks in Rocks mode and
+	// Postgres in Postgres mode.
+	processBoth := false
```
connorgorman, line 96: var processBoth bool

migrator/main.go
```diff
@@ -56,76 +62,146 @@ func run() error {
...
+	if replica != "" && pgReplica != "" {
+		processBoth = true
```
connorgorman, line 98: processBoth = replica ...

migrator/main.go
```diff
@@ -56,76 +62,146 @@ func run() error {
...
+		gormDB, err := postgreshelper.Load(conf, dbReplica)
+		if err != nil {
+			return errors.Wrap(err, "failed to connect to postgres DB")
+		}
+
+		// We need to close the gorm connection when we are finished with it so the replica
+		// movement can occur.
+		sqlDB, err := gormDB.DB()
+		if err != nil {
+			log.WriteToStderrf("Error closing GormDB connection: %v", err)
```
connorgorman, line 159: I think this message is wrong?

migrator/main.go
```diff
@@ -56,76 +62,146 @@ func run() error {
...
+
+		ver, err := migrations.ReadVersion(pgPool)
+		if err != nil {
+			return errors.Wrap(err, "failed to get version from the database")
+		}
+
+		// If Postgres has no version and we have no bolt then we have no populated databases at all and thus don't
+		// need to migrate
+		if ver.SeqNum == 0 && ver.MainVersion == "0" && !processBoth {
+			log.WriteToStderr("Initial. Nothing to migrate...")
```
connorgorman, line 187: Can we elaborate on this log. Fresh database, new database, etc

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	return nil
+}
+
+// GetReplicaToMigrate - finds a replica to migrate.
+// It returns the replica link, path to database, postgres database name and error if fails.
+func (d *DBReplicaManagerImpl) GetReplicaToMigrate() (string, string, string, error) {
+	log.Info("GetReplicaToMigrate")
```
connorgorman, line 68: rm?

central/centralhealth/service/service_impl.go
dashrews78: Looking down below, I think that is on purpose from where I copied it from.  I will log it though but we do not want to return this particular one.

migrator/replica/rocksdb/db_replica_manager.go
dashrews78: I think I have this covered now.  We can use the timestamp on current to tell us if Rocks or Postgres is the functional DB on the upgrade.  So even if there is residue in Postgres, we should be covered until we figure out how to solve ROX-12094

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,98 @@
+package replica
```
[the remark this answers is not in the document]
dashrews78: I'm covering rocks->postgres in replica-test already.  I will double check to see if I covered the rollback.

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		return "", "", pgReplica, err
+
+	}
+
+	return replica, replicaPath, "", err
+}
+
+// Persist - replaces current replica with upgraded one.
+func (d *DBReplicaManagerImpl) Persist(replicaName string) error {
```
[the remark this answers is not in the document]
dashrews78: Based on existing logic, if rollback is enabled, we are creating a TempReplica from Current.  This TempReplica is the replica we are working with through the migration process.  When we go to persist AND are in Postgres we do this
`if persistBoth && replicaName == rocksdb.TempReplica {
	if err := d.dbmRocks.Persist(replicaName); err != nil {
		log.Warnf("Unable to create a previous version of Rocks to rollback to: %v", err)
	}
}`

That will take the original current and move it to previous.  It will also move temp to the new current on Rocks.  What we wind up with is a previous with a pre-migration version and a current that has the updated migration sequence number.  I think the previous logic is correct.  I will run some tests on the current part as that may not be completely correct.  There are checks elsewhere that require a current Rocks to exist I was trying to avoid messing with those as there is a huge ripple if I do.

migrator/replica/postgres/db_replica_manager_impl.go
```diff
@@ -0,0 +1,297 @@
...
+	if prevExists && prevReplica.GetVersion() == version.GetMainVersion() {
+		return PreviousReplica, false, nil
+	}
+
+	log.Info("Fell through all checks to return current, meaning probably empty OR rollback disabled.")
+	return CurrentReplica, false, nil
+}
+
+// Persist - replaces current replica with upgraded one.
+func (d *DBReplicaManagerImpl) Persist(replicaName string) error {
```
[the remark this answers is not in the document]
dashrews78: It actually does, if rollback is enabled, all work will be done on Temp Replica and the act of persisting that will do that.  Now one thing we might want to consider in this situation is what does enabling rollback mean in this case?  Do we use different criteria to determine rollback if we are in this hybrid situation or do we follow the same rules?  Since rollback is based on space we would probably have to do something sneaky like move current to previous and have an empty current.  But it seemed simpler to just not allow rollback from postgres->rocks if rollback from rocks->rocks was not enabled.

migrator/replica/db_replica_manager_impl.go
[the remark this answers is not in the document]
dashrews78: @c-du After looking at this case some more, it is tough since i have migrations turned off to see the permutations of how the version works its way through the versions.  So I would like to either defer this case until ROX-11922 OR if I'm still writing unit tests when your change is merged, I can try to incorporate that in this PR.  I have a good idea of how I'll make it work so it shouldn't be that bad.

central/centralhealth/service/service_impl.go
```diff
@@ -83,6 +118,9 @@ func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v
 
 	// Get rollback to version
 	migVer, err := migrations.Read(filepath.Join(migrations.DBMountPath(), migrations.PreviousReplica))
+	if err != nil {
+		log.Infof("Unable to get previous version, leaving ForceRollbackTo empty")
```
connorgorman, line 122: log the error?

central/centralhealth/service/service_impl.go
```diff
@@ -46,6 +49,38 @@ func (s *serviceImpl) AuthFuncOverride(ctx context.Context, fullMethodName strin
...
+		if err != nil {
+			return nil, err
+		}
+		pool := pgadmin.GetReplicaPool(adminConfig, "central_previous")
+		defer pool.Close()
+
+		// Get rollback to version
+		migVer, err := migrations.ReadVersion(pool)
+		if err != nil {
+			log.Infof("Unable to get previous version, leaving ForceRollbackTo empty")
```
connorgorman, line 73: log the error?

central/globaldb/export/backup.go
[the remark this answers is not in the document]
connorgorman: Yeah I think we need to keep it self contained in the storage layer so no disk writes

central/version/store/store_impl.go
```diff
@@ -105,8 +103,7 @@ func (s *storeImpl) UpdateVersion(version *storage.Version) error {
 	if features.PostgresDatastore.Enabled() {
 		ctx := sac.WithGlobalAccessScopeChecker(context.Background(),
```
connorgorman, line 104: I would just sac.WithAllAccess

central/version/store/store_impl.go
```diff
@@ -71,8 +70,7 @@ func (s *storeImpl) GetVersion() (*storage.Version, error) {
...
```
connorgorman, line 71: sac.WithAllAccess

migrator/main.go
```diff
@@ -56,76 +62,144 @@ func run() error {
...
-		pkgSchema.ApplyAllSchemas(context.Background(), gormDB)
-		log.WriteToStderr("Applied all table schemas.")
+		dbm = replica.New(migrations.DBMountPath(), conf.Maintenance.ForceRollbackVersion)
 	}
 
-	return nil
-}
+	err := dbm.Scan()
+	if err != nil {
+		return errors.Wrap(err, "fail to scan replicas")
```
connorgorman, line 79: failed

migrator/main.go
```diff
@@ -56,76 +62,144 @@ func run() error {
...
+	}
 
-func upgrade(conf *config.Config) error {
-	if err := compact.Compact(conf); err != nil {
-		log.WriteToStderrf("error compacting DB: %v", err)
+	// Get the replica we are migrating
+	replica, replicaPath, pgReplica, err := dbm.GetReplicaToMigrate()
+	log.WriteToStderrf("Replica to Migrate %q, %q", replica, pgReplica)
+	if err != nil {
+		return errors.Wrap(err, "fail to get replica to migrate")
```
connorgorman, line 86: failed

central/centralhealth/service/service_impl.go
dashrews78: @connorgorman yeah, if you look below we never returned that particular error.  we would simply set the rollbackTo version if the err was nil and the version was populated.  So I logged it in the interest of keeping it functionally the same as before.  We can easily return it if you would rather.

migrator/main.go
```diff
@@ -56,76 +62,144 @@ func run() error {
...
+		defer pgPool.Close()
+
+		ver, err := migrations.ReadVersion(pgPool)
+		if err != nil {
+			return errors.Wrap(err, "failed to get version from the database")
+		}
+
+		// If Postgres has no version and we have no bolt then we have no populated databases at all and thus don't
+		// need to migrate
+		if ver.SeqNum == 0 && ver.MainVersion == "0" && !processBoth {
```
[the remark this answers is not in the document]
dashrews78: @connorgorman yeah I get it.  But for consistency I kept it that way.  if you look in migration_version.Read you will see that the Rocks version would set it to that "0"
`	if os.IsNotExist(err) {
		return &MigrationVersion{dbPath: dbPath, SeqNum: 0, MainVersion: "0"}, nil
	}`
I suppose we could make a constant for it.  Or i could change it to empty string all around.  I'm not picky

migrator/replica/db_replica_manager_impl.go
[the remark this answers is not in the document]
dashrews78: I have turned migrations on and I believe I have covered this scenario

dashrews78:
> I recommend you to keep one replica manager with hooks to both rocksdb and postgres. It would be very hard to manage replicas with two managers.

It has been easier for me to follow the logic by keeping the operations separate and letting a parent manager handle the collaboration.  That doesn't mean I'm right, just that is how I could visualize the code and the steps that needed to be accomplished.

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+// GetReplicaToMigrate - finds a replica to migrate.
+// It returns the replica link, path to database, postgres database name and error if fails.
+func (d *DBReplicaManagerImpl) GetReplicaToMigrate() (string, string, string, error) {
+	// Get the RocksDB replica we are migrating
+	replica, replicaPath, err := d.dbmRocks.GetReplicaToMigrate()
+	if err != nil {
+		return "", "", "", err
```
c-du, line 71: Why are we return no replica chosen here? We did not try to select a replica from Postgres yet.

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Persist - replaces current replica with upgraded one.
+func (d *DBReplicaManagerImpl) Persist(replicaName string, pgReplica string, persistBoth bool) error {
+	if features.PostgresDatastore.Enabled() {
+		// We need to persist the Rocks previous, so it is there in case of a rollback.  In the case of
+		// an upgrade that will generate a previous, the Temp Replica will be the one RocksDB persists.
+		// During the persist operation the Current replica will move to Previous and Temp will move to Current.
+		if persistBoth && replicaName == rocksdb.TempReplica {
+			if err := d.dbmRocks.Persist(replicaName); err != nil {
```
c-du, line 110: We did not remove the previous and current version after two migrations, do we? 

Like (rocksdb ver) -> (Postgres ver1) -> (postgres ver2). After two upgrades, we should be able to clean up all the replicas on rocksdb.

pkg/migrations/migration_version.go
```diff
@@ -45,16 +53,51 @@ func Read(dbPath string) (*MigrationVersion, error) {
 		return nil, err
 	}
 
+	// Get the last time file was updated
+	fileInfo, err := os.Lstat(path)
```
c-du, line 57: Is it possible that path is a symbolic link? Can you confirm if we can get correct mod time here.

migrator/replica/db_replica_manager_impl.go
```diff
@@ -0,0 +1,117 @@
...
+			return "", "", "", err
+		}
+
+		// If we need to migrate from rocks we need to return both
+		// rocks and postgres replicas.  Otherwise we just need one or
+		// the other.
+		if migrateFromRocks {
+			return replica, replicaPath, pgReplica, err
+		}
+		return "", "", pgReplica, err
```
c-du, line 96: If it is not migrateFromRocks, since we called rocks replica manager before to select replica, we may have created a temp replica from the "current rocksdb"?

c-du (commented):
Leave some comments. Thanks!

migrator/replica/db_replica_manager_impl.go
dashrews78: nice catch as that could occur if the PVC were removed.  Will fix.

migrator/replica/db_replica_manager_impl.go
dashrews78: yeah need to re-order that.  Since I use current to determine if we need to migrate, I simply need to process Postgres first here and only process rocks if we are in postgres mode and do need to migrate.  If we don't need to migrate, there is no reason to process Rocks.

migrator/replica/db_replica_manager_impl.go
dashrews78: Agreed.  I had looked at that a while ago and there are a lot of start up checks to ensure that current rocks exists.  I'd rather not delve into that in this PR because it is big enough.  So I wrote [ROX-12289](https://issues.redhat.com/browse/ROX-12289) to solve this issue.

migrator/replica/rocksdb/db_replica_manager_impl.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	return nil
+}
+
+// GetCurrentReplicaCreationTime - time current replica was created
+func (d *DBReplicaManagerImpl) GetCurrentReplicaCreationTime() time.Time {
+	path := d.getPath(CurrentReplica)
+	fileInfo, err := os.Lstat(path)
+	if err != nil {
+		return time.Time{}
```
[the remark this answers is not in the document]
dashrews78: yes and no.  If postgres is enabled that may not be an issue.   Should only be called in Postgres mode.  I will add some protection and panic if it is not in Postgres mode.

pkg/migrations/migration_version.go
dashrews78: Turns out I'm not using this value from here.  I'm pulling it in the replica manager as I can get the path and not worry about the links.  I'm going to remove this one.

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -21,8 +22,16 @@ const (
...
+	// EmptyDB - name of an empty database (automatically created by postgres)
+	EmptyDB = "template0"
+
+	// postgresOpenRetries - number of retries when trying to open a connection
+	postgresOpenRetries = 10
+
+	// postgresTimeBetweenRetries - time to wait between retries
 	postgresTimeBetweenRetries = 10 * time.Second
+
+	getReplicaStmt = "SELECT datname FROM pg_catalog.pg_database WHERE datname ~ '^%s_.*'"
```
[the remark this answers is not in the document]
dashrews78: Essentially I am restricting it, it just isn't as obvious.  When I use that statement I pass the value of `config.GetConfig().CentralDB.DatabaseName` into it.  That config value at this time is currently hard coded to "central"

pkg/postgres/pgtest/postgres.go
```diff
@@ -30,27 +30,54 @@ func CreateADatabaseForT(t testing.TB) string {
...
+
+	// Only create the test DB if it does not exist
+	if !exists {
+		_, err = db.Exec("CREATE DATABASE " + database)
+		if err != nil {
+			return err
+		}
+	}
 	require.NoError(t, err)
 	require.NoError(t, db.Close())
```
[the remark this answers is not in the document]
dashrews78: ha, well mostly because I copied some of this from test code and some of this from non-test code.  These should be require.NoError to match the rest of the test file.  will resolve.

c-du (commented):
Overall look good to me. I left a few comments and also please clean up the debug messages. Thanks!

migrator/clone/db_clone_manager_impl.go
```diff
@@ -0,0 +1,117 @@
...
+	if features.PostgresDatastore.Enabled() {
+		// We need to persist the Rocks previous, so it is there in case of a rollback.  In the case of
+		// an upgrade that will generate a previous, the Temp Clone will be the one RocksDB persists.
+		// During the persist operation the Current clone will move to Previous and Temp will move to Current.
+		if persistBoth && cloneName == rocksdb.TempClone {
+			if err := d.dbmRocks.Persist(cloneName); err != nil {
+				log.Warnf("Unable to create a previous version of Rocks to rollback to: %v", err)
+			}
+		}
+		return d.dbmPostgres.Persist(pgClone)
```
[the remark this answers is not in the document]
dashrews78: yeah that is an interesting case.  I suspect in that case it would try the migration from Rocks -> Postgres because the err coming back from the persist of Postgres would fail the migration.  Definitely a test case we need to look at when we introduce failures in the process

migrator/runner/version.go
```diff
@@ -61,13 +63,28 @@ func GetCurrentSeqNumRocksDB(db *gorocksdb.DB) (int, error) {
...
+		return 0, errors.Wrap(err, "getting current postgres sequence number")
+	}
+
+	return migVer.SeqNum, nil
+}
 func getCurrentSeqNum(databases *types.Databases) (int, error) {
+	// If Rocks and Bolt are passed into this function when Postgres is enabled, that means
+	// we are in a state where we need to migrate Rocks to Postgres.  In this case the Rocks
+	// sequence number will take precedence and drive the migrations
+	if features.PostgresDatastore.Enabled() && databases.RocksDB == nil && databases.BoltDB == nil {
```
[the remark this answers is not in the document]
dashrews78: it is not that simple.  if we are upgrading from Rocks -> Postgres we need to get the sequence number from Rocks to start the migration process.  The Postgres sequence won't get set until we start going through n01.  So I still need the other check to make sure I only get the postgres sequence when I need it.  The clone manager will only send all databases to the runner if it determines it needs to go from Rocks -> Postgres.  Otherwise it will only send one or the other.  So we shouldn't get in this situation.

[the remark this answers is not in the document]
dashrews78: I think I covered your intent with the latest update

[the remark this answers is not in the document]
c-du: Right. Got you

c-du (approved):

pkg/version/postgres/gen.go
```diff
@@ -1,3 +1,3 @@
 package postgres
 
-//go:generate pg-table-bindings-wrapper --type=storage.Version --singleton
```
janisz, line 3: Why do we need this change? If you added this then please remove `// Code generated by pg-bindings generator. DO NOT EDIT.` from corresponding `store.go`

dashrews78: @janisz I needed to move the simple version store to pkg so that both central and migrator could get at it.  Once I did that that I either had to remove some of the SAC stuff such as resources from the store or pull all that stuff to pkg as well.  Due to what this store is, we decided in the short term to just generate the schema, manually remove the SAC stuff causing issue, and re-visit in [ROX-12083](https://issues.redhat.com/browse/ROX-12083) if we desire.  We missed removing the comment from the store.  I will take care of that as I'm working other PRs that are doing follow on things in the feature.  Hopefully that answers your questions.

janisz: OK, sounds good. Pleas add todos so it will be obvious to everybody that there is something that will be changed

PR #2501, opened 2022-07-22
ROX-11136 persist dashboard widget config

dvail:
## Description

Adds a `useWidgetStorage` hook that will be used to persist widget configuration options across relevant dashboard widgets. Implementation for the current dashboard widgets will come in a follow up PR.

This hook implementation will use `localStorage` and store widget data under a single key in a structure like the following:
```javascript
{
    violationsByCategory: {
        '/main/dashboard': {
            sortBy: 'Severity',
            lifecycle: 'All',
        },
        '/some/other/page': {
            sortBy: 'Total',
            lifecycle: 'Deploy',
        },
    },
    agingImages: {
        '/main/dashboard': {
            ...
        },
    }
};
```

Note that this does not use our existing `useLocalStorage` hook due to the configurations being stored under a single key. I started out on that path but realized that I would run into issues with:
- Multiple unrelated changes overwriting existing configs with stale data
- **All** components using this hook on the same page re-rendering when any single one changed

## Known limitations

- If you were to use a widget on the same page more than one, with an identical `widgetId` and identical `routeId`, config changes would not update in-memory for the two widgets. I don't think this is a use case we will need to support but I wanted to mention it.

## Checklist

## Testing Performed

Unit tests for hook included.

### DIFF
--- a/ui/apps/platform/src/hooks/useWidgetConfig.test.tsx
+++ b/ui/apps/platform/src/hooks/useWidgetConfig.test.tsx
@@ -1,0 +1,242 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import React from 'react';
+import { render, act as reactAct } from '@testing-library/react';
+import { renderHook, act as hooksAct } from '@testing-library/react-hooks';
+import useWidgetConfig, { defaultStorageKey } from 'hooks/useWidgetConfig';
+
+beforeEach(() => {
+    localStorage.clear();
+});
+
+describe('useWidgetConfig hook', () => {
+    it('should persist a widget configuration in localStorage', () => {
+        const widgetId = 'single-component';
+        const routeId = '/';
+        const defaultConfig = {
+            sort: 'asc',
+            filter: '',
+        };
+        const { result: firstResult } = renderHook(() =>
+            useWidgetConfig(widgetId, routeId, defaultConfig)
+        );
+
+        // Expect the initial state to be clean and return the provided defaults
+        expect(firstResult.current[0]).toEqual(defaultConfig);
+
+        // By default changes should merge the new and old configs
+        hooksAct(() => {
+            const [, updateConfig] = firstResult.current;
+            updateConfig({ filter: 'Namespace:stackrox' });
+        });
+        expect(firstResult.current[0]).toEqual({ sort: 'asc', filter: 'Namespace:stackrox' });
+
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/hooks/useWidgetConfig.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+        );
+
+        expect(firstResult.current[0]).toEqual(defaultConfig);
+
+        hooksAct(() => {
+            const [, updateConfig] = firstResult.current;
+            updateConfig({ filter: 'Namespace:stackrox' });
+        });
+
+        // Even though the `sort` property was overwritten, the hook ensures type safety by
```
sachaudh, line 71: At first, I was thinking what if we want to clear the sort, but I guess technically that's something you should have regardless

sachaudh (approved):
Looks really good @dvail. Thanks for working on it! 🚢

ui/apps/platform/src/types/widgetConfig.ts
```diff
@@ -0,0 +1,31 @@
+// The main configuration storage type. Reading directly from localStorage would provide this type
+// of result at the top level.
+export type WidgetConfigStorage = Record<WidgetId, WidgetConfigMap>;
+
+// A unique identifier for a widget. Short term this will be the name of the widget, but if
+// we move to server side storage this could become a UUID.
```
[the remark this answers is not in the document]
dvail: Yeah how this is going to work on the server is pretty vaguely defined. PM and UX put together a long term vision on how dashboards will be capable of, both on the main page and throughout the app, but I don't think we are anywhere close to a timeline on implementation. The idea is that eventually users will have the option to do really detailed customization, i.e. create-your-own widgets, that define options, what data is loaded, dynamic titles etc. 

In that case I agree that the backend will pass along an id, name, description, query properties, etc. and these widgets will probably be migrated to a backend OOTB widget service of some sort.

ui/apps/platform/src/hooks/useWidgetConfig.test.tsx
dvail: That's a good point, I think in that case I'd probably update the union type to contain something like "NO_SORT". The idea here was that since it is stored in localStorage we can't really be sure what we are getting back, so this does the best it can to ensure we don't run into weird runtime errors.

PR #2505, opened 2022-07-24
Treat SYNC as CREATE for in namespace pipeline

connorgorman:
## Description

This fixes the issue with SAC

## Checklist

## Testing Performed

QA tests should pass


gavin-stackrox (approved):
Looks like it fixes the SACTest issue. The ImageScanning/gcr failures are already fixed so those failures just need a rebase.

PR #2509, opened 2022-07-25
ROX-11923: Generic set

0x656b694d:
## Description

Transform generated set to a generic one.

## Checklist

## Testing Performed

Unit tests, CI, benchmarks.


stehessel:
Looks great 🚀. "Only" need to get tests to pass now...

0x656b694d:
```
x509: certificate signed by unknown authority (possibly because of "x509: cannot verify signature: insecure algorithm SHA1-RSA (temporarily override with GODEBUG=x509sha1=1)" while trying to verify candidate authority certificate "Intermediate Server CA")
```
Unit test failure. Needs to be addressed.

connorgorman (changes requested):
I am requesting changes until I can confirm that adding this use of generics won't prevent us from using the FIPS go-toolset on our timeline

vikin91:
> I am requesting changes until I can confirm that adding this use of generics won't prevent us from using the FIPS go-toolset on our timeline (currently have a message out and waiting to hear back)

Chipping in here (not sure if fully relevant). The active builder image in downstream - [openshift-golang-builder-container-v1.18.2-202206222023.el8.gb30794b](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2060177) - has been selected from [this list of images](https://brewweb.engineering.redhat.com/brew/search?terms=*openshift-golang-builder*&type=build&match=glob). Some images on that list are marked with `nofips`, however our image is missing such annotation, so I deduct from that, that it may be FIPS compliant.

janisz (commented):
If possible I'd like to see it as a two PRs one to update go to 1.18 and another to use generic set

misberner:
@0x656b694d do you want to resurrect this one?

0x656b694d:
Yeah, why not. Rebasing and testing locally.

central/cluster/datastore/datastore_impl.go
```diff
@@ -512,7 +512,7 @@ func (ds *datastoreImpl) RemoveCluster(ctx context.Context, id string, done *con
 
 func (ds *datastoreImpl) postRemoveCluster(ctx context.Context, cluster *storage.Cluster, done *concurrency.Signal) {
 	// Terminate the cluster connection to prevent new data from being stored.
-	if ds.cm != nil {
+	if ds.cm != (connection.Manager)(nil) {
```
misberner, line 515:
```suggestion
	if ds.cm != nil {
```

central/graphql/resolvers/search_test.go
```diff
@@ -75,7 +75,8 @@ func TestSearchCategories(t *testing.T) {
 		case v1.SearchCategory_COMPLIANCE:
 			continue
 		default:
-			assert.True(t, searchFuncs[category] != nil, "search category %s does not have a search func", category.String())
+			_, ok := searchFuncs[category]
+			assert.True(t, ok, "search category %s does not have a search func", category.String())
```
misberner, line 79:
```suggestion
			assert.True(t, searchFuncs[category] != nil, "search category %s does not have a search func", category.String())
```

.golangci.yml
```diff
@@ -91,7 +91,7 @@ linters-settings:
     - 'os\.Stderr(# Disallowed output streams used\. Use environment\.InputOutput\(\).Out instead\.)?'
     - 'os\.Stdin(# Disallowed output streams used\. Use environment\.InputOutput\(\).Out instead\.)?'
   staticcheck:
-    go: "1.16"
+    go: "1.18"
```
misberner, line 94: Maybe it makes sense to factor out this as well as the changes in `qa-tests-backend/` and `tools/` into a separate PR. I assume we want to do some more thorough investigation as to how the set changes impact the performance, but the aforementioned changes block usage of go1.18 features otherwise (at least for some use cases). I probably should've done this when bumping the version in `go.mod`, but unfortunately I missed that.

0x656b694d: @misberner, would you suggest me to run some performance tests on the generic set versus generated?
Do you expect a noticeable degradation, or just to be sure there's none?

misberner: Well, I would suggest to start with a microbenchmark (`_bench_test.go`) to assess if there is much of a performance impact at all. If there's none, or none that is significant, then no, I don't think we need to do any further tests. However, if a microbenchmark suggests that there is some noticeable degradation, we have to see further.

0x656b694d:
Primitive add/remove benchmark
```go
func BenchmarkSet(b *testing.B) {
	a := NewIntSet()
	for i := 0; i < 1000000; i++ {
		a.Add(i)
	}
	for i := 0; i < 1000000; i++ {
		a.Remove(i)
	}
}
```

Results of both generated and generic versions are practically the same, on both go versions 1.18.6 and 1.19.1:
```
goos: linux
goarch: amd64
pkg: github.com/stackrox/rox/pkg/set
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkSet-8   	1000000000	         0.2433 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/stackrox/rox/pkg/set	3.674s
```

0x656b694d:
Thanks, TIL about go benchmark tests.

> Also, please add the benchmark to the source code directly, and _always_ include the results of all benchmarking results, even if they are very similar.
The results were different per every run, and showing several would be misleading, as there was no clear winner.

I will add better testing for the new code, but I'm not sure how to add tests for the removed generated code, if only you want me to add them in a separate PR to master.

0x656b694d:
Generated set:
```
goos: linux
goarch: amd64
pkg: github.com/stackrox/rox/pkg/set
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkIntSet-8      	   10000	    341897 ns/op	   40246 B/op	      30 allocs/op
BenchmarkStringSet-8   	    4888	    708540 ns/op	   86052 B/op	    2031 allocs/op
PASS
ok  	github.com/stackrox/rox/pkg/set	6.972s
```

Generic set:
```
goos: linux
goarch: amd64
pkg: github.com/stackrox/rox/pkg/set
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkIntSet-8          10000            327385 ns/op           40464 B/op         30 allocs/op
BenchmarkStringSet-8        4797            634919 ns/op           87368 B/op       2031 allocs/op
PASS
ok      github.com/stackrox/rox/pkg/set 6.411s
```

pkg/set/set_bench_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"strconv"
+	"testing"
+)
+
+var (
+	a IntSet
+	s StringSet
+)
```
misberner, line 11: Why are these global vars, and not just local variables defined in the `Benchmark..` function via `:=`?

misberner:
> The results were different per every run, and showing several would be misleading, as there was no clear winner.

If results vary widely, that's usually a sign that there's an issue with the benchmark like the one I pointed out. Benchmarks are supposed to be reproducible, at least on a single machine.

> I will add better testing for the new code, but I'm not sure how to add tests for the removed generated code, if only you want me to add them in a separate PR to master.

Yeah, totally fine to not add benchmarks/tests for the removed code. What I asked for was the code of a benchmark as part of the PR, and this benchmark ideally written in such a way (i.e., using a type like `StringSet`, not `Set[string]`) that it works with both the existing code and the new one. Just to ensure that we know exactly what benchmark we are talking about, that I can look at it, and that I see that there's nothing that structurally favors the old or the new set implementation.

Results look great. There's a code style concern in the `_bench_test.go`, but otherwise 🚢 it!

pkg/set/set_bench_test.go
0x656b694d: As recommended by the link you suggested, these global vars may help to ensure the benchmark code is not optimized out.

misberner: SG

PR #2513, opened 2022-07-25
ROX-11136 implementation for action widget config persistence

dvail:
## Description

This implements the `useWidgetConfig` hook to persist dashboard options in local storage.

Note: This hook makes a _synchronous_ read of local storage each time it is initialized as the starting state for an internal `useState` hook. If we transition this persistence to live on the backend in the future we will need to pull the initial call up to a higher level (Redux, useContext, react-query cache (?)) than it is in the current components.  An async read here will cause many of these components to make two network calls on mount; one when all the component has are the default options, and another when the saved options are loaded.

## Checklist

## Testing Performed

- Edit existing unit/integration tests in response to failures (dirty local storage state between test runs).

Load the dashboard and update widget options:

- In "Images at most risk", set Image vulns to "All CVEs" and Image status to "All images"
- In "Aging images", deselect time buckets 1 and 3, and also set time bucket 4 to 265 days
- In "Policy violations by severity", set Sort by to "Total" and Lifecycle to "Deploy"
- In "Policy violations by severity" toggle _off_ the Critical, Medium, and Low violations via the legend
- In "Compliance by standard", set Sort By to "descending"

Refresh the page, and check that the above options for each widget remain set.

Visit other pages in the app and then revisit the dashboard via the "Dashboard" navigation link and check that the above options for each widget remain set.

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
@@ -1,4 +1,5 @@
-import React, { useReducer } from 'react';
+import React from 'react';
+import { useLocation } from 'react-router-dom';
 import {
     Flex,
     FlexItem,
@@ -18,6 +19,7 @@ import pluralize from 'pluralize';
 
 import LinkShim from 'Components/PatternFly/LinkShim';
 import useURLSearch from 'hooks/useURLSearch';
+import useWidgetConfig from 'hooks/useWidgetConfig';
 import { getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import { SearchFilter } from 'types/search';
@@ -93,12 +95,18 @@ function getWidgetTitle(
     return `${totalImages} Aging ${pluralize('image', totalImages)}`;
 }
 
-const defaultTimeRanges: TimeRangeTuple = [
-    { enabled: true, value: 30 },
-    { enabled: true, value: 90 },
-    { enabled: true, value: 180 },
-    { enabled: true, value: 365 },
-];
+type Config = {
+    timeRanges: TimeRangeTuple;
+};
+
+const defaultConfig: Config = {
+    timeRanges: [
+        { enabled: true, value: 30 },
+        { enabled: true, value: 90 },
+        { enabled: true, value: 180 },
+        { enabled: true, value: 365 },
+    ],
+};
 
 type TimeRangeAction =
     | {
@@ -111,14 +119,16 @@ type TimeRangeAction =
           value: number;
       };
 
-function timeRangeReducer(state: TimeRangeTuple, action: TimeRangeAction) {
+function timeRangeReducer(state: Config, action: TimeRangeAction): Config {
     const nextState = cloneDeep(state);
+    const { timeRanges } = nextState;
+
     switch (action.type) {
         case 'toggle':
-            nextState[action.index].enabled = !nextState[action.index].enabled;
+            timeRanges[action.index].enabled = !timeRanges[action.index].enabled;
             return nextState;
         case 'update':
-            nextState[action.index].value = action.value;
+            timeRanges[action.index].value = action.value;
             return nextState;
         default:
             return nextState;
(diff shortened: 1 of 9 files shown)


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -223,10 +240,10 @@ function AgingImages() {
                                                     <TextInput
                                                         aria-label="Image age in days"
                                                         style={{ minWidth: '100px' }}
-                                                        onChange={(val) => {
+                                                        onChange={async (val) => {
                                                             const value = parseInt(val, 10);
                                                             if (!(value >= maxTimeRange)) {
-                                                                dispatch({
+                                                                await dispatch({
```
dvail, line 246: With the change to have this value backed by localStorage, the `dispatch` function now returns a Promise. The async/await here is purely to avoid the eslint "no dangling promises" error.

Is there another pattern that should be used instead here?

This API returns a Promise in order to make it easier to transition to server-side storage in the future if needed, but we currently have no reason to wait on the result of these promises.

ui/apps/platform/src/types/widgetConfig.ts
```diff
@@ -19,13 +19,13 @@ type WidgetConfigMap = Partial<Record<RouteId, WidgetConfig>>;
 export type RouteId = string;
 
 // Configuration read for a single instance of a widget.
-export type WidgetConfig = Partial<Record<string, ConfigOptionValue>>;
+export type WidgetConfig = Readonly<Partial<Record<string, ConfigOptionValue>>>;
```
dvail, line 22: These structures being mutable has already caused some annoying bugs in development, so I've updated them to be `readonly`.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
```diff
@@ -151,7 +161,14 @@ function getViewAllLink(searchFilter: SearchFilter) {
 function AgingImages() {
     const { isOpen: isOptionsOpen, onToggle: toggleOptionsOpen } = useSelectToggle();
     const { searchFilter } = useURLSearch();
-    const [timeRanges, dispatch] = useReducer(timeRangeReducer, defaultTimeRanges);
+    const { pathname } = useLocation();
+
+    const [{ timeRanges }, dispatch] = useWidgetConfig<Config, TimeRangeAction>(
```
vjwilson, line 166: Why `dispatch` here, and `updateConfig` without any async/await in all the other widgets?

vjwilson (commented):
The approach looks good. Just had a question about difference in one widget, and possible stale code in another.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useState, useCallback } from 'react';
-import { useHistory } from 'react-router-dom';
+import React, { useState, useMemo, useCallback } from 'react';
+import { useLocation, useHistory } from 'react-router-dom';
```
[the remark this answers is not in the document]
dvail: Yep, we use it down on line 201 to push to the history when the chart bar is clicked. https://github.com/stackrox/stackrox/blob/a76d0dc0ae478195ccc0c6c52eeec64e6af9e8d9/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx#L201

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.tsx
dvail: The `useWidgetConfig` hook's second return value is a function to update the config state, which defaults to a simple merge function. If you pass a reducer function as the hook's 4th argument though, it will be used as the state update function instead of the default merge function.

I kept this as `dispatch` to be the same as when this component used `useReducer`, but I don't love the inconsistency either. Do you think calling it `updateConfig` everywhere is better, even if the usage is slightly different?

The async/await is because this is the one case where the update call is not the last line of a function (in the other cases I just `returned` the function call.) This is to get around the eslint `no-dangling-promises` error. The other alternative to the async/await is to `return` the function call as well, but with an extra `return undefined` at the end. I can update to do that an add a short comment for consistency.

vjwilson: Ah, I didn't check the code inside `useWidgetConfig` because it was merged in an earlier PR.

I agree there is no really good solution. You could merge as-is; no need to belabor the point.

We should have a lunch-and-learn once all the team is back from PTO, and get feedback on the pattern. We may get some suggestions for future optimization.

vjwilson (approved):
Questions asked, questions answered.

(Gen X irony can't be turned off 😉 )

PR #2520, opened 2022-07-26
ROX-11573 Update network graph legend to use expanded PatternFly design

dvail:
## Description

Updates the Network Graph legend to use PatternFly styles and be more readable with the text auto-expanded instead of contained in tooltips.

## Notes

- The `Legend` and `LegendTile` components are _technically_ an update of the existing components, but the changes are large enough that they are probably better reviewed as if they are new files. The unified diff is not going to be terribly helpful. I also moved `LegendTile` to colocate with the `Legend` file, since both are single-use Network Graph components.

- The previous `LegendTile` component used a mix of SVG icons and text from the "stackrox" font, driven by props passed from the parent. For consistency, I updated this to use only SVG icons. This lets us get rid of the branching logic in `LegendTIle`, remove a few single use constants, and remove some global CSS styles.

- It looks like the "stackrox" font is only used in the Network Graph itself now, so it can be removed once the UI is completely updated to PatternFly. Assuming there are not future plans for this font - it looks like it was last added 4 years ago.

## Checklist

## Testing Performed

Visit the Network Graph and view the new legend in the bottom-left:
image

Clicking on the 'X' icon in the corner should collapse the legend:
image

Clicking the collapsed legend should re-open it.
image

Tabbing through the page elements should land you on the collapsed legend button. Pressing the 'enter' key should open the legend (note the blue "focused" style applied).
image


dvail:
I see the merge conflict from changes from 2+ years ago - trying to figure out what is causing that.

vjwilson:
> * It looks like the "stackrox" font is only used in the Network Graph itself now, so it can be removed once the UI is completely updated to PatternFly. Assuming there are not future plans for this font - it looks like it was last added 4 years ago.

^ @alanonthegit Do you know of any reason we can't remove the "stackrox" font from the app?

ui/apps/platform/cypress/constants/NetworkPage.js
```diff
@@ -20,9 +20,9 @@ export const selectors = {
     simulatorSuccessMessage: 'div[data-testid="message-body"]:contains("Policies processed")',
     panels: networkPanels,
     legend: {
-        deployments: '[data-testid="deployment-legend"] div',
-        namespaces: '[data-testid="namespace-legend"] div',
-        connections: '[data-testid="connection-legend"] div',
+        deployments: '[data-testid="deployment-legend"] li',
+        namespaces: '[data-testid="namespace-legend"] li',
+        connections: '[data-testid="connection-legend"] li',
```
vjwilson, line 25: nit: We've got a dev-defined precise selector target, but we still have to switch from one brittle child selector to another.

ui/apps/platform/src/Containers/Network/Graph/Overlays/Legend.tsx
```diff
@@ -70,23 +68,48 @@ function Legend(): ReactElement {
     }
 
     return (
-        <div
+        <Card
```
vjwilson, line 71: Now that I see we're using a PF Card, I wonder if, in the future we should adopt the [Expandable Card variant](https://www.patternfly.org/v4/components/card#expandable). It's less bespoke, and you get the indication that it's closed and can be clicked to open. (I'm not sure if we can have it open "up", though.)

@alanonthegit What do you think about that for the future?

vjwilson (approved):
Good job. 

(Just had a couple of thoughts for the future.)

ui/apps/platform/cypress/constants/NetworkPage.js
dvail: Good call, I've updated this to be selector agnostic and just check "the n-th item in the legend". I also update this to use less Cypress chaining, which is more in line with how we have been doing recent tests.

dvail:
For the `stackrox` font, we're still currently using that for some of the items in the network graph itself. If/when we switch to the PatternFly topology extension I believe all usage of the stackrox font will be removed.

PR #2523, opened 2022-07-26
[ROX-11912] Add support for clusters and cluster count for cluster vulnerabilities

md2119:
## Description

This is new support. We never had a clusters column on cluster/platform cves page.

## Checklist


charmik-redhat (approved):
lgtm

PR #2537, opened 2022-07-27
ROX-11984: Match network policies against pod labels instead of deployment labels

vikin91:
## Description

Matching Network Policies selectors against the `deployment.GetLabels()` caused issues* in the following cases:

- When deployment labels were different than the labels of pods created by that deployment  - for an example, see test case below

It worked correctly for:

- Rouge pods (pods with no deployment)
- Pods with deployments where labels of the deployment and its pods were identical.

The (*) issues are described in the ticket but they mainly include false-negative and false-positive alerts for violating the system policies for network policies.

The fix is a oneliner - calling `deployment.GetPodLabels()` ensures that the correct fields are analyzed in all mentioned cases. 

## Checklist

## Testing Performed

- Manually by starting a cluster and applying the scenarios from [the ticket comment](https://issues.redhat.com/browse/ROX-11984?focusedCommentId=20696841&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20696841)

### Case 1 - Policy matches deployment labels

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
    xname: nginx-deployment
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        xname: nginx-pod
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-deny-ingress-xname
spec:
  podSelector:
    matchLabels:
      xname: nginx-deployment
  policyTypes:
  - Ingress
```

1. Should cause policy violation ✅
2. Network graph should display `No network policies defined for this deployment` when clicking on the deployment in the graph and selecting the "Network Policies" tab ✅

### Case 2 - Policy matches pod labels

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
    xname: nginx-deployment
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        xname: nginx-pod
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-deny-ingress-xname
spec:
  podSelector:
    matchLabels:
      xname: nginx-pod
  policyTypes:
  - Ingress
```

1. Should NOT cause policy violation ✅
1. Network graph should display network policy  `default-deny-ingress-xname` when clicking on the deployment in the graph and selecting the "Network Policies" tab ✅

### DIFF
--- a/pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
+++ b/pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
@@ -65,11 +65,10 @@ func (suite *NetworkPolicySuite) Test_FilterForDeployment() {
 			expectedPoliciesMatched: 2,
 		},
 		"No Match: NW policies shall not match against the deployment label even if 0 pods match": {
-			deploymentLabels:      map[string]string{"app": "central", "env": "prod"},
-			specPodTemplateLabels: map[string]string{"app": "sensor", "env": "dev"},
+			deploymentLabels:      map[string]string{"app": "central-deployment", "env": "prod"},
+			specPodTemplateLabels: map[string]string{"app": "central-pod", "env": "prod"},
 			netpolSelectors: []map[string]string{
-				{"app": "central"},
-				{"env": "prod"},
+				{"app": "central-deployment", "env": "prod"},
 			},
 			expectedPoliciesMatched: 0,
 		},
--- a/pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
+++ b/pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
@@ -26,25 +26,29 @@ func policy(classificationEnums []storage.NetworkPolicyType) *storage.NetworkPol
 func (suite *NetworkPolicySuite) Test_FilterForDeployment() {
 	cases := map[string]struct {
 		deploymentLabels        map[string]string
+		specPodTemplateLabels   map[string]string
 		netpolSelectors         []map[string]string
 		expectedPoliciesMatched int
 	}{
 		"Match: one NW policy with one label": {
-			deploymentLabels: map[string]string{"app": "central"},
+			deploymentLabels:      map[string]string{"app": "central"},
+			specPodTemplateLabels: map[string]string{"app": "central"},
 			netpolSelectors: []map[string]string{
 				{"app": "central"},
 			},
 			expectedPoliciesMatched: 1,
 		},
 		"Match: one NW policy with two labels": {
-			deploymentLabels: map[string]string{"app": "central", "env": "prod"},
+			deploymentLabels:      map[string]string{"app": "central", "env": "prod"},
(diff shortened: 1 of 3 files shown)


pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
```diff
@@ -26,48 +26,81 @@ func policy(classificationEnums []storage.NetworkPolicyType) *storage.NetworkPol
...
+			specPodTemplateLabels: map[string]string{"app": "central", "env": "prod"},
 			netpolSelectors: []map[string]string{
 				{"app": "central"},
 				{"env": "prod"},
 			},
 			expectedPoliciesMatched: 2,
 		},
+		"No Match: NW policies shall not match against the deployment label even if 0 pods match": {
+			deploymentLabels:      map[string]string{"app": "central", "env": "prod"},
+			specPodTemplateLabels: map[string]string{"app": "sensor", "env": "dev"},
```
fredrb, line 69: _extreme nit:_ I think it's a bit confusing to use `central` as deployment label and `sensor` as pod label. I like to think of unit tests as examples of how things work, and that is not a "realistic" scenario that might make the reader more confused. I still think that dumb names are more helpful here, something like `label=a` vs `label=b`.

fredrb: After reading through the reamining of the tests I think you could use the same labels from test `"One Match: NW policies shall match against the pod labels only"` but just change the netpol selector to `central-deployment` instead.

fredrb (approved):
LGTM. Only a tiny suggestion.

pkg/booleanpolicy/networkpolicy/networkpolicy_test.go
vikin91: Okay, that makes sense. I will change this and will also do a rebase hoping for the same green CI as now.

sensor/common/detector/detector.go
```diff
@@ -414,7 +414,7 @@ func (d *detectorImpl) getNetworkPoliciesApplied(deployment *storage.Deployment)
 		// It is fine (from the Matcher perspective) to use nil augmented objects
 		return nil
 	}
-	networkPolicies := d.networkPolicyStore.Find(deployment.GetNamespace(), deployment.GetLabels())
+	networkPolicies := d.networkPolicyStore.Find(deployment.GetNamespace(), deployment.GetPodLabels())
```
md2119, line 417: :)

md2119 (approved):

PR #2542, opened 2022-07-27
ROX-12029 Update the images at most risk dashboard query to not use deprecated gql sub-resolver.

dvail:
## Description

Replaces the deprecated `vulnCounter` subresolver with `imageVulnerabilityCounter` in the main page dashboard "Images at most risk" widget's graphql query.

## Checklist

## Testing Performed

Set up an infra cluster with the following ENV var
`ROX_POSTGRES_DATASTORE=true`

Behavior before this change:
image

Behavior after this change:
image

With "Image Vulns:All CVEs" setting:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ImagesAtMostRisk.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ImagesAtMostRisk.test.tsx
@@ -6,7 +6,7 @@ import '@testing-library/jest-dom/extend-expect';
 
 import renderWithRouter from 'test-utils/renderWithRouter';
 import { vulnManagementImagesPath, vulnManagementPath } from 'routePaths';
-import ImagesAtMostRisk, { imagesQuery } from './ImagesAtMostRisk';
+import ImagesAtMostRisk, { getImagesQuery } from './ImagesAtMostRisk';
 
 function makeMockImage(
     id: string,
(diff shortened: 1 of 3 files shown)


vjwilson (approved):
/lgtm

md2119 (approved):
Recommend fork

vjwilson:
> Recommend fork

Ah, Mandar caught something.

We should make the field name conditional, based on the `ROX_FRONTEND_VM_UPDATES`

We can just gate the query, either `vulnCounter` or aliased as `vulnCounter: imageVulnerabilityCounter`, and then leave all the usages `vulnCounter`.

dvail:
Interesting @md2119 , the new resolver in fact _does_ return different data. The following screenshots are from a new install on a plain `gke-default` infra cluster with `ROX_POSTGRES_DATASTORE` _off_. Notice the number of reported CVEs for `calico-node`:

using `vulnCounts` sub-resolver
image

using `imageVulnerabilityCounts` sub-resolver
image

Running this again against a new cluster using Postgres, the results across all combinations for `calico-node` are:
|Sub-resolver|datastore|Critical CVEs|Important CVEs|
|---|---|---|---|
|`vulnCounter`|RocksDB|0|4|
|`imageVulnerabilityCounter`|RocksDB|2|5|
|`vulnCounter`|Postgres|ERROR|ERROR|
|`imageVulnerabilityCounter`|Postgres|0|4|

md2119 (approved):

md2119:
Verified on graphQL on demo clusters with and without Postgres. I see the same data.
with postgres
image

without postgres:
image

PR #2550, opened 2022-07-28
UI unit test debugging enhancements

dvail:
### Motivation:

We have been getting occasional test failures in CI due to a timeout of > 5000ms in some of the newer tests. It was difficult to pinpoint why this was occurring, and whether or not the issue was something unrelated to the timeout. These changes should help avoid the timeouts, as well as make debugging these failures in CI easier in the future.

### Proposal:
- Remove coverage reporting for ui unit tests when running in CI. We do not track the coverage in any way (e.g. via SonarQube or similar), and it slows down execution of the job. (roughly 50% slower in my dev environment)
- Add `verbose=true` to get timing details for individual tests to better pinpoint problem tests in CI. The total time of the test doesn't give an accurate view of the individual test runs, for reasons described below.
- Add `--slowTestThreshold=1`, another optional parameter for reporting purposes. This gives us a better view of total time for each test _file_. It does not impact test execution.
- Disable gql tag warnings due to duplicate fragment names that clutter test output.
- Bump up the jest timeout threshold for tests that are taking substantial time in CI (from 5s to 10s)

### Investigation:

Running the tests locally and in CI with these settings reveals the following:

- Jest setup time per test is where _the vast majority_ of time is spent in the ui-unit-tests job. The number next to the filename is the total time that file spent executing, while the numbers next to individual specs is how long that specific test spent executing. In the following example, the actual test execution was 10% or less of the total time Jest spent in the tests, with the remaining time caused by jest overhead. Almost all other tests are similar.

```
# Prow CI

PASS src/hooks/useURLParameter.test.tsx (10.846 s)
✓ should read/write scoped string value in URL parameter without changing existing URL parameters (64 ms)
✓ should read/write scoped complex object in URL parameter without changing existing URL parameters (66 ms)
✓ should implement push and replace state for history (16 ms) 
```

```
# Local

PASS  src/hooks/useURLParameter.test.tsx (1.088 s)
✓ should read/write scoped string value in URL parameter without changing existing URL parameters (10 ms)
✓ should read/write scoped complex object in URL parameter without changing existing URL parameters (9 ms)
✓ should implement push and replace state for history (4 ms)
```

- Some tests that were failing with a > 5000ms error might actually have been timing out in test runs. Jest does not account for the setup time in these failures, so the long total time is sometimes indicative of slow tests, and sometimes isn't. In the below case, the second `ScopeBar` test could easily cross the 5000ms mark if the system was under load. (Note that this test is 10x slower than a local run, while the other tests are 4-6x slower on average).

```
# Prow CI

PASS src/Containers/Dashboard/PatternFly/ScopeBar.test.tsx (16.716 s)
Resource scope bar
✓ should default to all clusters and namespaces selected (897 ms)
✓ allows selection of multiple clusters and namespaces (3556 ms)
✓ will track selected clusters and namespaces in the page URL (1296 ms) 
```

```
# Local

PASS  src/Containers/Dashboard/PatternFly/ScopeBar.test.tsx (2.237 s)
Resource scope bar
✓ should default to all clusters and namespaces selected (128 ms)
✓ allows selection of multiple clusters and namespaces (364 ms)
✓ will track selected clusters and namespaces in the page URL (391 ms)
```

Interestingly enough, not all tests scaled this poorly in CI, and overall the CI run for all tests was pretty much that same as the total time locally:

```
# Prow CI
@stackrox/platform-app: Test Suites: 101 passed, 101 total
@stackrox/platform-app: Tests:       1 skipped, 721 passed, 722 total
@stackrox/platform-app: Snapshots:   43 passed, 43 total
@stackrox/platform-app: Time:        44.962 s 
```

```
# Local
@stackrox/platform-app: Test Suites: 101 passed, 101 total
@stackrox/platform-app: Tests:       1 skipped, 721 passed, 722 total
@stackrox/platform-app: Snapshots:   43 passed, 43 total
@stackrox/platform-app: Time:        44.118 s
```

This makes me think that in general, Prow is more efficient at handling the tests, but experiences significant slowdowns in certain cases. (Complex imports in files under tests, usage of JSDOM, usage of `getByRole` seem to have the worst impact on performance.) One theory is that in CI we have many more cores available to handle Jest parallelization, but individual core speed is less, leading to slower test execution.

## Checklist

## Testing Performed

Run unit tests normally (`yarn test`)
- Individual test executions times should be displayed for all tests
- The console should not be flooded with duplicate gql fragment warnings


vjwilson (approved):

pedrottimark (approved):
Super, thank you!

PR #2554, opened 2022-07-28
refactor graphql deprecations on various components

keyallis:
## Description

This refactor is to more clearly define fields to be deprecated in the future and allow FE querying on the nested CVEInfo object to allow for migration towards in the future.

### DIFF
--- a/central/graphql/resolvers/common_vulnerabilities.go
+++ b/central/graphql/resolvers/common_vulnerabilities.go
@@ -13,49 +13,68 @@ import (
  * NOTE: This list is and should remain alphabetically ordered
  */
 var commonVulnerabilitySubResolvers = []string{
-	"createdAt: Time", // Discovered At System
-	"cve: String!",
+	"cveBaseInfo: CVEInfo",
 	"cvss: Float!",
 	"envImpact: Float!",
 	"fixedByVersion: String!",
 	"id: ID!",
 	"impactScore: Float!",
 	"isFixable(query: String): Boolean!",
-	"lastModified: Time",
 	"lastScanned: Time",
-	"link: String!",
-	"publishedOn: Time",
-	"scoreVersion: String!",
 	"severity: String!",
-	"summary: String!",
-	"suppressActivation: Time",
-	"suppressExpiry: Time",
-	"suppressed: Boolean!",
+	"snoozeExpiry: Time",
+	"snoozeStart: Time",
+	"snoozed: Boolean!",
 	"unusedVarSink(query: String): Int",
 	"vectors: EmbeddedVulnerabilityVectors",
+
+	// deprecated fields
+	"suppressActivation: Time @deprecated(reason: \"use 'snoozeStart'\")",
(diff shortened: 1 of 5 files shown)


keyallis:
\test all

central/graphql/resolvers/common_vulnerabilities.go
```diff
@@ -13,49 +13,68 @@ import (
...
+	"suppressed: Boolean! @deprecated(reason: \"use 'snoozed'\")",
+
+	// CVEInfo fields
+	"createdAt: Time @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"cve: String! @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"lastModified: Time @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"link: String! @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"publishedOn: Time @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"scoreVersion: String! @deprecated(reason: \"use 'cveBaseInfo'\")",
+	"summary: String! @deprecated(reason: \"use 'cveBaseInfo'\")",
```
md2119, line 43: I do not recommend deprecating these fields because it allows us to be agnostic of a stored object structure. In short, the service response structure need not be exact as the stored structure.

md2119 (changes requested):

central/graphql/resolvers/image_components.go
```diff
@@ -77,6 +78,11 @@ type ImageComponentResolver interface {
 	TopImageVulnerability(ctx context.Context) (ImageVulnerabilityResolver, error)
 	UnusedVarSink(ctx context.Context, args RawQuery) *int32
 	Version(ctx context.Context) string
+
+	// deprecated functions
+
+	FixedIn(ctx context.Context) string
+	ID(ctx context.Context) graphql.ID
```
charmik-redhat, line 85: No need to deprecate `ID`, can be directly removed as nothing changes for the GraphQL
schema. Request to `id` field in GraphQL query will be automatically routed to `Id` function.

central/graphql/resolvers/image_components.go
```diff
@@ -56,9 +58,8 @@ type ImageComponentResolver interface {
 	ActiveState(ctx context.Context, args RawQuery) (*activeStateResolver, error)
 	DeploymentCount(ctx context.Context, args RawQuery) (int32, error)
 	Deployments(ctx context.Context, args PaginatedQuery) ([]*deploymentResolver, error)
-	FixedIn(ctx context.Context) string
 	FixedBy(ctx context.Context) string
-	ID(ctx context.Context) graphql.ID
+	Id(ctx context.Context) graphql.ID
```
md2119, line 62: nit: `ID` because it is abbrev

md2119 (approved):

central/graphql/resolvers/image_components.go
keyallis: The reason I'm making this change is that the `Id` function is autogen'd based off the proto

PR #2555, opened 2022-07-28
Remove completed TODOs from code

md2119:
## Checklist


central/role/service/service_impl.go
```diff
@@ -117,12 +117,6 @@ func (s *serviceImpl) CreateRole(ctx context.Context, roleRequest *v1.CreateRole
 	}
 	role.Name = roleRequest.GetName()
 
-	// Empty access scope ID is deprecated. Fill the default during the adoption
-	// period.
-	// TODO(ROX-9510): remove this block.
```
rukletsov, line 122: This only partially resolves ROX-9510. Let's either address ROX-9510 completely or leave this out. I'm fine with any option.

central/graphql/resolvers/gen/main.go
```diff
@@ -90,12 +90,6 @@ var (
 				ParentType: reflect.TypeOf(storage.Node{}),
 				FieldName:  "Scan",
 			},
-			// TODO(ROX-6194): Remove this entirely after the deprecation cycle started with the 55.0 release.
```
rukletsov, line 93: I'd suggest removing everything related to ROX-6194 in one PR and close the ticket. Can happen here or separately.

md2119: Everything related to ROX-6194 has already been removed except this,

```
message Policy {
...
reserved                                            10; //was repeated Exclusion whitelists
...
}
```

central/role/service/service_impl.go
md2119: Will revert this change and defer the changes to Authz team.

central/graphql/resolvers/gen/main.go
rukletsov: I also see mentions in:
* `pkg/env/baseline.go`
* `qa-tests-backend/src/test/groovy/RiskTest.groovy`

PR #2562, opened 2022-07-31
ROX-11909: save kuttl JUnit XML to artifacts

ivan-degtiarenko:
## Description

Prow can display tests that fail in its JUnit panel. e.g. https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-go-unit-tests/1550162970292523008

For Prow to display tests that fail in its JUnit panel it needs files in JUnit XML format copied to $ARTIFACT_DIR. This PR does that for `openshift-4-operator-e2e-tests`, which are run via `kuttl`.

## Checklist

## Testing Performed

1. Verified in CI that artifacts are in place, see `ci/prow/openshift-4-operator-e2e-tests `

### DIFF
--- a/.openshift-ci/ci_tests.py
+++ b/.openshift-ci/ci_tests.py
@@ -55,7 +55,6 @@ class OperatorE2eTest(BaseTest):
     UPGRADE_TEST_TIMEOUT_SEC = 50 * 60
     E2E_TEST_TIMEOUT_SEC = 50 * 60
     SCORECARD_TEST_TIMEOUT_SEC = 20 * 60
-    ARTIFACTS_TIMEOUT = 3 * 60
 
     def __init__(self):
         self.test_outputs = [
@@ -76,26 +75,10 @@ class OperatorE2eTest(BaseTest):
             OperatorE2eTest.UPGRADE_TEST_TIMEOUT_SEC
         )
 
-        print("Storing test-upgrade results in JUnit format")
-        self.run_with_graceful_kill(
-            ["scripts/ci/store-artifacts.sh", "store_test_results",
-             "operator/build/kuttl-test-artifacts",
-             "operator-test-results-upgrade"],
-            OperatorE2eTest.ARTIFACTS_TIMEOUT
-        )
-
         print("Executing operator e2e tests")
         self.run_with_graceful_kill(
             ["make", "-C", "operator", "test-e2e-deployed"],
             OperatorE2eTest.E2E_TEST_TIMEOUT_SEC,
-        )
-
-        print("Storing test-e2e-deployed results in JUnit format")
-        self.run_with_graceful_kill(
-            ["scripts/ci/store-artifacts.sh", "store_test_results",
-             "operator/build/kuttl-test-artifacts",
-             "operator-test-results-e2e-deployed"],
-            OperatorE2eTest.ARTIFACTS_TIMEOUT
         )
 
         print("Executing Operator Bundle Scorecard tests")
--- a/.openshift-ci/ci_tests.py
+++ b/.openshift-ci/ci_tests.py
@@ -55,6 +55,7 @@ class OperatorE2eTest(BaseTest):
     UPGRADE_TEST_TIMEOUT_SEC = 50 * 60
     E2E_TEST_TIMEOUT_SEC = 50 * 60
     SCORECARD_TEST_TIMEOUT_SEC = 20 * 60
+    ARTIFACTS_TIMEOUT = 3 * 60
 
     def __init__(self):
         self.test_outputs = [
@@ -72,13 +73,29 @@ class OperatorE2eTest(BaseTest):
         print("Executing operator upgrade test")
         self.run_with_graceful_kill(
             ["make", "-C", "operator", "test-upgrade"],
-            OperatorE2eTest.UPGRADE_TEST_TIMEOUT_SEC,
+            OperatorE2eTest.UPGRADE_TEST_TIMEOUT_SEC
+        )
+
+        print("Storing test-upgrade results in JUnit format")
+        self.run_with_graceful_kill(
+            ["scripts/ci/store-artifacts.sh", "store_test_results",
+             "operator/build/kuttl-test-artifacts",
+             "operator-test-results-upgrade"],
+            OperatorE2eTest.ARTIFACTS_TIMEOUT
         )
 
         print("Executing operator e2e tests")
         self.run_with_graceful_kill(
             ["make", "-C", "operator", "test-e2e-deployed"],
             OperatorE2eTest.E2E_TEST_TIMEOUT_SEC,
+        )
+
+        print("Storing test-e2e-deployed results in JUnit format")
+        self.run_with_graceful_kill(
+            ["scripts/ci/store-artifacts.sh", "store_test_results",
+             "operator/build/kuttl-test-artifacts",
+             "operator-test-results-e2e-deployed"],
+            OperatorE2eTest.ARTIFACTS_TIMEOUT
         )
 
         print("Executing Operator Bundle Scorecard tests")
--- a/.openshift-ci/ci_tests.py
+++ b/.openshift-ci/ci_tests.py
@@ -12,6 +12,7 @@ from common import popen_graceful_kill
 class BaseTest:
     def __init__(self):
         self.test_outputs = []
+        self.test_results = {}
 
     def run_with_graceful_kill(self, args, timeout, post_start_hook=None):
         with subprocess.Popen(args) as cmd:
@@ -57,10 +58,10 @@ class OperatorE2eTest(BaseTest):
     SCORECARD_TEST_TIMEOUT_SEC = 20 * 60
 
     def __init__(self):
-        self.test_outputs = [
-            "operator/build/kuttl-test-artifacts",
-            "operator/build/kuttl-test-artifacts-upgrade",
-        ]
+        self.test_results = {
+            "operator/build/kuttl-test-artifacts": "kuttl-test-artifacts",
+            "operator/build/kuttl-test-artifacts-upgrade": "kuttl-test-artifacts-upgrade",
+        }
 
     def run(self):
         print("Deploying operator")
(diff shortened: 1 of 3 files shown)


.openshift-ci/ci_tests.py
```diff
@@ -75,6 +84,14 @@ def run(self):
             OperatorE2eTest.E2E_TEST_TIMEOUT_SEC,
         )
 
+        print("Storing test-e2e-deployed results in JUnit format")
+        self.run_with_graceful_kill(
```
porridge, line 88: Would be nice to refactor this `print`+`run` into a couple of methods:
- one in `OperatorE2eTest` which contains pieces specific to kuttl artifacts, or perhaps specific to operator tests; it would delegate to:
- another generic method in `BaseTest` that calls `store-artifacts.sh` and would potentially be useful to other subclasses.

porridge (approved):
Awesome!

.openshift-ci/ci_tests.py
```diff
@@ -66,7 +67,15 @@ def run(self):
         print("Executing operator upgrade test")
         self.run_with_graceful_kill(
```
gavin-stackrox, line 68: If a test fails I would expect this to raise an error and the junit logs would not be gathered. What about using the `post_start_hook=set_dirs_after_start` convention used elsewhere in this file? Or just set `test_output_dirs` in a `OperatorE2eTest` `__init__()`?

gavin-stackrox (changes requested):

.openshift-ci/ci_tests.py
porridge: Hm, good point, this will only gather logs when we don't need them, :-)
Another way would be to change this artifact gathering into a context manager, or something?

ivan-degtiarenko: @gavin-stackrox do you suggest creating new `post_test` like `StoreTestResults`? Because `StoreArtifacts` doesn't add them to JUnit panel

gavin-stackrox: @ivan-degtiarenko actually, now that I think about it a bit more, maybe a better approach for the operator e2e tests is something similar to what is done with `UIE2eTest`. i.e. it is just a wrapper around `tests/e2e/run-ui-e2e.sh` which can then do all the error handling and JUnit artifact saving via `store_test_results()`.

.openshift-ci/post_tests.py
```diff
@@ -204,6 +209,14 @@ def check_stackrox_logs(self):
             timeout=PostTestsConstants.CHECK_TIMEOUT,
         )
 
+    def add_test_output(self):
+        print("Storing operator test results in JUnit format")
+        self.run_with_best_effort(
+            ["scripts/ci/store-artifacts.sh", "store_test_results",
+             "operator/build/kuttl-test-artifacts"],
```
gavin-stackrox, line 216: I think based on @porridge changes there will be different dirs now for different tests. i.e. not all are "operator/build/kuttl-test-artifacts"

what about adding an array of junit test output passed to PostClusterTest() then this change is not so operator specific.

gavin-stackrox (changes requested):

.openshift-ci/ci_tests.py
```diff
@@ -57,10 +58,10 @@ class OperatorE2eTest(BaseTest):
     SCORECARD_TEST_TIMEOUT_SEC = 20 * 60
 
     def __init__(self):
-        self.test_outputs = [
-            "operator/build/kuttl-test-artifacts",
-            "operator/build/kuttl-test-artifacts-upgrade",
-        ]
+        self.test_results = {
+            "operator/build/kuttl-test-artifacts": "kuttl-test-artifacts",
```
porridge, line 62: I think it might be better to have the dict the other way, since we primarily want to make sure the destination name is unique..

.openshift-ci/post_tests.py
```diff
@@ -105,9 +106,11 @@ def __init__(
         collect_central_artifacts=True,
         check_stackrox_logs=False,
         artifact_destination_prefix=None,
+        add_test_result=True,
```
porridge, line 109: I think keeping this in plural to match the `test_results` argument name would be better for consistency.

.openshift-ci/post_tests.py
```diff
@@ -75,7 +76,7 @@ def __init__(
         self.artifact_destination_prefix = artifact_destination_prefix
         self.data_to_store = []
 
-    def run(self, test_outputs=None):
+    def run(self, test_outputs=None, test_results=None):
```
porridge, line 79: This may be a bit confusing, since it does process `test_outputs` but not `test_results`. I can think of two ways:
- barf if `test_results is not None` (better fail loud than silent), or
- move the functionality to this class

.openshift-ci/post_tests.py
```diff
@@ -204,6 +209,16 @@ def check_stackrox_logs(self):
             timeout=PostTestsConstants.CHECK_TIMEOUT,
         )
 
+    def add_test_result(self, test_results):
+        print("Storing operator test results in JUnit format")
```
porridge, line 213: This is in general not operator specific.
```suggestion
        print("Storing test results in JUnit format")
```

.openshift-ci/post_tests.py
```diff
@@ -204,6 +209,16 @@ def check_stackrox_logs(self):
...
+        for fromDir, toDir in test_results:
+            self.run_with_best_effort(
+                ["scripts/ci/store-artifacts.sh", "store_test_results",
+                 fromDir, toDir],
```
porridge, line 217: The rest of this file seems to avoid camelCase.
```suggestion
        for from_dir, to_dir in test_results:
            self.run_with_best_effort(
                ["scripts/ci/store-artifacts.sh", "store_test_results",
                 from_dir, to_dir],
```

.openshift-ci/post_tests.py
```diff
@@ -219,7 +234,7 @@ def __init__(
         self._check_for_errors_in_stackrox_logs = check_for_errors_in_stackrox_logs
         self.central_is_responsive = False
 
-    def run(self, test_outputs=None):
+    def run(self, test_outputs=None, test_results=None):
```
porridge, line 237: Same comment as with the `StoreArtifacts` implementation.

.openshift-ci/post_tests.py
```diff
@@ -287,7 +302,7 @@ def __init__(
             self.data_to_store.append(PostTestsConstants.QA_SPOCK_RESULTS)
         self._handle_e2e_progress_failures = handle_e2e_progress_failures
 
-    def run(self, test_outputs=None):
+    def run(self, test_outputs=None, test_results=None):
```
porridge, line 305: Ditto

porridge (commented):
LGTM overall, except it seems like this might better fit in the parent class...

porridge:
And let's see whether this actually works in the operator jobs :-)

porridge:
Unfortunately:
```
****
**** 09:54:36: final post completed
****
Traceback (most recent call last):
  File "/go/src/github.com/stackrox/stackrox/scripts/ci/jobs/openshift_newest_operator_e2e_tests.py", line 16, in <module>
    final_post=FinalPost(handle_e2e_progress_failures=False),
  File "/go/src/github.com/stackrox/stackrox/.openshift-ci/runners.py", line 83, in run
    raise hold
  File "/go/src/github.com/stackrox/stackrox/.openshift-ci/runners.py", line 57, in run
    self.run_test_set(test_set)
  File "/go/src/github.com/stackrox/stackrox/.openshift-ci/runners.py", line 113, in run_test_set
    raise hold
  File "/go/src/github.com/stackrox/stackrox/.openshift-ci/runners.py", line 104, in run_test_set
    test_set["post_test"].run(test_outputs=test_set["test"].test_outputs,
AttributeError: 'OperatorE2eTest' object has no attribute 'test_outputs'
```

.openshift-ci/post_tests.py
```diff
@@ -68,17 +68,29 @@ class StoreArtifacts(RunWithBestEffortMixin):
...
         self.data_to_store = []
 
-    def run(self, test_outputs=None):
+    def run(self, test_outputs=None, test_results=None):
         self.store_artifacts(test_outputs)
+        self.add_test_results(test_results)
         self.handle_run_failure()
 
+    def add_test_results(self, test_results):
+        if test_results is None:
```
porridge, line 84: It's generally considered more pythonic to say:
```suggestion
        if not test_results:
```
This catches the empty list case too (and I guess it makes sense because we don't want to log the following line if we actually have nothing to store).

.openshift-ci/post_tests.py
```diff
@@ -127,6 +139,7 @@ def run(self, test_outputs=None):
         self.collect_service_logs()
         if self._check_stackrox_logs:
             self.check_stackrox_logs()
+        self.add_test_results(test_results)
         self.store_artifacts(test_outputs)
```
porridge, line 143: To keep the order consistent with `StoreArtifacts.run`:
```suggestion
        self.store_artifacts(test_outputs)
        self.add_test_results(test_results)
```

porridge (approved):
NICE!
Just a couple of nitpicks.

gavin-stackrox (approved):
LGTM! 

I'd prefer to avoid the reformatting but I left the door open for that by not adding the formatter to CI: https://github.com/stackrox/stackrox/blob/master/.openshift-ci/README.md#workflow-aliases-for-testformatlint

It would be nice to have an example of a failed kuttl test result in the PR description as proof of work.

porridge:
> It would be nice to have an example of a failed kuttl test result in the PR description as proof of work.

Let's see how https://github.com/stackrox/stackrox/pull/3362 goes

porridge:
[Failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3362/pull-ci-stackrox-stackrox-master-openshift-oldest-operator-e2e-tests/1579337904126169088) [cases](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3362/pull-ci-stackrox-stackrox-master-openshift-newest-operator-e2e-tests/1579337903287308288).

ivan-degtiarenko:
@porridge thanks for testing 👍  Merging the PR

PR #2566, opened 2022-08-01
Split Violations by category chart and card components.

dvail:
## Description

This is a purely mechanical cleanup change that separates the outer part of the `ViolationsByPolicyCategory` component (widget card, options, data loading) from the inner part of the card (chart display). This was the first widget developed and the only one where both of the above components were contained in a single file, so this PR breaks them apart to be in line with the other widgets and keep the code consistent.

## Checklist

## Testing Performed

Manual testing of widget on the dashboard, local run of existing unit tests.

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategoryChart.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategoryChart.tsx
@@ -73,8 +73,6 @@ function sortBySeverity(groups: AlertGroup[]) {
     ]);
 }
 
-export type LifecycleOption = 'ALL' | Exclude<LifecycleStage, 'BUILD'>;
-
 type CountsBySeverity = Record<PolicySeverity, Record<string, number>>;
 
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategoryChart.tsx
```diff
@@ -0,0 +1,261 @@
...
+}
+
+type ViolationsByPolicyCategoryChartProps = {
+    alertGroups: AlertGroup[];
+    sortType: SortTypeOption;
+    lifecycle: LifecycleOption;
+    searchFilter: SearchFilter;
+    hiddenSeverities: Set<PolicySeverity>;
+    setHiddenSeverities: (severities: Set<PolicySeverity>) => Promise<Config>;
+};
```
pedrottimark, line 134: Can you move the Props type to immediately precede the component function?

On the other side of the coin, it moves the following helper function and module scope variable to be together with the others.

pedrottimark (approved):
Super separation to decrease complexity of component files and increase consistency of Widgets folder. One suggestion.

PR #2529, opened 2022-07-27
ROX-11986: Fix minimum access role updates

dhaus67:
## Description

When trying to update the minimum access role within the UI, the updated value of the minimum access role is not actually reflected within the request:
Screenshot 2022-07-27 at 02 03 32

Instead, the same value is being sent.

Additionally, when trying to update an existing default group via the `/v1/groupsbatch` API, the update will fail with the following error message:
```json
{
    "error": "a default group already exists for auth provider \"3de3432f-5e5b-4903-bbde-82bcc232b239\"",
    "code": 6,
    "message": "a default group already exists for auth provider \"3de3432f-5e5b-4903-bbde-82bcc232b239\"",
    "details": []
}
```

This is due to the fact that we only check for the existence of a default group in the update case. Instead, we should allow updates to the existing group.

This PR will introduce the following:
- Fixing the role name sent in the `groupsbatch` request from the UI.
- Adding E2E tests for changing the minimum access role.
- Allow updates to existing default groups.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed
- see Unit tests.
- E2E test for UI changes.
- Manual tests:
1. Create an auth provider of your choice.
2. Update the minimum access role.
3. Ensure the minimum access role is correctly set within UI + API requests.

PR #2561, opened 2022-07-30
Refactor bolt store for groups and move logic to datastore in preparation for postgres move

theencee:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #2569, opened 2022-08-01
Fix update default group by props

dhaus67:
## Description

Based on the comment left [here](https://github.com/stackrox/stackrox/pull/2529#discussion_r933699601).

If the ID is not set initially due to fetching groups via property, we should only check the default group after we are sure the ID is set.

## Testing Performed

- added Unit tests.


dhaus67:
* master
  * **PR #2569** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2569?utm_source=stack-comment).

dhaus67:
Closed since this will be addressed within #2561 and rebase wouldn't be worth the effort.

PR #2571, opened 2022-08-01
ROX-11320: Adding create policy category modal

alwayshooin:
## Description

Adding the policy create category modal to the policy categories page. 
Should look like the below and disable interacting with default categories. 
When creating a new category, if successful, should auto select the newly created category.

image
image
image

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
```diff
@@ -19,15 +19,15 @@ function PolicyCategoriesList({
...
             {policyCategories.map((category) => {
-                const { id, name } = category;
+                const { id, name, isDefault } = category;
                 return (
                     <SimpleListItem
                         key={id}
                         onClick={() => {
                             setSelectedCategory(category);
                         }}
                         isActive={false}
```
pedrottimark, line 29: Will `isActive` prop have a conditional value in a future pull request?

pedrottimark:
Off the topic of this pull request, while testing `isDisabledProp` of **Create policy** button:

1. Result of **Save** existing policy category with new name does not update the list. Because backend can adjust the user name according to its naming convention, the update needs to use the response instead of the form values, correct? Maybe also update the form values?
 
https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/PolicyCategories/PolicyCategorySidePanel.tsx#L33-L36

    ```js
            renamePolicyCategory(id, name)
                .then((response) => {
                    setSelectedCategory(response);
                    // TODO also update values?
                })
    ```

2. Result of **Create** new policy category needs to use response:
    * which has `id` prop to correctly support a subsequent update?
    * also because of possible backend user name adjustment?

```js
           postPolicyCategory(values)
                .then((response) => {
                    setSelectedCategory(response); // here instead of in finally block? reset the form values too?
                    setTimeout(refreshPolicyCategories, 200);
                    addToast('Successfully added category');
                })
```

pedrottimark (approved):
Keep up the steady steps forward. 2 questions which might affect future work.

ui/apps/platform/src/Containers/PolicyCategories/CreatePolicyCategoryModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    resetForm();
+                    onClose();
+                });
+        },
+        validateOnMount: true,
+        validationSchema: yup.object().shape({
+            name: yup
+                .string()
+                .min(5, 'Policy category name must be at least 5 characters long')
+                .max(128, 'Policy category name must be less than 128 characters long')
```
vjwilson, line 63: Is this max-length a backend requirement?
(128 seems like a random choice. I know relationship SQL, like Postgres, used to be optimized for a short string of up to 255 characters--haven't seen 128 before.)

vjwilson (approved):

ui/apps/platform/src/Containers/PolicyCategories/CreatePolicyCategoryModal.tsx
pedrottimark: Yes, backend requirement:

https://github.com/stackrox/stackrox/blob/master/central/policycategory/service/service_impl.go#L40

ui/apps/platform/src/Containers/PolicyCategories/PolicyCategoriesList.tsx
alwayshooin: oh no it's residual from styling i thought i had to do. good eyes, i'll remove it

PR #2572, opened 2022-08-01
ROX-11657: Scope autogenerated image integrations by the image pull secrets

connorgorman:
## Description

This PR adds a new message `Source` to the `ImageIntegration` proto with the goal to scope autogenerated integrations to a specific cluster, namespace, and pull secrets.

For this to work, a couple of changes were required:
- Proto changes to the `ImageIntegration` proto, adding the `Source` message.
- API changes to the `ScanImageInternalRequest`, allowing sending the `cluster_id, namespace, and pullsecrets` that should be used for scanning.
- A new `ServiceAccountStore/Dispatcher` within sensor, allowing to manage image pull secrets associated with a service account for a specific namespace.
- Changes in the enricher within sensor to send associated image pull secrets within the `ScanImageInternalRequest`.
- Changes in the image enricher within central to filter registries based on the attached image pull secrets within the enrichment context for internal images.
- Implementing a reconciliation for internal image integrations, ensuring changes from secured clusters are correctly reflected in central.

## Checklist

## Testing Performed

- see CI.


sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -133,16 +135,29 @@ func newSecretDispatcher(regStore *registry.Store) *secretDispatcher {
 	}
 }
 
+func derivedID(secret *v1.Secret, registry string) (string, error) {
+	rootUUID, err := uuid.FromString(string(secret.UID))
+	if err != nil {
+		return "", errors.Wrapf(err, "converting %q to uuid", secret.UID)
```
dhaus67, line 141:
```suggestion
		return "", errors.Wrapf(err, "converting secret ID %q to uuid", secret.UID)
```

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -133,16 +135,29 @@ func newSecretDispatcher(regStore *registry.Store) *secretDispatcher {
 	}
 }
 
+func derivedID(secret *v1.Secret, registry string) (string, error) {
```
dhaus67, line 138:
```suggestion
func deriveIDFromSecret(secret *v1.Secret, registry string) (string, error) {
```

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -183,6 +203,35 @@ func (s *secretDispatcher) processDockerConfigEvent(secret *v1.Secret, action ce
 		_ = utils.Should(errors.New("only Docker Config secrets are allowed"))
 		return nil
 	}
+	return &dockerConfig
+}
+
+func idSetFromSecret(secret *v1.Secret) (set.StringSet, error) {
```
dhaus67, line 209:
```suggestion
func imageIntegationIDSetFromSecret(secret *v1.Secret) (set.StringSet, error) {
```

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -183,6 +203,35 @@ func (s *secretDispatcher) processDockerConfigEvent(secret *v1.Secret, action ce
...
+	if secret == nil {
+		return nil, nil
+	}
+	dockerConfigPtr := getDockerConfigFromSecret(secret)
+	if dockerConfigPtr == nil {
+		return nil, nil
+	}
+	dockerConfig := *dockerConfigPtr
+	registrySet := set.NewStringSet()
```
dhaus67, line 218: Should be image integration.
```suggestion
	imageIntegrationIDSet := set.NewStringSet()
```

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -199,21 +250,43 @@ func (s *secretDispatcher) processDockerConfigEvent(secret *v1.Secret, action ce
...
-			sensorEvents = append(sensorEvents, &central.SensorEvent{
-				// Only update is supported at this time.
-				Action: central.ResourceAction_UPDATE_RESOURCE,
-				Resource: &central.SensorEvent_ImageIntegration{
-					ImageIntegration: ii,
-				},
-			})
+			ii, err := DockerConfigToImageIntegration(secret, registry, dce)
+			if err != nil {
+				log.Errorf("unable to create docker config for secret %s %v", secret.GetName(), err)
```
dhaus67, line 255: super nit:
```suggestion
				log.Errorf("unable to create docker config for secret %s: %v", secret.GetName(), err)
```

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -199,21 +250,43 @@ func (s *secretDispatcher) processDockerConfigEvent(secret *v1.Secret, action ce
...
 
 		registries = append(registries, &storage.ImagePullSecret_Registry{
 			Name:     registry,
 			Username: dce.Username,
 		})
 	}
+	// Compute diff between old and new secret to automatically clean up delete secrets
+	oldIntegrations, err := idSetFromSecret(oldSecret)
+	if err != nil {
+		log.Errorf("error getting ids from old secret: %v", err)
```
dhaus67, line 276:
```suggestion
		log.Errorf("error getting ids from old secret %q: %v", string(oldSecret.UID), err)
```

sensor/kubernetes/listener/resources/singleton.go
```diff
@@ -6,6 +6,9 @@ var (
 	dsInit   sync.Once
 	depStore *DeploymentStore
 
+	sasInit  sync.Once
+	sasStore *ServiceAccountStore
```
dhaus67, line 10: nit:
```suggestion
	saStore *ServiceAccountStore
```

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -54,14 +60,42 @@ func NewPipeline(integrationManager enrichment.Manager,
...
+	if !conn.HasCapability(centralsensor.ScopedImageIntegrations) {
+		return nil
+	}
+
+	integrations, err := s.datastore.GetImageIntegrations(ctx, &v1.GetImageIntegrationsRequest{})
+	if err != nil {
+		return errors.Wrap(err, "getting image integrations for reconciliation")
+	}
+	for _, integration := range integrations {
+		if integration.GetEcr() != nil {
```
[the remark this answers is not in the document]
connorgorman: The ECR autogenerated integrations are special and use AWS IAM

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -54,14 +60,42 @@ func NewPipeline(integrationManager enrichment.Manager,
...
+	if conn == nil {
+		return nil
+	}
+	// We should not reconcile image integrations unless the Sensor is on a version that can provide scoped integrations
+	// with consistent uuids
+	if !conn.HasCapability(centralsensor.ScopedImageIntegrations) {
+		return nil
+	}
+
+	integrations, err := s.datastore.GetImageIntegrations(ctx, &v1.GetImageIntegrationsRequest{})
```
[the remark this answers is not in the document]
connorgorman: Well, so long story, this doesn't work. I'll fix it in a follow up

connorgorman:
@dhaus67 I apologize for taking forever to clean this up, but I have made the requested changes and also fixed some issues. Fingers crossed on the e2es

proto/storage/image_integration.proto
```diff
@@ -35,6 +35,13 @@ message ImageIntegration {
     bool   skip_test_integration = 18;
 
     reserved 19; // Previously, scannerv2
+
+    message Source {
+        string cluster_id = 1;
```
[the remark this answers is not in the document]
connorgorman: I would leave it for now due to changing all the usages

connorgorman:
@dhaus67 This is good to go. The TestScan is a test that was written for summaries, but summaries are checked in qa-backend-e2e

dhaus67 (approved):

proto/storage/image_integration.proto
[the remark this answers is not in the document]
dhaus67: Sounds good, I've created [ROX-12996](https://issues.redhat.com/browse/ROX-12996) as a follow-up to remove exactly this.

dhaus67 (approved):

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -54,14 +61,42 @@ func NewPipeline(integrationManager enrichment.Manager,
...
+	if err != nil {
+		return errors.Wrap(err, "getting image integrations for reconciliation")
+	}
+	for _, integration := range integrations {
+		if integration.GetEcr() != nil {
+			continue
+		}
+		if integration.GetClusterId() != clusterID {
```
misberner, line 91: I looked at `v1.GetImageIntegrationsRequest` and it seems that it supports specifying a cluster (name only, not ID?). However, the handling of that field is extremely weird (`central/imageintegration/datastore/datastore_impl.go:55-64):
```
	integrations, err := ds.storage.GetAll(ctx)
	if err != nil {
		return nil, err
	}

	integrationSlice := integrations[:0]
	for _, integration := range integrations {
		if request.GetCluster() != "" {
			continue
		}
```
So if `request` specifies a non-empty cluster, it will always return an empty slice, while still doing a read on the DB. Not sure if there's a bug or this param was just never supported, but either way, this probably deserves a JIRA filed for it and addressed some time.

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -226,4 +234,64 @@ func (s *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+
+	if _, err := s.datastore.AddImageIntegration(ctx, imageIntegration); err != nil {
+		return errors.Wrapf(err, "adding image integration %q", imageIntegration.GetId())
+	}
+	if err := s.integrationManager.Upsert(imageIntegration); err != nil {
+		return errors.Wrapf(err, "notifying of update for image integration %q", imageIntegration.GetId())
+	}
+	if !exists {
+		s.enrichAndDetectLoop.ShortCircuit()
+	}
```
misberner, line 293: Tbh, I find it a bit ugly that the pipeline is in charge of all of this. IMHO, the manager should both abstract away the store insertion logic as well as the need to run detection. However, that may be hard to refactor. Still, let's factor this out into its own method, at least?

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -226,4 +234,64 @@ func (s *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+		return errors.Wrapf(err, "retrieving image integration %q", imageIntegration.GetId())
+	}
+
+	if _, err := s.datastore.AddImageIntegration(ctx, imageIntegration); err != nil {
+		return errors.Wrapf(err, "adding image integration %q", imageIntegration.GetId())
+	}
+	if err := s.integrationManager.Upsert(imageIntegration); err != nil {
+		return errors.Wrapf(err, "notifying of update for image integration %q", imageIntegration.GetId())
+	}
+	if !exists {
```
misberner, line 291: This doesn't make too much sense imho. A modified integration might just as well be a reason to run detection again. It would be nice if we could somehow scope the re-evaluation to images potentially using that integration, but that's far, far beyond the scope of this PR.

I wonder if the concern is that doing this on every update might be too frequent. I don't think however that updates of image pull secrets are very different in frequency from creations - most are probably never updated. A concern is the initial cluster connection. when we receive many new integrations and would start the detection loop several times. To avoid this, it might make sense to add a `hasSynced concurrency.Flag`, that is set to `true` on `Reconcile` followed by triggering detection, and then trigger detection here on every update whenever `s.hasSynced.Get()` is `true`.

pkg/images/enricher/enricher_impl.go
```diff
@@ -649,17 +655,38 @@ func (e *enricherImpl) checkRegistryForImage(image *storage.Image) error {
 	return nil
 }
 
-func (e *enricherImpl) getRegistriesForContext(ctx EnrichmentContext) (registries.Set, error) {
-	registrySet := e.integrations.RegistrySet()
+func (e *enricherImpl) getRegistriesForContext(ctx EnrichmentContext) ([]registryTypes.ImageRegistry, error) {
+	registries := e.integrations.RegistrySet().GetAll()
 	if ctx.Internal {
-		return registrySet, nil
+		if ctx.Source == nil {
```
misberner, line 661: The `then`-block might be worth to factor out into its own function, as the readability is suffering from the indentation level

proto/api/v1/image_service.proto
```diff
@@ -34,9 +34,12 @@ message ScanImageRequest {
...
 message ScanImageInternalRequest {
-    storage.ContainerImage image = 1;
-    reserved                       2;
-    bool cached_only             = 3;
+    storage.ContainerImage image              = 1;
+    reserved 2;
+    bool                   cached_only        = 3;
+    string                 cluster_id         = 4;
+    string                 namespace          = 5;
+    repeated string        image_pull_secrets = 6;
```
misberner, line 42: Move this to a sub-message `scope` or `source` or so? Tbh I found it very confusing that in `service_impl.go` we check for `GetClusterId() != ""` and then assume that `GetNamespace()` and `GetImagePullSecrets()` can be used, too. This would be much clearer if we used `if scope := request.GetScope(); scope != nil` and then just `scope.GetClusterId()` etc. This way, we wouldn't have to abuse `cluster_id` as a sentinel in the logic, as it is not more special than the other fields.

proto/storage/image_integration.proto
```diff
@@ -35,6 +35,13 @@ message ImageIntegration {
...
+        string namespace  = 2;
+        string name       = 3;
```
misberner, line 42: Call this `secret_name` or `image_pull_secret_name`? It's a bit unclear otherwise

misberner (changes requested):

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -54,14 +61,42 @@ func NewPipeline(integrationManager enrichment.Manager,
...
 	enrichAndDetectLoop reprocessor.Loop
 }
 
-func (s *pipelineImpl) Reconcile(_ context.Context, _ string, _ *reconciliation.StoreMap) error {
-	// Nothing to reconcile for image integrations
-	return nil
+func (s *pipelineImpl) Reconcile(ctx context.Context, clusterID string, storeMap *reconciliation.StoreMap) error {
+	existingIDs := set.NewStringSet()
+
+	conn := s.connManager.GetConnection(clusterID)
```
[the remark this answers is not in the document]
dhaus67: Seems like we _can_ skip it, the `clusterhealthupdate` does it in a similar fashion.

central/sensor/service/pipeline/imageintegrations/pipeline.go
dhaus67: So, seems like the original PR #658 was actually removing support for this parameter (but apparently, this wasn't done because there was still _something_ left for autogenerated image integrations that uses this parameter).

I do agree, it does justify a JIRA, I created one: [ROX-13071](https://issues.redhat.com/browse/ROX-13071)

central/sensor/service/pipeline/imageintegrations/pipeline.go
dhaus67: Hm, not sure I completely get it from your comment, sorry.
Your suggestion is to set the flag within `Reconcile()` and then only short-circuit if the flag is set, right?

Regarding the frequencies, I saw [this comment indicating a "higher" volume of updates due to service accounts](https://github.com/stackrox/stackrox/blob/d14d6e7b1a9a0e4b7c5b0e111cc0c6d330462ffb/central/sensor/service/pipeline/imageintegrations/pipeline.go#L210-L214), but to be fair I'm not quite sure whether this is of any concern.

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -139,11 +170,11 @@ func setUpIntegrationParams(ctx context.Context, imageIntegration *storage.Image
...
-		if err != nil {
-			err = errors.Wrapf(err, "reaching out for TLS check to %s", dockerCfg.GetEndpoint())
-			return
+		validTLS, tlsErr := tlscheck.CheckTLS(ctx, dockerCfg.GetEndpoint())
+		if tlsErr != nil {
+			log.Debugf("reaching out for TLS check to %s: %v", dockerCfg.GetEndpoint(), err)
+			// Assume TLS is true
+			validTLS = true
 		}
 		dockerCfg.Insecure = !validTLS
```
misberner, line 179:
```suggestion
		if tlsErr != nil {
			log.Debugf("reaching out for TLS check to %s: %v", dockerCfg.GetEndpoint(), err)
			// Not enough evidence that we can skip TLS, so conservatively require it
			dockerCfg.Insecure = false
		} else {
			dockerCfg.Insecure = !validTLS
		}
```
Don't overload variables if the name doesn't fit the usage any more. `validTLS` is a misnomer in l.177.

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -226,4 +230,75 @@ func (s *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+	if requiresShortCircuit {
+		s.enrichAndDetectLoop.ShortCircuit()
+	}
+
+	return nil
+}
+
 func (s *pipelineImpl) OnFinish(_ string) {}
+
+func (s *pipelineImpl) upsertImageIntegration(ctx context.Context, imageIntegration *storage.ImageIntegration) (bool, error) {
```
misberner, line 291: Use a named return value for the `bool` return value, or at least add a comment

central/sensor/service/pipeline/imageintegrations/pipeline.go
```diff
@@ -226,4 +230,75 @@ func (s *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+		return exists, errors.Wrapf(err, "retrieving image integration %q", imageIntegration.GetId())
+	}
+
+	if _, err := s.datastore.AddImageIntegration(ctx, imageIntegration); err != nil {
+		return exists, errors.Wrapf(err, "adding image integration %q", imageIntegration.GetId())
+	}
+	if err := s.integrationManager.Upsert(imageIntegration); err != nil {
+		return exists, errors.Wrapf(err, "notifying of update for image integration %q", imageIntegration.GetId())
+	}
+	return exists, nil
```
misberner, line 303: Didn't you now flip when to do a short-circuit? Why would we _not_ want to short circuit upon creation of an image integration?

central/sensor/service/pipeline/imageintegrations/pipeline.go
misberner: Let me try to rephrase:
- Ideally, if an image integration is updated, we would want to reprocess all deployments with previously unscannable/unretrievable images that potentially use this registry integration.
- As an approximation of the above, we could reprocess all deployments that potentially use this registry integration, unscannable images or not.
- As an approximation of the above, we could reprocess all deployments that are somewhat related to this registry integration (e.g., same cluster), even if they don't use it, and regardless of whether they have unscannable images.

The point is that an updated registry integration can only affect a deployment evaluation if we can now scan an image used by the deployment, that we previously could not scan (due to the content-addressing nature of IDs, "being able to scan an image" is a binary property that we can assume never reverts to `false` once it became `true). Otherwise, there is no information that can be gained from the updated image integration and that could change anything about the evaluation.

The existing code however only reprocesses a deployment when a new image integration is created. This is neither an over- nor under-approximation of the "ideal" handling stated at the top of this post. It **unnecessarily** reprocesses all deployments (even those not using the updated image integration nor using an image from one of the registries this integration covers, it might even reprocess deployments in other clusters, and it also reprocesses deployment for which we already have comprehensive scan results for all images). At the same time, it **misses** cases when a previously existing, erroneous image integration is fixed, and there actually _are_ deployments with still unscanned images due to the previously erroneous image integration. Those will only be fixed after ~1h, whereas those with a _missing_ image integration are fixed directly on creation.

My idea was therefore to just run reprocessing on every image integration update, in order to arrive at something that at least is a consistent over-approximation. Still reprocessing far too much, but at least not missing anything; a fully-overblown but sufficient solution rather than a half-overblown yet still insufficient one.

My suspicion was that the case this approach wasn't chosen was due to the fact that during reconciliation (i.e., a sensor reconnects), we would receive a flood of image integration updates, causing an avalanche of reprocessings. However: (a) this is not how `ShortCircuit()` works anyway, it just triggers a signal, but if the enricher/detector is still busy, a subseqent call will just do nothing (note: this alone makes my comment on setting a flag and calling `ShortCircuit` once after reconciliation pointless, so just ignore it), and (b) the comment you pointed out suggests that this is due to a high volume of (spurious?) updates on OpenShift rather than the reconciliation phase.

So, in summary, keep the existing behavior (which I think now has accidentally been flipped), until we either can do reprocessing in a more targeted way (unscanned images only?), or understand why image integration updates on OpenShift are so noisy.

misberner (changes requested):
Mostly looks good, except for what I think is an accidental flip in the behavior wrt when to reprocess

central/sensor/service/pipeline/imageintegrations/pipeline.go
dhaus67: Oh well, yup.

misberner (approved):

dhaus67:
Merging since the CI failures are fixed in master within #3415

PR #2576, opened 2022-08-01
ROX-9770: adding operating system to the vulnerability table columns

sachaudh:
## Description

This PR adds the operating system information to the table columns for the following tables:
1. Node CVEs
2. Image CVEs
3. Node Components
4. Image Components

The Nodes and Images tables have the operating system column already

This also adds the operating system information to the singles for each of those entity tables too

## Checklist

### Lists
Screen Shot 2022-08-02 at 4 21 45 PM
Screen Shot 2022-08-02 at 4 21 52 PM
Screen Shot 2022-08-02 at 4 22 03 PM
Screen Shot 2022-08-02 at 4 22 06 PM

### Singles
Screen Shot 2022-08-02 at 5 05 39 PM
Screen Shot 2022-08-02 at 5 05 51 PM
Screen Shot 2022-08-02 at 5 06 03 PM
Screen Shot 2022-08-02 at 5 06 09 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
@@ -247,7 +247,7 @@ export function getCveTableColumns(workflowState) {
     ];
 
     if (currentEntityType === entityTypes.NODE_CVE || currentEntityType === entityTypes.IMAGE_CVE) {
-        tableColumns.push({
+        tableColumns.splice(3, 0, {
             Header: `Operating System`,
             headerClassName: `w-1/10 ${defaultHeaderClassName}`,
             className: `w-1/10 ${defaultColumnClassName}`,
--- a/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
@@ -246,6 +246,17 @@ export function getCveTableColumns(workflowState) {
         },
     ];
 
+    if (currentEntityType === entityTypes.NODE_CVE || currentEntityType === entityTypes.IMAGE_CVE) {
+        tableColumns.push({
+            Header: `Operating System`,
+            headerClassName: `w-1/10 ${defaultHeaderClassName}`,
+            className: `w-1/10 ${defaultColumnClassName}`,
+            id: cveSortFields.OPERATING_SYSTEM,
+            accessor: 'operatingSystem',
+            sortField: cveSortFields.OPERATING_SYSTEM,
+        });
+    }
+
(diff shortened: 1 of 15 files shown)


sachaudh:
@md2119 @keyallis @charmik-redhat From the screenshots you can see that the Image Component's table was the only one that showed the operating system value. Seems like the other ones don't. Any idea why?

md2119:
Operating System is empty because this newly added attribute is supported only on Postgres.

sachaudh:
> * Add operating system column on the node components page as well.

I didn't see the `operatingSystem` field available for a `NodeComponent`. Do you see that? I added it for the ones that did

Screen Shot 2022-08-02 at 11 55 57 AM

sachaudh:
> 2\. Would be nice to make the operating system column second column on cves images. Vulns with the same cve but different distro becomes more obvious with such arrangment.

Sure we can re-arrange it

sachaudh:
> Operating System is empty for cves because this newly added attribute is supported only on Postgres.

Oh, I see. I'll use that Postgres feature flag then

sachaudh:
Quick question: Would we be enabling the frontend-only VM updates feature-flag while keeping the Postgres feature flag disabled? If so, I'll add conditionals to render the operating system stuff when Postgres is enabled. If we'll enable them both at the same time, then I guess it would be fine to keep things how they are since these pages don't show up unless the frontend-only VM updates feature-flag is enabled

md2119:
> Quick question: Would we be enabling the frontend-only VM updates feature-flag while keeping the Postgres feature flag disabled? If so, I'll add conditionals to render the operating system stuff when Postgres is enabled. If we'll enable them both at the same time, then I guess it would be fine to keep things how they are since these pages don't show up unless the frontend-only VM updates feature-flag is enabled

The UI feature flag will be turned on after Postgres feature flag is turned on.

> If we'll enable them both at the same time, then I guess it would be fine to keep things how they are since these pages don't show up unless the frontend-only VM updates feature-flag is enabled

Not sure I follow you. What do you mean by keeping things as is?

sachaudh:
> Not sure I follow you. What do you mean by keeping things as is?

Meaning I won't add any additional feature flagging using the postgres feature flag since the changes in the PR won't show up unless our frontend-only feature flag is enabled

ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
```diff
@@ -246,6 +246,17 @@ export function getCveTableColumns(workflowState) {
         },
     ];
 
+    if (currentEntityType === entityTypes.NODE_CVE || currentEntityType === entityTypes.IMAGE_CVE) {
+        tableColumns.splice(3, 0, {
```
vjwilson, line 250: I'd recommend including this column within the original list with a ternary : null, and let it get filtered out if not one of those two types, but still will work, since it is a short-lived solution.

vjwilson (approved):

PR #2589, opened 2022-08-02
ROX-11894: Skip unauthorized error messages when fetching image signatures

dhaus67:
## Description

When an issue occurs during authentication with a registry, i.e. the wrong credentials are used, the error returned by the call `FetchSignaturesForReference` will be of the type `registry.HttpStatusCode`.

Previously, we only expected either a `transport.Error` or a `error.Wrap` which contained a `url.Error` and subsequently a `registry.HttpStatusCode` error.

This led to additional logs being printed within central logs when credentials were used that didn't fit for a registry. Those logs are false, as they are not indicating an actual error within the signature fetching, as it will 
not stop the enriching of the image.

Now, instead of casting the errors, we use `errors.As`. This way, we can skip unwrapping errors manually, which was the issue beforehand of not casting the plain `registry.HttpStatusCode` error.

## Testing Performed
- see added unit tests.
- Manual tests:
1. Use a any registry of your liking (quay.io/docker.io/demo.harbor.io).
2. Create a private repository, if you do not already have one.
3. Create an image integration for the registry but with wrong credentials.
4. Deploy a pod, observe that no unauthorized log messages are printed.

### DIFF
--- a/pkg/signatures/public_key_fetcher.go
+++ b/pkg/signatures/public_key_fetcher.go
@@ -20,6 +20,7 @@ import (
 	"github.com/stackrox/rox/pkg/errox"
 	registryTypes "github.com/stackrox/rox/pkg/registries/types"
 	"github.com/stackrox/rox/pkg/retry"
+	"github.com/stackrox/rox/pkg/sliceutils"
 	"golang.org/x/time/rate"
 )
 
@@ -124,15 +125,17 @@ func (c *cosignPublicKeySignatureFetcher) FetchSignatures(ctx context.Context, i
 // Note: This takes into account the definition of the transport.Error, you can find more here:
 // https://github.com/google/go-containerregistry/blob/f1fa40b162a1601a863364e8a2f63bbb9e4ff36e/pkg/v1/remote/transport/error.go#L90
 func makeTransientErrorRetryable(err error) error {
+	var transportErr *transport.Error
+	var urlError *url.Error
 	// We don't expect any transient errors that are coming from cosign at the moment.
-	if transportErr, ok := err.(*transport.Error); ok && transportErr.Temporary() {
-		return retry.MakeRetryable(err)
+	if errors.As(err, &transportErr) && transportErr.Temporary() {
+		return retry.MakeRetryable(transportErr)
 	}
 
-	// We need to check for url.Error as well, since heroku-client will return those.
-	if urlErr := errToURLError(err); urlErr != nil && urlErr.Temporary() {
-		return retry.MakeRetryable(err)
+	if errors.As(err, &urlError) && urlError.Temporary() {
+		return retry.MakeRetryable(urlError)
 	}
+
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #2589** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2589" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2589?utm_source=stack-comment).

pkg/signatures/public_key_fetcher_test.go
```diff
@@ -244,11 +246,21 @@ func TestIsMissingSignatureError(t *testing.T) {
...
+			err: &emptyResponseErr,
+		},
+		"status error with status code unauthorized should not indicate missing signature": {
+			err: &unauthorizedErr,
+		},
 		"wrapped registry error with status code not found should indicate missing signature": {
 			err:              fmt.Errorf("something went wrong %w", &url.Error{Err: &notFoundErr}),
 			missingSignature: true,
 		},
-		"wraooed registry error with status code unauthorized should not indicate missing signature": {
```
janisz, line 251: ❤️

janisz (approved):

pkg/signatures/public_key_fetcher.go
```diff
@@ -124,15 +125,17 @@ func (c *cosignPublicKeySignatureFetcher) FetchSignatures(ctx context.Context, i
 // Note: This takes into account the definition of the transport.Error, you can find more here:
 // https://github.com/google/go-containerregistry/blob/f1fa40b162a1601a863364e8a2f63bbb9e4ff36e/pkg/v1/remote/transport/error.go#L90
 func makeTransientErrorRetryable(err error) error {
+	var transportErr *transport.Error
+	var urlError *url.Error
 	// We don't expect any transient errors that are coming from cosign at the moment.
-	if transportErr, ok := err.(*transport.Error); ok && transportErr.Temporary() {
-		return retry.MakeRetryable(err)
+	if errors.As(err, &transportErr) && transportErr.Temporary() {
+		return retry.MakeRetryable(transportErr)
```
misberner, line 132: Why do you return the unwrapped error here, instead of `err`?

pkg/signatures/public_key_fetcher.go
```diff
@@ -124,15 +125,17 @@ func (c *cosignPublicKeySignatureFetcher) FetchSignatures(ctx context.Context, i
...
 	}
 
-	// We need to check for url.Error as well, since heroku-client will return those.
-	if urlErr := errToURLError(err); urlErr != nil && urlErr.Temporary() {
-		return retry.MakeRetryable(err)
+	if errors.As(err, &urlError) && urlError.Temporary() {
+		return retry.MakeRetryable(urlError)
```
misberner, line 136: same here

misberner (approved):
LGTM, question on whether unwrapping is intentional

pkg/signatures/public_key_fetcher.go
dhaus67: Ah, good catch! You're right, we shouldn't return the unwrapped error here.

PR #2590, opened 2022-08-02
Update CVEs-By-CVSS-Score widget to use new GraphQL queries

vjwilson:
## Description

Fixes this bug from the bug list:

> CVES BY CVSS SCORE still points to old graphQL endpoint everywher

## Checklist
- [x] Investigated and inspected CI test results (failures are unrelated flakes)

## Testing Performed

Deployment scope => imageVulnerabilities
deployment_cves_by_cvss

Image scope => imageVulnerabilities
image_cves_by_cvss

Image Component scope => imageVulnerabilities 
**Note: UI is now using the correct query, but the API is not scoping the returned list correctly for an image component**
image_component_cves_by_cvss

Node Component scope => nodeVulnerabilities
**Note: UI is now using the correct query, but the API is not scoping the returned list correctly for a node component**
node_component_cves_by_cvss

Node scope => nodeVulnerabilities
node_cves_by_cvss

**Note: I allowed for using `clusterVulnerabilities` when this widget is used on a Cluster Overview page, but we don't currently use it on that type of page, so there is no example to show.**

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/widgets/CvesByCvssScore.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/widgets/CvesByCvssScore.js
@@ -16,6 +16,7 @@ import {
     cvssSeverityColorLegend,
 } from 'constants/severityColors';
 import { getScopeQuery } from 'Containers/VulnMgmt/Entity/VulnMgmtPolicyQueryUtil';
+import useFeatureFlags from 'hooks/useFeatureFlags';
 
 const CVES_QUERY = gql`
     query getCvesByCVSS($query: String, $scopeQuery: String) {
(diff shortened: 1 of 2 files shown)


md2119 (changes requested):
The query is incorrect. It is passing `IMAGE_COMPONENT ID` instead of `COMPONENT ID`

vjwilson:
> The query is incorrect. It is passing `IMAGE_COMPONENT ID` instead of `COMPONENT ID`

I've updated the query, based on our convo over Slack. Please re-review.

sachaudh (approved):
The code looks good to me here

md2119 (approved):

PR #2592, opened 2022-08-02
ROX-12066: Policy category edge store, datastore, searcher and indexer. Policy insert and update workflows changed to process categories differently under feature flag

clickboo:
## Description

- Added a policy category association table, and datastore, searcher and indexer layer
- Modified policy GET APIs to fill in category names by navigating the  new assoc table
- Future PRs will tackle policy add, update and import/export
- Deletes on policy and policy category will cause cascaded deletes on the assoc table
- Changes to policy insert and update workflows to process categories under a feature flag.
- 
## Checklist

## Testing Performed
Automated, full round of manual testing after import export of policies is worked upon in upcoming PR.


clickboo:
Tests failing due to unrelated admission controller timeout issue in adm controller tests.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -54,6 +57,25 @@ func (ds *datastoreImpl) buildIndex() error {
 	return ds.indexer.AddPolicyCategories(categories)
 }
 
+func (ds *datastoreImpl) GetPolicyCategoriesForPolicy(ctx context.Context, policyID string) ([]*storage.PolicyCategory, error) {
+	ds.categoryMutex.Lock()
+	defer ds.categoryMutex.Unlock()
+
+	edges, err := ds.policyCategoryEdgeDS.SearchRawEdges(ctx, searchPkg.NewQueryBuilder().AddStrings(searchPkg.PolicyID, policyID).ProtoQuery())
```
theencee, line 64: why search the join table? You should just be able to policyCategoryDS.Search(query) which would do a join for you.

clickboo: based on slack discussion, the suggestion ^ should work since the search will run on the edge and linked tables if the search term (in this case policy id) is not found in the category table.

clickboo:
Ping

md2119 (changes requested):
Just a note that search by `Category` won't work on postgres. The reason being, on postgres this field is served through new table and it is also present on `policy.proto` so all queries on policy store will stop at policy store because it serves the search key, but it is empty.

central/policycategoryedge/datastore/datastore_impl.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return imgs, nil
+}
+
+func (ds *datastoreImpl) Count(ctx context.Context, q *v1.Query) (int, error) {
+	return ds.searcher.Count(ctx, q)
+}
+
+func (ds *datastoreImpl) Get(ctx context.Context, id string) (*storage.PolicyCategoryEdge, bool, error) {
+	filteredIDs, err := ds.filterReadable(ctx, []string{id})
```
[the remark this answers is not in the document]
clickboo: Yes edge store and anything new categories will work postgres only. Will remove

proto/storage/relations.proto
```diff
@@ -84,3 +84,11 @@ message ClusterCVEEdge {
   string      cluster_id    = 4 [(gogoproto.moretags) = 'sql:"fk(Cluster:id)"'];
   string      cve_id        = 5 [(gogoproto.moretags) = 'sql:"fk(ClusterCVE:id),no-fk-constraint"'];
 }
+
+// Policy relations
+
+message PolicyCategoryEdge {
+    string        id                 = 1[(gogoproto.moretags) = 'sql:"pk,id"'];
+    string        policy_id          = 2[(gogoproto.moretags) = 'sql:"fk(Policy:id)" search:"Policy ID,store,hidden"'];
+    string        category_id        = 3[(gogoproto.moretags) = 'sql:"fk(PolicyCategory:id)" search:"Category ID,store,hidden"'];
```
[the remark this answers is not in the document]
clickboo: Yeah, was torn between going verbose or not. That was my verbose option ;-)

clickboo:
> Just a note that search by `Category` won't work on postgres. The reason being, on postgres this field is served through new table and it is also present on `policy.proto` so all queries on policy store will stop at policy store because it serves the search key, but it is empty.

Good point! So in this case, we should remove search tag on the policy proto when we move to postgres and for intermediate testing. I don't see a better way given how vastly different our search infra is for rocksdb vs postgres.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -54,6 +54,25 @@ func (ds *datastoreImpl) buildIndex() error {
...
+		return []*storage.PolicyCategory{}, err
+	}
+	categoryIds := make([]string, len(categories))
+	for _, c := range categories {
+		categoryIds = append(categoryIds, c.GetId())
+	}
+	categories, missingIDs, err := ds.storage.GetMany(ctx, categoryIds)
+	if len(missingIDs) > 0 {
+		panic(fmt.Sprintf("policy category edges present for categories %q, but categories not found", missingIDs))
+	}
```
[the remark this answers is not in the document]
clickboo: Aah yes, remnant from the time when I was not using search there. forgot to remove when I changed to using search.

clickboo:
> > Just a note that search by `Category` won't work on postgres. The reason being, on postgres this field is served through new table and it is also present on `policy.proto` so all queries on policy store will stop at policy store because it serves the search key, but it is empty.
> 
> Good point! So in this case, we should remove search tag on the policy proto when we move to postgres and for intermediate testing. I don't see a better way given how vastly different our search infra is for rocksdb vs postgres.

Transform added.

central/policycategoryedge/datastore/datastore_impl.go
```diff
@@ -0,0 +1,110 @@
+package datastore
```
md2119, line 1: I think we discuss this earlier, but let me draw it out here. There the is a way to avoid datastore layer for edges entirely if you follow the pkg structure below:
```
category
-- datastore
    -- store
         -- category
              -- postgres
         -- edge
              -- postgres

```
Category datastore impl:
```
{
  categoryStorage
  edgeStorage
  ...
}
```
You end up with a much simple structure and less code 😉 . The edge store becomes internal only because it does not need to be exposed as the functionality to set categories is abstracted behind category datastore.

central/policy/search/searcher_impl.go
```diff
@@ -104,6 +105,7 @@ func convertPolicy(policy *storage.Policy, result search.Result) *v1.SearchResul
 func formatSearcher(unsafeSearcher blevesearch.UnsafeSearcher) search.Searcher {
 	safeSearcher := blevesearch.WrapUnsafeSearcherAsSearcher(unsafeSearcher)
 	transformedSortFieldSearcher := sortfields.TransformSortFields(safeSearcher, policyMapping.OptionsMap)
-	paginatedSearcher := paginated.Paginated(transformedSortFieldSearcher)
+	transformedCategoryNameSearcher := policycategory.TransformCategoryNameFields(transformedSortFieldSearcher)
```
md2119, line 108: This transformation must be done behind feature flag.

md2119 (changes requested):

central/policy/search/searcher_impl.go
clickboo: Check transform func - it returns if feature flag is not set.

central/policycategoryedge/datastore/datastore_impl.go
clickboo: Yeah, I don't want to take this on just yet - working for correctness first, optimization later :). But noted :)

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +34,67 @@ var (
...
+	policyCategoryEdges := make([]*storage.PolicyCategoryEdge, 0, len(categoryIds))
+	for _, id := range categoryIds {
+		policyCategoryEdges = append(policyCategoryEdges, &storage.PolicyCategoryEdge{
+			Id:         uuid.NewV4().String(),
+			PolicyId:   policyID,
+			CategoryId: id,
+		})
+	}
+	return ds.policyCategoryEdgeDS.Upsert(ctx, policyCategoryEdges)
```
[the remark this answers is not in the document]
clickboo: Done

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +34,67 @@ var (
...
+	categories := make([]*storage.PolicyCategory, 0, len(categoryNames))
+	err = ds.storage.Walk(policyCategoryCtx, func(category *storage.PolicyCategory) error {
+		categories = append(categories, category)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	categoryIds := make([]string, 0, len(categoryNames))
+	categoryNameIDMap := categoryUtils.GetCategoryNameToIDs(categories)
```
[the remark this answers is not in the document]
clickboo: Yes, forgot to add that todo but that is in the intent. Stay tuned for some follow on optimization PR.

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +34,67 @@ var (
...
+
+	// disassociate all categories from given policy
+	results, err := ds.policyCategoryEdgeDS.Search(ctx, searchPkg.NewQueryBuilder().AddExactMatches(searchPkg.PolicyID, policyID).ProtoQuery())
+	if err != nil {
+		return err
+	}
+	err = ds.policyCategoryEdgeDS.Delete(ctx, searchPkg.ResultsToIDs(results)...)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
clickboo: Yes, that came in after or during my doing this. Done.

central/policy/search/searcher_impl.go
```diff
@@ -75,7 +77,7 @@ func (ds *searcherImpl) searchPolicies(ctx context.Context, q *v1.Query) ([]*sto
 
 func (ds *searcherImpl) Search(ctx context.Context, q *v1.Query) ([]search.Result, error) {
 	if ok, err := policySAC.ReadAllowed(ctx); err != nil || !ok {
-		return nil, err
+		return nil, errorsPkg.Wrap(err, "unauthorized")
```
md2119, line 80: check `errox/errors.go` for sentinel errors

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +35,81 @@ var (
...
+	ds.categoryMutex.Lock()
+	defer ds.categoryMutex.Unlock()
+
+	results, err := ds.policyCategoryEdgeDS.SearchEdges(ctx, searchPkg.NewQueryBuilder().AddExactMatches(searchPkg.PolicyID, policyID).ProtoQuery())
+	if err != nil {
+		return err
+	}
+	existingCategoryNames := make([]string, 0, len(results))
+	for _, result := range results {
+		existingCategoryNames = append(existingCategoryNames, result.Name)
```
md2119, line 55: This seems incorrect because edge `result.Name` is ID

```
func convertOne(obj *storage.PolicyCategoryEdge, result *search.Result) *v1.SearchResult {
	return &v1.SearchResult{
		Category:       v1.SearchCategory_POLICY_CATEGORY_EDGE,
		Id:             obj.GetId(),
		Name:           obj.GetId(),
		FieldToMatches: search.GetProtoMatchesMap(result.Matches),
		Score:          result.Score,
	}
}
```

md2119: or `existingCategoryNames` seems to be a misleading name

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +35,81 @@ var (
...
+	for _, c := range categoryNames {
+		if categoryNameIDMap[strings.Title(c)] != "" {
+			categoryIds = append(categoryIds, categoryNameIDMap[c])
+		} else {
+			newCategory := &storage.PolicyCategory{
+				Id:        uuid.NewV4().String(),
+				Name:      strings.Title(c),
+				IsDefault: false,
+			}
+			categoriesToAdd = append(categoriesToAdd, newCategory)
```
md2119, line 94: In incoming `categoryNames` is not deduped. You must dedupe it: `categoryNameIDMap[newCategory.Name] = newCategory.Id`

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +35,81 @@ var (
...
+			categoryIds = append(categoryIds, newCategory.GetId())
+		}
+	}
```
md2119, line 97:
```suggestion
	for _, c := range categoryNames {
		categoryID := categoryNameIDMap[strings.Title(c)]
		if categoryID == "" {
		       newCategory := &storage.PolicyCategory{
				Id:        uuid.NewV4().String(),
				Name:      strings.Title(c),
				IsDefault: false,
			}
			categoryID = newCategory.GetId()
			categoriesToAdd = append(categoriesToAdd, newCategory)
			categoryNameIDMap[newCategory.Name] = categoryID
		}
		categoryIds = append(categoryIds, categoryID)
	}
```

central/policycategory/datastore/datastore_impl.go
```diff
@@ -32,11 +35,81 @@ var (
...
-	searcher search.Searcher
+	storage              store.Store
+	indexer              index.Indexer
+	searcher             search.Searcher
+	policyCategoryEdgeDS datastore.DataStore
+	categoryMutex        sync.Mutex
+}
+
+func (ds *datastoreImpl) SetPolicyCategoriesForPolicy(ctx context.Context, policyID string, categoryNames []string) error {
+	ds.categoryMutex.Lock()
```
md2119, line 46: check for ` !features.NewPolicyCategories.Enabled() || !features.PostgresDatastore.Enabled()` and throw unsupported. I didn't realize earlier that this datastore also exists on rocksdb.

central/policycategoryedge/datastore/datastore_impl.go
md2119: Since you are not planning on changing the structure RN, you will need to lock edge datastore upserts and deletes.

central/search/service/service_impl_test.go
```diff
@@ -110,25 +112,46 @@ type SearchOperationsTestSuite struct {
...
 }
 
 func (s *SearchOperationsTestSuite) SetupSuite() {
 	s.envIsolator = envisolator.NewEnvIsolator(s.T())
 }
 
 func (s *SearchOperationsTestSuite) SetupTest() {
-	s.envIsolator.Setenv(features.NewPolicyCategories.EnvVar(), "true")
+	s.envIsolator.Setenv(features.NewPolicyCategories.EnvVar(), "false")
+	s.envIsolator.Setenv(features.PostgresDatastore.EnvVar(), "false")
```
md2119, line 126: remove?

central/search/service/service_impl_test.go
```diff
@@ -110,25 +112,46 @@ type SearchOperationsTestSuite struct {
...
+
 	s.mockCtrl = gomock.NewController(s.T())
 	s.rocksDB = rocksdbtest.RocksDBForT(s.T())
 	var err error
 	s.boltDB, err = bolthelper.NewTemp(s.T().Name() + "-bolt.db")
 	s.NoError(err)
+
+	s.ctx = context.Background()
+	if features.PostgresDatastore.Enabled() {
```
md2119, line 135: This is never true because of above `SetEnv`

pkg/search/policycategory/transform.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/features"
+	"github.com/stackrox/rox/pkg/search"
+)
+
+// TransformCategoryNameFields transforms category name fields for the new data layout for categories in postgres.
+func TransformCategoryNameFields(searcher search.Searcher) search.Searcher {
+	if !features.NewPolicyCategories.Enabled() {
```
md2119, line 13: This culprit I shall not forget!

md2119 (dismissed):

central/search/service/service_impl_test.go
clickboo: The search service never worked for postgres btw. The way the postgres instance was created for this test would never work with feature flag turned on.

clickboo: And "This is never true because of above SetEnv" has always been true :)

central/policy/datastore/datastore_impl.go
```diff
@@ -91,9 +93,30 @@ func (ds *datastoreImpl) GetPolicy(ctx context.Context, id string) (*storage.Pol
...
+	if !features.NewPolicyCategories.Enabled() || !features.PostgresDatastore.Enabled() {
+		return nil
+	}
+	for _, p := range policies {
+		categories, err := ds.categoriesDatastore.GetPolicyCategoriesForPolicy(ctx, p.GetId())
+		if err != nil {
+			return err
+		}
+		for _, c := range categories {
+			p.Categories = append(p.Categories, c.GetName())
```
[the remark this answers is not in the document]
clickboo: Can do, although this func is used only in the get path (and we've already set categories to `nil` during upserts).

central/policycategory/datastore/datastore_impl.go
clickboo: Ok I've blocked the calls to this func under feature flag check ^

central/policy/datastore/datastore_impl_postgres_test.go
```diff
@@ -0,0 +1,192 @@
+//go:build sql_integration
```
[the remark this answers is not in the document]
clickboo: The point of this test is to check the new category workflow so turning off the category feature flag doesn't make sense here.

pkg/search/policycategory/transform.go
clickboo: :)

central/policy/datastore/datastore_impl_postgres_test.go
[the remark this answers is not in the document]
md2119: @clickboo If you have had the test I recommend, we would have uncovered the issue with transformation very quickly. Also, what happens when Postgres is enabled but the feature is not? I think it there is no harm in adding that test.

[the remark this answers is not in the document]
clickboo: Aah, missed this comment earlier. So my point was I can't turn off the postgres feature flag in _this_ test. I understand from latest comment your ask to add another test? That would be a non postgres test, some other unit test for searching policies by category when postgres ff is turned off. Will see what can be done.

md2119 (approved):

PR #2595, opened 2022-08-03
 ROX-12416, ROX-12114: Enable email tests for non-TLS connections

theencee:
## Description

Email tests have been disabled for a while since the mailgun account that was used to test it was deactivated. This brings some of them back.

Uses a third party SMTP relay server called [maildev](https://github.com/maildev/maildev) ("As-Is" [license](https://github.com/maildev/maildev/blob/master/LICENSE)) which is deployed at the beginning of each test. The NodeJS SMTP server accepts emails and stores it in memory and it can be queried with a REST API. 

Unfortunately maildev doesn't support TLS or StartTLS and that was discovered late. So TLS related tests are still disabled. However, there is still benefit in enabling some of it as it tests the new unauthenticated SMTP feature.

Next up:
* Enable tests that actually read the email sent via notifiers (violations, NG, etc)
* Figure out how to test SMTP. Perhaps using a [TLS enabled mail forwarder](https://github.com/huan/docker-simple-mail-forwarder) that forwards to maildev?

## Checklist

## Testing Performed

This is a test

### DIFF
--- a/qa-tests-backend/src/test/groovy/IntegrationsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/IntegrationsTest.groovy
@@ -1,13 +1,17 @@
-import common.Constants
-import groups.BAT
-import groups.Integration
-import groups.Notifiers
+import java.util.concurrent.TimeUnit
+
 import io.grpc.StatusRuntimeException
+import org.apache.commons.lang3.RandomStringUtils
+
 import io.stackrox.proto.storage.ClusterOuterClass
 import io.stackrox.proto.storage.NotifierOuterClass
 import io.stackrox.proto.storage.PolicyOuterClass
 import io.stackrox.proto.storage.ScopeOuterClass
-import java.util.concurrent.TimeUnit
+
+import common.Constants
+import groups.BAT
+import groups.Integration
+import groups.Notifiers
 import objects.AzureRegistryIntegration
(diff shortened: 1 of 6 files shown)


qa-tests-backend/src/test/groovy/IntegrationsTest.groovy
```diff
@@ -68,54 +72,85 @@ class IntegrationsTest extends BaseSpecification {
...
+    @SuppressWarnings('LineLength')
     @Unroll
     @Category([BAT])
-    @Ignore("ROX-8113 - email tests are broken")
-    def "Verify create Email Integration (port #port, disable TLS=#disableTLS, startTLS=#startTLS)"() {
+    def "Verify create Email Integration (disableTLS=#disableTLS, startTLS=#startTLS, authenticated=#authenticated, sendCreds=#sendCreds)"() {
         given:
+        "mailserver is running"
+        def mailServer = MailServer.createMailServer(orchestrator, authenticated, !disableTLS)
+        sleep 15 * 1000 // wait 15s for service to start
```
gavin-stackrox, line 82: I'd prefer something other then a 15 second sleep but ok if you want to fix in ROX-12418

gavin-stackrox (approved):
Good to see these resurrected. Thanks!

PR #2601, opened 2022-08-03
ROX-11774: Add basic BATS tests for roxctl netpol generate

vikin91:
## Description

This PR adds basic tests for the new command `roxctl netpol generate`.
Note that we do not test here the output of the the NP-Guard library (e.g., whether the produced network policies are correct), but the wrapper code in `roxctl` binary.

These tests are simple e2e tests and can be extended by adding Go unit tests in cobra.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- ~[ ] Evaluated and added CHANGELOG entry if required~
- ~[ ] Determined and documented upgrade steps~
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed

- [x] Executed the tests locally with `bats /tests/roxctl/bats-tests/local/roxctl-netpol-generate.bats`
- [ ] CI

PR #2602, opened 2022-08-03
ROX-11775: Introduce different file outputs for roxctl generate netpol

Maddosaurus:
## Description

This PR introduces more output formats for roxctl generate netpol.
I'm not set with the names and combinations of flags, the focus was getting the functionality implemented

- To stdout (merged copy-paste-ready yaml)
  - If no additional flags are provided
- To a single file (merged, same as stdout)
  - If `--merge-policies` combined with `-output-file` is provided
- 1 file per network policy
  - If `--split-policies` combined with `-output-folder` is provided

## Checklist

~~We're not announcing the new functionality until the online mode is done and working, so no Changelog and Docs updates.~~
Also, as this is not touching existing functionality, no upgrade steps are needed.
Tests are added in a second PR, https://github.com/stackrox/stackrox/pull/2601 

## Testing Performed
To test the new functionality, point `roxctl generate netpol` to any folder containing any pod-creating YAML resources.
It will recursively scan all YAMLs and generate a list of recommended Network Policies to stdout.
```
git clone https://github.com/nadgowdas/microservices-demo /tmp/microservices-demo
./roxctl netpol generate /tmp/microservices-demo
```
Observe the recommended Network Policies being generated and output, depending on which combination of flags you use.

### DIFF
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -8,6 +8,7 @@ import (
 	"github.com/np-guard/cluster-topology-analyzer/pkg/controller"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/pkg/protoconv/networkpolicy"
+	v1 "k8s.io/api/networking/v1"
 )
 
 func (cmd *generateNetpolCommand) generateNetpol() error {
@@ -38,6 +39,12 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		}
 	}
 
+	if cmd.splitPolicies {
+		if err := cmd.saveNetpolsToFolder(recommendedNetpols); err != nil {
+			return errors.Wrapf(err, "Error writing split Network Policies to folder")
+		}
+	}
+
 	return nil
 }
 
@@ -56,6 +63,26 @@ func (cmd *generateNetpolCommand) saveNetpolsToMergedFile(combinedNetpols string
 
 	if err := writeFile(filename, dirpath, combinedNetpols); err != nil {
 		return errors.Wrapf(err, "Error writing policy to file")
+	}
+	return nil
+}
+
+func (cmd *generateNetpolCommand) saveNetpolsToFolder(recommendedNetpols []*v1.NetworkPolicy) error {
+	for _, netpol := range recommendedNetpols {
+		policyName := netpol.GetName()
+		if policyName == "" {
+			policyName = string(netpol.GetUID())
+		}
+		filename := policyName + ".yaml"
+
+		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
+		if err != nil {
+			return errors.Wrap(err, "Error converting Network Policy object to YAML")
+		}
+
+		if err := writeFile(filename, cmd.outputFolderPath, yamlPolicy); err != nil {
+			return errors.Wrapf(err, "Error writing policy to file")
+		}
 	}
 	return nil
 }
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -22,6 +22,8 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		}
 		cmd.env.Logger().WarnfLn("Removed output path %s", cmd.outputFolderPath)
 	}
+	cmd.env.Logger().InfofLn("Writing generated Network Policies to %s", cmd.outputFolderPath)
+
 	var mergedPolicy string
 	yamlPolicies := make([]string, 0, len(recommendedNetpols))
 	for _, netpol := range recommendedNetpols {
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -96,7 +96,7 @@ func (cmd *generateNetpolCommand) saveNetpolsToFolder(recommendedNetpols []*v1.N
 
 func writeFile(filename string, destDir string, content string) error {
 	outputPath := filepath.Join(destDir, filename)
-	if err := os.MkdirAll(filepath.Dir(destDir), 0755); err != nil {
+	if err := os.MkdirAll(filepath.Dir(outputPath), 0755); err != nil {
 		return errors.Wrapf(err, "error creating directory for file %q", filename)
 	}
 
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -28,22 +28,21 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 	}
 	mergedPolicy = strings.Join(yamlPolicies, "\n---\n")
 
-	if !cmd.mergePolicies && !cmd.splitPolicies {
-		cmd.printNetpols(mergedPolicy)
-		return nil
-	}
-
-	if cmd.mergePolicies {
+	if cmd.mergeMode {
 		if err := cmd.saveNetpolsToMergedFile(mergedPolicy); err != nil {
 			return errors.Wrapf(err, "error saving merged Network Policies")
 		}
+		return nil
 	}
 
-	if cmd.splitPolicies {
+	if cmd.splitMode {
 		if err := cmd.saveNetpolsToFolder(recommendedNetpols); err != nil {
 			return errors.Wrapf(err, "error saving split Network Policies")
 		}
+		return nil
 	}
+
+	cmd.printNetpols(mergedPolicy)
 
 	return nil
 }
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -23,9 +23,8 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		if err != nil {
 			return errors.Wrap(err, "error converting Network Policy object to YAML")
 		}
-		mergedPolicy = fmt.Sprintf("%s---\n%s", mergedPolicy, yamlPolicy)
 	}
-
+mergedPolicy = strings.Join(yamlPolicies, "\n---\n")
 	if !cmd.mergePolicies && !cmd.splitPolicies {
 		cmd.printNetpols(mergedPolicy)
 		return nil
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -42,8 +42,12 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		return nil
 	}
 
-	cmd.printNetpols(mergedPolicy)
+	if cmd.stdoutMode {
+		cmd.printNetpols(mergedPolicy)
+		return nil
+	}
 
+	cmd.env.Logger().WarnfLn("No mode selected, not providing any output.")
 	return nil
 }
 
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -30,14 +30,14 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 
 	if cmd.mergeMode {
 		if err := cmd.saveNetpolsToMergedFile(mergedPolicy); err != nil {
-			return errors.Wrapf(err, "error saving merged Network Policies")
+			return errors.Wrap(err, "error saving merged Network Policies")
 		}
 		return nil
 	}
 
 	if cmd.splitMode {
 		if err := cmd.saveNetpolsToFolder(recommendedNetpols); err != nil {
-			return errors.Wrapf(err, "error saving split Network Policies")
+			return errors.Wrap(err, "error saving split Network Policies")
 		}
 		return nil
 	}
@@ -62,7 +62,7 @@ func (cmd *generateNetpolCommand) saveNetpolsToMergedFile(combinedNetpols string
 	}
 
 	if err := writeFile(filename, dirpath, combinedNetpols); err != nil {
-		return errors.Wrapf(err, "error writing merged Network Policies")
+		return errors.Wrap(err, "error writing merged Network Policies")
 	}
 	return nil
 }
@@ -81,7 +81,7 @@ func (cmd *generateNetpolCommand) saveNetpolsToFolder(recommendedNetpols []*v1.N
 		}
 
 		if err := writeFile(filename, cmd.outputFolderPath, yamlPolicy); err != nil {
-			return errors.Wrapf(err, "error writing policy to file")
+			return errors.Wrap(err, "error writing policy to file")
 		}
 	}
 	return nil
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -1,9 +1,9 @@
 package netpol
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/np-guard/cluster-topology-analyzer/pkg/controller"
 	"github.com/pkg/errors"
@@ -18,13 +18,16 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 	}
 
 	var mergedPolicy string
+	yamlPolicies := make([]string, 0, len(recommendedNetpols))
 	for _, netpol := range recommendedNetpols {
 		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
 		if err != nil {
 			return errors.Wrap(err, "error converting Network Policy object to YAML")
 		}
+		yamlPolicies = append(yamlPolicies, yamlPolicy)
 	}
-mergedPolicy = strings.Join(yamlPolicies, "\n---\n")
+	mergedPolicy = strings.Join(yamlPolicies, "\n---\n")
+
 	if !cmd.mergePolicies && !cmd.splitPolicies {
 		cmd.printNetpols(mergedPolicy)
 		return nil
@@ -51,9 +54,6 @@ func (cmd *generateNetpolCommand) printNetpols(combinedNetpols string) {
 
 func (cmd *generateNetpolCommand) saveNetpolsToMergedFile(combinedNetpols string) error {
 	dirpath, filename := filepath.Split(cmd.outputFilePath)
-	if dirpath == "" {
-		dirpath = "./"
-	}
 	if filename == "" {
 		filename = "policies.yaml"
 	}
@@ -90,6 +90,5 @@ func writeFile(filename string, destDir string, content string) error {
 		return errors.Wrapf(err, "error creating directory for file %q", filename)
 	}
 
-	perms := os.FileMode(0644)
-	return errors.Wrap(os.WriteFile(outputPath, []byte(content), perms), "error writing file")
+	return errors.Wrap(os.WriteFile(outputPath, []byte(content), os.FileMode(0644)), "error writing file")
 }
--- a/roxctl/generate/netpol/cta.go
+++ b/roxctl/generate/netpol/cta.go
@@ -14,41 +14,39 @@ import (
 func (cmd *generateNetpolCommand) generateNetpol() error {
 	recommendedNetpols, err := controller.PoliciesFromFolderPath(cmd.folderPath)
 	if err != nil {
-		return errors.Wrap(err, "Error synthesizing policies from folder")
+		return errors.Wrap(err, "error synthesizing policies from folder")
 	}
 
-	var yamlPolicies []string
 	var mergedPolicy string
 	for _, netpol := range recommendedNetpols {
 		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
 		if err != nil {
-			return errors.Wrap(err, "Error converting Network Policy object to YAML")
+			return errors.Wrap(err, "error converting Network Policy object to YAML")
 		}
-		yamlPolicies = append(yamlPolicies, yamlPolicy)
 		mergedPolicy = fmt.Sprintf("%s---\n%s", mergedPolicy, yamlPolicy)
 	}
 
 	if !cmd.mergePolicies && !cmd.splitPolicies {
-		cmd.printNetpolsToStdout(mergedPolicy)
+		cmd.printNetpols(mergedPolicy)
 		return nil
 	}
 
 	if cmd.mergePolicies {
 		if err := cmd.saveNetpolsToMergedFile(mergedPolicy); err != nil {
-			return errors.Wrapf(err, "Error writing merged Network Policies to file")
+			return errors.Wrapf(err, "error saving merged Network Policies")
 		}
 	}
 
 	if cmd.splitPolicies {
 		if err := cmd.saveNetpolsToFolder(recommendedNetpols); err != nil {
-			return errors.Wrapf(err, "Error writing split Network Policies to folder")
+			return errors.Wrapf(err, "error saving split Network Policies")
 		}
 	}
 
 	return nil
 }
 
-func (cmd *generateNetpolCommand) printNetpolsToStdout(combinedNetpols string) {
+func (cmd *generateNetpolCommand) printNetpols(combinedNetpols string) {
 	cmd.env.Logger().PrintfLn(combinedNetpols)
 }
 
@@ -62,7 +60,7 @@ func (cmd *generateNetpolCommand) saveNetpolsToMergedFile(combinedNetpols string
 	}
 
 	if err := writeFile(filename, dirpath, combinedNetpols); err != nil {
-		return errors.Wrapf(err, "Error writing policy to file")
+		return errors.Wrapf(err, "error writing merged Network Policies")
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
+		}
+		filename := policyName + ".yaml"
+
+		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
+		if err != nil {
+			return errors.Wrap(err, "Error converting Network Policy object to YAML")
+		}
+
+		if err := writeFile(filename, cmd.outputFolderPath, yamlPolicy); err != nil {
+			return errors.Wrapf(err, "Error writing policy to file")
```
0x656b694d, line 84: Nit: the error message could read: _"Error writing policy to file: Error writing file: ..."_.
Looks like `writeFile` should not capitalize the first word in its error messages. And in general, we do not capitalize it.

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
+			return errors.Wrapf(err, "Error writing split Network Policies to folder")
 		}
-		cmd.env.Logger().PrintfLn("---\n\n%s", yamlPolicy)
 	}
+
 	return nil
 }
+
+func (cmd *generateNetpolCommand) printNetpolsToStdout(combinedNetpols string) {
+	cmd.env.Logger().PrintfLn(combinedNetpols)
```
0x656b694d, line 52: `cmd.env.Logger` is not necessarily `stdout`.

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
 
+	var yamlPolicies []string
+	var mergedPolicy string
 	for _, netpol := range recommendedNetpols {
 		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
 		if err != nil {
-			return errors.Wrap(err, "Error converting YAML into Network Policy")
+			return errors.Wrap(err, "Error converting Network Policy object to YAML")
+		}
+		yamlPolicies = append(yamlPolicies, yamlPolicy)
```
0x656b694d, line 27: Is `yamlPolicies` ever used?

0x656b694d:
Will that feature work with `*.json` files?

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		return errors.Wrap(err, "Error synthesizing policies from folder")
 	}
 
+	var yamlPolicies []string
```
vikin91, line 20: Could:

```suggestion
	yamlPolicies := make([]string, 0, len(recommendedNetpols))
```

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
 	for _, netpol := range recommendedNetpols {
 		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
 		if err != nil {
-			return errors.Wrap(err, "Error converting YAML into Network Policy")
+			return errors.Wrap(err, "Error converting Network Policy object to YAML")
+		}
+		yamlPolicies = append(yamlPolicies, yamlPolicy)
+		mergedPolicy = fmt.Sprintf("%s---\n%s", mergedPolicy, yamlPolicy)
+	}
+
```
vikin91, line 30: Should:

```suggestion
	}
mergedPolicy = strings.Join(yamlPolicies, "\n---\n")
```

roxctl/generate/netpol/cta.go
vikin91, line 52: I hope this logger cannot be disabled with some extreme log-level setting

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
+func (cmd *generateNetpolCommand) printNetpolsToStdout(combinedNetpols string) {
+	cmd.env.Logger().PrintfLn(combinedNetpols)
+}
+
+func (cmd *generateNetpolCommand) saveNetpolsToMergedFile(combinedNetpols string) error {
+	dirpath, filename := filepath.Split(cmd.outputFilePath)
+	if dirpath == "" {
+		dirpath = "./"
+	}
```
vikin91, line 59: Do we need this if clause? It should also work with an empty dir as it would default to `.`.

roxctl/generate/netpol/netpol.go
```diff
@@ -33,7 +37,13 @@ func Command(cliEnvironment environment.Environment) *cobra.Command {
 		},
 	}
 
-	c.Flags().BoolVar(&generateNetpolCmd.offline, "offline", false, "whether to connect to a central instace for additional information")
+	c.Flags().StringVar(&generateNetpolCmd.outputFolderPath, "output-dir", "./policies", "path to the output directory for generated policies")
+	c.Flags().StringVar(&generateNetpolCmd.outputFilePath, "output-file", "./policies.yaml", "path to the output file for merged policies")
+	c.Flags().BoolVar(&generateNetpolCmd.mergePolicies, "merge-policies", false, "Merge all generated Network Policies into a single file. Combine with -f for target file")
```
vikin91, line 42:
```suggestion
	c.Flags().BoolVar(&generateNetpolCmd.mergePolicies, "merge-policies", false, "Merge all generated Network Policies into a single file. Combine with --output-file for target file")
```

roxctl/generate/netpol/netpol.go
```diff
@@ -33,7 +37,13 @@ func Command(cliEnvironment environment.Environment) *cobra.Command {
 		},
 	}
 
-	c.Flags().BoolVar(&generateNetpolCmd.offline, "offline", false, "whether to connect to a central instace for additional information")
+	c.Flags().StringVar(&generateNetpolCmd.outputFolderPath, "output-dir", "./policies", "path to the output directory for generated policies")
+	c.Flags().StringVar(&generateNetpolCmd.outputFilePath, "output-file", "./policies.yaml", "path to the output file for merged policies")
```
vikin91, line 41: Could we maybe add the short version of the params? Like `-d` and `-f` maybe?

roxctl/generate/netpol/cta.go
```diff
@@ -12,12 +17,82 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
...
+}
+
+func writeFile(filename string, destDir string, content string) error {
+	outputPath := filepath.Join(destDir, filename)
+	if err := os.MkdirAll(filepath.Dir(outputPath), 0755); err != nil {
+		return errors.Wrapf(err, "Error creating directory for file %q", filename)
+	}
+
+	perms := os.FileMode(0644)
+	return errors.Wrap(os.WriteFile(outputPath, []byte(content), perms), "Error writing file")
```
vikin91, line 97: Could: I suggest to get rid of the var `perms` as we use it only locally and only once.

```suggestion
	return errors.Wrap(os.WriteFile(outputPath, []byte(content), os.FileMode(0644)), "Error writing file")
```

vikin91 (commented):
Looks good already! I went through the code but I still want to try it out in the console (will finish my review once I am done with this).

roxctl/generate/netpol/cta.go
Maddosaurus: That's correct, thanks! I've renamed the function to remove the notion of stdout.

roxctl/generate/netpol/cta.go
Maddosaurus: Thanks for the recommendation. I've updated the wrap messages.

roxctl/generate/netpol/cta.go
Maddosaurus: Well spotted, that's a leftover of dev time. Removed :)

Maddosaurus:
> Will that feature work with `*.json` files?

@0x656b694d It is first and foremost planned for static analysis of pod creating YAML resources, like Deployments, Pods, or ConfigMaps. 
The library we're using (NP-Guard) is doing a regex on the provided path to find YAML/YML, so no JSON support right now. If that is a use case that should be added, please let us know so I can forward that to the devs!

roxctl/generate/netpol/cta.go
Maddosaurus: This is a unused bit of code that I removed, but that's a great suggestion, thanks!

roxctl/generate/netpol/cta.go
Maddosaurus: Yes please. Would have accepted it but it got outdated because of another push. Thanks :)

roxctl/generate/netpol/netpol.go
Maddosaurus: Great idea! Done!

roxctl/generate/netpol/cta.go
Maddosaurus: We actually don't. Thanks!

roxctl/generate/netpol/cta.go
Maddosaurus: True. I based this on the Use Case that a User actually wants to see the output, so I wouldn't hope they crank the loglevel to something extreme. 
That said, is the roxctl loglevel configurable through an env variable? I don't think I've seen something like that so far.

vikin91:
Giving it another look now

lvalerom (changes requested):
The code looks good to me, I'm just blocking to discuss the following:

I don't understand the usability of `--merge-policies` and `--split-policies`. If these two flags are forced to be used with `-f` and `-d`, wouldn't it make more sense to remove them?

Also the user can technically type:

```
roxctl generate netpol -f netpols.yaml --split-policies --merge-policies -d tmp microservices-demo/
```

and the command will work. I'm not sure if this is intended but maybe it would make sense to make the flags exclusive.

```golang
c.MarkFlagsMutuallyExclusive("output-dit", "output-file")
```

WDYT?

vikin91:
That is a good testcase @lvalerom! I would also support the idea with
```
c.MarkFlagsMutuallyExclusive("output-dir", "output-file")
```

vikin91 (changes requested):
There are several issues unfortunately:

### 1. Defaults don't work

The help text suggests that `-f, --output-file string   path to the output file for merged policies (default "./policies.yaml")`
however, the following run ends with an error:

```sh
➜ bin/darwin/roxctl generate netpol --merge-policies /tmp/microservices-demo
ERROR:	if any flags in the group [merge-policies output-file] are set they must all be set; missing [output-file]
```

The error msg is a bit confusing - `if any flags (...) are set they must all be set` 🤷 The part `missing [output-file]` is clear but contradicts the default from the help.

### 2. We run into risk overwriting existing files

Consider the following scenario:

```shell
colima stackrox on  mm/rox-11776 [?]
➜ mkdir -p important
colima stackrox on  mm/rox-11776 [?]
➜ echo "xxxxx" > important/adservice-netpol.yaml
colima stackrox on  mm/rox-11776 [?]
➜ echo "xxxxx" > important/super-important.yaml
colima stackrox on  mm/rox-11776 [?]
➜ bin/darwin/roxctl generate netpol -d ./important --split-policies /tmp/microservices-demo/
colima stackrox on  mm/rox-11776 [?]
➜  ll important
total 16
-rw-r--r--  1 prygiels  staff  6 Aug  4 10:46 adservice-netpol.yaml
-rw-r--r--  1 prygiels  staff  6 Aug  4 10:47 super-important.yaml
colima stackrox on  mm/rox-11776 [?]
➜ echo "xxxxx" > important/super-important.yaml
colima stackrox on  mm/rox-11776 [?]
✗  bin/darwin/roxctl generate netpol -d ./important --split-policies /tmp/microservices-demo/
colima stackrox on  mm/rox-11776 [?]
➜ ll important
total 104
-rw-r--r--  1 prygiels  staff   344 Aug  4 10:47 adservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   655 Aug  4 10:47 cartservice-netpol.yaml
-rw-r--r--  1 prygiels  staff  1174 Aug  4 10:47 checkoutservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   485 Aug  4 10:47 currencyservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   357 Aug  4 10:47 emailservice-netpol.yaml
-rw-r--r--  1 prygiels  staff  1343 Aug  4 10:47 frontend-netpol.yaml
-rw-r--r--  1 prygiels  staff   395 Aug  4 10:47 loadgenerator-netpol.yaml
-rw-r--r--  1 prygiels  staff   362 Aug  4 10:47 paymentservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   632 Aug  4 10:47 productcatalogservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   557 Aug  4 10:47 recommendationservice-netpol.yaml
-rw-r--r--  1 prygiels  staff   349 Aug  4 10:47 redis-cart-netpol.yaml
-rw-r--r--  1 prygiels  staff   487 Aug  4 10:47 shippingservice-netpol.yaml
-rw-r--r--  1 prygiels  staff     6 Aug  4 10:47 super-important.yaml
```

The folder `important` already existed and it had two files `adservice-netpol.yaml` and `super-important.yaml`. After running the command, the file `adservice-netpol.yaml` got overwritten ❌  and the `super-important.yaml` was kept ✅ 

I think we should protect users against accidental overwrites. There are two options for this (and they are not exclusive - maybe we should do both?):

1. Expect the `-d` directory not to exist before the generation (we also expect this for other parts of `roxctl`, for example `central generate`) and return an error. We could do the same if the file specified with `-f` already exists.
2. Provide a flag that would allow overwrites. I think the other parts of `roxctl` use `--remove` flag for this.

### 3. Running in both modes

(Already mentioned by Luis)

```sh
➜ bin/darwin/roxctl generate netpol -d ./important --split-policies /tmp/microservices-demo/ -f policies.yaml --merge-policies
```

this actually works and produces both results - maybe it is not so bad?
However, let's try to construct a race like this:

```sh
➜ bin/darwin/roxctl generate netpol -d ./important --split-policies /tmp/microservices-demo/ -f important/adservice-netpol.yaml --merge-policies
```

What is the content of the `important/adservice-netpol.yaml`? 
- :one: huge yaml with all policies or 
- :two: only a singe policy named as the file itself?

Maddosaurus:
@lvalerom I've consolidated the flags to only use -f and -d. 
We do need to keep in mind to write very clear and concise feature documentation for this, as it's not directly intuitive.
Feel free to take a look :)

roxctl/generate/netpol/netpol.go
```diff
@@ -25,39 +29,60 @@ type generateNetpolCommand struct {
...
+	cmd.mergeMode = c.Flags().Changed("output-file")
+	return nil
+}
+
+func (cmd *generateNetpolCommand) validate() error {
+	if cmd.splitMode {
+		if err := cmd.setupPath(cmd.outputFolderPath); err != nil {
+			return errors.Wrapf(err, "failed to set up folder path")
+		}
+	} else if cmd.mergeMode {
```
lvalerom, line 63: [nit] here since `splitMode` and `mergeMode` are mutually exclusive you don't need to check the second variable.

roxctl/generate/netpol/netpol.go
```diff
@@ -13,8 +16,9 @@ type generateNetpolCommand struct {
 	folderPath       string
 	outputFolderPath string
 	outputFilePath   string
-	mergePolicies    bool
-	splitPolicies    bool
+	removeOutputPath bool
+	mergeMode        bool
+	splitMode        bool
```
lvalerom, line 21: [nit] Maybe `mergeMode` and `splitMode` can be combined into one variable now that they are mutually exclusive. You can have something like.

```golang
type writeMode int
const (
    splitMode writeMode = iota
    mergeMode
)

type generateNetpolCommand struct {
    // [...]
    mode writeMode
}

**Review by lvalerom — APPROVED** [review]
Looks good to me 🙂 Just a couple of optional suggestions.

File: roxctl/generate/netpol/netpol.go (hunk 11 above)
Comment (Maddosaurus) [reply roxctl/generate/netpol/netpol.go:21]: This is a tertiary situation, although I admit I'm not doing the best job of reflecting this in code :)
We have the third option to neither set merge nor split mode. If none are set, `generate netpol` outputs YAML to stdout. 

I could introduce a mode struct and switch depending on that (merge, split, stdout), if that's a bit more intuitive. 
What do you think?

File: roxctl/generate/netpol/cta.go (hunk 2 above)
Comment (0x656b694d) [reply roxctl/generate/netpol/cta.go:51]: I think this function should actually print to stdout. This is not a logging message that is printed, that's the output.

File: roxctl/generate/netpol/netpol.go (hunk 11 above)
Comment (lvalerom) [reply roxctl/generate/netpol/netpol.go:21]: You can also add another mode to the const list I mentioned (`stdoutMode`). I'm not sure what would be the added value of defining a struct for the mode but I might be missing something 🤔

**Comment by 0x656b694d** [comment]
> pod creating YAML resources

That sounds like an artificial limitation. Pod creating resources could be, and often times are, JSON files.

**Comment by Maddosaurus** [comment]
> > pod creating YAML resources
> 
> That sounds like an artificial limitation. Pod creating resources could be, and often times are, JSON files.

@0x656b694d That is a very good point. As the file discovery is done by the external lib, let me raise that issue over in their repo!

**Comment by vikin91** [comment]
> > > pod creating YAML resources
> > 
> > 
> > That sounds like an artificial limitation. Pod creating resources could be, and often times are, JSON files.
> 
> @0x656b694d That is a very good point. As the file discovery is done by the external lib, let me raise that issue over in their repo!

Do we also want to request API change to accept a list of files instead a folder path? Or shall we do this separately?

File: roxctl/generate/netpol/netpol.go (hunk 11 above)
Comment (vikin91) [reply roxctl/generate/netpol/netpol.go:21]: If we want to take it one step further, we could write it in a way that the function accepts a `writer` interface and we have 3 writers: file, files, stdout

Comment (vikin91) [reply roxctl/generate/netpol/netpol.go:21]: After going through the code, I like @lvalerom's idea :) Maybe you could consider applying it? :)

File: roxctl/generate/netpol/netpol.go [hunk 12]
```diff
@@ -21,33 +29,60 @@ type generateNetpolCommand struct {
...
+			if err := generateNetpolCmd.validate(); err != nil {
 				return err
 			}
 			return generateNetpolCmd.generateNetpol()
 		},
 	}
-
-	c.Flags().BoolVar(&generateNetpolCmd.offline, "offline", false, "whether to connect to a central instace for additional information")
+	c.Flags().BoolVar(&generateNetpolCmd.removeOutputPath, "remove", false, "remove the output path if it already exists")
+	c.Flags().StringVarP(&generateNetpolCmd.outputFolderPath, "output-dir", "d", "", "save generated policies into target folder - 1 file per policy")
```
vikin91, line 45: Numbers between 0-10 should be spelled out. Maybe use "single", or "one"?
```suggestion
	c.Flags().StringVarP(&generateNetpolCmd.outputFolderPath, "output-dir", "d", "", "save generated policies into target folder - single file per policy")
```

roxctl/generate/netpol/netpol.go
```diff
@@ -21,33 +29,60 @@ type generateNetpolCommand struct {
...
+	c.Flags().StringVarP(&generateNetpolCmd.outputFilePath, "output-file", "f", "", "save and merge generated policies into a single YAML file")
```
vikin91, line 46: I am still not sure about the capitalization - let's head for consistency with the rest of roxctl, but if there is no pattern, then I slightly prefer small caps.

```suggestion
	c.Flags().StringVarP(&generateNetpolCmd.outputFilePath, "output-file", "f", "", "save and merge generated policies into a single yaml file")
```

roxctl/generate/netpol/netpol.go
vikin91: Replying to Luis: there is the stdout mode as well and then both of these flags would be unset.

vikin91 (changes requested):
Looks much better, thanks for improving. 

My comments in the code are not blocking, but I think that the following issues discovered during manual testing need to be corrected:

### 1. This error message is unreadable

```sh
✗  bin/darwin/roxctl generate netpol -d ./important /tmp/microservices-demo/ -f important/adservice-netpol.yaml
ERROR:	if any flags in the group [output-dir output-file] are set none of the others can be; [output-dir output-file] were all set
```

Issues are:
- I have not used `output-dir output-file` but `-f -d` - maybe we could use: `[-d|--output-dir, -f|--output-file]`
- Wording - maybe simply: `ERROR: Flags [-d|--output-dir, -f|--output-file] cannot be used together`

### 2. Double error

```sh
✗  bin/darwin/roxctl generate netpol -d ./important /tmp/microservices-demo/
ERROR:	Output path ./important already exists. Use --remove to overwrite or select a different path.
ERROR:	failed to set up folder path: path ./important already exists
```

I mentioned this in the code - maybe we can fix that by removing the logging and keep only returning the error?

The rest looks great 🎉  but consider maybe addressing the detailed optional feedback ;)

roxctl/generate/netpol/netpol.go
```diff
@@ -21,33 +29,60 @@ type generateNetpolCommand struct {
...
+func (cmd *generateNetpolCommand) setupPath(path string) error {
+	if _, err := os.Stat(path); err == nil {
+		if cmd.removeOutputPath {
+			if err := os.RemoveAll(path); err != nil {
+				return errors.Wrapf(err, "failed to remove output path %s", path)
+			}
+			cmd.env.Logger().WarnfLn("Removed output path %s", path)
+		} else {
+			cmd.env.Logger().ErrfLn("Output path %s already exists. Use --remove to overwrite or select a different path.", path)
+			return errox.AlreadyExists.Newf("path %s already exists", path)
```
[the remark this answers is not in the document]
vikin91: Me from the future: It does lead to it - this should be corrected.

roxctl/generate/netpol/cta.go
```diff
@@ -1,23 +1,93 @@
...
 	}
 	return nil
 }
+
+func (cmd *generateNetpolCommand) saveNetpolsToFolder(recommendedNetpols []*v1.NetworkPolicy) error {
+	for _, netpol := range recommendedNetpols {
+		policyName := netpol.GetName()
+		if policyName == "" {
+			policyName = string(netpol.GetUID())
+		}
```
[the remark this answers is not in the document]
Maddosaurus: I've [checked with the NPG folks](https://coreos.slack.com/archives/C03719D0ABW/p1659687128685229?thread_ts=1659597334.976399&cid=C03719D0ABW) and they let me know they always use the workload pods' names with a `-netpol` suffix.
AFAIK pods cannot have  `/` in their name ([if I read this right](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/)), so this shouldn't be a problem.

roxctl/generate/netpol/netpol.go
```diff
@@ -21,33 +29,60 @@ type generateNetpolCommand struct {
...
 
-func (cmd *generateNetpolCommand) construct(args []string) error {
-	cmd.folderPath = args[0]
+func (cmd *generateNetpolCommand) setupPath(path string) error {
+	if _, err := os.Stat(path); err == nil {
+		if cmd.removeOutputPath {
+			if err := os.RemoveAll(path); err != nil {
+				return errors.Wrapf(err, "failed to remove output path %s", path)
+			}
+			cmd.env.Logger().WarnfLn("Removed output path %s", path)
```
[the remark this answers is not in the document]
Maddosaurus: I'm with you that `info` is more appropriate here, but I've designed this the same way `roxctl helm output` is built to keep consistency. I'm happy to change it, but maybe we should change this for all subcommands that do it this way, right?

roxctl/generate/netpol/cta.go
```diff
@@ -1,23 +1,93 @@
...
+		filename := policyName + ".yaml"
+
+		yamlPolicy, err := networkpolicy.KubernetesNetworkPolicyWrap{NetworkPolicy: netpol}.ToYaml()
+		if err != nil {
+			return errors.Wrap(err, "error converting Network Policy object to YAML")
+		}
+
+		if err := writeFile(filename, cmd.outputFolderPath, yamlPolicy); err != nil {
+			return errors.Wrapf(err, "error writing policy to file")
```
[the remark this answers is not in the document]
Maddosaurus: That is a great idea. Updated!

roxctl/generate/netpol/netpol.go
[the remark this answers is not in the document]
Maddosaurus: Thanks for spotting this - updated

Maddosaurus:
> 1. This error message is unreadable

@vikin91 Yes, that's what makes me a bit sad as well. This is the default error message of the cobra lib.
To date, I have not found a way to provide the validation functions custom error messages.
I'd rather stick to on-board validation done by cobra than to roll our own validation, but if the error messages are too cryptic, we should do it ourselves.
The third option would be a PR/refactor of cobra to allow custom error messages, but we should address this in a follow up, should we decide to go that route.
What do you think?

roxctl/generate/netpol/cta.go
```diff
@@ -1,23 +1,93 @@
...
+		}
+	}
+	return nil
+}
+
+func writeFile(filename string, destDir string, content string) error {
+	outputPath := filepath.Join(destDir, filename)
+	if err := os.MkdirAll(filepath.Dir(outputPath), 0755); err != nil {
```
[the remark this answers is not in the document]
Maddosaurus: Semantically, you're right. Funtionality-wise I can confirm that it works this way. 
I still changed it, thanks!

vikin91 (approved):
Unlocking as most of the comments were optional 👍

vikin91:
I have corrected that error, I couldn't stand it :) Maybe I should contribute this correction to the cobra upstream? 😂

roxctl/generate/netpol/cta.go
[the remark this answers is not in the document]
vikin91: I was wrong here - the initial version was correct. This was failing when we wanted to write a file into a dir that was missing. Now - by passing the `outputPath` as arg - the directory is created if required.

PR #2605, opened 2022-08-03
Add the capability to DeleteByQuery to Postgres stores

connorgorman:
## Description

There are some cases where we want to be able to delete by query and this allows it safely in the store (properly taking into account SAC)

## Checklist

## Testing Performed

Will run scale test to look at pod deletion time

### DIFF
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -85,6 +85,7 @@ type Store interface {
     Upsert(ctx context.Context, obj *{{.Type}}) error
     UpsertMany(ctx context.Context, objs []*{{.Type}}) error
     Delete(ctx context.Context, {{template "paramList" $pks}}) error
+    DeleteByQuery(ctx context.Context, q *v1.Query) error
 {{- end }}
 
 {{- if $singlePK }}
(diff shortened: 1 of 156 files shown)


md2119 (commented):
Is the use case primarily garbage collection?

theencee (approved):

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -635,6 +636,55 @@ func (s *storeImpl) Delete(ctx context.Context, {{template "paramList" $pks}}) e
...
 
+
+{{- if not .JoinTable }}
+// DeleteByQuery removes the objects based on the passed query
+func (s *storeImpl) DeleteByQuery(ctx context.Context, query *v1.Query) error {
+    {{- if not $inMigration}}
+	defer metrics.SetPostgresOperationDurationTime(time.Now(), ops.Remove, "{{.TrimmedType}}")
+    {{- end}}{{/* if not .inMigration */}}
+
+    var sacQueryFilter *v1.Query
```
md2119, line 647: Move this down to where it is actually used.

md2119: You actually do not need to declare it here. You can use short-var declaration because the `if-else` clause ensures the var is always init.

md2119 (approved):

tools/generate-helpers/pg-table-bindings/store.go.tpl
connorgorman: So the conjunction portion always has the var so it needs to be declared. Also, all of the other sac portions do it this way so want to keep it consistent

PR #2609, opened 2022-08-03
Update UI to react-scripts v5

dvail:
## Description

Updates us to react-scripts 5 and webpack 5.

If you want to test this locally, you very likely need to remove `ui/node_modules` and `ui/apps/platform/node_modules` and then `yarn install` as mentioned in the [release notes](https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#migrating-from-40x-to-500).

### **Update react-scripts to latest v5.0.1**

Of course.

### **Update redoc from 2.0.0.rc65 -> 2.0.0.rc74**

Webpack 5 no longer automatically polyfills native node built-in modules, and redoc rc65 uses the native `url` module. This is fixed in rc74 via https://github.com/Redocly/redoc/pull/1982.

### **Set resolution of transitive dependency `yaml` to newer patch version `1.10.2`.**

~~Webpack 5 is interpreting the export of `yaml` version `1.10.0` to be commonjs and throwing an error. This is fixed in 1.10.2 via https://github.com/eemeli/yaml/issues/208. I set a resolution override in the top level `package.json` - is there a better way to do this? `yaml` is only a transitive dependency in our project, used by a handful of packages.~~ This was resolved by upgrading the `yaml` dependency in yarn.lock to a compatible patch semver version, and the explicit resolution in the top level `package.json` was removed.

### **Update test consuming `Tab` component to comply with types**

This component specifies valid children as being of type `ReactElement`, but the test was passing a `string` child. I'm honestly not sure what was allowing this to pass before but making it fail now.

### **Remove explicit build for Tailwind and related package.json scripts**

CRA 5 ships with Tailwind support built in. We can now directly import the main CSS file and webpack will use our PostCSS and Tailwind config files to build the CSS on the fly. I changed the import from `index.css` to `app.tw.css` in our `index.tsx` to avoid an even larger PR diff due to moving the file.

### **Remove ESLint patch and set env var in package.json scripts**

react-scripts v5 now exposes an environment variable that can be used to disable ESLint running during the build process, which deprecates the need for the patch that we were using previously. This also adds the environment variable to package.json so that the behavior is the same across dev machines.

### **Add eslint dependencies implied by CRA usage to package.json `devDependencies`**

Running eslint manually fails by default because the `eslint-config-react-app` config dependency we are extending cannot be found after this change. This is due to that dependency being located under `ui/node_modules/react-scripts/node_modules/` instead of directly under `ui/node_modules`. https://github.com/eslint/eslint/issues/13283

I'm not sure what has changed in react-scripts or eslint (or another dependency) that is causing the package install location to change compared to react-scripts 4. The easiest fix is to add the missing dependencies to our `devDependencies` in package.json, which ensures they will be installed directly under `ui/node_modules`. I feel like there might be a better way, since in the [documentation](https://github.com/facebook/create-react-app/tree/main/packages/eslint-config-react-app#usage-in-create-react-app-projects) it is mentioned that reliance on these dependencies should be implied by our usage of CRA, and we shouldn't need to explicitly add these deps.

The counterpoint here is that the same can be said for eslint, since although we are using CRA, we are using eslint manually via a package.json script, so it might be valid to follow the "Usage Outside of Create React App" section of the docs above.

### Fix unit test failure due to Jest type change

The error:
```
Matcher error: received value must be a mock or spy function
Received has type:  function
Received has value: [Function clearTimeout]
```
is fixed by spying on `global.clearTimeout` and asserting that the spy was called instead of the native function.

### **Squash all `yarn.lock` changes above into single change**

Since multiple changes above modify the dependencies and the changes will be squashed on merge, running `git checkout master -- yarn.lock` followed by `yarn install` will give us a single yarn.lock change to go with the single commit on `master`

## TODO

- [x] Check generated bundle sizes and compare to react-scripts 4 output
- [x] Test embedded API docs after update to redoc library
- [x] Test image built by CI
- [x] Fix unit test
- [x] Squash yarn.lock changes as last commit

## Checklist

## Testing Performed

- [x] Local `yarn build` works
- [x] Local `yarn start` works and hot reloads changes
- [x] Local `yarn test` works and shows all passing tests
- [x] Local `yarn lint` works and correctly reports errors, if any
- [x] CI passes and correctly builds a working image

## Bundle sizes

Bundle sizes were generated with `yarn install source-map-explorer` followed by `yarn clean && yarn build` and then `source-map-explorer 'build/static/css/*.css'` and `source-map-explorer 'build/static/js/*.js'`. Note that we need to enable source maps in package.json for the build (`"build": "yarn tailwind && CI=false react-scripts build",`) in order for this to work.

This change slightly increases the bundled CSS size (+0.06 MB), and greatly decreases the bundled JS size (-1.94 MB), possibly due to better tree-shaking in webpack 5. (Side note: we should look into enabling Tailwind's purge feature if possible in the monorepo, as that should cut out a massive amount of our CSS bundle.)

Bundle sizes before:
JS
image

CSS
image

Bundle sizes after:
JS
image

CSS
image


vjwilson (commented):
Seems solid, but let's wait to merge until early next week. Some folks may be doing active development, and we don't want any unforeseen glitch to derail participation in our hackathon.

dvail:
Tested this build available on a gke-default cluster at https://104.197.181.98/main/dashboard (DM for credentials, today only unless we want to extend the duration). No apparent build issues after a spot check across pages of the application.

vjwilson (approved):

PR #2610, opened 2022-08-03
ROX-11018: Add option to make auth providers unmodifiable

ivan-degtiarenko:
## Description

This PR adds the option to create unmodifiable auth providers in Central. This is done to prevent ACSMS users from accidentally logging out of their ACS instance by modifying or deleting the `sso.redhat.com` auth provider.
Within this PR `unmodifiable` field is added to auth provider proto. 
When `unmodifiable` set to `true`, it's impossible to:
* modify auth provider
* delete auth provider via UI
* delete auth provider via API until `force` query parameter is set to `true`. 

`force` is a temporary parameter which is introduced so in worst case scenario it's possible to fix authentication within Managed Central.  It will be removed after we switch to declarative configuration for `sso.redhat.com` auth provider

## Checklist

#### Manual tests
1. Create unmodifiable auth provider -- verify the configuration is non-editable in the UI, but groups can be changed
2. Create unmodifiable auth provider -- verify it's impossible to delete from UI
3.  Create unmodifiable auth provider -- verify it's impossible to delete from API until `force` is set to `true`
4. Create unmodifiable auth provider -- verify it's impossible to `PUT`/`PATCH` auth provider


ui/apps/platform/src/services/AuthService/AuthService.ts
```diff
@@ -147,7 +148,7 @@ export function updateAuthProvider(authProvider: AuthProvider): Promise<AuthProv
  * Saves auth provider either by creating a new one (in case ID is missed) or by updating existing one by ID.
  */
 export function saveAuthProvider(authProvider: AuthProvider): string | Promise<AuthProvider> {
-    if (authProvider.active) {
+    if (authProvider.active || authProvider.unmodifiable) {
         return authProvider.id;
     }
     return authProvider.id
```
0x656b694d, line 154: Not yours, but you touched it. Why this condition?

ivan-degtiarenko: `saveAuthProvider` is called when you press `Save` in the auth provider form UI. This button can be pressed when auth provider config is unchanged, but groups change. 

-->This change prevents this code from returning an error when someone changes the groups for auth provider.

0x656b694d: Ah, I didn't expand the code below, sorry.
Not yours, but that's a bad practice to leave something on the next line after return.

0x656b694d (approved):
LGTM

ui/apps/platform/src/services/AuthService/AuthService.ts
pedrottimark: As Michaël said, not yours, but we are reviewing it :)

The only 2 calls of the `saveAuthProvider` service function are in authSagas.js file above.

@ivan-degtiarenko To avoid inconsistency at the risk of temporary code duplication, what do you think about:

1. Add comment in `saveAuthProvider` function that `if` statement might be deleted it the future, because redundant with condition in authSagas.js file. It seems like condition at point of call not to make a request is clearer than invisible early return in the service function.

2. Include both conditions in authSagas.js file:

```diff
-            if (!remaining.unmodifiable) {
+            if (!remaining.active && !remaining.unmodifiable) {
                 yield call(AuthService.saveAuthProvider, remaining);
             }
```

ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProvidersList.tsx
```diff
@@ -67,51 +67,61 @@ function AuthProvidersList({ entityId, authProviders }: AuthProvidersListProps):
                     </Tr>
                 </Thead>
                 <Tbody>
-                    {authProviders.map(({ id, name, type, defaultRole, groups = [] }) => {
-                        const typeLabel = getAuthProviderTypeLabel(type, availableProviderTypes);
+                    {authProviders.map(
+                        ({ id, name, type, defaultRole, unmodifiable, groups = [] }) => {
```
pedrottimark, line 71: An observation, not necessarily a change request: here is an idiom to avoid the change in indentation which makes it hard to find the few changes below.

```js
{authProviders.map((authProvider) => {
    const { id, name, type, defaultRole, unmodifiable, groups = [] } = authProvider;
```

pedrottimark:
@ivan-degtiarenko Something else came to mind. Is `unmodifiable: false` needed in the following line:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/AccessControl/AuthProviders/authProviders.utils.ts#L70

Or anywhere else in that file?

proto/storage/auth_provider.proto
```diff
@@ -33,6 +33,9 @@ message AuthProvider {
         string attribute_key   = 1;
         string attribute_value = 2;
     }
-    // EXPERIMENTAL.
     repeated RequiredAttribute required_attributes = 11;
+
+    // Modification via UI/API and deletion via UI are forbidden for this auth provider.
+    // Deletion via API is allowed only when `force` flag is set to true.
+    bool unmodifiable = 12;
```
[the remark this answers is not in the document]
misberner: I share the sentiment that there should be a more general way of highlighting where an object comes from/is stored. Given the discussion around possibly injecting checks at the SQL level, I am not sure if this wouldn't be defeating the purpose of doing something quick for service preview.

After all, is there really much utility in having objects that are unmodifiable stored in the SQL database? The way I was thinking about it, we would have a general datastore interface that provides an aggregated view onto a read-only in-memory store populated (and updated) from mounted declarative config layered on top of a SQL-based store.

This can get fairly complicated - we probably need code generation for both the in-memory store and the aggregation layer, some less-than-obvious for deciding to which underlying store writes/creates are directed, and a method to communicate to the UI where an object comes from/if it is modifiable (even though the last two could possibly be combined with a `source` or `origin` field).

If we store objects in the database, we need to support reconciliation efficiently. I.e., if the configmap gets deleted, we need to know which entities need to be deleted from the DB. If central comes back up, we need to have a way to maintain ID persistence (do we expect the user to specify a UUID in the configmap entry?). We need to have all kinds of rules how an object with a certain `origin` can be created, modified, deleted, but in the end, we take everything from a "dumb" SQL database without business logic, where it isn't clear how to deal with inconsistencies.


All in all, there are a lot of things to think about here. The use case of having unmodifiable objects that however can be created and deleted through the API might well be an artifact - and it _is_ weird in all kinds of ways, such as introducing an API parameter we seek to soon remove again, even though we generally want our APIs to be stable; or the fact that "unmodifiable" actually means "well, not really, but you have to say 'pretty please'". I do not see it as a stepping stone to supporting declarative configuration, to be honest, and that is not because of the implementation in this PR, but because it conceptually doesn't make a whole lot of sense.

There might be some way of evolving it to supporting the declarative configuration use case, but my current assessment would be that we will deprecate/delete everything related to this mechanism once we have support for declarative configuration. Looking to generalize here might not only slow down the implementation of the quick (albeit conceptually flawed) solution that was agreed upon, but also complicate the design of general support for declarative config objects, without carrying over any value.

central/authprovider/service/service_impl.go
```diff
@@ -241,6 +254,12 @@ func (s *serviceImpl) DeleteAuthProvider(ctx context.Context, request *v1.Resour
 	if authProvider == nil {
 		return nil, errors.Wrapf(errox.NotFound, "auth provider %q not found", request.GetId())
 	}
+	// Only delete unmodifiable auth provider if force=true.
+	if authProvider.StorageView().GetUnmodifiable() && !request.GetForce() {
+		return nil, errox.InvalidArgs.Newf("auth provider %q is unmodifiable. "+
+			"If you are sure you need to delete this auth provider,"+
+			" repeat request with `force` parameter set up to true.", request.GetId())
```
dhaus67, line 261: nit:
```suggestion
			" repeat request with `force` parameter set to true.", request.GetId())
```

dhaus67 (changes requested):
Since you've updated the proto message for the delete requests, the tests will now fail for this PR (see here for the [gke-qa-e2e-tests](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2610/pull-ci-stackrox-stackrox-master-gke-qa-e2e-tests/1555142179905605632)).

The fix itself would be to update the [deleteAuthProvider](https://github.com/stackrox/stackrox/blob/master/qa-tests-backend/src/main/groovy/services/AuthProviderService.groovy#L61-L63) to use the new message you've introduced.

Blocking the PR by requesting changes for this, as we shouldn't introduce another flake to our CI 🥴

proto/storage/auth_provider.proto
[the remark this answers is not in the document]
rukletsov: Thank you, Malte, for your quick and elaborate response, I appreciate the details.

I spoke to another advocate of declarative configuration, @connorgorman, and indeed, looks like we need to do some thinking on how we marry DB objects with declaratively created ones. With this in mind, I agree that designing an extensible solution will likely become poorly invested time.

With this out of the way, @ivan-degtiarenko let's do the following:
* introduce `storage/proto/traits.proto` with the following (we'll need the same for groups): 
    ```
    // EXPERIMENTAL.
    // 
    // MutabilityMode specifies whether and how an object can be modified. Default
    // is ALLOW and means there are no modification restrictions; this is equivalent
    // to the absence of MutabilityMode specification. ALLOW_FORCED forbids all
    // modifying operations except object removal with force bit on.
    //
    // Be careful when changing the state of this field. For example, modifying an
    // object from ALLOW to ALLOW_FORCED is allowed but will prohibit any further
    // changes to it, including modifying it back to ALLOW.
    enum MutabilityMode {
        ALLOW        = 0;
        ALLOW_FORCED = 1;
    }
    
    // EXPERIMENTAL.
    message Traits {
        MutabilityMode mutability = 1;
    }
    
    // EXPERIMENTAL.
    //
    // Used in combination with MutabilityMode.ALLOW_FORCED to enable forced removal.
    message DeleteByIDWithForce {
        string id  = 1;
        bool force = 2;
    }
    ```
* Mark `traits` field in auth provider as experimental as well, explaining how it works 
* Move the checks to the datastore level
* Cover the functionality with tests, including cases like not allowing to change the mutability mode from `ALLOW_FORCE` to `ALLOW`, forced and non-forced removal, updates to `ALLOW` and `ALLOW_FORCE` objects, etc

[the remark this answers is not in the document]
ivan-degtiarenko: Thank you for your collaborative effort on this, folks.

@rukletsov what is the use case to support `ALLOW_FORCED` mutability mode in groups?

AFAIR, the original idea was for `sso.redhat.com` auth provider groups to be open to modification - otherwise, customer would need to integrate with separate IdP to change default role etc.
SREs `sso.redhat.com` auth provider(if one will exist) will be hidden and also will not benefit from `ALLOW_FORCED` groups.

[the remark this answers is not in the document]
rukletsov: @ivan-degtiarenko if don't do that, users can remove both admin groups and lock themselves out. The idea is the same: all resource objects created during the bootstrap shall be immutable (aka unmodifiable aka ALLOW_FORCE). Any group for this auth provider added on top will not be immutable.

ivan-degtiarenko:
* master
  * **PR #2610** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2610" target="_blank">Graphite</a>  👈
    * **PR #2663** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2663" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2610?utm_source=stack-comment).

proto/storage/auth_provider.proto
```diff
@@ -33,6 +34,8 @@ message AuthProvider {
         string attribute_key   = 1;
         string attribute_value = 2;
     }
-    // EXPERIMENTAL.
     repeated RequiredAttribute required_attributes = 11;
+
+    // See traits.proto for documentation.
```
misberner, line 39: I would suggest to remove this comment, as it doesn't add much value.

ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProviderForm.tsx
```diff
@@ -314,7 +314,9 @@ function AuthProviderForm({
                                             isDisabled={action === 'edit'}
                                             isSmall
                                         >
-                                            {selectedAuthProvider.active
+                                            {selectedAuthProvider.active ||
+                                            selectedAuthProvider.traits?.mutabilityMode ===
+                                                'ALLOW_FORCED'
```
misberner, line 319:
```suggestion
                                            selectedAuthProvider.traits?.mutabilityMode !==
                                                'ALLOW'
```
Please change all the conditions accordingly, it is very to confusing to see the condition in the original form (who says the other option isn't `DONT_ALLOW`?)

proto/storage/traits.proto
```diff
@@ -0,0 +1,36 @@
...
+
+// EXPERIMENTAL.
+message Traits {
+  MutabilityMode mutabilityMode = 1;
+}
+
+// EXPERIMENTAL.
+//
+// Used in combination with MutabilityMode.ALLOW_FORCED to enable forced removal.
+message DeleteByIDWithForce {
```
misberner, line 33: I get that all of this is experimental, but shouldn't this still be in `v1`?

central/authprovider/datastore/datastore_impl.go
```diff
@@ -55,23 +55,37 @@ func (b *datastoreImpl) UpdateAuthProvider(ctx context.Context, authProvider *st
...
+
+func (b *datastoreImpl) verifyExistsAndMutable(ctx context.Context, id string, force bool) error {
+	provider, exists, err := b.storage.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return errox.NotFound.Newf("auth provider with id %q was not found", id)
+	}
+	if provider.GetTraits().GetMutabilityMode() == storage.MutabilityMode_ALLOW_FORCED && !force {
```
misberner, line 87: See my comment below on potential confusion arising from how the checks are performed. Here it's a bit less of an issue because it's directly related to the value of `force`, but for protecting (semi-)immutable objects I would still say that any check should _always_ start with checking if it is allowed or not (`if ... == storage.MutabilityMode_ALLOW { return nil }`), and then any value that is not `ALLOW` but still allows deletion to be performed under certain circumstances is checked in its separate condition.

ui/apps/platform/src/sagas/authSagas.js
```diff
@@ -282,7 +282,9 @@ function* saveAuthProvider(action) {
             yield call(fetchUsersAttributes);
             yield put(actions.selectAuthProvider({ ...remaining, id: savedAuthProvider.data.id }));
         } else {
-            yield call(AuthService.saveAuthProvider, remaining);
+            if (!remaining.active && !remaining.traits?.mutabilityMode === 'ALLOW_FORCED') {
```
dhaus67, line 285:
```suggestion
            if (!remaining.active && !(remaining.traits?.mutabilityMode === 'ALLOW_FORCED')) {
```

This currently leads to failing UI e2e tests ([see here](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2610/pull-ci-stackrox-stackrox-master-gke-ui-e2e-tests/1557292002603700224)), which try to update an auth provider whilst not having set the `traits` property at all.

dhaus67 (changes requested):
With the current changes, the UI e2e tests are failing ([specifically the tests that edit an auth provider](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2610/pull-ci-stackrox-stackrox-master-gke-ui-e2e-tests/1557292002603700224)).

Besides that, I don't have any additional comments apart from the already mentioned ones by Malte, so once the E2E failure is fixed this should be good to go from my side 🚀

proto/storage/traits.proto
dhaus67: Yup, I agree. Moved to v1 to common, since we'll need this for the groups APIs as well.

proto/api/v1/common.proto
```diff
@@ -8,3 +8,11 @@ package v1;
 message ResourceByID {
     string id = 1;
 }
+
+// EXPERIMENTAL.
+//
+// Used in combination with MutabilityMode.ALLOW_FORCED to enable forced removal.
+message DeleteByIDWithForce {
```
misberner, line 15: I was thinking that maybe this should be generalized to `DeleteByIDWithOptions`, as eventually we might want to add new delete options and then naming becomes a mess. However, that then leads to the question what we do if a delete handler doesn't understand any of the options (right now, with force being the only option, we make it very explicit in the service signature what options we accept). Probably fine to keep for now, we can always change the name if we start adding more options.

proto/storage/traits.proto
```diff
@@ -0,0 +1,28 @@
...
+//
+// MutabilityMode specifies whether and how an object can be modified. Default
+// is ALLOW and means there are no modification restrictions; this is equivalent
+// to the absence of MutabilityMode specification. ALLOW_FORCED forbids all
+// modifying operations except object removal with force bit on.
+//
+// Be careful when changing the state of this field. For example, modifying an
+// object from ALLOW to ALLOW_FORCED is allowed but will prohibit any further
+// changes to it, including modifying it back to ALLOW.
+enum MutabilityMode {
```
misberner, line 20: Please move this within `Traits`. Proto enums follow C scoping rules, which means `ALLOW` is visible at the `storage` package's top-level namespace. Additionally or alternatively, it might be good to call this something like `MUTABILITY_ALLOW`, `ALLOW_MUTATE` etc., as even within `Traits` there may be more use cases for `ALLOW`ing something.

misberner (approved):
Looks good overall, but the proto enum naming/namespacing needs addressing before merge.

central/authprovider/datastore/datastore_impl.go
```diff
@@ -55,23 +56,42 @@ func (b *datastoreImpl) UpdateAuthProvider(ctx context.Context, authProvider *st
...
+	}
+
+	if provider.GetTraits().GetMutabilityMode() == storage.Traits_ALLOW_MUTATE {
+		return nil
+	}
+
+	if provider.GetTraits().GetMutabilityMode() == storage.Traits_ALLOW_MUTATE_FORCED && !force {
+		return errox.InvalidArgs.Newf("auth provider %q is immutable", id)
+	}
+	return nil
```
misberner, line 96:
```suggestion
	switch provider.GetTraits().GetMutabilityMode() {
	case storage.Traits_ALLOW_MUTATE:
		return nil

	case storage.Traits_ALLOW_MUTATE_FORCED:
		if force {
			return nil
		}
		// Tricky - ideally, we would return a message here indicating to the user that they need to pass force=true in order to have the operation succeed. However, that could cause confusion for UI users?
	default:
		// Might be a good place for utils.Should
	}
	return errox.InvalidArgs.Newf("auth provider %q is immutable", id)
```
That way, when we add a new mutability mode but forget to handle it here, we conservatively prevent mutations.

misberner (approved):
LGTM except for the handling/enforcement of mutability modes

central/authprovider/datastore/datastore_impl.go
rukletsov: ➕

proto/storage/traits.proto
```diff
@@ -0,0 +1,26 @@
...
+option java_package = "io.stackrox.proto.storage";
+
+package storage;
+
+// EXPERIMENTAL.
+message Traits {
+  // EXPERIMENTAL.
+  //
+  // MutabilityMode specifies whether and how an object can be modified. Default
+  // is ALLOW and means there are no modification restrictions; this is equivalent
```
rukletsov, line 13:
```suggestion
  // is ALLOW_MUTATE and means there are no modification restrictions; this is equivalent
```

proto/storage/traits.proto
```diff
@@ -0,0 +1,26 @@
...
+  // to the absence of MutabilityMode specification. ALLOW_FORCED forbids all
```
rukletsov, line 14:
```suggestion
  // to the absence of MutabilityMode specification. ALLOW_MUTATE_FORCED forbids all
```

proto/storage/traits.proto
```diff
@@ -0,0 +1,26 @@
...
+  // modifying operations except object removal with force bit on.
+  //
+  // Be careful when changing the state of this field. For example, modifying an
+  // object from ALLOW to ALLOW_FORCED is allowed but will prohibit any further
```
rukletsov, line 18:
```suggestion
  // object from ALLOW_MUTATE to ALLOW_MUTATE_FORCED is allowed but will prohibit any further
```

proto/storage/traits.proto
```diff
@@ -0,0 +1,26 @@
...
+  // changes to it, including modifying it back to ALLOW.
```
rukletsov, line 19:
```suggestion
  // changes to it, including modifying it back to ALLOW_MUTATE.
```

proto/storage/traits.proto
```diff
@@ -0,0 +1,26 @@
...
+  enum MutabilityMode {
+    ALLOW_MUTATE        = 0;
+    ALLOW_MUTATE_FORCED = 1;
+  }
+
+  MutabilityMode mutabilityMode = 1;
```
rukletsov, line 25:
```suggestion
  MutabilityMode mutability_mode = 1;
```

rukletsov (approved):
Based on what is agreed upon in https://github.com/stackrox/stackrox/pull/2663#discussion_r945588805, you might need to add `force` option to the `update` call as well.

central/authprovider/datastore/datastore_impl.go
```diff
@@ -86,12 +86,18 @@ func (b *datastoreImpl) verifyExistsAndMutable(ctx context.Context, id string, f
...
 		return nil
+	case storage.Traits_ALLOW_MUTATE_FORCED:
+		if force {
+			return nil
+		}
+		return errox.InvalidArgs.Newf("auth provider %q is immutable and can only be removed"+
+			" via API and specifying the force flag", id)
+	default:
+		utils.Should(errors.Wrapf(errox.InvalidArgs, "unknown mutability mode given: %q",
+			provider.GetTraits().GetMutabilityMode().String()))
```
misberner, line 100:
```suggestion
			provider.GetTraits().GetMutabilityMode()))
```
`%q` and `%s` implicitly convert to `string`, via `fmt.Stringer` or casting to `string` as underlying type, if possible

central/authprovider/service/service_impl.go
```diff
@@ -180,41 +180,43 @@ func (s *serviceImpl) PostAuthProvider(ctx context.Context, request *v1.PostAuth
...
-	provider := s.registry.GetProvider(request.GetId())
+	provider := s.registry.GetProvider(request.GetProvider().GetId())
 	if provider == nil {
-		return nil, errox.NotFound.Newf("auth provider with id %q does not exist", request.GetId())
+		return nil, errox.NotFound.Newf("auth provider with id %q does not exist", request.GetProvider().GetId())
 	}
 
 	// Attempt to merge configs.
-	request.Config = provider.MergeConfigInto(request.GetConfig())
+	request.GetProvider().Config = provider.MergeConfigInto(request.GetProvider().GetConfig())
```
misberner, line 199:
```suggestion
	request.Provider.Config = provider.MergeConfigInto(request.GetProvider().GetConfig())
```

proto/api/v1/authprovider_service.proto
```diff
@@ -111,10 +119,10 @@ service AuthProviderService {
         };
     }
 
-    rpc PutAuthProvider (storage.AuthProvider) returns (storage.AuthProvider) {
+    rpc PutAuthProvider (PutAuthProviderRequest) returns (storage.AuthProvider) {
```
misberner, line 122: Note that this change is _not_ gRPC wire compatible! IMHO this is okay because we do not provide compatibility guarantees, and there are no gRPC customers of this method that we know of. However, in the future we might need to be more careful. One take-away would be that we should maybe officially disallow `storage.` types as input for service methods (i.e., _always_ follow the scheme `rpc MyRPCCall (MyRPCCallRequest) returns (MyRPCCallResponse)`, with an exception for common types such as `DeleteByID` and `Empty`).

However, (and I feel bad to only point this out now, after all those changes, but better late than never) I don't feel it was necessary to add a `force` parameter to the request. My comment https://github.com/stackrox/stackrox/pull/2663#discussion_r945588805 was about the following comment:
```
// We do not allow updates to groups even with the force flag set, hence explicitly setting force to false here.
```
This suggests that `force` is ignored, and it was magically being set to `false`. However, I do not see an issue with omitting `force` from the API. The point is that the mutability enforcement happens at the datastore, and there it would be good to implement a consistent logic. But if the service handler decides to not offer a `force` parameter, that is just an API choice - there is no logic inconsistency in `ALLOW_MUTATE_FORCED` suddenly not meaning what the word says.

I admit that I might have derailed the discussion a bit by questioning the motivation of why allow delete but not mutate _via API_, but my main comment was regarding the implementation at the datastore level, and I don't think it's a given that this needs to impact the API.

misberner (approved):

proto/api/v1/authprovider_service.proto
rukletsov: I don't think we should break API, as we don't do it in other places, e.g., [group service](https://github.com/stackrox/stackrox/pull/2663/files#diff-fe7c26168cea62173cff7ce47d14c3653edf35583ebb3ecd62f3ce4ff2da9f9bR50-R51). We indeed don't provide guarantees but there is some chance that someone's script is broken for no good reason.

dhaus67: Was aware that it wasn't gRPC wire compatible, but I haven't found any references to that specific function and the HTTP API was compatible.

Definitely +1 on forbidding `storage.*` as input.

However, I suppose I'll just revert this then, and stick with ignoring the force flag for update then.

central/authprovider/datastore/datastore_impl.go
```diff
@@ -57,7 +57,8 @@ func (b *datastoreImpl) UpdateAuthProvider(ctx context.Context, authProvider *st
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if err := b.verifyExistsAndMutable(ctx, authProvider.GetId(), force); err != nil {
+	// The API currently does not allow setting the force flag since it would require a non-compatible gRPC change.
```
misberner, line 60: This is datastore code. What the API allows or does not allow is of secondary importance for this code.

I would suggest to rephrase this as something like

```
// Currently, the data store does not support forcing updates. If we want to add a force flag to the respective API methods, we might need to revisit this.
```

misberner (approved):

ui/apps/platform/src/Containers/AccessControl/AuthProviders/AuthProvidersList.tsx
```diff
@@ -67,8 +67,10 @@ function AuthProvidersList({ entityId, authProviders }: AuthProvidersListProps):
                     </Tr>
                 </Thead>
                 <Tbody>
-                    {authProviders.map(({ id, name, type, defaultRole, groups = [] }) => {
+                    {authProviders.map((authProvider) => {
+                        const { id, name, type, defaultRole, traits, groups = [] } = authProvider;
                         const typeLabel = getAuthProviderTypeLabel(type, availableProviderTypes);
+                        const isImmutable = traits?.mutabilityMode !== 'ALLOW_MUTATE';
```
pedrottimark, line 73: @dhaus67 Here is my only question about the final version of this contribution:
* AuthProviderForm.tsx above consistently guards for the possibility that `traits` might have `undefined` value, for example:
     `values.traits && values.traits?.mutabilityMode !== 'ALLOW_MUTATE'`
* AuthProvidersList.tsx here seems double-minded because `?` implies the property might not exist, but `traits?.mutabilityMode !== 'ALLOW_MUTATE'` means `isImmutable` is true `undefined !== 'ALLOW_MUTATE'` in that case. But `traits` property absent means mutable, correct?
* authSagas.js similar `!(remaining.traits?.mutabilityMode !== 'ALLOW_MUTATE')` because `!(undefined !== 'ALLOW_MUTATE')` is `!true` for anonymous mutable condition, correct?
* AuthService seems similar to List above to return without saving if `traits` property is absent.

What do you think about consistent call of a helper function from AuthService.ts?

```ts
export function getIsAuthProviderImmutable(authProvider: AuthProvider) {
    return 'traits' in authProvider && authProvider.traits.mutabilityMode !== 'ALLOW_MUTATE';
}
```

dhaus67: That's a good suggestion, I added the helper function + refactored the existing conditions for the mutability mode.

PR #2623, opened 2022-08-04
Cache go modules on turbo-mode

janisz:
## Description

We can cache go modules to not download them from internet.

## Testing Performed

CI


gavin-stackrox (commented):
I don't think this is necessary because GOMODCACHE=/go/pkg/mod is already in the cache.

janisz:
@gavin-stackrox You're right there is `${{ steps.cache-paths.outputs.GOMODCACHE }}` because cache key is set to hash of go.mod my PR did not use it.

PR #2624, opened 2022-08-04
POC: Sensor resiliency to local crashes and also Central crashes

connorgorman:
## Description

Flow:
- Sensor processes events, sends them to central and writes out entries (id, hash) into a queue
- Every 10s Sensor sends a checkpoint with a uuid in the stream
- Central saturates all queues with this checkpoint and waits for all of them to acknowledge it, effectively guaranteeing everything on the stream has been processed by a pipeline
- Once all queues have acknowledged, central sends the checkpoint back to sensor with the acked uuid
The uuid is used to collect all of the acknowledged events and flush the (id, hash) tuples to pebble
- On restart, Sensor reads all of pebble and populates the (id, hash) deduper
- The caveat is that if the event coming through is a SYNC event, then we send a reconciliation shim to mark that ID as still active instead of the actual object
- Central does normal reconciliation (edited) 

## Checklist

## Testing Performed

POC, but needs rigorous testing on scaled clusters


connorgorman:
@misberner Ignore any weird patterns like the singleton creation, wanted to get a conceptual look before polishing

sensor/common/wal/wal.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (w *walImpl) Insert(id string, hash uint64) error {
+	if w.db == nil {
+		return nil
+	}
+	hashBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(hashBytes, hash)
+	return w.db.Set([]byte(id), hashBytes, &pebble.WriteOptions{
+		Sync: true,
```
misberner, line 50: Why sync? If I am not mistaken, `Insert()` is only ever called when flushing out a batch of elements every 10s. Syncing after every element seems very inefficient and also unneeded. I guess there is _some_ advantage wrt. making sure state remains consistent and you get incremental benefit even if sensor crashes while updating the pebble DB, but.. eh. Doesn't seem to be very meaningful.

That being said, I also don't think pebble is the right tool for the job. You are neither using random inserts nor random lookups. All you need is a batch write and a batch read into memory. Something mimicking [RecordIO](https://news.ycombinator.com/item?id=23589534) might be better suited. Since `id` and the little endian encoding of a uint64 are fixed size, you could even just write out the raw data without any size indicators or delimiters (not saying this is a good idea but it would be simple and way more efficient).

misberner (changes requested):

sensor/common/wal/wal.go
connorgorman: Yeah I definitely think not syncing would be fine here and can instead just batch everything. Would need to look at performance of other options

sensor/common/sensor/central_sender_impl.go
```diff
@@ -72,6 +77,8 @@ func (s *centralSenderImpl) send(stream central.SensorService_CommunicateClient,
 		}
 	}
 
+	checkpointTicker := time.NewTicker(10 * time.Second)
```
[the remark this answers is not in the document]
connorgorman: I think that the state should be properly managed even if we haven't initially synced. These are still fine events to consider

PR #2628, opened 2022-08-04
Prune cluster cves when cluster is removed

md2119:
## Checklist
- ~Documented user facing changes ~

## Testing Performed
Unit Test:

### DIFF
--- a/central/cluster/datastore/datastore.go
+++ b/central/cluster/datastore/datastore.go
@@ -15,6 +15,8 @@ import (
 	clusterHealthStore "github.com/stackrox/rox/central/cluster/store/clusterhealth"
 	clusterHealthPostgresStore "github.com/stackrox/rox/central/cluster/store/clusterhealth/postgres"
 	clusterHealthRocksDBStore "github.com/stackrox/rox/central/cluster/store/clusterhealth/rocksdb"
+	clusterCVEDS "github.com/stackrox/rox/central/cve/cluster/datastore"
+	clusterCVEDataStore "github.com/stackrox/rox/central/cve/cluster/datastore"
 	deploymentDataStore "github.com/stackrox/rox/central/deployment/datastore"
 	namespaceDataStore "github.com/stackrox/rox/central/namespace/datastore"
 	networkBaselineManager "github.com/stackrox/rox/central/networkbaseline/manager"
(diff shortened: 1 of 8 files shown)


central/cluster/datastore/datastore.go
```diff
@@ -293,5 +297,5 @@ func GetTestRocksBleveDataStore(t *testing.T, rocksengine *rocksdbBase.RocksDB,
 		alertStore, namespaceStore, deploymentStore,
 		nodeStore, podStore, secretStore, netFlowStore, netEntityStore,
 		serviceAccountStore, k8sRoleStore, k8sRoleBindingStore, sensorCnxMgr, nil,
-		dacky, clusterRanker, networkBaselineManager)
+		dacky, clusterRanker, networkBaselineManager, nil)
```
md2119, line 300: `nil` because this is only postgres path. The non-postgres path is handled thru cluster cve manager.

central/cluster/datastore/datastore_impl.go
```diff
@@ -545,7 +548,11 @@ func (ds *datastoreImpl) postRemoveCluster(ctx context.Context, cluster *storage
 	ds.removeK8SRoles(ctx, cluster)
 	ds.removeRoleBindings(ctx, cluster)
 
-	// TODO: Apparently, cluster cves are only cleaned up at next cve fetch cycle.
```
md2119, line 548: Actually, the cluster cve removal is handled through the fetcher for non-postgres scenario.

central/cluster/datastore/datastore.go
```diff
@@ -128,21 +131,18 @@ func New(
...
 	} else {
 		ds.searcher = search.New(clusterStorage, indexer, graphProvider, clusterRanker)
 	}
-	ctx := sac.WithGlobalAccessScopeChecker(context.Background(),
-		sac.AllowFixedScopes(
-			sac.AccessModeScopeKeys(storage.Access_READ_ACCESS),
-			sac.ResourceScopeKeys(resources.Cluster)))
-	if err := ds.buildIndex(ctx); err != nil {
+
+	if err := ds.buildIndex(sac.WithAllAccess(context.Background())); err != nil {
```
theencee, line 145: why the change here?

theencee (approved):

central/cluster/datastore/datastore.go
md2119: The internal instantiations are completely controlled by us. For instances like building an index on a startup, we really do not benefit from such instantiations. I remember Viswa, working on an issue in the network graph, and the issue was that deployment perm was not included while building the graph.

PR #2633, opened 2022-08-04
Postgres data migration: hook up

c-du:
## Description

Hookup data migration.
Includes:
1. Add rocksdb and postgres db instances to databases.
2. Conditionally register migration for postgres upgrade 
3. Enable database upgrade.
4. Fix some issues found.

## Checklist

## Testing Performed
Existing CI.
Manually tested data migration and checks the number of objects migrated.


pkg/migrations/seq_num.go
```diff
@@ -1,30 +1,23 @@
...
 // It is a shared constant between central and the migrator binary.
 func CurrentDBVersionSeqNum() int {
-	// If PostgresDatastore is enabled, we consider all migrations to Postgres are following
-	// the current migrations.
-	// if features.PostgresDatastore.Enabled() {
-	//	return internal.CurrentDBVersionSeqNum + internal.PostgresDBVersionPlus
-	// }
+	if features.PostgresDatastore.Enabled() {
+		return internal.CurrentDBVersionSeqNum + internal.PostgresDBVersionPlus
+	}
```
dashrews78, line 14: I'm curious about a scenario. 
CurrentDBVersionsSeqNum = 107
PostgresDBVersionPlus = 56

So we upgrade from Rocks to Postgres.  Migrations will run and get Rocks to level 107.  The migration loop will go until it hits 163.  Starting at 108 it will run the n_X_to_n_Y migrations.  When all that is done the sequence number stored in the postgres version will be 163.  But what happens if we need to add a new migration after Postgres is deployed while still supporting RocksDB?  

The way these numbers work, we would create m_107_m_108 that would handle the Rocks->Rocks migration.  
Then we would have an n_57_to_n_58 migration that would handle Rocks -> Postgres.
Where would the Postgres->Postgres migration fall?  We couldn't do it at 108 because a Postgres upgrade would never see it as Postgres would think it is at 163.  Would we want to do that in the n_57_to_n_58 migration?  I'm curious what that scenario looks like from a numbering scheme and what goes where.

c-du: Yes, what you thought is correct.

After we determine the version to release, new migrations would increase the PostgresDBVersionPlus only. So, this process is going to be three phrases:
1. Now till we are working on the first Postgres release, we increase CurrentDBVersionsSeqNum to add new migrations so we can release them before all legacy->Postgres migration.
2. When we work on the Postgres version to release, we add all migrations that should happen after Postgres migration with PostgresDBVersionPlus.
3. After Postgres version is releases or we are working on the postgres release +1, we merge this two sequence. We do not have to merge the directory names.

dashrews78: Are you saying all Postgres migrations go in the new directory structure?  Is this written down somewhere?  I took a look at your migration design doc and it didn't go in detail about sequence number and directory structure.  I think we should document it some place if it isn't.

pkg/migrations/seq_num.go
```diff
@@ -1,30 +1,23 @@
...
 	return internal.CurrentDBVersionSeqNum
 }
 
 // CurrentDBVersionSeqNumWithoutPostgres is the current DB version number
```
dashrews78, line 18: You are saying this should only be used in a testing environment, but you are using it in the migration.  So we probably need to clean up the comment.

pkg/migrations/seq_num.go
```diff
@@ -1,30 +1,23 @@
...
 // without Postgres migrations. This function should only be used in testing
 // environment.
 func CurrentDBVersionSeqNumWithoutPostgres() int {
-	if !features.PostgresDatastore.Enabled() {
-		utils.Must(errors.New("Unexpected call, ROX_POSTGRES_DATASTORE is not true"))
-	}
-	return internal.CurrentDBVersionSeqNum
+	return internal.CurrentDBVersionSeqNum - 1
```
dashrews78, line 22: Why the -1?

c-du: Because all valid migration numbers are "< CurrentDBVersionSeqNum" and postgres n is the count.

pkg/migrations/seq_num.go
c-du: Yes, I will update the doc.

pkg/migrations/seq_num.go
c-du: Will fix. Thanks!

pkg/migrations/seq_num.go
dashrews78: I get it.  It's fine.  I was thinking about it from a readability and maintenance perspective but I think the only way to get rid of it would be to start the first postgres migration at n_00_to_n01.  That is probably overkill and over thinking it.

migrator/main.go
```diff
@@ -56,26 +57,32 @@ func run() error {
...
-		gormDB, err := postgreshelper.Load(conf)
+	replicaName, replicaPath, err := dbm.GetReplicaToMigrate()
+	if err != nil {
+		return err
+	}
+
+	option.MigratorOptions.DBPathBase = replicaPath
+
+	// Close when needed
+	defer postgreshelper.Close()
```
connorgorman, line 74: hmm why does this need to be here if postgres is disabled?

migrator/postgreshelper/load.go
```diff
@@ -52,14 +52,23 @@ func Load(conf *config.Config) (*gorm.DB, error) {
 				return
 			}
 		}
+		adminConfig.ConnConfig.Database = activeDB
 
 		// Add the active database and password to the source
-		source := fmt.Sprintf("%s password=%s database=%s client_encoding=UTF-8", conf.CentralDB.Source, password, activeDB)
-		source = pgutils.PgxpoolDsnToPgxDsn(source)
+		gormSource := fmt.Sprintf("%s password=%s database=%s client_encoding=UTF-8", conf.CentralDB.Source, adminConfig.ConnConfig.Password, activeDB)
```
connorgorman, line 58: Does it make sense to put the db connection building the source, etc in a package that is just always shared between the migrator and central. We may want to tweak some parameters, but I think it would be nice to only have to change that in one place

migrator/runner/all.go
```diff
@@ -54,4 +54,60 @@ import (
 	_ "github.com/stackrox/rox/migrator/migrations/m_97_to_98_exclude_oauth_sa_kubeadmin_pol"
 	_ "github.com/stackrox/rox/migrator/migrations/m_98_to_m_99_process_alert_comments"
 	_ "github.com/stackrox/rox/migrator/migrations/m_99_to_m_100_violation_report_branding"
+	_ "github.com/stackrox/rox/migrator/migrations/n_01_to_n_02_postgres_clusters"
```
connorgorman, line 57: Recommend adding this to a different package called all_postgres.go to just separate them

pkg/migrations/seq_num.go
connorgorman: Can you add this as a comment to the function?

connorgorman (approved):

migrator/main.go
c-du: Added condition of Postgres flag. Thanks!

migrator/postgreshelper/load.go
c-du: The migrator would manage the database selection. And central may not need Gorm in the future. The need would diverge anyway.

janisz:
@c-du Looks like this change breaks TLS test https://issues.redhat.com/browse/ROX-12307

PR #2637, opened 2022-08-05
ROX-11776: Integrate with NP-Guard including error handling

vikin91:
## Description

This PR:
- Finishes the integration of NP-Guard and roxctl
- Adds basic e2e tests for `roxctl netpol generate`

TODO:
- ~[ ] Use released version of NP-Guard.~ it may be safer to stay on the master commit actually. We can bump the version later in a follow-up if we want to.

## Checklist

There is no official docs for this feature, as this feature is not GA nor tech-preview.

## Testing Performed

(All testing needs to be executed with feature-flag enabled: `export ROX_ROXCTL_NETPOL_GENERATE='true'`)

- [x] Locally `bats tests/roxctl/bats-tests/local/roxctl-netpol-generate-*.bats`
- [x] Bats tests on CI
- [x] Happy path & basic error cases manually: `make cli-darwin && ROX_ROXCTL_NETPOL_GENERATE=true bin/darwin_amd64/roxctl generate netpol tests/roxctl/bats-tests/test-data/np-guard/scenario-minimal-service --strict`


roxctl/generate/README.md
```diff
@@ -0,0 +1,115 @@
+# Static Network Policy Generator
+
+## Developer Preview Notice
+
+The static network policy generation feature is offered as a _developer preview_ feature.
+While we are open to receiving feedback about this feature, our technical support will be able to
```
Maddosaurus, line 6:
```suggestion
While we are open to receiving feedback about this feature, our technical support will not be able to
```

Maddosaurus: There is no support for DP, right?

roxctl/generate/README.md
```diff
@@ -0,0 +1,115 @@
...
+assist and answer questions about it.
+
+## About
+
+The static network policy generator a tool analyzes k8s manifests and generates a set of suggested network policies in form of yaml documents that may be directly applied to a k8s cluster. It is integrated with [NP-Guard](https://np-guard.github.io/) _analyze & synthesize_ module that discovers the network connections and generates the network policies. For more details, refer to the [NP-Guard webpage](https://np-guard.github.io/).
```
Maddosaurus, line 11:
```suggestion
The static network policy generator a tool analyzes k8s manifests and generates a set of suggested network policies in form of yaml documents that may be directly applied to a k8s cluster. It is integrated with NP-Guards _cluster-topology-analyzer_ module that discovers the network connections and generates the network policies. For more details, refer to the [NP-Guard webpage](https://np-guard.github.io/).
```
I think it is called the Cluster Topology Analyzer, but I might be wrong

Maddosaurus: Also, I've removed one of the NP-Guard links to not have the same link within one sentence distance two times

Maddosaurus (changes requested):
Love to see it!
Thanks for adding the Readme, that feels like the right place.
If we clarify the Dev Preview phrasing, I'm happy to merge it, as it looks great.

roxctl/generate/README.md
vikin91: Let me then link `Cluster Topology Analyzer` to its Github repo and keep the second link as it is.

vikin91: See [edfb373](https://github.com/stackrox/stackrox/pull/2637/commits/edfb373fe06d368440609f02b30e24003870cf9a)

roxctl/generate/README.md
vikin91: See [edfb373](https://github.com/stackrox/stackrox/pull/2637/commits/edfb373fe06d368440609f02b30e24003870cf9a)

Maddosaurus (approved):
Lovely! Thanks for the updates :)

vikin91:
We are green then!
Let's wait for @stackrox/merlin to give it a look.

roxctl/generate/netpol/cta.go
```diff
@@ -54,8 +84,7 @@ func (cmd *generateNetpolCommand) generateNetpol() error {
 		return nil
 	}
 
-	cmd.env.Logger().WarnfLn("No mode selected, not providing any output.")
-	return nil
+	return errors.New("no mode selected, not providing any output")
```
janisz, line 87: Is there any test for this change?

roxctl/generate/netpol/npgLogger.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+func newNpgLogger(l logger.Logger) *npgLogger {
+	return &npgLogger{
+		l: l,
+	}
+}
+
+func (nl *npgLogger) Debugf(format string, o ...interface{}) {}
+func (nl *npgLogger) Infof(format string, o ...interface{}) {
```
janisz, line 20: Why only info is supported? I think warn and error are even more interesting

roxctl/generate/netpol/cta.go
```diff
@@ -5,24 +5,54 @@ import (
...
+				roxerr = errNPGErrorsIndicator
+			}
+		}
 	}
+	return roxerr
+}
+
+func (cmd *generateNetpolCommand) ouputNetpols(recommendedNetpols []*v1.NetworkPolicy) error {
 	if _, err := os.Stat(cmd.outputFolderPath); err == nil {
 		if err := os.RemoveAll(cmd.outputFolderPath); err != nil {
```
janisz, line 48: Should we ask user about removing output first? What if by mistake user pass a file or `/` instead of `./`

roxctl/generate/netpol/cta.go
```diff
@@ -5,24 +5,54 @@ import (
...
 			return errors.Wrapf(err, "failed to remove output path %s", cmd.outputFolderPath)
 		}
 		cmd.env.Logger().WarnfLn("Removed output path %s", cmd.outputFolderPath)
 	}
-	cmd.env.Logger().InfofLn("Writing generated Network Policies to %s", cmd.outputFolderPath)
+	if cmd.outputFolderPath != "" {
```
janisz, line 53: If `outputFolderPath` is empty where files will be saved?

roxctl/generate/netpol/netpol.go
```diff
@@ -42,6 +44,8 @@ func Command(cliEnvironment environment.Environment) *cobra.Command {
 			return generateNetpolCmd.generateNetpol()
 		},
 	}
+	c.Flags().BoolVar(&generateNetpolCmd.treatWarningsAsErrors, "strict", false, "treat warnings as errors")
```
janisz, line 47: Cannot add commnet on line 62 but why not using https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive

janisz (changes requested):
Left some comments. My biggest concern is lack of unit tests.

roxctl/generate/netpol/cta.go
vikin91: This corner is actually impossible to reach. I removed this error and made output to stdout a default behavior.

roxctl/generate/netpol/npgLogger.go
vikin91: The library returns all errors and warnings on the output and we print them at the end of the output to make it easier for the user to read.
If we pass a unmuted logger to the library, we will see the errors and warnings twice - as they occur and at the end of the execution.

Do you think the following comment is enough to explain this, or shall I improve it?

```go
// npgLogger wraps the common logging interface in a NP-Guard compatible logger
// it mutes all warnings and errors as they are being returned explicitly by the library
```

roxctl/generate/netpol/cta.go
vikin91: This happens only when the user asks for it explicitly by providing `--remove` flag.
See https://github.com/stackrox/stackrox/blob/edfb373fe06d368440609f02b30e24003870cf9a/roxctl/generate/netpol/netpol.go#L82-L89

roxctl/generate/netpol/netpol.go
vikin91: We used to have it like this but the auto-generated error message was unreadable. I will find it and show an example in a moment.

roxctl/generate/netpol/cta.go
vikin91: This will be stdout. It should be clearer after adding [e23f3e5](https://github.com/stackrox/stackrox/pull/2637/commits/e23f3e58b451eafda654c03e4f1f71aa7365f07b)

roxctl/generate/netpol/cta.go
vikin91: Solved with  [e23f3e5](https://github.com/stackrox/stackrox/pull/2637/commits/e23f3e58b451eafda654c03e4f1f71aa7365f07b)

roxctl/generate/netpol/netpol.go
vikin91: Here: https://github.com/stackrox/stackrox/pull/2602#pullrequestreview-1063436976

vikin91: Using the `MarkFlagsMutuallyExclusive ` would give us:

```
ERROR:	if any flags in the group [output-dir output-file] are set none of the others can be; [output-dir output-file] were all set
```

which I find really difficult to read.

roxctl/generate/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Developer Preview Notice
+
+The static network policy generation feature is offered as a _developer preview_ feature.
+While we are open to receiving feedback about this feature, our technical support will not be able to
+assist and answer questions about it.
+
+## About
+
+The static network policy generator a tool analyzes k8s manifests and generates a set of suggested network policies in form of yaml documents that may be directly applied to a k8s cluster. It is integrated with [NP-Guard's Cluster Topology Analyzer](https://github.com/np-guard/cluster-topology-analyzer) component, which discovers the network connections and generates the network policies. For more details, refer to the [NP-Guard webpage](https://np-guard.github.io/).
```
dhaus67, line 11:
```suggestion
The static network policy generator is a tool that analyzes k8s manifests and generates a set of suggested network policies in form of yaml documents that may be directly applied to a k8s cluster. It is integrated with [NP-Guard's Cluster Topology Analyzer](https://github.com/np-guard/cluster-topology-analyzer) component, which discovers the network connections and generates the network policies. For more details, refer to the [NP-Guard webpage](https://np-guard.github.io/).
```

roxctl/generate/netpol/cta.go
```diff
@@ -1,28 +1,59 @@
...
 	"strings"
 
-	"github.com/np-guard/cluster-topology-analyzer/pkg/controller"
+	npguard "github.com/np-guard/cluster-topology-analyzer/pkg/controller"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/pkg/protoconv/networkpolicy"
 	v1 "k8s.io/api/networking/v1"
 )
 
+var errNPGErrorsIndicator = errors.New("there were errors during execution")
```
dhaus67, line 15: Shouldn't there be an additional error for warnings, i.e. `there were warnings during execution`?
Looking at the documentation and also the logic, it's weird that we state there were errors during execution when it is warnings that we fail on with `--strict`.

roxctl/generate/netpol/cta.go
```diff
@@ -1,28 +1,59 @@
...
+func (cmd *generateNetpolCommand) ouputNetpols(recommendedNetpols []*v1.NetworkPolicy) error {
 	if _, err := os.Stat(cmd.outputFolderPath); err == nil {
 		if err := os.RemoveAll(cmd.outputFolderPath); err != nil {
 			return errors.Wrapf(err, "failed to remove output path %s", cmd.outputFolderPath)
 		}
 		cmd.env.Logger().WarnfLn("Removed output path %s", cmd.outputFolderPath)
 	}
-	cmd.env.Logger().InfofLn("Writing generated Network Policies to %s", cmd.outputFolderPath)
+	if cmd.outputFolderPath != "" {
+		cmd.env.Logger().InfofLn("Writing generated Network Policies to %s", cmd.outputFolderPath)
```
dhaus67, line 55: nit:
```suggestion
		cmd.env.Logger().InfofLn("Writing generated Network Policies to %q", cmd.outputFolderPath)
```

roxctl/generate/netpol/cta.go
```diff
@@ -1,28 +1,59 @@
...
+
 func (cmd *generateNetpolCommand) generateNetpol() error {
-	recommendedNetpols, err := controller.PoliciesFromFolderPath(cmd.folderPath)
+	opts := []npguard.PoliciesSynthesizerOption{
+		npguard.WithLogger(newNpgLogger(cmd.env.Logger())),
+	}
+	if cmd.stopOnFirstError {
+		opts = append(opts, npguard.WithStopOnError())
+	}
+	npSynth := npguard.NewPoliciesSynthesizer(opts...)
```
dhaus67, line 24: Maybe it would be worthwhile having this within the struct `generateNetpolCommand` and construct it with the options (i.e. adding the logger + stop on first error) within `generateNetpolCommand.construct()`.

This way you can inject a fake npguard synthesizer if we decide to add unit tests and it also follows the way we do depedency injection (i.e. service clients for gRPC APIs) in other commands.

roxctl/generate/netpol/npgLogger.go
```diff
@@ -0,0 +1,24 @@
+package netpol
+
+import (
+	"github.com/stackrox/rox/roxctl/common/logger"
+)
+
+// npgLogger wraps the common logging interface in a NP-Guard compatible logger
+// it mutes all warnings and errors as they are being returned explicitly by the library
```
dhaus67, line 8: nit:
```suggestion
// npgLogger wraps the common logging interface in a NP-Guard compatible logger.
// It mutes all warnings and errors as they are being returned explicitly by the library.
```

roxctl/generate/netpol/cta.go
vikin91: Done in [50f3f50](https://github.com/stackrox/stackrox/pull/2637/commits/50f3f508c0d3c5aef79f70788b521056da4feb41)

roxctl/generate/netpol/cta.go
vikin91: Done in [3ae9aad](https://github.com/stackrox/stackrox/pull/2637/commits/3ae9aadf73c9b5e8e039e1737b2f9beafde03475)

dhaus67 (approved):
I agree with Tomek that unit tests should be added.

I'm inclined to approve this request, seeing that we have at least some bats tests for this and with the context that this is only on a "best effort" basis at the moment.

vikin91:
@janisz @dhaus67 I added unit tests in [dcb76b5](https://github.com/stackrox/stackrox/pull/2637/commits/dcb76b53301defa61767cda4a71f191b75509d16)

cc: @Maddosaurus in case you also would like to give it a look

roxctl/generate/netpol/npgLogger.go
janisz: My bad, did not read carefully.

janisz (approved):

PR #2638, opened 2022-08-05
Fix CVE id parsing for VM updates

vjwilson:
## Description

The new split CVE types--IMAGE_VULNERABILITY, NODE_VULNERABILITY, and CLUSTER_VULNERABILITY--now use a separate `id` field just like the other entity types.

Without this change when both the update and postgres feature flags are on, trying to visit an individual CVE page crashes the main area.

## Checklist
- [x] Investigated and inspected CI test results (CI failure is a flake, fixed in a different PR)

## Testing Performed

Screen Shot 2022-08-04 at 7 26 45 PM
Screen Shot 2022-08-04 at 7 26 59 PM
Screen Shot 2022-08-04 at 7 27 11 PM


sachaudh (approved):
LGTM 🥇

PR #2644, opened 2022-08-05
Use GetByQuery in Postgres stores

connorgorman:
## Description

Migrate over the SearchRawXYZ to GetByQuery with the exceptions below:

Not needed:
- SearchRawCVEs

Probably don't want due to caches:
- SearchRawDeployments
- SearchRawPods

Probably don't want due to complexities and nested nature of objects:
- SearchRawImages
- SearchRawNodes

SAC constraints:
- SearchRawProcessBaselines

## Checklist

## Testing Performed

CI should check for correctness

### DIFF
--- a/central/rbac/k8srolebinding/search/searcher_impl.go
+++ b/central/rbac/k8srolebinding/search/searcher_impl.go
@@ -55,7 +55,6 @@ func (ds *searcherImpl) Count(ctx context.Context, q *v1.Query) (int, error) {
 // SearchRawRoleBindings returns the rolebindings that match the query.
 func (ds *searcherImpl) SearchRawRoleBindings(ctx context.Context, q *v1.Query) ([]*storage.K8SRoleBinding, error) {
 	if env.PostgresDatastoreEnabled.BooleanSetting() {
-		log.Infof("Rolebindings search: %+v", q)
 		return ds.storage.GetByQuery(ctx, q)
 	}
 	bindings, _, err := ds.searchRoleBindings(ctx, q)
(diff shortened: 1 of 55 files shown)


central/rbac/k8srolebinding/search/searcher_impl.go
```diff
@@ -54,6 +54,10 @@ func (ds *searcherImpl) Count(ctx context.Context, q *v1.Query) (int, error) {
 
 // SearchRawRoleBindings returns the rolebindings that match the query.
 func (ds *searcherImpl) SearchRawRoleBindings(ctx context.Context, q *v1.Query) ([]*storage.K8SRoleBinding, error) {
+	if features.PostgresDatastore.Enabled() {
+		log.Infof("Rolebindings search: %+v", q)
```
md2119, line 58: why?

pkg/search/postgres/common.go
```diff
@@ -671,7 +671,7 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
 		return nil, err
 	}
 	if query == nil {
-		return nil, emptyQueryErr
+		return nil, nil
```
md2119, line 674: Should we log error on dev builds? Getting a nil query itself is a correctness problem in calling code.

connorgorman: Actually we do this in global search on purpose

central/rbac/k8srolebinding/search/searcher_impl.go
connorgorman: will remove

connorgorman:
@md2119 I rebased this, will also run scale test to get a feel for how much faster

PR #2648, opened 2022-08-07
Dependency hygiene: reduce transitive imports of gRPC client-related code

misberner:
## Description

This PR aims to reduce the set of transitive dependencies required for using gRPC client functionality (`pkg/clientconn`). It does so by:
- moving client-side logic for gRPC authentication methods into a package separate from the server-side identity extraction logic
- move the `CertInfo` type from `pkg/grpc/requestinfo` to `pkg/mtls`, as `requestinfo` contains a lot of server-side code.

The underlying motivation is an attempt to break the circular rox <-> scanner dependency by moving out shared code into its own module. This PR is a step in that direction.

## Checklist

## Testing Performed

Adjusted unit tests. No functional changes.


pkg/grpc/authn/servicecerttoken/token.go
```diff
@@ -12,7 +10,7 @@ import (
 	ctTLS "github.com/google/certificate-transparency-go/tls"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/generated/internalapi/central"
-	"github.com/stackrox/rox/pkg/cryptoutils"
+	"github.com/stackrox/rox/pkg/grpc/client/authn/servicecerttoken"
```
rukletsov, line 13: What is the benefit of splitting `servicecerttoken` into two packages? You still include the package and having `parseX()` and `createX()` in various files is suboptimal.

misberner: I'm not sure what you mean by "you still include the [which?] package". The problem is the server-side stuff, as that pulls in `pkg/sac`, which then pulls in `pkg/search` and a whole lot of other things. Hence the server-side importing the client stuff is fine, the other way round is the problem (scanner uses `pkg/clientconn` to establish authenticated connections to central, but none of the server-side identity extraction stuff).

I get the point about `parse` and `create` being in different files being suboptimal; I could probably change this because `parse...` does not require any complex dependencies. But I don't like having `Parse...` in a `client` package. Maybe I should add a `pkg/grpc/common/authn/servicecerttoken` package?

rukletsov: "the package" is  the newly created `pkg/grpc/client/authn/servicecerttoken`, which you include here. What in `pkg/grpc/authn/servicecerttoken/` requires `pkg/sac`? You only leave `extractor.go`, which seems lightweight in terms of dependencies. Can we _move_ `authn/servicecerttoken` to `pkg/grpc/common` as is without splitting it into two packages?

misberner: Well, `extractor.go` might _seem_ lightweight, but it does import `pkg/grpc/authn`, which in turns import `pkg/auth/authproviders`, which for the registry implementation (temporary privilege elevation for looking up auth providers) imports `pkg/sac`. I've tried to break the dependency chain in other places in several attempts, and all of those failed.

That obviously doesn't mean that you cannot break the chain in other places in a way that makes sense, but philosophically, I think separating client-side and server-side makes a lot of sense to me because of the strict client-server dichotomy. I understand that it would be nice to keep the symmetrical `create` and `parse` operations co-located, but that doesn't help with the dep woes.

rukletsov: > of the strict client-server dichotomy.

Maybe trichotomy will help here. Most from `pkg/grpc/authn/servicecerttoken` goes into `pkg/grpc/common/authn/servicecerttoken`, and `extractor.go` into `pkg/grpc/server/authn/servicecerttoken`. Not sure how much extra work this will mean especially if you want to re-sort everything under `pkg/grpc/` into 3 buckets.

rukletsov (approved):

pkg/grpc/authn/servicecerttoken/token.go
misberner: Yeah, I think that's the best path forward for now.

PR #2649, opened 2022-08-07
Update clean-autogen-protos.py script

misberner:
## Description

The current version of the script makes it somewhat hard to generalize the logic from proper extensions to suffixes (e.g., `_grpc.pb.go`). This is however necessary for some upcoming protobuf generation changes.

Additionally, the `find_in_list` logic to reconstruct the generated files was prone to false positives, since it did a pure prefix match (i.e., `foo_bar.pb.go` would not be removed if there was a `foo.proto`).

Refactor and simplify the logic to work on a file-by-file basis, determining the expected proto file for each individual generated file, rather than working on whole sets of stripped filenames with the above imprecisions.

## Checklist

## Testing Performed

No functional changes, tested manually by re-generating protos.

### DIFF
--- a/tools/clean_autogen_protos.py
+++ b/tools/clean_autogen_protos.py
@@ -4,33 +4,22 @@ import argparse
 import pathlib
 
 
-GENERATED_EXTENSIONS = ["pb.go", "pb.gw.go", "swagger.json"]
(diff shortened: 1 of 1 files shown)


Maddosaurus (approved):
LGTM!

PR #2656, opened 2022-08-08
ROX-12106: Replace FormMultiSelect with ToggleGroup for Type field in image integration forms

pedrottimark:
## Description

Prerequisite to support robot accounts as credentials for Quay image integrations.

### Goal

If a new integration starts with a valid type (for example, **Registry**) instead of the current invalid empty state, then it takes only one change to select a different desired type (for example, **Scanner** or **Registry + Scanner**). In that case, there is only one change in the credentials area of the form.

### Overview

```ts
export type ImageIntegrationCategory = 'REGISTRY' | 'SCANNER' | 'NODE_SCANNER';
```

* Clairify: `categories: ('NODE_SCANNER' | 'SCANNER')[]`
* Google: `categories: ('REGISTRY' | 'SCANNER')[]`
* Quay: `categories: ('REGISTRY' | 'SCANNER')[]`

Altough toggle group which has 3 mutually exclusive items is more intuitive for user, it requires more careful implementation than multi-select list.

Match `categories` array in form state compared to:
* array which corresponds to each item according to **deep equality** comparison
* array which corresponds to item with both categories in **either order** like `['REGISTRY', 'SCANNER']` or `['SCANNER', 'REGISTRY']`

Dave, critique is especially welcome whether or not the TypeScript is fluent.

### Changed files

1. Edit cypress/helpers/formHelpers.js
    * Export `getToggleGroupItem` function

2. Edit cypress/integration/integrations/imageIntegrations.test.js
    * Replace `getSelectButtonByLabel` and `getSelectOption` with `getToggleGroupItem` calls.
    * Change assumption about default from empty categories to initially selected category.

3. Move src/Components/FormMultiSelect.tsx to PatternFly folder to be more visible for possible future use.

4. Edit 3 files in src/Containers/Integrations/IntegrationForm/Forms/QuayIntegrationForm.tsx folder:
    * Use `validCategories` in validation schema.
    * Provide a valid `categories` value in `defaultValues` object
    * Replace `FormMultiSelect` element with `ToggleGroup` element

5. Edit src/Containers/Integrations/utils/integrationUtils.ts
    * Export `categoriesUtilsForClairifyScanner` and `categoriesUtilsForRegistryScanner` objects.

6. Edit style files so disabled toggle group item matches disabled input element.
    * acs.css: Add style rule so background color becomes consistent and text has more readable contrast.
    * trumps.css: Override classic half opacity.

### Residue

1. Verify intuitive initially selected default category, or categories.
2. Import `getCategoriesText` utility function for **Type** column in integrations tables:
    * StackRox Scanner
    * Google Container Registry
    * Red Hat Quay.io
3. Rewrite `getToggleGroupItem` helper function to use `for` attribute of label and `id` attribute of toggle group?

## Checklist

### manual testing

1. `yarn lint` in ui
2. `yarn build` in ui
3. `yarn start` in ui

    * Visit **StackRox Scanner** table

    * Click link to see read-only page for existing integration
        clairify-view

    * Click **Edit** to see form, and then click **Cancel**
        clairify-form

    * Create **New integration** to see form with default values, and then click **Cancel**
        clairify-new

    * Visit **Red Hat Quay.io** table

    * Click link to see read-only page for existing integration
        quay-view

    * Click **Edit** to see form, and then click **Cancel**
        quay-form

    * Create **New integration** to see form with default values, click a different value, and then click **Cancel**
        quay-new

    * Visit **Google Container Registry** table

    * Click **New integration** to see form with default values, and then click **Cancel**
        google-new

### integration testing

1. `yarn cypress-open` in ui/apps/platform
    * integrations/imageIntegrations.test.js


ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
```diff
@@ -74,3 +76,70 @@ const getTimes = () => {
 };
 
 export const timesOfDay = getTimes();
+
+// Utilities for image integrations which can have either or both of two categories.
+
+// Categories alternatives correspond to mutually exclusive toggle group items.
+type CategoriesAlternatives<
+    C0 extends ImageIntegrationCategory,
+    C1 extends ImageIntegrationCategory
```
vjwilson, line 85: I remember when variables could only be two letters long. 👴 

May be better to use more descriptive type names.

vjwilson (commented):
Looks OK, other than the terse var. names.

I'll leave the TS to Dave.

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
```diff
@@ -74,3 +76,70 @@ const getTimes = () => {
...
+// Utilities for image integrations which can have either or both of two categories.
+
+// Categories alternatives correspond to mutually exclusive toggle group items.
+type CategoriesAlternatives<
+    Category0 extends ImageIntegrationCategory,
+    Category1 extends ImageIntegrationCategory
+> = [
+    [[category0: Category0]],
+    [[category1: Category1]],
+    // The alternative for both categories includes both orders.
```
dvail, line 89: Aha! My initial reaction to this was why not use `ImageIntegrationCategory` directly, instead of creating two generic parameters, but this allows us to ensure that the third "alternatives" item contains both orderings at compile time.

It might be overkill, but if we want to ensure that these two items are always different as well we could update the type signatures in this file to:
```typescript
<
    Category0 extends ImageIntegrationCategory,
    Category1 extends Exclude<ImageIntegrationCategory, Category0>
>
```

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
```diff
@@ -74,3 +76,70 @@ const getTimes = () => {
...
+            [category1, category0],
+        ],
+    ];
+
+    // For robust behavior, do not assume that categories from response are limited to Category0 and Category1.
+    /* eslint-disable no-nested-ternary */
+    return {
+        categoriesAlternatives,
+
+        getCategoriesText: (categories: ImageIntegrationCategory[]) =>
```
dvail, line 117: I see the comment about robust behavior above, but is it possible to replace `ImageIntegrationCategory[]` here and below with `(Category0 | Category1)[]`? From what I can tell all usages of this are from client side code which can guarantee the the array will only contain the items passed to `getCategoriesUtils`.

Disregard if I missed another place where this function is called.

dvail (approved):
TypeScript usage looks good to me!

I made a couple of comments regarding type narrowing that may not be necessary, or even applicable.

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
pedrottimark: Super, thank you!

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
pedrottimark: Good eyes. This possible improvement will simmer in my slow cooker until step 2 below.

1. The first scenario is editing an existing integration, in which the starting value is from response.
2. The next pull request (which you cannot have missed) will add a **Type** column to the 3 integration tables. At the moment, TypeScript would report an error because of `categories: ImageIntegrationCategory[]` in `BaseImageIntegration` type. If I can override the property type for each of the 3 derived types, then I will change the argument type of the function, and keep the fallback case, just in case. I verified that TypeScript does not complain about the redundancy.

PR #2660, opened 2022-08-09
Disable store level SAC for network graph

rhybrillou:
## Description

Network graph is the last area where Scoped Access Checks have not been investigated in the postgres context.
As the NetworkEntity and NetworkFlow types are scoped by the namespace-scoped NetworkGraph resource, but do not have search tags on the proto message fields that could provide scoping information, generated SAC code in the store level cannot perform proper scope checks.
In order for the postgres migration to progress, scoped access control for network graphs will be disabled in generated store code and scope checks will be performed in the datastore and service layer, as they used to be done before the postgres migration.

## Checklist

## Testing Performed

CI should be sufficient


janisz (approved):
Please add information why this change

PR #2663, opened 2022-08-09
ROX-11018: Add immutable groups

ivan-degtiarenko:
## Description

This PR brings concept of immutability to `Group` resource. 

The same logic as in https://github.com/stackrox/stackrox/pull/2610 applies.

Current limitation is that UI changes only allow to show immutability of non-default groups. Default group can be immutable, but that will not be visible in the UI. The reason for that is because default group is represented as `defaultRole` in the UI, and changing to `defaultGroup` causes too many cascade changes.
The original use case of https://issues.redhat.com/browse/ROX-11017 suggests that we would only have unmodifiable groups for non-default groups(see discussion with @rukletsov here: https://github.com/stackrox/stackrox/pull/2610#discussion_r938651509). I suggest it's OK to merge the PR in its current state and create a backlog item to display the immutability of default groups in the UI.

Upd: we decided to forbid default immutable groups, at least within this PR.

## Checklist

## Testing Performed

1. Created immutable group
2. Impossible to update
3. Impossible to delete without `force` flag
4. Is greyed out in the UI


ivan-degtiarenko:
* master
  * **PR #2610** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2610" target="_blank">Graphite</a> 
    * **PR #2663** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2663" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2663?utm_source=stack-comment).

central/group/datastore/datastore_impl.go
```diff
@@ -174,6 +174,10 @@ func (ds *dataStoreImpl) Mutate(ctx context.Context, remove, update, add []*stor
 			// Use the id of the retrieved group to delete
 			propsID = id
 		}
+		err := ds.validateMutableGroupID(ctx, propsID, force)
```
dhaus67, line 177: I'd say it's time to create a function similar to `validateAndPrepGroupForUpdateNoLock`, which will handle validation of group, retrieval of ID etc. for the delete case.

central/group/datastore/datastore_impl.go
```diff
@@ -213,17 +217,22 @@ func (ds *dataStoreImpl) Remove(ctx context.Context, props *storage.GroupPropert
 		propsID = id
 	}
 
+	err := ds.validateMutableGroupID(ctx, propsID, force)
```
dhaus67, line 220: Same notion as above, when we create a `validateAndPrepGroupForDeleteNoLock` function we can simplify this here as well.

central/group/datastore/datastore.go
```diff
@@ -18,9 +18,9 @@ type DataStore interface {
 
 	Add(ctx context.Context, group *storage.Group) error
 	Update(ctx context.Context, group *storage.Group) error
-	Mutate(ctx context.Context, remove, update, add []*storage.Group) error
-	Remove(ctx context.Context, props *storage.GroupProperties) error
-	RemoveAllWithAuthProviderID(ctx context.Context, authProviderID string) error
+	Mutate(ctx context.Context, force bool, remove, update, add []*storage.Group) error
+	Remove(ctx context.Context, force bool, props *storage.GroupProperties) error
+	RemoveAllWithAuthProviderID(ctx context.Context, deleteReq *storage.DeleteByIDWithForce) error
```
dhaus67, line 23: I'm not sure I like the mixup of the function signatures, where with one we have `force bool, groups...` and a single one using the `storage.DeleteByIDWithForce` as parameter type. 
I'd favor it to be consistent with the signature and change it to 
`RemoveAllWithAuthProviderID(ctx context.Context, force bool, authProviderID string)`.

WDYT?

central/group/datastore/datastore_impl.go
```diff
@@ -378,3 +392,17 @@ func (ds *dataStoreImpl) getDefaultGroupForProps(ctx context.Context, props *sto
 	// 2. Filter for the default group.
 	return ds.getByProps(ctx, &storage.GroupProperties{AuthProviderId: props.GetAuthProviderId()})
 }
+
+func (ds *dataStoreImpl) validateMutableGroupID(ctx context.Context, id string, force bool) error {
```
dhaus67, line 396: We should note in a comment that this function assumes the function call is behind a lock, similar to the `...NoLock` functions we already have within this datastore.

central/group/datastore/datastore_test.go
```diff
@@ -396,11 +399,19 @@ func (s *groupDataStoreTestSuite) TestAllowsMutate() {
 		Key:            "123",
 		Value:          "123",
 	}, RoleName: "123"}
-	err = s.dataStore.Mutate(s.hasWriteAccessCtx, []*storage.Group{groupWithID}, []*storage.Group{groupWithID},
+	err = s.dataStore.Mutate(s.hasWriteAccessCtx, false, []*storage.Group{groupWithID}, []*storage.Group{groupWithID},
 		[]*storage.Group{grp})
 	s.NoError(err, "expected no error trying to write with Access permissions")
 }
 
+func fixtureGroupWithMutability(mode storage.MutabilityMode) *storage.Group {
```
dhaus67, line 407: I don't think we should keep this function within here, but instead add it to `pkg/fixtures/group.go`.
This way, in case other tests will require this for testing purposes, we have it in a shared place.

central/group/datastore/datastore_test.go
```diff
@@ -699,16 +719,110 @@ func (s *groupDataStoreTestSuite) TestValidateGroup() {
...
+	s.storage.EXPECT().Get(gomock.Any(), gomock.Any()).Return(fixtureImmutableGroup(), true, nil).Times(1)
+
+	err := s.dataStore.Remove(s.hasWriteCtx, false, &storage.GroupProperties{
+		Id:             "id",
+		AuthProviderId: "apid",
+	})
+	s.ErrorIs(err, errox.InvalidArgs)
+}
+
+func fixtureImmutableGroup() *storage.Group {
```
dhaus67, line 789: Same notion as the other unexported fixture function.

central/group/datastore/datastore_test.go
```diff
@@ -699,16 +719,110 @@ func (s *groupDataStoreTestSuite) TestValidateGroup() {
...
+			Traits: &storage.Traits{
+				MutabilityMode: storage.MutabilityMode_ALLOW_FORCED,
+			},
+		},
+	}
+	err := s.dataStore.Update(s.hasWriteCtx, group)
+	s.ErrorIs(err, errox.InvalidArgs)
+	err = s.dataStore.Add(s.hasWriteCtx, group)
+	s.ErrorIs(err, errox.InvalidArgs)
+}
```
dhaus67, line 828: For the mutable / immutable tests, could you also add some more tests for the `Mutate` call, to check that this also works as expected?

ui/apps/platform/src/Containers/AccessControl/AuthProviders/RuleGroups.tsx
```diff
@@ -57,6 +57,10 @@ function RuleGroups({
 }: RuleGroupsProps): ReactElement {
     const augmentedRuleKeys = getAugmentedRuleKeys(ruleAttributes, groups);
 
+    function isDisabled(group: Group) {
+        return disabled || group?.props?.traits?.mutabilityMode === 'ALLOW_FORCED';
```
dhaus67, line 61: Same notion as from @misberner's [comment within the previous PR](https://github.com/stackrox/stackrox/pull/2610#discussion_r942385111).

proto/api/v1/group_service.proto
```diff
@@ -43,6 +43,17 @@ message GroupBatchUpdateRequest {
...
+    bool force                             = 3;
+}
+
+message DeleteGroupRequest {
+    // We copy over parameters from storage.GroupProperties for seamless HTTP API migration.
+    string auth_provider_id = 1;
+    string key              = 2;
+    string value            = 3;
+    string id               = 4;
+    bool force              = 5;
```
dhaus67, line 56: super-nit:
```suggestion

    bool force              = 5;
```

proto/storage/group.proto
```diff
@@ -26,6 +27,7 @@ message Group {
 message GroupProperties {
```
dhaus67, line 27: Add a hint about modifying `GroupProperties` may require changes to `DeleteGroupRequest`.

pkg/fixtures/group.go
```diff
@@ -15,7 +15,17 @@ func GetGroup() *storage.Group {
 	}
 }
 
-// GetGroups returns a set of mock storage.Group objects, which in total represents the possible combinations of group properties and roles
+// GetGroupWithMutability returns a mock storage.Group with all possible properties filled out.
+func GetGroupWithMutability(mode storage.MutabilityMode) *storage.Group {
+	group := GetGroup()
+
+	group.GetProps().Traits = &storage.Traits{MutabilityMode: mode}
```
misberner, line 22:
```suggestion
	group.Props.Traits = &storage.Traits{MutabilityMode: mode}
```
`GetProps()` only makes sense if we think that `group` could be `nil`. However, in this case, `nil` props would be returned and the `.Traits` access would cause a panic.

central/group/datastore/datastore.go
```diff
@@ -18,9 +18,9 @@ type DataStore interface {
 
 	Add(ctx context.Context, group *storage.Group) error
 	Update(ctx context.Context, group *storage.Group) error
-	Mutate(ctx context.Context, remove, update, add []*storage.Group) error
-	Remove(ctx context.Context, props *storage.GroupProperties) error
-	RemoveAllWithAuthProviderID(ctx context.Context, authProviderID string) error
+	Mutate(ctx context.Context, force bool, remove, update, add []*storage.Group) error
```
misberner, line 21: nit but I would find it more logical if `force` were the last parameter. I think that's also the ordering I have seen in other methods, such as the helper methods introduced in this PR

central/group/datastore/datastore_impl.go
```diff
@@ -148,7 +148,8 @@ func (ds *dataStoreImpl) Mutate(ctx context.Context, remove, update, add []*stor
 	}
 
 	for _, group := range update {
-		if err := ds.validateAndPrepGroupForUpdateNoLock(ctx, group); err != nil {
+		// We do not allow updates to groups even with the force flag set, hence explicitly setting force to false here.
```
misberner, line 151: This seems surprising to me:
1. Why? Where is this documented? Should the respective traits enum then not be called `ALLOW_FORCE_DELETE_ONLY`?
2. Why does `validateAndPrepGroupForUpdateNoLock` even take a force parameter if it is never set to `true`?

central/group/datastore/datastore_impl.go
```diff
@@ -378,3 +394,24 @@ func (ds *dataStoreImpl) getDefaultGroupForProps(ctx context.Context, props *sto
 	// 2. Filter for the default group.
 	return ds.getByProps(ctx, &storage.GroupProperties{AuthProviderId: props.GetAuthProviderId()})
 }
+
+// validateMutableGroupID validates whether a group allows changes or not based on the mutability mode set.
+// NOTE: This function assumes that the call to this function is already behind a lock.
```
misberner, line 399: call it `...NoLock`, then?

central/group/datastore/datastore_impl.go
```diff
@@ -185,45 +179,33 @@ func (ds *dataStoreImpl) Mutate(ctx context.Context, remove, update, add []*stor
...
+	if err != nil {
+		return err
 	}
 
-	return ds.storage.Delete(ctx, propsID)
+	return ds.storage.Delete(ctx, groupID)
 }
 
-func (ds *dataStoreImpl) RemoveAllWithAuthProviderID(ctx context.Context, authProviderID string) error {
+func (ds *dataStoreImpl) RemoveAllWithAuthProviderID(ctx context.Context, force bool, id string) error {
```
misberner, line 201: Why the renaming `authProviderID => id`? I found `authProviderID` helpful as it's more explicit and `id` in the group datastore setting sounds like group ID

misberner (changes requested):

central/group/datastore/datastore_impl.go
dhaus67: I've removed the force parameter for `validateAndPrepGroupForUpdateNoLock`.

Based on what I got from [the conversations](https://github.com/stackrox/stackrox/pull/2610#discussion_r938571896) it seems we have agreed to not support the update for groups for now, simply due to the reason that we don't necessarily have a use-case. If people wish to desire to change the group, they would have to re-create it at the moment. 

Maybe @rukletsov can also share some additional thoughts on this.

misberner: What would be the use case for force delete in the first place, though? It would seem that making an API call with `ALLOW_FORCE` voids your warranty (i.e., you might soft-brick your instance), so calling it a "use case" seems a bit of a stretch to me in the first place. If we talk about the need to fix things with a broken instance, it seems illogical that we would require a delete and recreate workflow given that "delete" is already nothing that should be exposed to a user.

dhaus67: Personally, I don't mind adding the force flag also to the update API, but I know there was something agreed beforehand to not do it (and the reasoning for that was that there's no real _why_ to do this).

I'll add it for now, and once @rukletsov has shared some of the thoughts we can still decide to do it / not do it.

central/group/datastore/datastore_impl.go
```diff
@@ -378,3 +393,24 @@ func (ds *dataStoreImpl) getDefaultGroupForProps(ctx context.Context, props *sto
...
+func (ds *dataStoreImpl) validateMutableGroupIDNoLock(ctx context.Context, id string, force bool) error {
+	group, exists, err := ds.storage.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return errox.NotFound.Newf("group with id %q was not found", id)
+	}
+
+	if group.GetProps().GetTraits().GetMutabilityMode() == storage.Traits_ALLOW_MUTATE {
```
misberner, line 408: Please see [my comment on the other PR](https://github.com/stackrox/stackrox/pull/2610#discussion_r946295009) on proper enforcement of mutability modes

misberner (approved):
LGTM except for the handling of mutability modes

central/group/datastore/datastore_impl.go
rukletsov: Whatever, folks, we're building a kludge here : ). Indeed, I prefer kludges to feel and look like kludges, with as minimal changes as possible. At the same time, it is reasonable to have at least one option to remove a resource object because we're protecting against Murphy here, not Machiavelli. This is my reasoning for allowing force for delete only.

pedrottimark (approved):
* Good call by Ivan to repeat `isDisabled(group)` instead of factor out, which might have changed indentation and caused confusion in changed lines.
* Because this is a kludge, it seems unnecessary to make completely consistent changes in frontend service functions and request/response types. Good thing, because confusing lack of correspondence between frontend and backend files.

PR #2675, opened 2022-08-10
ROX-11018: remove admin password login option for ACSMS

ivan-degtiarenko:
## Description

This PR removes any mention of the `Basic` auth provider if the `ROX_MANAGED_CENTRAL` env variable is set to `true`.
"Any mention" in this context means:
* There is no `Log in with username/password` option in the UI
* API does not return `Basic` auth providers as part of any response
* API forbids to create/update `Basic` auth providers
It's still possible to login via username/password pair via `https://<central host>/sso/login/4df1b98c-24ed-4073-a9ad-356aec6bb62d`

Additional documents:
- [Design doc](https://docs.google.com/document/d/1Y2xuVpHbsjdNjNbT3ZggkWTnyHR4goIP5vxVpQsDykc/edit)

## Testing Performed
1. Manually tested that:
* There is no `Log in with username/password` option in the UI when ROX_MANAGED_CENTRAL is set.
* API does not return `Basic` auth providers as part of any response.
* API forbids to create/update `Basic` auth providers.


ivan-degtiarenko:
* master
  * **PR #2675** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2675" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2675?utm_source=stack-comment).

dhaus67 (commented):
Just minor comments + one question with regards to the deploy script.

Other than that, LGMT 🚀

deploy/common/k8sbased.sh
```diff
@@ -335,6 +341,10 @@ function launch_central {
         ${ORCH_CMD} -n stackrox set env deploy/central MODULE_LOGLEVELS="${MODULE_LOGLEVELS}"
       fi
 
+      if [["$ROX_MANAGED_CENTRAL" == "true" ]]; then
```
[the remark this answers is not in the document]
ivan-degtiarenko: These two hunks are in different branches of `if-else`

dhaus67 (approved):

misberner (changes requested):
I think the approach of having a wrapper service that suppresses everything about the basic auth provider is suboptimal.

For Put and Post, this should be the behavior already - nobody should be able to create a basic auth provider via API, or modify the existing one. If that is not ensured by the current code, the current code should be changed.

Regarding the listing, I think it would be preferable to have a `visible`, `selectable` etc. that governs whether a provider will show up in the list of login auth providers. In a managed services context, this would just be set to `false` in `central/auth/userpass/auth.go`, circa l. 78.

WDYT?

ivan-degtiarenko:
@misberner I find it an unfortunate practice that for every tweak we need to do in the behaviour of a specific auth provider within the MS context we add a new field(see `required attributes`, `traits`, now `visible`).  I'm not sure that supporting this field(which can be discovered by the customer) is a better long-term bet than creating a one-time wrapper which would have only one customer - us. But that's just me trying to predict the future. 
Also, unfortunately, there is no way to make this filtration on the database level since we don't store `basic` auth providers. So even if we add a `visible` flag, filtration logic goes to the service level.

Saying all that, a solution using `visible` mark is a valid alternative. For the next week, this PR is in @dhaus67 hands, so in the end, it's up to him to decide :)

dhaus67:
@misberner so, we do not currently have anything explicitly checking within the auth provider service for the basic auth type besides listing the available auth provider types.
That's due to the reason that we initialize the basic auth provider [with DoNotStore()](https://github.com/stackrox/stackrox/blob/master/central/auth/userpass/auth.go#L85). When you try to perform any CRUD operations on the existing basic auth provider, you get a not found error ([here's the initial PR that made this change](https://github.com/stackrox/rox/pull/1511)).
Personally, I feel that this is quite confusing (especially since the basic auth provider will be listed via `GET v1/authProviders`) and would expect a different error message (i.e. invalid args not allowed to modify basic auth provider).
When trying to create a new basic auth provider, there's no specific filter on the service level. Instead you will receive an error due to the basic auth manager missing from context.

Since I'm anyways going to make the changes to the service, I'll add the `visible` field to the auth provider. I do agree with Ivan's sentiment that its unfortunate that we add fields specifically for the managed service use-case and would be fine with the wrapper, but then again it could be that in the future we want to filter out other auth providers (talking about it, we probably should not show the OpenShift auth provider type & created instances for managed services 🥴).

I've made the changes and created a follow-up for the OpenShift auth provider topic ([ROX-12182](https://issues.redhat.com/browse/ROX-12182)), let me know what you think.

central/authprovider/service/service_impl.go
```diff
@@ -171,6 +190,10 @@ func (s *serviceImpl) PostAuthProvider(ctx context.Context, request *v1.PostAuth
 	if providerReq.GetLoginUrl() != "" {
 		return nil, errox.InvalidArgs.CausedBy("auth provider loginUrl field is not empty")
 	}
+	if providerReq.GetType() == basic.TypeName {
+		return nil, errox.InvalidArgs.CausedByf("auth provider of type %s can not be created",
```
misberner, line 194:
```suggestion
		return nil, errox.InvalidArgs.CausedByf("auth provider of type %s cannot be created",
```

central/authprovider/service/service_impl.go
```diff
@@ -184,6 +207,10 @@ func (s *serviceImpl) PutAuthProvider(ctx context.Context, request *storage.Auth
 	if request.GetId() == "" {
 		return nil, errox.InvalidArgs.CausedBy("auth provider id is empty")
 	}
+	if request.GetType() == basic.TypeName {
+		return nil, errox.InvalidArgs.CausedByf("auth provider of type %s can not be modified",
```
misberner, line 211:
```suggestion
		return nil, errox.InvalidArgs.CausedByf("auth provider of type %s cannot be modified",
```

proto/storage/auth_provider.proto
```diff
@@ -35,4 +35,14 @@ message AuthProvider {
     }
     // EXPERIMENTAL.
     repeated RequiredAttribute required_attributes = 11;
+
+    // EXPERIMENTAL.
+    // visibility allows to specify whether the auth provider should show up when listing the auth providers
+    // during login.
+    enum Visibility {
+        UNSET                    = 0;
```
misberner, line 43: Do we actually want an `UNSET` value here? Or should `VISIBLE` just be the default (value 0)?

If you keep this, please rename it to `UNSET_VISIBILITY`, `VISIBILITY_UNSET` or the like

proto/storage/auth_provider.proto
```diff
@@ -35,4 +35,14 @@ message AuthProvider {
...
+        VISIBLE                  = 1;
+        NOT_VISIBLE              = 2;
```
misberner, line 45: super nit: I would call it `HIDDEN` to highlight that visible is the default (i.e., hiding it is what requires an active choice)

misberner (changes requested):

proto/storage/auth_provider.proto
dhaus67: Used `VISIBLE` as default now, also removed the `DefaultVisible` option now as, by default, the provider will now be already visible.

misberner (approved):

proto/storage/auth_provider.proto
```diff
@@ -35,4 +35,13 @@ message AuthProvider {
     }
     // EXPERIMENTAL.
     repeated RequiredAttribute required_attributes = 11;
+
+    // EXPERIMENTAL.
+    // visibility allows to specify whether the auth provider should show up when listing the auth providers
+    // during login.
+    enum Visibility {
```
rukletsov, line 42: What do you think about putting this enum into `Traits` message?

rukletsov (approved):

PR #2680, opened 2022-08-10
Re-enable CVE test "Verify IsFixable is correct when scoped (#digest, #fixable)" in postgres mode

rhybrillou:
## Description

Some end-to-end tests were disabled at the early stages of the postgres migration because the state of things at the time made them fail or made them instable.
As the migration progresses, it is time to re-enable those that work again.

For the test at hand, some adjustments had to be introduced in the GraphQL queries to cope with some underlying re-design around vulnerability management.

## Checklist

## Testing Performed

CI run should be sufficient


janisz (approved):

janisz:
Could you add description for this change?

PR #2683, opened 2022-08-10
ROX-11009: Re-enable Vulnerability management SAC end-to-end test "Verify role based scoping on vuln mgmt" in postgres mode

rhybrillou:
## Description

Some end-to-end tests were disabled at the early stages of the postgres migration because the state of things at the time made them fail or made them instable.
As the migration progresses, it is time to re-enable those that work again.

For the test at hand, tow of the variations are easy to isolate and bring to work again. For the last one, due to redesign of the underlying image component and vulnerability management, the test should be re-written to split the retrieval per object class.

## Checklist

## Testing Performed

CI run should be sufficient.

### DIFF
--- a/qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
+++ b/qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
@@ -39,6 +39,34 @@ class VulnMgmtSACTest extends BaseSpecification {
     }
     """
 
+    private static final GET_IMAGE_CVES_QUERY = """
+    query getCves(\$query: String, \$pagination: Pagination)
+    {
+        results: imageVulnerabilities(query: \$query, pagination: \$pagination) {
+            ...cveFields
+            __typename
+        }
+        count: imageVulnerabilityCount(query: \$query)
+    }
+    fragment cveFields on ImageVulnerability {
+        cve
+    }
+    """
+
+    private static final GET_NODE_CVES_QUERY = """
+    query getCves(\$query: String, \$pagination: Pagination)
+    {
+        results: nodeVulnerabilities(query: \$query, pagination: \$pagination) {
+            ...cveFields
+            __typename
+        }
+        count: nodeVulnerabilityCount(query: \$query)
+    }
+    fragment cveFields on NodeVulnerability {
+        cve
+    }
+    """
+
     private static final GET_COMPONENTS_QUERY = """
     query getComponents(\$query: String, \$pagination: Pagination)
     {
@@ -50,6 +78,36 @@ class VulnMgmtSACTest extends BaseSpecification {
     }
 
     fragment componentFields on EmbeddedImageScanComponent {
+        name
+        version
+    }
+    """
+
+    private static final GET_IMAGE_COMPONENTS_QUERY = """
+    query getComponents(\$query: String, \$pagination: Pagination)
+    {
+        results: imageComponents(query: \$query, pagination: \$pagination) {
+            ...componentFields
+            __typename
+        }
+        count: imageComponentCount(query: \$query)
+    }
+    fragment componentFields on ImageComponent {
+        name
+        version
+    }
+    """
+
+    private static final GET_NODE_COMPONENTS_QUERY = """
+    query getComponents(\$query: String, \$pagination: Pagination)
+    {
+        results: nodeComponents(query: \$query, pagination: \$pagination) {
+            ...componentFields
+            __typename
+        }
+        count: nodeComponentCount(query: \$query)
+    }
+    fragment componentFields on NodeComponent {
         name
         version
     }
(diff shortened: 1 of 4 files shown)


qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -55,6 +83,124 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+        }
+        count: nodeComponentCount(query: \$query)
+    }
+    fragment componentFields on NodeComponent {
+        name
+        version
+    }
+    """
+
+    private static final IMAGES_TO_RESCAN = [
```
janisz, line 116: How did you get this list?

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -72,6 +218,11 @@ class VulnMgmtSACTest extends BaseSpecification {
         // where an image is orphaned
         ImageIntegrationService.addStackroxScannerIntegration()
         ImageService.scanImage(CENTOS_IMAGE)
```
janisz, line 220: This could be removed as this image is included in `IMAGES_TO_RESCAN`

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -100,25 +251,57 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+                if ( ! sortedVulns.contains(v) ) {
+                    log.info("Item found in baseVulnCallResult but not in vulnCallResults: " + v)
+                }
+            }
+            for ( v in sortedVulns ) {
+                if ( ! baseSortedVulns.contains(v) ) {
+                    log.info("Item found in vulnCallResults but not in baseVulnCallResult: " + v)
+                }
+            }
+        }
```
janisz, line 304: How about something like this?
```suggestion
        def baseSortedVulns = extractCVEsAndSort(baseVulnCallResult.value).toSet()
        def sortedVulns = extractCVEsAndSort(vulnCallResult.value).toSet()
        if ( baseSortedVulns != sortedVulns ) {
            log.info("Item found in baseVulnCallResult but not in vulnCallResults: " + baseSortedVulns-sortedVulns)
            log.info("Item found in vulnCallResults but not in baseVulnCallResult: " + sortedVulns-baseSortedVulns)
        }
```
Why do we need to log it? If it's not always happening the it should be warning or error

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -130,13 +313,120 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+    def "Verify role based scoping on vuln mgmt: #roleName #baseQuery"() {
+        when:
+        "Get Node CVEs and components"
+        BaseService.useBasicAuth()
+        def gqlService = new GraphQLService()
+        def imageCveQuery = getImageCVEQuery()
+        def imageComponentQuery = getImageComponentQuery()
+        def nodeCveQuery = getNodeCVEQuery()
+        def nodeComponentQuery = getNodeComponentQuery()
+        def imageBaseQuery = isPostgresRun() ? imageBaseQuery : baseQuery
```
janisz, line 387: Looks suspicious. Where `imageBaseQuery` is defined first?

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -130,13 +313,120 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+                    for ( img in tstImages ) {
+                        log.debug "Image ${img} has vulnerability ${v}"
+                    }
+                }
+            }
+            for ( v in sortedVulns ) {
+                if ( ! baseSortedVulns.contains(v) ) {
+                    log.info("Item found in vulnCallResults but not in baseVulnCallResult: " + v)
+                    def imageQuery = RawQuery.newBuilder().setQuery("CVE:${v}").build()
+                    def tstImages = ImageService.getImages(imageQuery)
```
janisz, line 356: It appears 2 times. How about extracting it to a func?
```

    private static List<ImageOuterClass.ListImage> getImagesWithCVE(String q) {
        def imageQuery = RawQuery.newBuilder().setQuery("CVE:${q}").build()
        def tstImages = ImageService.getImages(imageQuery)
        return tstImages
    }
```

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -130,13 +313,120 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+
+        and:
+        gqlService = new GraphQLService(getToken(IMAGE_ROLE, IMAGE_ROLE))
+        def vulnCallResult = gqlService.Call(cveQuery, [query: ""])
+        assert vulnCallResult.hasNoErrors()
+        def componentCallResult = gqlService.Call(componentQuery, [query: ""])
+        assert componentCallResult.hasNoErrors()
+        def baseSortedVulns = extractCVEsAndSort(baseVulnCallResult.value)
+        def sortedVulns = extractCVEsAndSort(vulnCallResult.value)
+        if ( baseSortedVulns != sortedVulns ) {
```
janisz, line 341: If we have sets it will be easier to write (aka need less indents)
```groovy
        baseSortedVulns-sortedVulns.each { String v -> 
            log.info("Item found in baseVulnCallResult but not in vulnCallResults: " + v)
            for ( img in getImagesWithCVE(v) ) {
                log.debug "Image ${img} has vulnerability ${v}"
            }
        }
```

janisz (changes requested):
Small comments. I think it will be easier to read if you add types instead of `def`s but it's up to you. I feel it's really important in functions where IDE not always is able to guess the type and sometimes puts `Object`

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
rhybrillou: I went through the list of previously run test files and looked for what looked like an image name.
The reason why I did this (and left debugging stuff in the test bodies) is that at some point, the image test was failing with one extra image vulnerability when using SAC than when using scoping query `"Image:*"`. I suspect reprocessing or pruning may have left an orphaned ImageCVE behind, hence the re-scan.

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
rhybrillou: Logging as error sounds good.
At the point where the test runs in e2e, there are about 5k vulnerabilities in the storage. Finding out which one is triggering the diff is not something doable manually anymore.

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
janisz: Not related to this PR but do we really need 85 images to do our E2E tests?
cc @gavin-stackrox @msugakov

janisz (approved):

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
msugakov: Not an answer to @janisz question, but a not for @rhybrillou . Please make sure you're aware of multi-platform enablement which soon will change images from `quay.io/rhacs-eng/qa` to `quay.io/rhacs-eng/qa-multi-arch`. https://github.com/stackrox/stackrox/pull/3461

qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -100,25 +250,107 @@ class VulnMgmtSACTest extends BaseSpecification {
...
+            (baseSortedVulns-sortedVulns).each {
+                item ->
+                log.error("Item found in baseVulnCallResult but not in vulnCallResults: " + item.cve)
+                for ( img in getImagesWithCVE(item.cve) ) {
+                    log.error("Vulnerability ${item.cve} is found in image {${img}}")
+                }
+            }
+            (tewakesortedVulnsdArray-baseSortedVulns).each {
```
janisz, line 346: Where it's defined?

PR #2688, opened 2022-08-10
Add Cluster foreign key to k8s objects

connorgorman:
## Description

Adds cluster foreign key so we don't need to worry about pruning when clusters are deleted

@c-du We will need to be very careful on migration here

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


md2119 (changes requested):
Cluster postgres store uses delete+copyFrom for batch operations. The delete will cause all the dependencies to be deleted. To avoid unintended cascade deletes from batch updates, add `no-fk-constraint` to proto or remove `copyFrom` from Cluster

connorgorman:
> Cluster postgres store uses delete+copyFrom for batch operations. The delete will cause all the dependencies to be deleted. To avoid unintended cascade deletes from batch updates, add no-fk-constraint to proto or remove copyFrom from Cluster

@md2119 There are no batch operations for Clusters so this is actually safe

c-du (approved):

md2119:
> > Cluster postgres store uses delete+copyFrom for batch operations. The delete will cause all the dependencies to be deleted. To avoid unintended cascade deletes from batch updates, add no-fk-constraint to proto or remove copyFrom from Cluster
> 
> @md2119 There are no batch operations for Clusters so this is actually safe, but do need to be careful about enabling that in the future

Let me rephrase the statement. This change relies on store layer batch functionality not being exposed. Furthermore, it requires the store's consumers to know the implementation details, which should not be the case. However, there is no check to prohibit that. I recommend we add a flag to generator to skip generating batch functionality.

md2119 (approved):
rebase and merge

PR #2690, opened 2022-08-11
[Hackathon] Enrich images with SBOMs + create policy criteria

dhaus67:
## Description

Hackathon project which adds support for detecting SBOMs stored in OCI registries as well as create a policy criteria that will alert on images that do not have a SBOM associated with it that covers the whole image.

The way the SBOM is stored within the OCI registry and what specific OCI descriptors to expect is defined within [cosign's SBOM spec](https://github.com/sigstore/cosign/blob/main/specs/SBOM_SPEC.md), which was followed during implementation.

The supported SBOM formats are also dependant on the formats outline in the SBOM spec, which currently are [SPDX](https://spdx.org), [CycloneDX](https://cyclonedx.org/) and [syft](https://github.com/anchore/syft).

A SBOM itself _may_ specify a scope, referring to the container image. The scope itself can either cover the whole container image, a specific layer, or a file under a specific path in the flattened container image (for more details, [see here](https://github.com/sigstore/cosign/blob/main/specs/SBOM_SPEC.md#scopes)). The scope itself is stored as an annotation on the descriptor of the OCI artifact.

A image counts as fully covered by a SBOM when:
- the scope on the SBOM(s) is set to `all` explicitly or no scope is set. At this point, its assumed that the SBOM refers to the entire object.
- the scope of the SBOM(s) contains all container image layers.

An image counts as partially covered by a SBOM when:
- the scope of the SBOM(s) contains layer digests. At least one layer digest is that of the container image.
- the scope of the SBOM(s) contains a specific file path.

The policy criteria itself currently allows to alert on images that are either not having a SBOM attached at all or are only partially covered by an SBOM.

What's not been achieved during the hackathon:
- Code is duplicated between `pkg/signatures` and `pkg/sbom`. This should be properly shared.
- Additional test harness is missing, i.e. qa-e2e tests testing SBOMs specifically, testing SBOMs with different scopes.
- The policy criteria is in a hackish state, some parts should be re-worked (i.e. setting a non-zero enum value).
- The SBOM currently is only fetched, no signature verification or attestation is done. Some tools do provide attestation in the form of [in-toto-attestations](https://github.com/in-toto/attestation) (i.e. [syft](https://github.com/anchore/syft#attestation-experimental)), which are also supported by the sigstore ecosystem.
- SBOMs that are scoped to a specific file path within the container image are currently not verified, i.e. its assumed to be partially covered by default.


dhaus67:
* master
  * **PR #2690** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2690" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2690?utm_source=stack-comment).

PR #2705, opened 2022-08-15
[ROX-12200] Add a postgres store generation flag for CopyFrom operation

md2119:
## Description

Add a flag in pg store generator that controls whether or not to use the CopyFrom operation. CopyFrom operation requires explicit record deletion before copying data into DB, hence, we are unable to attach referential constraints to some Postgres stores. This will enable attaching references to `clusters` table (#2688).

## Checklist

## Testing Performed

### DIFF
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -264,11 +264,13 @@ func (s *storeImpl) {{ template "copyFunctionName" $schema }}(ctx context.Contex
 {{- end}}
 
 {{- if not .JoinTable }}
+{{- if not .NoCopyFrom }}
 {{ template "copyObject" .Schema }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


connorgorman (approved):

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -264,11 +264,13 @@ func (s *storeImpl) {{ template "copyFunctionName" $schema }}(ctx context.Contex
...
 
 {{- if not .JoinTable }}
+{{- if not .NoCopyFrom }}
 {{ template "copyObject" .Schema }}
 {{- end }}
+{{- end }}
 
 {{- if not .JoinTable }}
-
+{{- if not .NoCopyFrom }}
```
clickboo, line 273: Can we name the flag CopyFrom? To avoid the double negative: not .NoCopyFrom?

PR #2706, opened 2022-08-15
Fix Observed, Deferred, False Positive CVE links, and Dockerfile

vjwilson:
## Description

Of the fixing of Postgres bugs, there is no end

## Checklist

## Testing Performed

Manual testing 
- CVE links in Observed, Deferred, and False Positive tables no longer crash
- Components show up in Dockerfile display again


sachaudh (approved):
LGTM 🚢

PR #2713, opened 2022-08-16
Add more images to fixtures

connorgorman:
## Description

Adds many quay.io images to the fixtures

## Checklist

## Testing Performed

Will run on scale test


dashrews78 (approved):
nit:  I really hate to say something like this but the rest of them are alphabetical.  I like having quay grouped separately, but It would be nice if they were ordered within the quay grouping.  If not I get it.  Otherwise looks good to me.

PR #2716, opened 2022-08-16
 Count on distinct primary keys instead of all table columns in Postgres

md2119:
## Description

## Checklist
- ~Documented user facing changes ~

## Testing Performed
Unit


connorgorman:
Also, distinct is unfortunately not free so if there are cases when the pks would be unique (not having been joined, etc) then this will incur the distinct cost. Also, I wonder if we should tweak the joins to be left joins

md2119:
> Also, distinct is unfortunately not free so if there are cases when the pks would be unique (not having been joined, etc) then this will incur the distinct cost. Also, I wonder if we should tweak the joins to be left joins

The left join will not give correct results.

connorgorman (approved):
see comment about falling back to *

gavin-stackrox:
NetworkFlowTest.Verify connections from external sources(NetworkFlowTest.groovy:500) started to fail after this PR merged. Do you think it is related? @md2119 

e.g.
https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-openshift-4-qa-e2e-tests/1559977391219019776
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-openshift-4-qa-e2e-tests/1559994778236489728

md2119:
@gavin-stackrox , this change is pertinent only for postgres world. Therefore, the test failure you pointed out is not related.

PR #2730, opened 2022-08-17
Add FocusTrap to Dashboard menus to allow tab navigation through option elements.

dvail:
## Description

This wraps the dropdown menus for each widget options in a PatternFly `FocusTrap`. The focus trap prevents items outside of the dropdown menu from being focus, or tabbed to, trapping the tab navigation inside the menu. This makes these elements more accessible and keeps them consistent with other similar PatternFly components like the Modal and Popover.

## Checklist

## Testing Performed

Non-accessible behavior is retained after opening the menu:

- All items within the menu can be interacted with via mouse
- Clicking outside of the menu closes the dropdown
- Clicking the "Options" toggle button closes the dropdown
- Clicking the "X" button closes the dropdown

Navigating across the page with Tab and Shift+Tab will allow you to pass over the "Options" toggle buttons.

Navigating to the "Options" toggle button via Tab and pressing "Enter" will open the menu, and the navigation focus will move inside the dropdown, starting on the "X/Close" button.

Pressing "Tab" within the menu will cycle through the tabbable elements of the menu, and they can be interacted with via keyboard.

A video demo is better than screenshots for this, so the video is attached [here](https://issues.redhat.com/secure/attachment/12752139/Screen%20Recording%202022-08-18%20at%209.23.47%20AM.mov).


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/OptionsDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+                focusTrapOptions={{
+                    // If there is no focusable element passed in `props.children`, this will put
+                    // the focus on the toggle button instead of throwing an error.
+                    fallbackFocus: `#${toggleId}`,
+                    clickOutsideDeactivates: true,
+                    escapeDeactivates: true,
+                }}
+            >
+                <Button
+                    style={{ position: 'absolute', top: '0.75rem', right: 0 }}
```
dvail, line 47: At first I attempted at cleaner layout with flex and grid, but there was always a large strip of extra whitespace on the side of the menu. The absolute positioning is a bit clunky but IMO results in the best layout.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/OptionsDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+        >
+            <FocusTrap
+                focusTrapOptions={{
+                    // If there is no focusable element passed in `props.children`, this will put
+                    // the focus on the toggle button instead of throwing an error.
+                    fallbackFocus: `#${toggleId}`,
+                    clickOutsideDeactivates: true,
+                    escapeDeactivates: true,
+                }}
+            >
```
pedrottimark, line 45: Here is a thought for consideration instead of a request for change. 

What is pro and con of an alternative encapsulation via `DropdownFocusTrap` component to wrap the children of a `Dropdown` element?
* Potential pro: Is there a reuse case for children of other than `Dropdown` element?
* Potential con: If `handleToggle` function is also needed for it to work. On the other side of the coin, is the callback function relevant for accessibility in non-focus-trap use of `useSelectToggle` hook?

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/OptionsDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+    children: ReactNode;
+    toggleId: string;
+};
+
+/**
+ * Component that uses a PatternFly `Dropdown` with the contents wrapped in a `FocusTrap` to
+ * enhance accessibility for interactive elements. Upon opening the dropdown, the browser will focus
+ * on the container's close button and only allow tab navigation within the container.
+ */
+function OptionsDropdown({ children, toggleId }: OptionsDropdownProps) {
```
sachaudh, line 17: I have a general question about focus trapping here. So does this mean that PatternFly uses focus trapping under the hood for components like the Modal and Popover but not for the Dropdown? If this is something that the Dropdown didn't have intentionally, is there a reason for that? If they didn't, and maybe they should have, is it worth bringing it up for consideration on their part?

pedrottimark: Because other occurrences have `dropdownItems` prop, in which you do not tab between items; but these have form as child, in which you tab between fields, correct?

pedrottimark: Having said that, System Health diagnostic bundle form is similar to dashboard:
* current classic-PatternFly hybrid: https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/SystemHealth/Components/DiagnosticBundleDialogBox.tsx#L146-L211
* future PatternFly is child of `Popover` element: https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/SystemHealth/PatternFly/Components/GenerateDiagnosticBundle.tsx#L102-L135

dvail: I think this is because the main use case for the Dropdown seems to be a list of items/buttons as shown in all of the demos like Mark mentioned, but I'm passing a form directly as child props. The [PF docs](https://www.patternfly.org/v4/components/dropdown#dropdown) say you can pass a `ReactNode` of "Anything which can be rendered in a dropdown". I can mention it to the PF team and see if this was an intentional omission, or if it is just a rare enough use case that they haven't implemented it.

Let me take a look at doing this with a `Popover` as well, and see if that implementation is cleaner and keeps the same functionality.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/OptionsDropdown.tsx
dvail: The `handleToggle` is necessary only in this specific use case (focus-trapping a `Dropdown`), since PatternFly's default behavior will close the Dropdown when the user presses "tab". This is opposed to the behavior of a Popout, where "tab" iterates through the interactive items contained in the Popout body. Possibly another good argument to just convert these components to use `Popout` instead.

dvail:
Updated the implementation to just use a Popper directly, which makes this change a lot more straightforward and gives us better accessibility out of the box. Functionality remains identical to the video linked at the top of the PR, with a slight visual change that I ran by Mansur regarding the display of the dropdown:
image

pedrottimark (approved):
Super! Thank you for attention to detail about accessibility.

PR #2736, opened 2022-08-18
Ignore unknown mime type error when fetching signatures

dhaus67:
## Description

Currently, older images that do not use [docker manifest schema type](https://github.com/google/go-containerregistry/blob/771a9b4471049aa1b25ef5c1e51907f86e4db07a/pkg/v1/types/types.go#L33-L34) or [OCI manifest 
schema](https://github.com/google/go-containerregistry/blob/main/pkg/v1/types/types.go#L24) are leading to an error when fetching image signatures, specifically:
`unknown mime type: application/vnd.docker.distribution.manifest.v1+prettyjws`

This leads to error logs being logged, although this doesn't necessarily indicate an error, simply that we cannot fetch signatures for the specific image.

### DIFF
--- a/pkg/signatures/public_key_fetcher.go
+++ b/pkg/signatures/public_key_fetcher.go
@@ -53,6 +53,12 @@ func init() {
 // NOTE: No error will be returned when the image has no signature available. All occurring errors will be logged.
 func (c *cosignPublicKeySignatureFetcher) FetchSignatures(ctx context.Context, image *storage.Image,
 	registry registryTypes.Registry) ([]*storage.Signature, error) {
+	// Short-circuit for images that do not have V2 metadata associated with them. These would be older images manifest
+	// schemes that are not supported by cosign, like the docker v1 manifest.
+	if image.GetMetadata().GetV2() == nil {
+		return nil, nil
+	}
+
 	// Since cosign makes heavy use of google/go-containerregistry, we need to parse the image's full name as a
 	// name.Reference.
 	imgFullName := image.GetName().GetFullName()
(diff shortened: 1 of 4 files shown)


dhaus67:
* master
  * **PR #2736** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2736" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2736?utm_source=stack-comment).

connorgorman (approved):
Makes sense, but wondering if we can use the property that causes this to fail to not even make the call to signatures

dhaus67:
@connorgorman added the short-circuit for `v2 == nil`, but will still keep filtering out the error in case we have in _some_ future the issue that we introduce other manifest schemes that are not supported by cosign and are `v2 != nil`.

connorgorman (approved):

pkg/signatures/public_key_fetcher.go
```diff
@@ -53,6 +53,12 @@ func init() {
 // NOTE: No error will be returned when the image has no signature available. All occurring errors will be logged.
 func (c *cosignPublicKeySignatureFetcher) FetchSignatures(ctx context.Context, image *storage.Image,
 	registry registryTypes.Registry) ([]*storage.Signature, error) {
+	// Short-circuit for images that do not have V2 metadata associated with them. These would be older images manifest
+	// schemes that are not supported by cosign, like the docker v1 manifest.
+	if image.GetMetadata().GetV2() == nil {
```
connorgorman, line 58: @dhaus67 Something just hit me, this will make it so that metadata must be pulled before signatures. This is desired right?

dhaus67: Yeah, I've checked the two places where the signature fetching is done ,would be a) [the enrichment for local scanner](https://github.com/stackrox/stackrox/blob/master/sensor/common/scan/scan.go#L39) and b) [the image enricher](https://github.com/stackrox/stackrox/blob/master/pkg/images/enricher/enricher_impl.go#L171). Both times we do just fetch the image metadata beforehand (we short-circuit in both places when image metadata cannot be fetched).

PR #2739, opened 2022-08-18
PF theme tweak test

dvail:
## Description

This adds support for dark theme on PatternFly pages and components. This feature is behind a feature flag and _disabled_ by default. The thought is that we roll this out first to the engineering teams as a preview, and then enable the feature flag once the 73 release is cut so that we have a full release cycle to find inconsistencies.

The main functional changes here are:
- The `light-theme` class that forces light styles on the `SearchFilterInput` and network graph header is now conditionally applied only when the feature flag is disabled. When the flag is enabled, we let these components get the globally applied light/dark theme.
- The integration tiles are set to have an off-white background when in dark mode, so that the variety of logos remain visible.

## Checklist

## Testing Performed

With ROX_PF_DARK_THEME **enabled**

NetGraph light mode:
image
NetGraph dark mode:
image

Integration page light mode:
image
Integration page dark mode:
image

Policies page light mode:
image
Policies page dark mode:
image

Dashboard light mode:
image
Dashboard dark mode:
image

With ROX_PF_DARK_THEME **disabled**

NetGraph light mode:
image
NetGraph dark mode:
image

Integrations, Policies, and Dashboard pages are identical to their light mode screenshots above, regardless of the user's selected theme.


ui/apps/platform/src/Containers/AccessControl/AccessScopes/RequirementRowAddKey.tsx
```diff
@@ -98,7 +98,7 @@ function RequirementRowAddKey({
                         className="pf-m-smallest"
                         onClick={handleRequirementKeyCancel}
                     >
-                        <TimesCircleIcon color="var(--pf-global--color--100)" />
```
dvail, line 101: This is the cancel icon for the label selector widget on the AccessControl page. This changes the color from black/black to the default of dark gray/light gray. The former black icon on dark background was very difficult to see.

pkg/features/list.go
```diff
@@ -28,6 +28,9 @@ var (
 	// SearchPageUI enables search page instead of search modal in UI. Frontend only.
 	SearchPageUI = registerFeature("Enable search page instead of search modal in UI", "ROX_SEARCH_PAGE_UI", false)
 
+	// PatternFlyDarkTheme enables dark theme support on PatternFly pages and components
+	PatternFlyDarkTheme = registerFeature("Enable dark theme support on PatternFly pages", "ROX_PF_DARK_THEME", false)
```
dvail, line 32: Note that this feature flag will _only_ toggle the ability to view dark mode on PatternFly components. If the flag is disabled, users can still use dark mode on Tailwind based components like they do today.

PR #2740, opened 2022-08-18
Dashrews/ROX-12252 postgres analyze after restore

dashrews78:
## Description

Simply adding a call to analyze after a restore.  Dmitri mentioned this was typically the best practice because pg_dump does not contain statistical data.  If the call fails for whatever reason we will just log that as a warning and move on as Postgres will eventually analyze the data.  

The CI failure is failing pretty everywhere and is related to migrations running in Postgres mode and failing.  That will be solved when the upgrade/replica PR is pushed.  

## Checklist

## Testing Performed

Just local testing to make sure the command was executed.

### DIFF
--- a/pkg/postgres/pgadmin/admin_utils.go
+++ b/pkg/postgres/pgadmin/admin_utils.go
@@ -153,6 +153,20 @@ func GetDatabaseClones(pgConfig *pgxpool.Config) []string {
 	return clones
 }
 
+// AnalyzeDatabase - runs ANALYZE on the database named dbName
+func AnalyzeDatabase(config *pgxpool.Config, dbName string) error {
+	log.Debugf("Analyze - %q", dbName)
+
+	// Connect to different database for admin functions
+	connectPool := GetClonePool(config, dbName)
+	// Close the admin connection pool
+	defer connectPool.Close()
+
+	_, err := connectPool.Exec(context.Background(), "ANALYZE")
+
+	return err
+}
+
 // GetAdminPool - returns a pool to connect to the admin database.
 // This is useful for renaming databases such as a restore to active.
 // THIS POOL SHOULD BE CLOSED ONCE ITS PURPOSE HAS BEEN FULFILLED.
@@ -183,6 +197,7 @@ func GetClonePool(pgConfig *pgxpool.Config, clone string) *pgxpool.Pool {
 	postgresDB := getPool(tempConfig)
 
 	log.Debugf("Got connection pool for database %q", clone)
+
 	return postgresDB
 }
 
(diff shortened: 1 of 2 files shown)


pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -107,20 +107,63 @@ func CheckIfDBExists(config *pgxpool.Config, dbName string) bool {
...
+	// Connect to different database for admin functions
+	connectPool := GetReplicaPool(config, dbName)
+	// Close the admin connection pool
+	defer connectPool.Close()
+
+	// Create a context with a timeout
+	ctx, cancel := context.WithTimeout(context.Background(), postgresQueryTimeout)
+	defer cancel()
+
+	_, err := connectPool.Exec(ctx, "ANALYZE")
```
c-du, line 123: How long does this take if we restored a large database? If it takes long, postgresQueryTimeout may not be enough. And are we holding grpc call waiting for it?

erthalion: Right, analyzing could be a heavy operation if there is a lot of data.

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -107,20 +107,63 @@ func CheckIfDBExists(config *pgxpool.Config, dbName string) bool {
...
 
+	postgresDB := getPool(tempConfig)
+
+	log.Debugf("Got connection pool for database %q", AdminDB)
+	return postgresDB
+}
+
+// GetReplicaPool - returns a connection pool for the specified database replica.
+// THIS POOL SHOULD BE CLOSED ONCE ITS PURPOSE HAS BEEN FULFILLED.
+func GetReplicaPool(pgConfig *pgxpool.Config, replica string) *pgxpool.Pool {
```
erthalion, line 146: [nit] Are there better options to name this database except "replica"? It will be inevitably confusing in the future when the actual replicas (streaming/logical replica instances) will be involved.

erthalion (commented):
Thanks for taking care of it. A few small notes below:

pkg/postgres/pgadmin/admin_utils.go
dashrews78: @erthalion yeah that has been bugging me a little as I've been working on the rollback/upgrade stuff.  I think I will solve this there.  Either as part of ROX-11677 or some other subtask under ROX-9884.  I'm leaning toward changing replica to clone as that probably more closely resembles what these things are.

connorgorman (approved):

erthalion (approved):

c-du (approved):

PR #2742, opened 2022-08-18
ROX-12263: fix authz plugin removal migrator crash on recent DB

misberner:
## Description

We stopped creating the bucket a while ago, but the migrator now fails if the `authzPlugins` bucket does not exist. This shouldn't happen, so make sure it doesn't fail if the error is a missing bucket.

Also fix some log messages to aid in debugging when a migration fails.

## Checklist

## Testing Performed

- Added a unit test
- Live test on scale test instance

### DIFF
--- a/migrator/migrations/m_107_to_m_108_remove_auth_plugin/migration.go
+++ b/migrator/migrations/m_107_to_m_108_remove_auth_plugin/migration.go
@@ -72,10 +72,11 @@ func migratePS(db *gorocksdb.DB) error {
 
 func deleteAuthPluginBucket(db *bbolt.DB) error {
 	return db.Update(func(tx *bbolt.Tx) error {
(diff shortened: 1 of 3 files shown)


migrator/migrations/m_107_to_m_108_remove_auth_plugin/migration.go
```diff
@@ -72,7 +72,10 @@ func migratePS(db *gorocksdb.DB) error {
 
 func deleteAuthPluginBucket(db *bbolt.DB) error {
 	return db.Update(func(tx *bbolt.Tx) error {
-		return tx.DeleteBucket(authPluginBucket)
+		if err := tx.DeleteBucket(authPluginBucket); err != nil && !errors.Is(err, bbolt.ErrBucketNotFound) {
```
connorgorman, line 75: I prefer an explicit bucket check

if tx.Bucket(authPluginBucket) == nil {
  return nil
}

connorgorman (approved):

PR #2748, opened 2022-08-19
ROX-11939: Expose scanner monitoring

stehessel:
## Description

This PR aims to enable Prometheus monitoring of Scanner pods. To this extent, it includes two necessary changes:

1. Add `exposeMonitoring` flag to the Scanner Helm chart. This is based on https://github.com/stackrox/rox/pull/7985, which adds this flag to the Central chart. 
2. Add the flag to the ACS operator. The operator populates the Helm chart with values, after which the Helm chart performs the actual installation. This is based on https://github.com/stackrox/stackrox/pull/2117, which adds Central monitoring to the operator.

After this PR has been merged, another PR in fleet manager will be required to enable the monitoring flag in data plane deployed Scanners, similar to https://github.com/stackrox/acs-fleet-manager/pull/158.

## Checklist

## Testing Performed

Helm tests and operator tests in CI.


operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -168,7 +168,7 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
 
 // Monitoring defines settings for monitoring endpoint.
 type Monitoring struct {
-	// Expose Central's monitoring endpoint. A new service, "monitoring",
+	// Expose the monitoring endpoint. A new service, "monitoring",
```
misberner, line 171: Consequently, this `struct` should be moved to `common_types.go`

misberner (approved):
I think this change looks good in principle, mirroring the monitoring exposure setting for central.

In the context of ACSMS, I have some concerns that I will discuss via Slack.

image/templates/helm/shared/templates/02-scanner-07-service.yaml
```diff
@@ -19,6 +19,11 @@ spec:
     - name: grpcs-scanner
       port: 8443
       targetPort: 8443
+    {{ if ._rox.scanner.exposeMonitoring -}}
+    - name: monitoring
+      port: 9090
+      targetPort: monitoring
```
RTann, line 25: feel free to ignore, but should we use a concrete number like the rest (or switch the others to named)?

operator/apis/platform/v1alpha1/common_types.go
```diff
@@ -166,3 +166,33 @@ const (
 	// ScannerAutoScalingDisabled means that scanner autoscaling should be disabled.
 	ScannerAutoScalingDisabled AutoScalingPolicy = "Disabled"
 )
+
+// Monitoring defines settings for monitoring endpoint.
+type Monitoring struct {
+	// Expose the monitoring endpoint. A new service, "monitoring",
+	// with port 9090, will be created as well as a network policy allowing
```
RTann, line 173: does this have to be 9090?

RTann (commented):
I'm curious how are we handling the fact there are 2 - 5 Scanners in a production deployment? I think this just looks at the metrics for the Scanner which is chosen by the Service, right?

image/templates/helm/shared/templates/02-scanner-07-service.yaml
stehessel: I used the `monitoring` name to be consistent with how monitoring is configured for Central. Having a consistent name is also useful when configuring the Prometheus pod monitors for both Central and Scanner. Perhaps we should add names for the other ports as well, but I don't think it would be in scope here.

operator/apis/platform/v1alpha1/common_types.go
stehessel: It doesn't have to be, but it's a very common choice for Prometheus. I don't think we gain much by making the port configurable, and we already hard code it for Central.

stehessel:
> I'm curious how are we handling the fact there are 2 - 5 Scanners in a production deployment? I think this just looks at the metrics for the Scanner which is chosen by the Service, right?

Good point. I think it depends on how the Prometheus scrape is configured (see the different role types in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config). Basically Prometheus either allows service discovery - then the scrape is routed to only one pod - or pod discovery, in which case all pods should be scraped. For the pod discovery role config the service definition should be unnecessary, but again I'm not sure it makes sense to allow individual configuration.

misberner:
> > I'm curious how are we handling the fact there are 2 - 5 Scanners in a production deployment? I think this just looks at the metrics for the Scanner which is chosen by the Service, right?
> 
> Good point. I think it depends on how the Prometheus scrape is configured (see the different role types in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config). Basically Prometheus either allows service discovery - then the scrape is routed to only one pod - or pod discovery, in which case all pods should be scraped. For the pod discovery role config the service definition should be unnecessary, but again I'm not sure it makes sense to allow individual configuration.

I think `pod` might work, but what is an even better choice here is `endpoints`. You can run `kubectl -n stackrox get endpoints scanner -o yaml` to see what the `endpoints` resource backing the `scanner` service looks like - it basically contains all possible (ip, port) combinations resulting from resolving the pod selector of the service to concrete pod IPs. IMHO, this is the best of all worlds, as:
1. You can select ports for scraping even if they don't show up in the `containerPorts` section of the pod spec, which might not always be complete. Rather than patching a missing `containerPort` into the deployment resource, a `Service` can be created as a new resource, not modifying any of the existing ones.
2. You do not miss any replicas, as can happen with the `service` sd role.
3. You still get the logical grouping imposed by the `Service` object (you can stitch that together with `pods` via labels, but with the `endpoints` sd role, you also get all the labels from the service sd role ("If the endpoints belong to a service, all labels of the role: service discovery are attached.")

janisz (approved):

PR #2749, opened 2022-08-19
Upgrade Operator SDK to v1.20.1

vladbologa:
## Description

This PR upgrades `operator-sdk` from v1.14.0 to v1.20.1, to enable building the Stackrox Operator on Darwin arm64 machines.

Applied migrations for all the intermediate versions.

The latest version of the `operator-sdk` is currently 1.23.0, but I did not upgrade all the way to that because:
* from v1.21 onwards, the migrations cause the build to fail on arm64 (the `operator-sdk` itself provides `darwin/arm64` deliverables, but the platform is not really supported by the `golang/v3` plugin which the operator is using)
* from v1.22 onwards, the go version in `go.mod` should be 1.18, and we're currently at 1.17 and I'm unsure that we can upgrade atm

Migration instructions are available here:
https://sdk.operatorframework.io/docs/upgrading-sdk-version/

## Checklist

## Testing Performed

Tested locally by building the various operator targets.
Tested the upgrade and E2E tests on an OpenShift cluster.
Otherwise, the CI tests should suffice.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -44,6 +44,19 @@ IMAGE_TAG_BASE ?= $(IMAGE_REPO)/stackrox-operator
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
+# BUNDLE_GEN_FLAGS are the flags passed to the operator-sdk generate bundle command
+# Version is hardcoded to 0.0.1 here because otherwise git-versioned files are changed which we want to avoid.
+# The correct version is updated later.
+BUNDLE_GEN_FLAGS ?= -q --overwrite --version 0.0.1 $(BUNDLE_METADATA_OPTS)
+
+# USE_IMAGE_DIGESTS defines if images are resolved via tags or digests
+# You can enable this value if you would like to use SHA Based Digests
+# To enable set flag to true
+USE_IMAGE_DIGESTS ?= false
+ifeq ($(USE_IMAGE_DIGESTS), true)
+    BUNDLE_GEN_FLAGS += --use-image-digests
+endif
+
 # INDEX_IMG defines the image:tag used for the index (a.k.a. catalog) image, containing the single operator bundle image.
 INDEX_IMG_BASE = $(IMAGE_TAG_BASE)-index
 INDEX_IMG ?= $(INDEX_IMG_BASE):v$(VERSION)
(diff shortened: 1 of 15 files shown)


operator/README.md
```diff
@@ -189,8 +189,8 @@ $ kubectl -n bundle-test patch serviceaccount default -p '{"imagePullSecrets": [
 # Use one-liner above.
 
 # 4. Run bundle.
-$ bin/operator-sdk-1.14.0 run bundle \
-  docker.io/stackrox/stackrox-operator-bundle:v$(make --quiet tag) \
+$ bin/operator-sdk-1.22.2 run bundle \
+  quay.io/rhacs-eng/stackrox-operator-bundle:v$(make --quiet tag) \
```
vladbologa, line 193: Had to use `rhacs-eng` here because the `stackrox-io/stackrox-operator-bundle` is empty atm.

operator/Makefile
```diff
@@ -236,20 +251,25 @@ CRDS := centrals.platform.stackrox.io securedclusters.platform.stackrox.io
 validate-crs:
 	./tests/scripts/validate-crs.sh ./tests $(CRDS)
 
+.PHONY: test-e2e
 test-e2e: build install validate-crs kuttl ensure-rox-main-image-exists ## Run e2e tests with local manager.
 	mkdir -p $(PROJECT_DIR)/build/kuttl-test-artifacts
-	KUTTL=$(KUTTL) ENABLE_WEBHOOKS=$(ENABLE_WEBHOOKS) $(KUTTL) test
+	KUTTL=$(KUTTL) REAL_KUBECONFIG=$${KUBECONFIG} ENABLE_WEBHOOKS=$(ENABLE_WEBHOOKS) $(KUTTL) test
+
```
porridge, line 258:
```suggestion
```

operator/Makefile
```diff
@@ -320,7 +346,7 @@ install: check-ci-setup manifests kustomize ## Install CRDs into the K8s cluster
 
 .PHONY: uninstall
 uninstall: check-ci-setup manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```
porridge, line 349:
```suggestion
	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```

operator/bundle/manifests/rhacs-operator.clusterserviceversion.yaml
```diff
@@ -894,6 +897,9 @@ spec:
   minKubeVersion: 1.15.0
   provider:
     name: Red Hat
+  relatedImages:
+  - image: ''
+    name: ''
```
porridge, line 902: Hm, this bit looks weird. Can we avoid it easily?

porridge (commented):
Looks good, modulo the duplicate `kubectl delete`.

Why a draft? Are there more changes to come? I noticed commits for migrations for just some versions - are the remainders pending or just didn't contain any migrations?

operator/Makefile
vladbologa: I made it a draft because there are more changes to come, and I was also waiting to see if I can merge this: https://github.com/stackrox/stackrox/pull/2784

The migration to v1.17 include go version 1.17 in go.mod, and I wasn't sure I could do that, due to some downstream problems. But now it's done.

operator/bundle/manifests/rhacs-operator.clusterserviceversion.yaml
vladbologa: I think this was generated. Let's see if it's still there when I'm done with this.

vladbologa: It's a generated file, not sure there's anything I can do about it.

operator/Makefile
```diff
@@ -372,14 +410,12 @@ ACTIVATE_PYTHON = python3 -m venv bundle_helpers/.venv ;\
 
 .PHONY: bundle
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
-# Version is hardcoded to 0.0.1 here because otherwise git-versioned files are changed which we want to avoid.
-# The correct version is updated later.
 # Likewise, we hardcode the image reference to quay.io/stackrox-io/stackrox-operator. If this is overridden via
```
porridge, line 413:
```suggestion
# We hardcode the image reference to quay.io/stackrox-io/stackrox-operator. If this is overridden via
```

porridge (approved):
Awesome job, just one nitpick inline.

This is worrying: `the platform is not really supported by the golang/v3 plugin which the operator is using` - can you please make sure the right people are aware of this?

Can you please also close https://issues.redhat.com/browse/ROX-9011 with a reference to this PR when it's merged?

misberner (approved):

vladbologa:
> This is worrying: `the platform is not really supported by the golang/v3 plugin which the operator is using` - can you please make sure the right people are aware of this?

@porridge Do you mean on the `operator-sdk` side? They are definitely aware of this, v1.23.0 adds the `v4_alpha` plugin, which officially supports `arm64`.

For now I'm fine with using v1.20.1, as it works on my arm machine. If we at some point officially support arm64 (which I hope we will), we should upgrade the plugin to v4. They changed the version because it introduces some breaking changes, such as updating kustomize from v3.8 to v4.5.

PR #2760, opened 2022-08-19
Dashrews/ROX-12271 check postgres for replica space

dashrews78:
## Description

Before creating clones to be used in upgrades and restores, we need to ensure that Postgres has enough space for the clone.  This PR does that calculation to ensure we have space.  There is one caveat to that.  

We do not yet know how to best get the available size of a Postgres installation because it could be local, managed, etc.  We would have to use provider mechanisms to get that information for a managed instance.  For a local instance it varies and may be difficult to do in a secure fashion.  So for now, we are doing the calculations based on the assumption we have 100GB of Postgres storage.  ROX-12059 will address that.

The other trick to this is backup/restores.  The backup generated by pg_dump does not contain database size information.  Additionally the size of the dump file is likely significantly smaller than the database itself.  So I had to write a new file `postgres.size` that gets written to the backup bundle first to ensure that it is processed first.  The restore process reads that file to determine if there is enough space or not to perform the backup.  The DB size is the only thing written to that file.

## Checklist

## Testing Performed

Manual testing to specifically drive cases like an invalid size in the backup bundle.  Modified the size constant to be incredibly small and ran upgrades and restores manully to make sure not enough space was reported as expected.

### DIFF
--- a/central/globaldb/v2backuprestore/backup/generators/dbs/postgres_size.go
+++ b/central/globaldb/v2backuprestore/backup/generators/dbs/postgres_size.go
@@ -1,0 +1,46 @@
+package dbs
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/stackrox/rox/pkg/migrations"
+	"github.com/stackrox/rox/pkg/postgres/pgadmin"
+	"github.com/stackrox/rox/pkg/postgres/pgconfig"
+)
+
+// NewPostgresSize returns a generator for Postgres backups.
+// We take in the connection to connect to the DB
+func NewPostgresSize(db *pgxpool.Pool) *PostgresSize {
+	return &PostgresSize{
+		db: db,
+	}
+}
+
+// PostgresSize is an implementation of a StreamGenerator which writes a backup of PostgresDB to the input io.Writer.
+type PostgresSize struct {
+	db *pgxpool.Pool
+}
+
+// WriteTo writes a backup of Postgres to the writer
+func (ps *PostgresSize) WriteTo(ctx context.Context, out io.Writer) error {
+	_, config, err := pgconfig.GetPostgresConfig()
+	if err != nil {
+		log.Fatalf("Could not parse postgres config: %v", err)
+		return err
+	}
+
+	size, err := pgadmin.GetDatabaseSize(config, migrations.GetCurrentClone())
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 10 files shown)


central/globaldb/v2backuprestore/formats/postgresv1/postgres.go
```diff
@@ -2,23 +2,66 @@ package postgresv1
...
+	}
+
+	availableBytes, err := pgadmin.GetRemainingCapacity(dbConfig)
+	if err != nil {
+		log.Warnf("Could not determine free disk space for Postgres: %v. Assuming free space is sufficient for %d bytes.", err, requiredBytes)
+		return nil
+	}
+
+	hasSpace := float64(availableBytes) > float64(requiredBytes)
+	if !hasSpace {
```
c-du, line 62: Are we going to have a more precise number or precise estimate on capacity soon? If we could not get a better number, we'd better not blocking restore process.

central/globaldb/v2backuprestore/formats/postgresv1/format.go
```diff
@@ -11,6 +11,7 @@ import (
 func init() {
 	formats.MustRegisterNewFormat(
 		"postgresv1",
+		common.NewFileHandler(backup.PostgresSize, false, checkPostgresSize),
```
c-du, line 14: how about PostgresSizeFile?

central/globaldb/v2backuprestore/backup/generators/dbs/postgres_size.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// PostgresSize is an implementation of a StreamGenerator which writes a backup of PostgresDB to the input io.Writer.
+type PostgresSize struct {
+	db *pgxpool.Pool
+}
+
+// WriteTo writes a backup of Postgres to the writer
+func (bu *PostgresSize) WriteTo(ctx context.Context, out io.Writer) error {
```
c-du, line 28: bu? How about p or ps?

central/globaldb/v2backuprestore/formats/postgresv1/postgres.go
dashrews78: @c-du I guess some of that depends on how you define soon.  I'd say that this is our best guess for 3.72.  For 3.73, we should spend some time on [ROX-12059](https://issues.redhat.com/browse/ROX-12059) and decide how we are going to arrive at that.  If you like, I could put a todo here and log a warning instead of returning the error.

PR #2763, opened 2022-08-19
Add comment about redirect for deprecated policies path in UI

pedrottimark:
## Description

The only known use of deprecated /main/policies/id was replaced by import from routePaths in #2464

Initial investigation of react-router-dom upgrade suggested that that this use of `Redirect` element might become an obstacle.

After discussion by the team, add a comment so we do not overlook it when planning for the upgrade.

## Checklist

## Testing Performed

1. `yarn build` in ui


alwayshooin:
so the deprecated path was so that if users had saved URLs or had bookmarked previous policies, the UI will redirect to said policy with the updated URL instead of throwing a 404 or redirecting to the dashboard. I intentionally left the deprecated policy URL to ensure backwards compatibility -- maybe we can discuss with the team to see whether that's a valid concern to address or not

alwayshooin (approved):

PR #2437, opened 2022-07-15
Delete obsolete processes reducers and sagas in UI

pedrottimark:
## Description

We are investigating **duplicate requests** because they increase risk of integration test failures (see manual testing below).

### Analysis

Risk container calls service functions directly instead of indirectly via reducers and sagas.

1. reducer actions
    * fetchProcesses: 3 calls of action in processSagas.js and 2 calls in Risk container of corresponding function from ProcessesService.js
    * fetchProcessesBaseline: 2 calls of action in processSagas.js
    * addDeleteProcesses: 1 call of action in processSagas.js and 2 calls in Risk container of corresponding function from ProcessService.js
    * lockUnlockProcesses: 1 call of action in processSagas.js and 2 calls in Risk container of corresponding function from ProcessService.js
2. reducer selectors
    * getProcessesByDeployment: 0 occurrences
    * getProcessesBaselineByDeployment: 0 occurrences
3. Risk container
    * mapDispatchToProps: 1 occurrence in CreatePolicyFromSearch which is unrelated to processes
    * mapStateToProps: 0 occurrences
    * useDispatch: 0 occurrences
    * useSelector: 0 occurrences

### Changed files

1. Edit reducers/index.js

2. Delete reducers/processes.js

3. Edit sagas/index.js

4. Delete sagas/processSagas.js

### Residue

1. Rewrite services/ProcessesService.js in TypeScript

## Checklist
- [x] Investigated and inspected CI test results
- [x] ~Unit test and regression tests added~

### manual testing

1. `yarn build` in ui
2. `wc build/static/js/*.chunk.js` in ui
    * branch - master: -4300 = 10410170 - 10414470
3. `yarn start` in ui

    * Visit /main/risk: see requests from RiskTablePanel
        GET /v1/deploymentswithprocessinfo?pagination.offset=0&pagination.limit=50&pagination.sortOption.field=Deployment%20Risk%20Priority&pagination.sortOption.reversed=false&query=Orchestrator%20Component%3Afalse
        GET /v1/deploymentscount?query=Orchestrator%20Component%3Afalse
    
    * Click a deployment: see **duplicate requests** before changes but not after changes
        GET /v1/processes/deployment/id/grouped/container from processSagas before changes but not after changes
        GET /v1/deploymentswithrisk/id from RiskSidePanel
        GET /v1/processes/deployment/id/grouped/container from RiskSidePanel

### integration testing

[...]

PR #2764, opened 2022-08-19
Delete obsolete Containers MainPage index file

pedrottimark:
## Description

No occurrence of `import MainPage from 'Containers/MainPage';`

1. src/index.tsx: `import AppPage from 'Containers/AppPage';`
2. src/Containers/AppPage.tsx: `import AuthenticatedRoutes from 'Containers/MainPage/AuthenticatedRoutes';`
3. src/Containers/MainPage/AuthenticatedRoutes.tsx: `import MainPage from './MainPage';`

## Checklist

## Testing Performed

1. `yarn build` in ui
    * branch - master: 0 = 5999821 - 5999821 confirms that is was not referenced

Dave, did you notice delta build size from webpack 5 upgrade included in react-scripts 5 upgrade?

Although I forgot to compare before and after, here is a similar measurement from #2437

> branch - master: -4300 = 10410170 - 10414470

1. In case there is more to lose, let’s investigate next week what is the project browserslist configuration.

    https://webpack.js.org/configuration/target/

    > If a project has a browserslist config, then webpack will use it to determinate ES-features that may be used to generate runtime-code.

2. Webpack 5 opens the door to review `"target": "es5"` in tsconfig.json file.

    https://www.typescriptlang.org/tsconfig#target

    > Modern browsers support all ES6 features, so ES6 is a good choice.

### DIFF
--- a/ui/apps/platform/src/Containers/MainPage/index.ts
+++ b/ui/apps/platform/src/Containers/MainPage/index.ts
@@ -1,1 +1,0 @@
-export { default } from './MainPage';


dvail (approved):
RE: bundle size

I did notice a significant change in bundle size in the move to react-scripts 5, with the new build being ~2MB smaller. I ran the bundle analyzer before and after the change against the CSS and JS files and included the notes at the bottom of the PR description:
https://github.com/stackrox/stackrox/pull/2609

pedrottimark:
@dvail Thank you for the careful analysis of the upgrade, far beyond what I would have thought to do!

> if we dynamically build TW class names from strings in any place in the app

Yeah, that seems likely. Finding and fixing seems like good exercise to break that habit. It will hurt us with globalization and was a subtle source of problem in search links when policies path changed.

PR #2766, opened 2022-08-19
Fix dependency conflict leading to failing tests.

dvail:
## Description

While working on the accessibility/Popover fixes for the dashboard, I was continuously running into the
```
When testing, code that causes React state updates should be wrapped into act(...):
```
error for existing tests. These tests as well as a very minimal repro case using `Popover` would fail consistently in ways that were contrary to all documentation on how `testing-library` should be used for tests.

Standing up a new repo and using the exact `ui/apps/platform/package.json` and `ui/yarn.lock` files from the stackrox repo would make the exact test setup pass.

I narrowed this down to the `ui-components` package dependency `eslint-plugin-jest-dom`, which brought in an incompatible `@testing-library/dom` version (7.x, we need 8.x) that was being picked up by the tests from the monorepo's root `node_modules`. This was agonizing to track down, but cleared all issues with tests related to `Popovers`. I have a hunch that a lot of previous `act` errors that needed a work around might have been related to this inconsistency as well.

## Checklist

## Testing Performed

ESLint dependency change - `yarn test && yarn lint` from the root pass as normal.

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -1475,7 +1475,7 @@
     core-js-pure "^3.0.0"
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.1.2", "@babel/runtime@^7.10.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.14.0", "@babel/runtime@^7.15.4", "@babel/runtime@^7.16.3", "@babel/runtime@^7.4.2", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.3", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2", "@babel/runtime@^7.9.6":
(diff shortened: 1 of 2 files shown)


ui/yarn.lock
```diff
@@ -3691,31 +3691,31 @@
...
     "@babel/code-frame" "^7.10.4"
     "@babel/runtime" "^7.12.5"
     "@types/aria-query" "^4.2.0"
     aria-query "^4.2.2"
     chalk "^4.1.0"
-    dom-accessibility-api "^0.5.4"
+    dom-accessibility-api "^0.5.6"
     lz-string "^1.4.4"
-    pretty-format "^26.6.2"
+    pretty-format "^27.0.2"
```
pedrottimark, line 3706: Something to discuss next week: sometimes I have merged close versions in `yarn.lock` file.

Do you have any experience good or bad with doing that?

```text
"@testing-library/dom@^8.0.0", "@testing-library/dom@^8.11.1":
    version "8.17.1"
```

pedrottimark (approved):
The thought that hit me last night: next week as a follow up to the react-scripts upgrade, let’s compare `devDependencies` in the 3 package.json files for consistency with each other and with versions in `yarn.lock` file.

ui/yarn.lock
dvail: I haven't done this a ton in the past, but did it once for the react-scripts 5 upgrade. Typically I leave the yarn.lock alone unless it is a relatively simple change. This one seems like a decent candidate to merge, and should be easy to tell if something breaks since it is purely a test dependency.

PR #2768, opened 2022-08-19
ROX-9766: Add severity column to Image CVEs and Node CVEs tables

sachaudh:
## Description

This change ensures that we show the severity column in the Image and Node CVEs tables

## Checklist

## Testing Performed

Screen Shot 2022-08-19 at 4 41 22 PM
Screen Shot 2022-08-19 at 4 41 28 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/ListCVEs.utils.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/ListCVEs.utils.js
@@ -33,6 +33,9 @@ export function getFilteredCVEColumns(columns, workflowState) {
     const shouldKeepCveType =
         currentEntityType === entityTypes.CVE || currentEntityType === entityTypes.CLUSTER_CVE;
 
+    const shouldKeepSeverity =
+        currentEntityType === entityTypes.IMAGE_CVE || currentEntityType === entityTypes.NODE_CVE;
+
     return columns.filter((col) => {
         switch (col.accessor) {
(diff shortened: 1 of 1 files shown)


vjwilson (approved):

ui/apps/platform/src/Containers/VulnMgmt/List/Cves/ListCVEs.utils.js
```diff
@@ -55,7 +58,7 @@ export function getFilteredCVEColumns(columns, workflowState) {
                 return shouldKeepEntitiesColumn;
             }
             case 'severity': {
-                return shouldKeepDiscoveredAtImageColumn;
+                return shouldKeepSeverity || shouldKeepDiscoveredAtImageColumn;
```
md2119, line 61: why are you checking for `shouldKeepDiscoveredAtImageColumn` to determine the visibility of severity?

sachaudh: It seems like we use that to determine whether to show the `severity` column. If you git blame you can track it down to this PR https://github.com/stackrox/rox/pull/8080. It seems like he added it according to that PR. All I know is that we want to show the severity for the Image CVEs and Node CVEs tables so I'm adding that conditional logic to display it for those. I don't want to mess around too much with what's here right now since we will eventually make changes to VM soon. If you want to dig into why it was added, maybe you can ask Gorman for clarification.

PR #2784, opened 2022-08-22
ROX-8056: upgrade to go 1.17 in go.mod

vladbologa:
## Description

* Changed go version in go.mod to 1.17, then ran `go mod tidy -go=1.16 && go mod tidy -go=1.17`.
* Removed hardcoded go version protection in `scripts/ci/libs.sh`

Note that running `go mod tidy` after changing  1.16 to1.17 results in error, and the tool outputs:
```
To upgrade to the versions selected by go 1.16:
	go mod tidy -go=1.16 && go mod tidy -go=1.17
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
```

## Checklist

## Testing Performed

CI should be sufficient for these changes. Should not be merged until also downstream is proven to work with these changes.


msugakov (approved):
LGTM
FTR, this replaces https://github.com/stackrox/rox/pull/9728

misberner:
Is there a reason you didn't go with
```
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
```
? Because to me it doesn't sound like this reproducibility is needed, but I also don't have a sense as to how the result would differ.

vladbologa:
> Is there a reason you didn't go with
> 
> ```
> If reproducibility with go 1.16 is not needed:
> 	go mod tidy -compat=1.17
> ```
> 
> ? Because to me it doesn't sound like this reproducibility is needed, but I also don't have a sense as to how the result would differ.

@misberner Actually the first time I went with `-compat=1.17`, then I changed my mind because:
* that one causes a lot of changes in `go.sum`, so I thought the chances of unwanted side effects are higher
* when @vikin91 did this change (which was then reverted) in https://github.com/stackrox/rox/pull/9728 he also didn't have any changes to `go.sum` so I assumed he also chose the same approach

I found this about the topic:
https://stackoverflow.com/questions/71973152/go-mod-tidy-error-message-but-go-1-16-would-select

I wasn't convinced after reading this that we **need** any off the approaches, so I just chose what I thought was slightly safer.

vladbologa (commented):
btw @misberner I just tried this:

If I run `go mod tidy -compat=1.17`, I cannot subsequently run `go mod tidy`, it keeps returning the same error that I have to choose between the two options. If I run `go mod tidy -go=1.16 && go mod tidy -go=1.17`, from that point on we can run `go mod tidy` in the repo, and it will succeed without changing anything.

vikin91:
> when @vikin91 did this change (which was then reverted) in https://github.com/stackrox/rox/pull/9728 he also didn't have any changes to go.sum so I assumed he also chose the same approach

I would not take that as an oracle 😂

vladbologa:
Downstream CI passed:
* https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/pipelines/2277395
* https://jenkins-cpaas-rhacs.apps.cpaas-poc.r6c9.p1.openshiftapps.com/job/gitlab-event-dispatcher/3/console

Which means that it should be safe to merge this PR.

PR #2794, opened 2022-08-23
ROX-11741: Wait for images on Quay.io before creating clusters

0x656b694d:
## Description

Introducing `wait-for-image.sh` script which waits (up to 20 minutes) for the given image to be available on Quay.io.

## Checklist

## Testing Performed

Tested in local and on `test-gh-actions` repository:
https://github.com/stackrox/test-gh-actions/actions/runs/2912132210

### DIFF
--- a/.github/workflows/scripts/wait-for-image.sh
+++ b/.github/workflows/scripts/wait-for-image.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Wait for an image to appear on Quay.io
+#
+set -euo pipefail
+
+NAME="$1"
+TAG="$2"
+
+check_not_empty \
+    NAME \
+    TAG \
+    \
+    QUAY_TOKEN \
+    DRY_RUN
+
+find_tag() {
+    curl --silent --show-error --fail --location \
+        -H "Authorization: Bearer $QUAY_TOKEN" \
+        -X GET "https://quay.io/api/v1/repository/rhacs-eng/$1/tag?specificTag=$2" |
+        jq -r ".tags[0].name"
+}
+
+# Seconds:
+TIME_LIMIT=1200
+INTERVAL=10
+
+# bash built-in variable
+SECONDS=0
+
+FOUND_TAG=""
+while [ "$SECONDS" -le "$TIME_LIMIT" ]; do
+    FOUND_TAG=$(find_tag "$NAME" "$TAG")
+    if [ "$FOUND_TAG" = "$TAG" ]; then
+        gh_log notice "Image '$NAME:$TAG' has been found on Quay.io."
+        break
+    fi
+    if [ "$DRY_RUN" == "true" ]; then
+        break
+    fi
+    echo "Waiting..."
+    sleep "$INTERVAL"
+done
+
+if [ "$FOUND_TAG" != "$TAG" ]; then
+    gh_log error "Image '$NAME:$TAG' has not been found on Quay.io."
+    exit 1
+fi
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/wait-for-image.sh
```diff
@@ -0,0 +1,48 @@
...
+find_tag() {
+    curl --silent --show-error --fail --location \
+        -H "Authorization: Bearer $QUAY_TOKEN" \
+        -X GET "https://quay.io/api/v1/repository/rhacs-eng/$1/tag?specificTag=$2" |
+        jq -r ".tags[0].name"
+}
+
+# Seconds:
+TIME_LIMIT=1200
+INTERVAL=10
```
vikin91, line 26: I think CircleCI used to wait up to 30 minutes, 20 mins is also fine.
I hope we will not run into rate-limiting - maybe interval 30 would be slightly better?

.github/workflows/scripts/wait-for-image.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# bash built-in variable
+SECONDS=0
+
+FOUND_TAG=""
+while [ "$SECONDS" -le "$TIME_LIMIT" ]; do
+    FOUND_TAG=$(find_tag "$NAME" "$TAG")
```
vikin91, line 33: We could quote the output to be sure that we compare the entire API reply instead of the first space-separated block.

```suggestion
    FOUND_TAG="$(find_tag "$NAME" "$TAG")"
```

.github/workflows/scripts/wait-for-image.sh
```diff
@@ -0,0 +1,48 @@
...
+    if [ "$FOUND_TAG" = "$TAG" ]; then
+        gh_log notice "Image '$NAME:$TAG' has been found on Quay.io."
+        break
+    fi
+    if [ "$DRY_RUN" == "true" ]; then
```
vikin91, line 38: Could we maybe unify the `=` and `==` in the if-clauses?

.github/workflows/scripts/wait-for-image.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# bash built-in variable
+SECONDS=0
+
+FOUND_TAG=""
+while [ "$SECONDS" -le "$TIME_LIMIT" ]; do
+    FOUND_TAG=$(find_tag "$NAME" "$TAG")
+    if [ "$FOUND_TAG" = "$TAG" ]; then
+        gh_log notice "Image '$NAME:$TAG' has been found on Quay.io."
+        break
```
vikin91, line 36: Maybe simply `exit 0`?

.github/workflows/scripts/wait-for-image.sh
```diff
@@ -0,0 +1,48 @@
...
+    fi
+    echo "Waiting..."
+    sleep "$INTERVAL"
+done
+
+if [ "$FOUND_TAG" != "$TAG" ]; then
+    gh_log error "Image '$NAME:$TAG' has not been found on Quay.io."
+    exit 1
+fi
```
vikin91, line 48: If we would use the `exit 0` variant in line 36, then we could spare here the if-clause and just stay with:

```suggestion
gh_log error "Image '$NAME:$TAG' has not been found on Quay.io."
exit 1
```

vikin91 (approved):
It looks good and would work the way it is, but I found several bits that could be corrected.

vikin91:
I rebased this against master, so that we can pass the checks.

vikin91:
Trying to rebuild without the `turbo-build` label

/retest-required

PR #2798, opened 2022-08-23
ROX-12015: Configure OSCI branch for stackrox upstream build

0x656b694d:
## Description

A new step in "Start Release" automation.

## Checklist

## Testing Performed

Locally and in `test-gh-actions` repository:
https://github.com/stackrox/test-gh-actions/actions/runs/2912915223

### DIFF
--- a/.github/workflows/start-release.yml
+++ b/.github/workflows/start-release.yml
@@ -144,25 +144,23 @@ jobs:
           git switch --create "stackrox-release-$RELEASE"
 
           # Duplicate the main config.
-          yq ".promotion.tag=\"$RELEASE\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
+          yq "del(.tests[] | select(.postsubmit != true or has(\"cron\"))) | .promotion.tag=\"$RELEASE\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
             "$CFG_DIR/stackrox-stackrox-master.yaml" \
             > "${PREFIX}.yaml"
 
           # Duplicate the opensource images config.
-          yq ".promotion.tag=\"$RELEASE-stackrox-branding\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
+          yq "del(.tests[] | select(.postsubmit != true or has(\"cron\"))) | .promotion.tag=\"$RELEASE-stackrox-branding\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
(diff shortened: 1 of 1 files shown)


.github/workflows/start-release.yml
```diff
@@ -109,6 +109,73 @@ jobs:
...
+          # Duplicate the main config.
+          yq ".promotion.tag=\"$RELEASE\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
+            "$CFG_DIR/stackrox-stackrox-master.yaml" \
+            > "${PREFIX}.yaml"
+
+          # Duplicate the opensource images config.
+          yq ".promotion.tag=\"$RELEASE-stackrox-branding\" | .zz_generated_metadata.branch=\"release-$RELEASE\"" \
+            "$CFG_DIR/stackrox-stackrox-master__stackrox_branding.yaml" \
+            > "${PREFIX}__stackrox_branding.yaml"
+
```
gavin-stackrox, line 155: I think you are going to need to `yq` out all the `tests` that have no `postsubmit: true` line and `cron: x` or openshift CI will fail. But I may be wrong and maybe it is ok with deleting the job files after the `make update`. Have you been able to test with a `0.x` release?

vikin91 (commented):
This looks okay at the first glimpse, but I would like to see how you tested it and what was the result - I think there is no harm in creating that branch earlier for some testing release?

vikin91 (approved):
We need to start testing things in some wetter runs :) When are we introducing the weekly upstream releases? 😬

.github/workflows/start-release.yml
0x656b694d: Oh, I didn't test it end-to-end, I just refactored the script you documented and observed `make update` doesn't fail. I'll need your assistance in broader testing.

0x656b694d: The `presubmits` files are still getting generated after cutting out all the 'non-postsumbit' and 'cron' tests with `--target=[images]`. So I'll keep the `rm` step.

0x656b694d: https://github.com/openshift/release/pull/31722

gavin-stackrox:
I'd like to see this run on a test release branch to see what happens. I've used a `0.` prefix as a test release indicator which e.g. sends slack messages to different channels: https://github.com/stackrox/stackrox/blob/4b3c61e55330bade1ad326ab6cd597623da0a29a/scripts/lib.sh#L51

PR #2796, opened 2022-08-23
Fix go.mod file

vladbologa:
## Description

Ran `go mod tidy` to update the `go mod` file.

## Checklist
- [ ] Investigated and inspected CI test results
~~- [ ] Unit test and regression tests added~~
~~- [ ] Evaluated and added CHANGELOG entry if required~~
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed

CI is sufficient

PR #2805, opened 2022-08-23
Prevent timing failures in auth providers integration tests

pedrottimark:
### Failing test

1562110050078560256 from master build for #2796 on 2022-08-23

> Timed out retrying after 4000ms: Expected to find element: `main button:contains("Create")`, but never found it.

`'add SAML 2.0'` in cypress/integration/accessControl/accessControlAuthProviders.test.js

prow-accessControlAuthProviders-saml

### Analysis

Not specific to the test, prerequisite requests to render the page were slow enough that the DOM assertion timed out.

Although it looks like the failing test waits on the correct /v1/authProviders request, the preceding test makes requests including /v1/authProviders which it does not wait on, which opens the door to the next test to get ahead of UI which might have been delayed by finishing rendering the preceding test scenario, plus any unfortunately timed delays in responses from central.

### Solution

Provide several time intervals for requests and assertions to prevent timing failures.

Apply composable convention for helper function arguments that was added in #2176

* `routeMatcherMap` https://docs.cypress.io/api/commands/intercept#Matching-with-RouteMatcher
* `permissionsStaticResponse` and `staticResponseMap` https://docs.cypress.io/api/commands/intercept#StaticResponse-objects

### Changed files

1. Edit cypress/constants/apiEndpoints.js
    * Delete unneeded wildcards from `list` properties for more precise matches and use in template literals.
    * Add `batch` property to `groups` object.

2. Edit cypress/fixtures/auth/mypermissionsMinimalAccess.json
    * Delete obsolete properties.
        https://github.com/stackrox/stackrox/blob/master/proto/api/v1/role_service.proto#L27-L31

3. Edit cypress/helpers/visit.js
    * Add `visitWithPermissions` helper function to test conditional rendering for user role permissions.

4. Edit cypress/integration/accessControl/accessControlAuthProviders.test.js
    * Replace temporary local endpoints with import from constants/apiEndpoints
    * Rewrite local function based on import from helpers/visit
    * Factor out generic assertions from tests into `visitAuthProviders` function to provide an additional time interval before test specific assertions.
    * Move `mockUserCertResponse` argument into the specific test.
    * Rewrite `cy.location` assertions.
    * Uncomment assertion which was waiting for cypress 7 or later.

## Checklist

## Testing Performed

For local deployment: ran each test individually, and then ran all tests in the file.

### DIFF
--- a/ui/apps/platform/cypress/helpers/visit.js
+++ b/ui/apps/platform/cypress/helpers/visit.js
@@ -30,3 +30,26 @@ export function visit(pageUrl, requestConfig, staticResponseMap) {
     cy.wait(['@featureflags', '@mypermissions', '@config/public']);
     waitForResponses(requestConfig);
 }
+
+/*
+ * Visit page to test conditional rendering for user role permissions.
+ *
+ * { body: { resourceToAccess: { … } } }
+ * { fixture: 'fixtures/wherever/whatever.json' }
+ */
+export function visitWithPermissions(
+    pageUrl,
+    permissionsStaticResponse,
+    requestConfig,
+    staticResponseMap
+) {
+    cy.intercept('GET', api.featureFlags).as('featureflags');
+    cy.intercept('GET', api.roles.mypermissions, permissionsStaticResponse).as('mypermissions');
+    cy.intercept('GET', api.system.configPublic).as('config/public');
+    interceptRequests(requestConfig, staticResponseMap);
+
+    cy.visit(pageUrl);
+
+    cy.wait(['@featureflags', '@mypermissions', '@config/public']);
+    waitForResponses(requestConfig);
+}
--- a/ui/apps/platform/cypress/helpers/visit.js
+++ b/ui/apps/platform/cypress/helpers/visit.js
@@ -18,6 +18,10 @@ import { interceptRequests, waitForResponses } from './request';
  * graphqlMultiAliasMap: { opname: { aliases, routeHandler }, … }
  *
  * Optionally wait for responses with waitOptions: { requestTimeout, responseTimeout }
+ *
+ * @param {string} pageUrl
+ * @param {{ routeMatcherMap?: Record<string, { method: string, url: string }>, opnameAliasesMap?: Record<string, (request: Object) => boolean>, waitOptions?: { requestTimeout?: number, responseTimeout?: number } }} [requestConfig]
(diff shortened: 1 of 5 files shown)


ui/apps/platform/cypress/helpers/visit.js
```diff
@@ -30,3 +30,26 @@ export function visit(pageUrl, requestConfig, staticResponseMap) {
...
+ * { fixture: 'fixtures/wherever/whatever.json' }
+ */
+export function visitWithPermissions(
+    pageUrl,
+    permissionsStaticResponse,
+    requestConfig,
+    staticResponseMap
+) {
+    cy.intercept('GET', api.featureFlags).as('featureflags');
+    cy.intercept('GET', api.roles.mypermissions, permissionsStaticResponse).as('mypermissions');
```
vjwilson, line 47: This is worth a dedicated slot in UI meeting, because it seems to be a low-cost way to start testing views which differ by permissions.

ui/apps/platform/cypress/integration/accessControl/accessControlAuthProviders.test.js
```diff
@@ -3,51 +3,57 @@ import {
...
-
     it('displays alert if no permission', () => {
-        cy.intercept('GET', permissionsApi.mypermissions, {
+        const permissionsStaticResponse = {
             fixture: 'auth/mypermissionsMinimalAccess.json',
-        }).as('GetMyPermissions');
-        cy.visit(authProvidersUrl);
-        cy.wait('@GetMyPermissions');
+        };
+        visitWithPermissions(authProvidersUrl, permissionsStaticResponse);
```
vjwilson, line 53: `visitWithPermissions()` takes 4 arguments, but we're only passing 2 parameters here?

vjwilson (commented):
General approach is good, just have question about func signature.

ui/apps/platform/cypress/integration/accessControl/accessControlAuthProviders.test.js
pedrottimark: Good eyes. Yes, the others are optional so tests can call same function:
* mock permissions only
* mock permissions and also wait on requests
* mock permissions and also wait on requests and also mock responses

I added JSDoc `param` types in case people recognize name in square brackets as notation for optional:
* `[requestConfig]`
* `[staticResponseMap]`

vjwilson (approved):

PR #2806, opened 2022-08-24
Add Postgres upgrade test

c-du:
## Description

Add Postgres upgrade test
1. Test rocksdb -> Postgres upgrade
2. Bounce central

## Checklist

## Testing Performed

Run upgrade tests.


tests/upgrade/postgres_run.sh
```diff
@@ -0,0 +1,218 @@
...
+    local log_output_dir="$1"
+
+    require_environment "KUBECONFIG"
+
+    export_test_environment
+
+    REPO_FOR_TIME_TRAVEL="/tmp/rox-postgres-upgrade-test"
+    DEPLOY_DIR="deploy/k8s"
+    QUAY_REPO="rhacs-eng"
+    REGISTRY="quay.io/$QUAY_REPO"
```
dashrews78, line 33: Do we need to put the is_CI check here like the other version of run has?  I'm not sure these are being executed outside of CI though.

    if is_CI; then
        REGISTRY="quay.io/$QUAY_REPO"
    else
        REGISTRY="stackrox"
    fi

tests/upgrade/postgres_run.sh
```diff
@@ -0,0 +1,218 @@
...
+
+    if [[ "$#" -ne 1 ]]; then
+        die "missing args. usage: test_upgrade_paths <log-output-dir>"
+    fi
+
+    local log_output_dir="$1"
+
+    EARLIER_SHA="9f82d2713cfec4b5c876d8dc0149f6d9cd70d349"
+    EARLIER_TAG="3.63.x-163-g2c4fe1563c"
+    #EARLIER_SHA="270d517e9572a8301a62e93d9299bd66c15ef4f8"
```
dashrews78, line 95: nit:  remove these

dashrews78 (approved):
Looks pretty good to me.  just a couple of nits.

tests/upgrade/postgres_run.sh
c-du: We cannot use docker.io any more.

PR #2715, opened 2022-08-16
Fix scope lookups in image and cluster Vulnerabilities sub-resolvers

charmik-redhat:
## Description

Some image and cluster vulnerability sub resolvers used to search the default ctx sent to the sub-resolver. The default ctx would not have the entity scoping added when vulnerabilities were queried from another entity. This PR updates the sub-resolvers in image and cluster vulnerability sub-resolvers to use the parent ctx to lookup scopes.

## Checklist
- [x] Investigated and inspected CI test results
- ~[ ] Unit test and regression tests added~
- ~[ ] Evaluated and added CHANGELOG entry if required~
- ~[ ] Determined and documented upgrade steps~
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

If any of these don't apply, please comment below.

## Testing Performed

Manual testing 
- Check if `fixedByVersion`, `isFixable` and `discoveredAtImage` in ImageVulnerability graphQL work as expected when `imageVulnerabilities` are queries from other entities.
- Check if `fixedByVersion` and `isFixable` in ClusterVulnerability graphQl work as expected.

PR #2812, opened 2022-08-24
Explicitly click to close actions menu in integration test for violations

pedrottimark:
### Test failure

1560369927628525568 from master build for #2715 on 2022-08-18

> should contain correct action buttons for the lifecycle stage

> Timed out retrying after 4050ms: `cy.click()` failed because this element:

> `<button aria-label="Actions" id="pf-dropdown-toggle-id-17" class="pf-c-dropdown__toggle pf-m-plain" type="button" aria-expanded="false" aria-haspopup="true">...</button>`

> is being covered by another element:

> `<button tabindex="-1" data-ouia-component-type="PF4/DropdownItem" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-DropdownItem-2" data-key="1" aria-disabled="false" type="button" class="pf-c-dropdown__menu-item" role="menuitem">Mark as...</button>`

Indeed, the kabob button of the second row is covered by the action menu of the first row.
prow

Although hard to say why it failed this time this way, it seems like a good habit to use most direct interaction to put UI back into a neutral state for each of the 2 sub-tests.
### Changed files

1. Edit cypress/constants/ViolationsPage.js
    * Make selector specific enough not to match the buttons in an open actions menu.
        Initial attempt to click button again to close the menu failed because the selector also matched the `button` elements for action menu items.

2. Edit cypress/integration/violations/violations.test.js
    * Replace indirect with direct interaction to close action menu.

## Checklist
- [x] Edited integration test

## Testing Performed

* For local deployment: ran the test individually, and then ran all tests in the file.
* Inspect elements in readonly demo to see the elements in open and closed action menu.

### DIFF
--- a/ui/apps/platform/cypress/constants/ViolationsPage.js
+++ b/ui/apps/platform/cypress/constants/ViolationsPage.js
@@ -13,7 +13,7 @@ export const selectors = {
     resultsFoundHeader: (number) =>
         `h2:contains("${number} result${number === 1 ? '' : 's'} found")`,
     actions: {
-        btn: 'td.pf-c-table__action button',
+        btn: 'td.pf-c-table__action button[aria-label="Actions"]',
         excludeDeploymentBtn: 'button:contains("Exclude deployment")',
         resolveBtn: 'button:contains("Mark as resolved")',
         resolveAndAddToBaselineBtn: 'button:contains("Resolve and add to process baseline")',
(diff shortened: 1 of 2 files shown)


dvail (approved):
> hard to say why it failed this time this way

True, but even if this was a fluke I agree that closing dropdowns like this should help the tests be more reliable in general.

PR #2814, opened 2022-08-24
Dedupe dackbox index writes

connorgorman:
## Description

Hashing the objects going to be indexed in dackbox should significantly alleviate load on the indexer. Something similar will need to be used for Postgres, but that can be done within Postgres

## Checklist

## Testing Performed

Will run in CI and look at the metrics

### DIFF
--- a/pkg/dackbox/indexer/lazy.go
+++ b/pkg/dackbox/indexer/lazy.go
@@ -1,6 +1,7 @@
 package indexer
 
 import (
+	"encoding/binary"
 	"fmt"
 	"hash"
 	"hash/fnv"
@@ -10,7 +11,6 @@ import (
 	"github.com/blevesearch/bleve"
 	"github.com/blevesearch/bleve/document"
 	"github.com/gogo/protobuf/proto"
-	"github.com/mitchellh/hashstructure"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/pkg/dackbox/utils/queue"
 	"github.com/stackrox/rox/pkg/logging"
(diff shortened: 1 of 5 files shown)


connorgorman:
5k images:
```
rox_central_dackbox_index_objects_deduped 2.425583e+06
rox_central_dackbox_index_objects_indexed 573601
```
This was actually not on the document like I have the PR now, but on the actual proto.Message which is theoretically worse but considering the data was static still representative

pkg/dackbox/indexer/lazy.go
```diff
@@ -83,6 +91,23 @@ func (li *lazyImpl) runIndexing() {
 	}
 }
 
+func (li *lazyImpl) isDeduped(doc *document.Document) bool {
```
misberner, line 94: Nit:  the naming is off. `isDuplicate` would be more adequate, but `is...` also suggests a side-effect freeness the this medic doesn't provide. Don't have a good idea off the top of my head on his to name this instead though

misberner (approved):

pkg/dackbox/indexer/lazy.go
connorgorman: Yeah, agreed. Let me think of a better name because I don't like the fact that this function modifies something. I should also note that the hashing of the doc itself is a bit of a tradeoff in terms of increasing the likelihood of deduping while potentially paying the cost of doc mapping (which isn't free). However, the true solution would be to hash fields with search labels that aren't `-`, but thats more involved then what I want to do here

connorgorman:
CPU is largely the same
```
process_cpu_seconds_total  (old: 6232.68, new 6244.48): change: 0.1893%
```

On QA run, which is a pretty bad proxy because we delete the images a bunch
```
dackbox_index_objects_deduped                                                    12344
dackbox_index_objects_indexed                                                    90737
```

```
bleve_index_CurOnDiskBytes Index=combined-persisted (old: 118780812.00, new 90645390.00): change: -23.6868%
```

Scaled up
```
rox_central_dackbox_index_objects_deduped 4.375776e+06
rox_central_dackbox_index_objects_indexed 2.225394e+06
```

pkg/dackbox/indexer/lazy.go
```diff
@@ -83,6 +91,44 @@ func (li *lazyImpl) runIndexing() {
...
+	}
+	for _, f := range doc.Fields {
+		hashableDoc.Fields = append(hashableDoc.Fields, hashableField{
+			Name:           f.Name(),
+			Value:          f.Value(),
+			ArrayPositions: f.ArrayPositions(),
+		})
+	}
+	hashValue, err := hashstructure.Hash(hashableDoc, &hashstructure.HashOptions{
+		Hasher: li.hasher,
```
misberner, line 118: Just double-checking that this is never called concurrently? Because it will always use the same hasher instance

pkg/dackbox/indexer/lazy.go
```diff
@@ -83,6 +91,44 @@ func (li *lazyImpl) runIndexing() {
...
+	ID     string
+	Fields []hashableField
+}
+
+func (li *lazyImpl) evaluateDeduping(doc *document.Document) bool {
+	hashableDoc := hashableDoc{
+		ID:     doc.ID,
+		Fields: make([]hashableField, 0, len(doc.Fields)),
+	}
+	for _, f := range doc.Fields {
```
misberner, line 110: Just curious, is `doc.Fields` order deterministic? I wonder if you really need hashstructure and couldn't just compute the hash on the fly, if performance is a concern. It's just a sequence of writes to the hasher, no?

misberner (approved):

pkg/dackbox/indexer/lazy.go
connorgorman: Yeah, never called concurrently. There is a single worker that consumes them

pkg/dackbox/indexer/lazy.go
connorgorman: doc.Fields is deterministic
```
		hasher.Write([]byte(doc.ID))
		for _, f := range doc.Fields {
			hasher.Write([]byte(f.Name()))
			hasher.Write(f.Value())
			positionBytes := make([]byte, 8)
			for _, pos := range f.ArrayPositions() {
				binary.LittleEndian.PutUint64(positionBytes, pos)
				hasher.Write(positionBytes)
			}
		}
		hasher.Sum64()
```
Could replace this and be much faster than hashstructure. ~35-40x faster

PR #2817, opened 2022-08-24
[ROX-12303][POSTGRES] Change the string search from prefix matching to substring matching

md2119:
## Description
Change the string search from prefix matching to substring matching for all search labels. Due to analyzers that we use with bleve indexer, some fields are prefixed matched whereas others are substring matched (esp those with whitespaces). Make substring matching defacto string search technique and remove the inconsistency in Postgres.

## Checklist
- ~Documented user-facing changes~

## Testing Performed
Unit


md2119:
ACS 4.0 w/ postgres as default DB is already shipped with default prefix matching-based search. Changing the search pattern must either have critical customer feedback or product requirement. I don't believe we have either. Hence, closing this ticket.

PR #2820, opened 2022-08-24
Support multi-arch binaries

yselkowitz:
## Description

Make all binary paths and names contain both OS and ARCH, as a prerequisite of multi-architecture enablement.

## Checklist

## Testing Performed

Relying on CI to verify.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -343,35 +343,39 @@ endif
 
 .PHONY: build-prep
 build-prep: deps
-	mkdir -p bin/{darwin,linux,windows}
+	mkdir -p bin/{darwin_amd64,linux_amd64,linux_ppc64le,linux_s390x,windows_amd64}
 
 .PHONY: cli-build
 cli-build: build-prep
 	RACE=0 CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 $(GOBUILD) ./roxctl
 	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GOBUILD) ./roxctl
+	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le $(GOBUILD) ./roxctl
+	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=s390x $(GOBUILD) ./roxctl
 ifdef CI
-	RACE=0 CGO_ENABLED=0 GOOS=windows $(GOBUILD) ./roxctl
+	RACE=0 CGO_ENABLED=0 GOOS=windows GOARCH=amd64 $(GOBUILD) ./roxctl
 endif
 
 .PHONY: cli
 cli: cli-build
 	# Copy the user's specific OS into gopath
-	cp bin/$(HOST_OS)/roxctl $(GOPATH)/bin/roxctl
+	cp bin/$(HOST_OS)_$(GOARCH)/roxctl $(GOPATH)/bin/roxctl
 	chmod u+w $(GOPATH)/bin/roxctl
 
 cli-linux: build-prep
 	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GOBUILD) ./roxctl
+	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le $(GOBUILD) ./roxctl
+	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=s390x $(GOBUILD) ./roxctl
 
 cli-darwin: build-prep
 	RACE=0 CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 $(GOBUILD) ./roxctl
 
-upgrader: bin/$(HOST_OS)/upgrader
+upgrader: bin/$(HOST_OS)_$(GOARCH)/upgrader
 
-bin/%/upgrader: build-prep
-	GOOS=$* $(GOBUILD) ./sensor/upgrader
+bin/$(HOST_OS)_$(GOARCH)/upgrader: build-prep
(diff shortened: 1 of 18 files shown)


yselkowitz:
could you `/ok-to-test` please
cc @robbycochran

Makefile
```diff
@@ -572,7 +576,7 @@ docker-build-data-image: docs-image
 
 .PHONY: docker-build-roxctl-image
 docker-build-roxctl-image:
-	cp -f bin/linux/roxctl image/roxctl/roxctl-linux
+	cp -f bin/linux_$(GOARCH)/roxctl image/roxctl/roxctl-linux
```
misberner, line 579: This is referenced as `roxctl-linux-amd64` in `image/roxctl/Dockerfile`, l.4. Curious why this doesn't fail the build - don't we invoke this target?

ui/apps/platform/src/Containers/MainPage/Header/CLIDownloadMenu.tsx
```diff
@@ -37,25 +37,39 @@ function CLIDownloadMenu({ addToast, removeToast }: CLIDownloadMenuProps): React
 
     const appLauncherItems = [
         <ApplicationLauncherItem
-            key="app-launcher-item-cli-mac"
+            key="app-launcher-item-cli-mac-amd64"
```
misberner, line 40: It's a bit ugly that all CLI variants are stored in two places. @vjwilson WDYT? @yselkowitz can I push to your PR to clean this up?

ui/apps/platform/src/services/CLIService.js
```diff
@@ -8,9 +8,9 @@ import { saveFile } from 'services/DownloadService';
 export default function downloadCLI(type) {
     let name = 'roxctl';
     let suffix = type;
-    if (type === 'windows') {
+    if (type === 'windows-amd64') {
```
misberner, line 11: not your code, but given that who knows what archs we support in the future, I would prefer the more general
```
if (type.startsWith('windows-')) {
  name += 'exe';
  suffix += 'exe';
}
```
up to you

tests/upgrade/run.sh
```diff
@@ -401,8 +401,8 @@ deploy_earlier_central() {
     info "Deploying: $EARLIER_TAG..."
 
     mkdir -p "bin/$TEST_HOST_OS"
-    gsutil cp "gs://stackrox-ci/roxctl-$EARLIER_TAG" "bin/$TEST_HOST_OS/roxctl"
-    chmod +x "bin/$TEST_HOST_OS/roxctl"
+    gsutil cp "gs://stackrox-ci/roxctl-$EARLIER_TAG" "bin/${TEST_HOST_OS}_amd64/roxctl"
+    chmod +x "bin/${TEST_HOST_OS}_amd64/roxctl"
     PATH="bin/$TEST_HOST_OS:$PATH" command -v roxctl
```
misberner, line 406: You need to adjust these lines to `${TEST_HOST_OS}_amd64` as well. The problem is that we copy the binary to `linux_amd64`, but we don't add it to the path, so the deploy scripts don't find the roxctl binary and thus assume we need to run docker to invoke roxctl, which doesn't work in CI. This is the cause of the upgrade tests failure.

misberner (changes requested):
This looks pretty good overall! I think the upgrade tests failure is real and I have commented below on what I think needs to be done.

What is still missing is the publishing logic (i.e., uploading the CLI to stackrox.io and openshift mirror). You find the relevant scripts in `scripts/ci/roxctl-publish/` (@gavin-stackrox can you briefly confirm that this is still the relevant location, even in the OSCI world?), and I think that `prepare.sh` is the one of interest, as `publish.sh` just works on the directory structure left behind by `prepare.sh`.

I would suggest to change things in such a way that for now, amd64 binaries without the arch suffix are still published in addition to the arch-specific names. The reason is that download links are referenced in the docs, and we cannot upgrade the docs atomically with releasing the new version. Therefore, make sure we have a transition period in which we can fix the docs (note to self: check if there's a JIRA for doc changes).

ui/apps/platform/src/Containers/MainPage/Header/CLIDownloadMenu.tsx
vjwilson: @misberner 
The `key` properties are an affordance for the React templating system. They just need to be unique IDs. In fact, let's not try to keep them in sync with variant in the click handler--that's gilding a lily. Just make them the type of system

For example,
* mac
* linux-amd
* linux-ppc

misberner: Oh, I was referring to it being referenced in the other file as well.

misberner (commented):
@yselkowitz FYI, we only allow merging PRs via "Rebase & squash" to this repo, so you can accumulate an arbitrary number of commits on your branch with no effect on the `master` branch post merging. I would suggest to force push only when rebasing, as otherwise I lose the ability to do incremental reviews on GitHub.

gavin-stackrox:
@misberner yes, the scripts in `scripts/ci/roxctl-publish` are still used to publish roxctl on release.

misberner:
@yselkowitz please rebase to fix CI failures

misberner:
@yselkowitz FYI I've for now decided against pushing the feature flags changed against your branch, I think we can easily take care of it at a later point. Let's focus on getting this merged first.

misberner:
Also, the changes look good to me! Let's see what CI says..

misberner (approved):

misberner:
Awesome. All tests passed. 🚢 it

misberner:
@yselkowitz tbh, I don't know how our repo is set up. Can you merge this yourself, or do you need me to do it?

PR #2821, opened 2022-08-24
ROX-12293: Disable DB backup and restore for managed central

clickboo:
## Checklist


connorgorman (approved):
I do think it'd be cleaner to add it as an option in the custom route, but this works

PR #2830, opened 2022-08-25
Dashrews/remove gorm source log statement

dashrews78:
## Description
Redact the password from the logging statement.

## Checklist
~- [ ] Investigated and inspected CI test results~

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/migrator/postgreshelper/load.go
+++ b/migrator/postgreshelper/load.go
@@ -71,7 +71,7 @@ func Load(conf *config.Config, databaseName string) (*pgxpool.Pool, *gorm.DB, er
 					return err
 				}
 			}
-			log.WriteToStderrf("connect to gorm: %v", gormSource)
+
 			gormDB, err = gorm.Open(postgres.Open(gormSource), &gorm.Config{
 				NamingStrategy:  pgutils.NamingStrategy,
 				CreateBatchSize: 1000})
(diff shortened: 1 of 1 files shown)


migrator/postgreshelper/load.go
```diff
@@ -71,7 +71,7 @@ func Load(conf *config.Config, databaseName string) (*pgxpool.Pool, *gorm.DB, er
 					return err
 				}
 			}
-			log.WriteToStderrf("connect to gorm: %v", gormSource)
+
```
theencee, line 74: It's useful to have the log statement though. At least keep the log line with the host and db name, but ideally it can be printed as-is with the password redacted.

migrator/postgreshelper/load.go
```diff
@@ -71,7 +72,8 @@ func Load(conf *config.Config, databaseName string) (*pgxpool.Pool, *gorm.DB, er
 					return err
 				}
 			}
-			log.WriteToStderrf("connect to gorm: %v", gormSource)
+			log.WriteToStderrf("connect to gorm: %v", strings.Replace(gormSource, adminConfig.ConnConfig.Password, "<REDACTED>", -1))
```
dashrews78, line 75: cockroach has a redact package that looked interesting, but seemed like it might be a bit much to pull in for this since I could just do a simple replace, since strings.Replace returns an updated copy of the source string.

janisz (approved):

theencee (approved):

PR #2831, opened 2022-08-25
Ensure that the postgres env var is reset in tests

md2119:
## Description

I scrubbed `central/` to verify if we were resetting the feature flags in all unit tests...and we did not.
 
## Checklist

## Testing Performed
Unit


dashrews78 (approved):
LGTM

PR #2839, opened 2022-08-26
ROX-12336: Harden SCCs for sensor to prevent running as root or privilege escalation

vikin91:
## Description

This PR hardens the security for sensor.
Based on the sensor-SCCs (that are optional in the installation), sensor was allowed to escalate the privileges from user 4000 (default) to 0 (root).
By changing this, we would most probably pass the validation that runs for `restricted-v2` SCC as defined [here](https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth).

The SCC being changed here was introduced [in this PR](https://github.com/stackrox/rox/pull/5129) most probably based on linter suggestions.
Creation of SCC resources was made optional in https://github.com/stackrox/rox/pull/8686.

## Checklist

## Testing Performed

- [x] Manually changing SCC on OpenShift 4.11, recreating the pod and verifying that sensor still works
- [x] CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ authentication is still required for an email notifier, but the user can now cho
 - ROX-11181: Any clusters that have been unhealthy (defined as central being unable to reach sensor running on those clusters) for a configured period of time will be automatically removed. The number of days after which an 'unhealthy' cluster is removed can be configured in the System Configuration page or using the cluster API.
   - Any cluster that is expected to be unavailable for a period of time (e.g. clusters used in disaster recovery), can be tagged with a customizable label. Clusters with those labels will never be removed automatically.
   - By default, this unhealthy cluster removal is disabled (number of days set to 0)
-- ROX-12336: If sensor is installed with the optional `SecurityContextConstraint`s (SCC), they now explicitly disallow now running as root and priviledges escalation.
+- ROX-12336: If sensor is installed with the optional `SecurityContextConstraint`s (SCC), they now explicitly disallow: running as root and priviledges escalation.
 
 ## [3.71.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Entries in this file should be limited to:
 Please avoid adding duplicate information across this changelog and JIRA/doc input pages.
 
 ## [NEXT RELEASE]
-- ROX-11348: The email notifier now allows for unauthenticated SMTP. By default, 
+- ROX-11348: The email notifier now allows for unauthenticated SMTP. By default,
 authentication is still required for an email notifier, but the user can now choose to turn it off.
 
 ### Removed Features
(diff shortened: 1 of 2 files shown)


mtesseract:
Hi. I am not too familiar with SCCs. What is the reason in the first place that we have a custom SCC instead of simply using e.g. the `restricted-v2` SCC?

Besides, if this hardening would pose any risk I would believe that this would be easy to catch in our test suite (because it would most likely render all applications for which this SCC is configured as dysfunctional, correct?). Do we have sufficient test coverage among different OpenShift versions for such a change?

porridge (commented):
This LGTM, except an "obscure side effect" of this change is that if any other customer's workload happened to be bound by this SCC (by mistake/conicidence), then this hardening could cause their workload to break.
Such things did happen in the past I think (cc @misberner ) so I think a note in the changelog is due.

vikin91:
> Hi @vikin91. I am not too familiar with SCCs. What is the reason in the first place that we have a custom SCC instead of simply using e.g. the `restricted-v2` SCC?

I have tried to find the reason for custom SCCs and found only this what I mentioned in the PR description - most probably, we added them based on a linter suggestion and never looked at them again until [ROX-9342
](https://issues.redhat.com/browse/ROX-9342). In [ROX-9342
](https://issues.redhat.com/browse/ROX-9342), the reporter suggested that if we create a cluster with custom SCCs, then the SCCs suggest that sensor may run as root or may gain root privileges. In the practice this is not the case, as we protect ourselves with additional settings `readOnlyRootFilesystem: true` `runAsNonRoot: true` in the deployment definition. I thought that we could express this in the custom SCC.

I have introduced here the minimal changes to the previous state to achieve this, but I can also give it a try with simply using `restricted-v2` directly. However, there may be an issue related to that change, as `restricted-v2` was introduced in OpenShift 4.11, before we had only `restricted` available.

> Besides, if this hardening would pose any risk I would believe that this would be easy to catch in our test suite (because it would most likely render all applications for which this SCC is configured as dysfunctional, correct?). Do we have sufficient test coverage among different OpenShift versions for such a change?

I think, we do not have any tests for this, as the original issue was discovered by Sam Fowler who reported [ROX-9342
](https://issues.redhat.com/browse/ROX-9342) and not by our test suites. The change itself should work along all versions of openshift as long as we stay with custom SCC (`restricted-v2` was introduced only recently).

vikin91:
> This LGTM, except an "obscure side effect" of this change is that if any other customer's workload happened to be bound by this SCC (by mistake/conicidence), then this hardening could cause their workload to break. Such things did happen in the past I think (cc @misberner ) so I think a note in the changelog is due.

The creation of these SCCs remains optional, to only a subset would be affected. But I agree about the changelog entry - will add one in a moment.

vikin91:
Closing as the proper changes will be done in https://github.com/stackrox/stackrox/pull/2928 and its follow-ups.

PR #2844, opened 2022-08-26
Dashrews/ROX-11357 process baseline issue with failure on slow environments

dashrews78:
## Description

The crux of the problem here appears to be speed.  Looks like the orchestrator is trying to execute the ls before the deployment is ready.  So I made a couple of adjustments.  

1. utilize the waitForDeployment call to hopefully ensure the deployment is ready before the exec call.
2. Move the validation of the container name being in the baseline until after the baseline is out of observation.  that just buys a little more time for the process indicator to come in and get processed.  And really, moving it simplifies the test as we were waiting on it to come out of observation anyway.

## Checklist
~- [ ] Investigated and inspected CI test results~

## Testing Performed

It is a test.  CI is sufficient.


gavin-stackrox (approved):
LGTM

PR #2846, opened 2022-08-26
ROX-12339: Untangle entities column in fixable CVEs tables (with updated queries)

vjwilson:
## Description

* Primary goal: show correct Entities column data in fixable CVEs table in Image Component Overview (and verified that other Entities columns still work, with and without feature flag ON)
* Piggybacking simplified queries for split CVE and split COMPONENT types, after brainstorming with Mandar 

## Checklist

## Testing Performed

Example on Image Component Overview, fixable CVEs table with correct entities
FixableCVEsTable_Component_correct_entities

Image CVE query
ImageCVE_query_with_CVE_ID

Node CVE query
NodeeCVE_query_with_CVE_ID

Cluster CVE query
ClusterCVE_query_with_CVE_ID

Image Component query
ImageCOMPONENT_query_with_COMPONENT_ID

Node Component query
NodeCOMPONENT_query_with_COMPONENT_ID

### DIFF
--- a/ui/apps/platform/src/Components/workflow/TableCountLinks/TableCountLinks.tsx
+++ b/ui/apps/platform/src/Components/workflow/TableCountLinks/TableCountLinks.tsx
@@ -3,6 +3,7 @@ import React, { ReactElement, useContext } from 'react';
 import entityTypes, { ResourceType, resourceTypes } from 'constants/entityTypes';
 import TableCountLink from 'Components/workflow/TableCountLink';
 import workflowStateContext from 'Containers/workflowStateContext';
+import fixableVulnTypeContext from 'Containers/VulnMgmt/fixableVulnTypeContext';
 
 type TableCountLinksProps = {
     row: {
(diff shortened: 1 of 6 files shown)


md2119 (approved):
The search query changes look good. For the entities col changes, you might want to check deployments, namespace and other pages as well.

ui/apps/platform/src/utils/queryService.test.js
```diff
@@ -102,37 +100,37 @@ describe('queryService.', () => {
...
             const entityContext = {
-                CVE: 'CVE-2020-8554#K8S_CVE',
+                CLUSTER_CVE: 'CVE-2020-8554#K8S_CVE',
             };
 
             const queryObject = queryService.entityContextToQueryObject(entityContext);
 
             expect(queryObject).toEqual({
-                CVE: 'CVE-2020-8554#K8S_CVE',
+                'CVE ID': 'CVE-2020-8554#K8S_CVE',
```
sachaudh, line 133: I noticed in the PR we are changing `CVE` to `CVE ID` and `COMPONENT` to `COMPONENT ID` but then I also see `CLUSTER_CVE` and `NODE_CVE`. Doesn't seem like we have a consistent pattern for these search queries. @md2119 do we have plans to address stuff like that in the near future?

ui/apps/platform/src/Components/workflow/TableCountLinks/TableCountLinks.tsx
```diff
@@ -15,9 +16,12 @@ type TableCountLinksProps = {
         id: string;
     };
     textOnly: boolean;
+    // fixableVulnType?: typeof entityTypes[keyof typeof entityTypes] | '' | null | undefined;
```
sachaudh, line 19: Do we need to keep this? If not, maybe we can delete

sachaudh (approved):
LGTM. Left a comment about inconsistent search query syntax 🤔

ui/apps/platform/src/utils/queryService.test.js
vjwilson: Mandar and I went over this face-to-face, because it was so confusing, and we decided 'CVE ID' was the best option for all three types. (Search will get a some major improvements in the "Vulnerability Management Re-imagined" epic.)

PR #2849, opened 2022-08-26
[ROX-12357] Follow up on suppress API change

md2119:
## Description
Make the change that was notified in `71.0`.

Prior to Postgres, we did not have separate ID and CVE field for vulns. We only had `id` field that stored CVE name (public identifer). In Postgres, CVE `id` differs from CVE name `cve`. 
We keep the definition of _**this CVE**_ the same with or without Postgres, for suppress and unsuppress operation. Therefore, with Postgres we act on `cve` not `id`. This PR makes those follow-up changes to make the API clear.
In short, from user's perspective this PR changes only field name in suppress and unsuppress API and not the behavior.

**Update 09/15/2022:**
**This PR performs the remainder changes to complete the rename. The frontend changes were pushed in 72.0 via #2898.**

## Checklist

## Testing Performed

**with postgres flag:**

Make sure that all instances of Image CVE are suppressed.

image

image

Make sure all instances of Image CVE are unsuppressed.

image

image

Supppress node CVEs
image

Unsuppress node cve
image
image

Platform CVEs:
image

image

image

**without postgres and vm feature flag**

suppress
image

image

unsuppress
image

image


connorgorman:
Question on this: should we have the ability to suppress a vuln by full ID? In the case that a vuln is not going to be fixed for a certain OS, but would be fixed in a later one. Like rhel 7 suppress, but keep for rhel 9?

md2119:
> Question on this: should we have the ability to suppress a vuln by full ID? In the case that a vuln is not going to be fixed for a certain OS, but would be fixed in a later one. Like rhel 7 suppress, but keep for rhel 9?

This certainly could be added in future, likely in VM2.0. However, for the current features—legacy suppression and risk acceptance, we will need to take this route.

theencee:
Was the deprecation notice put in 71 to remove in 72 or 71 to remove in 73? Excepting extenuating circumstances, we want to give 2 releases for deprecations

Also this is not in the published release notes: https://docs.openshift.com/acs/3.71/release_notes/371-release-notes.html

md2119:
> Was the deprecation notice put in 71 to remove in 72 or 71 to remove in 73? Excepting extenuating circumstances, we want to give 2 releases for deprecations
> 
> Also this is not in the published release notes: https://docs.openshift.com/acs/3.71/release_notes/371-release-notes.html

@kcarmichael08 I think some deprecations were missed out from `71.0` public release notes.
```
- `ids` field in `/v1/cves/suppress` and `/v1/cves/unsuppress` API payload will be renamed to `cves` in 72.0 release.
- `cves.ids` field of `storage.VulnerabilityRequest` object, which is in the response of `VulnerabilityRequestService` endpoints, will be renamed to `cves.cves` in 72.0 release.
```

vjwilson (approved):
UI changes look good.

kcarmichael08:
@md2119 @theencee I had put in the snoozed/unsnoozed changes - but missed this. How do you want to handle? We could do an update of the .71 Release Notes. If we do the normal change period, though, it would still be deprecated in .72 and then removed in .73, even with a changed .71 Release Notes.

Sorry it was missed - I think if we end up changing the format of the changelog, like we discussed (with headers for deprecation, etc.), it will help us better identify what needs to go in the RNs.

md2119:
@connorgorman @theencee, the PR is ready for review.

connorgorman (approved):

PR #2853, opened 2022-08-29
Add hash indexes to all vulnerability inner join columns

connorgorman:
## Description

Adding hash index speeds up the inner joins

## Checklist

## Testing Performed

Ran on a scaled up cluster and the CVEs page:

Before: `real	0m2.676s`
After: `real	0m0.493s`


md2119 (approved):
Weirdly, my go benchmark had shown slower performance with indexes, hence I skipped adding indexes. But that was way back and the performance could be rooted in some other issue.

Do you think you can capture the performance after indexing fks in other entities? For example, image id, namespace id, cluster id in deployment, etc. These are frequently joined entities in VM.

connorgorman:
> Do you think you can capture the performance after indexing fks in other entities? For example, image id, namespace id, cluster id in deployment, etc. These are frequently joined entities in VM.

Will do this in a follow up 👍  Need to have SAC there too as those will be very commonly joined

PR #2855, opened 2022-08-29
ROX-12309: Substitute ROX_IMAGE_FLAVOR based on ROX_PRODUCT_BRANDING.

porridge:
## Description

This is so that we can get rid of the `ROX_IMAGE_FLAVOR` variable in the CI definition file.

In CI building the `Dockerfile.gen` will happen during the operator artifacts build.

## Checklist

## Testing Performed

1. Tested locally that `make -C operator docker-build` works, both with clean anv and with `ROX_PRODUCT_BRANDING=RHACS_BRANDING` and that the resulting `Env` shown by `docker image inspect` is as expected.
2. Relying on CI to ensure nothing broken. Note that the path to the original `Dockerfile` is still used at the moment. The `build/Dockerfile.gen` will start getting used after [a change](https://github.com/openshift/release/pull/31762) to `openshift/release` which will be rehearsed separately.

### DIFF
--- a/.openshift-ci/build/build-operator-artifacts.sh
+++ b/.openshift-ci/build/build-operator-artifacts.sh
@@ -24,9 +24,9 @@ build_operator_bundle_and_binary() {
     # Otherwise we get the cryptic "Missing value for flag: -I" error when running protoc.
     go mod download
 
-    info "Preparing bundle sources and smuggled status.sh file"
+    info "Preparing build/Dockerfile.gen, bundle sources and smuggled status.sh file"
     # TODO(ROX-11889): get rid of the SILENT= once we gain some confidence (after release 3.72?)
-    make -C operator bundle bundle-post-process smuggled-status-sh SILENT=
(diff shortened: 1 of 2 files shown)


kovayur (dismissed):
LGTM

kovayur (commented):
I see there is a resolution of `ROX_IMAGE_FLAVOR` from `ROX_PRODUCT_BRANDING` in the root `Makefile`. But how it gets loaded to the operator’s `Makefile`?

I tried to change the variable but I think it doesn’t work as expected, it should change the `Dockerfile.gen`:
```
$ export ROX_PRODUCT_BRANDING=RHACS_BRANDING
$ make build/Dockerfile.gen
make: `build/Dockerfile.gen' is up to date.
```

kovayur:
> But how it gets loaded to the operator’s Makefile?
https://github.com/stackrox/stackrox/blob/master/operator/Makefile#L101

It's still unclear how to cope with "make: `build/Dockerfile.gen' is up to date." but it is not critical I think

kovayur (approved):

kovayur (approved):

PR #2859, opened 2022-08-29
Dashrews/ROX-12256 remove connections before create db

dashrews78:
## Description

In Postgres you cannot copy (via a create_db with a template) or rename a database if there are open connections on the source database.  Since those operations are being performed throughout the migration process, it is not really valid to be connected to those database at this time.  This PR will terminate any open connections on a source database before we try to copy it or move it.  

## Checklist

## Testing Performed

TODO(replace-me)
Moved a test file and renamed it to be admin_utils_test.go.  Then expanded it to cover cases to ensure connections were being terminated prior to copy and rename functions being completed.

### DIFF
--- a/pkg/postgres/pgadmin/admin_utils.go
+++ b/pkg/postgres/pgadmin/admin_utils.go
@@ -190,7 +190,7 @@ func AnalyzeDatabase(config *pgxpool.Config, dbName string) error {
 
 // terminateConnection - terminates connections to the specified database
 func terminateConnection(config *pgxpool.Config, dbName string) error {
-	log.Infof("terminateConnection - %q", dbName)
+	log.Debugf("terminateConnection - %q", dbName)
 
 	// Connect to different database for admin functions
 	connectPool := GetAdminPool(config)
@@ -199,7 +199,7 @@ func terminateConnection(config *pgxpool.Config, dbName string) error {
 
 	_, err := connectPool.Exec(context.Background(), terminateConnectionStmt, dbName)
 
-	log.Info("terminateConnection done")
+	log.Debug("terminateConnection done")
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -76,6 +82,13 @@ func CreateDB(sourceMap map[string]string, adminConfig *pgxpool.Config, dbTempla
 
 	SetPostgresCmdEnv(cmd, sourceMap, adminConfig)
 
+	// terminate connections to the source database.  You cannot copy from a database if
+	// there are open connections to it.
+	err := terminateConnection(adminConfig, dbTemplate)
```
connorgorman, line 87: Since we have the pool that is currently active, any reason we shouldn't call close on it? This will prevent new connections from being started right?

connorgorman: Also maybe we could throw back an error if someone is logged in despite this call?

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -32,6 +32,12 @@ const (
 	postgresTimeBetweenRetries = 10 * time.Second
 
 	getCloneStmt = "SELECT datname FROM pg_catalog.pg_database WHERE datname ~ '^%s_.*'"
+
+	// terminateConnectionStmt - terminates connections to the specified database
+	terminateConnectionStmt = "SELECT pg_terminate_backend(pg_stat_activity.pid) " +
+		"FROM pg_stat_activity " +
+		"WHERE datname = $1 " +
+		" AND pid <> pg_backend_pid();"
```
c-du, line 40: It looks like no need to have the space before "AND"

pkg/postgres/pgadmin/admin_utils.go
dashrews78: @connorgorman 
That begs a question of whether I'm over protecting us here.  When we make a copy of a non-empty database we are running exclusively from the migrator.  Central has not created its pool at that time.  In the migrator we create pools when we need them and close them because we are moving stuff around.  Where I've seen this issue and what I was trying to protect itself is the unknown connection.  For example, this issue occurred in the demo because I was in psql.  That is the only time I've seen it and what I was trying to protect against.  I do not we can have a connection to a database we are copying from through execution of the code, as it is written today.  So it seemed like a reasonable thing to do, but is it a valid scenario?

connorgorman (approved):

PR #2860, opened 2022-08-29
Postgres end-to-end tests - Fix VulnScanWithGraphQLTest GraphQL queries

rhybrillou:
## Description

With the re-design of the dackbox graph for the postgres migration, some graphQL queries in the end-to-end tests need to be adapted.
Here, the focus is on the VulnScanWithGraphQLTest.groovy test file.

In order to make debugging easier, additional logs were added to the GraphQL service, and a minor framework adjustment was also introduced (add a function to tell whether the current run is a Postgres run or not).

## Checklist

## Testing Performed

Test code fixes, CI run should be sufficient.

### DIFF
--- a/qa-tests-backend/src/test/groovy/BaseSpecification.groovy
+++ b/qa-tests-backend/src/test/groovy/BaseSpecification.groovy
@@ -382,6 +382,10 @@ class BaseSpecification extends Specification {
         orchestrator.deleteSecret("gcr-image-pull-secret", Constants.ORCHESTRATOR_NAMESPACE)
     }
 
+    static Boolean isPostgresRun() {
+        return Env.CI_JOBNAME.contains("postgres")
+    }
+
     static Boolean isRaceBuild() {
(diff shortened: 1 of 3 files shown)


qa-tests-backend/src/test/groovy/BaseSpecification.groovy
```diff
@@ -382,6 +382,10 @@ class BaseSpecification extends Specification {
         orchestrator.deleteSecret("gcr-image-pull-secret", Constants.ORCHESTRATOR_NAMESPACE)
     }
 
+    static Boolean isPostgresRun() {
```
janisz, line 385: ❤️ 
Could you replace all `Env.CI_JOBNAME.contains("postgres")` with this method.

janisz: Could be in another PR

janisz (approved):

qa-tests-backend/src/test/groovy/BaseSpecification.groovy
rhybrillou: Will do in a separate PR.

PR #2862, opened 2022-08-29
Disable updates for patch releases updates for aws-sdk-go

janisz:
## Description

aws-sdk-go has many patch releases that are noisy. Let's update only on security updates and minor and major releases.

Refs: https://github.com/stackrox/stackrox/pull/2854#issuecomment-1230096187

## Testing Performed

CI


misberner (approved):
Many thanks!

PR #2863, opened 2022-08-29
QA end-to-end tests : Centralize postgres run identification logic

rhybrillou:
## Description

Some end-to-end tests require knowledge of whether the current run targets a central instance running against postgres.
This knowledge can be used to disable some tests that are not ready yet to run against postgres, or to adjust the test flow to either target specific endpoints, or to adjust request elements passed to Central.
The change at hand reduces the number of locations in the test code where this logic (currently check of an environment variable) appears.

## Checklist

## Testing Performed

Test logic changes. CI run should be sufficient.


qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
```diff
@@ -217,11 +217,7 @@ class AdmissionControllerTest extends BaseSpecification {
         "Suppress CVE and check that the deployment can now launch"
 
         def cve = "CVE-2019-3462"
-        if (Env.CI_JOBNAME.contains("postgres")) {
-            CVEService.suppressImageCVE(cve)
-        } else {
-            CVEService.suppressCVE(cve)
-        }
+        CVEService.suppressImageCVE(cve)
```
janisz, line 220: I believe this is not a part of centralize postgres run identification logic

qa-tests-backend/src/main/groovy/services/CVEService.groovy
```diff
@@ -9,6 +9,10 @@ import io.stackrox.proto.api.v1.ImageCVEServiceGrpc
 import util.Env
 
 class CVEService extends BaseService {
+    static Boolean isPostgresRun() {
```
janisz, line 12: I think it's not needed here as it's defined in `BaseService` as well

qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
rhybrillou: The `if (Env.CI_JOBNAME.contains("postgres")) {...} else {...}` block actually duplicate the code of `CVEService.suppressImageCVE` (same for unsuppress).
I can move the change somewhere else if this is too confusing here.

qa-tests-backend/src/main/groovy/services/CVEService.groovy
rhybrillou: As far as I can tell, this function is only defined here and in `qa-tests-backend/src/test/groovy/BaseSpecification.groovy`.

rhybrillou:
The postgres migration is finished, and the pre-postgres datastore code has been cleaned up.
This change is not needed anymore.

PR #2864, opened 2022-08-29
Reinstate UI unit test coverage in 'yarn test' script

dvail:
## Description

This reverts the change that disabled ui unit test coverage from the `yarn test` script in the PR targeting performance improvements a couple of weeks ago.

An unfortunate side effect of removing the coverage (stored in a junit test report) is that it becomes more difficult to track down where errors are coming from when the ui-unit-tests prow job fails.

[This is a run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2864/pull-ci-stackrox-stackrox-master-ui-unit-tests/1564240856653762560) with two forced test failures, without the coverage report.

[This is a run ](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2864/pull-ci-stackrox-stackrox-master-ui-unit-tests/1564247272131137536)with the same failures, but coverage re-enabled.

For the second run, the prow UI displays which tests failed as well as the closest stack trace at the top of the page. In the first run we would have to search for the same errors in the full log text.

## Checklist

## Testing Performed

- Run ui-unit-tests with forced failures against `master` and check prow output
- Run ui-unit-tests with forced failures with coverage re-enabled and check prow output


bradr5 (approved):

pedrottimark (approved):
Thank you for improving developer experience when unit tests fail.

PR #2775, opened 2022-08-22
Remove TryAllowed from scope checker

janisz:
## Description

There is no need to try with only built in authorizer and no error scope checker.

## Testing Performed

CI

PR #2873, opened 2022-08-29
Wait for 3 more generic requests in visit helper functions for integration tests

pedrottimark:
### Test failure

> for API Docs: General sanity checks should have correct page titles based on URL for API Docs

> Timed out retrying after 10000ms: `cy.wait()` timed out waiting `10000ms` for the 1st response to the route: `apiDocs`. No response ever occurred.

cypress/integration/general.test.js
prow

* 1563129461082165248 from master build for #2775 on 2022-08-26
* 1564401364421840896 from master build for #2825 on 2022-08-29
* 1564522511587086336 from master build for #2883 on 2022-08-30

### Analysis

* /v1/api/docs/swagger request has green indicator at lower left of the picture.
* This test might some of the time interval on prerequisite requests. Fix that problem now. Only if it fails in the future, increase the timeout.

### Changed files

1. Edit helpers/visit.js
    * Factor out `requestConfigGeneric` object and add comments about code files which make requests.
    * Add requests to `visit` and `visitWithPermissions` functions: /v1/auth/status and /v1/credentialexpiry for CENTRAL and SCANNER.

1. Edit integration/general.test.js
    * Replace `cy.visit` with `visit` helper function which waits for generic prerequisite requests before time interval begins for test-specific responses.
    * Replace `timeout` with `responseTimeout` key to be more precise.

### Residue

1. Rewrite the classic test for API Docs and then unskip it. Because it is slow, combine the tests.
2. Investigate double pair of credentialexpiry requests.

## Checklist

## Testing Performed

For local deployment: ran tests individually before and after the change

### DIFF
--- a/ui/apps/platform/cypress/helpers/visit.js
+++ b/ui/apps/platform/cypress/helpers/visit.js
@@ -2,6 +2,36 @@ import * as api from '../constants/apiEndpoints';
 
 import { interceptRequests, waitForResponses } from './request';
 
+// generic requests to render the MainPage component
+const requestConfigGeneric = {
+    routeMatcherMap: {
+        featureflags: {
+            method: 'GET',
+            url: api.featureFlags,
+        }, // sagas/featureFlagSagas
+        mypermissions: {
+            method: 'GET',
+            url: api.roles.mypermissions,
+        }, // sagas/authSagas
+        'config/public': {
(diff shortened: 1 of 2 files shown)


vjwilson (approved):
/lgtm

PR #2877, opened 2022-08-29
Generic proto walker

misberner:
## Description

Yeah, not sure if we can replace the postgres walker with it, doesn't actually look like it. But two other instances, and I'll shortly be adding a third.

## Checklist

## Testing Performed

- Unit tests (existing)
- CI


pkg/protowalk/field_path.go
```diff
@@ -0,0 +1,141 @@
...
+	ContainingType reflect.Type
+	// StructField is the definition of the struct field within the containing type.
+	reflect.StructField
+}
+
+// ElemType is the type of element stored in this field. In case of slices (repeated fields), this is the slice
+// element type; in case of maps, this is the value type. Otherwise, it is the type of the respective struct field.
+func (f Field) ElemType() reflect.Type {
+	ty := f.Type
+	if ty.Kind() == reflect.Slice || ty.Kind() == reflect.Map {
```
connorgorman, line 38: Nit: feel free to not do this, but I prefer
```
switch ty.Kind(){
  case reflect.Slice, reflect.Map:
    
}
```

connorgorman: It's just two so I'm a bit torn and it does add an extra line

pkg/protowalk/field_path.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Kind returns the kind of this field.
+func (f Field) Kind() FieldKind {
+	elemTy := f.ElemType()
+	if elemTy.Kind() == reflect.Interface {
+		return OneofFieldKind
+	}
+	if elemTy.Kind() == reflect.Ptr && elemTy.Elem().Kind() == reflect.Struct {
+		return MessageFieldKind
```
connorgorman, line 51: hmm so one of the perf benefits of gogo proto (which we don't use) was the ability to use fewer pointers. Does vtproto have anything like that?

connorgorman (approved):
Looks good. Can you test the following by pulling the IndexMappings and ensuring that the serialized json is the same? This will be a nice validation that nothing has changed

connorgorman:
Ship?

janisz:
@connorgorman Are we going to merge it?

PR #2887, opened 2022-08-30
Increase golangci-lint timeout

janisz:
## Description

From time to time `style-checks` fails due to golangci-lint timeout. Let's increase it to prevent such failures as this check is required and fully predictable and repeatable hence should pass every time. 

## Testing Performed

CI


gavin-stackrox (approved):

janisz:
🔁  /retest

PR #2888, opened 2022-08-30
ROX-11891: Add an empty commit when retagging

0x656b694d:
## Description

Extraction of the `tag-rc.sh` script, which can also be used locally to retag a release branch if needed.

Allow running wet scripts locally by setting `DRY_RUN=false`.

## Checklist

## Testing Performed

TODO(replace-me)

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -220,13 +220,10 @@ jobs:
           fi
       - name: Tag release branch with "${{needs.variables.outputs.milestone}}"
         run: |
-          git tag --annotate "${{needs.variables.outputs.milestone}}" --message ""
-      - name: Push
-        if: env.DRY_RUN == 'false'
-        run: |
-          git push --follow-tags
-      - run: |
-          echo "Release branch has been tagged with ${{needs.variables.outputs.milestone}}." >> $GITHUB_STEP_SUMMARY
+          set -uo pipefail
+          gh api -H "$ACCEPT_RAW" "${{env.script_url}}" | bash -s -- \
+            tag-rc \
+            "${{needs.variables.outputs.milestone}}"
       - name: Create next milestone
         if: env.DRY_RUN == 'false'
         run: |
--- a/.github/workflows/scripts/tag-rc.sh
+++ b/.github/workflows/scripts/tag-rc.sh
@@ -18,7 +18,7 @@ check_not_empty \
 
 if [ "$(git tag "$TAG" --points-at HEAD)" = "$TAG" ]; then
     git push --delete origin "$TAG"
-    gh_log warning "Tag '$TAG' has been deleted."
+    gh_log warning "Existing tag '$TAG' has been deleted."
 fi
 
 git commit --allow-empty --message "Empty commit to trigger CI"
--- a/.github/workflows/scripts/tag-rc.sh
+++ b/.github/workflows/scripts/tag-rc.sh
@@ -3,6 +3,10 @@
 # Add an empty commit and delete the remote tag if the tag exists.
 # Tag and push the branch.
 #
+# Local wet run:
+#
+#   DRY_RUN=false bash local-env.sh tag-rc <tag>
+#
 set -euo pipefail
 
 TAG="$1"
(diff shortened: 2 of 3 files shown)


.github/workflows/scripts/tag-rc.sh
```diff
@@ -0,0 +1,31 @@
...
+    git push --delete origin "$TAG"
+    git commit --allow-empty --message "Empty commit to trigger CI"
+    gh_log warning "Tag '$TAG' has been deleted and an empty commit has been added to trigger CI."
+fi
+
+git tag --force --annotate "$TAG" --message "Upstream release automation"
+
+if [ "$DRY_RUN" = "false" ]; then
+    git push --follow-tags
+fi
```
vikin91, line 29: There are 2 cases that are uncovered yet, but we want to cover here:

1. `$TAG` is unknown yet, but there were no commits created in cherry-picking and no updates to the submodules.
2. Last commit on the release branch is untagged (e.g., a fix was merged to the release branch but not to `master` - this has happened once in 3.71)

To cover these both cases, we could do:

```suggestion
if git ls-remote --tags --exit-code origin "$TAG"; then
    git push --delete origin "$TAG"
    gh_log warning "Tag '$TAG' has been deleted"
fi

git commit --allow-empty --message "Empty commit to trigger CI"
gh_log warning "An empty commit has been added to trigger CI."
git tag --force --annotate "$TAG" --message "Upstream release automation"

if [ "$DRY_RUN" = "false" ]; then
    git push --follow-tags
fi
```

WDYT?

.github/workflows/cut-rc.yml
```diff
@@ -220,13 +220,10 @@ jobs:
...
       - name: Tag release branch with "${{needs.variables.outputs.milestone}}"
         run: |
-          git tag --annotate "${{needs.variables.outputs.milestone}}" --message ""
-      - name: Push
-        if: env.DRY_RUN == 'false'
-        run: |
-          git push --follow-tags
-      - run: |
-          echo "Release branch has been tagged with ${{needs.variables.outputs.milestone}}." >> $GITHUB_STEP_SUMMARY
+          set -uo pipefail
```
vikin91, line 223: No `set -e` here?

vikin91 (commented):
Looks good, but please check the 2 scenarios that I mentioned in the comment.

.github/workflows/scripts/tag-rc.sh
```diff
@@ -0,0 +1,31 @@
...
+    DRY_RUN
+
+if git ls-remote --tags --exit-code origin "$TAG"; then
+    git push --delete origin "$TAG"
+    git commit --allow-empty --message "Empty commit to trigger CI"
+    gh_log warning "Tag '$TAG' has been deleted and an empty commit has been added to trigger CI."
+fi
+
+git tag --force --annotate "$TAG" --message "Upstream release automation"
+
```
0x656b694d, line 26:
```suggestion
if git push --delete origin "$TAG"; then
    gh_log warning "Tag '$TAG' has been deleted."
fi

git commit --allow-empty --message "Empty commit to trigger CI"
git tag --force --annotate "$TAG" --message "Upstream release automation"

```

.github/workflows/scripts/tag-rc.sh
0x656b694d: Right, but let's make it simpler, see above.

vikin91: It feels strange to review a comment suggestion that is an answer to a comment suggestion, because I cannot comment inline :) 

The important thing is that we want to remove the tag ONLY if it is on the tip of the branch. If you run `git ls-remote --tags --exit-code origin "3.69.1-rc.4"` then it will return 0 and then you would simply delete something that is not the tip of the branch.

I would suggest to do the following:
- fetch branch from origin
- check if the last commit is tagged 
    - if not, then add empty commit and tag it
    - if yes, then is that tag the same as `$TAG`?
        - if yes, then remove the tag, add empty commit and tag it
        - if no, then only add empty commit and tag it

0x656b694d: Please take a look at the updated version. I simply check the HEAD for the tag before deleting it.

vikin91 (approved):
This looks really good now!

PR #2889, opened 2022-08-30
Dashrews/ROX-12289 cleanup rocks after postgres upgrade

dashrews78:
## Description

From the perspective of the migrator, this will clear out Rocks after a Postgres has been used twice.  At that point we will no longer be able to rollback to Rocks and the data within rocks will be removed.  For the time being, central will add the current directory back to the PVC for things like bolt.  Clearing this out from a global perspective is a bit more work, so the focus here is from the perspective of the migrator.  The rest of it will be handled by such tickets as ROX-9882.

## Checklist

## Testing Performed

Did various levels of upgrades and such in manual testing.  Also added a multiple upgrade steps in the clone_test to drive this condition.

### DIFF
--- a/migrator/clone/db_clone_manager_impl.go
+++ b/migrator/clone/db_clone_manager_impl.go
@@ -7,7 +7,7 @@ import (
 	"github.com/stackrox/rox/pkg/features"
 )
 
-// DBCloneManagerImpl - scans and manage database clones within central.
+// dbCloneManagerImpl - scans and manage database clones within central.
 type dbCloneManagerImpl struct {
 	forceRollbackVersion string
 	adminConfig          *pgxpool.Config
@@ -82,7 +82,8 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate() (string, string, string, error)
 		}
 
 		// If we need to migrate from rocks we need to continue processing and
-		// get the Rocks clones.  If we don't, there is no need to process Rocks
+		// get the Rocks clones.  If we don't, there is no need to process Rocks, but
+		// we will check to see if we can get rid of rocks
 		if !migrateFromRocks {
 			return "", "", pgClone, nil
 		}
@@ -111,7 +112,24 @@ func (d *dbCloneManagerImpl) Persist(cloneName string, pgClone string, persistBo
 				log.Warnf("Unable to create a previous version of Rocks to rollback to: %v", err)
 			}
 		}
-		return d.dbmPostgres.Persist(pgClone)
+
+		err := d.dbmPostgres.Persist(pgClone)
+		if err != nil {
+			return err
+		}
+
+		// Now that updated Postgres was persisted we can decommission RocksDB if necessary
+		if !persistBoth {
+			rocksVersion := d.dbmRocks.GetVersion(rocksdb.CurrentClone)
+			currentPostgresVersion := d.dbmPostgres.GetCurrentVersion()
+
+			// If the versions do not match, we have updated another time with Postgres,
+			// so we can no longer roll back to RocksDB.
+			if rocksVersion == nil || rocksVersion.MainVersion != currentPostgresVersion.MainVersion {
+				d.dbmRocks.DecommissionRocksDB()
+			}
+		}
+		return nil
 	}
 	return d.dbmRocks.Persist(cloneName)
 }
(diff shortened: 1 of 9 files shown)


migrator/clone/db_clone_manager_impl.go
```diff
@@ -82,8 +85,28 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate() (string, string, string, error)
...
+		// get the Rocks clones.  If we don't, there is no need to process Rocks, but
+		// we will check to see if we can get rid of rocks
 		if !migrateFromRocks {
+			currentPostgresVersion := d.dbmPostgres.GetCurrentVersion()
+			// We no longer need to migrate from Rocks so we should check to see if we have upgraded
+			// to the point where we should remove it entirely
+			if rocksVersion == nil || rocksVersion.SeqNum == 0 {
+				log.Info("rocks is empty or 0, going to decommission")
+				// Obviously if the Rocks version is nil or empty we need not carry the Rocks DB forward.
+				d.decommissionRocks = true
```
c-du, line 97: Does it also mean that there is no database to decomission?

pkg/migrations/paths.go
```diff
@@ -78,11 +78,20 @@ func GetRestoreClone() string {
...
+			log.Infof("in Postgres mode with no current RocksDB database: %v", err)
+		} else {
+			utils.CrashOnError(errors.Wrap(err, "no current database"))
+		}
+	}
 
 	switch path {
 	case CurrentPath(), currentDB:
-		utils.CrashOnError(errors.Errorf("Database in use. Cannot remove %s", path))
+		if !features.PostgresDatastore.Enabled() {
```
c-du, line 91: This function is to protect the current database from being accidentally removed from auto or manual linkage errors. Let's not change this. You can just use the dedicated function DecommissionRocksDB to remove them.

migrator/clone/db_clone_manager_impl.go
```diff
@@ -82,8 +85,28 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate() (string, string, string, error)
...
+			} else if currentPostgresVersion != nil &&
+				(version.CompareVersions(currentPostgresVersion.MainVersion, rocksVersion.MainVersion) > 0 &&
+					version.CompareVersions(version.GetMainVersion(), currentPostgresVersion.MainVersion) > 0) ||
```
c-du, line 100: So you want to make sure you are in upgrade here. If we have Postgres enabled, we are in upgrade and we are not upgrade from rocks then we decomission.

There are issues in line 101. We cannot strictly compare the seqnum to make sure we are in an upgrade. In cases there is no migration added in this version, we may have an upgrade with the same sequence number.  And also we may not reliable compare that we are upgrading from a lower version to a higher version. There are cases we cannot determine which version is "larger" than the other for example in engineering versions, check ROX_DONT_COMPARE_DEV_BUILDS.
Those cases will end up not matching this condition.

c-du (commented):
IMHO, the best place to decommission is at the Persist time. If you had selected a pgClone(no rocks clone) to migrate from and after the upgrade was completed successfully. You just need to differentiate it from normal boot-up like doing an exact string version comparision.

migrator/clone/db_clone_manager_impl.go
dashrews78: @c-du It means we found one but it has no version info, so it is either a fresh install and it is empty or we have already wiped it out from the migrator perspective, but central has added back in as it will until we finish clearing things back out.

migrator/clone/db_clone_manager_impl.go
dashrews78: @c-du The general idea is that I tried to follow the same conditions that the clone managers use to determine if there is a rollback or an upgrade.  Those conditions check that the versions are different OR that the sequence numbers are different.  The only thing I'm doing here is doing the extra check to make sure we are planning to go multiple steps since our last RocksDB.  Like the check below, only I'm looking in the opposite direction:

`if version.CompareVersions(currClone.GetVersion(), version.GetMainVersion()) > 0 || currClone.GetSeqNum() > migrations.CurrentDBVersionSeqNum() {
			// Force rollback
			return PreviousClone, d.getPath(d.cloneMap[PreviousClone].GetDirName()), nil
		}`
		
Also, does it really matter for engineering versions?  In those cases we would just not decommission until such time that those versions were obviously compare higher.  What should we do in those cases?  Is that why the current logic like I posted above uses sequence number OR version?

dashrews78:
> 

@c-du I think I'm following.  So if it is only a postgres persist, after the persist compare the versions.  If they differ, decommission.  The only trick with that is the version for the database we are using gets set in central NOT the migrator.  So the first postgres only persist would have the same version as rocks.  The 2nd postgres only persist would have a new version number.  I suppose that we could set the version in migrator to decommission 1 release faster.

c-du:
> r. I suppose that we could set the version in migrator to decommission 1 release faster.
Yes, that is good too.

c-du (approved):
Thanks!

PR #2891, opened 2022-08-30
Replace assertions about DOM elements in risk sort priority integration test

pedrottimark:
### Test failure

> should sort priority in the table: Risk page without mock API should sort priority in the table

> Timed out retrying after 4000ms: Expected to find element: `.rt-tr-group:first-child .rt-tr .rt-td:nth-child(5):contains("6")`, but never found it.

cypress/integration/risk/risk.test.js

* 1564197760603787264 from master build for #2854 on 2022-08-29

Here is the screenshot for the test failure in which **collector** has **Priority** of 7:
prow

Here is a frame from the video for the preceding test in which **collector** has **Priority** of 6:
frame

### Analysis

The test assumes that the priorities of deployments remains the same:
* first request which is sorted ascending (get values of first and last row)
* second request which is sorted descending (assert expected values of first and last row)
* third request which is sorted ascending (assert expected values of first and last row, but value of first row has changed)

### Solution

Replace assertions about DOM elements with assertions about requests: include expected search parameters for ascending or descending sort.

### Residue

1. While investigating why **Priority** table cell heading does not display the initial state of ascending sort:
    * `TableV2` component does not have `defaultSorted` prop like `Table` component.
    * This is the only use of `TableV2` component.
    * There is no use of `CheckboxTableV2` component.

## Checklist
- [x] Edited integration test

## Testing Performed

For local deployment: ran test individually before and after the change

### DIFF
--- a/ui/apps/platform/cypress/integration/risk/risk.test.js
+++ b/ui/apps/platform/cypress/integration/risk/risk.test.js
@@ -22,59 +22,45 @@ describe('Risk page', () => {
 
             const priorityColumnHeadingSelector = `${RiskPageSelectors.table.columnHeaders}:contains("Priority")`;
 
-            /*
-             * Because risk table displays all deployments on only one page in integration testing environment,
-             * sort descending just reverses the rows so first becomes last, and last becomes first.
-             */
-            cy.get(`${RiskPageSelectors.table.dataRows} .rt-td:nth-child(5)`).then(
-                ($priorityCells) => {
-                    const priorityFirst = $priorityCells.eq(0).text();
-                    const priorityLast = $priorityCells.eq($priorityCells.length - 1).text();
+            // Initial table state is not sorted by priority.
+            cy.get(priorityColumnHeadingSelector)
+                .should('not.have.class', '-sort-asc')
+                .should('not.have.class', '-sort-desc');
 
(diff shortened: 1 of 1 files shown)


dvail (approved):
Interesting that the risk priority changes after sorting. Is this consistent or just bad luck that something in the system has caused a risk modification?

Either way, this is a good example of how this test assertion can be flawed.

I can still see value in testing that the items in the table are in sorted order though, e.g. that each item is >= or <= the following one in the table depending on the sort order. It might be something worth adding in the future if we revisit this test.

pedrottimark:
@dvail Bad luck. Reminds me of an ironic saying in auto racing: If he didn’t have bad luck, he’d have no luck at all.

> I can still see value in testing that the items in the table are in sorted order though, e.g. that each item is >= or <= the following one in the table depending on the sort order. It might be something worth adding in the future if we revisit this test.

Yes, what you suggest is worthwhile. Just yesterday, I found in Vulnerability Management a (commented out) helper function to assert ascending or descending order of the values in a column. I will update this pull request.

bradr5 (approved):

pedrottimark:
One second thought, I will merge this now, and then open a new pull request for Risk and Vulnerability Management.

PR #2892, opened 2022-08-30
Add Postgres query tracer that logs the GraphQL query and the resulting queries

connorgorman:
## Description

Adds a top level context value that will trace the run queries in order
Add threshold for graphql query to be logged and also for individual postgres queries to be logged

Future work:
- extend into grpc interceptor
- extend to include rows affected / rows returned watchdog

## Checklist

## Testing Performed

Run with it on and check which queries are slow

### DIFF
--- a/central/graphql/handler/handler.go
+++ b/central/graphql/handler/handler.go
@@ -22,7 +22,8 @@ import (
 var (
 	log = logging.LoggerForModule()
 
-	queryTracerEnabled = features.PostgresDatastore.Enabled() && env.PostgresQueryTracer.BooleanSetting()
+	queryTracerEnabled    = features.PostgresDatastore.Enabled() && env.PostgresQueryTracer.BooleanSetting()
+	graphQLQueryThreshold = env.PostgresQueryTracerGraphQLThreshold.DurationSetting()
 )
 
 type logger struct {
@@ -83,7 +84,7 @@ func (h *relayHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if queryTracerEnabled {
+	if queryTracerEnabled && time.Since(startTime) > graphQLQueryThreshold {
 		postgres.LogTrace(log, ctx, fmt.Sprintf("GraphQL Op %s took %d ms: %s %+v", params.OperationName, time.Since(startTime).Milliseconds(), params.Query, params.Variables))
 	}
 	w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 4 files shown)


pkg/postgres/postgres.go
```diff
@@ -0,0 +1,83 @@
...
+	qt.events = append(qt.events, queryEvent{
+		query:    query,
+		args:     args,
+		duration: time.Since(start),
+		stack:    debug.Stack(),
+	})
+}
+
+func LogTrace(logger *logging.Logger, ctx context.Context, contextString string) {
+	tracer := ctx.Value(queryTracerKey{}).(*queryTracer)
```
misberner, line 46:
```suggestion
	tracer := GetTracerFromContext(ctx)
	if tracer == nil {
	  return
	}
```

pkg/postgres/postgres.go
```diff
@@ -0,0 +1,83 @@
...
+
+	qt.events = append(qt.events, queryEvent{
+		query:    query,
+		args:     args,
+		duration: time.Since(start),
+		stack:    debug.Stack(),
+	})
+}
+
+func LogTrace(logger *logging.Logger, ctx context.Context, contextString string) {
```
misberner, line 45: Would suggest to add a
```
LogTracef(logger *logging.Logger, ctx context.Context, format string, args... interface{}) {
  LogTrace(logger, ctx, fmt.Sprintf(format, args))
}
```

misberner, line 45: Also, `ctx context.Context` should always be the first parameter

misberner (approved):

md2119 (approved):
Nice!

central/graphql/handler/handler.go
```diff
@@ -63,8 +70,12 @@ func (h *relayHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Adds the data loader intermediates so that we can stop ourselves from loading the same data from the store
 	// many time.
 	ctx = loaders.WithLoaderContext(ctx)
+	if queryTracerEnabled {
+		ctx = postgres.WithTracerContext(ctx)
+	}
 
-	defer metrics.SetGraphQLQueryDurationTime(time.Now(), params.Query)
+	startTime := time.Now()
+	defer metrics.SetGraphQLQueryDurationTime(startTime, params.Query)
```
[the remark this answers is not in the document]
connorgorman: oh yeah, this is the current metrics so I'm not actually changing anything here

PR #2894, opened 2022-08-30
ROX-12368: Post query pagination for clusters/namespaces for correctness and performance

connorgorman:
## Description

Due to the priority sorted searcher, we need to post paginate. 

Filed ROX-12401 to roll this back eventually, but we'll need to determine how to handle the risk priorities

## Checklist

## Testing Performed

Run on a cluster with more than 25 namespaces and only 25 results appear


md2119 (approved):
Make the change for cluster also

PR #2896, opened 2022-08-30
Change AddStrings queries to AddExactMatches where possible

md2119:
## Description

This PR is In prep for changing the string search from prefix to substring. We should replace the occurrences in the code that should be exact matches instead of prefix queries such as, enum queries, ID queries.

## Checklist

## Testing Performed
CI

### DIFF
--- a/central/reports/common/query_builder.go
+++ b/central/reports/common/query_builder.go
@@ -63,9 +63,9 @@ func (q *queryBuilder) buildCVEAttributesQuery() (string, error) {
 	case storage.VulnerabilityReportFilters_BOTH:
 		break
 	case storage.VulnerabilityReportFilters_FIXABLE:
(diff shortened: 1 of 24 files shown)


central/reports/common/query_builder.go
```diff
@@ -63,17 +63,17 @@ func (q *queryBuilder) buildCVEAttributesQuery() (string, error) {
 	case storage.VulnerabilityReportFilters_BOTH:
 		break
 	case storage.VulnerabilityReportFilters_FIXABLE:
-		conjuncts = append(conjuncts, search.NewQueryBuilder().AddStrings(search.Fixable, "true").Query())
+		conjuncts = append(conjuncts, search.NewQueryBuilder().AddExactMatches(search.Fixable, "true").Query())
```
connorgorman, line 66: Can you change this to bool query?

connorgorman (approved):
see one suggestion for bool query

PR #2906, opened 2022-08-31
ROX-12015: Make osci config depend on release branch

0x656b694d:
## Description

As openshift/release configuration requires the existence of the release branch, let's make it so.

Changed order of the jobs and made `ci` need `branch`.


vikin91 (approved):
LGTM! (Didn't go into details assuming this is just a move of the code block)

PR #2907, opened 2022-08-31
ROX-11537: workaround a race in admission controller test

porridge:
## Description

See [the ticket for a description](https://issues.redhat.com/browse/ROX-11537?focusedCommentId=20877231&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20877231) of the root cause.

This is a stopgap solution to stop the pain, but we need to think of a better way to test this.
It also reminds me of a similar race (in sensor?) that @fredrb mentioned a while ago, where there was no way to tell whether all events had been processed.

## Checklist

## Testing Performed

Ran `go test . -test.v -test.run '^\QTestExecIntoPodNameEventPolicy\E$' -test.count=100` in `sensor/admission-control/service`.

### DIFF
--- a/sensor/admission-control/service/service_test.go
+++ b/sensor/admission-control/service/service_test.go
@@ -168,6 +168,9 @@ func (r serviceTestRun) execute() {
 	require.NotNil(r.t, r.mgr)
 	require.NotNil(r.t, r.handlerFunc)
 	require.True(r.t, r.mgr.IsReady(), "Manager is stopped or was not started")
+	// Wait for any events delivered to manager prior to this call to be processed.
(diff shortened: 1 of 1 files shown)


misberner (approved):
LGTM to de-flake but I think I can send out a proper fix in a few minutes

porridge:
Collector crash seems to be known and unrelated, merging.

porridge:
FTR this can mostly be reverted (modulo the docstring and the final timeout reduction) once https://github.com/stackrox/stackrox/pull/2909 is merged.

PR #2908, opened 2022-08-31
ColumnType option

erthalion:
## Description

Allow specifying an SQL type directly without being deduced via golang
type. In the proto definition it looks like this:

    string id = 1 [(gogoproto.moretags) = 'sql:"pk,type(uuid)"'];

## Checklist

## Testing Performed

Local testing on example of network_flows.


md2119 (approved):
Nice!

connorgorman:
@erthalion I think we need a rebase

pkg/postgres/walker/walker.go
```diff
@@ -212,6 +212,10 @@ func getPostgresOptions(tag string, topLevel bool, ignorePK, ignoreUnique, ignor
 			opts.Reference.Directional = true
 		case field == "ignore-fks":
 			opts.IgnoreChildFKs = true
+		case strings.HasPrefix(field, "type"):
+			typeName := field[strings.Index(field, "(")+1 : strings.Index(field, ")")]
```
connorgorman, line 216: Maybe in a follow up, but would be cool to add this to the stringutils package as a GetBetween function

connorgorman (approved):

PR #2909, opened 2022-08-31
ROX-11537: Fix flake in admission control test

misberner:
## Description

Use the `ResourcesSynced` request to realize a "Sync" helper on the manager, allowing us to wait until all messages sent earlier are processed.

It may be a bit of a hack, but IMHO it's a benign one. Realizing a callback-like mechanism isn't easily possible because `ResourceUpdatesC` is designed for proto types, and thus cannot be extended by a user to pass a callback channel/signal.

## Checklist

## Testing Performed

Ran test with
```
$ GOMAXPROCS=1 go test ./sensor/admission-control/service/ -test.v -test.run '^\QTestExecIntoPodNameEventPolicy\E$' -test.count=10
```

### DIFF
--- a/sensor/admission-control/manager/manager_impl.go
+++ b/sensor/admission-control/manager/manager_impl.go
@@ -218,8 +218,8 @@ func (m *manager) run() {
 				m.ProcessNewSettings(newSettings)
 			case req := <-m.resourceUpdatesC:
 				m.processUpdateResourceRequest(req)
-			case sig := <-m.syncC:
-				sig.Signal()
+			case syncSig := <-m.syncC:
+				syncSig.Signal()
 			}
 		}
 	}


porridge (approved):
My only worry is that I guess if we actually had a test for `_Synced` that used `serviceTestRun` then this would interfere, right?
Perhaps add a comment inside `Sync` that explains that this is a bit of a hack, and not an intended use for this facility. Someone might be a bit confused seeing how differently it's used in a test vs in production code.

porridge:
Also @fredrb this might be a pattern to follow in scanner, w.r.t. the race you mentioned some time ago...

SimonBaeumer (approved):
Thanks for fixing @porridge @misberner!

fredrb:
> Also @fredrb this might be a pattern to follow in scanner, w.r.t. the race you mentioned some time ago...

@porridge I assume you mean sensor?

porridge:
> > Also @fredrb this might be a pattern to follow in scanner, w.r.t. the race you mentioned some time ago...
> 
> @porridge I assume you mean sensor?

:100: :facepalm:

misberner:
@porridge PTAL, I've changed the approach to add an explicit "Sync" method for the manager. While primarily used for testing, I feel such a method is generally useful for any component that does asynchronous processing and thus doesn't pollute the public interface.

This makes it fully non-hacky and also avoids the problem you sketched.

sensor/admission-control/manager/manager_impl.go
```diff
@@ -181,20 +204,23 @@ func (m *manager) runSettingsWatch() {
...
+			// without reflection.
+			select {
+			case <-m.stopSig.Done():
+				m.ProcessNewSettings(nil)
+				return
+			case newSettings := <-m.settingsC:
+				m.ProcessNewSettings(newSettings)
+			case req := <-m.resourceUpdatesC:
+				m.processUpdateResourceRequest(req)
+			case sig := <-m.syncC:
```
porridge, line 221: With `stopSig` and `stoppedSig` already being mentioned in this function, perhaps call this `syncSig` to be a bit more specific?

porridge (approved):
Nice!

porridge:
```
[2022-09-01 12:54:08] panic: Import processing for file "/go/src/github.com/stackrox/stackrox/pkg/postgres/schema/cluster_health_statuses.go" took more than 1 second (1.069848009s). This typically indicates that an import was not added to the Go template, which forced import processing to search through all types and magically add the import. Please add the import to the template; you can compare the imports in the generated file with the ones in the template, and add the missing one(s)
```
/retest-required

sensor/admission-control/manager/manager_impl.go
misberner: done

PR #1769, opened 2022-05-18
Write frontend credential expiry proto and rewrite services in TypeScript

pedrottimark:
### Goals

* Increase type safety with compile-time checks.
* Simplify maintenance of consistent backend and frontend response types.

### Changed files

1. Edit cypress/integration/certexpiration.test.js
    * Replace `cy.visit` with `visitMainDashboard` helper function

2. Rename and edit src/Containers/CredentialExpiryBanners/CredentialExpiry.js
    * Add `nameOfComponent` to compute name because `component` prop value becomes non-presentational
    * Replace `expiryFetchFunc` prop with `fetchCertExpiryForComponent` function
    * Replace `userRolePermissions` with `hasServiceIdentityWritePermission` prop
    * Replace `downloadYAMLFunc` prop with `generateCertSecretForComponent` function

3. Delete src/Containers/CredentialExpiryBanners/CredentialExpiryBanners.js

4. Edit src/Containers/MainPage/MainPage.tsx
    * Replace `CredentialExpiryBanners` element with `CredentialExpiryBanner` elements

5. Rename and edit src/services/CertGenerationService.js
    * Import `CertExpiryComponent` type
    * Replace specific `generateCentralCertSecret` and `generateScannerCertSecret` with generic `generateCertSecretForComponent` function

6. Rename and edit src/services/CredentialExpiryService.js
    * Import `CertExpiryComponent` type
    * Replace specific `fetchCentralCertExpiry` and `fetchScannerCertExpiry` with generic `fetchCertExpiryForComponent` function

7. Add src/types/credentialExpiryService.proto.ts which corresponds to https://github.com/stackrox/stackrox/blob/master/proto/api/v1/credential_expiry_service.proto

## Checklist
- [x] Investigated and inspected CI test results
- [x] ~Unit test and regression tests added~

### manual testing

1. `yarn build` in ui
2. `yarn start` in ui

    * See same requests and returned values in banner components (with temporary console.log)

    * See same request to download secret file for CENTRAL and file in bottom bar (with temporary change to expiry)
        certgen-CENTRAL-2

    * See same request to download secret file for SCANNER and file in bottom bar (with temporary change to expiry)
        certgen-SCANNER-2

### integration testing

1. `yarn cypress-open` in ui/apps/platform
    * certexpiration.test.js

PR #2910, opened 2022-08-31
Prevent unneeded credentialexpiry requests

pedrottimark:
### Test failure

> Network Graph Search should filter to show only the deployments from the stackrox namespace and deployments connected to them

> Timed out retrying after 4000ms: Expected to find element: `h1:contains("Network Graph")`, but never found it.

at visitNetworkGraph
at visitNetworkGraphWithNamespaceFilter

cypress/integration/networkGraph/networkGraph.test.js
prow

Here is the same test in local deployment after change to prevent unneeded credentialexpiry requests:
local

* 1564230427168739328 from master build for #2946 on 2022-08-29

For comparison, here is a picture related to this item under **Residue** in #2873

> Investigate double pair of credentialexpiry requests.
prow-API-Docs

### Analysis

The test failed because of timeout before the container even made its requests. The /v1/clusters request is from network sagas, not the container.

Whether the unnecessary requests are an insult or the injury, prevent them.

My bad when rewriting `CredentialExpiry` component in #1769

The component has started making the request after every completed render, because I replaced `[expiryFetchFunc, component]` dependencies with no dependencies instead of `[component]` which does not change.

https://reactjs.org/docs/hooks-reference.html

### Residue

1. Improve auth.test.js and certexpiration.test.js tests which have mock response for generic requests.
2. Wait for a few more container-specific requests in `visitNetworkGraph` funciton.
3. Split network.test.js into separate test files with descriptive names and move into networkGraph folder. 

## Checklist

## Testing Performed

* certexpiration.test.js
* general.test.js
* network.test.js
* networkGraph/networkGraph.test.js

### DIFF
--- a/ui/apps/platform/src/Containers/MainPage/CredentialExpiryBanner.tsx
+++ b/ui/apps/platform/src/Containers/MainPage/CredentialExpiryBanner.tsx
@@ -44,7 +44,7 @@ function CredentialExpiryBanner({
                 // eslint-disable-next-line no-console
                 console.warn(`Failed to fetch certification expiration for ${component}`, e);
             });
-    });
+    }, [component]);
 
     if (!expirationDate) {
         return null;


vjwilson (approved):
/lgtm

PR #2916, opened 2022-08-31
ROX-13805: Add check-style to GHA

janisz:
## Description

Add `make style` to github actions workflow

## Testing Performed

CI

### DIFF
--- a/.github/actions/cache-gradle-dependencies/action.yaml
+++ b/.github/actions/cache-gradle-dependencies/action.yaml
@@ -10,7 +10,7 @@ runs:
           ~/.gradle/caches
           ~/.gradle/wrapper
           ~/.m2/repository
-        key: npm-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
+        key: gradle-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
         restore-keys: |
           gradle-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
           gradle-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-
--- a/.github/actions/cache-gradle-dependencies/action.yaml
+++ b/.github/actions/cache-gradle-dependencies/action.yaml
@@ -1,0 +1,17 @@
+name: Cache QA Tests Dependencies
+description: Cache QA Tests Dependencies
+runs:
+  using: composite
+  steps:
+    - name: Cache Gradle Dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+          ~/.m2/repository
+        key: npm-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
+        restore-keys: |
+          gradle-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
+          gradle-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-
+          gradle-v2-
--- a/.github/actions/cache-ui-dependencies/action.yaml
+++ b/.github/actions/cache-ui-dependencies/action.yaml
@@ -13,12 +13,15 @@ runs:
       with:
         path: |
           ui/deps
+          /github/home/.cache/yarn/v6
           /github/home/.cache/Cypress
           /usr/local/share/.cache
           ui/node_modules
           ui/apps/platform/node_modules
           ui/packages/ui-components/node_modules
           ui/packages/tailwind-config/node_modules
-        key: npm-v2-${{ hashFiles(inputs.lockFile) }}
+        key: npm-v2-${{ hashFiles(inputs.lockFile) }}-${{ github.job }}
         restore-keys: |
+          npm-v2-${{ hashFiles(inputs.lockFile) }}-${{ github.job }}
+          npm-v2-${{ hashFiles(inputs.lockFile) }}-
           npm-v2-
--- a/.github/actions/cache-ui-dependencies/action.yaml
+++ b/.github/actions/cache-ui-dependencies/action.yaml
@@ -13,7 +13,7 @@ runs:
       with:
         path: |
           ui/deps
-          /github/home/.cache/yarn/v6
+          /github/home/.cache/yarn
           /github/home/.cache/Cypress
           /usr/local/share/.cache
           ui/node_modules
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -1,0 +1,53 @@
+name: Style
+on:
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    - labeled
+
+env:
+  ROX_PRODUCT_BRANDING: RHACS_BRANDING
+
+jobs:
+
+  style-check:
+    if: |
+      (github.event.action == 'labeled' && github.event.label.name == 'turbo-build') ||
+      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'turbo-build'))
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.49
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Cache Go dependencies
+        uses: ./.github/actions/cache-go-dependencies
+
+      - uses: ./.github/actions/create-concatenated-ui-monorepo-lock
+
+      - name: Cache UI dependencies
+        uses: ./.github/actions/cache-ui-dependencies
+
+      - name: Cache QA Test dependencies
+        uses: ./.github/actions/cache-gradle-dependencies
+
+      - name: PR labels
+        uses: joerick/pr-labels-action@v1.0.6
+
+      - name: Setup Go build environment
+        run: |
+          if [[ -n "$GITHUB_PR_LABEL_CI_RELEASE_BUILD" ]]; then
+            cci-export GOTAGS release
+          fi
+          echo "GOTAGS: ${GOTAGS:-}"
+
+      - name: Fetch UI deps
+        run: make -C ui deps
+
+      - name: make style
+        run: make style
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -17,7 +17,7 @@ jobs:
       (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'turbo-build'))
     runs-on: ubuntu-latest
     container:
-      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.54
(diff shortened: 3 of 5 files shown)


gavin-stackrox:
I think this should be separate from builds i.e. a new workflow style.yaml. Though triggered by the same label. What do you think?

janisz:
Make sense. How about caches should this job has it's own cache? It updates go mod cache and golangci-lint cache

gavin-stackrox:
> Make sense. How about caches should this job has it's own cache? It updates go mod cache and golangci-lint cache

I think it should be fine to share the cache from the build steps but I'm not 100% clear what is cached :)

janisz:
> I think it should be fine to share the cache from the build steps but I'm not 100% clear what is cached :)

It's OK but it's not optimal as cache will be populated by the first job and `make style` install additional deps in go modules and populate golang ci cache. I think this step should have it's own cache.

.github/actions/cache-ui-dependencies/action.yaml
```diff
@@ -13,12 +13,15 @@ runs:
...
           ui/deps
+          /github/home/.cache/yarn/v6
           /github/home/.cache/Cypress
           /usr/local/share/.cache
           ui/node_modules
           ui/apps/platform/node_modules
           ui/packages/ui-components/node_modules
           ui/packages/tailwind-config/node_modules
-        key: npm-v2-${{ hashFiles(inputs.lockFile) }}
+        key: npm-v2--${{ hashFiles(inputs.lockFile) }}-${{ github.job }}
```
janisz, line 23: @gavin-stackrox Why do we need monorepo.lock? Why not using glob 
```
${{ hashFiles('**/package.json') }}
```
or even
```
${{ hashFiles('**/yarn.lock') }}
```

janisz:
@gavin-stackrox PTAL

gavin-stackrox:
@janisz style-checks in OSCI are now under 20 mins: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-style-checks?buildId=1575481918969876480

gavin-stackrox:
[hit send too early]
@janisz style-checks in OSCI are now under 20 mins: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-style-checks?buildId=1575481918969876480
So is there any real benefit to adding this?

janisz:
@gavin-stackrox That's great news! Let's close this PR. Month ago it was much longer sometimes 50m https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-style-checks?buildId=1568344902184472576

How was it possible to get so much speed up on OSCI? Are other builds faster as well?

gavin-stackrox:
@janisz the essential reason for the improvement is that the test image is mirrored into the CI cluster and the tests were changed to simpler 'single step' jobs so the ci-operator does not do as much image copying that was the source of a lot of the initial slowness. I did this for all the required jobs, so go-unit is also a lot quicker now. https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-go-unit-tests

Building though still is slow, so I think `turbo-build` shall live on unless that can get resolved. My best guess for building slowness is that it requires a few pods and allocating them is where OSCI can be arbitrary: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-images

msugakov:
Tomek, do you mind to try rebasing to the recent master?

janisz:
* master
  * **PR #2916** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2916" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2916?utm_source=stack-comment).

janisz:
@msugakov @gavin-stackrox @porridge

gavin-stackrox (approved):
The code change looks good to me. Any performance improvement?

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+
+  style-check:
+    if: |
+      (github.event.action == 'labeled' && github.event.label.name == 'turbo-build') ||
+      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'turbo-build'))
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.50
```
porridge, line 20: How do we make sure going forward that we bump this here consistently with other places? :thinking:

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
+name: Style
+on:
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    - labeled
```
msugakov, line 7: `reopened` is missing.

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.

From https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+env:
+  ROX_PRODUCT_BRANDING: RHACS_BRANDING
+
+jobs:
+
+  style-check:
+    if: |
+      (github.event.action == 'labeled' && github.event.label.name == 'turbo-build') ||
+      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'turbo-build'))
```
msugakov, line 17: Why not drop this `if` given that GitHub Actions migration is allowed and confirmed by now?

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: ./.github/actions/create-concatenated-ui-monorepo-lock
+
+      - name: Cache UI dependencies
+        uses: ./.github/actions/cache-ui-dependencies
+
+      - name: Cache QA Test dependencies
+        uses: ./.github/actions/cache-gradle-dependencies
+
+      - name: PR labels
+        uses: joerick/pr-labels-action@v1.0.6
```
msugakov, line 40: I would suggest not add 3-rd party dependency for the task to determine if a label is present on PR.
Some googling suggests that `${{contains( github.event.pull_request.labels.*.name, 'My Label')}}` will provide an answer if `My Label` is present.

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
...
+          if [[ -n "$GITHUB_PR_LABEL_CI_RELEASE_BUILD" ]]; then
+            cci-export GOTAGS release
+          fi
+          echo "GOTAGS: ${GOTAGS:-}"
+
+      - name: Fetch UI deps
+        run: make -C ui deps
+
+      - name: make style
+        run: make style
```
msugakov, line 53: Why not
```suggestion
        run: scripts/ci/jobs/style-checks.sh
```
?

.github/actions/cache-gradle-dependencies/action.yaml
```diff
@@ -0,0 +1,17 @@
...
+  using: composite
+  steps:
+    - name: Cache Gradle Dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+          ~/.m2/repository
+        key: npm-v2-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}-${{ github.job }}
```
msugakov, line 13: Why `npm-` or is it a typo?

.github/workflows/style.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Cache QA Test dependencies
+        uses: ./.github/actions/cache-gradle-dependencies
+
+      - name: PR labels
+        uses: joerick/pr-labels-action@v1.0.6
+
+      - name: Setup Go build environment
+        run: |
+          if [[ -n "$GITHUB_PR_LABEL_CI_RELEASE_BUILD" ]]; then
+            cci-export GOTAGS release
```
msugakov, line 45: Does `cci-export` work in GHA?

msugakov:
@gavin-stackrox 

> Any performance improvement?

:trollface: GHA 39 min, Prow 20 min.

.github/workflows/style.yaml
janisz: How about a script that will replace `image: quay.io/stackrox-io/apollo-ci:stackrox-test-*` with latest tag when rox images repo is tagged?

.github/workflows/style.yaml
janisz: I have a mixed feeling about this. In GHA we do not this wrapper and `make style` is enough.

.github/workflows/style.yaml
janisz: It's in our test container
```
docker run -it quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.54 /bin/bash -c cci-export
Usage: /bin/real-bash KEY VALUE
```

janisz: Ah, I got it. It's there but it's not working correctly :) Good catch

.github/workflows/style.yaml
janisz: I think this step is not required anyway but I'll fix build.yaml

.github/workflows/style.yaml
janisz: Let's do it

.github/workflows/style.yaml
janisz: https://github.com/stackrox/stackrox/pull/4049

.github/actions/cache-ui-dependencies/action.yaml
```diff
@@ -13,12 +13,15 @@ runs:
       with:
         path: |
           ui/deps
+          /github/home/.cache/yarn/v6
```
porridge, line 16: How about adding a note to wherever v6 is picked to update this line when changing?
Or strip `/v6` here?

porridge (approved):
LGTM but obviously we should not make this the default or required until we can reduce the execution time.

.github/actions/cache-ui-dependencies/action.yaml
janisz: I think we can drop version here so no need to update it. In theory we should obtain this path with `yarn cache dir` but then action need to have yarn installed

.github/workflows/style.yaml
msugakov: I'm ok if @gavin-stackrox is ok too

msugakov: Well, let me correct myself. The advantage of `scripts/ci/jobs/style-checks.sh` is that it saves junit reports in a standard location which _theoretically_ makes it easier to consume and/or mass-change logic. Maybe there's a value in that?

.github/workflows/style.yaml
msugakov: Well, GOTAGS thing is gone now, but it was done before. See e.g. this revision:
https://github.com/stackrox/stackrox/blob/ba9956b42f9b90169c2ee047a7c1723ea3eb5377/.circleci/config.yml#L2249-L2255

I wonder if had any effect or was just there for consistency?

msugakov: Reply from @janisz in DM:

> https://github.com/stackrox/stackrox/blob/master/Makefile#L137-L142 golangci runs for two tags It think other checks are not tags sensitive

PR #2919, opened 2022-08-31
Dashrews/ROX-12415 upgrade db copy fails

dashrews78:
## Description

Found an issue where if we failed to create a temp database we would return the temp clone anyway and thus it would create empty and cause all the data to be removed as the migration process continued.  While there discovered I could do renames of databases in a transaction so that we can prevent weird states where some clones were moved and others were not.  So that is the rationale behind the changes to doPersist so that we can wrap those calls in a transaction.

## Checklist

## Testing Performed

Added a unit test to drive some of the Postgres clone cases.

### DIFF
--- a/central/globaldb/v2backuprestore/restore/postgres.go
+++ b/central/globaldb/v2backuprestore/restore/postgres.go
@@ -84,33 +84,6 @@ func runRestoreStream(fileReader io.Reader, sourceMap map[string]string, config 
 	return nil
 }
 
-// SwitchToRestoredDB - switches the restore DB to be the active DB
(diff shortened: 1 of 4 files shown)


central/globaldb/v2backuprestore/restore/postgres.go
```diff
@@ -84,33 +84,6 @@ func runRestoreStream(fileReader io.Reader, sourceMap map[string]string, config
 	return nil
 }
 
-// SwitchToRestoredDB - switches the restore DB to be the active DB
```
dashrews78, line 87: was not being used from the initial upgrade work.

c-du (approved):

PR #2921, opened 2022-08-31
[ROX-6895] Remove Label search option ambiguity

md2119:
## Description

Remove ambiguity around _Label_ and _Annotation_ search for various resources. Add search labels that clearly inform what they are searching for. Furthermore, with this ambiguity, the search in Postgres does not work correctly. For example, searching deployments by namespace labels fails, because both have search label _Label_ and the search stops at deployment.

## Checklist

## Testing Performed

image

image

image

image


rukletsov (approved):
Thanks for picking this up!

PR #2926, opened 2022-09-01
[ROX-12212] : Optimize node and related GraphQL resolvers; Avoid pulling full node with scan and vulns by default in node datastore

charmik-redhat:
## Description

All below improvements are for Postgres.
- In node datastore and postgres store, add ability to avoid pulling full node with scan, components and vulns.
- In Node GraphQL, pull partial node excluding scan, until `Scan()` sub resolver is called, when full node is retrieved with scan, components and vulns.
- Cache embedded node scan component and other useful info in context and use that whenever possible instead of querying datastores.
- Cache embedded node scan vulns in context and use that when possible instead of querying datastores.

## Checklist

## Testing Performed

- Added unit tests and benchmark tests for node datastore and NodeScan GraphQL.
- Performed manual testing by calling nodes->nodeScan->nodeComponents->nodeVulnerabilities resolvers to verify the correctness of results.

### node_scan_benchmark_test results
```
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkNodeResolver
BenchmarkNodeResolver/GetNodeComponentsInNodeScanResolver
BenchmarkNodeResolver/GetNodeComponentsInNodeScanResolver-16         	      30	  40213754 ns/op
BenchmarkNodeResolver/GetNodeComponentsWithoutNodeScanResolver
BenchmarkNodeResolver/GetNodeComponentsWithoutNodeScanResolver-16    	       6	 179263362 ns/op
BenchmarkNodeResolver/GetNodeComponentsDerivedFieldsWithNodeScanResolver
BenchmarkNodeResolver/GetNodeComponentsDerivedFieldsWithNodeScanResolver-16         	      15	  85106791 ns/op
BenchmarkNodeResolver/GetNodeComponentsDerivedWithoutNodeScanResolver
BenchmarkNodeResolver/GetNodeComponentsDerivedWithoutNodeScanResolver-16            	       1	1384612231 ns/op
BenchmarkNodeResolver/GetNodeOnly
BenchmarkNodeResolver/GetNodeOnly-16                                                	    3465	    325572 ns/op
BenchmarkNodeResolver/GetNodeWithCounts
BenchmarkNodeResolver/GetNodeWithCounts-16                                          	      39	  28955174 ns/op
PASS
```

### Node datastore benchmark test results comparing GetNodesBatch (includes full node scan) and GetManyNodeMetadata (does not include node scan)
```
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkGetManyNodes
BenchmarkGetManyNodes/GetNodesBatch
BenchmarkGetManyNodes/GetNodesBatch-16         	      15	  98662333 ns/op
BenchmarkGetManyNodes/GetManyNodeMetadata
BenchmarkGetManyNodes/GetManyNodeMetadata-16   	    2220	    696879 ns/op
PASS
```

### DIFF
--- a/central/graphql/resolvers/embeddedobjs/node_component_context.go
+++ b/central/graphql/resolvers/embeddedobjs/node_component_context.go
@@ -1,0 +1,64 @@
+package embeddedobjs
+
+import (
+	"context"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// nodeComponentContextKey is the key for the *storage.EmbeddedNodeScanComponent value in the context.
+// All data is scoped to the embedding node.
+type nodeComponentContextKey struct{}
+
+// nodeComponentContextValue holds the value of the distro in the context.
+type nodeComponentContextValue struct {
+	component   *storage.EmbeddedNodeScanComponent
+	os          string
+	lastScanned *types.Timestamp
+}
+
+// NodeComponentContext returns a new context with the component attached.
+func NodeComponentContext(ctx context.Context, os string, lastScanned *types.Timestamp, component *storage.EmbeddedNodeScanComponent) context.Context {
+	return context.WithValue(ctx, nodeComponentContextKey{}, &nodeComponentContextValue{
+		component:   component,
+		os:          os,
+		lastScanned: lastScanned,
+	})
+}
+
+// NodeComponentFromContext returns the component from the input context.
+func NodeComponentFromContext(context context.Context) *storage.EmbeddedNodeScanComponent {
+	if context == nil {
+		return nil
+	}
+	value := context.Value(nodeComponentContextKey{})
+	if value == nil {
+		return nil
+	}
+	return value.(*nodeComponentContextValue).component
+}
+
+// NodeComponentLastScannedFromContext returns the last scanned time of the component, scoped to embedding node, from the input context.
+func NodeComponentLastScannedFromContext(context context.Context) *types.Timestamp {
+	if context == nil {
+		return nil
+	}
+	value := context.Value(nodeComponentContextKey{})
+	if value == nil {
+		return nil
+	}
+	return value.(*nodeComponentContextValue).lastScanned
+}
+
+// NodeComponentOSFromContext returns the operating system of the component, scoped to embedding node, from the input context.
+func NodeComponentOSFromContext(context context.Context) string {
+	if context == nil {
+		return ""
+	}
+	value := context.Value(nodeComponentContextKey{})
+	if value == nil {
+		return ""
+	}
+	return value.(*nodeComponentContextValue).os
+}
--- a/central/graphql/resolvers/embeddedobjs/node_component_context.go
+++ b/central/graphql/resolvers/embeddedobjs/node_component_context.go
@@ -14,15 +14,13 @@ type nodeComponentContextKey struct{}
 // nodeComponentContextValue holds the value of the distro in the context.
 type nodeComponentContextValue struct {
 	component   *storage.EmbeddedNodeScanComponent
-	os          string
 	lastScanned *types.Timestamp
 }
 
 // NodeComponentContext returns a new context with the component attached.
-func NodeComponentContext(ctx context.Context, os string, lastScanned *types.Timestamp, component *storage.EmbeddedNodeScanComponent) context.Context {
+func NodeComponentContext(ctx context.Context, lastScanned *types.Timestamp, component *storage.EmbeddedNodeScanComponent) context.Context {
 	return context.WithValue(ctx, nodeComponentContextKey{}, &nodeComponentContextValue{
 		component:   component,
-		os:          os,
 		lastScanned: lastScanned,
 	})
 }
@@ -49,16 +47,4 @@ func NodeComponentLastScannedFromContext(context context.Context) *types.Timesta
 		return nil
 	}
 	return value.(*nodeComponentContextValue).lastScanned
-}
-
-// NodeComponentOSFromContext returns the operating system of the component, scoped to embedding node, from the input context.
-func NodeComponentOSFromContext(context context.Context) string {
-	if context == nil {
-		return ""
-	}
-	value := context.Value(nodeComponentContextKey{})
-	if value == nil {
-		return ""
-	}
-	return value.(*nodeComponentContextValue).os
 }
--- a/central/graphql/resolvers/embeddedobjs/node_vuln_context.go
+++ b/central/graphql/resolvers/embeddedobjs/node_vuln_context.go
@@ -1,0 +1,36 @@
+package embeddedobjs
+
+import (
+	"context"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// nodeVulnContextKey is the key for the *storage.NodeVulnerability value in the context.
+type nodeVulnContextKey struct{}
+
+// nodeVulnContextValue holds the value of the distro in the context.
+type nodeVulnContextValue struct {
+	vuln        *storage.NodeVulnerability
+	lastScanned *types.Timestamp
+}
+
+// NodeVulnContext returns a new context with the vuln attached.
+func NodeVulnContext(ctx context.Context, vuln *storage.NodeVulnerability) context.Context {
+	return context.WithValue(ctx, nodeVulnContextKey{}, &nodeVulnContextValue{
+		vuln: vuln,
+	})
+}
+
+// NodeVulnFromContext returns the vuln from the input context.
+func NodeVulnFromContext(context context.Context) *storage.NodeVulnerability {
+	if context == nil {
+		return nil
+	}
+	value := context.Value(nodeVulnContextKey{})
+	if value == nil {
+		return nil
+	}
+	return value.(*nodeVulnContextValue).vuln
+}
--- a/central/graphql/resolvers/embeddedobjs/node_vuln_context.go
+++ b/central/graphql/resolvers/embeddedobjs/node_vuln_context.go
@@ -3,7 +3,6 @@ package embeddedobjs
 import (
 	"context"
 
-	"github.com/gogo/protobuf/types"
 	"github.com/stackrox/rox/generated/storage"
 )
 
@@ -12,8 +11,7 @@ type nodeVulnContextKey struct{}
 
 // nodeVulnContextValue holds the value of the distro in the context.
 type nodeVulnContextValue struct {
-	vuln        *storage.NodeVulnerability
-	lastScanned *types.Timestamp
+	vuln *storage.NodeVulnerability
 }
 
 // NodeVulnContext returns a new context with the vuln attached.
--- a/central/graphql/resolvers/loaders/nodes.go
+++ b/central/graphql/resolvers/loaders/nodes.go
@@ -78,8 +78,24 @@ func (ndl *nodeLoaderImpl) FromID(ctx context.Context, id string) (*storage.Node
 
 // FullNodeWithID loads full node from an ID.
 func (ndl *nodeLoaderImpl) FullNodeWithID(ctx context.Context, id string) (*storage.Node, error) {
-	//TODO implement me
-	panic("implement me")
+	node, err := ndl.FromID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	// Load the full node if full scan is not available.
+	if node.GetComponents() == 0 || len(node.GetScan().GetComponents()) > 0 {
+		return node, nil
+	}
+
+	ndl.lock.Lock()
+	delete(ndl.loaded, id)
+	ndl.lock.Unlock()
+
+	nodes, err := ndl.load(ctx, []string{id}, true)
+	if err != nil {
+		return nil, err
+	}
+	return nodes[0], nil
 }
 
 // FromQuery loads a set of nodes that match a query.
--- a/central/graphql/resolvers/loaders/nodes.go
+++ b/central/graphql/resolvers/loaders/nodes.go
@@ -82,7 +82,7 @@ func (ndl *nodeLoaderImpl) FullNodeWithID(ctx context.Context, id string) (*stor
 	if err != nil {
 		return nil, err
 	}
-	// Load the full node if full scan is not available.
+	// Do not load the full node if full scan is already available.
 	if node.GetComponents() == 0 || len(node.GetScan().GetComponents()) > 0 {
 		return node, nil
 	}
--- a/central/graphql/resolvers/loaders/nodes.go
+++ b/central/graphql/resolvers/loaders/nodes.go
@@ -9,6 +9,7 @@ import (
 	nodeDatastore "github.com/stackrox/rox/central/node/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/features"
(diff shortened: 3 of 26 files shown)


central/node/datastore/store/postgres/store.go
```diff
@@ -917,10 +917,68 @@ func (s *storeImpl) GetNodeMetadata(ctx context.Context, id string) (*storage.No
...
+		}
+	}
+	return elems, missingIndices, nil
+}
+
 //// Used for testing
 
 // CreateTableAndNewStore returns a new Store instance for testing
-func CreateTableAndNewStore(ctx context.Context, t *testing.T, db *pgxpool.Pool, gormDB *gorm.DB, noUpdateTimestamps bool) Store {
+func CreateTableAndNewStore(ctx context.Context, db *pgxpool.Pool, gormDB *gorm.DB, noUpdateTimestamps bool) Store {
```
keyallis, line 981: What's the reason for this change? I don't see a similar change in the Image equivalent PR

charmik-redhat: I removed that because testing.T argument was not used in the function, but now I think that the reason it was there was to ensure that the function gets called only during testing. I would revert the change.

dashrews78: looks like this was the only store that still had the testing reference.  So I presume that was a remnant from when the store was originally generated and before it diverged from the auto-generated code.  So the testing reference should probably be removed.

central/graphql/resolvers/embeddedobjs/node_component_context.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	value := context.Value(nodeComponentContextKey{})
+	if value == nil {
+		return nil
+	}
+	return value.(*nodeComponentContextValue).lastScanned
+}
+
+// NodeComponentOSFromContext returns the operating system of the component, scoped to embedding node, from the input context.
+func NodeComponentOSFromContext(context context.Context) string {
```
dashrews78, line 55: This doesn't seem to be used, so do we even need to store the os in the context?

central/graphql/resolvers/embeddedobjs/node_vuln_context.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// nodeVulnContextKey is the key for the *storage.NodeVulnerability value in the context.
+type nodeVulnContextKey struct{}
+
+// nodeVulnContextValue holds the value of the distro in the context.
+type nodeVulnContextValue struct {
+	vuln        *storage.NodeVulnerability
+	lastScanned *types.Timestamp
```
dashrews78, line 16: It does not look like you are ever setting or getting this field.  Do we need it?

central/graphql/resolvers/node_components.go
```diff
@@ -313,10 +340,71 @@ func (resolver *nodeComponentResolver) TopNodeVulnerability(ctx context.Context)
...
+			if err != nil {
+				return nil, err
+			}
+			resolver.ctx = embeddedobjs.NodeVulnContext(ctx, vuln)
+			idToVals[id] = resolver
+		}
+	}
+
+	// For now, sort by ID.
+	resolvers := make([]*nodeCVEResolver, 0, len(idToVals))
```
dashrews78, line 393: nit:  I think naming these so there is more distinction between resolvers and resolverI would be useful.  Also resolverI should probably be plural.

central/graphql/resolvers/node_scan.go
```diff
@@ -29,5 +37,55 @@ func (resolver *nodeScanResolver) NodeComponentCount(_ context.Context, args Raw
...
+			if err != nil {
+				return nil, err
+			}
+			resolver.ctx = embeddedobjs.NodeComponentContext(ctx, os, nodeScan.GetScanTime(), embeddedComponent)
+			idToComponent[id] = resolver
+		}
+	}
+
+	// For now, sort by IDs.
+	resolvers := make([]*nodeComponentResolver, 0, len(idToComponent))
```
dashrews78, line 74: nit: I think naming these so there is more distinction between resolvers and resolverI would be useful. Also resolverI should probably be plural.

central/graphql/resolvers/nodes.go
```diff
@@ -523,7 +527,23 @@ func (resolver *nodeResolver) PlottedNodeVulnerabilities(ctx context.Context, ar
...
+	scan := resolver.data.GetScan()
+	if features.PostgresDatastore.Enabled() {
+		// If scan is pulled, it is most likely to fetch all components and vulns contained in node.
+		// Therefore, load the node again with full scan.
+		nodeLoader, err := loaders.GetNodeLoader(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		// If Postgres is not enabled, node loader always pulls full node.
```
dashrews78, line 539: Not sure I'm following this comment since it is in the Postgres block

central/graphql/resolvers/testutils.go
```diff
@@ -46,7 +63,7 @@ func setupPostgresConn(t testing.TB) (*pgxpool.Pool, *gorm.DB) {
 	return pool, gormDB
 }
 
-func setupResolver(
+func setupResolverForImageGraphQLTestsWithPostgres(
```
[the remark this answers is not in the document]
dashrews78: agreed that name is really long.

central/node/datastore/store/dackbox/store_impl.go
```diff
@@ -112,7 +114,12 @@ func (b *storeImpl) GetNodeMetadata(_ context.Context, id string) (*storage.Node
 	return node, node != nil, err
 }
 
-// GetNodesBatch returns nodes with given ids.
+func (b *storeImpl) GetManyNodeMetadata(ctx context.Context, id []string) ([]*storage.Node, []int, error) {
```
dashrews78, line 117: you should add the function comment to be consistent.

central/node/datastore/store/postgres/store.go
```diff
@@ -67,8 +66,9 @@ type Store interface {
 	Delete(ctx context.Context, id string) error
 	GetIDs(ctx context.Context) ([]string, error)
 	GetMany(ctx context.Context, ids []string) ([]*storage.Node, []int, error)
-	// GetNodeMetadata gets the node without scan/component data.
+	// GetNodeMetadata and GetManyNodeMetadata returns the node without scan/component data.
```
dashrews78, line 69: nit:  my preference would be to put the comment on each function instead of grouping them.  But I understand why you are doing it this way.  So I leave it to you.

dashrews78 (changes requested):
I didn't see anything super major, but some stuff you should clean up.  Additionally, I think I'd like to see it after a rebase since this PR is pretty stale.

central/graphql/resolvers/testutils.go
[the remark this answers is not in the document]
charmik-redhat: Yeah, changed the name

central/graphql/resolvers/embeddedobjs/node_component_context.go
charmik-redhat: Ah good catch, I removed the function

central/graphql/resolvers/embeddedobjs/node_vuln_context.go
charmik-redhat: Yeah, removed it

central/graphql/resolvers/nodes.go
charmik-redhat: Ah yes, the comment should be before the Postgres block.

central/graphql/resolvers/loaders/nodes.go
```diff
@@ -58,16 +60,38 @@ type nodeLoaderImpl struct {
...
+	if err != nil {
+		return nil, err
+	}
+	// Load the full node if full scan is not available.
+	if node.GetComponents() == 0 || len(node.GetScan().GetComponents()) > 0 {
+		return node, nil
+	}
+
+	ndl.lock.Lock()
+	delete(ndl.loaded, id)
```
c-du, line 91: I did not see codes to expire entries in loaded, will it become full node eventually? Of course, that is not introduced in this PR.

central/graphql/resolvers/node_scan_benchmark_test.go
```diff
@@ -0,0 +1,176 @@
+package resolvers
```
c-du, line 1: Can you post some benchmark results here?

central/graphql/resolvers/loaders/nodes.go
```diff
@@ -58,16 +60,38 @@ type nodeLoaderImpl struct {
...
 		return nil, err
 	}
 	return nodes, nil
 }
 
-// FromID loads a node from an ID.
+// FromID loads a node from an ID. Does not load scan components and vulnerabilities when Postgres is enabled.
 func (ndl *nodeLoaderImpl) FromID(ctx context.Context, id string) (*storage.Node, error) {
-	nodes, err := ndl.load(ctx, []string{id})
+	nodes, err := ndl.load(ctx, []string{id}, false)
```
c-du, line 72: It would be helpful if we have metrics around load cache here. But this is out of scope for this PR.

c-du (approved):
LGTM overall, just a few questions.

central/graphql/resolvers/embeddedobjs/node_component_context.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// nodeComponentContextKey is the key for the *storage.EmbeddedNodeScanComponent value in the context.
+// All data is scoped to the embedding node.
+type nodeComponentContextKey struct{}
```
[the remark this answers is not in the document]
charmik-redhat: We use uniquely named structs as keys within contexts. Also, context doesn't recommend using string or any builtin type as keys.

central/graphql/resolvers/loaders/nodes.go
charmik-redhat: Yes it will become full node eventually, and the next time same node it queried, it will not fetch full node again.

central/graphql/resolvers/node_scan_benchmark_test.go
charmik-redhat: Do I add it in comment here? Or in the top of file?

central/graphql/resolvers/loaders/nodes.go
charmik-redhat: Yeah, it would be useful to know especially memory usages, to see if the cache takes up too much space.

dashrews78 (approved):
LGTM

central/graphql/resolvers/node_components.go
```diff
@@ -246,7 +304,22 @@ func (resolver *nodeComponentResolver) NodeCount(ctx context.Context, args RawQu
...
-	return resolver.root.NodeVulnerabilities(resolver.nodeComponentScopeContext(ctx), args)
+
+	if resolver.ctx == nil {
+		resolver.ctx = ctx
+	}
+
+	// Short path. Full node is embedded when node scan resolver is called.
+	embeddedComponent := embeddedobjs.NodeComponentFromContext(resolver.ctx)
+	if embeddedComponent == nil || !env.PostgresDatastoreEnabled.BooleanSetting() {
+		return resolver.root.NodeVulnerabilities(resolver.nodeComponentScopeContext(ctx), args)
```
keyallis, line 315: Should this line be using `ctx` or `resolver.ctx`?

central/graphql/resolvers/node_scan.go
```diff
@@ -29,5 +42,52 @@ func (resolver *nodeScanResolver) NodeComponentCount(_ context.Context, args Raw
 }
 
 func (resolver *nodeScanResolver) NodeComponents(_ context.Context, args PaginatedQuery) ([]NodeComponentResolver, error) {
```
keyallis, line 44: Do we need the:
```go
if resolver.ctx == nil {
resolver.ctx = ctx
}
```

central/graphql/resolvers/node_scan_benchmark_test.go
keyallis: If you could include it in the PR description

central/graphql/resolvers/nodes.go
```diff
@@ -546,3 +570,15 @@ func (resolver *nodeResolver) nodeScopeContext(ctx context.Context) context.Cont
 		ID:    resolver.data.GetId(),
 	})
 }
+
+//// Node scan-related fields pulled as direct sub-resolvers of node.
+
+func (resolver *nodeResolver) ScanNotes(ctx context.Context) []string {
+	value := resolver.data.GetScan().GetNotes()
+	return stringSlice(value)
```
keyallis, line 578: nit: this can be a one line, same with `ScanTime`

keyallis (commented):
Couple of comments

central/graphql/resolvers/node_components.go
charmik-redhat: No it should still use ctx because `resolver.nodeComponentScopeContext` func also does below assignment
```
if resolver.ctx == nil {
 		resolver.ctx = ctx
 	}
```
It is redundant in this case, but we need non-nil resolver.ctx when calling `getNodeCVEResolvers` on ln 322 below. resolver.nodeComponentScopeContext get called only within the if condition, so it won't be always executed and assign resolver.ctx to a non-nil value.

central/graphql/resolvers/node_scan.go
charmik-redhat: nodeScanResolver is created by `Scan()` sub-resolver is nodes graphQL and that sub-resolver adds the node's scope to nodeScanResolver's ctx before returning. So, nodeScanResolver should have a non-nil ctx at this point. But I think it is safe to add the check here in case the `Scan` sub-resolver changes.

central/graphql/resolvers/loaders/nodes.go
```diff
@@ -58,16 +60,38 @@ type nodeLoaderImpl struct {
...
+	return nodes[0], nil
+}
+
+// FullNodeWithID loads full node from an ID.
+func (ndl *nodeLoaderImpl) FullNodeWithID(ctx context.Context, id string) (*storage.Node, error) {
+	node, err := ndl.FromID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	// Load the full node if full scan is not available.
```
md2119, line 85: The comment does not describe the if block below. A converse comment would work or move the comment after the if block.

central/graphql/resolvers/nodes.go
```diff
@@ -519,7 +523,23 @@ func (resolver *nodeResolver) PlottedNodeVulnerabilities(ctx context.Context, ar
 }
 
 func (resolver *nodeResolver) Scan(ctx context.Context) (*nodeScanResolver, error) {
-	res, err := resolver.root.wrapNodeScan(resolver.data.GetScan(), true, nil)
+	// If Postgres is not enabled, node loader always pulls full node along with scan
+	scan := resolver.data.GetScan()
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		// If scan is pulled, it is most likely to fetch all components and vulns contained in node.
```
md2119, line 529: `...it is most likely for the user to fetch...`

md2119 (approved):
Nice job in getting this done!

keyallis (approved):
LGTM

PR #2928, opened 2022-09-01
ROX-12428: [refactor] Limit scope of the `edit` role and move the control over SCCs to a separate role

vikin91:
## Description

This PR creates a separate RBAC role to control the `use` verb for sensor.
Before, sensor was affected by the `use` setting from the `edit` role.

SCC `stackrox-sensor` is no longer assigned using the `users:` field. Instead the permission to use the SCCs is provided by the role `use-sccs-for-sensor` and role binding `sensor-use-sccs-for-sensor`.

Note that this PR is mainly refactoring as sensor can still run under `anyuid` SCC.
A PR to change this will be opened as a follow-up (or added here if reviewers wish so).

## Checklist

I have not added any additional tests for the SCCs, but we may investigate such possibility in a follow-up.

## Testing Performed

- [x] Manually OCP 4.11, install with helm, `--set=system.createSCCs=false` -> uses SCCs in the following order: `anyuid`  > `nonroot-v2` > `nonroot` 
- [x] Manually OCP 4.11, install with helm, `--set=system.createSCCs=true` -> uses SCCs in the following order: `anyuid`  > `stackrox-sensor` > `nonroot-v2`  > `nonroot` 
- [x] Manually on K8s with and without `createSCCs` - expect to see no errors (SCCs should be ignored)

### DIFF
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
@@ -53,6 +53,8 @@ roleRef:
   name: stackrox:view-cluster
   apiGroup: rbac.authorization.k8s.io
 ---
+# Role edit has all verbs but 'use' to disallow using any SCCs (resources: *).
+# The permission to 'use' SCCs should be defined at finer granularity in other roles.
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
@@ -69,7 +69,14 @@ rules:
   resources:
   - '*'
   verbs:
-  - '*'
+  - create
+  - get
+  - list
+  - watch
+  - update
+  - patch
+  - delete
+  - deletecollection
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -1,7 +1,7 @@
 {{- include "srox.init" . -}}
 
 {{- if and ._rox.env.openshift ._rox.system.createSCCs }}
-
+---
 apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
@@ -40,6 +40,8 @@ volumes:
   - downwardAPI
   - emptyDir
   - secret
+
+---
 
 {{- end }}
 
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -61,7 +61,7 @@ rules:
   resources:
   - securitycontextconstraints
   resourceNames:
-{{- if and ._rox.env.openshift ._rox.system.createSCCs }}
+{{- if ._rox.system.createSCCs }}
   - stackrox-sensor
 {{- end }}
   - nonroot-v2
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -63,9 +63,7 @@ rules:
 {{- if and ._rox.env.openshift ._rox.system.createSCCs }}
   - stackrox-sensor
 {{- end }}
-{{- if eq ._rox.env.openshift 4 }}
   - nonroot-v2
-{{- end }}
   - nonroot
   - anyuid
   verbs:
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -61,7 +61,9 @@ rules:
 {{- if and ._rox.env.openshift ._rox.system.createSCCs }}
   - stackrox-sensor
 {{- end }}
+{{- if eq ._rox.env.openshift 4 }}
   - nonroot-v2
+{{- end }}
   - nonroot
   - anyuid
   verbs:
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -13,6 +13,11 @@ metadata:
   annotations:
     {{- include "srox.annotations" (list . "securitycontextconstraints" "stackrox-sensor") | nindent 4 }}
     kubernetes.io/description: stackrox-sensor is the security constraint for the sensor
+{{- if eq ._rox.env.openshift 3 }}
+users:
+  - system:serviceaccount:{{ ._rox._namespace }}:sensor
+  - system:serviceaccount:{{ ._rox._namespace }}:sensor-upgrader
+{{- end }}
 priority: 0
 runAsUser:
   type: RunAsAny
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -65,9 +65,6 @@ rules:
   resources:
   - securitycontextconstraints
   resourceNames:
-{{- if ._rox.system.createSCCs }}
-  - stackrox-sensor
-{{- end }}
   - nonroot-v2
   - nonroot
   - anyuid
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -13,11 +13,9 @@ metadata:
   annotations:
     {{- include "srox.annotations" (list . "securitycontextconstraints" "stackrox-sensor") | nindent 4 }}
     kubernetes.io/description: stackrox-sensor is the security constraint for the sensor
-{{- if eq ._rox.env.openshift 3 }}
 users:
   - system:serviceaccount:{{ ._rox._namespace }}:sensor
   - system:serviceaccount:{{ ._rox._namespace }}:sensor-upgrader
-{{- end }}
 priority: 0
 runAsUser:
   type: RunAsAny
--- a/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
@@ -71,14 +71,33 @@ rules:
   - use
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: Role
 metadata:
-  name: sensor-use-scc
+  name: use-stackrox-sensor-scc
   namespace: {{ ._rox._namespace }}
   labels:
-    {{- include "srox.labels" (list . "rolebinding" "sensor-use-scc") | nindent 4 }}
+    {{- include "srox.labels" (list . "role" "use-stackrox-sensor-scc") | nindent 4 }}
   annotations:
-    {{- include "srox.annotations" (list . "rolebinding" "sensor-use-scc") | nindent 4 }}
+    {{- include "srox.annotations" (list . "role" "use-stackrox-sensor-scc") | nindent 4 }}
+rules:
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  resourceNames:
+  - stackrox-sensor
+  verbs:
+  - use
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: sensor-use-anyuid-scc
+  namespace: {{ ._rox._namespace }}
+  labels:
+    {{- include "srox.labels" (list . "rolebinding" "sensor-use-anyuid-scc") | nindent 4 }}
+  annotations:
+    {{- include "srox.annotations" (list . "rolebinding" "sensor-use-anyuid-scc") | nindent 4 }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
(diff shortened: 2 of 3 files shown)


vikin91:
Deploying this PR with the following:  

```shell
bin/darwin/roxctl central generate openshift --openshift-version 4 pvc
cd central-bundle
helm install -n stackrox --create-namespace stackrox-central-services chart/
cd ..
bin/darwin/roxctl sensor generate openshift --openshift-version 4 --name cluster1 --password $(cat central-bundle/password)
./sensor-cluster1/sensor.sh
```

results in `openshift.io/scc: stackrox-sensor`.

porridge:
```
cp: cannot stat 'bin/linux_amd64/roxctl': No such file or directory
error: build error: error building at STEP "RUN ./.openshift-ci/build/build-main-and-bundle.sh": error while running runtime: exit status 1
```
I think this was breakage reverted by @misberner late last week.
@vikin91 please rebase so we can make sure the tests pass

operator/tests/common/delete-securedcluster-errors.yaml
```diff
@@ -372,11 +372,6 @@ metadata:
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
-metadata:
-  name: sensor-use-scc
```
porridge, line 376: Shouldn't this be renamed to `  name: sensor-use-sccs-for-sensor` rather than removed?

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
```diff
@@ -69,7 +69,14 @@ rules:
   resources:
   - '*'
   verbs:
```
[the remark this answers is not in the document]
porridge: And by the way, did you make sure SCCs indeed the only use case for the `use` verb?
Otherwise we could be taking away some niche privilege...

porridge:
I also have a question: what exactly do you mean by `uses SCCs in the following order`? Were you deleting SCCs one by one and checking which one will be used next in turn?

porridge:
And FTR: I very much prefer PRs to be as small as possible so :+1: to keeping this refactoring separate!

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -44,51 +41,53 @@ volumes:
...
 - apiGroups:
   - security.openshift.io
   resources:
   - securitycontextconstraints
   resourceNames:
+{{- if and ._rox.env.openshift ._rox.system.createSCCs }}
+  - stackrox-sensor
+{{- end }}
+{{- if eq ._rox.env.openshift 4 }}
+  - nonroot-v2
```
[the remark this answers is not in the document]
vikin91: Yes, this is a heuristic and I arrived at it in the following way:
- We can write there any name, the SCC is not required to exist
- If we want to use `nonroot-v2` we must be on OCP 4
- If we run on OCP v4 we MAY be lucky to have `nonroot-v2` available

So based on this, we could:
- add `nonroot-v2` unconditionally
- add `nonroot-v2` conditionally 

I choose the second option but the first one would also work.

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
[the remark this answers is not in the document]
vikin91: You mean "mention" here in the PR or in the code as a comment?

We remove verb `use` from the `*`, because it would allow us to `use` any SCC and thus all other Roles that concern SCCs would be shadowed by this one.

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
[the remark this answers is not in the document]
porridge: I don't mind either way, but I think unconditional might end up slightly less confusing.

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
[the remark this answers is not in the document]
porridge: In the code as comment.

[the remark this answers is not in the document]
vikin91: > And by the way, did you make sure SCCs indeed the only use case for the `use` verb? Otherwise we could be taking away some niche privilege...

I am not sure how to verify whether there are any other verbs here - read the source? The docs mention only `use` everywhere when it comes to consuming the SCC by a pod. I guess `get`, `create`, `delete` could also have a meaning in this context, but whatever gets created won't be able to be used.

vikin91:
> I also have a question: what exactly do you mean by `uses SCCs in the following order`? Were you deleting SCCs one by one and checking which one will be used next in turn?

Yes, exactly. I see what SCC is used by default, then I disallow it and see what is the fallback, and so on.

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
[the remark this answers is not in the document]
porridge: My approach would be:
- see if this verb exists in vanilla k8s. SCCs are an openshift-only feature.
- if they do, ask in #sig-api-machinery on kubernetes slack perhaps, if the upstream docs don't mention it?

operator/tests/common/delete-securedcluster-errors.yaml
vikin91: I replaced 2 bindings with 1, so one should be removed and the other renamed. The renaming is in this file in line 296.

Note that the `Role` `use-sccs-for-sensor` is mentioned here twice: once for `rbac.authorization.k8s.io/v1` and once for `authorization.openshift.io/v1`

vikin91: I think I see now that the `RoleBinding` should be also twice for both API versions. I will fix that

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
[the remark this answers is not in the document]
misberner: I'm not sure if `use` has any use outside of PSPs and SCCs, but I am pretty sure that we do not rely on any `use` privileges in the sensor code.

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -44,51 +41,53 @@ volumes:
...
   annotations:
-    {{- include "srox.annotations" (list . "role" "use-anyuid-scc") | nindent 4 }}
+    {{- include "srox.annotations" (list . "role" "use-sccs-for-sensor") | nindent 4 }}
 rules:
 - apiGroups:
   - security.openshift.io
   resources:
   - securitycontextconstraints
   resourceNames:
+{{- if and ._rox.env.openshift ._rox.system.createSCCs }}
```
misberner, line 63: Note that the entire file should be conditional under `{{ if ._rox.env.openshift }}`, as there are no SCCs in vanilla k8s.

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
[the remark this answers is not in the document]
misberner: Keep in mind that this PR will require detailed update instructions for customers using the bundle installation method. Therefore, you generally want your resources to be as uniform as possible. Hence +1 for unconditional

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -44,51 +41,53 @@ volumes:
...
+  - stackrox-sensor
```
misberner, line 64: I guess this can be made unconditional as well? But I'm not sure if it should

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -44,51 +41,53 @@ volumes:
...
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: sensor-use-scc
+  name: sensor-use-sccs-for-sensor
   namespace: {{ ._rox._namespace }}
   labels:
-    {{- include "srox.labels" (list . "rolebinding" "sensor-use-scc") | nindent 4 }}
+    {{- include "srox.labels" (list . "rolebinding" "sensor-use-stackrox-sensor-scc") | nindent 4 }}
```
misberner, line 80: the argument must match the name of the resource

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -44,51 +41,53 @@ volumes:
...
   annotations:
-    {{- include "srox.annotations" (list . "rolebinding" "sensor-use-scc") | nindent 4 }}
+    {{- include "srox.annotations" (list . "rolebinding" "sensor-use-stackrox-sensor-scc") | nindent 4 }}
```
misberner, line 82: here again

operator/tests/common/delete-securedcluster-errors.yaml
```diff
@@ -293,7 +293,7 @@ metadata:
 apiVersion: authorization.openshift.io/v1
 kind: RoleBinding
 metadata:
-  name: sensor-use-scc
+  name: sensor-use-sccs-for-sensor
```
misberner, line 296: I appreciate the clear naming, but one thing to keep in mind is that in the manual upgrade instructions, you will need to specify `oc rm scc ... ; oc create -f - <<EOF ...` commands for all the renames.

image/templates/helm/stackrox-secured-cluster/templates/sensor-rbac.yaml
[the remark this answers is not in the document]
vikin91: - Comment added in the code 
- The verb `use` appears in the vanilla k8s in context of PodSecurityPolicies ([example](https://docs.giantswarm.io/getting-started/rbac-and-psp/)). They come however first in `k8s.io/api/policy/v1beta1` 
- As far as I can see in the k8s API code, the verbs are just strings and an API is allowed to define custom set of verbs.

So, if we want to disallow `use` for `SCC`s only, then we would need to additionally restrict the resources, so that `use` is allowed for all but `SCC`.
We could replace the `'*`` for `resources with a list of all resources, e.g., from [here](https://kubernetes.io/docs/reference/kubectl/#resource-types). This may be a lot of entries, as OpenShift will add their own types.
Alternatively, we could allow both `*-resource` and `*-verb` on selected api groups, but we would end up listing all of them but one as we cannot allow `*` minus one particular item.

What do you think?

operator/tests/common/delete-securedcluster-errors.yaml
vikin91: Ok, this convinces me to stay with old names :)

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
vikin91: I left it conditional for now, let me know if we should remove the `if`

porridge:
Note the qa test seems to have found an issue in your change:
```
      upgrade_status_detail: "Upgrader failed to execute InstantiateBundleStage of the roll-forward workflow: executing stage \"Instantiate objects from sensor bundle\": object metadata validation failed errors: [upgrade label auto-upgrade.stackrox.io/component of object stackrox/use-anyuid-scc[rbac.authorization.k8s.io/v1, Kind=Role] has invalid value \"\", expected: \"sensor\", upgrade label auto-upgrade.stackrox.io/component of object stackrox/sensor-use-scc[rbac.authorization.k8s.io/v1, Kind=RoleBinding] has invalid value \"\", expected: \"sensor\"]"
```

vikin91:
> Another potential problem is that using Roles/RoleBindings to assign SCC privileges is (IIRC) an OpenShift 4 feature. I seem to recall now that on OpenShift 3 the SCC needs the users list that you removed. Are we no longer supporting OpenShift 3 for sensor? 🤔

I think, we do support it. I just didn't know that the Roles are first available in OpenShift 4. I will check that and restore the `users` section if needed.

vikin91:
I have added the `auto-upgrade` labels as suggested in the error message and restored the block `users` in the SCC, but wrapped it in `if eq ._rox.env.openshift 3` - this will hopefully allow us to remember that this block is there for backwards compat with OCP v3.

vikin91:
We could also totally skip the Role and RoleBinding if we are not on OCP v4. This is how it used to be before. Maybe the Role and RoleBinding appearing on v3.11 would be confusing for the admins?

vikin91:
FYI, I am closing this one https://github.com/stackrox/stackrox/pull/2839 as we have found a better way of addressing the issue in this PR (or its follow-up where we probably disable the permission to run under `anyuid` SCC).

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -12,9 +13,11 @@ metadata:
   annotations:
     {{- include "srox.annotations" (list . "securitycontextconstraints" "stackrox-sensor") | nindent 4 }}
     kubernetes.io/description: stackrox-sensor is the security constraint for the sensor
+{{- if eq ._rox.env.openshift 3 }}
```
misberner, line 16: My 2c: leave this as unconditional. This change will only cause pain when having to write upgrade instructions, and unless I am not mistaken, does not provide any real benefit (even if we are adding RBAC bindings, these entries here don't hurt anything we want to accomplish). Also, I thought that we wanted to migrate away from custom SCCs soon-ish. Why would we make any changes to them?

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
misberner: With the `users` section coming back, I guess this can be removed? It will make upgrade instructions easier

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -66,6 +67,11 @@ rules:
   resources:
   - securitycontextconstraints
   resourceNames:
+{{- if ._rox.system.createSCCs }}
+  - stackrox-sensor
+{{- end }}
+  - nonroot-v2
+  - nonroot
```
misberner, line 74: I understand that you list both `nonroot-v2` and `nonroot` here, as the former might not always be available. But what about `nonroot` and `anyuid`? Would we ever want `anyuid` to be used? Can there be cases where only `anyuid` is available, but not `nonroot`? We should be specific here as to what SCC we want to be used.

misberner: I think I also don't quite understand why we remove the `use` verb from the `edit` role above, but then create an RBAC binding "officially" using the SCC which was previously accidentally used

misberner (changes requested):

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
vikin91: I wanted to split the work of getting rid of `anyuid` for sensor into 2 PRs:
- this one, where we refactor things to prepare everything and catch possible errors,
- and a follow-up where we remove a single line `- anyiud` and so make the sensor to use `nonroot` or `nonroot-v2`.

If we stayed with the full-powered `edit` role, the follow-up PR would not work, because the `edit` role would still be allowing to use the `anyuid` SCC.

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
vikin91: Ok, I can remove it. I wanted to avoid that OCP v4 users would get the permission to use the custom SCC from 2 different sources, but you are right, we do not hurt any of our goals making this unconditional.

vikin91:
@misberner could you please re-review my answers and recent changes? I would like to move forward with this PR and open the promised follow-up where we remove the permission to use `anyuid`.

misberner (approved):
Sorry for the delay

PR #2929, opened 2022-09-01
ROX-12032: Don't show edges for other deployments if Network Policy CIDR block is in public range

fredrb:
## Description

This is a proposal of how to mitigate a Network Graph issue related to showing allowed edges between deployments if the egress Network Policy associated with a deployment uses public CIDR blocks. More info about the problem can be found [in this doc](https://docs.google.com/document/d/1W27OxPRr32kYzi77JsP0WmO-IgmB8_cjfOGhs5GM0uA/edit?usp=sharing).

The proposed change is: **Don't show allowed edges if Network Policy's CIDR block is in public range**. 

Some considerations:
- Deployments are scheduled within private CIDR blocks, so a public CIDR range network policy won't match.
- Deployments could be behind a service with a public IP. This could be considered as an allowed edge, but we're ignoring this case for now.

The proposal in this PR doesn't solve all problems. The set of allowed edges might not be accurate for CIDR block, but it limits the cases where they are added. 

## Checklist

It's likely that we should add note in the docs and/or README about this change.

## Testing Performed

- New unit tests for this case

**Note for reviewers:** I had to change existing unit tests in `graph_builder_test` to use private IP range for external entities in the network tree wrapper. Because the graph builder should act the same way as before, but _only_ for private IPs. I've added two additional tests to cover the changed behavior. 

- Manual testing using the yaml files that exemplify the issue found [here](https://github.com/stackrox/yaml-examples/tree/main/network-graph-cidr-egress) 

Prior to this change, an edge between `ns3` and `ns2` would be shown. 

image

### DIFF
--- a/central/networkpolicies/graph/graph_builder.go
+++ b/central/networkpolicies/graph/graph_builder.go
@@ -2,6 +2,7 @@ package graph
 
 import (
 	"fmt"
+	"net"
 	"sort"
 	"strings"
 
@@ -114,15 +115,22 @@ func (b *graphBuilder) evaluatePeers(currentNS *storage.NamespaceMetadata, peers
 func (b *graphBuilder) evaluatePeer(currentNS *storage.NamespaceMetadata, peer *storage.NetworkPolicyPeer) []*node {
 	if peer.GetIpBlock() != nil {
 		var allNodes []*node
+		_, ipNet, err := net.ParseCIDR(peer.GetIpBlock().GetCidr())
+		if err != nil {
+			log.Warnf("Failed to parse CIDR block: %s", err)
+			return allNodes
+		}
+
 		// If the IP is in the private range, we add edges to other deployments in the cluster.
 		// This is still not the perfect approach, but it solves user issues where edges were being
 		// shown on *every* deployment with a Network Policy that uses CIDR block matchers. By
 		// limiting to private ranges only, the set of edges in the graph is likely more accurate, but
 		// still not the reality. There is an epic created to tackle this further and come up with a
 		// more elaborate logic for this: ROX-12120
-		if netutil.IsCIDRBlockInPrivateSubnet(peer.GetIpBlock().GetCidr()) {
+		if netutil.IsIPSubNetOverlapingPrivateRange(ipNet) {
 			allNodes = append(allNodes, b.allDeployments...)
 		}
+
 		allNodes = append(allNodes, b.evaluateExternalPeer(peer.GetIpBlock())...)
 		return allNodes
 	}
--- a/central/networkpolicies/graph/graph_builder.go
+++ b/central/networkpolicies/graph/graph_builder.go
@@ -9,6 +9,7 @@ import (
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/labels"
+	"github.com/stackrox/rox/pkg/netutil"
 	"github.com/stackrox/rox/pkg/networkgraph"
 	"github.com/stackrox/rox/pkg/networkgraph/tree"
 	"github.com/stackrox/rox/pkg/set"
@@ -112,11 +113,16 @@ func (b *graphBuilder) evaluatePeers(currentNS *storage.NamespaceMetadata, peers
 
 func (b *graphBuilder) evaluatePeer(currentNS *storage.NamespaceMetadata, peer *storage.NetworkPolicyPeer) []*node {
 	if peer.GetIpBlock() != nil {
-		// TODO(ROX-5370): We assume all CIDR blocks always match all deployments and overlapping external CIDRs.
-		//  This is probably wrong, but we don't really have a good way of determining otherwise. Except for maybe
-		//  look at Pod IPs. Which we actually could.
-		allNodes := make([]*node, 0, len(b.allDeployments)+1)
-		allNodes = append(allNodes, b.allDeployments...)
+		var allNodes []*node
+		// If the IP is in the private range, we add edges to other deployments in the cluster.
+		// This is still not the perfect approach, but it solves user issues where edges were being
+		// shown on *every* deployment with a Network Policy that uses CIDR block matchers. By
+		// limiting to private ranges only, the set of edges in the graph is likely more accurate, but
+		// still not the reality. There is an epic created to tackle this further and come up with a
+		// more elaborate logic for this: ROX-12120
+		if netutil.IsCIDRBlockInPrivateSubnet(peer.GetIpBlock().GetCidr()) {
+			allNodes = append(allNodes, b.allDeployments...)
+		}
 		allNodes = append(allNodes, b.evaluateExternalPeer(peer.GetIpBlock())...)
 		return allNodes
 	}
--- a/central/networkpolicies/graph/graph_builder.go
+++ b/central/networkpolicies/graph/graph_builder.go
@@ -127,7 +127,7 @@ func (b *graphBuilder) evaluatePeer(currentNS *storage.NamespaceMetadata, peer *
 		// limiting to private ranges only, the set of edges in the graph is likely more accurate, but
 		// still not the reality. There is an epic created to tackle this further and come up with a
 		// more elaborate logic for this: ROX-12120
-		if netutil.IsIPSubNetOverlapingPrivateRange(ipNet) {
+		if netutil.IsIPNetOverlapingPrivateRange(ipNet) {
 			allNodes = append(allNodes, b.allDeployments...)
 		}
 
--- a/central/networkpolicies/graph/graph_builder_test.go
+++ b/central/networkpolicies/graph/graph_builder_test.go
@@ -19,7 +19,7 @@ func TestMatchPolicyPeer(t *testing.T) {
 			Desc: &storage.NetworkEntityInfo_ExternalSource_{
 				ExternalSource: &storage.NetworkEntityInfo_ExternalSource{
 					Source: &storage.NetworkEntityInfo_ExternalSource_Cidr{
-						Cidr: "192.16.0.0/16",
+						Cidr: "192.168.0.0/16",
 					},
 				},
 			},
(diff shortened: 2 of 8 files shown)


pkg/netutil/cidr.go
```diff
@@ -3,9 +3,37 @@ package netutil
 import (
 	"net"
 
+	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/utils"
 )
 
+var (
+	privateSubnets = []*net.IPNet{
```
misberner, line 11: Imho it would be good to consolidate this list somewhere in the codebase, as it occurs multiple times (with some differences). The values in `pkg/net/addr.go` are probably a good starting point.

pkg/netutil/cidr.go
```diff
@@ -3,9 +3,37 @@ package netutil
...
+		MustParseCIDR("172.16.0.0/12"),  // class B
+		MustParseCIDR("192.168.0.0/16"), // class C
+		MustParseCIDR("::1/128"),        // IPv6 localhost
+		MustParseCIDR("fd00::/8"),       // IPv6 ULA
+	}
+	log = logging.LoggerForModule()
+)
+
+// IsCIDRBlockInPrivateSubnet parses cidrStr and checks if it falls under the RFC 1819 private IP range
+func IsCIDRBlockInPrivateSubnet(cidrStr string) bool {
```
misberner, line 23: For a library function in `netutil`, it would be more natural to operate on a `*net.IPNet` rather than possibly duplicating the parsing effort.

pkg/netutil/cidr.go
```diff
@@ -3,9 +3,37 @@ package netutil
...
+	_, cidr, err := net.ParseCIDR(cidrStr)
+	if err != nil {
+		log.Warnf("Parsing incorrect CIDR block string (%s) as CIDR %s", cidrStr, err)
```
misberner, line 26: Low-level library functions should not log, this can be very spammy.

central/networkpolicies/graph/graph_builder.go
```diff
@@ -112,11 +113,16 @@ func (b *graphBuilder) evaluatePeers(currentNS *storage.NamespaceMetadata, peers
...
-		allNodes := make([]*node, 0, len(b.allDeployments)+1)
-		allNodes = append(allNodes, b.allDeployments...)
+		var allNodes []*node
+		// If the IP is in the private range, we add edges to other deployments in the cluster.
+		// This is still not the perfect approach, but it solves user issues where edges were being
+		// shown on *every* deployment with a Network Policy that uses CIDR block matchers. By
+		// limiting to private ranges only, the set of edges in the graph is likely more accurate, but
+		// still not the reality. There is an epic created to tackle this further and come up with a
+		// more elaborate logic for this: ROX-12120
+		if netutil.IsCIDRBlockInPrivateSubnet(peer.GetIpBlock().GetCidr()) {
```
misberner, line 123: I forget what the outcome of our discussion was, but I would still say that overlap is the property that would need to be checked, not inclusion.

The way the code is currently written, `0.0.0.0/0` would _not_ result in edges to deployments being created.

pkg/netutil/cidr.go
```diff
@@ -3,9 +3,37 @@ package netutil
 import (
 	"net"
 
+	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/utils"
 )
 
+var (
+	privateSubnets = []*net.IPNet{
+		MustParseCIDR("127.0.0.0/8"),    // IPv4 localhost
```
misberner, line 12: Note that this IP range (along with `::1`) is completely irrelevant to the problem at hand. I'm not even sure that a `127.0.0.0/8` CIDR block in a network policy has any effect at all.

misberner (changes requested):

pkg/netutil/cidr.go
fredrb: Took a shoot at refactoring this. To avoid keeping too many different subsets the insecure_registries checks (`central/compliance/checks/common/insecure_registries.go` and `pkg/compliance/checks/common/insecure_registries.go`) are now also checking private IPv4 addresses mapped to IPv6 values. Let me know what do you think.

pkg/netutil/cidr.go
fredrb: Pulled the warning up to the caller of this function.

central/networkpolicies/graph/graph_builder.go
fredrb: Changed to overlap. Thanks!

pkg/netutil/cidr.go
fredrb: This is now removed as per the refactor I mentioned in the first comment. Check the latest commit: [0b1881a](https://github.com/stackrox/stackrox/pull/2929/commits/0b1881a15843dc5aca4d325b7454204a97efa327)

fredrb:
@misberner could you please take a look at this again? Thanks!

central/networkpolicies/graph/graph_builder_test.go
```diff
@@ -58,14 +58,29 @@ func TestMatchPolicyPeer(t *testing.T) {
...
+						Cidr: "30.30.0.0/32",
 					},
 				},
 			},
 		},
 	})
 	assert.NoError(t, err)
-	_, _, _ = t1, t2, t3
+
+	_, _, _, _ = t1, t2, t3, t4
```
misberner, line 83: I don't really understand what this is for. Can you try if removing it makes any difference?

pkg/netutil/private_subnet.go
```diff
@@ -0,0 +1,51 @@
...
+	IPV4MappedIPv6Loopback = MustParseCIDR("::ffff:127.0.0.1/104")
+)
+
+// GetPrivateSubnets returns a slice of IPv4 and IPv6 addresses considered as private ranges including localhost addresses.
+func GetPrivateSubnets() []*net.IPNet {
+	var subnets []*net.IPNet
+	subnets = append(subnets, IPV4PrivateNetworks...)
+	subnets = append(subnets, IPV4LocalHost)
+	subnets = append(subnets, IPV6PrivateNetworks...)
+	subnets = append(subnets, IPV6LocalHost)
```
misberner, line 49: This is missing `IPv4MappedIPv6Loopback`, no?

pkg/netutil/private_subnet.go
```diff
@@ -0,0 +1,51 @@
...
+		MustParseCIDR("::ffff:172.16.0.0/108"),
+		MustParseCIDR("::ffff:192.168.0.0/112"),
+
+		// Other reserved IPv4 ranges
+		MustParseCIDR("::ffff:100.64.0.0/106"),
+		MustParseCIDR("::ffff:169.254.0.0/112"),
+	}
+
+	// IPV6LocalHost is the local host IP range ::1/128.
+	IPV6LocalHost = MustParseCIDR("::1/128")
```
misberner, line 37:
```suggestion
	IPv6LocalHost = MustParseCIDR("::1/128")
```

pkg/netutil/private_subnet.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// IPV4MappedIPv6Loopback is the IPv4 loopback address mapped in IPv6 range.
+	IPV4MappedIPv6Loopback = MustParseCIDR("::ffff:127.0.0.1/104")
```
misberner, line 40:
```suggestion
	IPv4MappedIPv6Loopback = MustParseCIDR("::ffff:127.0.0.1/104")
```

pkg/netutil/private_subnet.go
```diff
@@ -0,0 +1,51 @@
...
+var (
+	// IPV4PrivateNetworks holds RFC1918 addresses plus other reserved IPv4 ranges.
+	IPV4PrivateNetworks = []*net.IPNet{
+		// private networks per RFC1918
+		MustParseCIDR("10.0.0.0/8"),
+		MustParseCIDR("172.16.0.0/12"),
+		MustParseCIDR("192.168.0.0/16"),
+
+		// Other reserved ranges
+		MustParseCIDR("100.64.0.0/10"),
```
misberner, line 14: Just a heads-up that the definition in `central/compliance/checks/common/insecure_registries.go` was missing those. I believe adding them causes no harm, but just wanted to point that out.

misberner (approved):
LGTM

central/networkpolicies/graph/graph_builder_test.go
fredrb: Done

pkg/netutil/private_subnet.go
fredrb: Added

pkg/netutil/private_subnet.go
fredrb: Renamed all other variables to use lower-case `v`

pkg/netutil/private_subnet.go
fredrb: Renamed all other variables to use lower-case `v`

pkg/compliance/checks/common/insecure_registries.go
```diff
@@ -31,7 +20,7 @@ func insecureRegistries(info *compliance.ContainerRuntimeInfo) []*storage.Compli
 		_, cidr, _ := net.ParseCIDR(cidrStr)
 		isPrivate := false
 		if cidr != nil {
-			for _, privateSubnets := range privateSubnets {
+			for _, privateSubnets := range netutil.GetPrivateSubnets() {
 				if netutil.IsIPNetSubset(privateSubnets, cidr) {
```
lvalerom, line 24:
```suggestion
			for _, privateSubnet := range netutil.GetPrivateSubnets() {
				if netutil.IsIPNetSubset(privateSubnet, cidr) {
```

lvalerom (approved):
Looks good to me.

pkg/compliance/checks/common/insecure_registries.go
lvalerom: [nit]☝️

PR #2935, opened 2022-09-01
Changelog for substring string search

md2119:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ authentication is still required for an email notifier, but the user can now cho
   - Any cluster that is expected to be unavailable for a period of time (e.g. clusters used in disaster recovery), can be tagged with a customizable label. Clusters with those labels will never be removed automatically.
   - By default, this unhealthy cluster removal is disabled (number of days set to 0)
 - ROX-7591: Policy `Fixable CVSS >= 6 and Privileged` disabled by default on new installations, new policy `Severity Important and Privileged` added and enabled by default.
-- Search by string expressions will support substring match starting 3.73, in addition to current support for prefix string match.
+- Starting 3.73, the default technique used by string expression searches will be substring matching.
 
 ## [3.71.0]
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ authentication is still required for an email notifier, but the user can now cho
   - Any cluster that is expected to be unavailable for a period of time (e.g. clusters used in disaster recovery), can be tagged with a customizable label. Clusters with those labels will never be removed automatically.
   - By default, this unhealthy cluster removal is disabled (number of days set to 0)
 - ROX-7591: Policy `Fixable CVSS >= 6 and Privileged` disabled by default on new installations, new policy `Severity Important and Privileged` added and enabled by default.
+- Starting 3.73, the default technique used by string expression searches will be substring matching.
```
theencee, line 23:
```suggestion
- Starting 3.73, the default technique used by string expression searches will be to match any substrings instead of the prefix.
```

maybe?

md2119: I thought about adding prefix case, but then it is actually not true  in all cases from user's perspective.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ authentication is still required for an email notifier, but the user can now cho
   - Any cluster that is expected to be unavailable for a period of time (e.g. clusters used in disaster recovery), can be tagged with a customizable label. Clusters with those labels will never be removed automatically.
   - By default, this unhealthy cluster removal is disabled (number of days set to 0)
 - ROX-7591: Policy `Fixable CVSS >= 6 and Privileged` disabled by default on new installations, new policy `Severity Important and Privileged` added and enabled by default.
+- Starting 3.73, the default technique used by string expression searches will be to match any substrings instead of the prefix.
```
connorgorman, line 23: this needs to be tied to postgres release because it will not change for bleve, how were we doing that before?

md2119: I think we should not state the actual release target and just call it out that this is a future deprecation.

theencee: This is assuming that we are shipping postgres in 73 and there's no more bleve. If that changes, we can just add an update to the changelog then, right?

theencee (approved):

CHANGELOG.md
connorgorman: Yes, but perhaps we should split out anything that depends on Postgres into a separate block so it's simpler to just remove that block and mark as 73?

PR #2937, opened 2022-09-01
Add visitWithGenericResponses helper function for integration tests

pedrottimark:
## Description

Extend and adjust naming convention for helper functions before presentation to the team.

### Changed files

1. Edit cypress/helpers/visit.js
    * Export alias constants for callers of `visitWithGenericResponses` function.
    * Add `availableAuthProviders` generic request for completeness.
    * Add `loginAuthProviders` generic request especially for auth tests.
    * Rename `visitWithPermissions` as `visitWithPermissionsResponse` function.
    * Add `visitWithGenericResponses` function.

2. Edit cypress/integration/accessControl/accessControlAuthProviders.test.js
    * Rename `visitWithPermissions` as `visitWithPermissionsResponse` function and also its argument.

3. Edit cypress/integration/certexpiration.test.js
    * Replace local helper with `requestConfig` and `staticResponseMap` arguments of `visitMainDashboard` function.

### Residue

1. ~Call `visitWithGenericResponses` function~ in auth.test.js file. Rewriting `stubAPIs` and `setupAuth` functions might be more subtle than that.

## Checklist
- [x] Edited integration test and helper functions

## Testing Performed

* accessControl/accessControlAuthProviders.test.js: 1 request with fixture has open green circle
    accessControlAuthProviders
* auth.test.js
* certexpiration.test.js: 2 requests with fixture and body have open green circle
    certexpiration

* network.test.js
* networkGraph/networkGraph.test.js

### DIFF
--- a/ui/apps/platform/cypress/integration/certexpiration.test.js
+++ b/ui/apps/platform/cypress/integration/certexpiration.test.js
@@ -1,71 +1,102 @@
 import dateFns from 'date-fns';
-import withAuth from '../helpers/basicAuth';
+
 import * as api from '../constants/apiEndpoints';
 import { selectors } from '../constants/CertExpiration';
-import { visitMainDashboard } from '../helpers/main';
+import { systemConfigUrl } from '../constants/SystemConfigPage';
+import withAuth from '../helpers/basicAuth';
+import {
+    credentialexpiryCentralAlias,
+    credentialexpiryScannerAlias,
+    mypermissionsAlias,
+    visitWithGenericResponses,
+} from '../helpers/visit';
+
+/*
+ * Visit System Configuration page for certificate expiration tests:
+ * It has a minimal number of page-specific requests.
+ * The credentialexpiry and mypermissions requests are generic for any page.
+ */
+function visitForCredentialExpiry(staticResponseMapGeneric) {
+    visitWithGenericResponses(systemConfigUrl, staticResponseMapGeneric, {
+        config: {
+            method: 'GET',
+            url: api.system.config,
+        },
+    });
+}
 
 describe('Cert Expiration Banner', () => {
     withAuth();
 
-    const mockCertExpiryAndVisitHomepage = (endpoint, expiry) => {
-        cy.intercept('GET', endpoint, {
-            body: { expiry },
-        }).as('certExpiry');
-        visitMainDashboard();
-        cy.wait('@certExpiry');
-    };
-
     describe('Central', () => {
         it('should not display banner if central cert is expiring more than 14 days later', () => {
             const expiry = dateFns.addHours(dateFns.addDays(new Date(), 15), 1);
-            mockCertExpiryAndVisitHomepage(api.certExpiry.central, expiry);
+            const staticResponseMapGeneric = {
+                [credentialexpiryCentralAlias]: {
+                    body: { expiry },
+                },
+            };
+            visitForCredentialExpiry(staticResponseMapGeneric);
 
             cy.get(selectors.centralCertExpiryBanner).should('not.exist');
         });
 
         it('should display banner without download button if user does not have the required permission', () => {
-            cy.intercept('GET', api.permissions.mypermissions, {
-                body: {
-                    globalAccess: 'READ_ACCESS',
-                    resourceToAccess: {
-                        VulnerabilityManagementRequests: 'READ_ACCESS',
-                        VulnerabilityManagementApprovals: 'READ_ACCESS',
-                    },
-                },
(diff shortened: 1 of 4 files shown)


ui/apps/platform/cypress/integration/certexpiration.test.js
```diff
@@ -1,125 +1,175 @@
...
     describe('Central', () => {
         it('should not display banner if central cert is expiring more than 14 days later', () => {
             const expiry = dateFns.addHours(dateFns.addDays(new Date(), 15), 1);
-            mockCertExpiryAndVisitHomepage(api.certExpiry.central, expiry);
+            const staticResponseMapGeneric = {
+                [credentialexpiryCentralAlias]: {
+                    body: { expiry },
+                },
+            };
+            visitForCredentialExpiry(staticResponseMapGeneric);
```
sachaudh, line 39: The naming for this function seems a bit different than the others. I think it's also a bit hard to understand what it's doing just by the name. If I had to define the purpose of this function I would say "visit the main dashboard page with credential expiry mocked". We have a number of options that would be descriptive enough. One could be
`visitMainDashboardWithCredentialExpiryResponse`

ui/apps/platform/cypress/integration/certexpiration.test.js
```diff
@@ -1,125 +1,175 @@
...
-            cy.wait('@permissions');
+            const staticResponseMapGeneric = {
+                [mypermissionsAlias]: {
+                    fixture: 'auth/mypermissionsMinimalAccess.json',
+                },
+                [credentialexpiryCentralAlias]: {
+                    body: { expiry },
+                },
+            };
+            visitForCredentialExpiry(staticResponseMapGeneric);
```
sachaudh, line 54: I guess considering this use case and my last comment, maybe using something like `visitMainDashboardWithCredentialExpiryResponse` as the function name would not be ideal because technically we're now mocking the `mypermissions` response too. Maybe a more generic function name would be better than to consider this and the previous use case

sachaudh: Also, the alias naming seems a bit odd because we've been using this weird casing that's not quite camel case (`mypermissionsAlias` vs. `myPermissionsAlias`). I believe that stemmed from the backend naming convention. I guess that's something worth discussing with the backend team as well. Maybe we can try to limit them on these case styles so we don't have these weird situations

ui/apps/platform/cypress/integration/certexpiration.test.js
```diff
@@ -1,125 +1,175 @@
...
 
             cy.get(selectors.centralCertExpiryBanner).should('have.class', 'pf-m-danger');
         });
 
         it('should download the YAML', () => {
             const expiry = dateFns.addDays(new Date(), 1);
-            mockCertExpiryAndVisitHomepage(api.certExpiry.central, expiry);
+            const staticResponseMapGeneric = {
+                [credentialexpiryCentralAlias]: {
```
sachaudh, line 92: I was wondering about the possibility of defining a reusable object like:
```
const centralCredentialExpiryResponse = {
    [credentialexpiryCentralAlias]: {
        body: { expiry },
    }
}

const mypermissionsResponse = {
    [mypermissionsAlias]: {
        fixture: 'auth/mypermissionsMinimalAccess.json',
    }
}
```
Then we can reuse blocks of these for the `staticResponseMapGeneric` objects
```
const staticResponseMapGeneric = {
    ...centralCredentialExpiryResponse,
    ...mypermissionsResponse
}
```

dvail (commented):
I was having trouble tracing the intended flow here at first, mostly due to the usage of `static` and `generic` qualifiers in the naming and how they related to `non-static` or `non-generic` arguments of similar structure.

It looks like the goal here is to separate the "intercept/mock response/wait" cycle for both a set of standard (generic) requests that are made on every page and a user provided list of specific requests for the page under test.

What are the trade offs in keeping both the `visit` and `visitWithGenericResponse` functions as a single function, with a signature similar to:
```typescript
function visit(pageUrl: string, options: {} = { requestConfig: {}, mockedResponseMap: {} }) {}
```
and then spreading the passed `options.requestConfig` into `requestConfigGeneric` for a single set of `intercetpRequests/waitForResponses` calls.

Would that give us the same functionality, without the need for an explicit separation between generic/static/specific config objects?

I know this would remove our ability to intercept a set of requests after the set of generic requests and would also remove the ability to wait for a non-mocked generic request followed by immediately waiting for a _mocked_ generic request.

pedrottimark:
Superseded by #3577 and #3603

PR #2938, opened 2022-09-01
ROX-11033: Clear selection to avoid table interaction crashes

vjwilson:
## Description

Brute-force fix for these crashes, rather than refactoring the several components/hooks involved, because we are going to be reworking all of vulnerability management.

## Checklist

## Testing Performed

Rather than provided screenshots of pages working-as-expected, I list the manual tests I did for each of the 3 tabbed tables on the Image Overview page: Observed CVEs, Deferred CVEs, and False Positive CVEs.

1. Visit VulnMgmt -> Images
2. Scroll down to the embedded PatternFly CVE list
3. Select one or more CVEs by clicking the checkboxes
4. Perform any of the following actions at the top of the table

- Change the table "page"
- Change the number of items per page
- Apply a search filter
- Click the top heading of a column


dvail (approved):
I know the ViolationsTable uses the same hooks in a similar way without crashing and without the explicit calls to `onClearAll()`, but it isn't immediately apparent why.

Is it too risky before release to modify `useTableSelection` to do a safe check in `getSelectedIds` for the value of `data[i]`? It seems like that shouldn't have any ill effect on other components that use the hook that aren't already crashing.

If so, an easily understandable brute-force fix seems like a good option to fix some pretty bad crashes to get this into 3.72.

vjwilson:
Closing in favor of a simpler solution,
https://github.com/stackrox/stackrox/pull/3037

PR #2945, opened 2022-09-01
Revert "Support multi-arch binaries (#2820)"

misberner:
This reverts commit 8773303de209234ae5988ada7793c865d96d44f0.

PR #2946, opened 2022-09-01
Support multi-arch binaries

yselkowitz:
This reverts #2945 and reinstates #2820.

/cc @misberner


misberner:
This seems to have succeeded:
```
$ docker run quay.io/stackrox-io/roxctl:3.71.x-473-g99ebf5aeb2 version
3.71.x-473-g99ebf5aeb2
```

misberner (approved):

misberner:
Merging because this is basically the same as the earlier PR which passed all tests, and I have manually verified that the roxctl image works

janisz:
This breaks `turbo-build`

PR #2950, opened 2022-09-02
Fix local operator build/start

misberner:
## Description

I tried to `make -C operator install run` yesterday, and this failed for several reasons:
- `.SHELLFLAGS` are set to `-u`, so the check for CI only works if the CI variable is defined (it not being defined is kind of the point of the check, though)
- `ensure_image_exists.sh` was working with a garbled `$main_image_tag` (it said `Ensuring quay.io/stackrox-io/main:make[1] Entering directory ...`). Tell `make` to really shut up by passing `--no-print-directory` in addition to `--quiet`.
- More of a related improvement, only try to build the image if the build would actually result in the desired image.

## Checklist

## Testing Performed

- Manual testing, it worked afterwards

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -290,7 +290,7 @@ stackrox-image-pull-secret: ## Create default image pull secret for StackRox ima
 
 .PHONY: check-ci-setup
 check-ci-setup: ## Make sure this target is not started in CI environment.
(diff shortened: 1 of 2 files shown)


vladbologa (approved):

misberner:
Remaining tests don't use operator

PR #2951, opened 2022-09-02
Delete spec.relatedImages from operator CSV

misberner:
## Description

Since [recently upgrading to a newer operator-sdk version](https://github.com/stackrox/stackrox/pull/2749), the generation of the ClusterServiceVersion included an [empty `relatedImages` entry](https://github.com/stackrox/stackrox/pull/2749/files#diff-31342b68c83ec8bf58029a16b4c4047c13b714cdb2b245a94d1ef2ea89d4b076R918-R920). We don't know why this was added, but it breaks the downstream image pinning logic, where we set the `RELATED_IMAGE_` environment variables and the relevant pipeline stage cannot handle presence of both the env vars and the `spec.relatedImages` field.

Since the value of the new entry is unclear and it doesn't seem to affect upstream builds, let's remove it in post-processing.

## Checklist

## Testing Performed

- Ran generation locally, observed that entry was gone

### DIFF
--- a/operator/bundle_helpers/patch-csv.py
+++ b/operator/bundle_helpers/patch-csv.py
@@ -75,6 +75,8 @@ def patch_csv(csv_doc, version, operator_image, first_version, no_related_images
         else:
             csv_doc["spec"]["replaces"] = f'{raw_name}.v{x}.{y}.{z-1}'
 
+    # We don't know what this does or why it is there, but it breaks downstream builds.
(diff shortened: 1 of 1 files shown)


vikin91 (approved):
Love the comment 😄

msugakov:
Context thread https://srox.slack.com/archives/C02AD1GUA0Y/p1662079598390879

PR #2957, opened 2022-09-02
Fix variable unbound in upgrade

c-du:
## Description

Fix a break to the gke-upgrade test in variable unbond.

## Checklist

## Testing Performed
Run existing tests.

### DIFF
--- a/tests/upgrade/lib.sh
+++ b/tests/upgrade/lib.sh
@@ -60,7 +60,7 @@ validate_upgrade() {
     local upgrade_cluster_id="$3"
     local policies_dir="../pkg/defaults/policies/files"
 
-    if [[ -n ${API_TOKEN:-} ]]; then
+    if [[ -n "${API_TOKEN:-}" ]]; then
         info "Verifying API token generated can access the central"
         echo $API_TOKEN | $TEST_ROOT/bin/$TEST_HOST_OS/roxctl --insecure-skip-tls-verify --insecure -e "$API_ENDPOINT" --token-file /dev/stdin central whoami > /dev/null
     fi
--- a/tests/upgrade/lib.sh
+++ b/tests/upgrade/lib.sh
@@ -62,7 +62,7 @@ validate_upgrade() {
 
     if [[ -n "${API_TOKEN:-}" ]]; then
         info "Verifying API token generated can access the central"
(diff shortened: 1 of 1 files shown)


tests/upgrade/lib.sh
```diff
@@ -60,7 +60,7 @@ validate_upgrade() {
     local upgrade_cluster_id="$3"
     local policies_dir="../pkg/defaults/policies/files"
 
-    if [ -n $API_TOKEN ]; then
+    if [[ -n ${API_TOKEN:-} ]]; then
```
dashrews78, line 63: I think you need quotes around it too.
"${API_TOKEN:-}"

c-du: yes

dashrews78 (approved):
LGTM as long as it passes.

tests/upgrade/lib.sh
```diff
@@ -60,7 +60,7 @@ validate_upgrade() {
     local upgrade_cluster_id="$3"
     local policies_dir="../pkg/defaults/policies/files"
 
-    if [ -n $API_TOKEN ]; then
+    if [[ -n "${API_TOKEN:-}" ]]; then
         info "Verifying API token generated can access the central"
         echo $API_TOKEN | $TEST_ROOT/bin/$TEST_HOST_OS/roxctl --insecure-skip-tls-verify --insecure -e "$API_ENDPOINT" --token-file /dev/stdin central whoami > /dev/null
```
gavin-stackrox, line 65: Can you also quote these things to keep bash happy? 
```suggestion
        echo "$API_TOKEN" | "$TEST_ROOT/bin/$TEST_HOST_OS/roxctl" --insecure-skip-tls-verify --insecure -e "$API_ENDPOINT" --token-file /dev/stdin central whoami > /dev/null
```

gavin-stackrox (changes requested):

gavin-stackrox:
Hi @c-du, LGTM but unfortunately it needs a label to force the upgrade test to actually do something (I added that). And then:
/test gke-upgrade-tests
should get a result in an hour or so.

gavin-stackrox (approved):

gavin-stackrox:
If this passes it should be good to commit: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/2957/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1565810956024418304

PR #2958, opened 2022-09-02
Add changelog for label and annotation search option deprecation

md2119:
## Description

We will need to remove the ambiguity around searching by `Label` and `Annotation` for the namespace, deployments, pods, and images at the very least for search on those search options to work correctly. I tested out the change in #2921 and things look good in CI and manual testing.

Given that we will need to do the change for aforementioned resources, we might as well fix it for all the resources that use `Label` and `Annotation` search label.

## Checklist
- [ ] Documented user facing changes

## Testing Performed
Refer to the CI of #2921


md2119:
Reviewers, ping on this PR. We are closer to 72 release cut and we need to make a decision if we want to carry over the ambiguity or not.

theencee (approved):

PR #2961, opened 2022-09-03
Fix chains_test.go by avoiding deprecated SHA1-based signature algorithm

misberner:
## Description

Re-generate the certificates with SHA256 used for signatures (and add a script to facilitate future re-generation, should that ever be necessary).

SHA1 is deprecated in Go 1.18+, which makes the test fail.

## Checklist

## Testing Performed

- Ran adapted unit test


connorgorman (approved):

connorgorman:
Ship?

misberner:
Shipped

PR #2964, opened 2022-09-05
Bump gradle to 7.5

janisz:
## Description

https://docs.gradle.org/7.5.1/release-notes.html

## Testing Performed

CI


rukletsov (approved):
Could you please include the "why" in PRs like this? Is it to fix a CVE, to get a feature we need, or just bump to the most recent version out of boredom. This is help for both reviewing and triaging when something goes wrong.

janisz:
> just bump to the most recent version out of boredom

I like to keep dependencies up to date as it's usually easier to keep them that way than once a year or two do a big update and resolving multiple issues with deprecated API.

PR #2975, opened 2022-09-06
Fix dashboard slicing of compliance data when there are more than 6 standards

dvail:
## Description

As discussed in this [Slack thread](https://srox.slack.com/archives/C02TXEA9WKF/p1662034816902929), the display of compliance data on the dashboard when there are more than six standards is incorrect. The current implementation sorts the list of all standards either ascending or descending, and then takes the first six items. However, since the PF charts display data items from bottom to top this results in the ascending sort showing the _top_ six standards from lowest to highest, and the _bottom_ six standards from highest to lowest.

What makes more sense is effectively the opposite of this:

- “Sort by Ascending” shows the six _lowest_ standards by compliance percentage, ordered low to high (this is the default)
- “Sort by Descending” shows the six _highest_ standards by compliance percentage, ordered high to low

### Follow Up

Add test cases that cover situations where more than six standards are returned via the API. (e.g. when the openshift compliance operator is installed.)

## Checklist

## Testing Performed

Against an OpenShift installation containing the compliance operator, which adds additional compliance standards, the default "ascending" view should show the _worst_ standards by compliance percentage, ordered from lowest to highest.
image

Sorting the list by "descending" in turn displays the six _best_ standards by compliance percentage, ordered from highest to lowest. 
image

Note that in this instance there are 9 total standards and the middle 3 (ocp4-cis-node, CIS Kubernetes, HIPAA 164) are visible in the appropriate places in each chart.

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.tsx
@@ -90,7 +90,7 @@ function processData(
 
     const sorted = sortBy(modifiedData, [(datum) => datum.passing]);
 
-    if (sortDirection === 'asc') {
+    if (sortDirection === 'desc') {
         sorted.reverse();
     }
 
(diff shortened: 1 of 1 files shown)


dvail:
Note: this is still awaiting approval by PM, but I believe it is the desired behavior so I wanted to get this PR out asap in case we were able to get it in before the release.

pedrottimark (approved):
Thank you for following through.

After release, it seems worthwhile to take another look at the helper functions. Although I might be missing some relevant context because of the focused code change, being able to slice up to 6 standards in an array which has been filtered and sorted seems clearer to read and safer to maintain.

dvail:
@pedrottimark you are absolutely right, this was a bug with the way I was sorting the data and your suggestion is much clearer. This PR didn't sit right with me in the back of my mind last night and the bad sorting is the reason why.

PR #2987, opened 2022-09-07
ROX-11751: Implement SAC tests for ComponentCVEEdge and NodeComponentCVEEdge

rhybrillou:
## Description

With the migration to postgres, both dackbox and scoped access control are being redesigned.
In order to ensure the provided functionality remains stable, some test are being added at datastore level for dackbox objects.
The change at hand focuses on the Component -> CVE edge of the dackbox graphs:
Cluster -> Namespace -> Deployment -> Image -> ImageComponent -> CVE
Cluster -> Node -> NodeComponent -> CVE

## Checklist

## Testing Performed

This Change only adds tests. Code changes if any is required by the tests will come in a prefactor PR.


central/nodecomponentcveedge/datastoretest/nodecomponentcveedge_from_componentcveedge_datastore.go
```diff
@@ -0,0 +1,60 @@
+package datastoretest
```
janisz, line 1: If it's used only in `datastore` package then maybe better is to put it in `internal` package or `_test` file

janisz (approved):

PR #2988, opened 2022-09-07
ROX-12487: Fix handling of proxy-related environment variables in the operator

misberner:
## Description

See bug for a more detailed description of the issue.

This PR changes the proxy-related environment variable injection logic to always give precedence to environment variables specified in the CR in the customize section, ignoring the respective values from the Operator's environment if both are set.

## Checklist

## Testing Performed

- Unit tests
- CI (tbd)

### DIFF
--- a/operator/pkg/proxy/translation.go
+++ b/operator/pkg/proxy/translation.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-func getProxyConfigHelmValues(obj k8sutil.Object, proxyEnvVars map[string]string) (chartutil.Values, error) {
+func getProxyConfigEnvVars(obj k8sutil.Object, proxyEnvVars map[string]string) (map[string]interface{}, error) {
 	if len(proxyEnvVars) == 0 {
 		return nil, nil
 	}
@@ -37,11 +37,7 @@ func getProxyConfigHelmValues(obj k8sutil.Object, proxyEnvVars map[string]string
 		}
 	}
 
-	return chartutil.Values{
-		"customize": map[string]interface{}{
-			"envVars": envVarsMap,
-		},
-	}, nil
+	return envVarsMap, nil
 }
 
 // InjectProxyEnvVars wraps a Translator to inject proxy configuration environment variables.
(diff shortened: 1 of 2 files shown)


operator/pkg/proxy/translation.go
```diff
@@ -52,7 +48,33 @@ func InjectProxyEnvVars(translator values.Translator, proxyEnv map[string]string
 			return nil, err
 		}
 
-		proxyVals, _ := getProxyConfigHelmValues(obj, proxyEnv) // ignore errors for now
-		return chartutil.CoalesceTables(vals, proxyVals), nil
+		proxyEnvVars, _ := getProxyConfigEnvVars(obj, proxyEnv) // ignore errors for now
```
porridge, line 51: Um, what do you mean by "for now"? :confused:

porridge (commented):
Just one thing that jumped at me, unfortunately the soonest I'd be able to take a proper look will be Monday.

SimonBaeumer (approved):

operator/pkg/proxy/translation.go
misberner: Valid question, given that the comment had been there for over a year now.

I've added an explicit log plus an explanation as to why we're not bubbling up the error.

operator/pkg/proxy/translation.go
```diff
@@ -37,22 +40,54 @@ func getProxyConfigHelmValues(obj k8sutil.Object, proxyEnvVars map[string]string
...
 		}
 
-		proxyVals, _ := getProxyConfigHelmValues(obj, proxyEnv) // ignore errors for now
-		return chartutil.CoalesceTables(vals, proxyVals), nil
+		proxyEnvVars, err := getProxyConfigEnvVars(obj, proxyEnv)
+		if err != nil {
+			// Simply log the error, we do not want to fail reconciliation based on this (the check for
+			// len(proxyEnvVars) == 0) should catch a complete failure).
+			// While this log can be spammy (emitted on every reconciliation), an error here is extremely unlikely
+			// and thus we deem this acceptable.
```
porridge, line 59: I'm not sure I agree. I'm worried about the conversion to unstructured failing for some obscure reason. My experience with the kubernetes types/converstions voodoo is that it is  notorious for causing failures at a distance - you forget to register some scheme somewhere, and another part of your program blows up with mysterious message :-(
Ignoring an error here, causing part of the functionality to disappear, would be even worse.

misberner: Hmm, but `corev1.EnvVar` is a fairly simple structure that is also unlikely to change significantly in the future. I find it unlikely that something goes wrong at runtime that is not detected in the unit test (*) (because the configuration surface error exposed to the user is even smaller). If this breaks, something must be really weird, and I don't know if just disrupting the entire functionality of the operator is a better choice, especially given that there exists an escape hatch of specifying proxy env vars via `spec.customize`.

However, I do admit that given that we have a separate status condition for proxy configuration, it would have been better to surface any errors there rather than just in the logs. But I would still say that this is unlikely to be consequential.

(*) Yes, I realize(d after typing the above) that we don't actually check for an error in the unstructured conversion, but we do check that values are correctly piped through.

misberner: Updating the status would require a helm operator library change, unfortunately..

misberner:
@porridge I'd like to get this merged, as I think it is better to have this change than not have this change. As per your concern regarding just logging the error, I would still say that this is better than breaking reconciliation entirely - because _if_ there's some unexpected error, I don't know what a user could do to get things back to a working state, whereas manually fixing proxy config by setting env vars in the CR is doable. LMK if you agree.

porridge:
> @porridge I'd like to get this merged, as I think it is better to have this change than not have this change. As per your concern regarding just logging the error, I would still say that this is better than breaking reconciliation entirely - because _if_ there's some unexpected error, I don't know what a user could do to get things back to a working state, whereas manually fixing proxy config by setting env vars in the CR is doable. LMK if you agree.

I think the errors that `ToUnstructured` could return are of the "should never happen" kind, i.e. something is very broken, and we should not ship an operator in that state.

However, how about [this approach](https://github.com/stackrox/stackrox/pull/3693)?

misberner:
> However, how about [this approach](https://github.com/stackrox/stackrox/pull/3693)?

Yeah, that should work 😅

PR #2989, opened 2022-09-07
Ensure read from dackbox join tables is scoped-access-controlled in postgres mode

rhybrillou:
## Description

The integration tests written for SAC against ComponentCVEEdge and NodeComponentCVEEdge have highlighted a gap in the SAC coverage on some join tables from the re-designed dackbox.
The goal with this PR is to generate SAC code at store code level for the join tables to ensure scope filtering happens as expected in postgres mode.

## Checklist

## Testing Performed

CI run should ensure the change does not introduce major bugs.
Further testing will come in other pull requests, such as https://github.com/stackrox/stackrox/pull/2987


janisz:
@rhybrillou I think you need to rebase

PR #2991, opened 2022-09-07
Add a visual indicator to dashboard option menus when the defaults have changed

dvail:
## Description

This adds a visual indication that default widget options have changed, as well as a reset button to restore the defaults.

Changes:

- The "Options" text in the dropdown has been changed to a standard "Cog" icon.
- The cog icon will be displayed as grey when the options are equal to their defaults, and blue when the default options have changed.
- A reset icon will appear next to the options dropdown that when clicked will restore the default options.

## Checklist

## Testing Performed

For each of the four widgets with options:
- Load the main dashboard and see that the cog is grey and the reset icon is not visible
- Change each individual option in the widget, observing that the cog turns blue and the reset icon appears (for the Violations by policy category widget, also show/hide non-default severities)
- Manually revert the options to the defaults, observing that the cog turns grey and the rest icon disappears
- Change an option to a non-default option and hover over the reset icon, displaying the icon's tooltip
- Click the reset icon to restore the widget options to the default
- Ensure that non-default options that are stored in local storage correctly display the cog and reset icon on page load

For the sake of brevity, screenshots for only a single widget are below.

Page load:
image

Default option changed:
image
image

Reset icon hover:
image

Clicking the reset icon:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
@@ -85,7 +85,7 @@ describe('AgingImages dashboard widget', () => {
             await screen.findByText(`${result0 + result1 + result2 + result3} Aging images`)
         ).toBeInTheDocument();
 
-        await user.click(await screen.findByText(`Options`));
+        await user.click(await screen.findByLabelText('Options'));
         const checkboxes = await screen.findAllByLabelText('Toggle image time range');
         expect(checkboxes).toHaveLength(4);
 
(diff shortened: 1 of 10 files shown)


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
```diff
@@ -147,12 +147,49 @@ describe('AgingImages dashboard widget', () => {
...
+        await user.dblClick(inputs[2]);
+        await user.type(inputs[2], '200', { skipClick: true });
+
+        expect(checkboxes[0]).not.toBeChecked();
+        expect(checkboxes[1]).not.toBeChecked();
+        expect(checkboxes[2]).toBeChecked();
+        expect(checkboxes[3]).toBeChecked();
+        expect(inputs.map(({ value }) => parseInt(value, 10))).toEqual(
+            expect.arrayContaining([30, 100, 200, 365])
+        );
```
vjwilson, line 186: Why is this expectation repeated here, and again below? It looks to me like these labels are always there; once we verify them the first time, we can just check the Checkboxes after changes.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
```diff
@@ -59,6 +61,11 @@ function ViolationsByPolicyCategory() {
     const query = getRequestQueryStringForSearchFilter(queryFilter);
     const { data: alertGroups, loading, error } = useAlertGroups(query, 'CATEGORY');
 
+    const isOptionsChanged =
+        lifecycle !== defaultConfig.lifecycle ||
+        sortType !== defaultConfig.sortType ||
+        xor(hiddenSeverities, defaultConfig.hiddenSeverities).length !== 0;
```
vjwilson, line 67: Use of `xor` is probably clever enough to warrant a code comment.

https://florian.github.io/xor-trick/

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/WidgetOptionsResetButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const buttonLabel = 'Revert to default options';
+
+function WidgetOptionsResetButton({ onClick }: WidgetOptionsResetButtonProps) {
+    return (
+        <Tooltip content={buttonLabel}>
+            <Button
+                aria-label={buttonLabel}
+                variant="plain"
+                icon={<RedoIcon />}
```
vjwilson, line 18: The "refresh" semantics of this icon concerns me. I searched for different alternative icons, but despite finding several discussions about a "restore defaults" icon, everyone's opinion seems to land on "there isn't a good icon for this."

What do you think about using text like "Reset" or "Restore default" ?

pedrottimark: Agree that refresh is not intuitive for reset.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/WidgetOptionsMenu.tsx
```diff
@@ -18,12 +19,18 @@ function WidgetOptionsMenu({ bodyContent }: OptionsMenuProps) {
...
                 variant="secondary"
                 className="pf-u-mr-sm"
                 icon={<CaretDownIcon />}
                 iconPosition="right"
             >
-                Options
+                <CogIcon
+                    className="pf-u-display-inline"
+                    color={
+                        isChanged ? 'var(--pf-global--link--Color)' : 'var(--pf-global--Color--200)'
```
pedrottimark, line 31: Thinking about pro and con of references to global colors brought a question to mind:
* Does gray suggest disabled?
* What do you think about leaving well enough alone with whatever is the solution to reset, instead of redundant conditional rendering. That leads to question whether cog even needs an explicit color.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ViolationsByPolicyCategory.tsx
dvail: Fair enough, I was hoping for a `Set.prototype.equals()` method but sadly it does not exist. I'll add a comment here.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/WidgetOptionsResetButton.tsx
dvail: Here's a screenshot of what it looks like with the change:

image

I'll run this as well as Mark's comments by UX to make sure everyone is on board and then I can push a quick update. In addition to the icon looking almost identical to the browser "refresh" icon, this would change the UI from a single word "Options", to two new icons. Having one icon button and one text button is likely a smoother transition.

dvail: There is also the undo icon which is the reverse of the screenshots above, and seems a little more semantically appropriate.

"Indicates the ability to undo a step in a historical log"

This isn't actually a "historical log", but it does undo the step of the user modifying the widget options.

image

pedrottimark (approved):
It looks clearer and simpler.

PR #2661, opened 2022-08-09
WIP: [DO NOT MERGE] Investiage risk reprocessing in SAC tests

dhaus67:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #2994, opened 2022-09-07
Add debug logs

janisz:
## Description

This PR add couple of debug log statements to narrow down issue with missing deployment.
We cannot reproduce it so we hope we get more logs if it fails on nightly or other PR.
Refs: https://github.com/stackrox/stackrox/pull/2661
## Testing Performed

CI

Refs:  ROX-12096

### DIFF
--- a/central/risk/manager/manager.go
+++ b/central/risk/manager/manager.go
@@ -19,6 +19,7 @@ import (
 	nodeScorer "github.com/stackrox/rox/central/risk/scorer/node"
 	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/debug"
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/sac"
 	"github.com/stackrox/rox/pkg/scancomponent"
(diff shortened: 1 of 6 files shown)


janisz:
* master
  * **PR #2994** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2994" target="_blank">Graphite</a>  👈
    * **PR #2995** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2995" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2994?utm_source=stack-comment).

central/risk/manager/manager.go
```diff
@@ -145,6 +148,7 @@ func (e *managerImpl) ReprocessDeploymentRisk(deployment *storage.Deployment) {
 	}
 
 	if oldScore == risk.GetScore() {
+		log.Debugf("Old riks equals existing risk, meaning we do not upsert for deployment %q", deployment.GetName())
```
gavin-stackrox, line 151: nit:
```suggestion
		log.Debugf("Old risk equals existing risk, meaning we do not upsert for deployment %q", deployment.GetName())
```

gavin-stackrox (commented):
It looks OK to me but I don't have any insight into potential performance impacts of debug statements in central so I'll leave it with you guys to approve.

dhaus67 (commented):
How about doing the same conditional log like before, so we only log the sac deployment? Otherwise the log will be quite big.

Additionally, let's wait for the first RC to be created before merging this one.

janisz:
> How about doing the same conditional log like before, so we only log the sac deployment? Otherwise the log will be quite big.

But then we will wait on repro only in this one test we spotted it. What if it happens in other tests?

janisz:
@dhaus67 @gavin-stackrox PTAL

janisz:
@dhaus67 Good point. Maybe we can enable logs for module

janisz:
@dhaus67 How about it now? Added filtering and keep everything in single function so it should be easy to remove.

janisz (commented):
/restest

dhaus67 (approved):

PR #2996, opened 2022-09-07
Clone values to prevent infinite loop and fix edge case in QuayIntegrationForm

pedrottimark:
## Description

1. Nakul found that QuayIntegrationForm can get stuck in an infinite loop of rendering.
    Scenario: Change only `oauthToken` value.
    Hypothesis: `if (initialValues) { … }` within the component looks like a side-effect (with shallow spread instead of deep clone) which I probably changed from naughty to nasty in #2974
    Solution:
    * Factor out pure `computeInitialValues` function and call it only one time (if needed) to initialize `useIntegrationForm` hook.
    * Replace `{ ...initialValues }` with `cloneDeep(initialValues)` because `updatePassword` logic did not work correctly without the change. This strengthens my feeling that the additional imperative assignment statements changed the code from brittle to broken.
    * Therefore also call `cloneDeep(defaultValues)` to create a new integration.

2. I found an edge case which needed improved frontend logic.
    Scenario: Edit an existing integration to clear Robot username and password causes `registryRobotCredentials: {username: "", password: ""}` to fail backend validation.
    Solution:
    * Add `onChangeRobotUsername` and `onChangeRobotPassword` to set `registryRobotCredentials: null` if both values are empty.

### Residue

1. Change from another type with robot account to **Scanner** has correct conditional rendering, but will fail backend validation. I will leave well enough alone for now.

## Checklist

## Testing Performed

1. Create new integration with **Scanner** type and **OAuth token** value. Save it without page crash

2. Create new integration with **Registry + Scanner** type and provide all credentials. Save it. Edit to clear **Robot username** field. Save it and see correct `registryRobotCredentials: null` instead of incorrect `registryRobotCredentials: {username: "", password: ""}` value.

### DIFF
--- a/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/QuayIntegrationForm.tsx
+++ b/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/QuayIntegrationForm.tsx
@@ -12,6 +12,7 @@ import {
     ToggleGroup,
     ToggleGroupItem,
 } from '@patternfly/react-core';
+import cloneDeep from 'lodash/cloneDeep';
 import * as yup from 'yup';
 
 import { QuayImageIntegration } from 'types/imageIntegration.proto';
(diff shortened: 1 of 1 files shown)


vjwilson (approved):
Oy vey!

PR #2997, opened 2022-09-07
Dashrews/ROX-12059-get available space from local postgres

dashrews78:
## Description

a lot going on here.  Took one of the approaches I found https://stackoverflow.com/questions/29806332/how-to-get-the-current-free-disk-space-in-postgres and implemented it.  Used go to parse the results figuring that would be simpler.  I also went a step beyond what this solution did in stackoverflow because the block size can differ even if using -P which is supposed to be portability mode.  For instance in the container -P got a block size of 1024.  On my intel Mac it got a block size of 512.  To to be safe I covered that.  Additionally if for whatever reason we get more results than we expect that would indicate that $PGDATA is not defined and we have no idea where the data is stored.  In that case I just use the same 100GB limit we started with.  Some of that is to get the conversation started in this PR.  While I think that position makes sense, I could be convinced of other approaches.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


c-du:
LOL, yes that is troublesome to have a different version for mac.

On my mac
     -P      Use (the default) 512-byte blocks.  This is only useful as a way to override an BLOCKSIZE specification from the
             environment.
     -k      Use 1024-byte (1-Kbyte) blocks, rather than the default.  Note that this overrides the BLOCKSIZE specification
             from the environment.
In a container, it is
  -P, --portability     use the POSIX output format
      --sync            invoke sync before getting usage info
      --total           elide all entries insignificant to available space,
                          and produce a grand total

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -265,9 +269,111 @@ func getPool(postgresConfig *pgxpool.Config) *pgxpool.Pool {
...
+	defer conn.Release()
+
+	// Start a transaction
+	tx, err := conn.Begin(ctx)
+	if err != nil {
+		log.Error(err)
+		return 0, err
+	}
+
+	// COPY needs some place to write the data.  This table will be deleted when the transaction ends.
```
c-du, line 296: Nice!

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -265,9 +269,111 @@ func getPool(postgresConfig *pgxpool.Config) *pgxpool.Pool {
...
+	if err != nil {
+		log.Errorf("Unable to create tmp table: %v", err)
+		if err := tx.Rollback(ctx); err != nil {
+			return 0, err
+		}
+		return 0, err
+	}
+
+	// COPY can execute a system level program and stream the results to a table.
+	_, err = tx.Exec(ctx, "COPY tmp_sys_df FROM PROGRAM 'df -P $PGDATA | tail'")
```
c-du, line 307: Is it better to use -k -P in this case?

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -278,12 +384,27 @@ func GetRemainingCapacity(postgresConfig *pgxpool.Config) (int64, error) {
...
+		capacity, err = getAvailablePostgresCapacity(postgresConfig)
+		if err != nil {
+			log.Error(err)
+			// If we cannot calculate the capacity, assume it based on the recommended starting
+			// capacity
+			return pgconfig.GetPostgresCapacity() - sizeUsed, err
+		}
+	}
+
+	log.Infof("remaining capacity = %d", capacity)
```
c-du, line 406: It is better to differentiate the results for managed central in log?

dashrews78:
> LOL, yes that is troublesome to have a different version for mac.
> 
> On my mac -P Use (the default) 512-byte blocks. This is only useful as a way to override an BLOCKSIZE specification from the environment. -k Use 1024-byte (1-Kbyte) blocks, rather than the default. Note that this overrides the BLOCKSIZE specification from the environment. In a container, it is -P, --portability use the POSIX output format --sync invoke sync before getting usage info --total elide all entries insignificant to available space, and produce a grand total

Yeah, it is 512 blocks if POSIXLY_CORRECT  is set and 1024 otherwise.

c-du (commented):
Looks good to me overall. Leave some comments to see if that would be helpful.

pkg/postgres/pgadmin/admin_utils.go
dashrews78: I'm not sure what you are meaning exactly.  But I wouldn't think that would be necessary in the log unless we want to put another caveat saying it is based off a value for now.  We will know that a managed central is using a managed Postgres at least initially.

c-du: I mean to put another (different) log within
`if env.ManagedCentral.BooleanSetting() {
}
`
so that we know if the capacity here is an estimate or not.

c-du (approved):

PR #2998, opened 2022-09-07
ROX-12400: Do not use digest of imageID when image names do not match

dhaus67:
## Description

Within detecting deployments, we use the `imageID` field to associate an image with a digest. This will help us potentially short-circuiting and using cached values for already existing images.

However, when using images that have been re-tagged, but are stored within different registries, it _may_ occur that the reference within the `imageID` field is for a different image.

As an example, when running two pods on the _same_ node with the following two images:
```bash
k run orig-img --image=docker.io/nginx:1.23

k run re-tagged-img --image=quay.io/dhaus/repro:1.23 # This is a retag of docker.io/nginx:1.23
```

You will observe the following within the output for kubectl describe:
```bash
k describe pod re-tagged-image
...
Containers:
  internal-image-signed:
    Container ID:   cri-o://3aaa27524519b64944721c91c89c73fd4c27eceb4efca435a5e47f7ab452cd67
    Image:          quay.io/dhaus/repro:1.23
    Image ID:       docker.io/nginx@sha256:89020cd33be2767f3f894484b8dd77bc2e5a1ccc864350b92c53262213257dfc
```

Although the repository digest _may_ be the valid for the quay.io/dhaus/repro repository, it is not guaranteed that it is.

Instead, we should skip setting the digest if the image name within the `imageID` field does not match with the image within the pod spec.

## Checklist

## Testing Performed
- added unit tests.

### DIFF
--- a/sensor/kubernetes/listener/resources/convert.go
+++ b/sensor/kubernetes/listener/resources/convert.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	openshift_appsv1 "github.com/openshift/api/apps/v1"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #2998** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/2998" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/2998?utm_source=stack-comment).

janisz:
Looks like this is not such a simple change as so many tests fails :(

PR #3001, opened 2022-09-07
Fix Mac ARM64 local build that broke due to multiarch

connorgorman:
## Description

See changes. the nogpgcheck is due to a package issue but because this is only used for local dev then it doesn't pose much of a risk

## Checklist

## Testing Performed

build locally + CI still needs to pass


connorgorman:
@misberner sanity check on the risks of using nogpgcheck?

image/rhel/Dockerfile.envsubst
```diff
@@ -43,7 +44,7 @@ RUN ln -s entrypoint-wrapper.sh /stackrox/admission-control && \
     ln -s entrypoint-wrapper.sh /stackrox/compliance && \
     ln -s entrypoint-wrapper.sh /stackrox/kubernetes-sensor && \
     ln -s entrypoint-wrapper.sh /stackrox/sensor-upgrader && \
-    ln -s /assets/downloads/cli/roxctl-linux-amd64 /stackrox/roxctl && \
+    ln -s /assets/downloads/cli/roxctl-linux-${TARGET_ARCH} /stackrox/roxctl && \
```
misberner, line 47: What's wrong with
```suggestion
    ln -s /assets/downloads/cli/roxctl-linux-$(uname -m) /stackrox/roxctl && \
```
? Why do you need an explicit build arg?

image/postgres/create-bundle.sh
```diff
@@ -30,8 +30,10 @@ chmod -R 755 "${bundle_root}"
 # =============================================================================
 # Get latest postgres minor version
 arch="x86_64"
+dnf_list_args=()
 if [[ $(uname -m) == "arm64" ]]; then
   arch="aarch64"
+  dnf_list_args=('--nogpgcheck')
```
misberner, line 36: Can you mention the `package issue` you refer to in the PR description in a comment here? Also: because this is only applied to `dnf list`, I would expect that this is relatively benign. When installing the downloaded RPM in the image, we will do a GPG check, no?

Makefile
```diff
@@ -69,7 +69,7 @@ BUILD_IMAGE := quay.io/stackrox-io/apollo-ci:$(shell sed 's/\s*\#.*//' BUILD_IMA
 ifeq ($(UNAME_S),Darwin)
 ifeq ($(UNAME_M),arm64)
 	# TODO(ROX-12064) build these images in the CI pipeline
-	BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-stackrox-build-0.3.44-arm64
+	BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-stackrox-build-0.3.46-arm64
```
tommartensen, line 72: Can you mention how to build this image (since it is not done with CI at the moment)? 🙂

tommartensen (commented):
General FYI, I have Mac M1 as well - `make image` was fixed with the changes provided in this PR.

image/rhel/Dockerfile.envsubst
connorgorman: Because uname -m ends up returning aarch64 when inside the Dockerfile :(

Makefile
connorgorman: 🪄  I build it on an ARM VM in GCP with some hacks to remove ossls

Makefile
```diff
@@ -357,6 +360,8 @@ endif
 
 .PHONY: cli
 cli: cli-build
+	# Workaround a bug on MacOS
+	rm $(GOPATH)/bin/roxctl
```
connorgorman, line 364:
```suggestion
	rm $(GOPATH)/bin/roxctl || true
```

misberner (approved):

image/rhel/Dockerfile.envsubst
misberner, line 47: While you're at it, you could also add
```suggestion
    ln -s /assets/downloads/cli/roxctl-linux-${TARGET_ARCH} /stackrox/roxctl && \
    ln -s /assets/downloads/cli/roxctl-linux-${TARGET_ARCH} /stackrox/roxctl-linux && \
```

to fix an issue `infra` is observing (see https://srox.slack.com/archives/CELUQKESC/p1662669354128059 for details).

I'm on the fence on whether it is a good idea to combine these two, but the change suggested above will have to be cherry-picked, and it seems fairly conflict-prone with your change so I'd rather keep it in one commit/PR. WDYT?

misberner (approved):

misberner:
@connorgorman If you decide to incorporate this change, please mark it with a release milestone as it should be fixed for the upcoming release.

image/rhel/Dockerfile.envsubst
connorgorman: Malte is addressing this in a separate PR then I will rebase as that will ensure a clean cherry pick onto release branch

msugakov (commented):
Non-blocking, cosmetic comments.
I just came here because I saw this is related to https://github.com/stackrox/stackrox/pull/3004

msugakov (commented):
BTW, it's worth to include `arch64` binaries in copying done in `copy-go-binaries-to-image-dir:` target.

Makefile
```diff
@@ -357,6 +360,8 @@ endif
 
 .PHONY: cli
 cli: cli-build
+	# Workaround a bug on MacOS
+	rm $(GOPATH)/bin/roxctl || true
```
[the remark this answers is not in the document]
connorgorman: yeah I'll do -f

connorgorman:
@msugakov 
> BTW, it's worth to include arch64 binaries in copying done in copy-go-binaries-to-image-dir: target.

Do you mean arm64? The problem is that inside the container it may say aarch64 as a result of uname -m whereas the valid GOARCH is arm64 so we're constantly distinguishing

msugakov:
Yes, these ways of calling ARM are confusing. I meant to suggest to put ARM binaries into the resulting image too.

connorgorman:
> I meant to suggest to put ARM binaries into the resulting image too.
Going to leave this to a follow up because this is out of scope of this particular change as it requires UI changes

image/postgres/create-bundle.sh
```diff
@@ -46,7 +49,7 @@ else
     docker build -q -t postgres-minor-image "${build_dir}" -f - <<EOF
 FROM registry.access.redhat.com/ubi8/ubi:${pg_rhel_version}
 RUN dnf install --disablerepo='*' -y "${postgres_repo_url}"
-ENTRYPOINT dnf list --disablerepo='*' --enablerepo=pgdg${postgres_major} -y postgresql${postgres_major}-server.$arch | tail -n 1 | awk '{print \$2}'
+ENTRYPOINT dnf list "${dnf_list_args[@]-}" --disablerepo='*' --enablerepo=pgdg${postgres_major} -y postgresql${postgres_major}-server.$arch | tail -n 1 | awk '{print \$2}'
```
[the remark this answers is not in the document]
connorgorman: Yeah, I'll do this

connorgorman:
I am going to merge this one shortly. Feel free to ping me once https://github.com/stackrox/stackrox/pull/3004 is completely done so I can test on my M1

PR #3004, opened 2022-09-07
Multi-arch non-CI roxctl build fix

yselkowitz:
/cc @msugakov 

## Checklist

## Testing Performed

Built locally with `CI` undefined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -429,7 +429,7 @@ main-build-nodeps:
 	$(GOBUILD) central migrator sensor/kubernetes sensor/admission-control compliance/collection
 	CGO_ENABLED=0 $(GOBUILD) sensor/upgrader
 ifndef CI
-    CGO_ENABLED=0 $(GOBUILD) roxctl
+	CGO_ENABLED=0 $(GOBUILD) roxctl
 endif
 
 .PHONY: scale-build
@@ -595,9 +595,9 @@ ifdef CI
 	cp bin/windows_amd64/roxctl.exe image/bin/roxctl-windows-amd64.exe
 else
 ifneq ($(HOST_OS),linux)
-	cp bin/linux_$(GOARCH)/roxctl image/bin/roxctl-linux-$(GOARCH)
+	cp bin/linux_amd64/roxctl image/bin/roxctl-linux-amd64
 endif
-	cp bin/$(HOST_OS)_amd64/roxctl image/bin/roxctl-$(HOST_OS)-amd64
+	cp bin/$(HOST_OS)_$(GOARCH)/roxctl image/bin/roxctl-$(HOST_OS)-$(GOARCH)
 endif
 	cp bin/linux_$(GOARCH)/migrator image/bin/migrator
 	cp bin/linux_$(GOARCH)/kubernetes        image/bin/kubernetes-sensor


Makefile
```diff
@@ -595,9 +595,9 @@ ifdef CI
 	cp bin/windows_amd64/roxctl.exe image/bin/roxctl-windows-amd64.exe
 else
 ifneq ($(HOST_OS),linux)
-	cp bin/linux_$(GOARCH)/roxctl image/bin/roxctl-linux-$(GOARCH)
+	cp bin/linux_amd64/roxctl image/bin/roxctl-linux-amd64
 endif
-	cp bin/$(HOST_OS)_amd64/roxctl image/bin/roxctl-$(HOST_OS)-amd64
+	cp bin/$(HOST_OS)_$(GOARCH)/roxctl image/bin/roxctl-$(HOST_OS)-$(GOARCH)
 endif
```
msugakov, line 601: [I'm nitpicking, but still]
I think there are too many `if`-s here. The above thing (lines 590-596) copies all possible (for now) combinations of roxctl to be included in the image, in `CI` context.
This thing here (lines 597-601) copies linux-amd64 for all non-linux hosts and and linux-$(GOARCH) for linux hosts.

The current code of `cli-build` unconditionally builds everything for all platforms except of windows.
https://github.com/stackrox/stackrox/blob/53dba5c9e4737f7cd64fa86341eaa301a630828c/Makefile#L348-L356

Therefore, I think the code of `copy-go-binaries-to-image-dir` should simply follow that:

```Makefile
copy-go-binaries-to-image-dir:
	cp bin/linux_$(GOARCH)/central image/bin/central

	cp bin/linux_amd64/roxctl image/bin/roxctl-linux-amd64
	cp bin/linux_ppc64le/roxctl image/bin/roxctl-linux-ppc64le
	cp bin/linux_s390x/roxctl image/bin/roxctl-linux-s390x
	cp bin/darwin_amd64/roxctl image/bin/roxctl-darwin-amd64
ifdef CI
	cp bin/windows_amd64/roxctl.exe image/bin/roxctl-windows-amd64.exe
else

	cp bin/linux_$(GOARCH)/migrator image/bin/migrator
	# and everything else
```

WDYT, does this make sense?

misberner: I'm not sure I read your snippet correctly, why is there still an `ifdef CI` and why is there an unterminated `else`? I'm also not sure it is super well justified to exclude windows from non-CI image builds.

I think there are two options:
1. Make local builds match CI builds, i.e., include `roxctl` for all platforms (including windows OSes).
2. Only include `roxctl` for the target arch in the image. I think that target arch is a concept we need to introduce at the Makefile level; it's what `uname -m` will show in the built container. On M1 Mac, target platform is still `linux/amd64`, but it might become `linux/arm64` in the future. Given this upcoming change, we should not reference the concrete `amd64` platform anywhere in new Makefile instructions, but rather `$(TARGET_ARCH)`.

The advantage of (1) is obviously a closer match between CI builds and local builds - is this relevant, though? The downside is that build times are even longer for local images, due to binaries we will never need (compiling `roxctl` for the host platform is relevant for the local deployment workflow, but this host `roxctl` still doesn't need to be part of the image).

I'm in favor of 2. I expect that nobody will download a roxctl binary from a locally built image

msugakov: > I'm not sure I read your snippet correctly, why is there still an ifdef CI and why is there an unterminated else? I'm also not sure it is super well justified to exclude windows from non-CI image builds.

The only part of this feedback I accept is about unterminated `else`. It was meant to be `endif` but I did not pay attention.
Why do I dismiss everything else in your first paragraph? Because what I suggested follows the logic of `cli-build` which I did not invent. What's the justification to exclude windows there? Well, that may be a fun exercise finding out.

msugakov: Question: what do we want Yaakov to do here and what do we take to do ourselves?

misberner: I don't think following logic that we have identified as questionable has any value, at least not for the sense of consistency. Keeping the scope of this PR small because we don't want to burden Yaakov with fixing our build stuff is fair, but imho that means: if the existing PR is sufficient, let's approve it and get it merged. If it breaks something, like M1 Mac builds, let's fix that. But *if* we're opening up a discussion on logic, consistency and what "should" be done, let's discuss based on what actually makes sense for us rather than adequate PR scope.

Yes, I forgot to add the "for local builds" qualifier in point 2, apologies. In my mind it followed from (a) the subject of the work here being non-CI builds, and (b) 1 and 2 being alternatives implying that 2 requires a discrepancy between the CI and local builds. But I could have been clearer about that.

My point also was that the download feature is already broken because we don't include the windows binary in the image. So obviously we are okay with it being broken on local builds.

msugakov: > if the existing PR is sufficient, let's approve it and get it merged. If it breaks something [...], let's fix that.

It would be great to discuss this specific part of the suggestion as we are still evolving the way we work with open source contributions. (Thanks Yaakov for your patience here.)
We have something that's visually and logically entangled with a longer-term technical debt sitting below it.

Should we be willing to have open source contributors commit things that don't pass reviewer comments that would appear if the contributor was in the team? If yes, what's the reason to dismiss comments and lower the bar for OSS? If not, how deep do we want OSS contributors to go addressing tech debt?

yselkowitz: In the meantime, and in line with my eventual answer to your question, I recommend proceeding with this patch as is unless there is a directly related issue with it.

msugakov: There are recent and related PRs regarding `roxctl` https://github.com/stackrox/infra/pull/696 and https://github.com/stackrox/stackrox/pull/3001 . At this point I'm reading and trying to understand what the Makefile is supposed to do w.r.t. `roxctl` build, e.g. why we build it multiple times and where does it go after.

msugakov, line 601: Okay, I think this entire thing (lines 596-601) should be simply:

```Makefile
else
	cp bin/linux_$(GOARCH)/roxctl image/bin/roxctl-linux-$(GOARCH)
endif
```

Am I wrong?

msugakov: Please check https://github.com/stackrox/stackrox/pull/3051

misberner: I would agree, yes

yselkowitz:
Closing in favour of #3051

PR #3011, opened 2022-09-08
Dashrews/ROX-12538-flows use union2

dashrews78:
## Description

Deployments can exist in the src portion of the flow OR the destination portion of the flow.  When processing baselines we need to grab the flows that may relate to that deployment.  So by definition we need the query of the flows to function as an or.  It was written as an or.  That is not being particularly performant because it is causing full table scans.  The first optimization trick to deal with ors is to change it to a union.  Query for the deployment in the source and union that with a query for the deployment in the destination.  This will utilize the indexes and cut down on the full table scans as well as the amount of data that is pulled in to do the aggregation to grab the latest flows that match.

## Checklist

## Testing Performed

Added a simple unit test.


dashrews78:
ROX-12538

c-du (approved):

PR #3012, opened 2022-09-08
Improve operator build QoL

misberner:
## Description

Minor improvements:
- Remove obsolete `GOBIN` setting (one less unconditional shell invocation).
- Make all tools explicitly versioned. This ensures that after an upgrade, one does not accidentally keep using the previously installed tool version.
- Use `-x` instead of `-f` checks to ensure tools can be executed, and make `chmod +x` the last operation when downloading a tool from GitHub. Combined these two ensures that an interrupted/failed `curl` download does not leave behind a state that makes the `Makefile` assume the tool is installed.

## Checklist

## Testing Performed
- Ran targets manually

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -100,13 +100,6 @@ ENABLE_WEBHOOKS ?= false
 # Use the value figured in the parent directory Makefile, unless provided explicitly in the environment.
 ROX_IMAGE_FLAVOR ?= $(shell $(MAKE) --quiet --no-print-directory -C .. image-flavor)
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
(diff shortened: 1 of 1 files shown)


msugakov (approved):

porridge (approved):
LGTM as long as the tests pass this time. Previously failed with:
```
    logger.go:42: 14:02:18 | upgrade/20-upgrade-operator | 2022-09-08T14:02:18Z Target image tag has -dirty suffix.
    logger.go:42: 14:02:18 | upgrade/20-upgrade-operator | 2022-09-08T14:02:18Z Cannot install from *-dirty image tag. Please, use 'deploy-dirty-tag-via-olm' command or add '--allow-dirty-tag' flag if you need to install dirty tagged image.
```

misberner:
The failure was due to not having rebased after a submodule (docs) change. This makes OSCI fail, see https://srox.slack.com/archives/C02JG23A93L/p1663018953775469 for an explanation

PR #3015, opened 2022-09-08
ROX-12512: Use operator framework version without memory leak

misberner:
## Description

This uses the `helm-operator` version where the memory leak issue is fixed. See https://github.com/stackrox/helm-operator/pull/30 for more details.

I'd like to decouple reducing the operator container resources from this, as it will require some more detailed investigation/load testing to see how bad the usage can become in a busy cluster with several CRDs, and this PR on its own represents a sufficient improvement.

## Checklist

## Testing Performed

- CI (tbd)


connorgorman (approved):
full send

PR #3019, opened 2022-09-08
External configmap for Central DB to allow for easier configuration

connorgorman:
## Description

Externalize the config map to allow customers to change it more easily. Also, bump default resources as they were a bit low

## Checklist

## Testing Performed

CI


janisz (approved):

c-du (approved):

dashrews78 (approved):

image/templates/helm/stackrox-central/templates/01-central-08-db-configmap.yaml
```diff
@@ -0,0 +1,939 @@
...
+metadata:
+  name: central-db-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "srox.labels" (list . "configmap" "central-db-config") | nindent 4 }}
+  annotations:
+    {{- include "srox.annotations" (list . "configmap" "central-db-config") | nindent 4 }}
+data:
+  postgresql.conf: |
+    {{- if ._rox.central.enableCentralDB }}
```
misberner, line 14: I'm pretty sure this is misplaced, because you only `{{- end }}` it after `pg_hba.conf`, and I don't see why one file would be conditional, whereas the other one would exist unconditionally (but possibly be empty).

I think what you want here is to make the creation of the entire `ConfigMap` resource conditional, no?

image/templates/helm/stackrox-central/templates/01-central-08-db-configmap.yaml
```diff
@@ -0,0 +1,939 @@
...
+    # -----------------------------
```
misberner, line 15: Rather than putting the entire config inline in a YAML file, put the files in `config/central-db/postgresql.conf` and `config/central-db/pg_hba.conf`, and include them here via, e.g.,
```
data:
  postgresql.conf: |
    {{- .Files.Get "config/central-db/postgresql.conf" | nindent 4 }}
```

misberner (changes requested):

image/templates/helm/stackrox-central/templates/01-central-08-db-configmap.yaml
connorgorman: correct

image/templates/helm/stackrox-central/templates/01-central-08-db-configmap.yaml
connorgorman: will do

connorgorman: Does this preclude that postgresql.conf from being templated?

connorgorman:
Actually, let's just hold on this one because I'm going to rework Central-DB helm almost entirely. I think it should be a top level object and not mixed with Central

connorgorman:
Replacing this PR with https://github.com/stackrox/stackrox/pull/3038

PR #3022, opened 2022-09-08
Create legacy symlink for roxctl

misberner:
## Description

With a recent change in preparation of multiarch, `roxctl` filenames in the `/assets/downloads/cli` directory were made architecture-specific (i.e., `roxctl-linux-amd64` instead of `roxctl-linux`).

Since apparently [some automation](https://srox.slack.com/archives/CELUQKESC/p1662669354128059) depends on the presence of `/assets/downloads/cli/roxctl-linux`, create a symlink for backwards compatibility.

## Checklist

## Testing Performed

- [ ] will check out that the old path works in the image once it's built


misberner:
This will probably have to be replicated downstream.

misberner:
Downstream MR: https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/171

connorgorman (approved):
just check that go file server will serve this through symlink (it should)

PR #3024, opened 2022-09-09
ROX-12182: Do not enable OpenShift Auth provider for managed centrals

dhaus67:
## Description

Within the managed service context, the instances will run on a OpenShift Dedicated Cluster (OSD). This cluster will be, as the name suggests, use Openshift.
Specifically, OpenShift > 4.X.

Since the operator is auto-sensing during installation and setting the variable `env.openshift=4`, this means the OpenShift Auth provider is enabled.

This means that customers actually have the possibility to create an OpenShift Auth provider using the cluster their instance is hosted upon.
We definitely do not want to give this possibility to the customer and disclose such information.

We already have the environment variable `ROX_MANAGED_CENTRAL` highlighting whether we are running in "managed central" mode. 

Within the helm chart, a new setting `env.managedServices` has been added. If this variable is set, the Openshift auth provider will not be enabled (irrespective of whether the `env.openshift` variable is set).
The operator will inject this variable if the annotation `platform.stackrox.io/managed-services` is set to true.

What's left to do is to adjust the central creation in fleetshard synchronize to include this annotation, making sure the Openshift auth provider is disabled and `ROX_MANAGED_CENTRAL` is set.

## Checklist

## Testing Performed

- see CI, added unit tests for the Helm chart itself.
- manual test:
  - create a central CR with annotation.
  - verify ROX_MANAGED_CENTRAL is set and Openshift auth provider is disabled.

### DIFF
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -341,6 +341,11 @@ function launch_central {
         ${ORCH_CMD} -n stackrox set env deploy/central MODULE_LOGLEVELS="${MODULE_LOGLEVELS}"
       fi
 
+      if [[ "$ROX_MANAGED_CENTRAL" == "true" ]]; then
+        echo "ROX_MANAGED_CENTRAL=true is only supported in conjunction with OUTPUT_FORMAT=helm"
+        exit 1
+      fi
+
       if [[ "$SCANNER_SUPPORT" == "true" ]]; then
           echo "Deploying Scanner..."
           if [[ -n "${REGISTRY_USERNAME}" ]]; then
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -304,7 +304,7 @@ function launch_central {
 
       if [[ "$ROX_MANAGED_CENTRAL" == "true" ]]; then
         helm_args+=(
-          --set customize.central.envVars.ROX_MANAGED_CENTRAL="${ROX_MANAGED_CENTRAL}"
+          --set env.managedServices=true
         )
       fi
 
@@ -339,10 +339,6 @@ function launch_central {
       fi
       if [[ -n $MODULE_LOGLEVELS ]]; then
         ${ORCH_CMD} -n stackrox set env deploy/central MODULE_LOGLEVELS="${MODULE_LOGLEVELS}"
-      fi
-
-      if [[ "$ROX_MANAGED_CENTRAL" == "true" ]]; then
-        ${ORCH_CMD} -n stackrox set env deploy/central ROX_MANAGED_CENTRAL="${ROX_MANAGED_CENTRAL}"
       fi
 
       if [[ "$SCANNER_SUPPORT" == "true" ]]; then
--- a/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
@@ -8,11 +8,11 @@ metadata:
   labels:
     {{- include "srox.labels" (list . "serviceaccount" "central") | nindent 4 }}
   annotations:
+    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
     {{- if and (eq ._rox.env.openshift 4) (not ._rox.env.managedServices) }}
     serviceaccounts.openshift.io/oauth-redirectreference.main: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"central"}}'
     serviceaccounts.openshift.io/oauth-redirecturi.main: "sso/providers/openshift/callback"
     {{- end }}
-    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
 imagePullSecrets:
 {{- range $secretName := ._rox.imagePullSecrets._names }}
 - name: {{ quote $secretName }}
--- a/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
@@ -8,7 +8,7 @@ metadata:
   labels:
     {{- include "srox.labels" (list . "serviceaccount" "central") | nindent 4 }}
   annotations:
-    {{- if and (eq ._rox.env.openshift 4) (not (hasKey ._rox.customize.central.envVars "ROX_MANAGED_CENTRAL" )) }}
+    {{- if and (eq ._rox.env.openshift 4) (not ._rox.env.managedServices) }}
     serviceaccounts.openshift.io/oauth-redirectreference.main: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"central"}}'
     serviceaccounts.openshift.io/oauth-redirecturi.main: "sso/providers/openshift/callback"
     {{- end }}
--- a/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
@@ -8,9 +8,11 @@ metadata:
   labels:
     {{- include "srox.labels" (list . "serviceaccount" "central") | nindent 4 }}
   annotations:
-    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
+    {{- if and (eq ._rox.env.openshift 4) (not (hasKey ._rox.customize.central.envVars "ROX_MANAGED_CENTRAL" )) }}
     serviceaccounts.openshift.io/oauth-redirectreference.main: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"central"}}'
     serviceaccounts.openshift.io/oauth-redirecturi.main: "sso/providers/openshift/callback"
+    {{- end }}
+    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
 imagePullSecrets:
 {{- range $secretName := ._rox.imagePullSecrets._names }}
 - name: {{ quote $secretName }}
--- a/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
@@ -100,7 +100,7 @@ spec:
           value: {{ ._rox.central.disableTelemetry | not | quote }}
         - name: ROX_OFFLINE_MODE
           value: {{ ._rox.env.offlineMode | quote }}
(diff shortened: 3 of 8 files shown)


dhaus67:
* master
  * **PR #3024** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3024" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3024?utm_source=stack-comment).

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -100,7 +100,7 @@ spec:
           value: {{ ._rox.central.disableTelemetry | not | quote }}
         - name: ROX_OFFLINE_MODE
           value: {{ ._rox.env.offlineMode | quote }}
-        {{- if eq ._rox.env.openshift 4 }}
+        {{- if and (eq ._rox.env.openshift 4) (not (hasKey ._rox.customize.central.envVars "ROX_MANAGED_CENTRAL" )) }}
```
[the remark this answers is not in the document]
dhaus67: >How do we know that ROX_MANAGED_CENTRAL is always already set here?

I'm not sure I understand your concern here. The `rox.customize.central.envVars` is supposed to be set within the `values.yaml` of the Helm Chart. At this point, of rendering the template, we can be sure that its either set / or not set, due to the `values.yaml` being used as input for rendering. So I'm not sure about your concern.

I get that it's weird that we set a env var based on the value of another env var, but what would be the alternative? We don't have anything besides `ROX_MANAGED_CENTRAL` that signals whether we are running in managed service context or not. Also, we already do set the env variable based on another env variable: `env.openshift=4`.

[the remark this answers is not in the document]
ivan-degtiarenko: But why do we need to set `ROX_ENABLE_OPENSHIFT_AUTH` to `false`, if there is already this check:

```
if env.EnableOpenShiftAuth.BooleanSetting() && !env.ManagedCentral.BooleanSetting()
```

I think I would prefer enforcement on the code level rather than the deployment level. That way it will work for all deployment methods.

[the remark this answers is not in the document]
dhaus67: Yeah, I can remove it for the deployment template, but we will have to keep it within the service account template.

image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
```diff
@@ -8,9 +8,11 @@ metadata:
   labels:
     {{- include "srox.labels" (list . "serviceaccount" "central") | nindent 4 }}
   annotations:
-    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
+    {{- if and (eq ._rox.env.openshift 4) (not (hasKey ._rox.customize.central.envVars "ROX_MANAGED_CENTRAL" )) }}
```
misberner, line 11: I don't think this is the right approach. I know why you've chosen it instead of adding a control knob at the level of Helm values, or rather, I can imagine two reasons:
- because you don't want to expose managed services mode as an "official" setting to customers
- because we are limited by which values can be set through the operator, and we certainly don't want to add a `managedServices` field to our Central CRD.

Imho, the first reason is a non-reason. Nobody says such a Helm chart value would need to be documented, and I don't think an undocumented value that is checked somewhere in the template via `{{- if ._rox._managedService }}` or so is any more exposed than the "magical" env var which is also referenced here in the template.

The second reason is understandable, but the implementation here won't work. As you can see in https://github.com/stackrox/stackrox/blob/master/operator/pkg/central/values/translation/translation.go#L74 in conjunction with https://github.com/stackrox/stackrox/blob/fa4592ee07ad8be1f8664edb66c87bb7324c74c3/operator/pkg/values/translation/translation.go#L37 , the operator only supports setting env vars _globally_, for all components. That is fine for the part inside central which reads the setting via `os.Getenv`, but it means that in the Helm chart rendering done by the operator, with values set from the translation logic linked above, the setting here is actually called `._rox.customize.envVars`.

Changing `._rox.customize.central.envVars` to `._rox.customize.envVars` is certainly a quick fix that should get the job done, and it might be the right choice depending on the time-sensitivity. I would argue that it's not the right approach because it's a leaky abstraction - we write the Helm chart based on some detail knowledge about how the operator implementation works. Meh.

One idea I have off the top of my head on how to address this issue is to read annotations in the operator. This could be either really specific, such as changing the translator logic to inject `env._managedServices = true` whenever it detects the presence of an `platform.stackrox.io/managed-services: true` annotation on the Central CR, or more general, via an annotation such as `platform.stackrox.io/inject-helm-values: '{"env": {"_managedServices": true}}'`.

I can see that this might seem like a bit of overengineering for the problem at hand, but I think it's not at all unlikely that we will see ourselves wanting to adjust the chart templates for managed services, beyond environment variables and beyond the knobs that the operator supports. But I want to hear your thoughts on that!

misberner (changes requested):

image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
dhaus67: I'd say that captures my thought process pretty much :-)
Thanks for the hints regarding the `GetCustomize` details!

I don't think the annotation way is over-engineered per-se, looking at other operators like the infamous ingress controllers this seems a typical solution to address this and it also seems better in the long-rung than doing the quick-way of using `._rox.customize.envVars`. 

I chose the `platform.stackrox.io/managed-services` annotation for that and the new helm setting `env.managedServices`. Now, we set `ROX_ENABLE_OPENSHIFT_AUTH / ROX_MANAGED_CENTRAL` only if the new setting is set.

deploy/common/k8sbased.sh
```diff
@@ -341,10 +341,6 @@ function launch_central {
         ${ORCH_CMD} -n stackrox set env deploy/central MODULE_LOGLEVELS="${MODULE_LOGLEVELS}"
       fi
 
-      if [[ "$ROX_MANAGED_CENTRAL" == "true" ]]; then
```
misberner, line 344: Do you just want to error out here, and indicate that deploying with `ROX_MANAGED_CENTRAL=true` is only supported when choosing the Helm route (`OUTPUT_FORMAT=helm`)?

image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
```diff
@@ -8,9 +8,11 @@ metadata:
   labels:
     {{- include "srox.labels" (list . "serviceaccount" "central") | nindent 4 }}
   annotations:
-    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
+    {{- if and (eq ._rox.env.openshift 4) (not ._rox.env.managedServices) }}
     serviceaccounts.openshift.io/oauth-redirectreference.main: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"central"}}'
     serviceaccounts.openshift.io/oauth-redirecturi.main: "sso/providers/openshift/callback"
+    {{- end }}
+    {{- include "srox.annotations" (list . "serviceaccount" "central") | nindent 4 }}
```
misberner, line 15: Is there a particular reason you moved this to the bottom of the annotations list?

operator/pkg/central/values/translation/translation.go
```diff
@@ -92,6 +101,19 @@ func getEnv(egress *platform.Egress) *translation.ValuesBuilder {
 			}
 		}
 	}
+
+	if annotations != nil {
+		if annotation, ok := annotations[managedServicesAnnotation]; ok {
+			managedServices, err := strconv.ParseBool(annotation)
+			if err != nil {
+				return env.SetError(fmt.Errorf("invalid annotation.%s %q", managedServicesAnnotation, annotation))
```
misberner, line 109:
```suggestion
				return env.SetError(fmt.Errorf("invalid annotation value %q for %s", annotationValue, managedServicesAnnotation))
```

operator/pkg/central/values/translation/translation.go
```diff
@@ -92,6 +101,19 @@ func getEnv(egress *platform.Egress) *translation.ValuesBuilder {
 			}
 		}
 	}
+
+	if annotations != nil {
+		if annotation, ok := annotations[managedServicesAnnotation]; ok {
```
misberner, line 106:
```suggestion
		if annotationValue, ok := annotations[managedServicesAnnotation]; ok {
```

misberner (approved):

image/templates/helm/stackrox-central/templates/01-central-00-serviceaccount.yaml
dhaus67: No reason at all, reverted it.

SimonBaeumer (approved):
Generally speaking I am not a fan on having configurations labeled as "Managed Services" because StackRox should not be aware in which exact product it is deployed, especially not as an opensource product which may be contributed to the CNCF.

With that being said I am not vetoing against it, but want to raise awareness that introducing overly general global flags is a pattern we might want to avoid long-term.

misberner:
@SimonBaeumer , I fully agree. However, there is some tension here between two competing objectives: one, to not expose bespoke configuration options like "managed service" that are only relevant for a single user (us), and two, to not expose knobs for every tiny setting, as that would cause the configuration surface to grow significantly. Because if we take what you say to heart, we would need an individual setting for everything that's different in managed services, and then apply a "managed services" config at fleetshard deploy time.

In the concrete example, adding a knob for enabling/disabling OpenShift auth seems fairly unnatural, because at the Helm chart you merely configure the _potential_ to use OpenShift auth. An administrator still needs to configure OpenShift auth in-product, and so it is unlikely that anyone would want anything other than auto-sense at the _Helm_ level. Of course we could add a setting for this and not document it, as I suggested for the managed services setting, but that then kind of defeats the point you're making.

SimonBaeumer:
> In the concrete example, adding a knob for enabling/disabling OpenShift auth seems fairly unnatural, because at the Helm chart you merely configure the _potential_ to use OpenShift auth. An administrator still needs to configure OpenShift auth in-product, and so it is unlikely that anyone would want anything other than auto-sense at the _Helm_ level. Of course we could add a setting for this and not document it, as I suggested for the managed services setting, but that then kind of defeats the point you're making.

I agree with you for this use-case, adding so many knobs would be too much. What I am observing is that a configuring the authprovider from a Helm chart / operator would make this not necessary in the first place.
So in this case, Central deployed to OpenShift the OpenShift auth provider is enabled by default, in a managed service environment a custom auth provider (RHSSO) is configured and the default AuthProvider is disabled.
Imho not having a declarative auth provider has effect on which and how configurations are exposed already.
I should have been more clear on what I am referring to.
With that being said in our current state introducing managed services flag is reasonable to me.

PR #3031, opened 2022-09-09
Update compliance widget test to be more robust against larger data input

dvail:
## Description

This updates the compliance dashboard widget unit test to better verify behavior.

The previous test only verified that the bars were in ascending or descending order, _not_ that the data in the bars was correct in regards to the server response. (Additionally, the previous check tested `>` and `<` instead of the more accurate `>=` and `<=`).

This new test mocks a data response with >6 results and tests that the exact 6 highest or lowest are displayed in the correct order.

## Checklist

## Testing Performed

`yarn test`

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { MockedProvider } from '@apollo/client/testing';
-import { screen, within, waitFor } from '@testing-library/react';
+import { screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom/extend-expect';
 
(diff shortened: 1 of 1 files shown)


ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
```diff
@@ -108,36 +95,45 @@ describe('Compliance levels by standard dashboard widget', () => {
...
 
-        async function getBarPercentages() {
-            // eslint-disable-next-line testing-library/no-node-access
-            const svgBarsElement = document.querySelector('svg > g:nth-of-type(3)');
-            const barPercentages = await within(svgBarsElement as HTMLElement).findAllByText(
-                /\d+%/
-            );
-            expect(barPercentages).toHaveLength(6);
+        async function getBarTitles() {
+            const titlesRegex = new RegExp(`${Object.values(standardNames).join('|')}`);
```
pedrottimark, line 99: If I understand correctly, find all possible standard names here, and verify below only the expected subset?

Is there an HTML element or accessibility attribute to search for titles without a pre-expectation of their content?

pedrottimark (approved):
2 comments for your consideration.

Nice to see how you reduced the number of imports from testing-library.

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
dvail: Unfortunately there isn't an accessibility attribute to grab onto, or another RTL utility that can be used to query these elements, since the HTML is SVG code generated by the Victory library:

Roughly
```
<g role="presentation">
  <line vector-effect="non-scaling-stroke" role="presentation" shape-rendering="auto" style="..."></line>
  <a href="...">
    <text direction="inherit" dx="0" x="135" y="214.97" id="chart-axis-0-tickLabels-0">
      <tspan x="135" dx="0" dy="0" text-anchor="end" style="...">CIS Kubernetes v1.5</tspan>
    </text>
  </a>
</g>
```

We could grab onto the `role="presentation"` attribute, but that would also pick up the other elements of the chart and we would still need to filter out the percentages, etc.

We could query directly by DOM structure, but I figured that would be more subject to breakages outside of our control than using the name constants defined earlier in the test file.

pedrottimark: Yeah, better to leave well enough alone. A lesson for the future when you contribute to an open source library!

ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/ComplianceLevelsByStandard.test.tsx
```diff
@@ -13,6 +13,34 @@ import ComplianceLevelsByStandard from './ComplianceLevelsByStandard';
 
 jest.setTimeout(10000);
 
+const standards = [
+    /*
+    These standards have been formatted for easier verification of the expected ordering in
+    tests compared to a direct hard coding in the mocked response below.
+
+    [numFailing, numPassing, id]
```
pedrottimark, line 21: What do you think about an order of items so reader even more quickly sees that `standards` `standardNames` have parallel structure?
* Replace `[numFailing, numPassing, id]` with `[id, numFailing, numPassing]` here.
* Replace `[, , id]` with `[id]` below.

pedrottimark: Optional extra credit: declare that strings in `standards` are keys in `standardNames` object?

dvail: The `standards` const existed before the `standardNames`, with the goal of listing them by `numFailures` to easily tell what the intended ordering would be. (Although having them intentionally out of order would be a better test of the client side sorting...)

To your point though, it might be better to just combine the two now into something like:
```typescript
// id: [name, numFailing, numPassing]
const standards = {
    CIS_Docker_v1_2_0: ['CIS Docker v1.2.0', 9, 1],
    ...
};
```
and iterate with `Object.entries` when needed.

PR #3037, opened 2022-09-09
ROX-10986-ROX-11033: Prevent checkbox crashes on Image Overview CVEs tables

vjwilson:
## Description

* Considered refactoring useTableSelection hook, and its usage, but rejected as too risky at this time.
* Created [draft brute-force fix](https://github.com/stackrox/stackrox/pull/2938) of clearing selections before every action that changed the tables in question
* This PR's simpler solution: check and short-circuit where crash occurred before during component update cycle (suggested by @dvail in [comment to draft PR](https://github.com/stackrox/stackrox/pull/2938#pullrequestreview-1099250127))

## Checklist

## Testing Performed

Existing crash when changing view of table (paging, filtering) after selecting one or more checkboxes.
Screen Shot 2022-09-09 at 3 12 20 PM

After this code change,
select one or more checkboxes
Screen Shot 2022-09-09 at 3 12 43 PM

then page or filter--and page behaves as expected (does not crash)
Screen Shot 2022-09-09 at 3 12 48 PM

### DIFF
--- a/ui/apps/platform/src/hooks/useTableSelection.ts
+++ b/ui/apps/platform/src/hooks/useTableSelection.ts
@@ -55,7 +55,7 @@ function useTableSelection<T extends Base>(
     function getSelectedIds() {
         const ids: string[] = [];
         for (let i = 0; i < selected.length; i += 1) {
-            if (selected[i]) {
+            if (selected[i] && data[i]?.id) {
                 ids.push(data[i].id);
             }
         }


dvail (approved):

ui/apps/platform/src/hooks/useTableSelection.ts
```diff
@@ -55,7 +55,7 @@ function useTableSelection<T extends Base>(
     function getSelectedIds() {
         const ids: string[] = [];
         for (let i = 0; i < selected.length; i += 1) {
-            if (selected[i]) {
+            if (selected[i] && data[i]?.id) {
```
sachaudh, line 58: This is a good fix for the immediate issue. I looked at the hook again and was considering what causes this issue and seems like its mainly because we do something like this:
```
const { ..., getSelectedIds, ... } = useTableSelection(...);
...
const selectedIds = getSelectedIds();
```
Seems like the `getSelectedIds` function is referencing the wrong `data`. I think it might be better to actually do the `const selectedIds = getSelectedIds();` part within the hook. We would call that function on every render anyways. At least we can prevent this closure issue from happening if we pull it into the hook. The API for that hook would then look something like this:
```
const { ..., getSelectedIds, selectedIds ... } = useTableSelection(...);
...
```
Now the hook will return an updated `selectedIds`, that uses the proper `data` to calculate the selected ids, whenever `data` is updated

sachaudh: You don't have to refactor it now, but maybe a `@TODO` for fixing the hook's API in the future would be good to add

sachaudh (approved):
LGTM 🚢  Just a suggestion as well

PR #3038, opened 2022-09-09
ROX-12493, ROX-12370, ROX-11262: Rework Central DB helm

connorgorman:
## Description

Lots of different changes at play here:
1. Make the configuration of Postgres a part of a config map that gets mounted in the Central DB container
2. Move Central DB configuration out from underneath the Central section and into it's own section
3. Add support for preexisting databases to be used by allowing the specification of a connection string (db password will still be required)

## Checklist

## Testing Performed

Will test with an external database, but CI should test with Postgres and without Postgres


image/templates/helm/stackrox-central/config/centraldb/pg_hba.conf
```diff
@@ -81,11 +81,11 @@
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
 
 # "local" is for Unix domain socket connections only
-local   all             all                                     trust
+local   all             all                                     scram-sha-256
```
connorgorman, line 84: This is to require password locally

image/templates/helm/stackrox-central/internal/config-shape.yaml
```diff
@@ -94,6 +115,12 @@ customize:
     podLabels: {} # dict
     podAnnotations: {} # dict
     envVars: {} # dict
+  centralDB:
```
SimonBaeumer, line 118: scanner's db is `scanner-db`. The helm chart is a bit inconsistent already, but let's try to keep it as consistent as possible here.

```suggestion
  central-db:
```

SimonBaeumer (dismissed):
Can you protect the central-db resources and documentation by a feature flag please or setting this `centralDB` to disabled?
Also ignoring the newly created PVC file from the `.helmtplignore.htpl` would be good.

You can check for feature flags on the meta-templating stage like:
```
[< if .FeatureFlags.ROX_LOCAL_IMAGE_SCANNING >]
```

image/templates/helm/stackrox-central/templates/01-central-12-central-db.yaml
```diff
@@ -160,8 +159,9 @@ spec:
   selector:
     app: central-db
   type: ClusterIP
-{{- if ._rox.central._dbPassword }}
-{{- if not (kindIs "invalid" ._rox.central._dbPassword.value) }}
+{{ end }}
```
misberner, line 162:
```suggestion
{{- end }}
```

image/templates/helm/stackrox-central/config/centraldb/pg_hba.conf
misberner, line 84: Rename this file to `pg_hba.conf.default`

image/templates/helm/stackrox-central/config/centraldb/postgresql.conf
```diff
@@ -42,7 +42,7 @@
 #data_directory = 'ConfigDir'		# use data in another directory
 					# (change requires restart)
 #hba_file = 'ConfigDir/pg_hba.conf'	# host-based authentication file
-hba_file = '/etc/pg_hba.conf'		# StackRox customized
+hba_file = '/etc/stackrox.d/config/pg_hba.conf'		# StackRox customized
```
misberner, line 45: Rename this file to `postgresql.conf.default`

image/templates/helm/stackrox-central/internal/config-shape.yaml
```diff
@@ -41,11 +41,6 @@ central:
     name: null # string
     tag: null # string
     fullRef: null # string
-  dbImage:
```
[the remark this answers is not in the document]
connorgorman: I think that it's cleaner at the top level to not embed different sets of values for different services. I would suggest that for each service/deployment that we keep the configuration separate from the other services. so that each service has it's own resources field as opposed to dbResources field. Where this does get a bit hairy is when talking about the connection string that links Central <-> Central DB. I opted for having this defined in centralDB but perhaps that is not the right place

image/templates/helm/stackrox-central/templates/01-central-11-db-pvc.yaml
```diff
@@ -0,0 +1,66 @@
...
+    storage: {{ include "srox.formatStorageSize" $pvcCfg.size }}
+  accessModes:
+  - ReadWriteOnce
+  claimRef:
+    namespace: {{ .Release.Namespace }}
+    name: {{ $claimName }}
+  {{- toYaml $pvcCfg.volume.volumeSpec | nindent 2 }}
+---
+  {{- end }}
+  {{- /* TODO(ROX-9807): Move customized PVC from Central to Central DB */}}
```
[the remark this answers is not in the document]
connorgorman: thanks, goland is so annoying and indents everything

tests/upgrade/postgres_run.sh
```diff
@@ -183,7 +183,7 @@ helm_upgrade_to_current_with_postgres() {
     fi
 
     # enable postgres
-    yq e '.defaults.central.enableCentralDB=true' -i /tmp/stackrox-central-services-chart/internal/defaults.yaml
+    yq e '.defaults.centralDB.enabled=true' -i /tmp/stackrox-central-services-chart/internal/defaults.yaml
```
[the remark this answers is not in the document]
connorgorman: yeah good point

image/templates/helm/stackrox-central/internal/config-shape.yaml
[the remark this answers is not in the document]
connorgorman: I will actually move this underneath central as a db: setting

image/templates/helm/stackrox-central/internal/config-shape.yaml
connorgorman: renamed to db:

janisz (commented):
How about splitting this into 3 different PRs?

roxctl/central/generate/validate.go
```diff
@@ -25,7 +25,17 @@ func validateHostPath(hostpath *renderer.HostPathPersistence) error {
 	if hostpath == nil {
 		return nil
 	}
-	if (hostpath.NodeSelectorKey == "") != (hostpath.NodeSelectorValue == "") {
+	if err := validateHostPathInstance(hostpath.Central); err != nil {
+		return err
+	}
+	return validateHostPathInstance(hostpath.DB)
```
janisz, line 31: Could you add a unit tests for this change?

connorgorman: done

deploy/common/k8sbased.sh
```diff
@@ -54,7 +54,7 @@ function hotload_binary {
   echo "**********"
   echo
 
-  binary_path=$(realpath "$(git rev-parse --show-toplevel)/bin/linux_amd64/${local_name}")
+  binary_path=$(realpath "$(git rev-parse --show-toplevel)/bin/linux_$(uname -m)/${local_name}")
```
connorgorman, line 57: will remove

connorgorman:
@misberner @SimonBaeumer This should be ready for review again. Thanks in advance!

image/templates/helm/stackrox-central/templates/01-central-08-configmap.yaml.htpl
```diff
@@ -12,15 +12,21 @@ metadata:
 data:
   central-config.yaml: |
     {{- ._rox.central._config | nindent 4 }}
-    {{- if ._rox.central.enableCentralDB }}
+[< if .FeatureFlags.ROX_POSTGRES_DATASTORE >]
```
misberner, line 15: I wonder if this shouldn't have better gone into a separate config file.

image/templates/helm/stackrox-central/values.yaml.htpl
```diff
@@ -193,7 +184,68 @@
 #      # Specify a custom hostname if desired, otherwise accept the default from OpenShift.
 #      host: null
 #
+[< if .FeatureFlags.ROX_POSTGRES_DATASTORE >]
+#  db:
+#    enabled: false
+#
+#    # Preexisting signifies that a Postgres wire-compatible database has already been deployed and a Central DB pod
+#    # does not need to be deployed
+#    preexisting: false
```
misberner, line 193: Before getting to this line, I had a misconception about what `preexisting` meant. It's confusing because `helm` values aren't used for newly deploying only, but also for upgrading, and in case of anything but the first upgrade, the DB will also be pre-existing, yet still is Helm-managed. Maybe call this `external`?

image/templates/helm/stackrox-central/values.yaml.htpl
```diff
@@ -193,7 +184,68 @@
 #      # Specify a custom hostname if desired, otherwise accept the default from OpenShift.
 #      host: null
 #
+[< if .FeatureFlags.ROX_POSTGRES_DATASTORE >]
+#  db:
+#    enabled: false
```
misberner, line 189: Does `central.db.enabled` even make sense as a setting, in addition to the feature flag? With the feature flag turned on, will it be at all possible to opt out of postgres?

misberner (changes requested):

image/templates/helm/stackrox-central/values.yaml.htpl
connorgorman: External works

image/templates/helm/stackrox-central/values.yaml.htpl
connorgorman: Maybe this is a little future proofing, but in the staged rollout, the feature flag will become an env var (this will also cause a series of problems in this PR with the changes to special casing things with the feature flag) so enabling and disabling will be possible in that context. I can remove in this PR if that's cleaner

image/templates/helm/stackrox-central/templates/01-central-08-configmap.yaml.htpl
connorgorman: Yeah will file this as a follow up

misberner: image

misberner (approved):

PR #3043, opened 2022-09-11
ROX-12491: Fix link between deployments to processes in Postgres

connorgorman:
## Description

Fix search on risk page for processes

## Checklist

## Testing Performed

Will write an e2e once the image builds the first time


sensor/kubernetes/fake/deployment.go
```diff
@@ -268,9 +268,12 @@ func getContainer(workload ContainerWorkload) corev1.Container {
 			},
 			{
 				Name:  "ROX_SECRET_PASSWORD",
-				Value: "roxapitoken",
+				Value: "secretpassword",
 				ValueFrom: &corev1.EnvVarSource{
 					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
```
connorgorman, line 274: This was not a valid deployment in the fake k8s, but fixes the issue where the secret query was empty which was due to an empty name

PR #3051, opened 2022-09-12
Improve consistency between `cli` and `copy-go-binaries-to-image-dir`

msugakov:
## Description

Improve consistency between `cli` and `copy-go-binaries-to-image-dir` targets in the Makefile.

This follows up on https://github.com/stackrox/stackrox/pull/3001 and https://github.com/stackrox/stackrox/pull/3004.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


connorgorman:
Will test this locally, not sure this will be that simple because if I build an image locally then I actually need both linux arm64 and darwin arm64. One for local run and one for the one that goes in the container

Makefile
```diff
@@ -347,19 +347,23 @@ ifeq ($(UNAME_S),Darwin)
...
+HOST_BIN_PLATFORM := bin/$(HOST_OS)_$(GOARCH)
+
 .PHONY: build-prep
 build-prep: deps
 	mkdir -p bin/{darwin_amd64,darwin_arm64,linux_amd64,linux_arm64,linux_ppc64le,linux_s390x,windows_amd64}
 
 .PHONY: cli-build
 cli-build: build-prep
+ifndef CI
+	RACE=0 CGO_ENABLED=0 GOOS=$(HOST_OS) GOARCH=$(GOARCH) $(GOBUILD) ./roxctl
```
connorgorman, line 359: Yeah so for Darwin and arm64, will need to create both linux/arm64 and darwin/arm64 so that local roxctl will work and then the image will work as well because that is linux/arm64

connorgorman: Actually, nvm, it seems to work correctly due to the later build :)

msugakov: Somehow, there's `main-build-nodeps` target that takes care of linux/$(GOARCH) version when `CI` isn't set.

misberner: So, since we are cleaning up here, there's maybe a more fundamental issue we should fix:
There are two (three) use cases for building `roxctl`:
1. Building `roxctl` for the host platform, for use in scripts etc.
2. Building `roxctl` for all platforms for embedding in the image
3. (Building `roxctl` for the target platform for a quicker CI build)

2 and 3 are alternatives and 1 might overlap with 2/3.

However, the current set of targets doesn't follow that structure. Instead, there's a target to accomplish 1+2 in CI or just 1 in non-CI, and 3 is (if desired) accomplished via `main-build-nodeps`. Also, 1 is accomplished in conjunction with the target `cli`, of which `cli-build` (that might also accomplish 2 in CI, but not in non-CI) is a dependency.

I'd propose to restructure the target such that `cli-build` addresses 2/3 (i.e., in the `ifndef CI` case, build `roxctl` not for the _host_ but for the _target_ platform), make `cli-build` a direct dependency of `main-build-nodeps` (because it is responsible for producing the Go binaries that go into the main image; if you take offence at the fact that `-nodeps` has a dependency, just build it as a sub-target via `$(MAKE) cli-build`), remove the conditional build of `roxctl` from `main-build-nodeps`, and rephrase `cli` in terms of `go install` and make it build `roxctl` for the _host_ platform only.

IMHO this would be much clearer. This means we might build `roxctl` twice for the host platform (once for embedding in the image if the host platform matches the target platform or we're in CI), but at least the binaries in the `bin/{platform}` directories will never be in conflict, because we circumvent this for the (1) case via `go install`. I don't think there's an easy way to avoid this duplication and it seems tolerable; the only way around it seems to be to introduce a dedicated target for _each_ platform.

Makefile
```diff
@@ -600,15 +596,14 @@ copy-go-binaries-to-image-dir:
 	cp bin/linux_$(GOARCH)/central image/bin/central
 ifdef CI
 	cp bin/linux_amd64/roxctl image/bin/roxctl-linux-amd64
+	cp bin/linux_arm64/roxctl image/bin/roxctl-linux-arm64
```
misberner, line 599: WDYT about having a makefile variable contain a list of all roxctl platforms? this would make it easier to keep things consistent, especially with the CI/non-CI flavors.

tests/roxctl/bats-tests/helpers.bash
```diff
@@ -9,10 +9,7 @@ fi
 load "${bats_helpers_root}/bats-support/load.bash"
 load "${bats_helpers_root}/bats-assert/load.bash"
 
-# luname outputs uname in lowercase
-luname() {
-  uname | tr '[:upper:]' '[:lower:]'
-}
+host_bin_platform="$(make --quiet host-bin-platform)"
```
misberner, line 12:
```suggestion
host_bin_platform="$(make --quiet --no-print-directory host-bin-platform)"
```
At least I've been tripped up by this a couple of time, I think, when there is an ancestor `make` process calling this script that runs in a separate working directory. Or is `--no-print-directory` only needed in conjunction with `-C`?

Makefile
msugakov: @misberner Could you please take a look at recent commits and tell if I'm going in the right direction?
There may be CI failures, test failures, absent reusable values and other things at the moment, but the most important for me at this point is to validate if I got your idea right.

tests/roxctl/bats-tests/helpers.bash
msugakov: I learned that `--quiet` is superset of `--no-print-directory`, at least with my version of `make`.
At least it works the same with `-C`:

```sh
$ cdrox
$ make -C operator image-flavor 
make: Entering directory '/home/misha/go/src/github.com/stackrox/stackrox/operator'
opensource
make: Leaving directory '/home/misha/go/src/github.com/stackrox/stackrox/operator'
$ make --quiet -C operator image-flavor
opensource
$ make --no-print-directory -C operator image-flavor
opensource
```

The difference between `--quiet` and `--no-print-directory` becomes more apparent if you use something like `make -C operator bundle-post-process`.
`--no-print-directory` suppresses messages about changing directory.
`--quiet` suppresses messages about changing directory plus echoing the command that's about to be ran.

misberner: Hmm, as I said I've gotten tripped up by this in surprising ways. I experimented a bit, and it seems to me that (on GNU Make 4.3) the behavior is such that `--quiet` does only suppress printing directories if none of the ancestor `make`s printed directories, but not if an ancestor already printed directories - then you need `--no-print-directories`. We should never assume anything about how a caller invoked an ancestor `make`, so let's pass `--no-print-directory` in addition to `--quiet`.

```
$ cat Makefile 
.PHONY: bar bar-cd bar-npd bar-cd-npd

bar:
	$(MAKE) --quiet -C bar

bar-cd:
	cd bar && $(MAKE) --quiet

bar-npd:
	$(MAKE) --quiet --no-print-directory -C bar

bar-cd-npd:
	cd bar && $(MAKE) --quiet --no-print-directory

$ cat bar/Makefile
.PHONY: bar

bar:
	echo bar

$ make --quiet bar
bar # ok, expected
$ make --quiet -C . bar
bar #ok, expected
$ make bar
make --quiet -C bar # ok, expected due to omitting --quiet
bar
$ make -C . bar
make: Entering directory '/private/tmp/maketest' #ok, because of -C .. and no --quiet in first invocation
make --quiet -C bar
make[1]: Entering directory '/private/tmp/maketest/bar' # surprise! printed in spite of --quiet for this invocation
bar
make[1]: Leaving directory '/private/tmp/maketest/bar'
make: Leaving directory '/private/tmp/maketest'
$ make -C . bar-npd
make: Entering directory '/private/tmp/maketest' #ok, because of -C .. and no --quiet in first invocation
make --quiet --no-print-directory -C bar
bar #ok, no directories printing for second invocation
make: Leaving directory '/private/tmp/maketest'
```
The behavior of the `-cd` versions is the same as those using an explicit `-C`.

Because directories seem to be printed to stdout instead of stderr, this can break a `$(make --quiet -C .. tag)` expansion (as we use in the operator `Makefile`), but this breakage will _only_ be apparent when `make` on the `operator/` directory is invoked either with an explicit `-C` and no `--quiet`, or via `cd && make` from another `Makefile` that was invoked without `--quiet`.

Pretty complicated stuff, so I decided to just always pass both when using command expansion.

msugakov: Thanks, this got reproduced for me with `make` 4.2.1.

Makefiles calling scripts that call Makefiles which will start calling into other Makefiles some time later after refactoring.
If I was smarter I would come up with a comprehensive ruleset when `--quiet` or `--no-print-directory` is needed and when it can safely be omitted, but this can be a tiny problem that's not worth time spent.
I agree, let's do `--quiet --no-print-directory` together.

Makefile
```diff
@@ -431,16 +410,27 @@ ifeq ($(CIRCLE_JOB),build-race-condition-debug-image)
...
 	docker cp builder:/go/src/github.com/stackrox/rox/bin/linux bin/
 else
-	docker run -i -e RACE -e CI -e CIRCLE_TAG -e GOTAGS -e DEBUG_BUILD --rm $(GOPATH_WD_OVERRIDES) $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) make main-build-nodeps
+	docker run -i -e RACE -e CI -e CIRCLE_TAG -e GOTAGS -e DEBUG_BUILD --rm $(GOPATH_WD_OVERRIDES) $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) make cli-build main-build-nodeps
+endif
+
+.PHONY: cli-build
+cli-build:
+ifndef CI
+	RACE=0 CGO_ENABLED=0 GOOS=$(HOST_OS) GOARCH=$(GOARCH) $(GOBUILD) ./roxctl
```
misberner, line 419: I believe this should be
```suggestion
	RACE=0 CGO_ENABLED=0 GOOS=linux GOARCH=$(GOARCH) $(GOBUILD) ./roxctl
```
because we want to build a `roxctl` that works inside the container, and the container is built with the host arch, but will always be linux-based.

misberner: It's also required to match what is copied below

Makefile
```diff
@@ -600,15 +590,14 @@ copy-go-binaries-to-image-dir:
 	cp bin/linux_$(GOARCH)/central image/bin/central
 ifdef CI
```
misberner, line 591: Nit: I would prefer to use `ifndef CI ... else ... fi` here, to match the shape of the `cli-build` target

msugakov:
Test


| First Header  | Second Header |
| ------------- | ------------- |
| Content Cell  | Content Cell  |
| Content Cell  | Content Cell  |

msugakov:
One commit got landed in https://github.com/stackrox/stackrox/pull/6546

PR #3054, opened 2022-09-12
Upstream release automation: use #eng-release Slack channel for notifications

0x656b694d:


vikin91 (approved):
No more dry-run tests without notifying everyone 🙂 
Or maybe we could redirect dry runs to the `#test-release-automation` channel? WDYT?

.github/workflows/cut-rc.yml
```diff
@@ -36,7 +36,7 @@ env:
   main_branch: ${{github.event.repository.default_branch}}
   docs_repository: openshift/openshift-docs
   jira_projects: ROX, RS, RTOOLS
-  slack_channel: C03KSV3N6N8 #test-release-automation
+  slack_channel: CMH5M8MHN #eng-release
```
tommartensen, line 39: What do you think about putting the channel ID in a secret?
This might make it easier to copy/develop the scripts in `test-gh-actions` repo.

0x656b694d: I avoid secrets only because they're out of my control.
I liked the @vikin91 idea above about dry-run condition, so maybe we can detect if running in stackrox and chose the right channel. This way we can keep hardcoded channel IDs and save time on sync. I'll think how to better do that.

PR #3055, opened 2022-09-12
Dashrews/ROX-12547 migration fk enable disable

dashrews78:
## Description

Encountered issue running migrations where a deployment had been deleted, but that deletion hadn't bee propagated to the active components yet, so that migration failed. so we need to disable the triggers on the table while processing the migration.  I evaluated the custom migrations and only the active components migration had an external foreign key.  So I left the other custom migrations alone.

Tackling this at a migration level.  The first step is to use GORM to turn off the foreign keys of the table(s) being migrated.  Once the data is pulled in, run a prune task that will remove rows violating the foreign key.  Then apply the keys.  Children need to be accounted for.  

So a couple of notes:
1. This should only be a problem when a top level table references another.  For example, active_components references deployment.  For other migrations where there are parent/child relationships, this should not be an issue as the migration is localized to a single migration and the migration uses the store.  So for example a migration of Image and all its children should be OK because the store handles the insertions based on the data in the parent.
2. I put in generation code that only applied to one migration, but I figured we should have it just in case we add others.  I only went down 1 level of child because I don't think we will need it to go deeper.
3. Most the reason I didn't go deeper is this is probably only really an issue for Rocks to Postgres.  Once we are doing Postgres to Postgres migrations will evolve as it is and it would probably make sense to handle all related things that change in an isolated migration.

## Checklist

## Testing Performed

set up long running cluster and ran migrations.

### DIFF
--- a/migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
+++ b/migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
@@ -31,14 +31,6 @@ var (
 				return errors.Wrap(err,
 					"moving clusters from rocksdb to postgres")
 			}
-			// Now that migrations are complete, turn the constraints back on
-			gormConfig := databases.GormDB.Config
-			gormConfig.DisableForeignKeyConstraintWhenMigrating = false
-			err := databases.GormDB.Apply(gormConfig)
-			if err != nil {
-				return errors.Wrap(err, "failed to turn on foreign key constraints")
-			}
-			pkgSchema.ApplySchemaForTable(context.Background(), databases.GormDB, schema.Table)
 			return nil
 		},
 	}
@@ -50,16 +42,9 @@ var (
 func move(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacy.Store) error {
 	ctx := sac.WithAllAccess(context.Background())
 	store := pgStore.New(postgresDB)
-	// We need to migrate so turn off foreign key constraints
-	gormConfig := gormDB.Config
-	gormConfig.DisableForeignKeyConstraintWhenMigrating = true
-	err := gormDB.Apply(gormConfig)
-	if err != nil {
-		return errors.Wrap(err, "failed to turn off foreign key constraints")
-	}
 	pkgSchema.ApplySchemaForTable(context.Background(), gormDB, schema.Table)
 	var clusters []*storage.Cluster
-	err = walk(ctx, legacyStore, func(obj *storage.Cluster) error {
+	err := walk(ctx, legacyStore, func(obj *storage.Cluster) error {
 		clusters = append(clusters, obj)
 		if len(clusters) == batchSize {
 			if err := store.UpsertMany(ctx, clusters); err != nil {
--- a/migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
+++ b/migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
@@ -31,6 +31,14 @@ var (
 				return errors.Wrap(err,
 					"moving clusters from rocksdb to postgres")
 			}
+			// Now that migrations are complete, turn the constraints back on
+			gormConfig := databases.GormDB.Config
+			gormConfig.DisableForeignKeyConstraintWhenMigrating = false
+			err := databases.GormDB.Apply(gormConfig)
+			if err != nil {
+				return errors.Wrap(err, "failed to turn on foreign key constraints")
+			}
+			pkgSchema.ApplySchemaForTable(context.Background(), databases.GormDB, schema.Table)
 			return nil
 		},
 	}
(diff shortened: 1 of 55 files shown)


connorgorman (changes requested):
Does re-enabling triggers in a faulty state make sense? Those active components will never be cleaned up? We should drop those that do not have deployments in the migration right? This exact issue is also why I've been avoiding adding the cluster_id fk on most objects

md2119:
Second Connor's concern. The migration orders reference tables before the referencing tables. If we come across a case that referenced record does not exist, then we should just drop violating records under the pretext that clean-up was underway and did not complete when migration kicked in. There is a chance that the sensor events could be racy and we may lose a legit event record, but it is better than central init failing.

md2119 (commented):
Second Connor's concern. The migration orders reference tables before the referencing tables. If we come across a case that referenced record does not exist, then we should just drop violating records under the pretext that clean-up was underway and did not complete when migration kicked in. There is a chance that the sensor events could be racy and we may lose a legit event record, but it is better than central init failing.

migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
```diff
@@ -43,8 +44,13 @@ func move(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacy.Store) e
 	ctx := sac.WithAllAccess(context.Background())
 	store := pgStore.New(postgresDB)
 	pkgSchema.ApplySchemaForTable(context.Background(), gormDB, schema.Table)
+	_, err := postgresDB.Exec(ctx, fmt.Sprintf("ALTER TABLE %s DISABLE TRIGGER ALL", schema.Table))
```
c-du, line 47: No need to do this. Checkout `pghelper.ForT(s.T(), true)`
`		&gorm.Config{
			NamingStrategy:                           pgutils.NamingStrategy,
			DisableForeignKeyConstraintWhenMigrating: disableConstraint,
		},`

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
```diff
@@ -30,6 +30,17 @@ var (
 				return errors.Wrap(err,
 					"moving active_components from rocksdb to postgres")
 			}
+			if err := prune(databases.PostgresDB); err != nil {
+				return errors.Wrap(err,
+					"pruning cluster_health_statuses")
```
connorgorman, line 35: I think this is the wrong name

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
```diff
@@ -119,6 +136,28 @@ func storeWalk(ctx context.Context, s legacy.Store, fn func(obj *storage.ActiveC
 	return nil
 }
 
+func prune(postgresDB *pgxpool.Pool) error {
+	ctx := sac.WithAllAccess(context.Background())
+	deleteStmt := `DELETE FROM active_components child WHERE NOT EXISTS
```
connorgorman, line 141: I think this should be more exact?
```
DELETE From active_components child where child.<referencing field> not in (select <referenced field> from deployments where child.deploymentid = parent.id)
```

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
```diff
@@ -119,6 +136,28 @@ func storeWalk(ctx context.Context, s legacy.Store, fn func(obj *storage.ActiveC
...
+	deleteStmt := `DELETE FROM active_components child WHERE NOT EXISTS 
+		(SELECT * from deployments parent WHERE child.deploymentid = parent.id)`
+	_, err := postgresDB.Exec(ctx, deleteStmt)
+	if err != nil {
+		log.WriteToStderrf("failed to clean up orphaned data for %s", schema.Table)
+		return err
+	}
+
+	// Now get the child
+	deleteStmt = `DELETE FROM active_components_active_contexts_slices child WHERE NOT EXISTS
```
connorgorman, line 150: This is not necessary as long as triggers are on due to delete cascade

connorgorman (changes requested):
see comments

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
dashrews78: @connorgorman Since I turned them off in Gorm before applying the schema for the parent table.  Gorm then applied the necessary table and children with foreign keys off.  So they are off when I move the data in.  Tradeoffs.  If we do this via Gorm it will turn the keys off for everything it creates in `pkgSchema.ApplySchemaForTable(context.Background(), gormDB, schema.Table)`. Then when we turn them back on, it will reapply them and thus ensure that all the data is properly constrained.  

We could be more finely controlled using sql with an alter on just active_components which would not impact its child.    Simple enough change to make if that is the preference.  Instead of disabling triggers I would drop the constraint which is what Gorm is doing behind the scenes.  That way we find out at that time if we have reasonable data integrity.

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
connorgorman: Nvm, the where not exists is twice as fast!

migrator/migrations/n_05_to_n_06_postgres_active_components/migration.go
connorgorman: @dashrews78 The thought then is that we should just do this after we have re-enabled triggers?

dashrews78: if we do it with SQL by disabling triggers, then yes.  We could drop the constraint instead.  That would work in the same order as this version.  That might be simpler and more obvious.  It also wouldn't have any side effects with other things like unique constraints.  I'm thinking if we want to do it more controlled, that would probably be a better option

tools/generate-helpers/pg-table-bindings/migration.go.tpl
```diff
@@ -49,6 +50,19 @@ var (
 				return errors.Wrap(err,
 					"moving {{.Table|lowerCase}} from rocksdb to postgres")
 			}
+			{{- if gt (len .Schema.RelationshipsToDefineAsForeignKeys) 0 }}
+			if err := prune(databases.PostgresDB); err != nil {
+				return errors.Wrap(err,
+				"pruning {{.Table|lowerCase}}")
+			}
+			// Now that migrations are complete, turn the constraints back on
```
c-du, line 58: We do not need to turn on the constraints in each migration. We can turn them on altogether after persist.

dashrews78:
Closing this PR.  We changed directions and the work was complete under https://github.com/stackrox/stackrox/pull/3132.

PR #3062, opened 2022-09-13
ROX-9342: Disallow using `anyuid` SCC for sensor and sensor-upgrader

vikin91:
## Description

~~This is a follow-up to #2928 . After applying this, sensor will use SCCs in the following order of preference: `stackrox-sensor` (if created) > `nonroot-v2` (on OCP v4.11) > `nonroot` (in all other cases).~~

This is a follow-up to #2928 . After applying this, sensor and admission-control will be able to use `restricted` or `restricted-v2` (on OCP v4.11) SCCs. This allows us to delete the Roles and RoleBindings and not worry about complex upgrade procedures.

## Checklist

- [x] Determined and documented upgrade steps - see https://docs.engineering.redhat.com/display/StackRox/Upgrade+to+73.0

### Upgrade steps

```shell
# Edit role `edit` from #2928
oc -n stackrox patch role edit -p '{"rules":[{"apiGroups":["*"],"resources":["*"],"verbs":["create","get", "list", "watch", "update", "patch", "delete","deletecollection"]}]}'

# cleanup roles and rolebindings (if exist)
oc -n stackrox delete RoleBinding admission-control-use-scc || true
oc -n stackrox delete RoleBinding sensor-use-scc || true
oc -n stackrox delete Role use-anyuid-scc || true

oc -n stackrox patch deploy sensor -p '{"spec":{"template":{"spec":{"securityContext":null}}}}'
oc -n stackrox patch deploy admission-control -p '{"spec":{"template":{"spec":{"securityContext":null}}}}'
# pods will be recreated automatically, sensor can experience a short downtime
```

## Testing Performed

- [x] Manual tests performed on #2928 during working on #2928
- [x] CI 
- [x] Tested upgrade commands manually


misberner (commented):
The change per se looks good, but I disagree that there are no manual upgrade steps to be done. Existing users will definitely not have a binding to use `nonroot` and `nonroot-v2`, because I think we also forgot upgrade steps in https://github.com/stackrox/stackrox/pull/2928

vikin91:
@misberner could you elaborate about the required manual upgrade steps resulting from #2928 ? We kept existing roles and role bindings and we changed only their properties. 
The RoleBinding `sensor-use-scc` exists already ([code@master](https://github.com/stackrox/stackrox/blob/master/image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml#L77)) and it has not been touched in #2928 . It binds the Role `use-anyuid-scc` (which we edited) to the `sensor` and `sensor-upgrader` service accounts.

I could try to do an upgrade test myself to verify that no manual upgrade steps are required, but maybe you see something that is missing by looking at the code?

image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml
```diff
@@ -49,6 +49,7 @@ volumes:
 
 {{- end }}
 
+# Despite the lack of 'anyuid' in this role, the name is kept as 'use-anyuid-scc' to avoid manual upgrade steps
```
porridge, line 52: What is the long-term plan for this role? If it's with us to stay for good, then I think leaving the `anyuid` in the name might be a wrong trade-off.

Otherwise, if we plan getting rid of it at some point (sorry, I lost track of the discussion), then I think the name is confusing enough that it makes sense to put this comment in a `kubernetes.io/description` annotation so that there is at least a chance that it will be visible to users.

vikin91: If we make sensor & sensor-upgrader run under `restricted` SCC, then this role will not be needed. If we need more permissive SCC, then we need to keep it.

vikin91: Let's wait for @misberner 's comments regarding required manual upgrade steps, cause if we need to do something special for the upgrade to work, then we can rename it already now.

vikin91:
Thanks for the answer!  I will do a manual upgrade test for a cluster created with yamls and extract the required manual upgrade steps out of that.

vikin91:
@misberner I tried to create a cluster from yaml manifests and then upgrade it to a newer version to extract the upgrade steps, but I am unable to get a cluster with the Role `use-anyuid-scc`. I may have missed a step or two, but for now I do not see any possibility to create a cluster using roxctl with yaml manifests and get the SCC-related roles and rolebindings. 

1. Is this possible, that the `use-anyuid-scc` role will be created only when installing with Helm or the operator?
2. Is this possible that a customer would **install** the product with helm but **upgrade** using yaml manifests?

This were my steps to create a cluster using roxctl with yaml manifests:

```shell
cdrox
git switch --detach f48dea4378163ba93e443a2d12ddf7d1416bcca6

rm -f bin/darwin/roxctl bin/darwin_amd64/roxctl
make cli-darwin

bin/darwin_amd64/roxctl central generate openshift --openshift-version 4 pvc
cd central-bundle
./central/scripts/setup.sh
oc -n stackrox apply -f central/
oc -n stackrox port-forward svc/central 8443:8443 >/dev/null &
cd ..

# now secured cluster
bin/darwin_amd64/roxctl --password "$(cat central-bundle/password)" --insecure-skip-tls-verify sensor generate openshift --openshift-version=4 --name=cluster1

cd sensor-cluster1/
./sensor.sh
``` 

Checking what has been created:

```
$ oc -n stackrox get roles,rolebindings,scc

NAME                                                          CREATED AT
role.rbac.authorization.k8s.io/edit                           2022-09-15T09:14:39Z
role.rbac.authorization.k8s.io/stackrox-central-diagnostics   2022-09-15T08:22:17Z
role.rbac.authorization.k8s.io/use-central-db-scc             2022-09-15T08:22:15Z
role.rbac.authorization.k8s.io/watch-config                   2022-09-15T09:15:46Z

NAME                                                                   ROLE                                         AGE
rolebinding.rbac.authorization.k8s.io/admission-control-watch-config   Role/watch-config                            23h
rolebinding.rbac.authorization.k8s.io/central-db-use-scc               Role/use-central-db-scc                      24h
rolebinding.rbac.authorization.k8s.io/manage-namespace                 Role/edit                                    23h
rolebinding.rbac.authorization.k8s.io/stackrox-admission-control-psp   ClusterRole/stackrox-admission-control-psp   23h
rolebinding.rbac.authorization.k8s.io/stackrox-central-diagnostics     Role/stackrox-central-diagnostics            24h
rolebinding.rbac.authorization.k8s.io/stackrox-central-psp             ClusterRole/stackrox-central-psp             24h
rolebinding.rbac.authorization.k8s.io/stackrox-collector-psp           ClusterRole/stackrox-collector-psp           23h
rolebinding.rbac.authorization.k8s.io/stackrox-sensor-psp              ClusterRole/stackrox-sensor-psp              23h
rolebinding.rbac.authorization.k8s.io/system:deployers                 ClusterRole/system:deployer                  24h
rolebinding.rbac.authorization.k8s.io/system:image-builders            ClusterRole/system:image-builder             24h
rolebinding.rbac.authorization.k8s.io/system:image-pullers             ClusterRole/system:image-puller              24h

NAME                                                                               PRIV    CAPS                   SELINUX     RUNASUSER          FSGROUP     SUPGROUP    PRIORITY     READONLYROOTFS   VOLUMES
securitycontextconstraints.security.openshift.io/anyuid                            false   <no value>             MustRunAs   RunAsAny           RunAsAny    RunAsAny    10           false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/hostaccess                        false   <no value>             MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","hostPath","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/hostmount-anyuid                  false   <no value>             MustRunAs   RunAsAny           RunAsAny    RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","hostPath","nfs","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/hostnetwork                       false   <no value>             MustRunAs   MustRunAsRange     MustRunAs   MustRunAs   <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/hostnetwork-v2                    false   ["NET_BIND_SERVICE"]   MustRunAs   MustRunAsRange     MustRunAs   MustRunAs   <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/machine-api-termination-handler   false   <no value>             MustRunAs   RunAsAny           MustRunAs   MustRunAs   <no value>   false            ["downwardAPI","hostPath"]
securitycontextconstraints.security.openshift.io/node-exporter                     true    <no value>             RunAsAny    RunAsAny           RunAsAny    RunAsAny    <no value>   false            ["*"]
securitycontextconstraints.security.openshift.io/nonroot                           false   <no value>             MustRunAs   MustRunAsNonRoot   RunAsAny    RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/nonroot-v2                        false   ["NET_BIND_SERVICE"]   MustRunAs   MustRunAsNonRoot   RunAsAny    RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/privileged                        true    ["*"]                  RunAsAny    RunAsAny           RunAsAny    RunAsAny    <no value>   false            ["*"]
securitycontextconstraints.security.openshift.io/restricted                        false   <no value>             MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/restricted-v2                     false   ["NET_BIND_SERVICE"]   MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
securitycontextconstraints.security.openshift.io/stackrox-admission-control        false   []                     RunAsAny    RunAsAny           RunAsAny    RunAsAny    0            true             ["configMap","downwardAPI","emptyDir","secret"]
securitycontextconstraints.security.openshift.io/stackrox-central                  false   []                     MustRunAs   MustRunAs          MustRunAs   RunAsAny    0            true             ["awsElasticBlockStore","azureDisk","azureFile","cephFS","cinder","configMap","csi","downwardAPI","emptyDir","ephemeral","fc","flexVolume","flocker","gcePersistentDisk","gitRepo","glusterfs","iscsi","nfs","persistentVolumeClaim","photonPersistentDisk","portworxVolume","projected","quobyte","rbd","scaleIO","secret","storageOS","vsphere"]
securitycontextconstraints.security.openshift.io/stackrox-collector                true    []                     RunAsAny    RunAsAny           RunAsAny    RunAsAny    0            true             ["configMap","downwardAPI","emptyDir","hostPath","secret"]
securitycontextconstraints.security.openshift.io/stackrox-sensor                   false   []                     RunAsAny    RunAsAny           RunAsAny    RunAsAny    0            true             ["configMap","downwardAPI","emptyDir","secret"]
```

misberner:
It looks like SCCs are created by default in all non-Operator deployments (see `image/templates/helm/stackrox-secured-cluster/internal/defaults/60-sccs.yaml.htpl`). Furthermore, it seems that for `roxctl` YAML bundle deployment (the only relevant case for upgrade instructions), the user has no possibility to opt out of SCC creation.

However, https://github.com/stackrox/stackrox/pull/2928/files#diff-d1f6195d4ec0c9d91a79d33542e189ad37606672106d39ea52a8d301c9d6e479R4 caused the creation of role&role bindings to become unconditional even when `createSCCs = true`. Therefore, the adequate update instructions for #2928 would have been to include `oc create -f - <<EOF ... EOF` commands to create these for _all_ OpenShift users in _all_ secured clusters that have been deployed via YAML bundle.

However, since the above change isn't released yet, you are right that there is no need to write upgrade instructions for anything but the final version. But nonetheless, there will have to be [upgrade instructions](https://docs.openshift.com/acs/3.71/upgrading/upgrade-roxctl.html#upgrade-secured-clusters). I'm not sure what you mean by "(and upgrade) method described in <access.rh link>", because on that link I only find installation, not upgrade instructions.

I therefore think that at least a placeholder entry in the upgrade instructions wiki page is required.

vikin91:
I agree, the upgrade instructions are necessary as the upgrade procedure is not simply "rerun the installation steps from a newer version" (this is what I did in the previous step).  I will add them to the wiki.

vikin91:
I think the reason why the roles and rolebindings were missing in `f48dea` is caused by this change in `image/templates/helm/stackrox-secured-cluster/templates/sensor-scc.yaml`.

In `f48dea` we have:
```yaml
{{- if and ._rox.env.openshift ._rox.system.createSCCs }}

apiVersion: security.openshift.io/v1
kind: SecurityContextConstraints
metadata:
  name: stackrox-sensor
(...)

{{- else if eq ._rox.env.openshift 4 }}

 kind: Role
metadata:
  name: use-anyuid-scc
(...)
---
kind: RoleBinding
metadata:
  name: sensor-use-scc
(...)

{{- end }}
```

Note that if we create `stackrox-sensor` SCC, then the Role and RoleBinding will not be created. As the `createSCCs=true` for yaml bundles, we will never enter into the `{{- else if eq ._rox.env.openshift 4 }}` branch.

In the current commit, this is fixed and the roles will be created on all OpenShift versions, regardless of `createSCCs` setting.

misberner:
Oh wow, it looks like `roleRef` [is immutable](https://stackoverflow.com/a/69069182). Don't think we've ever had such a case in our manifests. In this case I would suggest to either skip the rename, or rename the role binding as well.

misberner:
I guess lesson learnt is that a role binding should _always_ include the syntactic name of the role being bound, not a semantic generalization.

CHANGELOG.md
```diff
@@ -28,6 +28,11 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Technical Changes
 - ROX-11937: The Splunk integration now processes all additional standards of the compliance operator (ocp4-cis & ocp4-cis-node) correctly.
 
+### Technical Changes
```
misberner, line 31: This heading is duplicated

image/templates/helm/stackrox-secured-cluster/templates/admission-controller-scc.yaml
```diff
@@ -49,14 +49,34 @@ volumes:
 # "fake" document separator to aid GVK extraction heuristic
 ---
 {{- end }}
-
+apiVersion: rbac.authorization.k8s.io/v1
```
misberner, line 52: Why are we adding this in the admission controller SCC yaml file?

vikin91: Apparently admission controller uses the `anyuid` SCC and we deleted that from sensor, so I needed to restore it for the admission-controller to not introduce additional changes. But we could also check whether the admission-controller would be happy with `nonroot`.

vikin91:
@misberner take a look at [2ab61b3](https://github.com/stackrox/stackrox/pull/3062/commits/2ab61b3dd5e1a8bdf2dce8d92684f35cd51cec14) - I removed the hardcoded `runAsUser` and `fsGroup` settings to allow sensor and admission-control running with `restricted` SCC, which does not require any special roles and rolebindings.
This should simplify this PR, the upgrade steps, and improve the security - what do you think? 

(I may need to update the docs and some tests if we decide to go that way, but let's see what CI would say first)

image/templates/helm/stackrox-secured-cluster/templates/admission-controller.yaml
```diff
@@ -60,9 +60,11 @@ spec:
       nodeSelector:
         {{- ._rox.admissionControl._nodeSelector | nindent 8 }}
       {{- end}}
+      {{- if not ._rox.env.openshift }}
```
misberner, line 63: Is it intended to leave these settings for non-OpenShift platforms? Is it because in OpenShift, an SCC is responsible for filling in "good" values, but without an SCC, omitting these would mean to run as root?

misberner (approved):

image/templates/helm/stackrox-secured-cluster/templates/admission-controller.yaml
vikin91: I am not fully sure what would be the consequence of removing this for vanilla k8s clusters, so I felt safer leaving it as it is. I will give it a quick read maybe to learn what this change would mean for non-OpenShift clusters.

PR #3070, opened 2022-09-13
Disabling hovering on user events in flaky dashboard tests

dvail:
## Description

**Re-enables flaky tests, with hovering disabled during the click events.**

These tests would fail with the occasional `Warning: Can't perform a React state update on an unmounted component.` error during CI runs.

This was determined to be due to two things:

1. The fact that we spy on `console.error` and automatically fail any test that calls it, regardless of passing assertions.
2. PatternFly's `Tooltip` component having a 300ms fade in/fade out effect that is called by `setTimeout`.

These errors originally started due to the call to `user-event`'s
```typescript
        await user.click(resetButton);
```
which in order to better simulate a real user, dispatches a `hover`event internally before the `click` event. Since the element was being hovered, this would cause the Tooltip to display after 300ms via a `setTimeout` and `setState`, which throws an error due to the state update after the test exits.

A fix was merged that immediately unhovered the element after clicking in an attempt to prevent the tooltip from rendering:
```typescript
        await user.click(resetButton);
        await user.unhover(resetButton);
```

This mostly worked, but would still cause occasional flakes. My theory is that CI slow downs were taking long enough that the implicit `hover` was causing the tooltip to display before the `unhover` event was applied. I was able to replicate this reliably locally with:

```typescript
        await user.click(resetButton);
        await new Promise((res) => setTimeout(res, 300));
        // always calls `console.error`
        await user.unhover(resetButton);
```

The solution in this PR disables the automatic `hover` event entirely for these tests, instead of having to deal with the timing of intermediate tooltip renders.

## Checklist

## Testing Performed

`yarn test` locally

Manually executing `/test ui-unit-tests` twice

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/PatternFly/Widgets/AgingImages.test.tsx
@@ -156,9 +156,9 @@ describe('AgingImages dashboard widget', () => {
         expect(history.location.search).toContain('s[Image Created Time]=30d-180d');
     });
 
-    // Disabled due to React state updates on unmounted components from PatternFly/Tooltip
-    it.skip('should contain a button that resets the widget options to default', async () => {
-        const { user } = setup();
+    it('should contain a button that resets the widget options to default', async () => {
+        setup();
+        const user = userEvent.setup({ skipHover: true });
 
         await user.click(await screen.findByLabelText('Options'));
         const checkboxes = await screen.findAllByLabelText('Toggle image time range');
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #3070** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3070" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3070?utm_source=stack-comment).

pedrottimark (approved):
Tests are sharpening your trouble-shooting skills, the better to solve problems in the product.

PR #3071, opened 2022-09-13
[Perf][GraphQL] Use count instead of search

md2119:
## Checklist

### DIFF
--- a/central/graphql/resolvers/deployments.go
+++ b/central/graphql/resolvers/deployments.go
@@ -743,12 +743,11 @@ func (resolver *deploymentResolver) unresolvedAlertsExists(ctx context.Context, 
 	if err != nil {
 		return false, err
 	}
-	q.Pagination = &v1.QueryPagination{Limit: 1}
-	results, err := resolver.root.ViolationsDataStore.Search(ctx, q)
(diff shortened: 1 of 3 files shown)


central/graphql/resolvers/deployments.go
```diff
@@ -743,12 +743,11 @@ func (resolver *deploymentResolver) unresolvedAlertsExists(ctx context.Context,
 	if err != nil {
 		return false, err
 	}
-	q.Pagination = &v1.QueryPagination{Limit: 1}
-	results, err := resolver.root.ViolationsDataStore.Search(ctx, q)
+	count, err := resolver.root.ViolationsDataStore.Count(ctx, q)
```
connorgorman, line 746: I'm cool with this change, but what if we added Exists for a query? this would stop very early then and be extremely cheap

connorgorman (approved):
Count > Search, see potential comment about exists function

central/graphql/resolvers/deployments.go
md2119: Fair enough. Likely cannot accommodate it before my PTO.

connorgorman: Yeah, can you just cut a ticket for it? Feel free to ship in the meantime

PR #3077, opened 2022-09-14
ROX-11717: GitHub workflow use proper gh env vars

tommartensen:
## Description

* replace `GITHUB_TOKEN` with `GH_TOKEN` (preferred by `gh` CLI) 
* replace `secrets.GITHUB_TOKEN` with `github.token`
* add `GH_NO_UPDATE_NOTIFIER` where required

See https://cli.github.com/manual/gh_help_environment for the details.

> GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com API requests. Setting this avoids being prompted to authenticate and takes precedence over previously stored credentials.

> GH_NO_UPDATE_NOTIFIER: set to any value to disable update notifications. By default, gh checks for new releases once every 24 hours and displays an upgrade notice on standard error if a newer version was found

## Checklist
  - not required because CI change

## Testing Performed

Tested in https://github.com/stackrox/test-gh-actions/pull/25.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -229,7 +229,7 @@ jobs:
 
       - name: Comment on the PR
         env:
-          GITHUB_TOKEN: "${{ secrets.ROBOT_ROX_GITHUB_TOKEN }}"
+          GH_TOKEN: "${{ secrets.ROBOT_ROX_GITHUB_TOKEN }}"
         run: |
             source ./scripts/ci/lib.sh
             add_build_comment_to_pr
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -229,7 +229,7 @@ jobs:
 
       - name: Comment on the PR
         env:
-          GH_TOKEN: "${{ secrets.ROBOT_ROX_GITHUB_TOKEN }}"
+          GITHUB_TOKEN: "${{ secrets.ROBOT_ROX_GITHUB_TOKEN }}"
         run: |
             source ./scripts/ci/lib.sh
             add_build_comment_to_pr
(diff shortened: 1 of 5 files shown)


tommartensen:
@0x656b694d @vikin91 there is one more `GITHUB_TOKEN` where I am not sure if we should change it or not.

It is in [`build.yaml` Github workflow](https://github.com/stackrox/stackrox/blob/master/.github/workflows/build.yaml#L232)

and referencing `add_build_comment_to_pr()`:
https://github.com/stackrox/stackrox/blob/b1d8844d54711ecc433958e483a3a7b34a1fe7d1/scripts/ci/lib.sh#L1350-L1370

which in turn references `get_pr_details()`:
https://github.com/stackrox/stackrox/blob/b1d8844d54711ecc433958e483a3a7b34a1fe7d1/scripts/ci/lib.sh#L793-L854

From my understanding `get_pr_details()` is 1) expecting the `GITHUB_TOKEN` 2) compatible with different CIs - is that correct?

0x656b694d:
> From my understanding `get_pr_details()` is 1) expecting the `GITHUB_TOKEN` 2) compatible with different CIs - is that correct?

That could be it. I'd suggest to not touch those scripts.

0x656b694d (approved):
LGTM

vikin91:
> From my understanding get_pr_details() is 1) expecting the GITHUB_TOKEN 2) compatible with different CIs - is that correct?

We used to run under CircleCI and this may be a leftover. But I think, that this code may still be used in some scenarios in which we still rely on the free CircleCI plan, so I would rather keep it as it is. Optionally you can check with @gavin-stackrox , but this should not block this PR.

vikin91 (approved):

PR #2202, opened 2022-06-24
ROX-11014: New sensor integration tests

fredrb:
## Description

This PR introduces a new test structure for sensor that allows us to more easily write tests for individual resource testing. 
It requires a **real** kubernetes cluster to run. 

The role test below applied all three files with all possible orders and runs the assertion callback after each test. The tests will run on a new namespace that is created from scratch. The test namespace is deleted at the end of each permutation run. 

We currently force a 2 seconds sleep time at the beginning of the assertion to "give time for the re-sync to happen". Which is configured for 1 second. 

```golang
func (s *RoleDependencySuite) Test_PermutationTest() {
	s.testContext.RunWithResourcesPermutation(
		[]resource.YamlTestFile{
			resource.NginxDeployment,
			resource.NginxRole,
			resource.NginxRoleBinding,
		}, "Role Dependency", func(t *testing.T, testC *resource.TestContext) {
			// Test context already takes care of creating and destroying resources
			time.Sleep(2 * time.Second)
			assertLastDeploymentHasPermissionLevel(
				t,
				testC.GetFakeCentral().GetAllMessages(),
				storage.PermissionLevel_ELEVATED_IN_NAMESPACE,
			)
			testC.GetFakeCentral().ClearReceivedBuffer()
		},
	)
}
```

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #3080, opened 2022-09-14
Generate JUnit report for sensor tests

janisz:
## Description

Sensors tests are stored in different files then other go tests and junit report is not generated for them.
This PR stores all sensors tests output into a single file (the same as other tests) and generate junit report from it saving it in prow so it will be visible with lenses.

## Testing Performed
CI

## Refs
https://github.com/stackrox/stackrox/pull/2202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ endif
 DOCS_IMAGE = $(DEFAULT_IMAGE_REGISTRY)/docs:$(shell make --quiet --no-print-directory docs-tag)
 
 GOBUILD := $(CURDIR)/scripts/go-build.sh
-
+GO_TEST_OUTPUT_PATH=$(CURDIR)/test-output/test.log
 GOPATH_VOLUME_NAME := stackrox-rox-gopath
 GOCACHE_VOLUME_NAME := stackrox-rox-gocache
 
@@ -489,7 +489,7 @@ go-unit-tests: build-prep test-prep
 	set -o pipefail ; \
 	CGO_ENABLED=1 GODEBUG=cgocheck=2 MUTEX_WATCHDOG_TIMEOUT_SECS=30 GOTAGS=$(GOTAGS),test scripts/go-test.sh -p 4 -race -cover -coverprofile test-output/coverage.out -v \
 		$(shell git ls-files -- '*_test.go' | sed -e 's@^@./@g' | xargs -n 1 dirname | sort | uniq | xargs go list| grep -v '^github.com/stackrox/rox/tests$$' | grep -Ev $(UNIT_TEST_IGNORE)) \
-		| tee test-output/test.log
+		| tee $(GO_TEST_OUTPUT_PATH)
 	# Exercise the logging package for all supported logging levels to make sure that initialization works properly
 	for encoding in console json; do \
 		for level in debug info warn error fatal panic; do \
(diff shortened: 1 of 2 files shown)


janisz:
* master
  * **PR #3080** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3080" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3080?utm_source=stack-comment).

Makefile
```diff
@@ -491,7 +491,7 @@ sensor-integration-test: build-prep test-prep
 	set -eo pipefail ; \
 	for package in $(shell git ls-files ./sensor/tests | grep '_test.go' | xargs -n 1 dirname | uniq | sort | sed -e 's/sensor\/tests\///'); do \
 		CGO_ENABLED=1 GODEBUG=cgocheck=2 MUTEX_WATCHDOG_TIMEOUT_SECS=30 GOTAGS=$(GOTAGS),test scripts/go-test.sh -p 4 -race -cover -coverprofile test-output/coverage.out -v ./sensor/tests/$$package \
-		| tee test-output/$$(echo $$package | sed -e 's/\//\_/').integration.log; \
+		| tee $(GO_TEST_OUTPUT_PATH); \
```
lvalerom, line 494: I'm not sure if there was a reason to have the logs of each test in separate files. I'm ok with having them in one file though.

lvalerom (approved):
LGTM. I added a comment regarding the log files for `sensor-integration-test`.

lvalerom (changes requested):
I think `sensor-integration-test` should stay they way they were, otherwise we are overwriting the previous logs.

Makefile
```diff
@@ -489,18 +489,17 @@ go-unit-tests: build-prep test-prep
 .PHONE: sensor-integration-test
 sensor-integration-test: build-prep test-prep
 	set -eo pipefail ; \
-	for package in $(shell git ls-files ./sensor/tests | grep '_test.go' | xargs -n 1 dirname | uniq | sort | sed -e 's/sensor\/tests\///'); do \
-		CGO_ENABLED=1 GODEBUG=cgocheck=2 MUTEX_WATCHDOG_TIMEOUT_SECS=30 GOTAGS=$(GOTAGS),test scripts/go-test.sh -p 4 -race -cover -coverprofile test-output/coverage.out -v ./sensor/tests/$$package \
-		| tee test-output/$$(echo $$package | sed -e 's/\//\_/').integration.log; \
-	done \
+	CGO_ENABLED=1 GODEBUG=cgocheck=2 MUTEX_WATCHDOG_TIMEOUT_SECS=30 GOTAGS=$(GOTAGS),test scripts/go-test.sh -p 4 -race -cover -coverprofile test-output/coverage.out -v \
+		$(shell git git ls-files -- 'sensor/tests/*_test.go'  | sed -e 's@^@./@g' | xargs -n 1 dirname | sort | uniq | xargs go list) \
```
lvalerom, line 493: You have `git` twice in this line.

lvalerom (changes requested):
The `sensor-integration-tests` won't work like this. We rely on the test context to close some things for us, that's why we needed the `for` loop to execute the tests individually. If they run all at the same time, they won't pass. This behavior is something that we are planning to change at some point, but not right now.

lvalerom (approved):
LGTM

PR #3081, opened 2022-09-14
ROX-12434: do not use stackrox.io flavor for tagged builds

johannes94:
## Description

Nightly CI operator tests are failing because it tries to pull images from stackrox.io which is no longer available. This PR changes it to use "development_build" image flavor instead for the test.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


porridge (approved):
Screenshot from 2022-09-14 17-20-10
:ship: !!
Looks good from operator PoV, I haven't looked at the failures to see whether they are related.

vikin91:
Yes, they are related, but it should be easy to address them - the error messages suggest what is the issue.

tests/roxctl/bats-tests/helpers.bash
```diff
@@ -226,10 +227,13 @@ image_reference_regex() {
   local component="$2"
   local version="${3:-$any_version}"
 
-  case $registry_slug in
-    quay.io)
+  case "$registry_slug" in
+    quay.io/rhacs-eng)
```
vikin91, line 231: I think, there may be still some test cases that would use the old slug `quay.io`. We need to either change the assertions to assert against `quay.io/rhacs-eng` or  change this line to handle both cases: 
```suggestion
    quay.io|quay.io/rhacs-eng)
```

msugakov:
I want to open this for early review while waiting for the CI results.

janisz:
Can you split it into series of smaller PRs?

tests/roxctl/bats-tests/local/central-generate-development-postgres.bats
```diff
@@ -5,10 +5,11 @@ load "../helpers.bash"
 out_dir=""
 
 setup_file() {
+  # remove binaries from the previous runs
+  [[ -n "$NO_BATS_ROXCTL_REBUILD" ]] || rm -f "${tmp_roxctl}"/roxctl*
```
vikin91, line 9: Cool for speeding-up local testing!

tests/roxctl/bats-tests/local/central-generate-release.bats
```diff
@@ -88,7 +93,7 @@ teardown() {
 }
 
 @test "roxctl-release central generate k8s should not support --central-db-image" {
-  run roxctl-development central generate k8s pvc --output-dir "$out_dir" --central-db-image example.com/central-db:1.2.5
+  run roxctl-release central generate k8s pvc --output-dir "$out_dir" --central-db-image example.com/central-db:1.2.5
```
vikin91, line 96: Good catch!

vikin91 (approved):
Quite many changes in the bats tests, but I skimmed through them and they look good!
The main change is about hiding the dev flavor in release builds instead of preventing to use it, right?

msugakov:
Sorry, @janisz saw your commit just now. I try making semantic commits. It is similar to stacked PRs but all the context is in one PR. Try reviewing per commit.

janisz:
@msugakov got it but there's no way to approve by commit.

msugakov:
@janisz Yes, it's only to be sold entirely.

PR #3082, opened 2022-09-14
ROX-11323: Local-Sensor resource consumption monitoring: recording pprof

vikin91:
## Description

This PR adds automated recording of memory and CPU profile to `local-sensor`.
The feature is enabled by default and can be disabled by using flags: `--no-cpu-prof` and `--no-mem-prof`.

⚠️ This is one of potentially many PRs that will be needed to cover the scope of [ROX-11323](https://issues.redhat.com/browse/ROX-11323), but I wanted to deliver the features in small chunks.

## Checklist

Changes apply to a tool used locally.

## Testing Performed

```shell
$ go run tools/local-sensor/main.go -replay -replay-in=sensor/tests/replay/data/safety-net-alerts-k8s-trace.jsonl -resync=10s -delay=0s -format=raw -central-out=/dev/null

$ ls -l local-sensor-*.prof
-rw-r--r--  1 prygiels  staff   7952 Sep 16 17:10 local-sensor-cpu-2022-09-16T15:10:01Z.prof
-rw-r--r--  1 prygiels  staff  24648 Sep 16 17:10 local-sensor-mem-2022-09-16T15:10:39Z.prof

$ go tool pprof local-sensor-cpu-2022-09-16T15:10:01Z.prof 
Type: cpu
Time: Sep 16, 2022 at 5:10pm (CEST)
Duration: 38.43s, Total samples = 380ms ( 0.99%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) web
(pprof) q

$ go tool pprof local-sensor-mem-2022-09-16T15:10:39Z.prof 
Type: alloc_space
Time: Sep 16, 2022 at 5:10pm (CEST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) web
(pprof) q
```

### DIFF
--- a/tools/local-sensor/main.go
+++ b/tools/local-sensor/main.go
@@ -6,11 +6,14 @@ import (
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
 	"log"
 	"net"
 	"os"
 	"os/signal"
 	"path"
+	"runtime"
+	"runtime/pprof"
 	"syscall"
 	"time"
 
@@ -79,6 +82,8 @@ type localSensorConfig struct {
 	PoliciesFile       string
 	FakeWorkloadFile   string
 	WithMetrics        bool
+	NoCPUProfile       bool
+	NoMemProfile       bool
 }
 
 const (
@@ -113,7 +118,9 @@ func writeOutputInBinaryFormat(messages []*central.MsgFromSensor, _, _ time.Time
 		_, err = file.Write(d)
 		utils.CrashOnError(err)
 	}
-	utils.CrashOnError(file.Sync())
+	if outfile != "/dev/null" {
+		utils.CrashOnError(file.Sync())
+	}
 }
 
 var validFormats = map[string]func([]*central.MsgFromSensor, time.Time, time.Time, string){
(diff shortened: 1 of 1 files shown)


tools/local-sensor/main.go
```diff
@@ -171,13 +183,30 @@ func mustGetCommandLineArgs() localSensorConfig {
...
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 	<-ctx.Done()
 	// We cancel the creation of Events
 	cancelFunc()
 	endTime := time.Now()
+	if writeMemProfile {
+		writeMemoryProfile()
+	}
+	pprof.StopCPUProfile()
```
fredrb, line 209: Shouldn't this be guarded by the the `NoCPUProfile` flag as well? I wonder if `pprof` would panic/err if `StopCPUProfile` is called when `StartCPUProfile` wasn't.

vikin91: I tested all combinations and there was no panic. I needed to place it here, because when I put it in the deferred statement, the CPU profiles were empty. Maybe because we need to kill sensor with ctrl+c? I didn't look deeper

lvalerom (approved):

lvalerom:
@fredrb I'll merge this today if there are no objections 🙃

PR #3087, opened 2022-09-14
CI: Do not fail the build if the PR comment fails

gavin-stackrox:
## Description

This will no longer fail the build if a comment cannot be added to the PR. As the build is a required job and commenting on the PR is a non critical feature. This is motivated by recent github API limits but is a better approach in any case.

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/scripts/ci/jobs/push-images.sh
+++ b/scripts/ci/jobs/push-images.sh
@@ -67,7 +67,9 @@ push_images() {
     fi
 
     if is_in_PR_context && [[ "$brand" == "STACKROX_BRANDING" ]] && [[ -n "${MAIN_IMAGE}" ]]; then
-        add_build_comment_to_pr
+        add_build_comment_to_pr || {
+            info "Could not add a comment to the PR"
+        }
     fi
 }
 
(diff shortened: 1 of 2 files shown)


connorgorman (approved):

RTann (approved):

scripts/ci/jobs/push-images.sh
```diff
@@ -67,7 +67,9 @@ push_images() {
     fi
 
     if is_in_PR_context && [[ "$brand" == "STACKROX_BRANDING" ]] && [[ -n "${MAIN_IMAGE}" ]]; then
-        add_build_comment_to_pr
+        add_build_comment_to_pr || {
+            info "Could not add a comment to the PR"
```
RTann, line 71: should this be `warn` instead?

gavin-stackrox:
I jumped to the wrong conclusion pursuing this PR. The failure to add a comment to a PR will cause `ci/prow/stackrox_branding-push-images` images to be marked as failed (the images are still pushed) and that is *not* a required job. (`ci/prow/stackrox_branding-images` being the required job). But this is still a useful change to diminish noise and confusion.

PR #3094, opened 2022-09-15
ROX-10928: Deprecate release-tools scripts

tommartensen:
## Description

The release-tools scripts are deprecated and will be removed on October 10, 2022. The functionality is now implemented by GitHub Actions.

## Checklist

## Testing Performed

* scripts run manually to check for warning output

### DIFF
--- a/scripts/release-tools/README.md
+++ b/scripts/release-tools/README.md
@@ -1,3 +1,7 @@
+# DEPRECATION NOTICE
+
+:warning: The scripts are deprecated and will be removed on October 10, 2022. The functionality is now implemented by GitHub Actions.
+
 # Upstream Release Tools
 
 This directory contains a selection of scripts that automate selected mundane manual steps (a.k.a toil) in the upstream release process.
--- a/scripts/release-tools/clusters.sh
+++ b/scripts/release-tools/clusters.sh
@@ -27,7 +27,10 @@ main() {
   PS3="Choose action: "
   RED='\033[0;31m'
   NC='\033[0m' # No Color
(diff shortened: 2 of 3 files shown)


scripts/release-tools/clusters.sh
```diff
@@ -38,7 +41,7 @@ exec_option() {
   local num_options="${#MENU_OPTIONS[@]}"
   local last_option="$((num_options-1))"
   case "$1" in
-    "${MENU_OPTIONS[$last_option]}"|"$((last_option+1))"|q|Q) exit 0;;&
+    "${MENU_OPTIONS[$last_option]}"|"$((last_option+1))"|q|Q) exit 0;;
```
tommartensen, line 44:
```
./scripts/release-tools/clusters.sh: line 43: syntax error near unexpected token `&'
`
``

--> Remove `&`

File: scripts/release-tools/README.md [hunk 2]
```diff
@@ -1,3 +1,7 @@
+# DEPRECATION NOTICE
+
+:warning: The scripts are deprecated and will be removed on October 10, 2022. The functionality is now implemented by GitHub Actions.
```
vikin91, line 3: 10/10 looks pretty soonish. I have nothing against deleting them on that date, but I would propose the following steps to handle this:
- Unlist those scripts from the release manuals: upstream, downstream, tracker. Just remove all texts that mention them. This can happen on 10.10 👍 
- Delete the `release-tools` folder from the repo after 3.73*

*) I think that is the latest date, we can as well do it earlier.

vikin91 (approved):
Changes look good, thanks!

scripts/release-tools/README.md
tommartensen: I will invite you to a reminder on 10.10.

PR #3098, opened 2022-09-15
ROX-11101: Remove deprecated resources from UI

rhybrillou:
## Description

The set resources used for access control is being simplified.
A number of resources has been deprecated.
The goal here is to get rid of the deprecated resources, both in the database and in the codebase (UI, QA, central, DB). This PR focuses on the UI codebase removal.

## Checklist

## Testing Performed

CI is sufficient


ui/apps/platform/cypress/fixtures/auth/mypermissionsMinimalAccess.json
```diff
@@ -1,7 +1,7 @@
 {
     "resourceToAccess": {
         "Access": "NO_ACCESS",
-        "Administration": "NO_ACCESS",
+        "Administration": "READ_ACCESS",
```
rhybrillou, line 4: This change is because of `Config` which has read access.
Now allowed by `Administration` are the services that used to be tied to the `DebugLogs` permission, and which could allow data leaks. Keeping NO_ACCESS here could also be a good idea.

PR #3101, opened 2022-09-15
Force resolution of autoprefixer to 10.4.5

dvail:
## Description

Forces resolution to a lower version of `autoprefixer` to suppress deprecation warnings during build time.

These warning occur on every webpack rebuild during hot module reloading and were introduced by the react-scripts upgrade to v5. The upgrade switched our usage of PostCSS to the built-in version within react-scripts, which brought an updated version of autoprefixer (10.2.5 -> 10.4.8).

This downgrades us from 10.4.8 to 10.4.5, which according to the [changelog](https://github.com/postcss/autoprefixer/blob/main/CHANGELOG.md#1048) only contains changes related to this specific error message.

This seems like the safest fix that did not involve upgrading Tailwind, etc. since those packages will eventually be removed along with the deprecated generated code.

## Checklist

## Testing Performed

Spot check of Tailwind pages as this is a fairly superficial build change.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -44,7 +44,8 @@
         "forward": "../scripts/port-forward-ui.sh"
     },
     "resolutions": {
-        "@typescript-eslint/parser": "^4.32.0"
+        "@typescript-eslint/parser": "^4.32.0",
+        "autoprefixer": "10.4.5"
     },
     "devDependencies": {
         "chalk-cli": "^5.0.0",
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #3101** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3101" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3101?utm_source=stack-comment).

PR #3103, opened 2022-09-15
ROX-12672: Fix auth providers shown as immutable

dhaus67:
## Description

When creating auth providers via API instead of within UI, the field `traits` is set to `null` instead of omitted.

This lead to the auth provider being incorrectly shown as immutable, although the traits field was null.

It's due to the way we check for immutability. We should also check, if the traits field exists, whether its != null (this way checking if the value is either `null` or `undefined`).

This PR will add those additional checks (for both the auth provider and groups) and additionally changes the E2E tests to use `traits: null` as value, so we have this covered in the future.

## Testing Performed

- see E2E tests.

### DIFF
--- a/ui/apps/platform/cypress/fixtures/auth/authProviders-id1.json
+++ b/ui/apps/platform/cypress/fixtures/auth/authProviders-id1.json
@@ -13,9 +13,7 @@
             "loginUrl": "irrelevant",
             "validated": false,
             "active": false,
-            "traits": {
-                "mutabilityMode": "ALLOW_MUTATE"
-            }
+            "traits": null
         }
     ]
 }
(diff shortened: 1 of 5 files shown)


dhaus67:
* master
  * **PR #3103** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3103" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3103?utm_source=stack-comment).

pedrottimark (approved):

ui/apps/platform/cypress/fixtures/auth/authProviders-id1.json
```diff
@@ -13,9 +13,7 @@
             "loginUrl": "irrelevant",
             "validated": false,
             "active": false,
-            "traits": {
```
ivan-degtiarenko, line 16: Do we have **some** tests that work with "ALLOW_MUTATE"/"ALLOW_MUTATE_FORCE"  mutability modes?

ivan-degtiarenko (approved):
LGTM - I would add tests if possible, but if it's not that easy let's merge as it is before release is finalised

PR #3105, opened 2022-09-15
Call assertSortedItems in policies and risk integration tests

pedrottimark:
## Description

Follow up on comment by Dave

> I can still see value in testing that the items in the table are in sorted order though, e.g. that each item is >= or <= the following one in the table depending on the sort order. It might be something worth adding in the future if we revisit this test.

### History

1. The original test sometimes failed because of timing problems: get element too soon.
2. The simplified test has intercept-wait and assertions for search query and column heading indicator. I deleted the contains assertion, because the risk priority can change between the initial state and the sorted state.
3. The strengthened test depends on wait and assertions to get column elements in sorted state and assert the order.

### Overview

1. Write `assertSortedItems` function whose arguments are `items` array of unknown type and corresponding `callbackForPairOfSortedItems` function which returns a non-empty string if a comparison fails.

2. Write `createCallbackForPairOfSortedItems` higher-order function which returns a callback function from reusable functions:

    * `getValueFromItem`
    * `isValidValue`
    * `isPairOfSortedValues`

3. Export generic callback functions for tests in multiple containers:

    * `callbackForPairOfAscendingNumberValuesFromElements`
    * `callbackForPairOfDescendingNumberValuesFromElements`
    * `callbackForPairOfAscendingSeverityValuesFromElements`
    * `callbackForPairOfDescendingSeverityValuesFromElements`

Because backend sorting algorithm for strings might differ from JavaScript comparison operators or methods, especially considering possible effect of locale, it seems better not to test that arbitrary string values are sorted.

Here is an example of an artificial test failure when I temporarily edited sort.js to replace  `<=` with `<` in the `isPairOfAscendingSeverityValues` function.
policiesTable

Here is an example of an artificial test failure when I temporarily edited risk.test.js to replace `callbackForPairOfAscendingNumberValuesFromElements` with `callbackForPairOfDescendingNumberValuesFromElements` in the test as if get got ahead of re-rendering the table. Which happened during development before I added wait for /v1/deploymentscount request.
risk

### Changed files

1. Edit cypress/helpers/risk.js
    * Export whateverAlias constants.

2. Edit cypress/helpers/policiesPatternFly.js
    * Export whateverAlias constants.

3. Add cypress/helpers/sort.js

4. Edit cypress/integration/policies/policiesTable.test.js

5. Edit cypress/integration/risk/risk.test.js
    * Add `'should have table columns'` test.
    * Edit `'should sort the Priority column'` test:
        * Define `thSelector` and `tdSelector` constants to reduce the chance of copy-paste-edit errors when applying the pattern in other containers.
        * Wait for /v1/deploymentscount request to reduce the risk to get elements before the table has been re-rendered.
        * Call `assertSortedItems` helper function.

### Residue

1. Delete redundant `describe` blocks in risk.test.js file. Respect the changed files diff to review this pull request.
2. Call `assertSortedItems` in vulnmamagement integration tests.
    * Several tables have **Risk Priority** column.
    * Although some tables have **Severity** column, it looks like backend does not sort it yet. For now, write tests for **CVSS Score** decimal number value.

## Checklist

## Testing Performed

* risk.test.js

### DIFF
--- a/ui/apps/platform/cypress/helpers/sort.js
+++ b/ui/apps/platform/cypress/helpers/sort.js
@@ -1,7 +1,7 @@
 /*
  * Assert that each pair of adjacent items are in sorted order.
  * items might be DOM elements from cy.get(…) call.
- * callbackForPairOfSortedItems(itemA, itemB) is returned from create higher-order function below.
+ * callbackForPairOfSortedItems(itemA, itemB) is returned from higher-order function below.
  */
 export function assertSortedItems(items, callbackForPairOfSortedItems) {
     for (let indexB = 1; indexB < items.length; indexB += 1) {
--- a/ui/apps/platform/cypress/helpers/sort.js
+++ b/ui/apps/platform/cypress/helpers/sort.js
@@ -1,0 +1,120 @@
+/*
+ * Assert that each pair of adjacent items are in sorted order.
+ * items might be DOM elements from cy.get(…) call.
+ * callbackForPairOfSortedItems(itemA, itemB) is returned from create higher-order function below.
+ */
+export function assertSortedItems(items, callbackForPairOfSortedItems) {
+    for (let indexB = 1; indexB < items.length; indexB += 1) {
+        const itemA = items[indexB - 1];
+        const itemB = items[indexB];
+        const result = callbackForPairOfSortedItems(itemA, itemB);
+
+        // No news is good news. One incorrect result is enough bad news.
+        if (result !== '') {
+            expect(result).to.equal('');
+            break;
+        }
+    }
+}
+
+/*
+ * Create callback function for assertSortedItems function above.
+ * notValidDescription looks like a function name
+ * - Its prefix is not.
+ * - Its root is a description in PascalCase of valid and sorted values from items.
+ * getValueFromItem encapsulates access like item?.innerText property.
+ * isValidValue encapsulates expected type of value like string.
+ * isPairOfSortedValues encapsulates comparison like ascending or descending.
+ *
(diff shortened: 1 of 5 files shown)


pedrottimark:
The integration test failures are unrelated to the changes in this pull request and I am investigating them.

Therefore, we can keep moving forward to review now, but I will rebase and look for success before merging.

ui/apps/platform/cypress/helpers/sort.js
```diff
@@ -0,0 +1,120 @@
...
+    getValueFromItem,
+    isValidValue,
+    isPairOfSortedValues
+) {
+    return function callbackForPairOfItems(itemA, itemB) {
+        const valueA = getValueFromItem(itemA);
+        const valueB = getValueFromItem(itemB);
+
+        if (isValidValue(valueA) && isValidValue(valueB) && isPairOfSortedValues(valueA, valueB)) {
+            return ''; // return empty string for correctly sorted pair
```
bradr5, line 43: I like the way this file is setup. Returning the empty string felt a little odd at first but it makes sense now that I understand whats going on.

bradr5 (approved):

ui/apps/platform/cypress/helpers/sort.js
pedrottimark: The return value was an unexpected change in direction.
The first draft imitated `compareFn` for array sort method:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description

PR #3106, opened 2022-09-15
Update local/hack.go to run against local postgres DB

theencee:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


local/hack.go
```diff
@@ -1,26 +1,67 @@
...
+		//   docker run -it -p 5432:5432 -e POSTGRES_USER=${USER} -e POSTGRES_HOST_AUTH_METHOD=trust postgres:14.2-alpine
+		// 3. Restore the SQL dump from #1 to this local DB
+		//   cat local/central_active.sql |  docker exec -i $(docker ps -lq) sh -c 'psql -h localhost -U '$USER' -p 5432'
+		// 4. Update central/globaldb/postgres.go and add the following function:
+		//		func SetPostgres(pool *pgxpool.Pool) {
+		//			postgresDB = pool
+		//		}
+		// TODO: Expose a nicer way to set the DB connection without having to manually add the function
+
+		source := conn.GetConnectionStringWithDatabaseName(env.GetString("POSTGRES_DB", "central_active"))
```
md2119, line 36: The db name should really be a global const, if it is not already

local/hack.go
```diff
@@ -1,26 +1,69 @@
...
-	option.CentralOptions.DBPathBase = "local/database-restore/full"
+		// PREREQUISITES TO RUNNING POSTGRES
+		// 1. Get a SQL dump of DB that has the data you want]
+		//    kubectl exec -ti -n=stackrox $(kubectl get pod -n=stackrox -l app=central-db --output 'jsonpath={.items..metadata.name}') -- sh -c '/usr/bin/pg_dumpall' > local/central_active.sql
+		// 2. Run a docker container with postgres:
+		//   docker run -it -p 5432:5432 -e POSTGRES_USER=local-postgres -e POSTGRES_PASSWORD=local-pg-password postgres:14.2-alpine
+		// 3. Restore the SQL dump from #1 to this local DB
+		//   cat local/central_active.sql |  docker exec -i $(docker ps -lq) sh -c 'PGPASSWORD=local-pg-password psql -h localhost -U local-postgres -p 5432'
+		// 4. Go to pkg/config/config.go, and update the variable "defaultDBSource" to:
+		//		host=localhost port=5432 user=local-postgres sslmode=disable statement_timeout=600000 pool_min_conns=1 pool_max_conns=90 client_encoding=UTF-8
```
theencee, line 30: This would be a whole lot easier if the default config was a exported variable, but I'm not sure what risks that exposes. Any thoughts?

local/hack.go
```diff
@@ -1,26 +1,71 @@
...
 )
 
 func main() {
 	if features.PostgresDatastore.Enabled() {
-		return
-	}
 
-	option.CentralOptions.DBPathBase = "local/database-restore/full"
+		// PREREQUISITES TO RUNNING POSTGRES
+		// 1. Get a SQL dump of DB that has the data you want]
```
c-du, line 23: nit: Remove the ] at the end.

local/hack.go
c-du: You can check configmap central-config, in central-config.yaml centralDB.source

local/hack.go
```diff
@@ -1,26 +1,71 @@
...
+		//    kubectl exec -ti -n=stackrox $(kubectl get pod -n=stackrox -l app=central-db --output 'jsonpath={.items..metadata.name}') -- sh -c '/usr/bin/pg_dumpall' > local/central_active.sql
+		// 2. Run a docker container with postgres:
+		//   docker run -it -p 5432:5432 -e POSTGRES_USER=local-postgres -e POSTGRES_PASSWORD=local-pg-password postgres:14.2-alpine
+		// 3. Restore the SQL dump from #1 to this local DB
+		//   cat local/central_active.sql |  docker exec -i $(docker ps -lq) sh -c 'PGPASSWORD=local-pg-password psql -h localhost -U local-postgres -p 5432'
```
c-du, line 28: Can this be done with pg_restore?

local/hack.go
c-du: Oh, you do not have those.

theencee: Nope this is run locally (ie via IntelliJ or with go run) so no configmap is available

local/hack.go
theencee: I would need to change to using pg_dump and then export as a custom format or tar in line 24. And that is giving me troubles when running via exec because the pipe is dropping lines for whatever reason. The suggestions seems to be to [write it to a local file](https://stackoverflow.com/a/51073680) and then kubectl cp it out.

Problem is that the central-db container doesn't have permissions to write. Not even to /tmp:

```
pg_dump: error: could not open TOC file "/tmo/central_active2.tar" for output: No such file or directory
```

PR #3107, opened 2022-09-15
ROX-12670,ROX-12671: Fix Postgres dry-run by propagating proper context

connorgorman:
## Description

Copy over the context scope

## Checklist

## Testing Performed

Will add a test

### DIFF
--- a/central/policy/service/service_impl.go
+++ b/central/policy/service/service_impl.go
@@ -348,7 +348,7 @@ func (s *serviceImpl) SubmitDryRunPolicyJob(ctx context.Context, request *storag
 
 	t := func(c concurrency.ErrorWaitable) (interface{}, error) {
 		// Copy over SAC scopes into background context
-		ctx := sac.CopyScopeCheckerCore(context.Background(), ctx)
+		ctx := sac.CopyGlobalAccessScopeChecker(context.Background(), ctx)
 		return s.predicateBasedDryRunPolicy(ctx, c, request)
 	}
 
--- a/central/policy/service/service_impl.go
+++ b/central/policy/service/service_impl.go
@@ -30,6 +30,7 @@ import (
 	"github.com/stackrox/rox/pkg/booleanpolicy/networkpolicy"
 	"github.com/stackrox/rox/pkg/booleanpolicy/policyversion"
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/contextutil"
 	"github.com/stackrox/rox/pkg/detection"
 	"github.com/stackrox/rox/pkg/errorhelpers"
 	"github.com/stackrox/rox/pkg/errox"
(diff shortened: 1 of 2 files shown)


pkg/sac/context.go
```diff
@@ -37,3 +37,8 @@ func WithAllAccess(ctx context.Context) context.Context {
 func WithNoAccess(ctx context.Context) context.Context {
 	return WithGlobalAccessScopeChecker(ctx, denyAllScopeCheckerCore)
 }
+
+// CopyScopeCheckerCore from the source context to the destination context
+func CopyScopeCheckerCore(dstCtx, srcCtx context.Context) context.Context {
+	return context.WithValue(dstCtx, globalAccessScopeContextKey{}, srcCtx.Value(globalAccessScopeContextKey{}))
```
dhaus67, line 43:
```suggestion
	return WithGlobalAccessScopeChecker(dstCtx, GlobalAccessScopeChecker(srcCtx).Core())
```

dhaus67 (approved):

pkg/sac/context.go
```diff
@@ -37,3 +37,8 @@ func WithAllAccess(ctx context.Context) context.Context {
 func WithNoAccess(ctx context.Context) context.Context {
 	return WithGlobalAccessScopeChecker(ctx, denyAllScopeCheckerCore)
 }
+
+// CopyScopeCheckerCore from the source context to the destination context
```
dhaus67, line 41: super-nit:
```suggestion
// CopyScopeCheckerCore from the source context to the destination context.
```

dhaus67 (approved):

pkg/sac/context.go
```diff
@@ -37,3 +37,8 @@ func WithAllAccess(ctx context.Context) context.Context {
...
+func CopyScopeCheckerCore(dstCtx, srcCtx context.Context) context.Context {
```
dhaus67, line 42:
```suggestion
func CopyGlobalAccessScopeChecker(dstCtx, srcCtx context.Context) context.Context {
```

dhaus67 (approved):

pkg/sac/context.go
connorgorman: I actually prefer to not do this as someone could change the return value of Core and this function would not change (or if if GlobalAccessScopeChecker is nil then it would panic). I think the current code is a very clear copy

connorgorman:
Created https://issues.redhat.com/browse/ROX-12693 for the dry run async test which will be non-trivial

central/policy/service/service_impl.go
```diff
@@ -347,6 +347,8 @@ func (s *serviceImpl) SubmitDryRunPolicyJob(ctx context.Context, request *storag
 	}
 
 	t := func(c concurrency.ErrorWaitable) (interface{}, error) {
+		// Copy over SAC scopes into background context
+		ctx := sac.CopyGlobalAccessScopeChecker(context.Background(), ctx)
```
misberner, line 351: Note, there also already exists a generic function for stuff like this, not SAC-specific:
```suggestion
		ctx := contextutil.WithValuesFrom(context.Background(), ctx)
```
from `pkg/contextutil`

misberner (approved):

central/policy/service/service_impl.go
connorgorman: 😍 thx will use it

misberner (approved):

PR #3036, opened 2022-09-09
ROX-11101: Remove deprecated resources

rhybrillou:
## Description

The set resources used for access control is being simplified. A number of resources has been deprecated.

The goal here is to get rid of the deprecated resources, both in the database and in the codebase (UI, QA, central, DB). This PR focuses on the database content cleanup. Full PR chain:
https://github.com/stackrox/stackrox/pull/3036
https://github.com/stackrox/stackrox/pull/3115
https://github.com/stackrox/stackrox/pull/3116
https://github.com/stackrox/stackrox/pull/3428

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

If any of these don't apply, please comment below.

## Testing Performed

Added automated tests for the migration.

Manual testing
- Deploy locally a 3.72.0 `central` in RocksDB mode.
- Manually create permission sets using the UI.
- Create permission sets using the API.
example:
```
curl --insecure -XPOST -u admin:$admin_password https://localhost:8000/v1/permissionsets \
--data '{"id":"","name":"OneAccessSubAPIPermissionSet", "description":"", "resourceToAccess":{"AuthProvider": "READ_ACCESS"}}'
```
- Check in the UI all created permission sets are present and contain the expected resource access.
- Check using the API that the created permission sets are present and contain the expected resource access.
example:
```
curl --insecure -XGET -u admin:$admin_password https://localhost:8000/v1/permissionsets
```
- Update the image tag in the deployment to `3.72.x-533-g1614bca3ad`.
- Check the migrator run in the central deployment logs, make sure the migration from sequence number `110` to `111` happened, and that the database sequence number at the end is at least `111`.
```
% kubectl logs -n stackrox deploy/central
No certificates found in /usr/local/share/ca-certificates
No certificates found in /etc/pki/injected-ca-trust
Migrator: 2022/11/07 12:44:40.057224 log.go:13: Info: In migrator.run()
[...]

PR #3115, opened 2022-09-16
ROX-11101: Remove deprecated resources from central

rhybrillou:
## Description

The set resources used for access control is being simplified. A number of resources has been deprecated.

The goal here is to get rid of the deprecated resources, both in the database and in the codebase (UI, QA, central, DB). This PR focuses on the central codebase removal. Full PR chain:
https://github.com/stackrox/stackrox/pull/3036
https://github.com/stackrox/stackrox/pull/3115
https://github.com/stackrox/stackrox/pull/3116
https://github.com/stackrox/stackrox/pull/3428

## Checklist

## Testing Performed

CI should be sufficient


central/role/datastore/singleton.go
```diff
@@ -98,12 +98,12 @@ var defaultRoles = map[string]roleAttributes{
...
-			permissions.View(resources.AuthProvider),
+			permissions.View(resources.Access),
 			permissions.View(resources.Cluster),
 			permissions.View(resources.Namespace),
-			permissions.View(resources.Role),
-			permissions.Modify(resources.Role),
-			permissions.View(resources.AuthProvider),
+			permissions.View(resources.Access),
+			permissions.Modify(resources.Access),
+			permissions.View(resources.Access),
```
rukletsov, line 106: Please de-dupe, one `permissions.View(resources.Access),` is enough : )

central/integrationhealth/datastore/datastore_impl.go
```diff
@@ -12,9 +12,9 @@ import (
 )
 
 var (
-	imageSAC    = sac.ForResource(resources.ImageIntegration)
-	notifierSAC = sac.ForResource(resources.Notifier)
-	backupSAC   = sac.ForResource(resources.BackupPlugins)
+	imageSAC    = sac.ForResource(resources.Integration)
+	notifierSAC = sac.ForResource(resources.Integration)
+	backupSAC   = sac.ForResource(resources.Integration)
```
rukletsov, line 17: I think we can now have a single var and simplify `readAllowed()` and `writeAllowed()`.

central/graphql/resolvers/resolver.go
```diff
@@ -180,36 +180,36 @@ func New() *Resolver {
...
 	readK8sRoleBindings                  = readAuth(resources.K8sRoleBinding)
 	readK8sSubjects                      = readAuth(resources.K8sSubject)
-	readRisks                            = readAuth(resources.Risk)
-	readRoles                            = readAuth(resources.Role)
+	readRisks                            = readAuth(resources.DeploymentExtension)
+	readRoles                            = readAuth(resources.Access)
 	readSecrets                          = readAuth(resources.Secret)
 	readServiceAccounts                  = readAuth(resources.ServiceAccount)
-	readBaselines                        = readAuth(resources.ProcessWhitelist)
+	readBaselines                        = readAuth(resources.DeploymentExtension)
```
rukletsov, line 205: Please de-dupe the list.

rukletsov (changes requested):

tools/generate-helpers/pg-table-bindings/list.go
```diff
@@ -17,6 +17,7 @@ var typeRegistry = make(map[string]string)
 func init() {
 	for s, r := range map[proto.Message]permissions.ResourceHandle{
 		&storage.ActiveComponent{}:                              resources.Deployment,
+		&storage.AuthProvider{}:                                 resources.Access,
```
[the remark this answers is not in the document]
rhybrillou: This list is the `Proto -> scoping resource` mapping. In order to simplify things, the default is to map `&storage.ObjType{}` to `resources.ObjType`.
The additions in this map are actually equivalent to:
```
- &storage.SomeType{}: resources.SomeType,
+ &storage.SomeType{}: resources.NewScopingType,
```
The additions being done here should only have Protos matching the deprecated resource type names, and now map to the replacing resource types.

central/role/resources/list.go
```diff
@@ -74,53 +74,8 @@ var (
 	// To-be-deprecated resources. For now, the resources should still be used in favor of the newly
 	// introduced ones.
```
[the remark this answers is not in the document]
rhybrillou: It does make sense, despite the fact that there still is a deprecated resource: the `ClusterCVE` one, which should be removed in 3.74.

central/role/default.go
```diff
@@ -78,7 +78,7 @@ func IsDefaultAccessScope(id string) bool {
 func GetAnalystPermissions() []permissions.ResourceWithAccess {
 	resourceToAccess := resources.AllResourcesViewPermissions()
 	for i, resourceWithAccess := range resourceToAccess {
-		if resourceWithAccess.Resource.GetResource() == resources.DebugLogs.GetResource() {
+		if resourceWithAccess.Resource.GetResource() == resources.Administration.GetResource() {
```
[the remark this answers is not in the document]
rhybrillou: Good point.

central/role/datastore/singleton.go
rhybrillou: That is a good point. I had it on the radar.
There are a few cases where there were dupes before the change. Considering the order, I wonder if some resources were meant to be writeable. Now this is not the point of the change at hand. I'll do a deduplication pass (here, on the graphql resolver side, on some store cases where the grouped permission simplifies the access control, and on the existing unit test cases where test logic for the replacing resources was added).

central/clusterinit/backend/backend_test.go
```diff
@@ -390,16 +390,16 @@ func (s *clusterInitBackendTestSuite) TestCheckAccess() {
 		shouldFail  bool
 		expectedErr error
 	}{
-		"read access to both ServiceIdentity and APIToken should allow read access": {
+		"read access to both ServiceIdentity and Integration should allow read access": {
```
dhaus67, line 393: same applies for other test cases
```suggestion
		"read access to both Administration and Integration should allow read access": {
```

dhaus67 (commented):
Feel like we should rename the variables we create for datastores such as 
```
signatureSAC = sac.ForResource(resources.Integration)
```
as well to match the resource its checking. I find it a bit confusing seeing a different resource there than what the name of the variable indicates (`signature != integration`)

central/role/default.go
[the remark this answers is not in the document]
rhybrillou: Suggested changelog entry:
```
### Technical Changes
- ROX-11101: The Analyst permission set which used to have read access on all permissions except the now 
deprecated `DebugLogs` permission now has read access to all permissions except `Administration`.
```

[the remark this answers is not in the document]
rhybrillou: Now, the deprecation of `DebugLogs` for `Administration` is only announced in https://github.com/stackrox/stackrox/pull/3036 (was unfortunately missed in 3.71) which means this is a sample to be used in the 3.75 pass on deprecated resources.

rukletsov:
@md2119 that's the second PR in the chain, see https://github.com/stackrox/stackrox/pull/3036

rukletsov (approved):

tools/generate-helpers/pg-table-bindings/list.go
```diff
@@ -16,50 +16,70 @@ var typeRegistry = make(map[string]string)
...
-		&storage.ImageComponentEdge{}:                           resources.Image,
-		&storage.ImageCVE{}:                                     resources.Image,
-		&storage.ImageCVEEdge{}:                                 resources.Image,
-		&storage.K8SRoleBinding{}:                               resources.K8sRoleBinding,
-		&storage.K8SRole{}:                                      resources.K8sRole,
-		&storage.LogImbue{}:                                     resources.DebugLogs,
-		&storage.NamespaceMetadata{}:                            resources.Namespace,
-		&storage.NetworkEntity{}:                                resources.NetworkGraph,
-		&storage.NetworkFlow{}:                                  resources.NetworkGraph,
+		&storage.ActiveComponent{}:                      resources.Deployment,
```
md2119, line 19: Maybe a comment on the top to keep this list sorted?

md2119 (approved):
Changes LGTM. 

Do SREs need (read-only/read-write) `Administration` access?

dhaus67 (approved):

PR #3121, opened 2022-09-16
ROX-10206: Use modified alerts option map on Postgres to fix autocomplete

connorgorman:
## Description

This fixes _some_ autocomplete on alerts. I have another PR that fixes it more generically for different types. But autocomplete on alerts were failing due to field path differences between alert and list alert. This resolves them by using the alert option path

I filed [ROX-12673](https://issues.redhat.com/browse/ROX-12673) for Post-MVP

## Checklist

## Testing Performed

CI covers lots of alert searching


proto/storage/alert.proto
```diff
@@ -126,7 +126,7 @@ message Alert {
     repeated Violation        violations        = 5  [(gogoproto.moretags) = 'sensorhash:"ignore" search:"-"'];
     ProcessViolation          process_violation = 13 [(gogoproto.moretags) = 'search:"-"'];
 
-    Enforcement               enforcement       = 6 [(gogoproto.moretags) = 'search:"Enforcement"'];
+    Enforcement               enforcement       = 6;
```
connorgorman, line 129: Enforcement is a message so this is actually unsupported and caused the search walker to crash

rhybrillou (approved):

central/alert/mappings/options.go
```diff
@@ -3,10 +3,28 @@ package mappings
 import (
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/search"
 )
 
 // TODO: [ROX-10206] Reconcile storage.ListAlert search terms with storage.Alert
```
dashrews78, line 10: nit:  remove?

dashrews78 (approved):

central/alert/mappings/options.go
connorgorman, line 10:
```suggestion

```

PR #3122, opened 2022-09-16
ROX-12566: Proof-of-concept URL-driven network graph

vjwilson:
## Description

****_Never merge!_****

Reduced test case for StackRox "classic" Search query as cluster/namespace selectors for Network Graph

To try out, deploy and visit `<hostname>//main/network-graph`

1. On first load, Cluster Select is populated with clusters names, none selected
2. On first load, Namespace Select is empty
3. Select a Cluster, and its ID will appear as part of the search string in the URL, and its namespaces will be loaded as options in the Namespace Select
4. Select one or more namespaces, and those selected names will be added to the search string in the URL
5. With cluster and namespace options in the URL, refresh the page and watch the controls pre-populate with those selections

Screen Shot 2022-09-16 at 6 27 07 PM

Screen Shot 2022-09-16 at 6 27 15 PM

Screen Shot 2022-09-16 at 6 27 27 PM

Screen Shot 2022-09-16 at 6 27 36 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -2,8 +2,8 @@ import React from 'react';
 import { Flex, FlexItem, PageSection } from '@patternfly/react-core';
 
 import useURLSearch from 'hooks/useURLSearch';
-import { fetchClustersAsArray } from 'services/ClustersService';
-import { Cluster } from 'types/cluster.proto';
+import useClusters from './useClusters';
+import useNamespaces from './useNamespaces';
 import ClusterSelect from './ClusterSelect';
 import NamespaceSelect from './NamespaceSelect';
 
@@ -15,20 +15,16 @@ export type Namespace = {
 };
 
 function NetworkGraphPage() {
-    const [clusters, setClusters] = React.useState<Cluster[]>([]);
     const { searchFilter, setSearchFilter } = useURLSearch();
     const selectedClusterId = (searchFilter.Cluster as string) || '';
     const selectedNamespaces = (searchFilter.Namespace as string[]) || [];
 
-    React.useEffect(() => {
-        fetchClustersAsArray()
-            .then((data) => {
-                setClusters(data as Cluster[]);
-            })
-            .catch(() => {
-                // TODO
-            });
-    }, []);
+    const { clusters, error: clusterError, isLoading: isLoadingCluster } = useClusters();
+    const {
+        namespaces,
+        error: namespaceError,
+        loading: isLoadingNamespace,
+    } = useNamespaces(selectedClusterId);
 
     function updateSelectedClusterId(selection) {
         const newSearchFiliter = { ...searchFilter, Cluster: selection };
@@ -49,11 +45,13 @@ function NetworkGraphPage() {
                         clusters={clusters}
                         selectedClusterId={selectedClusterId}
                         setSelectedClusterId={updateSelectedClusterId}
+                        isLoading={isLoadingCluster}
+                        error={clusterError}
                     />
                 </FlexItem>
                 <FlexItem>
                     <NamespaceSelect
-                        selectedClusterId={selectedClusterId}
+                        namespaces={namespaces}
                         selectedNamespaces={selectedNamespaces}
                         setSelectedNamespaces={updateSelectedNamespaces}
                     />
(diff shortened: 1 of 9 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+import NamespaceSelect from './NamespaceSelect';
+
+export type Namespace = {
+    metadata: {
+        id: string;
+        name: string;
+    };
+};
+
+function NetworkGraphPage() {
```
sachaudh, line 17: Overall, the URL-driven Network Graph looks good. I can select a cluster and namespace and see the URL change

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+        fetchClustersAsArray()
+            .then((data) => {
+                setClusters(data as Cluster[]);
+            })
+            .catch(() => {
+                // TODO
+            });
+    }, []);
+
+    function updateSelectedClusterId(selection) {
```
sachaudh, line 33: Since we only have one cluster, I was wondering what would happen if we did the following:
1. Selected cluster `A`
2. Selected namespace `1`, `2`, `3`
3. Selected a different cluster `B`

I think I would expect the `selectedNamespaces` to be reset too. From the code it looks like the Namespace Select will fetch the namespaces for the new cluster, but I think the selection would still be for the previous namespaces

sachaudh: Looking ahead, I think we'd have to make sure we only show the graph if both the cluster and namespaces are selected. If only the cluster is selected or we pick a new cluster (clearing the namespaces as a result), we wouldn't show the graph + side panel. I think we already do that in the current network graph though

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <Flex>
+                <FlexItem>
+                    <ClusterSelect
+                        clusters={clusters}
+                        selectedClusterId={selectedClusterId}
+                        setSelectedClusterId={updateSelectedClusterId}
+                    />
+                </FlexItem>
+                <FlexItem>
+                    <NamespaceSelect
```
sachaudh, line 55: This Namespace Select component seems to do `data fetching` + `rendering` while the Cluster Select component only does `rendering` and does `data fetching` in this parent component. I think maybe we should try to keep the data fetching within this level and transform the data as necessary to be rendered in these Select components. I was thinking these Select components could even be a little more generic so that we can reuse them for all the other selects we'll need to have for this feature.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+            .then((data) => {
+                setClusters(data as Cluster[]);
+            })
+            .catch(() => {
+                // TODO
+            });
+    }, []);
+
+    function updateSelectedClusterId(selection) {
+        const newSearchFiliter = { ...searchFilter, Cluster: selection };
```
sachaudh, line 34: I think there might have been a typo here for `newSearchFilter`. Same a few lines below

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+    };
+};
+
+function NetworkGraphPage() {
+    const [clusters, setClusters] = React.useState<Cluster[]>([]);
+    const { searchFilter, setSearchFilter } = useURLSearch();
+    const selectedClusterId = (searchFilter.Cluster as string) || '';
+    const selectedNamespaces = (searchFilter.Namespace as string[]) || [];
+
+    React.useEffect(() => {
```
sachaudh, line 23: Just to make things cleaner and consistent, we could have hooks to get the clusters and namespaces within this component.

```
const availableClusters = useFetchClusterNames();
const availableNamespaces = useFetchNamespaceNames({ clusterId });
```

and then render the select components like the following:

```
<ClusterSelect selected={selectedClusterId} onSelect={setSelectedClusterId}>
    {availableClusters.map(cluster => {
          return (
              <SelectOption key={cluster.id} value={cluster.id}>{cluster.name}</SelectOption>
          );
    })}
</ClusterSelect>
```
Or something like that. Just throwing out ideas

PR #3124, opened 2022-09-17
ROX-11592: Remove support to retrieve groups without ID

dhaus67:
## Description

Remove support to retrieve groups without specifying an ID.

This was kept previously to not introduce a breaking change when re-working the groups API. Previously, groups where identified via a composite key of `auth provider ID, key, value`.
With the 3.71 release, a new field was introduced within the group properties called `id` which aims to uniquely identify a group.

Since the grace period has been reached, we can now remove the existing code that allowed retrieving the groups without specifying the id field.

## Testing Performed

- see unit tests and E2E tests passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,13 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Technical Changes
 - ROX-11937: The Splunk integration now processes all additional standards of the compliance operator (ocp4-cis & ocp4-cis-node) correctly.
 
+### Removed Features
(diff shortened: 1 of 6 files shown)


dhaus67:
* master
  * **PR #3124** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3124" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3124?utm_source=stack-comment).

rukletsov (approved):

theencee (approved):
Would be good to add a test (to datastore) that fails on Get if ID is not present (assuming the test doesn't already exist because we had to account for the backwards compatible functionality previously)

PR #3126, opened 2022-09-19
ROX-12709: Use mirrored golang in CI and document mirroring.

porridge:
## Description

[Turns out](https://coreos.slack.com/archives/CBN38N3MW/p1663231189451949) our hacks for overriding golang image didn't really work.
Use a different approach and mostly document the necessary steps for bumping golang image.

## Checklist

## Testing Performed

Ran the `make build/Dockerfile.gen` command manually with and without the variable to make sure it behaves as expected.
Relying on CI otherwise.

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -3,17 +3,24 @@ ARG roxpath=/workspace/src/github.com/stackrox/rox
 
 # NOTE 1: the following version needs to be in sync with the ../EXPECTED_GO_VERSION file.
 FROM golang:1.17.12 as builder
-# NOTE 2: Also, to keep CI from flakes, the above image needs to be mirrored into openshift CI registry and the CI
-# configured to use the mirrored image instead of the specification above.
-# This is because pulling from docker.io/library/... (which is what happens with a bare image specification above) is
-# subject to rate limits that our CI tends to hit, causing intermittent build failures.
+# NOTE 2: Also, the above image MUST be mirrored into openshift CI registry.
 #
-# Ideally, the mirroring and overrides for new image should be put in place first, and then once the above line is
-# updated, the old ones removed. But with some luck, doing it post-facto might also be possible without causing too
-# many failures, depending on the load.
-# See example PRs for how to achieve that:
-# - https://github.com/openshift/release/pull/31790/files
-# - https://github.com/openshift/release/pull/31791/files
+# Why?
+# This is because pulling from docker.io/library/... (which is what happens with a bare image specification above) is
+# subject to rate limits that our CI tends to hit, causing intermittent build failures. To prevent that, this file
+# processed in CI at build time to prefix the image specification with the CI registry.
+# See .openshift-ci/build/build-operator-artifacts.sh
+#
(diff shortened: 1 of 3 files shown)


porridge:
@msugakov in @gavin-stackrox 's absence, perhaps you know how the mirroring works (see the added TODO)?

operator/Dockerfile
```diff
@@ -3,17 +3,24 @@ ARG roxpath=/workspace/src/github.com/stackrox/rox
 
 # NOTE 1: the following version needs to be in sync with the ../EXPECTED_GO_VERSION file.
 FROM golang:1.17.12 as builder
```
msugakov, line 5: I suppose all issue is around pull throttling for docker.io/library/golang, right? Why not use quay.io/stackrox-io/apollo-ci which should be already used in CI and virtually shouldn't need mirroring?

porridge: @msugakov this is an excellent question, to which I do not have an answer. I know however, that **all** current uses of `apollo-ci` in this repo are from the mirror on `registry.ci.openshift.org`. So I think there is a reason. I bet @gavin-stackrox knows, but he's away this week.

msugakov: Well, you can always try and see if it works. I kicked off a build https://github.com/stackrox/stackrox/pull/3137

msugakov: The operator build succeeded there https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3137/pull-ci-stackrox-stackrox-master-images/1572130327395045376 and the image got pushed to `quay.io/rhacs-eng/stackrox-operator:3.72.0-94-g1a6a49c906`. I'd say that's the way to go instead of trying to find mirroring that works.

porridge: While that might be a good long-term goal, there are some open questions:
- I might be wrong, but we don't seem to rely on the apollo-ci images for builds that happen outside CI, so this is setting a new precedent
- [one of the e2e jobs failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3137/pull-ci-stackrox-stackrox-master-openshift-newest-operator-e2e-tests/1572130329634803712) and I think this requires some investigation
- the check job failed and fixing it requires some thinking

While this PR works now, and will help us avoid flakes should the OSCI worker nodes be replaced (or otherwise lose their image cache).

msugakov: Re (1): local builds depend on it already, see [Makefile](https://github.com/stackrox/stackrox/blob/1d9e588e4383067278be987dad599bd5dafc5f1c/Makefile#L66), but fortunately it's freely pullable. Would this change your mind?

porridge: I took the liberty of pushing a couple of commits to your branch to address (3) and I have to admit the diffstat looks tempting.
Let's see how that goes.

porridge:
Superseded by https://github.com/stackrox/stackrox/pull/3137

PR #3127, opened 2022-09-19
ROX-12569 Change local-sensor cluster id

lvalerom:
## Description

`sensor-integration-test` were tracing the following due to invalid `cluster_id`:
```
kubernetes/listener/resources: 2022/09/12 14:13:45.462640 convert.go:48: Error: uuid: incorrect UUID length 4 in string "1234"
kubernetes/listener/resources: 2022/09/12 14:13:45.486336 convert.go:48: Error: uuid: incorrect UUID length 4 in string "1234"
kubernetes/listener/resources: 2022/09/12 14:13:46.459133 convert.go:48: Error: uuid: incorrect UUID length 4 in string "1234"
```

## Checklist

## Testing Performed

Manual testing and CI:

```
go test -race -count=1 github.com/stackrox/rox/sensor/tests/replay -v
```
```
make sensor-integration-test
```


msugakov (approved):
The change looks ok.
You may want to calm down GitGuardian that complains about finding secrets in `.pem` files, not sure how it's done.
Also, I have no clue how to capture data for `safety-net-*` files and I wasn't able to spot instructions. Could you please link it somewhere visibly or write it down if it is not yet written?

sensor/tests/replay/replay_test.go
```diff
@@ -57,7 +57,7 @@ func (suite *ReplayEventsSuite) SetupSuite() {
 		panic(err)
 	}
 	suite.fakeCentral = centralDebug.MakeFakeCentralWithInitialMessages(
-		message.SensorHello("1234"),
+		message.SensorHello("12345678-1234-1234-1234-123456789abc"),
```
vikin91, line 60: This is not valid UUIDv4. Here is the validation rule that I found after quick googling: https://stackoverflow.com/a/19989922

Maybe let's use something like this?

```suggestion
		message.SensorHello("00000000-0000-4000-A000-000000000000"),
```

vikin91: ... and apply to all 3 occurrences

lvalerom: Fair. I'll change it. It's interesting though,`pkg/uuid.go` is not complaining with `12345678-1234-1234-1234-123456789abc` (haven't look at the code)

lvalerom:
> You may want to calm down GitGuardian that complains about finding secrets in `.pem` files, not sure how it's done.

@gavin-stackrox Do you know how to do this? We use this secrets for test purposes only.

> I have no clue how to capture data for `safety-net-*` files and I wasn't able to spot instructions. Could you please link it somewhere visibly or write it down if it is not yet written?

@msugakov The instruction are in this PR #2473. I think it's a good idea to have them written down somewhere else. Maybe in a comment in the tests code?

msugakov:
I think a README.md in `sensor/tests/replay` may be even better.

fredrb (approved):
I'm fine with this change. But for the record, I think a better way of addressing this is to mock the TLS configuration for local-sensor, as described here: https://issues.redhat.com/browse/ROX-11015

The only reason local-sensor needs these files is to extract the ID from the cert CN. The gRPC connection is being mocked anyway so the contents of the certificate are irrelevant.

lvalerom:
> I think a better way of addressing this is to mock the TLS configuration for local-sensor, as described here: https://issues.redhat.com/browse/ROX-11015

I agree with you, ROX-11015 should be the way to go to fix this properly. However, ROX-11015 might end up being a rabbit hole and this PR gives a quick patch to keep the log from spamming those uuid errors 🙂

PR #2799, opened 2022-08-23
Add triggerScan call to configmanagement integration tests which depend on it

pedrottimark:
## Description

13 tests in 5 files under configmanagement fail intermittently since cypress 10.6.0 upgrade, for example:

* 1561658257045458944 from master build for #2743 on 2022-08-22
* 1561807085534973952 from master build for #2781 on 2022-08-23
* 1562047103482466304 from master build for #2745 on 2022-08-23

### Failing tests

1. configmanagement/entitiespage.test.js
    * should not modify the URL when clicking the Overview tab when in the Overview section: Config Management Entity Page should not modify the URL when clicking the Overview tab when in the Overview section
2. configmanagement/dashboard.test.js
    * clicking the "CIS Standard Across Clusters" widget's "passing controls" link should take you to the controls list and filter by passing controls: Config Management Dashboard Page
    * clicking the "CIS Standard Across Clusters" widget's "failing controls" link should take you to the controls list and filter by failing controls: Config Management Dashboard Page
3. configmanagement/clusters.test.js
    * should click on the controls link in the clusters list and open the side panel with the controls list: Config Management Entities (Clusters) should click on the controls link in the clusters list and open the side panel with the controls list
4. configmanagement/ciscontrols.test.js
    * should render the controls list and open the side panel when a row is clicked: Config Management Entities (CIS controls) should render the controls list and open the side panel when a row is clicked
    * should take you to a control single when the "navigate away" button is clicked: Config Management Entities (CIS controls) should take you to a control single when the "navigate away" button is clicked
    * should have the correct count widgets for a single entity view: Config Management Entities (CIS controls) should have the correct count widgets for a single entity view
    * should click on the nodes count widget in the entity page and show the nodes tab: Config Management Entities (CIS controls) should click on the nodes count widget in the entity page and show the nodes tab
    * should have the correct tabs for a single entity view: Config Management Entities (CIS controls) should have the correct tabs for a single entity view
    * should have the same number of Nodes in the count widget as in the Nodes table: Config Management Entities (CIS controls) should have the same number of Nodes in the count widget as in the Nodes table
    * should show no failing nodes in the control findings section of a passing control: Config Management Entities (CIS controls) should show no failing nodes in the control findings section of a passing control
    * should show failing nodes in the control findings section of a failing control: Config Management Entities (CIS controls) should show failing nodes in the control findings section of a failing control
5. configmanagement/nodes.test.js
[...]

PR #3131, opened 2022-09-19
Replace endpoint address constants with helper functions in vulnmanagement integration tests

pedrottimark:
## Description

Encapsulate endpoints to reduce the sources of supporting information needed to read or write tests.

Through the experience of fixing 2 minor mistakes that I made in the first draft, I do see a benefit to my understanding from explicitly declaring entity keys for each interaction within a test.

1. Factor out multiple or parallel references in tests as helper functions which have semantic names:
    * `visitVulnerabilityManagementEntityInSidePanel`
    * `interactAndWaitForVulnerabilityManagementEntity`
    * `interactAndWaitForVulnerabilityManagementEntities`
    * `interactAndWaitForVulnerabilityManagementSecondaryEntities`

2. If only one test requires some container-specific information, consider pro and con to write it as inline string in the test, and then factor it out later if there is need to reuse it in other tests.
    * For example, `api.vulnMgmt.graphqlOps.getFixableCvesForEntity` has only 1 reference.

3. Delete a few redundant tests which had been skipped for a long time.

### Changed files

1. Edit cypress/constants/apiEndpoints.js
    * Delete `vulnMgmt` object.
    * Move `opnameForEntity`, `opnameForEntities`, `opnamePrefixForPrimaryAndSecondaryEntities` objects to entities.js file.

2. Edit cypress/constants/VulnManagementPage.js
    * Delete `sidepanel` property of `url` object.
    * Delete `applicationAndInfrastructureDropdown` property of `dashboardSelectors` object.

3. Edit cypress/helpers/vulnmanagement/entities.js
    * Move `opnameForEntity`, `opnameForEntities`, `opnamePrefixForPrimaryAndSecondaryEntities` objects from apiEndpoints.js file.
    * Rename `keyForEntity` as `typeOfEntity` because it is the type in workflow state.
    * Add helper functions.

4. Edit cypress/integration/vulnmanagement/dashboard.test.js
    * Factor out local helper functions for menu items.
    * Call `interactAndWaitForVulnerabilityManagementEntities` function to wait on relevant request for entities list.
    * Move `pathname` assertions into helper functions.
    * Replace method chains with pseudo-selectors to reduce risk that `cy.get` gets DOM elements too soon, and therefore delete `timeout` options.
    * Replace `parseInt` function calls with `:contains(value)` pseudo-selector to reduce risk that `cy.get` gets DOM elements too soon.
    * Replace skip classic test for CVEs button with 3 tests for CVEs drop-down.
    * Delete `'should properly navigate to the policies list'` test which is redundant with the first test.
    * Add test for Top Risky Nodes.

5. Edit cypress/integration/vulnmanagement/entitypages.test.js
    * Call helper functions to encapsulate intercept-visit-wait sandwich and optional static response for request.
    * Replace `parseInt` function calls with `:contains(value)` pseudo-selector to reduce risk that `cy.get` gets DOM elements too soon.
    * Delete skipped `'should filter deployment count in failing policies section in namespace findings by namespace'` test because it is redundant with a test in namespacesListPages.test.js file.
    * Delete skipped `'should filter component count in images list and image overview by cve when coming from cve list'` test because it is redundant with tests in imageCvesListPages.test.js file.
    * Replace `api.vulnMgmt.graphqlOps.getFixableCvesForEntity` with inline string.

6. Edit cypress/integration/vulnmanagement/imageSection/imageOverview.test.js
    * Get entity `id` from fixture file.
    * Call `visitVulnerabilityManagementEntityInSidePanel` function with static response argument for body from fixture file.

7. Edit cypress/integration/vulnmanagement/nodeSection/nodeOverview.test.js
    * Get entity `id` from fixture file.
    * Call `visitVulnerabilityManagementEntityInSidePanel` function with static response argument for body from fixture file.

## Checklist

## Testing Performed

Ran changed tests one at a time in local deployment and used inspector with stagingdb for tests which fail because no vulnerabilities:
* vulnmanagement/dashboard.test.js
* vulnmanagement/entitypages.test.js
* vulnmanagement/imageSection/imageOverview.test.js
* vulnmanagement/nodeSection/nodeOverview.test.js


pedrottimark:
The following unrelated test failure looks like my bad editing cypress/helpers/compliance.js in #2799

> Timed out retrying after 4050ms: expected '<button.flex.items-center.justify-center.border-2.btn.btn-base.h-10.uppercase.lg:min-w-32.xl:min-w-43.text-base-100.hover:bg-primary-800>' not to have attribute 'disabled'
compliance-triggerScan

There was a second click of the **Scan** button, which makes no sense. Apparently a copy-paste error.

ui/apps/platform/cypress/integration/vulnmanagement/dashboard.test.js
```diff
@@ -1,122 +1,188 @@
...
-                    .invoke('text')
-                    .then((panelHeaderText) => {
-                        expect(parseInt(panelHeaderText, 10)).to.equal(parseInt(numCves, 10));
-                    });
-            });
+    // TODO Delete skip when we delete ROX_FRONTEND_VM_UPDATES feature flag.
+
+    it.skip('should show same number of Image CVEs in menu item and entities list', () => {
+        verifyVulnerabilityManagementDashboardCVEs('image-cves', /^\d+ Image CVEs?$/);
+    });
```
vjwilson, line 93: You abstracted the setup into a function `verifyVulnerabilityManagementDashboardCVEs()`, but did you mean to leave all three of these tests that now use it `skip`'ed ?

vjwilson (approved):
Just one question about some still-skipped tests.

ui/apps/platform/cypress/integration/vulnmanagement/dashboard.test.js
pedrottimark: Good eyes. I tested them locally and decided on temporary skip instead of feature flag logic.

When we turn on the feature flag for nightly and release builds, we can remove the skip.

PR #3132, opened 2022-09-19
Dashrews/ROX-12547 migration remove certain foreign keys

dashrews78:
## Description
Encountered issue running migrations where a deployment had been deleted, but that deletion hadn't bee propagated to the active components yet, so that migration failed. so we need to disable the triggers on the table while processing the migration. I evaluated the custom migrations and only the active components migration had an external foreign key. So I left the other custom migrations alone.

Initially looked into tackling this at a migration level. That required turning off constraints or triggers and then re-enabling them after the fact.  Since the initial migration is everything form RocksDB to Postgres, we determined there was risk associated with the various ways we could accomplish that.

So after discussions we decided that in order to make the migration process safer in the RocksDB -> Postgres migration, we would remove the foreign keys in active_component and cluster_health_status because the key reaches outside the confines of their stores.  To alleviate the problem with inconsistent data, we added a prune for these to eliminate the out of sync data after the migration.  The inconsistent data is an issue today.  ROX-12710 and ROX-12711 have been filed to utilize the primary keys after we get rid of RocksDB and have more confidence in the data consistency and integrity.  That will also allow us to do this sort of thing on smaller more easily tested migrations than the big bang of RocksDB->Postgres.

Between this ticket and ROX-12710 we will make ROX-11763 obsolete.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


central/pruning/pruning.go
```diff
@@ -806,6 +809,27 @@ func (g *garbageCollectorImpl) removeRisks(riskType storage.RiskSubjectType, ids
 	}
 }
 
+func (g *garbageCollectorImpl) removeOrphanedComponents() {
```
dashrews78, line 812: seemed like overkill to push this down to the store/datastore.

central/pruning/pruning.go
```diff
@@ -806,6 +809,27 @@ func (g *garbageCollectorImpl) removeRisks(riskType storage.RiskSubjectType, ids
...
+	if features.PostgresDatastore.Enabled() {
+		pruneStmt := `DELETE FROM active_components child WHERE NOT EXISTS
```
c-du, line 814: It would be better to move these to central/activecomponent/datastore with other Postgres operations.

central/pruning/pruning.go
```diff
@@ -806,6 +809,27 @@ func (g *garbageCollectorImpl) removeRisks(riskType storage.RiskSubjectType, ids
...
+		if _, err := globaldb.GetPostgres().Exec(pruningCtx, pruneStmt); err != nil {
+			log.Error(err)
+		}
+	}
+}
+
+func (g *garbageCollectorImpl) removeOrphanedStatuses() {
+	if features.PostgresDatastore.Enabled() {
+		pruneStmt := `DELETE FROM cluster_health_statuses child WHERE NOT EXISTS
+		(SELECT * FROM clusters parent WHERE
```
c-du, line 825: Dido.

central/pruning/pruning.go
dashrews78: Yeah I suppose that is the correct thing to do, just more cumbersome to get there.

central/pruning/pruning.go
```diff
@@ -810,23 +813,14 @@ func (g *garbageCollectorImpl) removeRisks(riskType storage.RiskSubjectType, ids
 }
 
 func (g *garbageCollectorImpl) removeOrphanedComponents() {
```
c-du, line 815: nit: removeOrphanedActiveComponents

c-du (approved):

connorgorman (changes requested):
Find this a bit hard to review due to extraneous changes that are just for running postgres. if you set ROX_POSTGRES_DATASTORE=true when you deploy you don't need to comment out all the usages of it. Also, we should not write sql in datastores. If we want to provide pruning functions then I think we should separate these out into a specific package where we are explicitly calling raw SQL. Something like central/postgres/pruning.go where then we add explicit functions that do this pruning. Also, I prefer that the prune function itself is wrapped by the postgres flag instead of being within the function because it seems like we would be pruning active components when we're not in RocksDB. This begs the question, if we're not pruning in RocksDB then I would have expected it to increase forever?

dashrews78:
> Find this a bit hard to review due to extraneous changes that are just for running postgres. if you set ROX_POSTGRES_DATASTORE=true when you deploy you don't need to comment out all the usages of it. Also, we should not write sql in datastores. If we want to provide pruning functions then I think we should separate these out into a specific package where we are explicitly calling raw SQL. Something like central/postgres/pruning.go where then we add explicit functions that do this pruning. Also, I prefer that the prune function itself is wrapped by the postgres flag instead of being within the function because it seems like we would be pruning active components when we're not in RocksDB. This begs the question, if we're not pruning in RocksDB then I would have expected it to increase forever?

The extra stuff was for starting a Postgres so I can easily start with Rocks and then migrate to postgres.  Of course it was not meant to be checked in, I was testing something and got bit by `git add -u`

central/postgres/pruning.go
```diff
@@ -0,0 +1,37 @@
...
+	if _, err := globaldb.GetPostgres().Exec(ctx, pruneActiveComponentStmt); err != nil {
+		log.Error(err)
+	}
+}
+
+// PruneClusterHealthStatus - prunes cluster health statuses
+// TODO (ROX-12711):  This will no longer be necessary when the foreign keys are added back
+func PruneClusterHealthStatus(ctx context.Context) {
+	if _, err := globaldb.GetPostgres().Exec(ctx, pruneClusterHealthStatusStmt); err != nil {
+		log.Error(err)
```
connorgorman, line 35: Add context to what this is err is? The postgres errors can be vague. like "failed to prune cluster health statuses: %v"

central/postgres/pruning.go
```diff
@@ -0,0 +1,37 @@
...
+
+var (
+	log = logging.LoggerForModule()
+)
+
+// PruneActiveComponents - prunes active components
+// TODO (ROX-12710):  This will no longer be necessary when the foreign keys are added back
+func PruneActiveComponents(ctx context.Context) {
+	if _, err := globaldb.GetPostgres().Exec(ctx, pruneActiveComponentStmt); err != nil {
+		log.Error(err)
```
connorgorman, line 27: Same comment as below about context to the error

central/postgres/pruning.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+// PruneClusterHealthStatus - prunes cluster health statuses
+// TODO (ROX-12711):  This will no longer be necessary when the foreign keys are added back
+func PruneClusterHealthStatus(ctx context.Context) {
+	if _, err := globaldb.GetPostgres().Exec(ctx, pruneClusterHealthStatusStmt); err != nil {
```
connorgorman, line 34: Plural

central/postgres/pruning.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+
+	"github.com/stackrox/rox/central/globaldb"
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+const (
+	pruneActiveComponentStmt = `DELETE FROM active_components child WHERE NOT EXISTS
+		(SELECT * from deployments parent WHERE child.deploymentid = parent.id)`
```
connorgorman, line 12: select 1

central/postgres/pruning.go
```diff
@@ -0,0 +1,37 @@
...
+
+	pruneClusterHealthStatusStmt = `DELETE FROM cluster_health_statuses child WHERE NOT EXISTS
+		(SELECT * FROM clusters parent WHERE
```
connorgorman, line 15: select 1

central/pruning/pruning.go
```diff
@@ -134,6 +135,10 @@ func (g *garbageCollectorImpl) pruneBasedOnConfig() {
 	g.removeOrphanedResources()
 	g.removeOrphanedRisks()
 	g.removeExpiredVulnRequests()
+	if features.PostgresDatastore.Enabled() {
+		postgres.PruneActiveComponents(pruningCtx)
+		postgres.PruneClusterHealthStatus(pruningCtx)
```
connorgorman, line 140: I think maybe we should move this below the removal of decommissioned clusters?

connorgorman (approved):
LG, small comments. thanks for the restructure

proto/storage/cluster.proto
```diff
@@ -258,7 +258,7 @@ message AuditLogFileState {
 }
 
 message ClusterHealthStatus {
-    string                    id                               = 9 [(gogoproto.moretags) = 'sql:"pk,fk(Cluster:id)"'];
+    string                    id                               = 9 [(gogoproto.moretags) = 'sql:"pk,fk(Cluster:id),no-fk-constraint"'];
```
janisz, line 261: @rhybrillou Do you think it has any impact on SAC?

PR #3134, opened 2022-09-19
Fix hotreload script by deriving arch

connorgorman:
## Description

Can't always be amd64 for hotreload

## Checklist

## Testing Performed

Ran on my mac

### DIFF
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -54,7 +54,7 @@ function hotload_binary {
   echo "**********"
   echo
 
-  binary_path=$(realpath "$(git rev-parse --show-toplevel)/bin/linux_amd64/${local_name}")
+  binary_path=$(realpath "$(git rev-parse --show-toplevel)/bin/linux_$(uname -m)/${local_name}")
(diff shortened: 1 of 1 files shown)


deploy/common/k8sbased.sh
```diff
@@ -54,7 +54,12 @@ function hotload_binary {
   echo "**********"
   echo
 
-  binary_path=$(realpath "$(git rev-parse --show-toplevel)/bin/linux_amd64/${local_name}")
+  arch="amd64"
+  if [[ "$(uname -m)" == "arm64" ]]; then
+    arch="arm64"
+  fi
```
misberner, line 60: Tbh, I get super confused with respect to where to use which arch flavor (`uname -m` vs `$GOARCH`). I would prefer if we only had a single source of truth, e.g., by putting a
```
.PHONY: print-target-arch
print-target-arch:
	@echo "$(TARGET_ARCH)"
```
rule in the `Makefile`, and then rewriting this as
```suggestion
  arch="$(make --quiet --no-print-directory -C "$(git rev-parse --show-toplevel)" print-target-arch)"
```
or maybe even `print-target-bin-dir` or so

connorgorman: yeah, I think this would be way better

PR #3136, opened 2022-09-19
ROX-12611: Add support to Postgres store generator for ON DELETE RESTRICT

keyallis:
## Description

Add support to the postgres store generator to support a protobuf tag denoting a delete restriction for foreign key references.

This is necessary due to the fact that collections can have references to other collections and we want to prevent the deletion of nested collections that are in use. Errors encountered by attempted deletions will be percolated up to the UI to inform users that a given collection is a dependency of another collection.

## Checklist

## Testing Performed

Locally added the flag to a proto definition, built an image, and verified that the table had the correct configuration.

### DIFF
--- a/pkg/postgres/walker/schema.go
+++ b/pkg/postgres/walker/schema.go
@@ -65,6 +65,9 @@ type SchemaRelationship struct {
 	// NoConstraint indicates that this relationship is not enforced at the SQL
 	// level by a foreign key constraint.
 	NoConstraint bool
+
+	// DeleteRestrict indicates that this relationship should restrict deletion
+	DeleteRestrict bool
 }
 
 // ThisSchemaColumnNames generates the sequence of column names for this schema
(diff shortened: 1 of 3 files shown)


pkg/postgres/walker/schema.go
```diff
@@ -65,6 +65,9 @@ type SchemaRelationship struct {
 	// NoConstraint indicates that this relationship is not enforced at the SQL
 	// level by a foreign key constraint.
 	NoConstraint bool
+
+	// DeleteRestrict indicates that this relationship should restrict deletion
+	DeleteRestrict bool
```
connorgorman, line 70: Nit: I would flip the naming here. RestrictDelete seems more apt

connorgorman (approved):

pkg/postgres/walker/schema.go
keyallis: Renamed the field to bring it in line with this and reverified that the behavior is the same

PR #3139, opened 2022-09-20
Use a `http.Server` with a timeout

0x656b694d:
## Description

`gosec` reports *G114: Use of net/http serve function that has no support for setting timeouts*.

(Some versions of `gosec` would actually crash trying to report that.)

## Checklist

## Testing Performed
```
$ gosec webhookserver
[gosec] 2022/09/20 10:32:51 Including rules: default
[gosec] 2022/09/20 10:32:51 Excluding rules: default
[gosec] 2022/09/20 10:32:51 Import directory: /home/mipetrov/go/src/github.com/stackrox/stackrox/webhookserver
[gosec] 2022/09/20 10:32:51 Checking package: main
[gosec] 2022/09/20 10:32:51 Checking file: /home/mipetrov/go/src/github.com/stackrox/stackrox/webhookserver/server.go
Results:


Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 99
  Nosec  : 0
  Issues : 0
```

### DIFF
--- a/webhookserver/server.go
+++ b/webhookserver/server.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"time"
 
 	"github.com/stackrox/rox/pkg/sync"
 )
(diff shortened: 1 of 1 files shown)


misberner:
I guess this change doesn't hurt, but just wanted to remark the `webhookserver` is test-only software (run during e2e tests) and thus probably not super interesting from a security standpoint.

0x656b694d:
> I guess this change doesn't hurt, but just wanted to remark the `webhookserver` is test-only software (run during e2e tests) and thus probably not super interesting from a security standpoint.

Right, thank you Malte. This fix will remove the only appearance of G114 from gosec report, which is just nice to get rid of.

vikin91 (approved):

PR #3141, opened 2022-09-20
ROX-12692: Remove pods/attach resource from admission controller resources, and fail open on unsupported request kinds/api verbs

misberner:
## Description

The admission controller configuration specifies that we want to listen on `pods/attach` resource access. However, [we don't actually handle that](https://github.com/stackrox/stackrox/blob/master/pkg/kubernetes/event.go#L45).

This PR does two things: first, it removes the `pods/attach` resource from the admission controller configuration. This will not cause anything to break, as we're not handling pod attach events anyway, but it will ensure customers `kubectl attach` calls will not break when they have configured enforcement for K8s events. Second, it makes sure we simply log an error and fail open if we receive a request that we should not receive (per `ValidatingWebhookConfiguration`), instead of blocking the request.
 
## Checklist
- [x] Determined and documented upgrade steps: https://docs.engineering.redhat.com/display/StackRox/Upgrade+to+73.0

## Testing Performed

TBD

### DIFF
--- a/image/templates/helm/stackrox-secured-cluster/templates/admission-controller.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/admission-controller.yaml
@@ -247,7 +247,6 @@ webhooks:
         - CONNECT
         resources:
           - pods
-          - pods/attach
           - pods/exec
           - pods/portforward
     failurePolicy: Ignore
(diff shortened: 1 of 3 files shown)


misberner:
@md2119 do you have any historic insight as to why `pods/attach` was added here?

md2119 (approved):

misberner:
Test failures are known flakes

PR #3148, opened 2022-09-20
[ROX-12294] : Disable debug dump and diagnostic bundle endpoints when ROX_MANAGED_CENTRAL env var is set to true

charmik-redhat:
## Description

Disabled debug dump and diagnostic bundle for Managed services preview

## Checklist

## Testing Performed

Get debug dump and diagnostic bundle with ROX_MANAGED_CENTRAL env set and verify that they fail.

### DIFF
--- a/central/auth/userpass/auth.go
+++ b/central/auth/userpass/auth.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/authproviders"
 	basicAuthProvider "github.com/stackrox/rox/pkg/auth/authproviders/basic"
-	"github.com/stackrox/rox/pkg/buildinfo"
 	"github.com/stackrox/rox/pkg/env"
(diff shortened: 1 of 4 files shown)


misberner (changes requested):
This will certainly get the job done, but I am very honestly not a fan of having managed services check logic pop up in fairly random places. If we have to change behavior, I would much rather expose disabling debug info as a more general-purpose knob rather than making this behavior depend on managed services setting.

We can set arbitrary env vars differently for managed centrals, so we can just introduce a boolean setting `ROX_DISABLE_DEBUG_LOGS`. WDYT?

misberner:
See https://github.com/stackrox/stackrox/pull/3024#discussion_r967130719 for related thoughts.

Also, is there a plan to remove RBAC bindings in a follow-up?

charmik-redhat:
Closing this PR for now since we'll be releasing MS service preview with v3.72. For limited availability, I will figure out how to keep parts of diagnostics and exclude just the managed central's cluster specific info.

PR #3151, opened 2022-09-21
build hygiene: Get cgo flags for rocksdb from explicit env vars

misberner:
## Description

Rather than assuming the build container has `CGO_CFLAGS` and `CGO_LDFLAGS` set, explicitly augment those env vars by appending the contents of `$ROCKSDB_CFLAGS` and `$ROCKSDB_LDFLAGS` to them for the binaries that need rocksdb.

This change is designed such that it can go in immediately, without waiting for a change of the builder image to set those vars.

## Checklist

## Testing Performed

- CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,9 +202,12 @@ ci-config-validate:
 	circleci config validate --org-slug gh/stackrox
 
 .PHONY: fast-central-build
-fast-central-build:
+fast-central-build: central-build-nodeps
+
+.PHONY: central-build-nodeps
+central-build-nodeps:
 	@echo "+ $@"
-	$(GOBUILD) central
+	CGO_CFLAGS="$(CGO_CFLAGS) $(ROCKSDB_CFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS) $(ROCKSDB_LDFLAGS)" $(GOBUILD) central
 
 .PHONY: fast-central
 fast-central: deps
@@ -230,9 +233,12 @@ fast-migrator:
 	docker run --rm $(GOPATH_WD_OVERRIDES) $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) make fast-migrator-build
 
 .PHONY: fast-migrator-build
-fast-migrator-build:
+fast-migrator-build: migrator-build-nodeps
+
+.PHONY: migrator-build-nodeps
+migrator-build-nodeps:
 	@echo "+ $@"
-	$(GOBUILD) migrator
+	CGO_CFLAGS="$(CGO_CFLAGS) $(ROCKSDB_CFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS) $(ROCKSDB_LDFLAGS)" $(GOBUILD) migrator
 
 .PHONY: check-service-protos
 check-service-protos:
@@ -444,8 +450,8 @@ else
 endif
 
 .PHONY: main-build-nodeps
-main-build-nodeps:
-	$(GOBUILD) central migrator sensor/kubernetes sensor/admission-control compliance/collection
+main-build-nodeps: central-build-nodeps migrator-build-nodeps
+	$(GOBUILD) sensor/kubernetes sensor/admission-control compliance/collection
 	CGO_ENABLED=0 $(GOBUILD) sensor/upgrader
 ifndef CI
     CGO_ENABLED=0 $(GOBUILD) roxctl


Makefile
```diff
@@ -202,9 +202,12 @@ ci-config-validate:
...
 
 .PHONY: fast-central-build
-fast-central-build:
+fast-central-build: central-build-nodeps
+
+.PHONY: central-build-nodeps
+central-build-nodeps:
 	@echo "+ $@"
-	$(GOBUILD) central
+	CGO_CFLAGS="$(CGO_CFLAGS) $(ROCKSDB_CFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS) $(ROCKSDB_LDFLAGS)" $(GOBUILD) central
```
msugakov, line 210: Would it make sense to define `ROCKSDB_CFLAGS` and `ROCKSDB_LDFLAGS` in the same `Makefile` so that both upstream and downstream share these settings?

msugakov:
Link to downstream change for the record keeping https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/203

Makefile
misberner: I had thought about this (and at first was confused why I didn't find mentions of those env vars in all of the stackrox repo), but thinking about this, location of shared objects and include headers is inherently system specific, so it needs to be in the respective Dockerfile. Put it differently, we generally do not want to depend on our upstream and downstream build containers to use an identical layout/package paths (it's quite possible that we do depend on this already, but I don't want to further this)

msugakov: I think, it should be treated as a feature, not a bug: we should strive to have upstream and downstream containers as similar as possible so that once the code is tested upstream, the level of certainty about downstream is also higher.

`ROCKSDB_*FLAGS` could impact local development that's different from ubi container but it seems that these makefile targets are used only in containerized builds so I would guess it shouldn't break local development, especially given that we don't launch `central` on MacOS directly.

I am not strongly attached to what I propose and the change looks good so I would approve it if I wasn't able to change your mind. One thing: you'd probably need to update `BUILD_IMAGE_VERSION` to use the one built in https://github.com/stackrox/rox-ci-image/pull/167 for testing this change in CI.

misberner: I guess for me the decoupling is a feature, not a bug :) However, I think you (inadvertently?) managed to convince me that this is the wrong approach.

In cgo builds, one can actually specify the `CFLAGS`, `CXXFLAGS`, and `LDFLAGS` directly in the Go source code via a [`// #cgo` comment](https://pkg.go.dev/cmd/cgo#hdr-Using_cgo_with_the_go_command). Interestingly, the rocksdb library we use already does that, in our current dependency version, `dynflag.go` looks like this:
```
// +build !linux !static

package gorocksdb

// #cgo LDFLAGS: -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy -llz4 -lzstd -ldl
import "C"
```
This is almost the same as what we define:
```
ENV CGO_CFLAGS="-I/lib/rocksdb/include"
ENV CGO_LDFLAGS="-L/lib/rocksdb -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy -llz4 -lzstd -ljemalloc"
```
(caveat: a newer version of gorocksdb drops the compression libraries, one has to add them manually. I think we only need `lz4`)

I conclude that we can drop _most of_ the `LDFLAGS` (be it `CGO_` or `ROCKSDB_`) because it's already defined by the `tecbot/rocksdb` library (`-ljemalloc` will have to be reverted anyway).

As for the `-L` and `-I` options, I would still say that they probably should still be set in the container, and they can remain as `CGO_CFLAGS` and `CGO_LDFLAGS`. They just define where stuff is _on the system_, they do not however affect any builds.

So my course of action now would be:
- Change the `rox-ci-image` and downstream PR to only set `-L` and `-I` in `LDFLAGS`/`CFLAGS` and see if that works
- Disregard this PR, or maybe just reduce it to the `Makefile` refactorings. Or _maybe_ augment it to add `-llz4` to the `LDFLAGS` in `pkg/rocksdb/rocksdb.go`, because that is where we select the compression option.

Thoughts?

msugakov: I would suggest to re-state all rocksdb's `LDFLAGS` (except of `-L`) that we depend on in our code. This way accidental changes in `rocksdb` version would not affect product functioning. I suppose specifying `LDFLAGS` in `pkg/rocksdb/rocksdb.go` will work due to this

> All the `LDFLAGS` directives in **any** package in the program are concatenated and used at link time.

----

> As for the -L and -I options, I would still say that they probably should still be set in the container, and they can remain as CGO_CFLAGS and CGO_LDFLAGS. They just define where stuff is on the system, they do not however affect any builds.

That's not how I understand it. See https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html:

> CFLAGS
>     Extra flags to give to the C compiler.

> LDFLAGS
>     Extra flags to give to compilers when they are supposed to invoke the linker, ‘ld’, such as -L. Libraries (-lfoo) should be added to the LDLIBS variable instead.

Specifically `-I` in `CGO_CFLAGS="-I/lib/rocksdb/include"` configures the path to header files. These are only used during build, not needed at runtime.

I don't know the behavior of `-L` in `CGO_LDFLAGS="-L/lib/rocksdb"` whether this affects runtime but it definitely matters at the build-time also.

I agree that both `CGO_CFLAGS=-I` and `CGO_LDFLAGS` are specific to the system in a sense that dependencies or options provided in these flags should be present and should be present in the provided paths.

I'm still with the opinion that `-I/lib/rocksdb/include` and `-L/lib/rocksdb` apply only to binaries with rocksdb support and should go as part of commands building that.

----

Why reverting `-ljemalloc`? I thought Connor wanted it to see how memory usage can get improved.

misberner: > I would suggest to re-state all rocksdb's LDFLAGS (except of -L) that we depend on in our code. This way accidental changes in rocksdb version would not affect product functioning.

I'm not sure what you mean by this. Are you suggesting that in addition to `-llz4` I also specify `-lrocksdb -ldl -lm` etc.? I don't think it's a reasonable assumption that changing `rocksdb` version would omit `-lrocksdb` from the `LDFLAGS`. As for other libraries like `-lm -ldl`, I would consider them implementation details. We don't explicitly depend on that functionality, `rocksdb` happens to use and need those libraries. Even if we re-state them explicitly, they could just as well become wrong if we change the `rocksdb` (not `gorocksdb`) version, so I don't think we win anything. I think restating `-llz4` is important and needed, though.

>  These are only used during build, not needed at runtime.

Right. What I meant by "they do not however affect any builds" was that passing `-L..` or `-I...` to a link/compile process does not generally change the build _output_ (unless we shadow a header or library of the same name that would otherwise get included), as opposed to `-l`, which does create references for the linker. Put differently, `-L` typically only plays a role when `-l` is given with a library in that path. Currently, the `ENV CGO_LDFLAGS` causes even things like the `protoc-gen-gofast` tool to be linked against RocksDB. That doesn't make sense, and that's what I want to avoid.

> I'm still with the opinion that -I/lib/rocksdb/include and -L/lib/rocksdb apply only to binaries with rocksdb support and should go as part of commands building that.

I think a counterargument to that is unit tests. When running unit tests in the IDE, we depend on system rocksdb paths. And for people who run unit tests via `Makefile`, we also don't want to prescribe the location of libraries, because these are typically _not_ run in a container and may even be run on non-linux systems (I've run unit tests using rocksdb natively on my Mac). That's the point where I think a general abstraction of possibly heterogeneous build environments is necessary, rather than a streamlined centos-or-ubi environment assumption.

---

Because we would need to jump through a couple of hoops to use `jemalloc` downstream (probably build from source) and given that `rocksdb` is soon-to-be-retired, it was deemed not worth it.

connorgorman (approved):

PR #3152, opened 2022-09-21
Build out service functions for collections API

dvail:
## Description

Implements the front end collection service based on the [backend API design document](https://docs.google.com/document/d/1Yb9lFAZRTYNF-JCt-GbuSidzeI8y3Qf4r3XVkpksDRQ/edit#heading=h.2btpaey9h5q4). The details of these APIs might shift slightly during development, but the overall architecture should remain the same.

## Follow ups
- [Unit test for query param util function](https://github.com/stackrox/stackrox/pull/3155)
- [Refactor existing `Empty` types to use the top level `services/types` export](https://github.com/stackrox/stackrox/pull/3154)
- Add MSW mocks for these endpoints (either in the codebase or externally)

## Checklist

## Testing Performed

Manual testing of Violations page (sorting, searching, pagination) due to minor util function refactor.

### DIFF
--- a/ui/apps/platform/src/services/AlertsService.ts
+++ b/ui/apps/platform/src/services/AlertsService.ts
@@ -3,7 +3,7 @@ import queryString from 'qs';
 import { Alert, ListAlert } from 'Containers/Violations/types/violationTypes';
 
 import { ApiSortOption, SearchFilter } from 'types/search';
-import { getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
+import { getListQueryParams, getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
 import axios from './instance';
 import { CancellableRequest, makeCancellableAxiosRequest } from './cancellationUtils';
 
@@ -83,19 +83,7 @@ export function fetchAlerts(
     page: number,
     pageSize: number
 ): CancellableRequest<ListAlert[]> {
-    const offset = page > 0 ? page * pageSize : 0;
-    const query = getRequestQueryStringForSearchFilter(searchFilter);
-    const params = queryString.stringify(
-        {
-            query,
-            pagination: {
-                offset,
-                limit: pageSize,
-                sortOption,
-            },
-        },
-        { arrayFormat: 'repeat', allowDots: true }
-    );
+    const params = getListQueryParams(searchFilter, sortOption, page, pageSize);
     return makeCancellableAxiosRequest((signal) =>
         axios
             .get<{ alerts: ListAlert[] }>(`${baseUrl}?${params}`, { signal })
--- a/ui/apps/platform/src/services/CollectionsService.ts
+++ b/ui/apps/platform/src/services/CollectionsService.ts
@@ -7,10 +7,10 @@ import { CancellableRequest, makeCancellableAxiosRequest } from './cancellationU
 import axios from './instance';
 import { Empty, Pagination } from './types';
 
-const collectionsBaseUrl = '/v1/collections';
-const collectionsCountUrl = '/v1/collections/count';
-const collectionsDryRunUrl = '/v1/collections/dryrun';
-const collectionsAutocompleteUrl = '/v1/collections/autocomplete';
+export const collectionsBaseUrl = '/v1/collections';
+export const collectionsCountUrl = '/v1/collections/count';
+export const collectionsDryRunUrl = '/v1/collections/dryrun';
+export const collectionsAutocompleteUrl = '/v1/collections/autocomplete';
 
 type CollectionResource = 'Cluster' | 'Namespace' | 'Deployment';
 
@@ -29,14 +29,14 @@ type ResourceSelector = {
     rules: SelectorRule[];
 };
 
-type CollectionRequest = {
+export type CollectionRequest = {
     name: string;
     description: string;
     resourceSelectors: ResourceSelector[];
     embeddedCollectionIds: string[];
 };
 
-type CollectionResponse = {
+export type CollectionResponse = {
     id: string;
     name: string;
     description: string;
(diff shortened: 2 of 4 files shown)


dvail:
* master
  * **PR #3152** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3152" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3152?utm_source=stack-comment).

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -0,0 +1,234 @@
...
+ * @param dryRunRequest.options.skipDeploymentMatching
+ *      This flag will skip the resolution of matching deployments on the back end
+ *      in order to do more efficient error checking. Used in order to determine if
+ *      a config is valid, without returning a full data payload.
+ *
+ * @returns A list of deployments that are resolved by the collection.
+ */
+export function dryRunCollection(
+    dryRunRequest: CollectionDryRunRequest
+    // TODO `ListDeployment` will make this impossible to paginate without loading the entire
```
dvail, line 195: In order to implement the "Selector results" sidebar as designed in the mocks we need to aggregate all clusters and namespaces that match a deployment _name_, under that deployment name. The standard `ListDeployment` type will return individual deployments with a single cluster and namespace, even if there is a matching deployment name in another cluster.

Essentially we need something with a structure similar to the results of this GQL query:
```gql
query q {
  deployments(query: "Deployment:kube") {
    name
    namespaceObject {
      metadata { name }
      cluster { name }
    }
  }
}
```
but with the ability to accept `ResourceSelectors` to filter the results.

The BE team is currently looking to see if this is feasible to do over the `dryRunCollections` REST API.

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -0,0 +1,234 @@
...
+import { CancellableRequest, makeCancellableAxiosRequest } from './cancellationUtils';
+import axios from './instance';
+import { Empty, Pagination } from './types';
+
+export const collectionsBaseUrl = '/v1/collections';
+export const collectionsCountUrl = '/v1/collections/count';
+export const collectionsDryRunUrl = '/v1/collections/dryrun';
+export const collectionsAutocompleteUrl = '/v1/collections/autocomplete';
+
+type CollectionResource = 'Cluster' | 'Namespace' | 'Deployment';
```
pedrottimark, line 15: Remembering the discussion which motivated change from resource sets to collections:

Is it worthwhile to ask whether there is a meaningful alternative to **Resource** in collection API?

https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/

> A resource is an endpoint in the Kubernetes AP that stores a collection of API objects of a certain kind; for example, the built-in pods resource contains a collection of Pod objects.

> A custom resource is an extension of the Kubernetes API that is not necessarily available in a default Kubernetes installation. It represents a customization of a particular Kubernetes installation. However, many core Kubernetes functions are now built using custom resources, making Kubernetes more modular.

pedrottimark: Here is a thought:
1. Replace `CollectionResource` with `SelectorEntityType`
2. Replace `resourceSelectors: ResourceSelector[]` with `collectionSelectors: CollectionSelector[]` which implies discussion with core workflows team

dvail: Sure I can bring this up for discussion during our weekly sync and refactor after if needed.

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -0,0 +1,234 @@
...
+ *      The updated collection object
+ *
+ */
+export function updateCollection(
+    id: string,
+    collection: CollectionRequest
+): CancellableRequest<CollectionResponse> {
+    return makeCancellableAxiosRequest((signal) =>
+        axios
+            .post<CollectionResponse>(`${collectionsBaseUrl}/${id}`, collection, { signal })
```
vjwilson, line 144: Maybe I missed this in the design doc reviews, but why is "update" functionality a `POST`, instead of a `PUT` or a `PATCH` ?

vjwilson (commented):
Looks good overall, just have a couple of questions

ui/apps/platform/src/services/AlertsService.ts
```diff
@@ -3,7 +3,7 @@ import queryString from 'qs';
 import { Alert, ListAlert } from 'Containers/Violations/types/violationTypes';
 
 import { ApiSortOption, SearchFilter } from 'types/search';
-import { getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
+import { getListQueryParams, getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
```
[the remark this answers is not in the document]
dvail: `getRequestQueryStringForSearchFilter` is used in quite a few places, but it looks like in almost all of them it is eventually being put into an object of the form `{ query: <the_query> }` and passed to `qs.stringify()`, much like `getListQueryParams` is doing. There are a couple of places where it is put into an object of the form `{ where: <the_query> }` or passed directly as a GQL variable though. I can investigate and see if there is a way that makes sense to consolidate these as a follow up.

ui/apps/platform/src/services/CollectionsService.ts
dvail: I had noticed this too in the backend design doc, since I initially had a `PUT` here. I brought it up during the Tuesday collections sync and Oscar mentioned that he thought that it was intentional but was going to follow up with the team to find out, along with some other questions.

Do we want to hold off merging for now, or change it later if that backend confirms it should be a different method?

vjwilson: @dvail @keyallis 
Unless there's a obstacle to using the more semantic PUT or PATCH verbs for updates, I think we should get both front and back-end teams aligned on that now.

dvail: For this, since the entire payload is being sent on each request to update all fields of the collection I'd say we go with `PUT`. @keyallis I'll go ahead and update this PR if that works on your end.

PR #3159, opened 2022-09-21
ROX-12627: Add postgres store for collections

keyallis:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


central/resourcecollection/store/store.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Store provides storage functionality for resource collections.
+//
+//go:generate mockgen-wrapper
+type Store interface {
+	Count(ctx context.Context) (int, error)
+	Exists(ctx context.Context, id string) (bool, error)
+	Get(ctx context.Context, id string) (*storage.ResourceCollection, bool, error)
+
+	DryRun(context.Context, *storage.ResourceCollection) error
```
md2119, line 17: `DryRun` should be datastore method, not store.

proto/storage/resource_collection.proto
```diff
@@ -0,0 +1,54 @@
...
+package storage;
+
+message ResourceCollection {
+  message EmbeddedResourceCollection {
+    // this is searchable to force a separate table
+    string id = 1 [(gogoproto.moretags) = 'search:"Embedded Collection ID",hidden'];
+  }
+
+  string id                                                = 1 [(gogoproto.moretags) = 'sql:"pk"']; // primary key
+  string name                                              = 2 [(gogoproto.moretags) = 'search:"Collection Name"'];
```
md2119, line 20: Mark `name` as `unique`. We must enforce Collections to have unique names otherwise the workflow would become utterly confusing.

proto/storage/resource_collection.proto
```diff
@@ -0,0 +1,54 @@
...
+  string description                                       = 3;
+  google.protobuf.Timestamp created_at                     = 4;
+  google.protobuf.Timestamp last_updated                   = 5;
+  SlimUser created_by                                      = 6;
+  SlimUser updated_by                                      = 7;
+  // `resource_selectors` resolve as disjunction (OR) with each-other and with selectors from `embedded_collections`. For MVP, the size of resource_selectors will at most be 1 from UX standpoint.
+  repeated ResourceSelector resource_selectors             = 8;
+  repeated EmbeddedResourceCollection embedded_collections = 9 [(gogoproto.moretags) = 'sql:"fk(EmbeddedResourceCollection:id)"'];
```
md2119, line 28: Entire message cannot be a foreign key. A message is composite type and can have multiple fields. Move the constraint to scalar field `id`.

proto/storage/resource_collection.proto
```diff
@@ -0,0 +1,54 @@
...
+  // - Cluster
+  // - Cluster Label
+  // - Namespace
+  // - Namespace Label
+  // - Namespace Annotation
+  // - Deployment
+  // - Deployment Label
+  // - Deployment Annotation
+  string field_name         = 1;
+  BooleanOperator operator  = 2;
```
md2119, line 47: Add a comment on this that only `OR` is supported for now.

tools/generate-helpers/pg-table-bindings/list.go
```diff
@@ -54,6 +54,7 @@ func init() {
 		&storage.ProcessBaselineResults{}:                       resources.ProcessWhitelist,
 		&storage.ProcessBaseline{}:                              resources.ProcessWhitelist,
 		&storage.ProcessIndicator{}:                             resources.Indicator,
+		&storage.ResourceCollection{}:                           resources.Image, // TODO change this
```
md2119, line 57: Recommend adding  `Workflow Administration` resource first and putting the correct value here from the very beginning.

tools/generate-helpers/pg-table-bindings/resource_type_test.go
```diff
@@ -77,6 +77,7 @@ func TestGetResourceType(t *testing.T) {
 		{typ: &storage.K8SRole{}, resourceType: directlyScoped},
 		{typ: &storage.K8SRoleBinding{}, resourceType: directlyScoped},
 		{typ: &storage.ReportConfiguration{}, resourceType: globallyScoped},
+		{typ: &storage.ResourceCollection{}, resourceType: indirectlyScoped},
```
md2119, line 80: Resource Collection is globally scoped. From design document:

> Customers can use it to control which users are allowed to create, update and view collections. This resource is a globally scoped resource.

proto/storage/resource_collection.proto
keyallis: Do you mean for this comment to be on the `operator` field? That's where I added it. The field below also has some kind of comment about `AND` and `OR` that came from the design doc that probably should get cleaned up

proto/storage/resource_collection.proto
```diff
@@ -0,0 +1,55 @@
...
+  // `field_name` can be one of the following:
+  // - Cluster
+  // - Cluster Label
+  // - Namespace
+  // - Namespace Label
+  // - Namespace Annotation
+  // - Deployment
+  // - Deployment Label
+  // - Deployment Annotation
+  string field_name         = 1;
```
theencee, line 46: Any reason this isn't an enum?

keyallis: From direct messages with Mandar where I asked this of him:
Define the field_names as global string consts. It just keeps it the model flexible. TAL at features/list.go or pkg/booleanpolicy/fieldnames/list.go. The ListCollectionSelectors endpoint must serve these values.

md2119 (approved):
Approving to unblock, but address self reference. If the code generator does not handle it. Add a JIRA TODO.

PR #3163, opened 2022-09-21
Enable setting extra arch labels for operator

yselkowitz:
## Description

These labels are required in order to display an operator in the OperatorHub on architectures other than x86_64 (amd64).

## Checklist

### DIFF
--- a/operator/bundle_helpers/patch-csv.py
+++ b/operator/bundle_helpers/patch-csv.py
@@ -47,7 +47,7 @@ def must_replace_suffix(str, suffix, replacement):
     return splits[0] + replacement
 
 
-def patch_csv(csv_doc, version, operator_image, first_version, no_related_images, rbac_proxy_replacements):
+def patch_csv(csv_doc, version, operator_image, first_version, no_related_images, extra_supported_arches, rbac_proxy_replacements):
     csv_doc['metadata']['annotations']['createdAt'] = datetime.now(timezone.utc).isoformat()
 
     placeholder_image = csv_doc['metadata']['annotations']['containerImage']
@@ -68,6 +68,10 @@ def patch_csv(csv_doc, version, operator_image, first_version, no_related_images
     first_x, first_y, first_z = (int(c) for c in first_version.split('-', maxsplit=1)[0].split('.'))
     # An olm.skipRange doesn't hurt if it references non-existing versions.
     csv_doc["metadata"]["annotations"]["olm.skipRange"] = f'>= {x}.{y-1}.0 < {version}'
+
+    # multi-arch
+    for arch in extra_supported_arches:
+        csv_doc["metadata"]["labels"][f"operatorframework.io/arch.{arch}"] = "supported"
 
     if (x, y, z) > (first_x, first_y, first_z):
         if z == 0:
@@ -91,6 +95,9 @@ def parse_args():
     parser.add_argument("--replace-rbac-proxy", required=False, metavar='original-tag:replacement-image',
                         nargs='+', help='Replacement directives for the RBAC proxy image',
                         default=[])
+    parser.add_argument("--add-supported-arch", action='append', required=False,
+                        help='Add label to CSV for an additional operator (may be passed multiple times)',
+                        default=[])
     return parser.parse_args()
 
 
@@ -102,6 +109,7 @@ def main():
               version=args.use_version,
               first_version=args.first_version,
               no_related_images=args.no_related_images,
+              extra_supported_arches=args.add_supported_arch,
               rbac_proxy_replacements={
                     tag: img
                     for tag, img in (spec.split(':', maxsplit=1) for spec in args.replace_rbac_proxy)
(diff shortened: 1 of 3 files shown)


operator/bundle_helpers/patch-csv.py
```diff
@@ -47,7 +47,7 @@ def must_replace_suffix(str, suffix, replacement):
     return splits[0] + replacement
 
 
-def patch_csv(csv_doc, version, operator_image, first_version, no_related_images, rbac_proxy_replacements):
+def patch_csv(csv_doc, version, operator_image, first_version, no_related_images, downstream_arches, rbac_proxy_replacements):
```
misberner, line 50: While we use this script downstream, we typically try to keep the entire downstream terminology out of the upstream repo (and should keep doing that). Therefore, I would ask you to avoid (a) a `downstream` variable/parameter/flag name and (b) an assumption about what happens downstream. I leave the changes I would suggest throughout the PR as extra comments.

misberner, line 50:
```suggestion
def patch_csv(csv_doc, version, operator_image, first_version, no_related_images, extra_supported_arches, rbac_proxy_replacements):
```

operator/bundle_helpers/patch-csv.py
```diff
@@ -69,6 +69,11 @@ def patch_csv(csv_doc, version, operator_image, first_version, no_related_images
     # An olm.skipRange doesn't hurt if it references non-existing versions.
     csv_doc["metadata"]["annotations"]["olm.skipRange"] = f'>= {x}.{y-1}.0 < {version}'
 
+    # multi-arch
+    if downstream_arches:
+        csv_doc["metadata"]["labels"]["operatorframework.io/arch.ppc64le"] = "supported"
+        csv_doc["metadata"]["labels"]["operatorframework.io/arch.s390x"] = "supported"
```
misberner, line 75:
```suggestion
    for arch in extra_supported_arches:
        csv_doc["metadata"]["labels"][f"operatorframework.io/arch.{arch}"] = "supported"
```

operator/bundle_helpers/patch-csv.py
```diff
@@ -91,6 +96,8 @@ def parse_args():
     parser.add_argument("--replace-rbac-proxy", required=False, metavar='original-tag:replacement-image',
                         nargs='+', help='Replacement directives for the RBAC proxy image',
                         default=[])
+    parser.add_argument("--downstream-arches", action='store_true',
+                        help='Add labels for downstream architectures')
```
misberner, line 100:
```suggestion
    parser.add_argument("--extra-arch", action='append', required=False,
                        help='Mark the operator as supporting an additional architecture')
```

operator/bundle_helpers/patch-csv.py
```diff
@@ -102,6 +109,7 @@ def main():
               version=args.use_version,
               first_version=args.first_version,
               no_related_images=args.no_related_images,
+              downstream_arches=args.downstream_arches,
```
misberner, line 112:
```suggestion
              extra_supported_arches=args.extra_arch,
```

misberner (changes requested):

operator/config/manifests/bases/rhacs-operator.clusterserviceversion.yaml
```diff
@@ -12,6 +12,8 @@ metadata:
     operatorframework.io/suggested-namespace: rhacs-operator
     operators.openshift.io/infrastructure-features: '["disconnected", "proxy-aware"]'
     support: Red Hat
+  labels:
+    operatorframework.io/arch.amd64: supported
```
msugakov, line 16: Why don't we just add all architectures here unconditionally?

```suggestion
    operatorframework.io/arch.amd64: supported
    operatorframework.io/arch.ppc64le: supported
    operatorframework.io/arch.s390x: supported
```

Yes, the upstream operator does not come built for these architectures (today) but does this make anyone upset?

yselkowitz: This is the other option.  However, it may be confusing, particularly if the community operator CSV gets imported into another index (e.g. OpenShift's community-operator-index).  Since you already have this script to modify the CSV, it doesn't seem hard to handle this as proposed here (style nits aside).

operator/bundle_helpers/patch-csv.py
yselkowitz: Testing showed that this also needs `default=[]` for the case when the flag is not passed.

operator/config/manifests/bases/rhacs-operator.clusterserviceversion.yaml
misberner: I would agree. The operator build process certainly isn't super elegant already, but this seems a fairly needless (as in: doing it at least somewhat better than that seems a fairly low-hanging fruit) breaking of abstraction.

yselkowitz:
This error doesn't seem related to this PR?
```
Resolving definitions.stackrox.io (definitions.stackrox.io)... failed: Name or service not known.
wget: unable to resolve host address 'definitions.stackrox.io' 
```

misberner (approved):

PR #3168, opened 2022-09-22
Build with go1.18.4 by updating CI image version to 0.3.49

misberner:
## Description

Change the builder image to 0.3.49, which uses go1.18.4.

See the comments in the changed files view for an explanation of some of the more surprising changes.

DO NOT MERGE: This PR can only be merged after a builder image change lands on `openshift/release` `main` branch.

## Checklist

## Testing Performed

https://github.com/openshift/release/pull/32511


scripts/ci/lib.sh
```diff
@@ -1119,14 +1119,6 @@ validate_expected_go_version() {
         echo "Got unexpected go version ${roxctl_go_version} (wanted ${expected_go_version})"
         exit 1
     fi
-
```
misberner, line 1122: I don't know why this was added, but I do not think it makes sense. The entry in `go.mod` is the source compatibility version, its only relation to the go version used for building is that the source compatibility level must not be above the builder go version - but in that case, builds would fail anyway.

Whether or not we leave this at 1.17 or set to 1.18 is more similar to whether or not we use generics in the code.

pkg/mtls/testutils/testdata/central-certs/ca-key.pem
```diff
@@ -1,28 +1,52 @@
...
-8B4HhwXGoQKBgQCfPyAGT1mwlP51OuXNOaB+UdabJ5NRYLa0ErG7SQjbrxlR9re1
-0ENUVLEL/AukI6igYCb/BHrpmyaJDaBWobjxkOZd9w/plmyKA1GHILqnqrLir1W0
-V3B4PBB51fdwqdAZEdlgOlGOh60HPf5bm3mPlfRjJtgocmkpSrbiER9BqwKBgAQG
-6tSvVB78M784bGV6mNTrfuZtiPIm/Xm2aH5HY0aGgAxUe68wGU6p4u/mUwBrN/hV
-zsQj26Qu+Olw+8yUpLq+EFt72gR65Gy5Lre0ZjpkxuYpo9Ds4j7PvmEQtYu3pCHC
-QYH3S6IxmNvswWnYcmXVSGGfgqR4eLu+Qag+EXPhAoGAQj89wNgQ4WOZHnJOEORN
-bD7W9bmzBLJGWi8SiQs95ofRLrtS8KUiJuNlVFt0Cczqmmv02ynlPSlr522mHBj/
-QTjv5oXXO0pWZc84oq5kHgvoomu3FxEeuSreVKsQs1QiEhsjecStmTZeplehPkMf
-ty+nEkCChJ3Kj8p37JVPmKY=
+MIIJQgIBADANBgkqhkiG9w0BAQEFAASCCSwwggkoAgEAAoICAQC/Ibsc33Rcg1r3
```
misberner, line 2: These and the following files needed to be changed because go1.18+ will refuse to validate SHA-1 signatures. The generation logic has been adjusted to pass `-sha256` to `openssl x509` in order to fix this.

docs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.35 AS builder
```
RTann, line 1: wow that's old

connorgorman (approved):

pkg/mtls/testutils/testdata/generate-certs.sh
```diff
@@ -10,12 +10,12 @@ root_ca_exts="
   basicConstraints=CA:TRUE,pathlen:0
 "
 
-openssl req -config <(echo "${root_ca_exts}") -nodes -new -x509 -newkey 2048 -keyout ca-key.pem -out ca.pem -days 1825 -subj "/CN=StackRox Certificate Authority"
+openssl req -config <(echo "${root_ca_exts}") -nodes -new -x509 -newkey 4096 -keyout ca-key.pem -out ca.pem -days 1825 -subj "/CN=StackRox Certificate Authority"
```
RTann, line 13: curious why we updated to 4096. Is this using rsa? Also, this is just testdata, so it doesn't matter much, right? Just curious

pkg/mtls/testutils/testdata/generate-certs.sh
```diff
@@ -10,12 +10,12 @@ root_ca_exts="
...
 
 
 leaf_ca_exts="subjectAltName=DNS:central.stackrox"
-openssl genrsa -out leaf-key.pem 2048
+openssl genrsa -out leaf-key.pem 4096
```
RTann, line 17: ah guess this answers the rsa question

scripts/ci/lib.sh
connorgorman: Yes, I actually do not like this generally speaking

RTann (approved):
do we not want to bump the go.mod version as well?

misberner:
The main reason for not bumping `go.mod` version is the fact that the process of updating the CI image is so complicated that I really want to keep the changes minimal. Not bumping it ensures that we will break no single developer (this breakage could easily be remediated by upgrading to go1.18, but still). The bump can happen when the CI setup is in order.

PR #3180, opened 2022-09-23
Add operator architectures explicitly in Makefile

msugakov:
## Description

This follows up on #3163. That PR introduces some code which is for downstream to enable multiple architectures there. The issue is the code is never called upstream and so we're risking it gets rusty or may even get removed as unnecessary.
Having a code that always works is better than having conditional code that works rarely.

Also couple other tiny maintenance changes (see commit descriptions).

## Checklist

## Testing Performed

- Hoping for CI to cover it.
- Ran `make bundle bundle-build` locally and checked tat `build/bundle/manifests/rhacs-operator.clusterserviceversion.yaml` has `amd64` architecture label.

### DIFF
--- a/operator/bundle_helpers/patch-csv.py
+++ b/operator/bundle_helpers/patch-csv.py
@@ -98,7 +98,7 @@ def parse_args():
                         nargs='+', help='Replacement directives for the RBAC proxy image',
                         default=[])
     parser.add_argument("--add-supported-arch", action='append', required=False,
-                        help='Enable additional operator architecture via CSV labels (may be passed multiple times)',
+                        help='Enable specified operator architecture via CSV labels (may be passed multiple times)',
                         default=[])
     return parser.parse_args()
 
--- a/operator/bundle_helpers/patch-csv.py
+++ b/operator/bundle_helpers/patch-csv.py
@@ -70,6 +70,8 @@ def patch_csv(csv_doc, version, operator_image, first_version, no_related_images
     csv_doc["metadata"]["annotations"]["olm.skipRange"] = f'>= {x}.{y-1}.0 < {version}'
 
     # multi-arch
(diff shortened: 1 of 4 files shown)


misberner (approved):
👍

operator/bundle_helpers/patch-csv.py
```diff
@@ -70,6 +70,8 @@ def patch_csv(csv_doc, version, operator_image, first_version, no_related_images
     csv_doc["metadata"]["annotations"]["olm.skipRange"] = f'>= {x}.{y-1}.0 < {version}'
 
     # multi-arch
+    if "labels" not in csv_doc["metadata"]:
+        csv_doc["metadata"]["labels"] = {}
```
yselkowitz, line 74: Why is this necessary?

operator/bundle_helpers/patch-csv.py
```diff
@@ -96,7 +98,7 @@ def parse_args():
                         nargs='+', help='Replacement directives for the RBAC proxy image',
                         default=[])
     parser.add_argument("--add-supported-arch", action='append', required=False,
-                        help='Add label to CSV for an additional operator (may be passed multiple times)',
+                        help='Enable additional operator architecture via CSV labels (may be passed multiple times)',
```
yselkowitz, line 101: "additional" doesn't seem to make sense if we're starting without any.  (Note that if no such labels are declared, then `amd64` is implied, but once *any* flags are, only those specified are enabled.)

operator/bundle_helpers/patch-csv.py
msugakov: Because `labels` key does not exist in `metadata` map since I removed default `amd64` there.

yselkowitz (approved):

PR #3173, opened 2022-09-22
Add the ability to specify custom proxy endpoints during development

dvail:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #3157, opened 2022-09-21
MSW endpoint handlers for collections

dvail:
## Description

This adds `msw` and defines mock endpoints for the collections API. This is an alternative approach to the simple solution of [allowing development-time custom proxies](https://github.com/stackrox/stackrox/pull/3173).

Mock service worker can be used to mock endpoint seamlessly in the browser or node by standing up a service worker when enabled that listens for network requests, intercepts them, and returns custom responses. This allows 100% of the front end code to be run without mocking at a higher level (mocking fetch, axios, or service functions).

**Pros**
- Gives us complete control over the endpoints and how they are mocked, almost equivalent to a custom built mocking app
- Keeps the mocks bundled with the code base, making it easier for the team to view and collaborate
- Mocks can be used fairly interchangeably for development, Cypress testing, and Jest testing (although I'm not sure we want or need this)

**Cons**
- Adds new dependencies to the project
- If we don't use mocks for testing, adds code that is destined to be deleted in the near future to the project
- Adds a new file to `/public` that we need to strip out in CI
- Adds code to `index.tsx` (should be tree-shaken, but still)
- Service workers require a valid certificate. (I had to create a local self signed cert, tell webpack to use it, and add it to the Mac OS keychain in order to use this...)

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

In downstream branches, test service calls against the collection endpoints and see the correct results returned.

PR #3155, opened 2022-09-21
Add unit tests for search util getListQueryParamns

dvail:
## Description

Adds tests for util function.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added

## Testing Performed

`yarn test`

PR #3154, opened 2022-09-21
Refactor service api `Record<string, never>` to use exported `Empty` type

dvail:
## Description

Refactors duplicated `Record<string, never>` to use `Empty` type from common service API types file.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

N/A - nonfunctional type-level refactor

PR #3181, opened 2022-09-23
Add basic collections data loading and table display

dvail:
## Description

Adds a bare-bones table for collections display on the landing page.

_As an aside, given the recent UX changes to the "Append collections" functionality, I don't think there is value in making the link behavior, etc. configurable for this component from the start. The new mocks are sufficiently different enough that I think they should be separate components._

**Follow ups**

- Implement sorting by column
- Implement a table toolbar with pagination and filtering
- Implement correct selection handling and delete functionality
- File a PatternFly issue for the cut off checkbox layout in an empty table with the `variant='compact'` prop

## Checklist

## Testing Performed

(The following tests were performed with a mock server running locally to provide data.)

Visit the collection page, and see a spinner while data loads:
image

If an error occurs, display a brief message to the user. (This message is the descriptive message returned by the server.)
image

After loading, the collections data should be displayed in the table:
image
image

If no data is returned by the request, and EmptyState is displayed with a possible user action. (No effect right now).
image

The table menu should display three options, with the "delete" option disabled if the collection is in use:
image
image

Selecting the "edit" option should route the user to the edit page for that collection.( No effect right now other than URL change.)
image

Selecting the "clone" option should route the user to the clone page for that collection.( No effect right now other than URL change.)
image

Clicking on the name of a collection should route the user to that collection's page, by ID. (No effect right now other than URL change.)
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -1,0 +1,133 @@
+import React from 'react';
+import { useHistory } from 'react-router-dom';
+import { Button, ButtonVariant, Truncate, Bullseye, Text } from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+import { TableComposable, TableVariant, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
+
+import EmptyStateTemplate from 'Components/PatternFly/EmptyStateTemplate';
+import LinkShim from 'Components/PatternFly/LinkShim';
+import useTableSelection from 'hooks/useTableSelection';
+import { CollectionResponse } from 'services/CollectionsService';
+import { collectionsPath } from 'routePaths';
+
+export type CollectionsTableProps = {
+    collections: CollectionResponse[];
+    hasWriteAccess: boolean;
+};
+
+function CollectionsTable({ collections, hasWriteAccess }: CollectionsTableProps) {
+    const history = useHistory();
+    const { selected, allRowsSelected, onSelect, onSelectAll } = useTableSelection(collections);
+
+    function onEditCollection(id: string) {
+        history.push({
+            pathname: `${collectionsPath}/${id}`,
+            search: 'action=edit',
+        });
+    }
+
+    function onCloneCollection(id: string) {
+        history.push({
+            pathname: `${collectionsPath}/${id}`,
+            search: 'action=clone',
+        });
+    }
+
+    return (
+        <>
+            <TableComposable variant={TableVariant.compact}>
+                <Thead>
+                    <Tr>
+                        {hasWriteAccess && (
+                            <Th
+                                select={{
+                                    onSelect: onSelectAll,
+                                    isSelected: allRowsSelected,
+                                }}
+                            />
+                        )}
+                        <Th modifier="wrap" width={25}>
+                            Collection
+                        </Th>
+                        <Th modifier="wrap">Description</Th>
+                        <Th modifier="wrap" width={10}>
+                            In use
+                        </Th>
+                        <Th aria-label="Row actions" />
+                    </Tr>
+                </Thead>
+                <Tbody>
+                    {collections.length === 0 && (
+                        <Tr>
+                            <Td colSpan={hasWriteAccess ? 5 : 3}>
+                                <Bullseye>
+                                    <EmptyStateTemplate
+                                        title="No collections found"
+                                        headingLevel="h2"
+                                        icon={SearchIcon}
+                                    >
+                                        <Text>Clear all filters and try again.</Text>
+                                        <Button variant="link" onClick={() => {}}>
+                                            Clear all filters
+                                        </Button>
+                                    </EmptyStateTemplate>
+                                </Bullseye>
+                            </Td>
+                        </Tr>
+                    )}
+                    {collections.map(({ id, name, description, inUse }, rowIndex) => {
+                        const actionItems = [
+                            {
+                                title: 'Edit collection',
+                                onClick: () => onEditCollection(id),
+                            },
+                            {
+                                title: 'Clone collection',
+                                onClick: () => onCloneCollection(id),
+                            },
+                            {
+                                isSeparator: true,
+                            },
+                            {
+                                title: inUse ? 'Cannot delete (in use)' : 'Delete collection',
+                                onClick: () => {},
+                                isDisabled: inUse,
+                            },
+                        ];
+
+                        return (
+                            <Tr key={id}>
+                                {hasWriteAccess && (
+                                    <Td
+                                        select={{
+                                            rowIndex,
+                                            onSelect,
+                                            isSelected: selected[rowIndex],
+                                        }}
+                                    />
+                                )}
+                                <Td dataLabel="Collection">
+                                    <Button
+                                        variant={ButtonVariant.link}
+                                        isInline
+                                        component={LinkShim}
+                                        href={`${collectionsPath}/${id}`}
+                                    >
+                                        {name}
+                                    </Button>
+                                </Td>
+                                <Td dataLabel="Description">
+                                    <Truncate content={description || '-'} tooltipPosition="top" />
+                                </Td>
+                                <Td dataLabel="In Use">{inUse ? 'Yes' : 'No'}</Td>
+                                {hasWriteAccess && <Td actions={{ items: actionItems }} />}
+                            </Tr>
+                        );
+                    })}
+                </Tbody>
+            </TableComposable>
+        </>
+    );
+}
+
+export default CollectionsTable;
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import {
     PageSection,
     Title,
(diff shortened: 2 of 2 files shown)


dvail:
* master
  * **PR #3152** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3152" target="_blank">Graphite</a> 
    * **PR #3173** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3173" target="_blank">Graphite</a> 
      * **PR #3181** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3181" target="_blank">Graphite</a>  👈
    * **PR #3155** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3155" target="_blank">Graphite</a> 
    * **PR #3157** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3157" target="_blank">Graphite</a> 
    * **PR #3154** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3154" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3181?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+                                        href={`${collectionsPath}/${id}`}
+                                    >
+                                        {name}
+                                    </Button>
+                                </Td>
+                                <Td dataLabel="Description">
+                                    <Truncate content={description || '-'} tooltipPosition="top" />
+                                </Td>
+                                <Td dataLabel="In Use">{inUse ? 'Yes' : 'No'}</Td>
+                                {hasWriteAccess && <Td actions={{ items: actionItems }} />}
```
dvail, line 121: Having this `{hasWriteAccess && ` check three times throughout here doesn't feel great, but seems more straightforward than abstracting the permission checks in some way.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+                                    onSelect: onSelectAll,
+                                    isSelected: allRowsSelected,
+                                }}
+                            />
+                        )}
+                        {collectionsTableColumns.map(({ Header, width }) => (
+                            <Th key={Header} modifier="wrap" width={width}>
+                                {Header}
+                            </Th>
+                        ))}
```
pedrottimark, line 54: What do you think about inline `Th` elements parallel to `Td` elements below?

Descriptors seem more abstract than needed here, especially unless we expect conditional rendering.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -0,0 +1,131 @@
...
+                    </Tr>
+                </Thead>
+                <Tbody>
+                    {collections.length === 0 && (
+                        <Tr>
+                            <Td colSpan={8}>
```
pedrottimark, line 60: If I remember correctly, UX team prefers either empty state or table @vjwilson can you comment?

If so, you can lift up `EmptyStateTemplate` to conditional rendering in `CollectionsTablePage`

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
pedrottimark: Yes, it is clear and findable.

This comment reminds me there might be another, for a corresponding `Th` element.
* `<Th aria-label="Row actions" />`
* `Td` might like to nave `className="pf-u-text-align-right"` prop so kabob aligns at the right, in case descriptions are short and page is wide.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
dvail: Good call, this is another artifact from before the recent tweaks to the "Append collections" portion of the mocks when the plan was to make this table more reusable. I'll simplify here.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
dvail: Good catch on the missing `<Th>`. It looks like the description column always fills the available space though, so the kebab is always right aligned. Is there a situation you've seen where the kebab does not right align?

pedrottimark: Oh, good. Aha, description does not have width. Maybe because I specified width for every column.

janisz:
@dvail Why did you requested a review from stackrox/merlin team?

dvail:
@janisz I didn't request a review myself. I'm guessing the GitGuardian security checks added the merlin team automatically?

Nothing related to certificates was changed in this PR but it seems to be triggering on commits https://github.com/stackrox/stackrox/commit/2de65f09db1a67bd73025a584c340e0f9df92b91 and https://github.com/stackrox/stackrox/commit/fbb2196e997764ef8506a57132d0a93d6bf3af12 for some reason.

I'll remove the merlin team from review.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
vjwilson: I believe that is correct. Let's just mention that we're doing it in the UX-UI joint meeting.

dvail: I went with this approach since this table will have a toolbar in a follow up PR that can be used to filter the results. Looking into this more though, this is a good one to bring up in the joint meeting today since behavior is fairly inconsistent in other places:

- It looks like the CVE table in Vuln Mgmt uses the same approach here (EmptyState in place of `<tbody>`)
- Violations and Policies simply leave the table empty, with headers/toolbar intact
- Integrations and Access Control replace the entire table with an EmptyState (these pages do not have a table toolbar though)
- Risk Acceptance is the same as Integrations/Access Control, but _does_ have a toolbar with filtering, which leads to an inconsistent state. I'll file a bug for this one.

- [ ] Create ticket for Risk Acceptance after guidance from UX

dvail: Created the following related issues after meeting with UX:
https://issues.redhat.com/browse/ROX-12821
https://issues.redhat.com/browse/ROX-12822

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+                            />
+                        )}
+                        <Th modifier="wrap" width={25}>
+                            Collection
+                        </Th>
+                        <Th modifier="wrap">Description</Th>
+                        <Th modifier="wrap" width={10}>
+                            In use
+                        </Th>
+                        <Th aria-label="Row actions" />
```
pedrottimark, line 56: Conditional rendering for **Row actions** column head like corresponding data cell?

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+                    </Tr>
+                </Thead>
+                <Tbody>
+                    {collections.length === 0 && (
```
pedrottimark, line 60: Commented a few lines upstream to keep separate from thread below.

What are pro and con to replace “8 is enough” (could not resist the allusion) with something like:

```jsx
<Td colSpan={hasWriteAccess ? 3 : 5}>
```

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
```diff
@@ -7,17 +7,66 @@ import {
...
+        []
+    );
+    const { data: listData, loading: listLoading, error: listError } = useRestQuery(listQuery);
+
+    const countQuery = useCallback(() => getCollectionCount({}), []);
+    const { data: countData, loading: countLoading, error: countError } = useRestQuery(countQuery);
+
+    const isDataAvailable = typeof listData !== 'undefined' && typeof countData !== 'undefined';
+    const isLoading = !isDataAvailable && (listLoading || countLoading);
+    const loadError = listError || countError;
```
pedrottimark, line 39: A question to learn from your experience instead of a change request.

What do you think are pro and con to distinguish errors from primary and secondary requests?
Because I have not yet written a table which has PatternFly pagination, I am not sure how robust it would be to render if list requests succeeds and count request fails.

pedrottimark (approved):
2 questions might might motivate small changes plus 1 bonus for my learning.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
dvail: The only real con I can think of the possibility that a change down the line updates the number of columns and the empty state isn't well tested, leading to a mis-aligned UI.

According to the spec, this could be any value up to `1000` which we could use to simulate "full width". I did see some results online where this led to a massive page slow down in some versions of Firefox. Probably very unlikely, but I think I'd prefer an off-center error state to a page breaking edge case.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
dvail: I tried this originally, but applied to both sides (the checkbox header cell and the row actions cell). I kept these in since the removal of the first cell caused some layout shift as data appeared/disappeared. It seemed semantically equivalent to keep these headers in since we still display the "collection/description/in use" headers despite there being no values in the table.

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
dvail: That is a good question. The count data eventually needs to get passed into the PatternFly `<Pagination>` component, but we could probably fallback and do something more graceful than placing the entire section into an "error" state.

It looks like the Pagination `itemCount` prop is optional, which gives us the "of 0" UI:
image
image

We could also pass `Infinity` to get a similar behavior:
image
image

Note that in both these cases the pagination still works correctly, with the exception of allowing the user to "overpage".

PF also provides a `toggleTemplate` option we could use, if we really wanted to be robust, which lets us specify what is shown in the case of a `NaN` `itemCount`:
```typescript
toggleTemplate={({ firstIndex, lastIndex, itemCount, ofWord }) => {
  // When itemCount is not an integer
  // TODO Implement happy path
  return (
    <span>
      <b>{firstIndex}</b>-<b>{lastIndex}</b> {ofWord}{' '}
      <b>unknown</b>
    </span>
  );
}}
```
image

If this is something we wanted to do, it might actually be one of the cases where we would want a wrapper around a PF component so that we could get this on other pages that also use PF pagination, since it would be a bit verbose to write it out each time.

pedrottimark: Yeah, allowing overpage seems like a more serious problem, therefore leave well enough alone for now.

Although there is a prop to customize the count with text, it seems much cost for small benefit:
https://www.patternfly.org/v4/components/pagination#indeterminate

PR #3184, opened 2022-09-23
pass on check_docs failure

davdhacs:
## Description

allow push_images.sh check_docs() call to pass on tags not matching


0x656b694d (approved):
I'm not sure why did you choose to go the fancy way instead of just commenting out the `check_docs`, but LGTM.

PR #3196, opened 2022-09-26
go 1.18 follow up

0x656b694d:
## Description

Bump go version for static check.

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -91,7 +91,7 @@ linters-settings:
     - 'os\.Stderr(# Disallowed output streams used\. Use environment\.InputOutput\(\).Out instead\.)?'
     - 'os\.Stdin(# Disallowed output streams used\. Use environment\.InputOutput\(\).Out instead\.)?'
   staticcheck:
-    go: "1.16"
+    go: "1.18"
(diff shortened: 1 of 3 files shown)


roxctl/helm/output/output_test.go
```diff
@@ -108,23 +108,23 @@ func (suite *helmOutputTestSuite) TestValidate() {
 			outputDir: "test_output_dir",
 		},
 		"should not fail for valid chartName and non provided outputDir": {
-			chartName: common.ChartCentralServices,
+			chartName:    common.ChartCentralServices,
```
janisz, line 111: 🤔  
```
roxctl/helm/output/output_test.go:111: File is not `gofmt`-ed with `-s` (gofmt)
			chartName:    common.ChartCentralServices,
```

rukletsov (changes requested):
I'd stick to making only necessary changes. It seems to be accepted by our CI: https://github.com/stackrox/stackrox/pull/3207

janisz:
> I'd stick to making only necessary changes. It seems to be accepted by our CI

@rukletsov I think it's fine to change formatting. If we do not do this then this changes will randomly appear when someone runs gofmt on file and commit it.

rukletsov (approved):

PR #3198, opened 2022-09-26
ROX-12272: Don't allow auto-upgrades in managed centrals

fredrb:
## Description

This PR disables auto-upgrades for any sensor connected to a managed central.

1. Toggle is disabled by default
2. API rejects request that tries to set it to true if `ROX_MANAGED_CENTRAL` is set

There's a possible UX discussion here. The problem is that in a MS scenario, central version will most likely always be *higher* than sensor. Therefore, there will be a period, after upgrading managed centrals and before releasing the image for sensors to be upgraded, where we identify that a manual upgrade is required, but there's no image available for sensor's to be upgraded to. Probably for limited availability won't be an issue. 

Additionally, we could disable the toggle in the Front-End on a managed central. Because even though it's enabled the auto-upgrade will not be triggered and the warning will be present (covered in #3331)

## Checklist

## Testing Performed

When on managed centrals, the toggle won't be allowed to be set since the request returns 500:

```json
{
   "error":"auto-upgrade not allowed on managed central",
   "code":13,
   "message":"auto-upgrade not allowed on managed central",
   "details":[]
}
```

https://user-images.githubusercontent.com/6811076/194374104-905392b3-a9c3-4ad3-bfac-2167c83f7318.mov

### DIFF
--- a/central/sensorupgradeconfig/datastore/datastore_impl.go
+++ b/central/sensorupgradeconfig/datastore/datastore_impl.go
@@ -2,11 +2,13 @@ package datastore
 
 import (
 	"context"
+	"github.com/pkg/errors"
 
 	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/central/sensorupgradeconfig/datastore/internal/store"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/sac"
 )
 
@@ -18,6 +20,7 @@ type dataStore struct {
 
 var (
 	sacHelper = sac.ForResource(resources.SensorUpgradeConfig)
+	ErrAutoUpgradeNotAllowed = errors.New("auto-upgrade not allowed on managed central")
 )
 
 func (d *dataStore) initialize() error {
@@ -49,6 +52,10 @@ func (d *dataStore) UpsertSensorUpgradeConfig(ctx context.Context, sensorUpgrade
 		return err
 	} else if !ok {
 		return sac.ErrResourceAccessDenied
+	}
+
+	if env.ManagedCentral.BooleanSetting() && sensorUpgradeConfig.EnableAutoUpgrade {
+		return ErrAutoUpgradeNotAllowed
 	}
 
 	if err := d.store.Upsert(ctx, sensorUpgradeConfig); err != nil {
--- a/central/sensorupgradeconfig/datastore/datastore_impl.go
+++ b/central/sensorupgradeconfig/datastore/datastore_impl.go
@@ -2,8 +2,8 @@ package datastore
 
 import (
 	"context"
-	"github.com/pkg/errors"
 
+	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/central/sensorupgradeconfig/datastore/internal/store"
 	"github.com/stackrox/rox/generated/storage"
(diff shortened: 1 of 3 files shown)


vikin91 (approved):
LGTM

misberner:
Thanks for tagging me @rukletsov . I think this indeed is a good illustration of why I think that managed services checks spread throughout the code base are a bit of an anti-pattern.

We already have a global option to disable auto-upgrade, which is accessible from the `Clusters` menu:
Screen Shot 2022-09-28 at 19 39 12

I would expect that a managed services customer:
- does not see this toggle at all
- sees this toggle, but is unable to change it to "on" (with an adequate error response)

Instead, they will see this toggle and be able to click on it and change its states, but that will not do anything.

To be fair, I don't think that instance of checking for managed services in the middle of the code is fairly benign. Its effects are not subtle, in fact, whenever this check has an effect (preventing upgradability of a cluster), the user will immediately see an error message relating to this effect. However, the placement is less than ideal: this portion of the code is responsible for checking if a _single_ sensor connection is eligible for auto-upgrade. The error furthermore states that this sensor is connect to "a" managed central. This is a bit confusing, because the customer will only see this error message through _the_ (not _a_) managed central instance.

I think it would therefore be cleaner to actually add the managed services check in `central/sensorupgradeconfig/datastore/datastore_impl.go`, ensuring that in a managed services setting, we always have the desired value for this setting (and the user is unable to change it). That way, the logic is inserted much much higher up the callstack, and we leverage an already existing flag for a global setting in the day-to-day sensor connection handling logic.

misberner:
Note, what I said above obviously depends on what the intended shelf life of disabling auto upgrades for managed services is, and I don't object to it in its current form, considering the above. But I think the change suggested by me isn't really that much more complicated (<10 lines total) and therefore is worth the small user experience win.

fredrb:
@misberner I've made the update according to your suggestions. But the FE doesn't show any error messages if BE returns an error on `POST /v1/sensorupgrades/config`. The toggle is simply not set. This means we need changes in the FE to provide feedback to the user on why this isn't allowed. 

Another issue I see with this approach is that _auto_ upgrade is not set -- but the upgrade button will still show up if sensor version is behind central. This can be problematic if images used in managed central aren't published downstream yet.  

image
image

central/sensorupgradeconfig/datastore/datastore_impl.go
```diff
@@ -51,6 +53,10 @@ func (d *dataStore) UpsertSensorUpgradeConfig(ctx context.Context, sensorUpgrade
 		return sac.ErrResourceAccessDenied
 	}
 
+	if env.ManagedCentral.BooleanSetting() && sensorUpgradeConfig.EnableAutoUpgrade {
```
misberner, line 56:
```suggestion
	if env.ManagedCentral.BooleanSetting() && sensorUpgradeConfig.GetEnableAutoUpgrade() {
```
always prefer getters in protobufs for `nil`-safety

central/sensorupgradeconfig/datastore/singleton.go
```diff
@@ -35,7 +35,9 @@ func addDefaultConfigIfEmpty(d DataStore) error {
 	if currentConfig != nil {
```
misberner, line 35: I think what you want here is to also check
```suggestion
	if currentConfig != nil && (!env.ManagedCentral.BooleanSetting() || !currentConfig.GetEnableAutoUpgrade()) {
```
because in managed mode, you want to make sure that auto-upgrade is disabled, even if for some reason it is stored as enabled in the DB.

misberner (changes requested):

central/sensorupgradeconfig/datastore/singleton.go
fredrb: That makes sense. Thanks!

misberner (approved):

PR #3202, opened 2022-09-26
ROX-12818: Postgres store support for cycle references

keyallis:
## Description

Add support for cyclical references in Postgres store generation.
Because references into the same table as the parent table would be named the same I added a special case naming scheme for cyclical references. This seemed less intrusive than redefining the name of parent references since those appear more commonly and could cause many conflicts.

## Checklist

## Testing Performed

Built and deployed locally, verified the cyclical relationship was present in the postgres table definitions. Checked for any suspicious log entries on central.


theencee (approved):
Let's merge this but please add a JIRA (use this [epic](https://github.com/stackrox/stackrox/pull/3136)) that is about doing this in a more analytical way

PR #3208, opened 2022-09-27
ROX-12826: Improve sensor integration tests to remove `time.Sleep`

fredrb:
## Description

We wanted for a while to remove the `time.Sleep` call in sensor integration tests to reduce the possible flakiness to it. 

This PR introduces a new set of functions `LastDeploymentState` and `LastViolationStateForDeployment` that will try an assertion multiple times until it times-out. 

The problem was originally pointed out by @porridge during the show-and-tell presentation. 

Previously, a test was written as:
```go
time.Sleep(2 * time.Second)
lastNginxDeploymentUpdate := GetLastMessageWithDeploymentName(messages, "nginx-deployment")
deployment := lastNginxDeploymentUpdate.GetEvent().GetDeployment()
assert.Equal(
		t,
		deployment.ServiceAccountPermissionLevel,
		storage.PermissionLevel_ELEVATED_IN_NAMESPACE,
		fmt.Sprintf("permission level has to be %s", storage.PermissionLevel_ELEVATED_IN_NAMESPACE),
)
```
Now can be written as:
```go
testC.LastDeploymentState("nginx-deployment",
        func(deployment *storage.Deployment) bool {
		return deployment.ServiceAccountPermissionLevel == storage.PermissionLevel_ELEVATED_IN_NAMESPACE
	},
	"Permission level has to be elevated in namespace")
```

## Checklist

## Testing Performed

- Current sensor integration tests ran with new code.

### DIFF
--- a/sensor/tests/resource/helper.go
+++ b/sensor/tests/resource/helper.go
@@ -40,7 +40,7 @@ import (
 )
 
 const (
-	// DefaultNamespace - the default namespace used to create the resources
+	// DefaultNamespace the default namespace used to create the resources
 	DefaultNamespace string = "sensor-integration"
 )
 
@@ -74,10 +74,13 @@ func objByKind(kind string) k8s.Object {
 	}
 }
 
-// TestCallback the test callback
+// TestCallback represents the test case function written in the go test file.
 type TestCallback func(t *testing.T, testContext *TestContext, objects map[string]k8s.Object)
 
-// TestContext the test context
+// TestContext holds all the information about the cluster and sensor under test. A TestContext represents
+// a test case run where the input is a set of resources applied to the cluster and the output is a set of
+// messages emitted by Sensor. Each Go test should use a single TestContext instance to manage cluster interaction
+// and assertions.
 type TestContext struct {
 	t               *testing.T
 	r               *resources.Resources
@@ -100,12 +103,12 @@ func defaultCentralConfig() CentralConfig {
 	}
 }
 
-// NewContext creates a new test context
+// NewContext creates a new test context with default configuration.
 func NewContext(t *testing.T) (*TestContext, error) {
 	return NewContextWithConfig(t, defaultCentralConfig())
 }
 
-// NewContextWithConfig creates a new test context with custom central configuration
+// NewContextWithConfig creates a new test context with custom central configuration.
 func NewContextWithConfig(t *testing.T, config CentralConfig) (*TestContext, error) {
 	envConfig := envconf.New().WithKubeconfigFile(conf.ResolveKubeConfigFile())
 	r, err := resources.New(envConfig.Client().RESTConfig())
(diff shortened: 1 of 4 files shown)


porridge (commented):
Nice, but perhaps consider throwing in `time.Sleep(<a few ms>)` into the loops to prevent burning CPU busy-looping.

sensor/tests/resource/helper.go
```diff
@@ -229,6 +233,80 @@ func runPermutation(files []YamlTestFile, i int, cb func([]YamlTestFile)) {
 	}
 }
 
+// AssertFunc is the deployment state assertion function signature.
+type AssertFunc func(deployment *storage.Deployment) bool
```
lvalerom, line 237: Maybe it would be better to pass here `testing.T` to be able to assert inside or maybe return an error? 🤔  For these tests this is ok since there's only one assertion but in the service tests there are multiple assertions for each deployment and there is no way to know which assertion failed if we return a `bool`

sensor/tests/resource/helper.go
```diff
@@ -229,6 +233,80 @@ func runPermutation(files []YamlTestFile, i int, cb func([]YamlTestFile)) {
...
+			if deployment != nil && assertion(deployment) {
+				// Assertion matched the case. We can return here without failing the test case
+				return
+			}
+		}
+	}
+}
+
+// AlertAssertFunc is the alert assertion function signature.
+type AlertAssertFunc func(alertResults *central.AlertResults) bool
```
lvalerom, line 265: Similar comment to `AssertFunc`

lvalerom (changes requested):
Looks good, just a few a comments.
> Nice, but perhaps consider throwing in `time.Sleep(<a few ms>)` into the loops to prevent burning CPU busy-looping.

+1

fredrb:
> Nice, but perhaps consider throwing in time.Sleep(<a few ms>) into the loops to prevent burning CPU busy-looping.

Good catch. I've added a `time.Ticker` to the loops to mitigate that.

sensor/tests/resource/helper.go
fredrb: Yea I thought about that. But then the test would fail on the first check of the state. The reason why I explicitly didn't want to use a test assertion here is because the semantic of this function is "eventually this assertion should be true". If Go test has a single `assert.Equal` that fails the complete test fails -- even if the most recent deployment/violation state passes the assertion. 

If you have a suggestion of how to workaround that or design in a way that we can assert multiple state but only consider the last assertion as the test result, I'd be happy to change this.

fredrb: > maybe return an error?

Returning an error might be an improvement. If there are no other ideas of how to restructure this I'll change it to return an error.

lvalerom: That makes sense. Then maybe returning an `error` or allowing the `AssertFunc` to modify the `message`.

fredrb:
@lvalerom updated

lvalerom (approved):
LGTM

PR #3211, opened 2022-09-27
ROX-12823: Fix links in Most Common Image Vulnerabilities

vjwilson:
## Description

Problem addressed:
- On Vulnerability Management Dashboard, in a Postgres environment, click a link in Most Common Image Vulnerabilities.
- See that the link URL is still trying to go to the old CVE type, `/cve/...`, instead of to the new IMAGE_CVE type, `image-cve`

This change makes those links conditional on whether you are in a Postgres environment or not.

## Checklist
- [-] Unit test and regression tests added (@pedrottimark is fixing the skipped tests that caused us to miss this problem, in another PR)

## Testing Performed

Before
Screen Shot 2022-09-26 at 4 03 04 PM

After
Screen Shot 2022-09-26 at 4 57 15 PM


pedrottimark (approved):
Thank you very much!

PR #3185, opened 2022-09-23
Add sorting to collections table.

dvail:
## Description

Adds sorting to the collection table column headers.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

Load the collections page. The table should default to sorted by name, ascending, with the "Collection" column header highlighted in blue.
image

Clicking on the column header reverses the sort.
image

The description and inUse fields follow the same behavior:
image
image
image
image

If no results are returned from the server, the sorting functionality of the headers is disabled:
image

PR #3212, opened 2022-09-27
Implement toolbar with search and pagination for collections table.

dvail:
## Description

Adds a toolbar with search by name and pagination functionality to the collections table.

## Checklist

## Testing Performed

Visit the collections page to see the table with toolbar, defaulting to no search value and the first page of results.
image

Begin editing text into the search field:
- The results in the table should update to reflect a case-insensitive match on the name field. 
- Result requests should be debounced and only populate x00 ms after the user finishes typing. 
- The search value should be present in the URL.
- The pagination numbers should reflect the new number of items displayed, if less than the page size.
image

Clicking the back button should remove the search filter:
image

Pagination limits the amount of data shown in the table at one time. Page size and page number should be tracked in the URL. (Duplicate mock data was used here to increase the result set).
image
image
image

Test that sorting works in conjunction with filtering and pagination:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -101,7 +101,12 @@ function CollectionsTable({
                             page={page}
                             perPage={perPage}
                             onSetPage={(_, newPage) => setPage(newPage)}
-                            onPerPageSelect={(_, newPerPage) => setPerPage(newPerPage)}
+                            onPerPageSelect={(_, newPerPage) => {
+                                if (collectionsCount < (page - 1) * newPerPage) {
+                                    setPage(1);
+                                }
+                                setPerPage(newPerPage);
+                            }}
                         />
                     </ToolbarItem>
                 </ToolbarContent>
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -1,16 +1,16 @@
 import React, { useMemo } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
+    Bullseye,
+    Button,
+    ButtonVariant,
+    Pagination,
+    SearchInput,
+    Text,
     Toolbar,
     ToolbarContent,
     ToolbarItem,
-    Pagination,
-    Button,
-    ButtonVariant,
     Truncate,
-    SearchInput,
-    Bullseye,
-    Text,
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
 import { TableComposable, TableVariant, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
@@ -35,8 +35,7 @@ export type CollectionsTableProps = {
     hasWriteAccess: boolean;
 };
 
-// TODO We are using this value elsewhere in the app - extract to a central location?
-const TYPING_DELAY = 800;
+const SEARCH_INPUT_REQUEST_DELAY = 800;
 
 function CollectionsTable({
     collections,
@@ -71,7 +70,11 @@ function CollectionsTable({
     }
 
     const onSearchInputChange = useMemo(
-        () => debounce((value: string) => setSearchFilter({ Collection: value }), TYPING_DELAY),
+        () =>
+            debounce(
+                (value: string) => setSearchFilter({ Collection: value }),
+                SEARCH_INPUT_REQUEST_DELAY
+            ),
         [setSearchFilter]
     );
 
@@ -86,7 +89,7 @@ function CollectionsTable({
                     <ToolbarItem variant="search-filter" className="pf-u-flex-grow-1">
                         <SearchInput
                             aria-label="Search by name"
-                            placeholder="Search by name..."
+                            placeholder="Search by name"
                             value={searchValue}
                             onChange={onSearchInputChange}
                         />
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -78,6 +78,9 @@ function CollectionsTable({
         [setSearchFilter]
     );
 
+    // Currently, it is not expected that the value of `searchFilter.Collection` will
+    // be an array even though it would valid. This is a safeguard for future code
+    // changes that might change this assumption.
     const searchValue = Array.isArray(searchFilter.Collection)
         ? searchFilter.Collection.join('+')
         : searchFilter.Collection;
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -1,24 +1,54 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { useHistory } from 'react-router-dom';
-import { Button, ButtonVariant, Truncate, Bullseye, Text } from '@patternfly/react-core';
+import {
+    Toolbar,
+    ToolbarContent,
+    ToolbarItem,
+    Pagination,
+    Button,
+    ButtonVariant,
+    Truncate,
+    SearchInput,
+    Bullseye,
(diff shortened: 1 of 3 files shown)


dvail:
* master
  * **PR #3152** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3152" target="_blank">Graphite</a> 
    * **PR #3173** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3173" target="_blank">Graphite</a> 
      * **PR #3181** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3181" target="_blank">Graphite</a> 
        * **PR #3185** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3185" target="_blank">Graphite</a> 
          * **PR #3212** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3212" target="_blank">Graphite</a>  👈
    * **PR #3155** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3155" target="_blank">Graphite</a> 
    * **PR #3157** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3157" target="_blank">Graphite</a> 
    * **PR #3154** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3154" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3212?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -40,8 +66,34 @@ function CollectionsTable({ collections, getSortParams, hasWriteAccess }: Collec
         });
     }
 
+    const searchValue = Array.isArray(searchFilter.Collection)
```
dvail, line 69: An array value shouldn't happen in normal circumstances via the UI, but is here in case of a future logic change or someone manually editing URL search parameters.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -1,24 +1,54 @@
...
+    pagination: UseURLPaginationResult;
+    searchFilter: SearchFilter;
+    setSearchFilter: (searchFilter: SearchFilter) => void;
     getSortParams: GetSortParams;
     hasWriteAccess: boolean;
 };
 
-function CollectionsTable({ collections, getSortParams, hasWriteAccess }: CollectionsTableProps) {
+// TODO We are using this value elsewhere in the app - extract to a central location?
+const TYPING_DELAY = 800;
```
dvail, line 39: This would be the third place this line shows up in the app. I can make a follow up PR to extract the constant to keep this consistent.

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
```diff
@@ -31,17 +34,30 @@ const sortOptions = {
...
+    const countQuery = useCallback(() => getCollectionCount(searchFilter), [searchFilter]);
     const { data: countData, loading: countLoading, error: countError } = useRestQuery(countQuery);
-
     const isDataAvailable = typeof listData !== 'undefined' && typeof countData !== 'undefined';
     const isLoading = !isDataAvailable && (listLoading || countLoading);
     const loadError = listError || countError;
 
+    useEffectAfterFirstRender(() => {
+        // Prevent viewing a page beyond the maximum page count
+        if (typeof countData !== 'undefined' && page > Math.ceil(countData / perPage)) {
```
dvail, line 56: It would be nice to roll this logic into the `useURLPagination` hook if possible. Maybe as an additional follow up.

ui/apps/platform/src/hooks/useURLPagination.ts
```diff
@@ -1,7 +1,7 @@
 import { useCallback } from 'react';
 import useURLParameter from './useURLParameter';
 
-type UseURLPaginationResult = {
+export type UseURLPaginationResult = {
```
dvail, line 4: I'm finding that any time a function is exported we will eventually need the parameter and return types to be exported as well...

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -1,24 +1,54 @@
...
+    ToolbarContent,
+    ToolbarItem,
+    Pagination,
+    Button,
+    ButtonVariant,
+    Truncate,
+    SearchInput,
+    Bullseye,
+    Text,
+} from '@patternfly/react-core';
```
pedrottimark, line 14: Because prettier indentation affects changed files anyway, would you be willing to alphabetize the names?

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
pedrottimark: Please make sure all occurrences have similar effects on search, to avoid subtle problems if single shared ever changes in the future.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -40,8 +70,39 @@ function CollectionsTable({ collections, getSortParams, hasWriteAccess }: Collec
...
+        : searchFilter.Collection;
+
     return (
         <>
+            <Toolbar>
+                <ToolbarContent>
+                    <ToolbarItem variant="search-filter" className="pf-u-flex-grow-1">
+                        <SearchInput
+                            aria-label="Search by name"
+                            placeholder="Search by name..."
```
pedrottimark, line 89: 3 periods in placeholder seems to be a new trend. Only other occurrence is in policy categories.
* Placeholder examples for `SearchInput` element do not include it:
    https://www.patternfly.org/v4/components/search-input
* UX writing guidelines do not include it:
    https://www.patternfly.org/v4/ux-writing/punctuation#ellipses-...

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
pedrottimark: Yes, `useEffectAfterFirstRender` hooks feels awkward.

pedrottimark: Forgive me for asking instead of studying. Does change to search filter reset to page 1?

pedrottimark (approved):
A few comments for your consideration.

Reuse of hooks is repaying your investment.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
dvail: Hmm good point. Other uses have this to handle _pagination_ requests, but this is for a search filter. Same general goal I suppose (easing number of requests to the server), but maybe I'll keep this separate instead for the reason you mentioned.

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
dvail: It does not, but I think maybe it should. As for prior art:

- Violations _does_ reset page when the search filter changes
- Risk Acceptance _does not_ reset page when the search filter changes
- Policies doesn't support server side pagination

Let me take another look at this and see if it is worth changing to remove some of the awkwardness in this PR, or if I can come up with a good follow up plan.

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
pedrottimark: As a delayed reaction, can you write a comment, so future readers understand the reason?

ui/apps/platform/src/Containers/Collections/CollectionsTablePage.tsx
dvail: I've got a simple alternate approach to this that I'll push up instead that relies on https://github.com/stackrox/stackrox/pull/3228 to function properly.

Essentially just calling `setPage(1); setSearchFilter(value);` when the filter changes, but requires a fix to the `useURLParameter()` hook to work.

PR #3214, opened 2022-09-27
Set Postgres connect timeout to 15s instead of 2m

connorgorman:
## Description

The default connect timeout is 2m. This is excruciatingly long as it happens almost every time the product is deployed. Cut this down so the retry will happen faster

## Checklist

## Testing Performed

CI


dashrews78 (approved):
LGTM

PR #3222, opened 2022-09-28
Dashrews/ROX-12733 move persistent logs to postgres take 2

dashrews78:
## Description

Take on moving persistent logs to Postgres.  The approach here was to update the logger and add a PostgresWriter that would be added as a Sync to the logger.  Then that writer would write the data to Postgres.  A lot of things had to be moved around to avoid circular dependencies because with this approach the logger depends on Postgres. 

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


pkg/postgres/schema/persistent_logs.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+const (
+	// PersistentLogsTableName holds the database table name
+	PersistentLogsTableName = "persistent_logs"
+)
+
+// PersistentLogs holds the Gorm model for Postgres table `network_flows`.
+type PersistentLogs struct {
+	LogID     string     `gorm:"column:log_id;type:bigserial;primaryKey"`
```
c-du, line 38: LogLevel and Module

pkg/persistentlog/store/store.go
```diff
@@ -0,0 +1,308 @@
...
+	countStmt = "SELECT COUNT(*) FROM persistent_logs"
+
+	getStmt = `SELECT log, timestamp 
+	FROM persistent_logs 
+	WHERE timestamp between $1 and $2`
+	deleteStmt = "DELETE FROM persistent_logs WHERE timestamp < $1"
+	walkStmt   = "SELECT log, timestamp FROM persistent_logs"
+
+	createTableStmt = `
+		CREATE TABLE IF NOT EXISTS persistent_logs (
```
c-du, line 29: Use gorm to manage schema so that we are compatible to future changes.

PR #3223, opened 2022-09-28
Implement deletion functionalty for collections table

dvail:
## Description

Adds the ability to delete collections from the collection table.

In addition to the core functionality, there are a couple of UX items that could use feedback that I didn't find an exact example of in the app:

### 1. Disabled table checkboxes for collections that are "in use"

Currently the only available action in "Bulk actions" is to delete a collection, which is not a valid option for collections that are in use. This is handled by disabling the checkboxes for these rows to give the user an indication that they cannot be interacted with. The checkboxes have title text explaining this when hovered.
image
Additionally, if the user clicks the "select all" checkbox at the top of the table, the items are selected but still disabled. The collections are unable to be deleted from the bulk actions dropdown and the count reflects this.
image
This is somewhat based on how the Policies table works, except that there isn't a need to disable checkboxes in that table due to there being valid options other than "Delete".

### 2. Alert toasts when delete requests return an error

When one or more collections are deleted, the happy path show a single "success" toast message and refreshes the data. Since each delete request is sent separately, it is possible that some requests will fail while other requests succeed.

Each request that fails to delete will individually display a single "error" toast message with the reason the request failed. If one or more requests fail during a bulk delete action, but _at least one succeeds_, an additional "warning" toast will be displayed that shows the user how many of the collections were deleted.
image

## Checklist

## Testing Performed

Collections that are `inUse` cannot be deleted.

Load the collections table and see that any collection listed as `inUse` has its selection checkbox disabled. The "Bulk actions" dropdown will also be disabled when no collections are selected.
image

Opening the menu of an `inUse` collection has the delete option disabled:
image
Collections that are not `inUse` have the delete option enabled:
image

If the user selects all rows via the "select all" checkbox at the top of the table, the `inUse` row checkboxes are still disabled. Opening the "Bulk actions" dropdown will prompt for deletion of only selected collections that are not `inUse`.
image

If the user clicks the "select all" checkbox, and then individually unchecks each collection that is not `inUse`, the "Bulk actions" dropdown will show a message that the collections cannot be deleted:
image

Selecting the "Delete collection" item from the menu of a single collection will prompt the user to confirm deletion:
image

Selecting the cancel option will close the modal and keep the data intact. Selecting the confirmation option will delete the collection and present a toast to the user:
image

Selecting multiple collections via checkbox and using the "Bulk action" delete option will present the user with the confirmation modal. The modal will specify how many selected collections that are not `inUse` will be deleted:
image

If one or more of the collections fails to delete, each failure will result in a toast message telling the user about the error. In the case of a partial success, an additional modal will tell the user how many of the selected collections were successfully deleted. If no deletions were successful, this additional modal will not be shown.
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -1,23 +1,30 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
     Bullseye,
     Button,
     ButtonVariant,
+    Dropdown,
+    DropdownItem,
+    DropdownToggle,
     Pagination,
     SearchInput,
     Text,
     Toolbar,
     ToolbarContent,
     ToolbarItem,
+    ToolbarItemVariant,
     Truncate,
 } from '@patternfly/react-core';
-import { SearchIcon } from '@patternfly/react-icons';
+import { CaretDownIcon, SearchIcon } from '@patternfly/react-icons';
 import { TableComposable, TableVariant, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
 import debounce from 'lodash/debounce';
+import pluralize from 'pluralize';
 
+import ConfirmationModal from 'Components/PatternFly/ConfirmationModal';
 import EmptyStateTemplate from 'Components/PatternFly/EmptyStateTemplate';
 import LinkShim from 'Components/PatternFly/LinkShim';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import useTableSelection from 'hooks/useTableSelection';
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #3181** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3181" target="_blank">Graphite</a> 
    * **PR #3185** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3185" target="_blank">Graphite</a> 
      * **PR #3212** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3212" target="_blank">Graphite</a> 
        * **PR #3223** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3223" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3223?utm_source=stack-comment).

pedrottimark (approved):

PR #3224, opened 2022-09-28
ROX-11323 Enable fake workloads in local-sensor

lvalerom:
## Description

These changes enable the fake workloads in `local-sensor`

## Checklist

## Testing Performed

```
go run tools/local-sensor/main.go -record -resync=0s -with-fakeworkload=scale/workloads/default.yaml
```

generates a trace file (`k8s-trace.jsonl`) with fake events.

### DIFF
--- a/sensor/kubernetes/fake/fake.go
+++ b/sensor/kubernetes/fake/fake.go
@@ -85,14 +85,33 @@ type WorkloadManager struct {
 	networkManager      manager.Manager
 }
 
+// WorkloadManagerConfig WorkloadManager's configuration
+type WorkloadManagerConfig struct {
+	workloadFile string
+	client       client.Interface
+}
+
+// ConfigDefaults default configuration
+func ConfigDefaults() *WorkloadManagerConfig {
+	return &WorkloadManagerConfig{
+		workloadFile: workloadPath,
+	}
+}
+
+// WithWorkloadFile configures the WorkloadManagerConfig's WorkloadFile field
+func (c *WorkloadManagerConfig) WithWorkloadFile(file string) *WorkloadManagerConfig {
+	c.workloadFile = file
+	return c
+}
+
 // Client returns the mock client
 func (w *WorkloadManager) Client() client.Interface {
 	return w.client
 }
 
 // NewWorkloadManager returns a fake kubernetes client interface that will be managed with the passed Workload
-func NewWorkloadManager() *WorkloadManager {
-	data, err := os.ReadFile(workloadPath)
+func NewWorkloadManager(config *WorkloadManagerConfig) *WorkloadManager {
+	data, err := os.ReadFile(config.workloadFile)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil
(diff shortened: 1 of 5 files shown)


sensor/kubernetes/fake/fake.go
```diff
@@ -85,14 +85,33 @@ type WorkloadManager struct {
...
 }
 
+// WorkloadManagerConfig WorkloadManager's configuration
+type WorkloadManagerConfig struct {
+	workloadFile string
+	client       client.Interface
+}
+
+// ConfigDefaults default configuration
+func ConfigDefaults() *WorkloadManagerConfig {
```
vikin91, line 95: How about renaming this to `NoFakeWorklod()` to highlight that the default behavior is not to use any fake workload?

sensor/kubernetes/fake/fake.go
```diff
@@ -85,14 +85,33 @@ type WorkloadManager struct {
 	networkManager      manager.Manager
 }
 
+// WorkloadManagerConfig WorkloadManager's configuration
+type WorkloadManagerConfig struct {
+	workloadFile string
+	client       client.Interface
```
vikin91, line 91: Why do we have client here when no methods use that field? This suggests that maybe it should be a var outside of this struct (or shall simply be reamoved)?

vikin91 (commented):
Looks really cool!

Let's maybe clarify the `client       client.Interface` in the `WorkloadManagerConfig` before we merge.

sensor/kubernetes/fake/fake.go
lvalerom: I see where you're coming from but I think that renaming this function to `NoFakeWorkload` could lead to a misunderstanding. `ConfigDefaults` takes the default path as `/var/scale/stackrox/workload.yaml` from `workloadPath`. The decision whether we use fake workloads or not is based on the existence of the configured file (whether it is the default file previously mentioned or a configured one using `WithWorkloadFile`)

sensor/kubernetes/fake/fake.go
lvalerom: Oops that's left over from when I was spiking on this 😅 I'll remove it.

sensor/kubernetes/fake/fake.go
vikin91: Ah, I see. Okay then 👍

vikin91 (approved):

tools/local-sensor/main.go
```diff
@@ -191,11 +199,19 @@ func registerHostKillSignals(startTime time.Time, fakeCentral *centralDebug.Fake
 // If a KUBECONFIG file is provided, then local-sensor will use that file to connect to a remote cluster.
 func main() {
 	localConfig := mustGetCommandLineArgs()
+	var fakeClient client.Interface
 	fakeClient, err := k8s.MakeOutOfClusterClient()
 	// when replying a trace, there is no need to connect to K8s cluster
 	if localConfig.ReplayK8sEnabled {
 		fakeClient = k8s.MakeFakeClient()
 	}
+	var workloadManager *fake.WorkloadManager = nil
```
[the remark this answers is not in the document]
gavin-stackrox: > I wonder why it didn't in the pre-merge check? 🤔

Yeah, me too 🤔 Did anyone figure that out?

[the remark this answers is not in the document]
porridge: My assumption was that the linter or its configuration was changed concurrently. I bet our lint master @janisz would know :-)

PR #3225, opened 2022-09-28
[ROX-1162] : Delete images before test to clear previous scans

charmik-redhat:
## Description

Delete images from db before test starts so that the DefaultPolicies test picks up fresh scans.

## Checklist

## Testing Performed

### DIFF
--- a/qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
+++ b/qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
@@ -1,5 +1,8 @@
 import static Services.getPolicies
 import static Services.waitForViolation
+
+import io.stackrox.proto.api.v1.SearchServiceOuterClass
+
 import common.Constants
 import groups.BAT
 import groups.SMOKE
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -128,6 +137,16 @@ class DefaultPoliciesTest extends BaseSpecification {
         gcrId = GCRImageIntegration.createDefaultIntegration()
         assert gcrId != ""
 
+        ImageService.clearImageCaches()
```
theencee, line 140: Is it possible to do this just for the test that failed instead of all of them. This is a pretty hefty test file and this deletion can slow it down quite a bit.

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -71,6 +74,12 @@ class DefaultPoliciesTest extends BaseSpecification {
             "webhookserver - Kubernetes Actions: Port Forward to Pod",
     ]
 
+    static final private String STRUTS_IMG = "quay.io/rhacs-eng/qa:struts-app"
```
theencee, line 77: What are these for? I don't see them being used anywhere.

charmik-redhat: Yeah, intended to use the coasts, just extracted the images to delete from the deployments.

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
charmik-redhat: I think setupSpec() runs just for the test file, so it should only run once. Do you mean remove and add back just the struts image which causes the specific failure instead of all images ?

theencee: Oh I missed the fact that it was setupSpec

theencee (approved):
approved pending test results. Try rerunning the e2e tests at least once to make sure it really did succeed.

keyallis:
I think these build issues can be solved by a rebase on master

PR #3226, opened 2022-09-28
Properly propagate context to Postgres delete request

connorgorman:
## Description

Propagate context correctly to delete request

## Checklist

## Testing Performed

Trivial, but CI


c-du (approved):

dashrews78 (changes requested):
Do we need to catch the handful of migrations I mentioned and any stores that are not generated?

dashrews78 (approved):
LGTM now.

PR #3227, opened 2022-09-28
Add Postgres query retry mechanism to handle transient errors

connorgorman:
## Description

There is really no free way to do this. I have added the logic and wired it into searches, gets and deletions. Upserts are in the stores and will be added in a different PR because that will modify autogen and muddy the PR. Effectively, if Postgres is unavailable with a transient error then we should try to be successful in the background while still respecting contexts

## Checklist

## Testing Performed

All postgres tests should pass and will add some more chaos-like tests in a follow up

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,6 +58,7 @@ require (
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/heroku/docker-registry-client v0.0.0
+	github.com/jackc/pgconn v1.13.0
 	github.com/jackc/pgtype v1.12.0
 	github.com/jackc/pgx/v4 v4.17.2
 	github.com/jinzhu/now v1.1.5 // indirect
@@ -248,7 +249,6 @@ require (
 	github.com/itchyny/gojq v0.12.5 // indirect
 	github.com/itchyny/timefmt-go v0.1.3 // indirect
 	github.com/jackc/chunkreader/v2 v2.0.1 // indirect
-	github.com/jackc/pgconn v1.13.0 // indirect
 	github.com/jackc/pgio v1.0.0 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgproto3/v2 v2.3.1 // indirect
--- a/pkg/errorhelpers/utils.go
+++ b/pkg/errorhelpers/utils.go
@@ -1,0 +1,16 @@
+package errorhelpers
+
+import "github.com/pkg/errors"
+
+// IsAny returns a bool if it matches any of the target errors
+// This helps consolidate code from
+// errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.ErrClosedPipe)
+// to errors.IsAny(err, io.EOF. io.ErrUnexpectedEOF, io.ErrClosedPipe)
+func IsAny(err error, targets ...error) bool {
+	for _, target := range targets {
+		if errors.Is(err, target) {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -1,12 +1,8 @@
 package pgutils
 
 import (
-	"io"
-	"net"
-
 	"github.com/jackc/pgconn"
 	"github.com/pkg/errors"
-	"github.com/stackrox/rox/pkg/errorhelpers"
 	"github.com/stackrox/rox/pkg/set"
 )
 
@@ -38,7 +34,6 @@ var transientPGCodes = set.NewFrozenStringSet(
 	"57P01", // admin_shutdown
 	"57P02", // crash_shutdown
 	"57P03", // cannot_connect_now
-	"57P04", // database_dropped
 	"57P05", // idle_session_timeout
 
 	// Class 58 — System Error (errors external to PostgreSQL itself)
@@ -51,8 +46,6 @@ func isTransientError(err error) bool {
 	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
 		return transientPGCodes.Contains(pgErr.Code)
 	}
-	if netErr := (*net.OpError)(nil); errors.As(err, &netErr) {
-		return netErr.Temporary() || netErr.Timeout()
-	}
-	return errorhelpers.IsAny(err, io.EOF, io.ErrUnexpectedEOF, io.ErrClosedPipe)
+	// Assume all other errors are transient
+	return false
 }
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -1,0 +1,58 @@
+package pgutils
+
+import (
+	"io"
+	"net"
+
+	"github.com/jackc/pgconn"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/errorhelpers"
+	"github.com/stackrox/rox/pkg/set"
+)
+
+var transientPGCodes = set.NewFrozenStringSet(
+	// Class 08 — Connection Exception
+	"08000", // connection_exception
+	"08003", // connection_does_not_exist
+	"08006", // connection_failure
+	"08001", // sqlclient_unable_to_establish_sqlconnection
+	"08004", // sqlserver_rejected_establishment_of_sqlconnection
+	"08007", // transaction_resolution_unknown
+	"08P01", // protocol_violation
+
+	// Class 40 — Transaction Rollback
+	"40000", // transaction_rollback
+	"40002", // transaction_integrity_constraint_violation
+	"40001", // serialization_failure
+	"40003", // statement_completion_unknown
+	"40P01", // deadlock_detected
+
+	// Class 55 — Object Not In Prerequisite State
+	"55000", // object_not_in_prerequisite_state
+	"55006", // object_in_use
+	"55P03", // lock_not_available
+
+	// Class 57 — Operator Intervention
+	"57000", // operator_intervention
+	"57014", // query_canceled
+	"57P01", // admin_shutdown
+	"57P02", // crash_shutdown
+	"57P03", // cannot_connect_now
+	"57P04", // database_dropped
+	"57P05", // idle_session_timeout
+
+	// Class 58 — System Error (errors external to PostgreSQL itself)
+	"58000", // system_error
+	"58030", // io_error
+)
+
+// isTransientError specifies if the passed error is transient and should be retried
+func isTransientError(err error) bool {
+	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
+		return transientPGCodes.Contains(pgErr.Code)
+	}
+	if netErr := (*net.OpError)(nil); errors.As(err, &netErr) {
+		return netErr.Temporary() || netErr.Timeout()
+	}
+	return errorhelpers.IsAny(err, io.EOF, io.ErrUnexpectedEOF, io.ErrClosedPipe)
+}
--- a/pkg/postgres/pgutils/retry.go
+++ b/pkg/postgres/pgutils/retry.go
@@ -26,8 +26,8 @@ func RetryExecQuery(fn func() error) error {
 // that fails with Transient errors
 func RetryQuery[T any](fn func() (T, error)) (T, error) {
 	// Run query immediately
-	if T, err := fn(); err == nil || !isTransientError(err) {
-		return T, err
+	if val, err := fn(); err == nil || !isTransientError(err) {
+		return val, err
 	}
 
 	expirationTimer := time.NewTimer(timeout)
--- a/pkg/postgres/pgutils/retry.go
+++ b/pkg/postgres/pgutils/retry.go
@@ -1,0 +1,54 @@
+package pgutils
+
+import (
+	"time"
+
+	"github.com/stackrox/rox/pkg/timeutil"
+)
+
+const (
+	interval = 5 * time.Second
+	timeout  = 5 * time.Minute
+)
+
+// RetryExecQuery is used to specify how long to retry to successfully run an exec query
+// that fails with Transient errors
+func RetryExecQuery(fn func() error) error {
+	// Shape fn to match the RetryQuery below
+	fnWithReturn := func() (struct{}, error) {
+		return struct{}{}, fn()
+	}
+	_, err := RetryQuery(fnWithReturn)
+	return err
+}
+
+// RetryQuery is used to specify how long to retry to successfully run a query
+// that fails with Transient errors
+func RetryQuery[T any](fn func() (T, error)) (T, error) {
+	// Run query immediately
+	if T, err := fn(); err == nil || !isTransientError(err) {
+		return T, err
+	}
+
+	expirationTimer := time.NewTimer(timeout)
+	defer timeutil.StopTimer(expirationTimer)
+
+	intervalTicker := time.NewTicker(interval)
+	defer intervalTicker.Stop()
+
+	var err error
+	for {
+		select {
+		case <-expirationTimer.C:
+			log.Fatalf("unsuccessful in reconnecting to the database: %v. Exiting...", err)
+		case <-intervalTicker.C:
+			// Uses err outside the for loop to allow for the expiration to show the last err received
+			// and provide context for the expiration
+			var ret T
+			ret, err = fn()
+			if err == nil || !isTransientError(err) {
+				return ret, err
+			}
+		}
+	}
+}
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/pointers"
 	"github.com/stackrox/rox/pkg/postgres"
+	"github.com/stackrox/rox/pkg/postgres/pgutils"
 	"github.com/stackrox/rox/pkg/postgres/walker"
 	"github.com/stackrox/rox/pkg/random"
 	searchPkg "github.com/stackrox/rox/pkg/search"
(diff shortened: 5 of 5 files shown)


pkg/postgres/pgutils/error.go
```diff
@@ -0,0 +1,59 @@
...
+	"io"
+	"net"
+
+	"github.com/jackc/pgconn"
+	"github.com/stackrox/rox/pkg/set"
+)
+
+var transientSQLCodes = set.NewFrozenStringSet(
+	// Class 08 — Connection Exception
+	"08000", // connection_exception
```
connorgorman, line 13: indentation

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	closure := func() error {
+		err := fn()
+		if isTransientError(err) {
+			return retry.MakeRetryable(err)
+		}
+		return err
+	}
+	numTries := int(timeout / 1 * time.Second)
+	err := retry.WithRetry(closure,
```
connorgorman, line 28: one comment here is that we have a 15s connect timeout so maybe retrying in this way isn't the best

connorgorman:
@erthalion Any of the PG codes you think that we should not retry on?

pkg/search/postgres/common.go
```diff
@@ -738,6 +767,24 @@ func RunGetManyQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
+	var results [][]byte
+	err = pgutils.RetryQuery(ctx, func() error {
+		results, err = retriableRunGetManyQueryForSchema(ctx, query, db)
+		return err
+	})
+	return results, err
+}
+
 // RunCursorQueryForSchema creates a cursor against the database
 func RunCursorQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (fetcher func(n int) ([][]byte, error), closer func(), err error) {
```
connorgorman, line 789: These are not retriable

pkg/search/postgres/common.go
```diff
@@ -797,9 +844,11 @@ func RunDeleteRequestForSchema(schema *walker.Schema, q *v1.Query, db *pgxpool.P
 		return err
 	}
 
-	_, err = db.Exec(context.Background(), query.AsSQL(), query.Data...)
-	if err != nil {
-		return errors.Wrapf(err, "could not delete from %q", schema.Table)
-	}
-	return nil
+	return pgutils.RetryQuery(context.Background(), func() error {
```
connorgorman, line 847: PR for propagation will be merged and I'll rebase

pkg/search/postgres/common.go
```diff
@@ -623,7 +599,21 @@ func RunSearchRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1
...
+	rows, err := tracedQuery(ctx, db, queryStr, query.Data...)
+	if err != nil {
+		if !buildinfo.ReleaseBuild {
+			debug.PrintStack()
+			log.Errorf("Query issue: %s %+v: %v", queryStr, query.Data, err)
+		}
+		return nil, err
+	}
+	defer rows.Close()
+	log.Debugf("SEARCH: ran query %s; data %+v", queryStr, query.Data)
```
connorgorman, line 615: I think I probably want to remove this debug log or at least the query.Data portion because this could log secrets

pkg/search/postgres/common.go
```diff
@@ -668,10 +658,52 @@ func RunSearchRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1
...
+func RunSearchRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (searchResults []searchPkg.Result, err error) {
+	var query *query
+	// Add this to be safe and convert panics to errors,
+	// since we do a lot of casting and other operations that could potentially panic in this code.
+	// Panics are expected ONLY in the event of a programming error, all foreseeable errors are handled
+	// the usual way.
+	defer func() {
+		if r := recover(); r != nil {
+			if query != nil {
+				log.Errorf("Query issue: %s %+v: %v", query.AsSQL(), query.Data, r)
```
connorgorman, line 671: This can't be on releases due to logging secret data

pkg/postgres/pgutils/error.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if pgErr, ok := err.(*pgconn.PgError); ok {
+		return transientSQLCodes.Contains(pgErr.Code)
+	}
+	if err == io.EOF || err == io.ErrUnexpectedEOF || err == io.ErrClosedPipe {
+		return true
+	}
+	if netErr, ok := err.(*net.OpError); ok {
+		return netErr.Temporary() || netErr.Timeout()
+	}
```
misberner, line 60: Use modern error handling
```suggestion
	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
		return transientSQLCodes.Contains(pgErr.Code)
	}
	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.ErrClosedPipe) {
		return true
	}
	if netErr := (*net.OpError)(nil); errors.As(err, &netErr) {
		return netErr.Temporary() || netErr.Timeout()
	}
```

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+const (
+	interval = 5 * time.Second
+	timeout  = 5 * time.Minute
+)
+
+// RetryQuery is used to specify how long to retry to successfully run a query
+// that fails with Transient errors
+func RetryQuery(fn func() error) error {
```
misberner, line 14: I think it would be nicer to rewrite this as
```suggestion
func RetryQuery(ctx context.Context, fn func(ctx context.Context) error) error {
```
that way, the timeout could be made universal by using a `context.WithTimeout` instead of the expiration timer.

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,37 @@
...
+	// Run query immediately
+	if err := fn(); !isTransientError(err) {
+		return err
+	}
+
+	expiration := time.NewTimer(timeout)
+	defer expiration.Stop()
```
misberner, line 21: If you don't use a context, use 
```suggestion
	defer timeutil.StopTimer(expiration)
```
to prevent a resource leak

pkg/postgres/pgutils/retry.go
misberner: Nevermind, since you fatal on timeout, I guess a consistent timeout isn't all that relevant.

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	interval = 5 * time.Second
+	timeout  = 5 * time.Minute
+)
+
+// RetryQuery is used to specify how long to retry to successfully run a query
+// that fails with Transient errors
+func RetryQuery(fn func() error) error {
+	// Run query immediately
+	if err := fn(); !isTransientError(err) {
```
misberner, line 16: I find it a bit subtle to rely on the fact that `isTransientError` returns `false` for `nil`. Maybe better to explicitly rewrite this as
```suggestion
	if err := fn(); err == nil || !isTransientError(err) {
```

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,37 @@
...
+	intervalTicker := time.NewTicker(interval)
+	defer intervalTicker.Stop()
+
+	var err error
+	for {
+		select {
+		case <-expiration.C:
+			log.Fatalf("unsuccessful in reconnecting to the database: %v. Exiting...", err)
+		case <-intervalTicker.C:
+			if err = fn(); !isTransientError(err) {
```
misberner, line 32: Same here. Also, remove the declaration in l.26 in favor of a more tightly-scoped `err`:
```suggestion
			if err := fn(); err == nil || !isTransientError(err) {
```

pkg/search/postgres/common.go
misberner: Add a `TODO` in some form?

misberner (commented):
partial review, flushing comments

pkg/postgres/pgutils/error.go
connorgorman: Honestly, I see that they added this in go1.13, but I don't think it looks much better or is more clear generally speaking. Is there a functional benefit to the new style?

pkg/postgres/pgutils/retry.go
connorgorman: The declaration in line 26 allows the final error to be logged in case of expiration

pkg/search/postgres/common.go
connorgorman: I created a jira ticket for this and marked it as a blocker for Postgres

pkg/search/postgres/common.go
```diff
@@ -680,16 +712,21 @@ func RunCountRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1.
...
-	if err := row.Scan(&count); err != nil {
-		debug.PrintStack()
-		log.Errorf("Query issue: %s %+v: %v", queryStr, query.Data, err)
-		return 0, err
-	}
-	return count, nil
+	err = pgutils.RetryQuery(func() error {
+		row := tracedQueryRow(ctx, db, queryStr, query.Data...)
+		if err := row.Scan(&count); err != nil {
+			if !buildinfo.ReleaseBuild {
```
connorgorman, line 721: I can roll this back is it's not directly pertinent to this PR

pkg/search/postgres/common.go
```diff
@@ -797,9 +844,11 @@ func RunDeleteRequestForSchema(schema *walker.Schema, q *v1.Query, db *pgxpool.P
...
 
-	_, err = db.Exec(context.Background(), query.AsSQL(), query.Data...)
-	if err != nil {
-		return errors.Wrapf(err, "could not delete from %q", schema.Table)
-	}
-	return nil
+	return pgutils.RetryQuery(func() error {
+		_, err = db.Exec(context.Background(), query.AsSQL(), query.Data...)
+		if err != nil {
+			return errors.Wrapf(err, "could not delete from %q", schema.Table)
```
misberner, line 850: Well actually you do wrap the error here, so the `==` checks or direct casts would already fail here.

pkg/postgres/pgutils/retry.go
misberner, line 14: Looking at the use of `RetryQuery`, I think you could benefit from generics:
```
func RetryQuery[T any](fn func() (T, error)) {
```
to avoid the extra `var` declaration outside of the closure

pkg/postgres/pgutils/retry.go
connorgorman: 👍

pkg/postgres/pgutils/retry.go
connorgorman: oh yeah, this is so nice :)

go.mod
```diff
@@ -135,6 +135,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/jackc/pgconn v1.13.0
```
connorgorman, line 138: very odd this is moved into it's own line 🤔

pkg/search/postgres/common.go
connorgorman: done

pkg/search/postgres/common.go
connorgorman: Yup, fixed the error handling

pkg/search/postgres/common.go
```diff
@@ -668,10 +658,52 @@ func RunSearchRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1
...
 
+// RunSearchRequestForSchema executes a request against the database for given schema
+func RunSearchRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (searchResults []searchPkg.Result, err error) {
+	var query *query
+	// Add this to be safe and convert panics to errors,
+	// since we do a lot of casting and other operations that could potentially panic in this code.
+	// Panics are expected ONLY in the event of a programming error, all foreseeable errors are handled
+	// the usual way.
+	defer func() {
+		if r := recover(); r != nil {
```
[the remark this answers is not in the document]
connorgorman: Right, I think it's mostly just pointer exceptions, etc

connorgorman (commented):
Appreciate the review @misberner. Have moved to using generics and cleaned up a good chunk of the code. Please re-review when you get the chance :)

go.mod
connorgorman: fixed

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,54 @@
...
+const (
+	interval = 5 * time.Second
+	timeout  = 5 * time.Minute
+)
+
+// RetryExecQuery is used to specify how long to retry to successfully run an exec query
+// that fails with Transient errors
+func RetryExecQuery(fn func() error) error {
+	// Shape fn to match the RetryQuery below
+	fnWithReturn := func() (struct{}, error) {
```
misberner, line 18: I would just inline this into the `RetryQuery` argument, but up to you

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,54 @@
...
+	_, err := RetryQuery(fnWithReturn)
+	return err
+}
+
+// RetryQuery is used to specify how long to retry to successfully run a query
+// that fails with Transient errors
+func RetryQuery[T any](fn func() (T, error)) (T, error) {
+	// Run query immediately
+	if T, err := fn(); err == nil || !isTransientError(err) {
+		return T, err
```
misberner, line 30:
```suggestion
	if val, err := fn(); err == nil || !isTransientError(err) {
		return val, err
```

Don't think you intended to use the type param as variable name

pkg/postgres/pgutils/retry.go
```diff
@@ -0,0 +1,54 @@
...
+	defer timeutil.StopTimer(expirationTimer)
+
+	intervalTicker := time.NewTicker(interval)
+	defer intervalTicker.Stop()
+
+	var err error
+	for {
+		select {
+		case <-expirationTimer.C:
+			log.Fatalf("unsuccessful in reconnecting to the database: %v. Exiting...", err)
```
misberner, line 43: It might make sense, but it's a bit odd that a repeated failure of a transient error is _worse_ than a permanent error. I guess that a permanent error is assumed to mean that there's something wrong with the input?

misberner (approved):
Just some questions and stylistic suggestions in the comments.

pkg/postgres/pgutils/retry.go
c-du, line 43: Shall we return last error after this?

pkg/postgres/pgutils/error.go
```diff
@@ -0,0 +1,58 @@
...
+	"08006", // connection_failure
+	"08001", // sqlclient_unable_to_establish_sqlconnection
+	"08004", // sqlserver_rejected_establishment_of_sqlconnection
+	"08007", // transaction_resolution_unknown
+	"08P01", // protocol_violation
+
+	// Class 40 — Transaction Rollback
+	"40000", // transaction_rollback
+	"40002", // transaction_integrity_constraint_violation
+	"40001", // serialization_failure
```
c-du, line 26: How do you tell which codes are transient? Like this one and database_dropped, it does not look like they can be corrected by retries.

pkg/postgres/pgutils/retry.go
connorgorman: This is a fatal. Central exists

pkg/postgres/pgutils/error.go
connorgorman: I was reading some notes on the Postgres page. I think we could remove the dropped database one

connorgorman: Both Repeatable Read and Serializable isolation levels can produce errors that are designed to prevent serialization anomalies. As previously stated, applications using these levels must be prepared to retry transactions that fail due to serialization errors. Such an error's message text will vary according to the precise circumstances, but it will always have the SQLSTATE code 40001 (serialization_failure).

pkg/postgres/pgutils/error.go
```diff
@@ -0,0 +1,58 @@
...
+	"57P05", // idle_session_timeout
+
+	// Class 58 — System Error (errors external to PostgreSQL itself)
+	"58000", // system_error
+	"58030", // io_error
+)
+
+// isTransientError specifies if the passed error is transient and should be retried
+func isTransientError(err error) bool {
+	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
```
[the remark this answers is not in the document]
connorgorman: I only need this in one place since I changed it so will keep it like before

pkg/postgres/pgutils/retry.go
connorgorman: thx

pkg/postgres/pgutils/retry.go
connorgorman: New logic basically says all errors are transient unless they are specifically postgres errors that are not transient in which case crashing won't do anything

pkg/errorhelpers/utils.go
```diff
@@ -0,0 +1,16 @@
+package errorhelpers
+
+import "github.com/pkg/errors"
+
+// IsAny returns a bool if it matches any of the target errors
+// This helps consolidate code from
+// errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.ErrClosedPipe)
+// to errors.IsAny(err, io.EOF. io.ErrUnexpectedEOF, io.ErrClosedPipe)
+func IsAny(err error, targets ...error) bool {
```
connorgorman, line 9: not used, but gonna keep it cause it's useful

PR #3228, opened 2022-09-28
Fixes useURLParameter hook to allow multiple sequential mutations.

dvail:
## Description

The useURLParameter hook previously did not allow sequential updates to different URL parameters if they were in a single tick of the event loop, due to asynchronous code in `react-router-dom`.

```typescript
// e.g.
const [, setPage] = useURLParameter('page', 1);
const [, setPerPage] = useURLParameter('perPage', 10);

setPage(2);
setPerPage(20);

// Results in a URL search query of `?page=1&perPage=20` instead of the 
// expected `?page=2&perPage=20`, since the first call is overwritten by the second.
```

## Checklist

## Testing Performed

Added unit test to cover this case.
Manual testing of the Violations page, which uses this hook for search/sort/pagination.

### DIFF
--- a/ui/apps/platform/src/hooks/useURLParameter.ts
+++ b/ui/apps/platform/src/hooks/useURLParameter.ts
@@ -36,6 +36,12 @@ function useURLParameter(keyPrefix: string, defaultValue: QueryValue): UseURLPar
     // as the URL parameters do not change
     const setValue = useCallback(
         (newValue: QueryValue, historyAction: Action = 'push') => {
+            // Note that we use the version of `location` on `history` here, since it is mutable (compared
+            // to the immutable `location` when used directly.) In this case, we aren't looking
+            // for a reference change to trigger a rerender, we are looking for the current up-to-date
+            // location search parameters.
+            //
+            // https://v5.reactrouter.com/web/api/history/history-is-mutable
             const previousQuery = getQueryObject(history.location.search) || {};
             const newQuery = {
                 ...previousQuery,
--- a/ui/apps/platform/src/hooks/useURLParameter.ts
+++ b/ui/apps/platform/src/hooks/useURLParameter.ts
@@ -36,23 +36,23 @@ function useURLParameter(keyPrefix: string, defaultValue: QueryValue): UseURLPar
     // as the URL parameters do not change
     const setValue = useCallback(
         (newValue: QueryValue, historyAction: Action = 'push') => {
-            const previousQuery = getQueryObject(location.search) || {};
-            const newQueryString = getQueryString({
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/hooks/useURLParameter.ts
```diff
@@ -36,23 +36,23 @@ function useURLParameter(keyPrefix: string, defaultValue: QueryValue): UseURLPar
...
+            const newQuery = {
                 ...previousQuery,
                 [keyPrefix]: newValue,
-            });
+            };
 
             // If the value passed in is `undefined`, don't display it in the URL at all
             if (typeof newValue === 'undefined') {
-                delete newQueryString[keyPrefix];
+                delete newQuery[keyPrefix];
```
dvail, line 47: In the previous commit, the type of `newQueryString` was in fact a `string`. So this line was `delete string[string]`...

ui/apps/platform/src/hooks/useURLParameter.ts
```diff
@@ -36,23 +36,23 @@ function useURLParameter(keyPrefix: string, defaultValue: QueryValue): UseURLPar
     // as the URL parameters do not change
     const setValue = useCallback(
         (newValue: QueryValue, historyAction: Action = 'push') => {
-            const previousQuery = getQueryObject(location.search) || {};
-            const newQueryString = getQueryString({
+            const previousQuery = getQueryObject(history.location.search) || {};
```
dvail, line 39: According to the react router docs, `history.location` is mutable while `location` is mutable, and they generally recommend the latter because of this. However in this case we want to rely on the mutability in order to get the latest version of `location.search` when this function is called multiple times from application code.

pedrottimark: You are super searcher! Even more important, is knowing when you need to ;)

Please write a comment which summarizes what what you learned, maybe even with link to the docs, so we can adjust if needed when we replace `useHistory` with `useNavigate` in React Router 6.

pedrottimark (approved):

ui/apps/platform/src/hooks/useURLParameter.ts
bradr5: @dvail I assume you meant to say `location` is immutable which is why it's recommended?

bradr5 (approved):

ui/apps/platform/src/hooks/useURLParameter.ts
dvail: Whoops, yeah `location` is immutable, I'll edit.

Also Mark, I believe React Router 6 has a `useParams` hook or similar which might even render this function obsolete.

PR #3232, opened 2022-09-29
Separate central-db yaml's from Central

connorgorman:
## Description

This is an initial PR that will have follow ups.  We need to split these so we can more easily generate a set of YAMLs that will deploy just central-db and not influence Central's YAMLs

In a follow up, we should split the configuration out as well which will allow us to upgrade more easily and also separate the database connection config from the general Central config.

## Checklist

## Testing Performed

Postgres tests should still work


image/templates/helm/stackrox-central/templates/01-central-00-db-serviceaccount.yaml
```diff
@@ -0,0 +1,17 @@
+{{- include "srox.init" . -}}
+{{ if and ._rox.central.db.enabled (not ._rox.central.db.external) -}}
+---
```
c-du, line 3: Remove this line?

c-du (approved):

image/templates/helm/stackrox-central/templates/01-central-00-db-serviceaccount.yaml
connorgorman, line 3:
```suggestion
```

PR #3242, opened 2022-09-29
Fix a struct tag in a proto

misberner:
## Description

The double quotation mark is placed at the wrong place, making `go vet` complain.

## Checklist

## Testing Performed

- CI


misberner:
@keyallis @connorgorman it looks like using a field of a parent as a foreign key doesn't work. I've tried fixing this in the PG generator, but wasn't able to do so successfully in the self-allotted time. @connorgorman just from your intuition, should this be a trivial fix, or will it be harder?

connorgorman:
@misberner Realistically, this will probably be challenging. The generator graph code is pretty complicated so I would punt it

misberner:
Ok, this isn't a cosmetic issue though. The fact that it works today is only due to the fact that the struct tag is not being parsed correctly (or at all). @keyallis I think in this case it should be removed altogether, but I don't know what the implications are for the collections feature, or whether it's really possible to punt on this.

keyallis:
Just now seeing this PR. It took me a while to find this typo but then I was able to make progress, see #3202 for the relevant changes to allow the store generator to support embedded cyclical foreign keys. This change is already in master and this PR can be closed.

PR #3243, opened 2022-09-29
ROX-11279: Disable removed default policies during migration to Postgres

c-du:
## Description
We store removed_default_policies in Rocksdb which would block default policies to be added to policies primary table. We do not allow default policies to be removed since 3.66.
While migrating to Postgres, we plan to remove this table and disable the default policies that were removed before. 
The default policies are a copy of current of default policies without the new default policies added during 3.66 to 3.73. 
The following is a list of default policies that are NOT included in this migration.
```
automount_service_account_token.json
deployment_has_ingress_network_policy.json
external_exposure.json
log4shell.json
privilege_escalation.json
severity_high_privileged.json
spring4shell.json
```
## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


pkg/booleanpolicy/policyversion/convertor.go
```diff
@@ -27,16 +33,20 @@ func EnsureConvertedToLatest(p *storage.Policy) error {
...
-	// If it's not the latest version, delegate to the upgrader
-	// CurrentVersion should always be the latest, thus this will always involve an upgrade.
-	if !IsCurrentVersion(ver) {
-		if err := upgradePolicyTo(p, CurrentVersion()); err != nil {
+	switch diff := Compare(ver, toVersion); {
+	case diff > 0:
+		// No downgrade
+		utils.CrashOnError(errors.Errorf("Unexpected version %s, cannot downgrade policy version to %s", ver.String(), toVersion.String()))
+	case diff < 0:
+		// If it's blow the requested version, delegate to the upgrader
```
dashrews78, line 41: nit:  below

migrator/migrations/n_39_to_n_40_postgres_policies/legacy/store.go
```diff
@@ -24,6 +28,7 @@ type Store interface {
 // New returns a new Store instance using the provided bolt DB instance.
 func New(db *bolt.DB) Store {
 	bolthelper.RegisterBucketOrPanic(db, policyBucket)
+	bolthelper.RegisterBucketOrPanic(db, removedDefaultPolicyBucket)
```
dashrews78, line 31: Does this bucket still exist after 66?  For instance if a customer started at 70, do we have the bucket just empty?

c-du: Yes, it still exists. If a customer started at 70, it is just empty but we cannot assume that.

dashrews78 (approved):
LGTM

migrator/migrations/n_39_to_n_40_postgres_policies/legacy/remove_test.go
```diff
@@ -0,0 +1,96 @@
+package legacy
+
+import (
+	"context"
+	"encoding/json"
```
connorgorman, line 5: i think this needs to be jsonpb

migrator/migrations/n_39_to_n_40_postgres_policies/legacy/store_impl.go
```diff
@@ -5,20 +5,24 @@ package legacy
 
 import (
 	"context"
+	"encoding/json"
```
connorgorman, line 8: jsonpb I think

migrator/migrations/n_39_to_n_40_postgres_policies/legacy/remove_test.go
c-du: This is only used to marshal a boolean, not a proto message.

migrator/migrations/n_39_to_n_40_postgres_policies/legacy/store_impl.go
c-du: This is existing boolean in json.

PR #3245, opened 2022-09-29
Dashrews/ROX-12874 fix int type mappings to postgres

dashrews78:
## Description

We were mapping uint32, int64 and uint64 types to the postgres integer.  The size and value ranges of a Postgres integer map to the go int and int32.  The max value of a uint32 requires us to use the bigint Postgres type for that.  int64 and uint64 should map to Postgres bigint as well.

## Checklist

## Testing Performed

TODO(replace-me)
Added a couple of cases in the query area.  Mostly the existing test cases handled the verification.


pkg/postgres/schema/deployments.go
```diff
@@ -106,7 +106,7 @@ type Deployments struct {
 	ClusterId                     string                  `gorm:"column:clusterid;type:varchar"`
 	ClusterName                   string                  `gorm:"column:clustername;type:varchar"`
 	Annotations                   map[string]string       `gorm:"column:annotations;type:jsonb"`
-	Priority                      int64                   `gorm:"column:priority;type:integer"`
+	Priority                      int64                   `gorm:"column:priority;type:bigint"`
```
dashrews78, line 109: This feels like a large type for a Priority field.  But I'm not an expert on the data.  If that is something we should look as to whether some of the fields are right sized, that should be done separately.

connorgorman (approved):
remove bigintarray then looks good

pkg/postgres/walker/walker.go
```diff
@@ -360,13 +360,16 @@ func handleStruct(ctx context, schema *Schema, original reflect.Type) {
...
 				continue
-			case reflect.Uint32, reflect.Uint64, reflect.Int32, reflect.Int64:
+			case reflect.Int32:
 				if typeIsEnum(elemType) {
 					schema.AddFieldWithType(field, EnumArray, opts)
 				} else {
 					schema.AddFieldWithType(field, IntArray, opts)
 				}
 				continue
+			case reflect.Uint32, reflect.Uint64, reflect.Int64:
```
c-du, line 370: I would suggest to keep using integer to store uint32, just convert it to/from int32 before and after update/write the data.

We need to support uint64 which is also beyond max int64.

dashrews78: We don't currently use uint64.  I have concerns about using integer for uint32.  I feel like that is asking for trouble from a maintenance perspective as it would be easy for a conversion of uint32 to/from int32 to be missed and introduce a bug.  IMO using bigint is safer and allows us to more quickly resolve the blocker.  It also seems to be in line with how the various ORMs map the types.  the ORMs also seem to map Uint64 to bigint due to the performance degradation of using numeric.

proto/storage/test.proto
```diff
@@ -23,6 +23,9 @@ message TestSingleKeyStruct {
     bool bool                    = 4 [(gogoproto.moretags) = 'search:"Test Bool"'];
     // repeated bool bool_slice: currently unsupported
 
+    // When using this ensure your precision needs are met and accounted for.  This will use a
```
c-du, line 26: Double spaces.

c-du (approved):

PR #3246, opened 2022-09-29
ROX-12860: Remove flag ROX_FRONTEND_VM_UPDATES and use postgres env var instead

theencee:
## Description

ROX_FRONTEND_VM_UPDATES cannot ship without postgres and vice versa. So just remove it and gate on the postgres flag.

## Checklist

## Testing Performed

Deployed manually to a gke cluster and saw that the CVEs and components were split into Image, Platform and Node.


ui/apps/platform/src/Containers/VulnMgmt/List/Cves/VulnMgmtListCves.js
```diff
@@ -296,7 +296,6 @@ const VulnMgmtCves = ({
     const [selectedCveIds, setSelectedCveIds] = useState([]);
     const [bulkActionCveIds, setBulkActionCveIds] = useState([]);
     const { isFeatureFlagEnabled } = useFeatureFlags();
-    const showVMUpdates = isFeatureFlagEnabled('ROX_FRONTEND_VM_UPDATES');
     const usesPostgres = isFeatureFlagEnabled('ROX_POSTGRES_DATASTORE');
```
vjwilson, line 299: I'd bring this into alignment with the pattern everywhere else.

```suggestion
    const showVMUpdates = isFeatureFlagEnabled('ROX_POSTGRES_DATASTORE');
```

And likewise just use `idAttribute={showVMUpdates ? 'id' : 'cve'}` in the change below.

ui/apps/platform/src/Containers/VulnMgmt/List/Images/VulnMgmtListImages.js
```diff
@@ -39,7 +39,7 @@ export const defaultImageSort = [
 ];
 
 export function getCurriedImageTableColumns(watchedImagesTrigger, isFeatureFlagEnabled) {
-    const isFrontendVMUpdatesEnabled = isFeatureFlagEnabled('ROX_FRONTEND_VM_UPDATES');
+    const isFrontendVMUpdatesEnabled = isFeatureFlagEnabled('ROX_POSTGRES_DATASTORE');
```
vjwilson, line 42: Ooops, we were checking the flag twice in this file.

vjwilson (approved):
Ok as is if working, but two suggestions for clarity (which I can do in a later PR).

ui/apps/platform/src/Containers/VulnMgmt/List/Images/VulnMgmtListImages.js
theencee: discussed offline. Won't make changes because that cannot be refactored outside of callback below

PR #3254, opened 2022-09-29
Update Scanner dependency

RTann:
## Checklist

## Testing Performed

CI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,10 +2,6 @@ module github.com/stackrox/rox
 
 go 1.18
 
-// CAVEAT: This introduces a circular dependency. If you change this line, you MUST change the "exclude"
-// directive at the bottom of the file as well.
-require github.com/stackrox/scanner v0.0.0-20220929215850-334d4658d5e6
-
 require (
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 // CAVEAT: This introduces a circular dependency. If you change this line, you MUST change the "exclude"
 // directive at the bottom of the file as well.
-require github.com/stackrox/scanner v0.0.0-20220426001230-9ab6777c9581
+require github.com/stackrox/scanner v0.0.0-20220929215850-334d4658d5e6
```
misberner, line 7: It looks like the above comment is outdated, since you didn't actually update the `exclude` directive. Remove the comment and the `exclude` directive

RTann:
Updating the `go.mod` to remove the exclude did nothing to the `go.sum`, which I guess is exactly what is expected since go (I think since 1.17?) now prunes the graph to only include up to immediate dependencies of dependencies

misberner (approved):

PR #3256, opened 2022-09-29
CI: fix 3.70.x

gavin-stackrox:
## Description

Changes:
- Removed setup_remote_docker from run-qa-tests. It is not needed and was causing Circle CI to barf.
- Removed uses of docker.io and stackrox.io.

## Checklist

## Testing Performed

CI is sufficient. There are failures in the e2e tests and the upgrade tests that are to be expected and will be justified elsewhere.


0x656b694d (approved):
Looks enigmatic to me!

PR #3259, opened 2022-09-30
ROX-12505 Bump hashstructure to v2.0.2

lvalerom:
## Description

Bump `hashstructure` to v2.0.2

## Checklist

## Testing Performed

CI


central/sensor/service/pipeline/deploymentevents/pipeline.go
```diff
@@ -166,7 +166,7 @@ func (s *pipelineImpl) runGeneralPipeline(ctx context.Context, deployment *stora
 	}
 
 	var err error
-	deployment.Hash, err = hashstructure.Hash(deployment, &hashstructure.HashOptions{})
+	deployment.Hash, err = hashstructure.Hash(deployment, hashstructure.FormatV2, &hashstructure.HashOptions{})
```
connorgorman, line 169: So if we move this to v2 then this is identical to the code in the deduper. Should we set the deployment hash in sensor instead so we don't end up computing it in two places? Make sure the hash is ignored by sensorhash so that it'll still default the state timestamp. Then we can remove this instance?

lvalerom: I added `sensorhash` to the `hash` field but I think it wasn't necessary since it already has `hash:"ignore"` which is the default used by the library, so the field shouldn't be taken into account in the detector deduper.

go.mod
```diff
@@ -454,3 +454,5 @@ replace (
 exclude k8s.io/client-go v12.0.0+incompatible
 
 exclude github.com/openshift/client-go v3.9.0+incompatible
+
+replace github.com/mitchellh/hashstructure => github.com/mitchellh/hashstructure/v2 v2.0.2
```
misberner, line 458: Is there a reason you're not just importing this as `github.com/mitchellh/hashstructure/v2`?

lvalerom: Not really. I'll change it to use `v2` directly.

misberner (approved):

PR #3262, opened 2022-09-30
ROX-12808: Fix diagnostic test for rosa

janisz:
## Description

It looks like cluster names on rosa are different then default cluster name and central prefix is used to generate them . This PR check if running on rosa and expect different path in diagnostic bundle. It also changes the way error will be printed to be more obvious what's wrong. 
https://github.com/stackrox/stackrox/blob/f9a64978f195eaa16bd60e000222304d44303370/central/debug/service/service.go#L61

## Testing Performed

CI

### DIFF
--- a/qa-tests-backend/src/test/groovy/DiagnosticBundleTest.groovy
+++ b/qa-tests-backend/src/test/groovy/DiagnosticBundleTest.groovy
@@ -1,17 +1,23 @@
 import static com.jayway.restassured.RestAssured.given
+
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+
 import com.jayway.restassured.config.RestAssuredConfig
 import com.jayway.restassured.config.SSLConfig
-import groups.BAT
+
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/DiagnosticBundleTest.groovy
```diff
@@ -102,20 +108,22 @@ class DiagnosticBundleTest extends BaseSpecification {
...
-                    }
+                    entries.add(entry.name)
                 }
             } finally {
                 zis.close()
             }
-            assert foundK8sInfo
+
+            def entry = "kubernetes/" +
+                    "${Env.CI_JOBNAME.contains('rosa') ? '_central-cluster' : ClusterService.DEFAULT_CLUSTER_NAME}" +
```
janisz, line 124: @misberner any idea why this has changed on ROSA?

janisz:
@misberner You are right. It looks like `remote` file is on rosa so there must be a different cause

PR #3264, opened 2022-09-30
RS-544 Fix compatability with new infractl json output

BradLugo:
## Description

These changes are to make sure the changes from https://github.com/stackrox/infra/pull/706 will be compatible with the rest of stackrox's CI.

## Testing Performed

CI via [stackrox/test-gh-actions](https://github.com/stackrox/test-gh-actions): 
- Test PR: https://github.com/stackrox/test-gh-actions/pull/49 
- Test run: https://github.com/stackrox/test-gh-actions/actions/runs/3225066721

### DIFF
--- a/.github/workflows/scripts/create-cluster.sh
+++ b/.github/workflows/scripts/create-cluster.sh
@@ -31,8 +31,7 @@ function cluster_status() {
 }
 
 function cluster_destroying() {
-    STATUS=$(cluster_status "$1")
-    [ "$STATUS" -eq 3 || "$STATUS" -eq "DESTROYING" ]
+    [ "$(cluster_status "$1")" = "DESTROYING" ]
 }
 
 function infra_status_summary() {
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/create-cluster.sh
```diff
@@ -49,27 +50,32 @@ case $(cluster_status "$CNAME") in
 "")
     gh_summary "Cluster $CNAME doesn't exist."
     ;;
-1)
+0|FAILED)
+    # Cluster exists already.
```
gavin-stackrox, line 54: comment does not match state
```suggestion
```

.github/workflows/scripts/create-cluster.sh
```diff
@@ -49,27 +50,32 @@ case $(cluster_status "$CNAME") in
...
+    infra_status_summary "$CNAME" "Cluster failed to create"
+    exit 0
```
gavin-stackrox, line 56: In the current state, infractl get for a cluster that failed to create will have no .Status field which will result in a  "null" value from `jq` and thus fall through to the `*)` case and it will still try to create the cluster. I'd say we want to keep the existing behavior i.e. do not exit 0 here. A failed cluster might be due to something that was rectified and we are trying again.

gavin-stackrox (changes requested):

gavin-stackrox (approved):
LGTM! Hold off merge for Michaël as to whether this is 👍 and/or how to test with stackrox/test-gh-actions (it at all).

0x656b694d:
The workflow downloads `infractl` every time it is needed. Is it `infractl` version or server version change? I wonder if we need to support the old format, or can switch directly to the new one.
We probably want to copy the change to test-gh-actions first for smoke-testing.
Otherwise LGTM.

BradLugo:
> Is it infractl version or server version change?
The former. 

If `.github/workflows/scripts/create-cluster.sh` doesn't run on push triggers, I agree we can get rid of the old format. I didn't want to break backward compatibility if it was going to be disruptive to the daily workflow.

@0x656b694d Do you have an example of using test-gh-actions for smoke testing?

0x656b694d:
Please cherry-pick this change to https://github.com/stackrox/actions.

PR #3267, opened 2022-09-30
Add page parameter parsing and component shell for collection form page

dvail:
## Description

This adds the empty shell of the main collection form page as well as parsing out the page action (view/create/edit/clone) from the URL.

## Checklist

## Testing Performed

For visual testing purposes, the page action and ID are rendered in the form page component.

Visit the collection page and load the table of collections in the system.
image

Click the name of one of the collections in the table, this should update the page URL and take you to an empty "view mode" page for the collection.
image

From the table, open the menu of one of the collection and click the "Edit" button:
image

From the table, open the menu of one of the collection and click the "Clone" button:
image

From the table page, click the "Create collection" button in the header:
image

Error handling checks - for each of the below URLs, enter the value manually in the address bar:

`/main/collections/2?action=create` (will update the URL to `/main/collections/2` and display "view 2")
`/main/collections/2?action=bogus` (will update the URL to `/main/collections/2` and display "view 2")
`/main/collections?action=edit` (will update the URL to `/main/collections` and display the collections table)
`/main/collections?action=clone` (will update the URL to `/main/collections` and display the collections table)
`/main/collections?action=view` (will update the URL to `/main/collections` and display the collections table)
`/main/collections?action=bogus` (will update the URL to `/main/collections` and display the collections table)

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
@@ -1,11 +1,32 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+
+import useURLParameter from 'hooks/useURLParameter';
 
 import CollectionsTablePage from './CollectionsTablePage';
+import CollectionsFormPage from './CollectionFormPage';
+import { parsePageActionProp } from './collections.utils';
 
 function CollectionsPage() {
     // TODO Implement permissions once https://issues.redhat.com/browse/ROX-12619 is merged
     // const { hasWriteAccess } = usePermissions();
     const hasWriteAccessForCollections = true; // hasWriteAccess('TODO');
+
+    const [pageAction, setPageAction] = useURLParameter('action', undefined);
+    const { collectionId } = useParams();
+    const validPageActionProp = parsePageActionProp(pageAction, collectionId);
+
+    useEffect(() => {
+        // If the URL structure somehow gets into a state with an invalid action, clear
+        // the parameter to avoid confusing the user.
+        if (typeof pageAction !== 'undefined' && !validPageActionProp) {
+            setPageAction(undefined);
+        }
+    }, [pageAction, validPageActionProp, setPageAction]);
+
+    if (validPageActionProp) {
+        return <CollectionsFormPage pageAction={validPageActionProp} />;
+    }
 
     return <CollectionsTablePage hasWriteAccessForCollections={hasWriteAccessForCollections} />;
 }
--- a/ui/apps/platform/src/Containers/Collections/collections.utils.ts
+++ b/ui/apps/platform/src/Containers/Collections/collections.utils.ts
@@ -1,0 +1,41 @@
+export type CollectionPageAction =
+    | {
+          type: 'create';
+      }
+    | {
+          type: 'edit' | 'clone' | 'view';
+          collectionId: string;
+      };
+
+/**
+ * Parses raw URL parameters into a valid page action for the collections form.
+ *
+ * This ensures that page cannot get into invalid states, such as loading a collection ID
+ * with 'action=create', or having 'action={edit,clone,view}' without a collection ID.
+ *
+ * @param action The URL action search parameter
+ * @param collectionId The collection ID passed in the URL path
+ *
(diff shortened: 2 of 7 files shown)


dvail:
* master
  * **PR #3267** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3267" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3267?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/collections.utils.ts
```diff
@@ -0,0 +1,41 @@
+export type CollectionPageAction =
+    | {
+          type: 'create';
+      }
+    | {
+          type: 'edit' | 'clone' | 'view';
```
dvail, line 6: Note that `"view"` is implicit in the URL for collection pages, but is included here for a couple of reasons:

- Clarity when reading/writing code, instead of having to work with `'edit' | 'clone' | undefined`.
- Better type narrowing, since TypeScript can't use the type of a property to narrow in a discriminated union (TIL!)

Both `/collections/<collection_id>` and `/collections/<collection_id>?action=view` are technically valid, but when creating links we should prefer the former.

ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
```diff
@@ -1,12 +1,33 @@
...
     // const { hasWriteAccess } = usePermissions();
     const hasWriteAccessForCollections = true; // hasWriteAccess('TODO');
 
+    const [pageAction, setPageAction] = useURLParameter('action', undefined);
+    const { collectionId } = useParams();
+    const validPageActionProp = parsePageActionProp(pageAction, collectionId);
+
+    useEffect(() => {
+        // If the URL structure somehow gets into a state with an invalid action, clear
+        // the parameter to avoid confusing the user.
```
pedrottimark, line 21: This reminds me to ask if search query has page action, is anything else valid? For example, search filter.

A counter-example, in case it might ever arise in the future, is generate policy from search.

At the moment risk transfers the policy via Redux store, but it will soon pass via search query, to minimize the coupling between the containers (plus opening the door to generate from other contexts).

```text
/main/risk?s[Add%20Capabilities]=NET_BIND_SERVICE
/main/policy-management/policies?action=generate&s[Add%20Capabilities]=NET_BIND_SERVICE

File: ui/apps/platform/src/routePaths.js [hunk 3]
```diff
@@ -159,7 +160,7 @@ export const basePathToLabelMap = {
     [policyManagementBasePath]: 'Policy Management',
     [policiesBasePath]: 'Policy Management',
     [policyCategoriesPath]: 'Policy Categories',
-    [collectionsPath]: 'Collections',
+    [collectionsBasePath]: 'Collections',
```
pedrottimark, line 163: As a follow up, let’s discuss page title in browser history and tab.

Unlike path, which must be global, containers can take responsibility for page title.

pedrottimark (approved):

ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
dvail: I think an action with a search query makes sense as a valid use case (at least for policies, not collections so far). This was more addressing the (slim) possibility of someone visiting a page with a collection ID _plus_ action=create. These combinations aren't restricted on the policies page, and even though I think the only way for someone to get there is to hand-edit the URL, it does place the page into a funky state where the form doesn't quite work correctly.

ui/apps/platform/src/routePaths.js
dvail: Yes good call, I'll create a Jira task to address titles before I merge this PR.

ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
pedrottimark: Yeah, I can believe that I overlooked an inconsistent state for policies.

ui/apps/platform/src/routePaths.js
dvail: https://issues.redhat.com/browse/ROX-12952 ✅

PR #3276, opened 2022-10-02
Unknown errors and some pgx errors should be treated as transient

connorgorman:
## Description

The comment is correct, but the return should have been true

## Checklist

## Testing Performed

Was running some bounce tests and found that non-pg errors were being treated incorrectly

### DIFF
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -47,5 +47,5 @@ func isTransientError(err error) bool {
 		return transientPGCodes.Contains(pgErr.Code)
 	}
 	// Assume all other errors are transient
-	return false
+	return true
 }
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -52,6 +52,9 @@ func isTransientError(err error) bool {
 	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
 		return transientPGCodes.Contains(pgErr.Code)
 	}
+	if pgconn.SafeToRetry(err) {
+		return true
+	}
 	switch err {
 	case pgx.ErrTxClosed, pgx.ErrTxCommitRollback:
 		return true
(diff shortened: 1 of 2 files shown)


theencee (approved):

pkg/postgres/pgutils/error.go
```diff
@@ -46,6 +52,20 @@ func isTransientError(err error) bool {
...
+	case pgx.ErrTxClosed, pgx.ErrTxCommitRollback:
+		return true
+	case pgx.ErrNoRows:
+		return false
+	}
+	if netErr := (*net.OpError)(nil); errors.As(err, &netErr) {
+		if netErr.Temporary() || netErr.Timeout() {
+			return true
+		}
+		return errorhelpers.IsAny(err, syscall.ECONNREFUSED, syscall.ECONNRESET, syscall.ECONNABORTED)
```
connorgorman, line 68: There's gotta be a better way to handle these besides just enumerating them 🤔

pkg/postgres/pgutils/retry.go
```diff
@@ -39,6 +41,9 @@ func Retry2[T any](fn func() (T, error)) (T, error) {
 func Retry3[T any, U any](fn func() (T, U, error)) (T, U, error) {
 	// Run query immediately
 	if val1, val2, err := fn(); err == nil || !isTransientError(err) {
+		if err != nil && err != pgx.ErrNoRows {
```
connorgorman, line 44: won't be shipped, for testing

pkg/postgres/pgutils/error.go
```diff
@@ -46,6 +52,20 @@ func isTransientError(err error) bool {
 	if pgErr := (*pgconn.PgError)(nil); errors.As(err, &pgErr) {
 		return transientPGCodes.Contains(pgErr.Code)
 	}
-	// Assume all other errors are transient
-	return false
+	if pgconn.SafeToRetry(err) {
+		return true
+	}
+	switch err {
+	case pgx.ErrTxClosed, pgx.ErrTxCommitRollback:
```
theencee, line 59: This doesn't feel retryable. It could be any reason within the txn that caused it to fail.

pkg/postgres/pgutils/retry.go
```diff
@@ -52,6 +57,7 @@ func Retry3[T any, U any](fn func() (T, U, error)) (T, U, error) {
 	for {
 		select {
 		case <-expirationTimer.C:
+			debug.PrintStack()
```
theencee, line 60: Should only do in debug builds not release builds

theencee (approved):

pkg/postgres/pgutils/retry.go
misberner: Doesn't really hurt in release builds as we're fataling anyway

misberner (approved):

PR #3280, opened 2022-10-03
ROX-9617: Fold two functions into a single generic one.

porridge:
## Description

This was supposed to be an exercise in Go generics, but turned out that:
- generics don't actually help, since you cannot do `func bla[T whatever]() { obj := &T{} }` that would be required here (`invalid composite literal type`)
- generics are not required, since as of more recent controller-runtime versions one can achieve this result using an `UnstructuredList` and pass the GVK at runtime

## Checklist

## Testing Performed

Relying on CI now that I added an e2e test for this functionality.
Also ran the e2e test manually and made sure it works.


porridge:
https://github.com/stackrox/stackrox/pull/3281

misberner (approved):

porridge:
Thanks for the tip, @misberner . However the result is still problematic for another reason:
```
pkg/reconciler/reconciler_factory.go:99:26: list.Items undefined (type T has no field or method Items)
```
AFAICT accessing the `Items` field is the only way to iterate over the items :-/

porridge:
The operator tests and all required tests passed, which is all matters here, merging.

PR #3284, opened 2022-10-03
ROX-12808: Increase diagnostic bundle timeouts

janisz:
## Description

We experience multiple timeouts on diagnostic bundle creation on rosa cluster.
This PR doubles the time for sensor metrics gathering. 

## Testing Performed
CI

### DIFF
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -60,8 +60,8 @@ const (
 
 	centralClusterPrefix = "_central-cluster"
 
-	metricsPullTimeout     = 10 * time.Second
-	diagnosticsPullTimeout = 10 * time.Second
+	metricsPullTimeout     = 20 * time.Second
+	diagnosticsPullTimeout = 20 * time.Second
 	layout                 = "2006-01-02T15:04:05.000Z"
 	logWindow              = 20 * time.Minute
 )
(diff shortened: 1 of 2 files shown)


janisz:
> Wouldn't it be better to pass --timeout 60 in the respective roxctl invocation, given that this only seems to occur on Rosa?

I thought about this but then I thought that if we had this in e2e tests than maybe our customers also could run into this issue and since it's little change it's better to change default for everybody then just our tests. Especially, it's not resulting in any performance degradation.

central/debug/service/service.go
```diff
@@ -60,8 +60,8 @@ const (
 
 	centralClusterPrefix = "_central-cluster"
 
-	metricsPullTimeout     = 10 * time.Second
-	diagnosticsPullTimeout = 10 * time.Second
+	metricsPullTimeout     = 20 * time.Second
```
connorgorman, line 63: Does this really take 20 seconds to load data from a local http server?

central/debug/service/service.go
```diff
@@ -60,8 +60,8 @@ const (
...
+	diagnosticsPullTimeout = 20 * time.Second
```
connorgorman, line 64: it seems like this could be dynamic based on ctx timeout passed via roxctl or the ui

central/debug/service/service.go
janisz: Now it has timeouts. Maybe the issue is somewhere else. But after couple of timeouts one diagnostic bundle is created correctly.

central/debug/service/service.go
janisz: Issue created https://issues.redhat.com/browse/ROX-12940

janisz: And another https://issues.redhat.com/browse/ROX-12939

PR #3287, opened 2022-10-03
ROX-12760: Use passed context if nested resolver context is null

keyallis:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/graphql/resolvers/node_components.go
+++ b/central/graphql/resolvers/node_components.go
@@ -157,6 +157,12 @@ func (resolver *nodeComponentResolver) nodeComponentScopeContext() context.Conte
 		log.Errorf("attempted to scope context on nil")
 		return nil
 	}
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		return scoped.Context(resolver.ctx, scoped.Scope{
+			Level: v1.SearchCategory_IMAGE_COMPONENTS,
+			ID:    resolver.data.GetId(),
+		})
+	}
 	return scoped.Context(resolver.ctx, scoped.Scope{
 		Level: v1.SearchCategory_NODE_COMPONENTS,
 		ID:    resolver.data.GetId(),
--- a/central/graphql/resolvers/node_components.go
+++ b/central/graphql/resolvers/node_components.go
@@ -152,15 +152,13 @@ func queryWithNodeIDRegexFilter(q string) string {
 		search.NewQueryBuilder().AddRegexes(search.NodeID, ".+").Query())
 }
 
-func (resolver *nodeComponentResolver) withNodeComponentScope() context.Context {
-	if env.PostgresDatastoreEnabled.BooleanSetting() {
-		return scoped.Context(resolver.ctx, scoped.Scope{
-			Level: v1.SearchCategory_NODE_COMPONENTS,
-			ID:    resolver.data.GetId(),
-		})
+func (resolver *nodeComponentResolver) nodeComponentScopeContext() context.Context {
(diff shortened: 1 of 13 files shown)


central/graphql/resolvers/node_components.go
```diff
@@ -159,15 +154,13 @@ func queryWithNodeIDRegexFilter(q string) string {
...
-		return scoped.Context(ctx, scoped.Scope{
-			Level: v1.SearchCategory_NODE_COMPONENTS,
-			ID:    resolver.data.GetId(),
-		})
+func (resolver *nodeComponentResolver) nodeComponentScopeContext() context.Context {
+	if resolver.ctx == nil {
+		log.Errorf("attempted to scope context on nil")
+		return nil
 	}
-	return scoped.Context(ctx, scoped.Scope{
```
charmik-redhat, line 169: We use nodeComponentResolver when postgres is disabled too. In that case, we convert image components to node components and wrap inside nodeComponentResolver. So we should keep both scopes.

https://github.com/stackrox/stackrox/blob/3a5c247b14aba154d27879ff5d01582be24324d6/central/graphql/resolvers/components_v2.go#L79

central/graphql/resolvers/nodes.go
```diff
@@ -318,7 +319,7 @@ func (resolver *nodeResolver) Components(ctx context.Context, args PaginatedQuer
 	}
 	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getNodeRawQuery())
 
-	return resolver.root.componentsV2(resolver.withNodeScopeContext(ctx), PaginatedQuery{Query: &query, Pagination: args.Pagination})
+	return resolver.root.componentsV2(resolver.nodeScopeContext(), PaginatedQuery{Query: &query, Pagination: args.Pagination})
```
charmik-redhat, line 322: Does this need if resolver.ctx == nil check ?

central/graphql/resolvers/nodes.go
```diff
@@ -330,25 +331,25 @@ func (resolver *nodeResolver) ComponentCount(ctx context.Context, args RawQuery)
 
 	query := search.AddRawQueriesAsConjunction(args.String(), resolver.getNodeRawQuery())
 
-	return resolver.root.componentCountV2(resolver.withNodeScopeContext(ctx), RawQuery{Query: &query})
+	return resolver.root.componentCountV2(resolver.nodeScopeContext(), RawQuery{Query: &query})
```
charmik-redhat, line 334: Same as above

central/graphql/resolvers/nodes.go
```diff
@@ -523,20 +542,26 @@ func (resolver *nodeResolver) PlottedNodeVulnerabilities(ctx context.Context, ar
...
 func (resolver *nodeResolver) Scan(ctx context.Context) (*nodeScanResolver, error) {
-	res, err := resolver.root.wrapNodeScan(resolver.data.GetScan(), true, nil)
+	res, err := resolver.root.wrapNodeScanWithContext(resolver.ctx, resolver.data.GetScan(), true, nil)
+	if resolver.ctx == nil {
+		resolver.ctx = ctx
+	}
 	if err != nil || res == nil {
 		return nil, err
 	}
-	res.ctx = resolver.withNodeScopeContext(ctx)
```
charmik-redhat, line 530: Should leave this because node scope would be needed inside nodeScanResolver's sub resolvers. You can use resolver.ctx instead of ctx while adding node scope.

charmik-redhat (approved):

PR #3292, opened 2022-10-03
Do not depend on tools in system-/user-wide paths (or install tools into those)

misberner:
## Description

This is a long-standing pet peeve of mine: most of the Go-based tools that we use (`golangci-lint`, `mockgen` etc.) are installed in system-wide or user-wide paths, e.g., the default `$GOBIN` (which is `$HOME/go/bin`). With other projects having a similar annoying behavior, this in the best case leads to frequent rebuilds of tools with changing versions, and in the worst case causes code to be generated with incorrect or even inconsistent versions of tools, causing unstable behavior (that is hopefully caught in CI via the generated files up-to-date check).

It doesn't need to be this way. This PR changes things in such a way that (a) Go tools are installed in a project-specific location (`.gotools/bin`) and (b) sets up `Makefile` rules to ensure that tools are always rebuilt when necessary (version bump, code change), but - in most cases - not rebuilt more often than necessary. Note that the last point is not a strict requirement, because `go install` already caches most of its work, so running `go install` repeatedly is really fast (~0.2s on my machine). However, that's still worse than 0s, which is why whenever we can easily prevent rebuilds by correct formulation of `Makefile` targets, we will.

The functionality is exposed via a reusable macro invoked via `$(call go-tool, ...)`. Please refer to the [documentation of that](https://github.com/stackrox/stackrox/pull/3292/files#diff-6680f14ebf202ac4f73d3741109aa75cdd3aad8fe66b1cbc9e389377694b896dR1) for further details.

This PR affects just the tools referenced in the top-level Makefile. For tools used in `protogen.mk`, the situation is already different; I plan to unify things in a follow-up PR and use `$(call go-tool, ...)` in `protogen.mk` as well.

## Checklist

## Testing Performed

- Local testing with `make reinstall-dev-tools` and the individual targets
- CI
- [ ] needs status check: Downstream testing with this commit at https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/285

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,47 +111,16 @@ all: deps style test image
 ###### Binaries we depend on (need to be defined on top) ############
 #####################################################################
 
-GOLANGCILINT_BIN := $(GOBIN)/golangci-lint
-$(GOLANGCILINT_BIN): deps
-	@echo "+ $@"
-	@cd tools/linters/ && go install github.com/golangci/golangci-lint/cmd/golangci-lint
+include make/gotools.mk
 
-EASYJSON_BIN := $(GOBIN)/easyjson
-$(EASYJSON_BIN): deps
-	$(SILENT)echo "+ $@"
-	go install github.com/mailru/easyjson/easyjson
-
-CONTROLLER_GEN_BIN := $(GOBIN)/controller-gen
-$(CONTROLLER_GEN_BIN): deps
-	$(SILENT)echo "+ $@"
-	@# We need to install a legacy version for compatibility reasons.
-	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
-
-ROXVET_BIN := $(GOBIN)/roxvet
-.PHONY: $(ROXVET_BIN)
-$(ROXVET_BIN): deps
-	@echo "+ $@"
-	go install ./tools/roxvet
-
-STRINGER_BIN := $(GOBIN)/stringer
-$(STRINGER_BIN): deps
-	@echo "+ $@"
-	go install golang.org/x/tools/cmd/stringer
-
-MOCKGEN_BIN := $(GOBIN)/mockgen
-$(MOCKGEN_BIN): deps
-	@echo "+ $@"
-	go install github.com/golang/mock/mockgen
-
-GO_JUNIT_REPORT_BIN := $(GOBIN)/go-junit-report
-$(GO_JUNIT_REPORT_BIN): deps
-	@echo "+ $@"
-	$(SILENT)cd tools/test/ && go install github.com/jstemmer/go-junit-report/v2
-
-PROTOLOCK_BIN := $(GOBIN)/protolock
-$(PROTOLOCK_BIN): deps
-	@echo "+ $@"
-	$(SILENT)cd tools/linters/ && go install github.com/nilslice/protolock/cmd/protolock
+$(call go-tool, GOLANGCILINT_BIN, github.com/golangci/golangci-lint/cmd/golangci-lint, tools/linters)
+$(call go-tool, EASYJSON_BIN, github.com/mailru/easyjson/easyjson)
+$(call go-tool, CONTROLLER_GEN_BIN, sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
+$(call go-tool, ROXVET_BIN, ./tools/roxvet)
+$(call go-tool, STRINGER_BIN, golang.org/x/tools/cmd/stringer)
+$(call go-tool, MOCKGEN_BIN, github.com/golang/mock/mockgen)
+$(call go-tool, GO_JUNIT_REPORT_BIN, github.com/jstemmer/go-junit-report/v2, tools/test)
+$(call go-tool, PROTOLOCK_BIN, github.com/nilslice/protolock/cmd/protolock, tools/linters)
 
 ###########
 ## Style ##
@@ -166,14 +135,14 @@ ifdef CI
 	@echo 'The environment indicates we are in CI; running linters in check mode.'
 	@echo 'If this fails, run `make style`.'
 	@echo "Running with no tags..."
-	golangci-lint run
+	$(GOLANGCILINT_BIN) run
 	@echo "Running with release tags..."
 	@# We use --tests=false because some unit tests don't compile with release tags,
 	@# since they use functions that we don't define in the release build. That's okay.
-	golangci-lint run --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
+	$(GOLANGCILINT_BIN) run --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
 else
-	golangci-lint run --fix
-	golangci-lint run --fix --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
+	$(GOLANGCILINT_BIN) run --fix
+	$(GOLANGCILINT_BIN) run --fix --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
 endif
 
 .PHONY: qa-tests-style
@@ -254,7 +223,7 @@ no-large-files:
 .PHONY: storage-protos-compatible
 storage-protos-compatible: $(PROTOLOCK_BIN)
 	@echo "+ $@"
-	$(SILENT)protolock status -lockdir=$(BASE_DIR)/proto/storage -protoroot=$(BASE_DIR)/proto/storage
+	$(SILENT)$(PROTOLOCK_BIN) status -lockdir=$(BASE_DIR)/proto/storage -protoroot=$(BASE_DIR)/proto/storage
 
 .PHONY: blanks
 blanks:
@@ -294,10 +263,11 @@ include make/protogen.mk
 .PHONY: go-easyjson-srcs
 go-easyjson-srcs: $(EASYJSON_BIN)
 	@echo "+ $@"
-	$(SILENT)easyjson -pkg pkg/docker/types/types.go
-	$(SILENT)easyjson -pkg pkg/docker/types/container.go
-	$(SILENT)easyjson -pkg pkg/docker/types/image.go
-	$(SILENT)easyjson -pkg pkg/compliance/compress/compress.go
+	@# Files are ordered such that repeated runs of `make go-easyjson-srcs` don't create diffs.
+	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/image.go
+	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/container.go
+	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/types.go
+	$(SILENT)$(EASYJSON_BIN) pkg/compliance/compress/compress.go
 
 .PHONY: clean-easyjson-srcs
 clean-easyjson-srcs:
@@ -313,7 +283,7 @@ pkg/complianceoperator/api/v1alpha1/zz_generated.deepcopy.go: $(CONTROLLER_GEN_B
 .PHONY: go-generated-srcs
 go-generated-srcs: deps clean-easyjson-srcs go-easyjson-srcs $(MOCKGEN_BIN) $(STRINGER_BIN) pkg/complianceoperator/api/v1alpha1/zz_generated.deepcopy.go
 	@echo "+ $@"
-	PATH="$(PATH):$(BASE_DIR)/tools/generate-helpers" go generate -v -x ./...
+	PATH="$(PATH):$(BASE_DIR)/tools/generate-helpers" MOCKGEN_BIN="$(MOCKGEN_BIN)" go generate -v -x ./...
 
 proto-generated-srcs: $(PROTO_GENERATED_SRCS) $(GENERATED_API_SWAGGER_SPECS)
 	@echo "+ $@"
@@ -551,8 +521,17 @@ integration-unit-tests: build-prep test-prep
 		$(shell go list ./... | grep  "registries\|scanners\|notifiers") \
 		| tee $(GO_TEST_OUTPUT_PATH)
 
-generate-junit-reports: $(GO_JUNIT_REPORT_BIN)
-	$(BASE_DIR)/scripts/generate-junit-reports.sh
+.PHONY: generate-junit-reports
+generate-junit-reports: junit-reports/report.xml
+
+$(GO_TEST_OUTPUT_PATH):
+	@echo "The test output log cannot be created via a direct Makefile rule. You must make the desired test targets"
+	@echo "first to ensure the file's existence."
+	@exit 1
+
+junit-reports/report.xml: $(GO_TEST_OUTPUT_PATH) $(GO_JUNIT_REPORT_BIN)
+	@mkdir -p junit-reports
+	$(SILENT)$(GO_JUNIT_REPORT_BIN) <"$<" >"$@"
 
 ###########
 ## Image ##
@@ -768,11 +747,9 @@ docs-tag:
 scanner-tag:
 	@cat SCANNER_VERSION
 
-GET_DEVTOOLS_CMD := $(MAKE) -qp | sed -e '/^\# Not a target:$$/{ N; d; }' | egrep -v '^(\s*(\#.*)?$$|\s|%|\(|\.)' | egrep '^[^[:space:]:]*:' | cut -d: -f1 | sort | uniq | grep '^$(GOBIN)/'
 .PHONY: clean-dev-tools
-clean-dev-tools:
+clean-dev-tools: gotools-clean
 	@echo "+ $@"
-	$(SILENT)$(GET_DEVTOOLS_CMD) | xargs rm -fv
 
 .PHONY: reinstall-dev-tools
 reinstall-dev-tools: clean-dev-tools
@@ -780,10 +757,8 @@ reinstall-dev-tools: clean-dev-tools
 	$(SILENT)$(MAKE) install-dev-tools
 
 .PHONY: install-dev-tools
-install-dev-tools:
+install-dev-tools: gotools-all
 	@echo "+ $@"
-	$(SILENT)test -n "$(GOPATH)" || { echo "Set GOPATH before installing dev tools"; exit 1; }
-	$(SILENT)$(GET_DEVTOOLS_CMD) | xargs $(MAKE)
 ifeq ($(UNAME_S),Darwin)
 	@echo "Please manually install RocksDB if you haven't already. See README for details"
 endif
--- a/make/gotools.mk
+++ b/make/gotools.mk
@@ -1,0 +1,128 @@
+# gotools.mk
+# Simplified installation & usage of Go-based tools
+#
+# Input variables:
+#   GOTOOLS_PROJECT_ROOT: the project root directory; defaults to $(CURDIR)
+#   GOTOOLS_ROOT: the directory in which this file stores auxiliary data (should be .gitignore'd); defaults to
+#                 $(GOTOOLS_PROJECT_ROOT)/.gotools
+#   GOTOOLS_BIN: the directory in which binaries are stored; defaults to $(GOTOOLS_ROOT)/bin.
+#
+# This file defines a single (user-facing) macro, `go-tool`, which can be invoked via
+#   $(call go-tool VARNAME, go-pkg [, module-root])
+# where go-pkg can be:
+# - an absolute Go import path with an explicit version, e.g.,
+#     github.com/golangci/golangci-lint/cmd/golangci-lint@v1.49.0. In this case, the tool is installed via `go install`,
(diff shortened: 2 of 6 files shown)


janisz (approved):

tools/generate-helpers/mockgen-wrapper
```diff
@@ -25,6 +25,10 @@ set -o pipefail
 #   This does not work when generating mocks for code that will be generated by
 #     another go:generate command.
 
+PROJECT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+MOCKGEN_BIN="${PROJECT_ROOT}/.gotools/bin/mockgen"
+[[ -x "$MOCKGEN_BIN" ]] || make -C "$PROJECT_ROOT" "$MOCKGEN_BIN"
```
porridge, line 30: This is a bit ugly:
- requires knowing the `./gotools/...` path
- will not DTRT if the version of mockgen changes in `go.mod`

How about introducing a `mockgen: $(MOCKGEN_BIN)` target in the `Makefile` and calling `make -C "$PROJECT_ROOT" mockgen` here **unconditionally**?

porridge (approved):
Niiiiiiiiice! This is something I will definitely remember to steal for other projects :-)
Some nitpicks inline.

porridge:
Do you have plans to also use this in `operator/Makefile`?

misberner:
> Do you have plans to also use this in `operator/Makefile`?

Possibly. There are some trade-offs to consider, the main one being that we diverge further from the `operator-sdk` scaffolding's Makefile. And I have to see how well the rather idiosyncratic way of installing tools (which is however necessary due to the usage of `replace` directives in the respective `go.mod` files). We could solve that by just having the versions listed in a `operator/tools/go.mod` file, but this would only increase the divergence.

tools/generate-helpers/mockgen-wrapper
misberner: Yeah, I'll add something like `${tool-basename}-bin` or `install-${tool-basename}` phony target.

porridge: Why everywhere? If `which-mockgen` only depends on `$(MOCKGEN_BIN)` then you only need to make sure `>&2` is on those `@echo "+ $$(notdir $$@)"` lines in `gotools.mk`? What am I missing?
If there indeed is another reason, then I guess it's acceptable, as long as:
- it works correctly in all funny `-C` cases
- you document the reason in a comment here :-)

make/gotools.mk
```diff
@@ -0,0 +1,126 @@
...
+# This file also defines two static, global targets:
+#   gotools-clean: this removes all gotools-related data
+#   gotools-all: this builds all gotools.
+
+GOTOOLS_PROJECT_ROOT ?= $(CURDIR)
+GOTOOLS_ROOT ?= $(GOTOOLS_PROJECT_ROOT)/.gotools
+GOTOOLS_BIN ?= $(GOTOOLS_ROOT)/bin
+
+PATH := $(GOTOOLS_BIN):$(PATH)
+export PATH
```
msugakov, line 53: Does this have any effect on the parent `Makefile`? I suppose that's the way due to which `generate-junit-reports.sh` works when called from the `Makefile`. This seems too sneaky, implicit and prone to differences between user-default environment and the environment constructed from the `Makefile`.

It seems, that you're constructing something similar to Python's VirtualEnv here. VirtualEnv has `activate` script which developers source in shell, and get all local binaries available for interactive use.
Maybe we don't need to be that fancy and just hardcode binaries to be called from `.gotools/` directory. If the binary is there - the script would use it. If the binary isn't there, the developer needs to get it through invoking `Makefile`.

make/gotools.mk
```diff
@@ -0,0 +1,126 @@
...
+# To make the tool accessible in the canonical location, we create a symlink. This only depends on the versioned path,
+# i.e., only needs to be recreated when the version is bumped.
+$$(_gotools_canonical_bin_path): $$(_gotools_versioned_bin_path)
+	@mkdir -p "$(GOTOOLS_BIN)"
+	$$(SILENT)ln -sf "$$<" "$$@"
+
+else
+
+# Tool is specified with an absolute path without a version. Take info from go.mod file in the respective directory.
+$$(_gotools_canonical_bin_path): $$(_gotools_mod_root)/go.mod
```
msugakov, line 103: Why not also `go.sum`?

pkg/docker/types/types_easyjson.go
```diff
@@ -65,7 +62,7 @@ func easyjson6601e8cdDecodeGithubComStackroxRoxPkgDockerTypes(in *jlexer.Lexer,
 				}
 				for !in.IsDelim(']') {
 					var v1 ContainerJSON
-					easyjson6601e8cdDecodeGithubComStackroxRoxPkgDockerTypes1(in, &v1)
+					(v1).UnmarshalEasyJSON(in)
```
msugakov, line 65: Do you know, why is there such a large diff on `types_easyjson.go`?

msugakov (commented):
`gotools.mk` Makefile magic is too much beyond my knowledge. Don't know if I am ready to dive that deep.

tools/generate-helpers/mockgen-wrapper
misberner: Well, I guess I'd also have to apply it to the `go install` commands, no? I don't know if I can rely on the fact that they print all output to `stderr`, as there is no `stdout` output that would be usable for `$(...)` command substitution anyway.

make/gotools.mk
misberner: `go.mod` used to be mostly a requirements file. This changed in (I believe) go 1.17, now `go.mod` contains all dependencies, direct and transitive, and `go.sum` is mostly just for checksum verification (`go mod verify`). Therefore, no version of a tool dependency can change without a change in `go.mod`.

I guess you could argue that a tool should be rebuilt when the checksum of the respective module changes. But checksum changes are error cases, they should not normally happen and require human review and intervention (probably resulting in a `go.mod` change) anyway.

pkg/docker/types/types_easyjson.go
misberner: Yeah, this was a fun puzzle to solve while working on this: I observed that the generation apparently was unstable, i.e., I ran `make go-easyjson-srcs`, checked in the code, and saw the "check that generated sources are up to date" check fail. At first I thought there must be some version mismatch. Then I ran the whole `make go-generated-srcs` (which I almost never run locally, and also didn't in this PR before), and voila, I got matching easyjson sources again. Then I re-ran `make go-easyjson-srcs`, and it broke.

I then took a closer look and saw that `make go-generated-srcs` depends on `clean-easyjson-srcs`, and that `go-easyjson-srcs` generates the "right" (wrt. CI) sources whenever the easyjson srcs don't exist. I looked up the `easyjson` help, and saw the help text
```
  -pkg
        process the whole package instead of just the given file
```
so I became suspicious that all files are listed individually. Apparently, "process" means "generate optimized marshal/unmarshal code", but the exported methods will still only be generated for the annotated types specified in the `.go` file. However, `-pkg` also causes `easyjson` to pick up any optimized marshaling/unmarshaling code that's already there in other files for some of the used types (but if it doesn't find optimized methods, it will generate marshaling/unmarshaling for types outside of the `.go` file with unexported names). Running `make go-easyjson-srcs` when the sources were already there just picked up those types/methods that weren't picked up by it running in CI in the course of `make go-generated-srcs`, that cleaned them beforehand. So I changed the generation order to process files in topological order, making sure that easyjson generation for types in one file has always finished by the time the file in which those types are used is processed.

This is ugly, to say the least, and extremely brittle - especially because you absolutely can have circular dependencies between `.go` files, and thus you can break this in a way that's not only hard but impossible to repair (well, I guess one could always run all generation commands twice). But other than put a warning in the `Makefile` to honor the order, I don't know a better way to fix it. And before you ask, no - specifying `-pkg` and then _all_ files in the directory doesn't work, either. This would probably be a good change in `easyjson`, but unfortunately, I [didn't have any luck with PRs in the past](https://github.com/mailru/easyjson/pull/376).

So yeah, the diff is IMHO what it should be, without the duplicate generated code that was there before. It's also stable now, in the sense that running `make go-easyjson-srcs` always gives the same result, regardless of whether easyjson files are there or not. But I'm not at all happy at how I've reached that state.

make/gotools.mk
misberner: > Does this have any effect on the parent `Makefile`?

TTBOMK, yes, `include` in `Makefiles` are just inserted inline.

> This seems too sneaky, implicit and prone to differences between user-default environment and the environment constructed from the `Makefile`.

Is this about this happening in an `include`d file, or about the `PATH` being modified in the `Makefile`? It sounds like the latter, and I don't think I agree. "User-default environment" is an anti-pattern when it comes to building software, the build logic has every right to shield itself from the user-default environment, and that necessarily means deviating from it. One of the points of using a `Makefile` is that it provides a uniform way and entry-point to build commands. If somebody can't just copy something out of the `Makefile` into their terminal without observing different behavior, well, tough luck, the `Makefile` is there for a reason (yes, there are cases when you want to run stuff directly in the terminal, but those are corner-cases).

In the concrete case, I don't see any evidence that `generate-junit-reports.sh` is intended to be invoked outside of the respective `Makefile` rule, and just because it's there doesn't mean you can just run it and expect it to work. Is your concern that it's easy to run this script and get the _wrong_ result (using a different version) without noticing? Would you be okay with instead `export` the `XYZ_BIN` variables, and do a `[[ -x "$GO_JUNIT_REPORT_BIN" ]] || die "..."` in the script? I don't think there's a big difference in the two approaches, in both cases we rely on environment set up by the `Makefile`. The one advantage I can see is that we now _reserve_ the invocation by basename only for when we depend on the user-default environment, so it makes things somewhat less ambiguous.

But there will always be cases when we can't do this, for example, if we invoke a tool that we don't control that depends on the `PATH`. We invoke our protobuf compilation with `protoc ... --gofast_out=...`. This causes `protoc` to look for a `protoc-gen-gofast` binary in the path (generally, `--{x}_out` causes `protoc-gen-{x}` to be used). Here, it seems a requirement to set a custom `PATH`.

Makefile
```diff
@@ -111,56 +111,18 @@ all: deps style test image
...
-PROTOLOCK_BIN := $(GOBIN)/protolock
-$(PROTOLOCK_BIN): deps
-	@echo "+ $@"
-	$(SILENT)cd tools/linters/ && go install github.com/nilslice/protolock/cmd/protolock
+include make/gotools.mk
+
+$(call go-tool, GOLANGCILINT_BIN, github.com/golangci/golangci-lint/cmd/golangci-lint,tools/linters)
+$(call go-tool, EASYJSON_BIN, github.com/mailru/easyjson/easyjson)
+$(call go-tool, CONTROLLER_GEN_BIN, sigs.k8s.io/controller-tools/cmd/controller-gen)
+$(call go-tool, GOVERALLS_BIN, github.com/mattn/goveralls,tools/test)
```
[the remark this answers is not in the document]
msugakov: That was https://github.com/stackrox/stackrox/pull/3375 . Resolving.

make/gotools.mk
msugakov: With regards to `generate-junit-reports.sh` file, I suggest to simply inline its content in the `Makefile`. That would be correct according to the logic that the `Makefile` should be the entry point that you propose. There `$GO_JUNIT_REPORT_BIN` can be safely used instead of `go-junit-report`.

make/gotools.mk
msugakov: Thanks, I did not know about that go 1.17 change.

> checksum changes are error cases, they should not normally happen and require human review and intervention

I suppose that's only for packages that are cached in Go Module Mirror. Packages that are consumed directly from git(hub) which are mostly the ones under github.com/stackrox/, can have their tags force-pushed which would change `go.sum` but not change in `go.mod`.

Several probabilities need to be multiplied to get the overall chance of such issue affecting anyone. However, thinking that the issue may happen specifically during local development, it could be very annoying to run into such edge case and see binaries not updated when you expect them to.

No strong feelings as the chance is low and one has an option to just nuke `.gotools/` in anger when things go wrong, which should recover the situation.

pkg/docker/types/types_easyjson.go
msugakov: Thanks for the explanation. I find `-pkg` to be very confusing. I would expect that it can be sufficient to run `$(SILENT)$(EASYJSON_BIN) -pkg pkg/docker/types/<blah>.go` only for **one file** in that directory and it will pick up two others as they have matching `package types`, but that doesn't happen.

Moreover, if I take your branch and remove `-pkg` from arguments like this

```makefile
.PHONY: go-easyjson-srcs
go-easyjson-srcs: $(EASYJSON_BIN)
	@echo "+ $@"
	@# Don't change the order of these, as otherwise the generation might become unstable.
	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/image.go
	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/container.go
	$(SILENT)$(EASYJSON_BIN) pkg/docker/types/types.go
	$(SILENT)$(EASYJSON_BIN) pkg/compliance/compress/compress.go
```

I see the diff is empty after running `make clean-easyjson-srcs` and `make go-easyjson-srcs` in any combinations.

Looks like `-pkg` doesn't have any effect for us. Therefore I suggest dropping it to avoid misleading ourselves.

Makefile
```diff
@@ -292,10 +254,11 @@ include make/protogen.mk
 .PHONY: go-easyjson-srcs
 go-easyjson-srcs: $(EASYJSON_BIN)
 	@echo "+ $@"
-	$(SILENT)easyjson -pkg pkg/docker/types/types.go
-	$(SILENT)easyjson -pkg pkg/docker/types/container.go
-	$(SILENT)easyjson -pkg pkg/docker/types/image.go
-	$(SILENT)easyjson -pkg pkg/compliance/compress/compress.go
+	@# Don't change the order of these, as otherwise the generation might become unstable.
```
msugakov, line 257: I suggest to reword this comment to make it more actionable for future generations:
```suggestion
	@# Files are ordered such that repeated runs of `make go-easyjson-srcs` don't create diffs.
```
Or something like this.

pkg/docker/types/types_easyjson.go
misberner: I agree with your expectation on what `-pkg` does, and the frustration about this not being the reality. The fact that it isn't even needed in our case leaves me completely puzzled as to what its point is. Will remove.

make/gotools.mk
misberner: Yes, the checksum can change in such cases, but I don't know what Go would to do then. I have a hard time imagining that it would silently change the checksum. It would probably ask you to remove the offending checksum entry manually or through a go tool invocation.

However, there is probably no harm at all in specifying `go.sum` in addition (I wanted to avoid a spurious up to date when somebody changed the `go.mod` file w/o running `go mod tidy` afterwards, but nobody says I need to remove `go.mod` as a dep). So added it as well to make it more robust.

As a side note, the `GOPRIVATE` stems from the time when our repos were private. We should probably remove it (will happen in a different PR, though) and get everything through the Go package proxy.

make/gotools.mk
misberner: Yeah, good point about remove `generate-junit-reports.sh`. I've removed it and also made the respective rules more `make`-y`.

Good point about `PATH` being potentially unneeded, especially after the above change. I have removed it now, let's see if things break.

We still have the reference to the tool in the `mockgen-wrapper` script. Do you have an opinion on the `which-<tool>` target I sketched in the conversation with Marcin above?

Makefile
```diff
@@ -111,56 +111,18 @@ all: deps style test image
...
+
+$(call go-tool, GOLANGCILINT_BIN, github.com/golangci/golangci-lint/cmd/golangci-lint,tools/linters)
+$(call go-tool, EASYJSON_BIN, github.com/mailru/easyjson/easyjson)
+$(call go-tool, CONTROLLER_GEN_BIN, sigs.k8s.io/controller-tools/cmd/controller-gen)
+$(call go-tool, GOVERALLS_BIN, github.com/mattn/goveralls,tools/test)
+$(call go-tool, ROXVET_BIN, ./tools/roxvet)
+$(call go-tool, STRINGER_BIN, golang.org/x/tools/cmd/stringer)
+$(call go-tool, MOCKGEN_BIN, github.com/golang/mock/mockgen)
+$(call go-tool, GENNY_BIN, github.com/mauricelam/genny)
+$(call go-tool, GO_JUNIT_REPORT_BIN, github.com/jstemmer/go-junit-report/v2,tools/test)
```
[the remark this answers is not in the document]
misberner: Fixed now

Makefile
misberner: Done

Makefile
```diff
@@ -111,56 +111,18 @@ all: deps style test image
...
+$(call go-tool, PROTOLOCK_BIN, github.com/nilslice/protolock/cmd/protolock,tools/linters)
```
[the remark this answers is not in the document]
misberner: Done

make/gotools.mk
```diff
@@ -0,0 +1,126 @@
...
+# This file defines a single (user-facing) macro, `go-tool`, which can be invoked via
+#   $(call go-tool VARNAME, go-pkg [, module-root])
+# where go-pkg can be:
+# - an absolute Go import path with an explicit version, e.g.,
+#     github.com/golangci/golangci-lint/cmd/golangci-lint@v1.49.0. In this case, the tool is installed via `go install`,
+#     and module information from the local workspace is ignored, in accordance with the normal behavior of go install
+#     with an explicit version given.
+# - an absolut Go import path WITHOUT a version, e.g., github.com/golangci/golangci-lint/cmd/golangci-lint. In this
+#     case, the tool is installed via `go install` from the module rooted at $(GOTOOLS_PROJECT_ROOT), or, if
+#     module-root is given, from the module rooted at that (relative) path. I.e., go-pkg must be a provided by a module
```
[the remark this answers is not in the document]
misberner: I just removed the stray "a". It is still the module providing the package that must be listed in `go.mod`, not the package itself

make/gotools.mk
```diff
@@ -0,0 +1,126 @@
...
+#                to PATH.
+#
+# This file defines a single (user-facing) macro, `go-tool`, which can be invoked via
+#   $(call go-tool VARNAME, go-pkg [, module-root])
+# where go-pkg can be:
+# - an absolute Go import path with an explicit version, e.g.,
+#     github.com/golangci/golangci-lint/cmd/golangci-lint@v1.49.0. In this case, the tool is installed via `go install`,
+#     and module information from the local workspace is ignored, in accordance with the normal behavior of go install
+#     with an explicit version given.
+# - an absolut Go import path WITHOUT a version, e.g., github.com/golangci/golangci-lint/cmd/golangci-lint. In this
```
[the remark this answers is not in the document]
misberner: Done

make/gotools.mk
msugakov: Would be nice to remove `GOPRIVATE` because I've seen GitHub throwing 429-s when we try to build the product in CI. Do you plan (have time) doing it yourself?

tools/generate-helpers/mockgen-wrapper
msugakov: I suppose you won't be extremely happy in a third opinion.
I'd suggest to take it as granted that `.gotools/` is the location for binaries and binaries are expected there.
This fragment would look like this then:

```sh
MOCKGEN_BIN="${PROJECT_ROOT}/.gotools/bin/mockgen"
if [[ ! -x "$MOCKGEN_BIN" ]]; then
  die "Did not detect mockgen executable, run the following command to materialize it: make -C $PROJECT_ROOT $MOCKGEN_BIN"
fi
```

Alternatively, I'm for `which-mockgen` approach which does the same but is robust against executable renames. I think, a side effect of downloading and building something is a bit too heavy especially since that may fail e.g. due to network outage.

make/gotools.mk
msugakov: Thanks.
Commented on `mockgen-wrapper` thread.
Resolving this one.

Makefile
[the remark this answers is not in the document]
msugakov: Since this is done, I'd resolve this thread to save space and scrolling time.

make/gotools.mk
msugakov: https://github.com/stackrox/stackrox/pull/3402

msugakov (approved):
I have no concerns that this is merged as-is.

misberner:
I have taken the `which-<binary>` approach and used that in `mockgen-wrapper`. I have realized though that sometimes we call tools directly in `go:generate` without a wrapper script, so I will send out a companion change to `github.com/stackrox/workflow` to add `<project root>/.gotools/bin` to the `PATH` when running `gogen`.

Since everyone has approved, I have made no changes that should come as a surprise, and CI is passing, I will merge this in the course of today.

PR #3294, opened 2022-10-04
ROX-11927:Remove unsupported 'License' sub resolver from NodeComponent graphQL

ksurabhi91:
Removed license field from node components graphql query. Tested graphql query on  postgres db cluster
Screen Shot 2022-10-04 at 5 23 40 PM

### DIFF
--- a/central/graphql/resolvers/node_components.go
+++ b/central/graphql/resolvers/node_components.go
@@ -25,7 +25,6 @@ func init() {
 		schema.AddExtraResolvers("NodeComponent", []string{
 			"fixedIn: String!",
 			"lastScanned: Time",
-			"license: License",
 			"location(query: String): String!",
 			"nodes(query: String, scopeQuery: String, pagination: Pagination): [Node!]!",
 			"nodeCount(query: String, scopeQuery: String): Int!",
@@ -50,7 +49,6 @@ type NodeComponentResolver interface {
 	FixedIn(ctx context.Context) string
 	Id(ctx context.Context) graphql.ID
 	LastScanned(ctx context.Context) (*graphql.Time, error)
-	License(ctx context.Context) (*licenseResolver, error)
 	Location(ctx context.Context, args RawQuery) (string, error)
 	Name(ctx context.Context) string
 	Nodes(ctx context.Context, args PaginatedQuery) ([]*nodeResolver, error)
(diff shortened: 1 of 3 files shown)


vjwilson (commented):
Does this change pass `ui-e2e-tests` when the Postgres flag is OFF, too?

keyallis:
@vjwilson NodeComponent is a graphql only resolver so it shouldn't cause issue in non-postgres world

ksurabhi91:
@vjwilson @keyallis  does CI test on postgres cluster ?

vjwilson:
> @vjwilson NodeComponent is a graphql only resolver so it shouldn't cause issue in non-postgres world

We use GraphQL in the non-Postgres app, too, in several sections.

The two instances I see where this is referenced are 

- https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/ConfigManagement/Entity/Image.js#L71
- https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/queries/image.js#L36

I'm running @ksurabhi91 's PR image locally, and it seems to run fine, those fields just come back as `null`, and they are no longer used in the display template anywhere, apparently.
Since that's the case, we should probably remove those references in the two JS files, just to prevent "code-rot".

Screen Shot 2022-10-05 at 1 21 03 PM

vjwilson:
> @vjwilson @keyallis does CI run tests on postgres cluster ?

Yes, CI for PRs, and their subsequent master merges, run with all feature flags ON.

The nightly and release builds run with all feature flags OFF, so if we merge this, it will get tested with Postgres OFF overnight.

Based on my manual testing, I think it will be fine--I was afraid GraphQL would be too strict and throw errors with Postgres OFF, but it's not.. See my reply to Oscar's comment.

theencee:
+1 to what Van said. Please remove the two references in JS as well 

```
license {
                            name
                            type
                            url
                        }
```

Thanks for testing @vjwilson

vjwilson (approved):
FE changes
/lgtm

ksurabhi91:
@keyallis @charmik-redhat all tests have passed. can you please approve this CR

charmik-redhat (approved):
Lgtm. Nice work.

keyallis (approved):
LGTM

theencee (approved):

PR #3297, opened 2022-10-04
ROX-12837: Add Feature flag for the Full Node Scanning

vikin91:
## Description

This PR adds a new feature-flag. The flag is unused yet, but will be used in the context of epic [ROX-10818](https://issues.redhat.com/browse/ROX-10818)

## Checklist

## Testing Performed

- CI is sufficient

### DIFF
--- a/pkg/features/list.go
+++ b/pkg/features/list.go
@@ -43,6 +43,6 @@ var (
 	// NetworkGraphPatternFly enables the PatternFly version of NetworkGraph. (used in the front-end app only)
 	NetworkGraphPatternFly = registerFeature("Enable PatternFly version of NetworkGraph", "ROX_NETWORK_GRAPH_PATTERNFLY", false)
 
-	// FullNodeScanning RHCOS full node scanning function
-	FullNodeScanning = registerFeature("Enable RHCOS full node scanning function", "ROX_FULL_NODE_SCANNING", false)
+	// RHCOSNodeScanning enables phase 1 functions of "Full host level vulnerability scanning for RHCOS nodes" (ROX-10818)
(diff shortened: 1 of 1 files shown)


pkg/features/list.go
```diff
@@ -42,4 +42,7 @@ var (
 
 	// NetworkGraphPatternFly enables the PatternFly version of NetworkGraph. (used in the front-end app only)
 	NetworkGraphPatternFly = registerFeature("Enable PatternFly version of NetworkGraph", "ROX_NETWORK_GRAPH_PATTERNFLY", false)
+
+	// FullNodeScanning RHCOS full node scanning function
+	FullNodeScanning = registerFeature("Enable RHCOS full node scanning function", "ROX_FULL_NODE_SCANNING", false)
```
msugakov, line 47: Nit: node scanning is more than RHCOS node scanning. Also, the level of scanning we're implementing is not "full node" but rather OS, kernel versions and rpm packages.
I suggest to have some more specific name here.

Maddosaurus (approved):
I'd add a +1 to @msugakov's nit, otherwise that's a great addition!

pkg/features/list.go
vikin91: I am fine renaming this. Below an explanation why I am using the `Full` word in the names of other objects in this epic.

I am using `Full` in the namer (here, the proto message) as we aim at full node scan in the near future (including language components). Moreover, we do have a proto message called `NodeScan` already (for the partial node scan that is currently available) and I want to avoid confusing these two flows.

msugakov (approved):

PR #3298, opened 2022-10-04
Introduce Compliance Full Node Scan message

Maddosaurus:
## Description

As a prereq for later work, a new Protobuf message is introduced to represent a full node scan result (generated in compliance).
When the timer fires, Compliance scans the Node and then sends a `FullNodeScanRequest` with the scan results to Scanner.

## Checklist

As this PR only introduces a new protobuf message, no specific upgrade steps or user-facing changes are introduced

## Testing Performed
None


proto/internalapi/sensor/compliance_nodescan.proto
```diff
@@ -0,0 +1,22 @@
...
+
+
+import "gogoproto/gogo.proto";
+
+package sensor;
+
+// Names and data types are designed to follow proto/storage/node.proto as close as possible
+// Next tag: 7
+message FullNodeScan {
+    string                         id         = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk"'];
```
msugakov, line 17: 1. I suggest to clarify that's node id to prevent confusing reader that the id may belong to some other entity, e.g. scan id.

```suggestion
    string                         node_id         = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk"'];
```

2. I'm not sure if `gogoproto.moretags` will be used for this message, but at this point they don't hurt and we can remove them later.

proto/internalapi/sensor/compliance_nodescan.proto
```diff
@@ -0,0 +1,22 @@
...
+    string                         name       = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
```
msugakov, line 18: Same for node name
```suggestion
    string                         node_name       = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
```

proto/internalapi/sensor/compliance_nodescan.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
```
msugakov, line 1: About the file name `compliance_nodescan.proto`. I think, the fact that we put Scanner in Compliance is just an implementation detail that we shouldn't promote to the contract. As a thought exercise, `compliance_` part will become misleading if we decide to set up a dedicated DaemonSet for node scanner.

proto/internalapi/sensor/compliance_nodescan_iservice.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+option go_package = "sensor";
+
+import "internalapi/sensor/compliance_nodescan.proto";
+
+package sensor;
+
+message FullNodeScanRequest {
+    string       node = 1;
```
msugakov, line 10: How did you intend to use `string node`?

proto/internalapi/sensor/compliance_nodescan.proto
RTann: I think it would make more sense to actually put this in `proto/storage/node.proto` next to `Node` and the older `NodeScan`. That way we can store it in Central once all the vulns are filled in, and we can use it in other places as well. It also decouples it from the internal API. I believe `internalapi/...` can import from `storage/`, right?

proto/internalapi/sensor/compliance_nodescan.proto
```diff
@@ -0,0 +1,22 @@
...
+import "scanner/api/v1/component.proto";
+
+
+import "gogoproto/gogo.proto";
+
+package sensor;
+
+// Names and data types are designed to follow proto/storage/node.proto as close as possible
+// Next tag: 7
+message FullNodeScan {
```
RTann, line 16: Mentioned on Slack, but I wonder if something like `NodeScanV2` would also work? Makes it clear this is the newer version of node scanning

proto/internalapi/sensor/compliance_nodescan_iservice.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
```
RTann, line 1: Why not just update `internalapi/sensor/compliance_iservice.proto`? It already is implemented to control the communication between Compliance pods and Sensor. Node scanning can just be added to the `oneof` in `MsgFromCompliance`

Maddosaurus: Originally, I wanted to make sure we have completely separated processing pipelines for the two different NodeScan messages.
With a better idea of the work involved to set up a new bidi stream, your idea makes way more sense. Let me set that up!

proto/storage/node.proto
```diff
@@ -70,7 +73,7 @@ message Node {
 
 // Next tag: 5
 message NodeScan {
-    google.protobuf.Timestamp          scan_time        = 1 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
+    google.protobuf.Timestamp          san_time        = 1 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
```
msugakov, line 76: Let's put it back :-)

```suggestion
    google.protobuf.Timestamp          scan_time        = 1 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
```

proto/internalapi/sensor/compliance_nodescan.proto
msugakov: My logic is that "NodeScan" combines two things: 1) information about the node, 2) information about the scan. While information about the scan is more or less self-evident when you look at the message, information about the node isn't. Just `name` and `id` could mean anything: "name" and "id" of the scan or "name" and "id" of the node, or "name" of the node but "id" of the scan. Therefore I suggest to make things more explicit and clarify directly in the name.

The example of `Node` and `NodeScan` doesn't match here. `Node` rightfully has `id` and `name` because there hardly can be a confusion what these mean. `NodeScan` at the same time doesn't have `id` or `name` attributes, unlike the new `NodeScanV2`.

proto/storage/node.proto
Maddosaurus: Thanks, well spotted!

proto/storage/node.proto
```diff
@@ -82,6 +85,15 @@ message NodeScan {
     repeated Note                      notes            = 4;
 }
 
+// Names and data types are designed to follow NodeScan as close as possible
```
msugakov, line 88: I think this comment now is mostly redundant.

msugakov (approved):

PR #3301, opened 2022-10-04
Basic structure of collection form

dvail:
## Description

Adds the basic structure of the collection form page, with empty child components and page layout.

The most interesting (complicated?) piece of this PR is related to the layout of the collection selector results panel on the righthand side of the page. The ideal layout planned with UX is to have the following properties:

- The sidebar results fill the vertical space in the container, minus the content header and footer bars.
- The sidebar results does not overflow the available vertical space; its contents are always 100% visible.
- The sidebar results has its own scroll bar so users can browse the collection results while editing rules without having to scroll the page up and down.

This is done through a custom hook with an `IntersectionObserver` that does the following:
- As the user scrolls, calculate the amount of vertical space the obstructing elements take up in the main container
- Using the value above, set the `maxHeight` of the selector results to be the full viewport height, minus the masthead,  minus padding, and minus the obstructed height

The following screenshots demonstrate this behavior (Ipsum is for layout testing and not part of this PR):

At the top of the page, the selector results fill the right side column:
image

As the user scrolls down, the sidebar is sticky and grows to fill the now empty header space:
image

At the bottom of the page, the sidebar remains sticky to the top of the container, but gives back space for the bottom toolbar.
image

With unequal content between sidebar and main area, the sidebar will still fill the vertical space at a maximum:
image
image

## Checklist

## Testing Performed

Testing layout of PR as-is (no content, no ipsum).

Clicking the "Create collection" button on the main collection page renders:
image

Clicking the collection name in the main collection page table renders:
image

Selecting either "Edit" or "Clone" from the main collection page table renders:
image


dvail:
* master
  * **PR #3301** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3301?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionFormPage.tsx
```diff
@@ -1,16 +1,90 @@
...
+import BreadcrumbItemLink from 'Components/BreadcrumbItemLink';
+import { collectionsBasePath } from 'routePaths';
 import { CollectionPageAction } from './collections.utils';
+import CollectionForm from './CollectionForm';
 
 export type CollectionsFormPageProps = {
     pageAction: CollectionPageAction;
 };
 
+const noopRequest = {
```
dvail, line 23: To keep the data fetching declarative, we use this object as a response in place of a `fetch` call when the user is on the "create" version of the form page.

ui/apps/platform/src/Containers/Collections/CollectionFormPage.tsx
```diff
@@ -1,16 +1,90 @@
...
-    ) : (
+    const action = pageAction.type;
+    const collectionId = action !== 'create' ? pageAction.collectionId : undefined;
+    const collectionFetcher = useCallback(() => {
+        return collectionId ? getCollection(collectionId) : noopRequest;
+    }, [collectionId]);
+
+    const { data } = useRestQuery(collectionFetcher);
+
+    const watchElements = Array.from(document.getElementsByClassName(observedClass));
```
dvail, line 51: The `useLayoutSpaceObserver` hook and the `IntersectionObserver` inside it require a reference to DOM nodes in order to track the space usage in the container. There are three main approaches I looked at when setting this up:

1. Use the imperative `document.getElementsByClassName` method to access the DOM directly.
This is probably the least React-idiomatic, but most straightforward, way to do this and it what I have now. This simply queries the DOM for a class that is defined in this component to get a reference to the correct elements. The `undefined` values on the first render do not matter, since the `useEffect` in the hook triggers an update when the DOM elements are mounted.

2. Use `Refs` part one. Ideally I'd like to get a ref to the `<section>` in the PF `<PageSection>` components, but since those are React function components, the only way to do so is via `forwardRef`. Unfortunately, PF doesn't use `forwardRef` for this component so this option is out.

3. Use `Refs` part two. For this use case, I could wrap the sections I need reference to with a simple `<div ref={myRef}>` and the functionality should be identical. The downside here is that it would make the markup a bit clunky, as well as introduce CSS conflicts that would need to be overridden in order to display the PF components correctly. (We have a global CSS rule that messes up styles if the immediate descendant of a `<Page>` is not a `<PageSection>`.)

ui/apps/platform/src/hooks/useLayoutSpaceObserver.ts
```diff
@@ -0,0 +1,47 @@
...
+ *      The parent element container to observe. An `undefined` value will use `document.body`.
+ * @param observationTargets
+ *      An array of descendent elements of `root` whose space should be monitored.
+ * @param granularity
+ *      How frequently space updates should be reported. Higher numbers will result in more frequent
+ *      and more accurate updates but with a larger volume of function calls. Must be an integer >= 1.
+ * @return
+ *      The total `height` and `width` the target elements take up in the visible portion of the root element.
+ */
+export default function useLayoutSpaceObserver(
```
dvail, line 17: This is used to track the visible space a list of elements occupy in a parent container. Although this was needed for the specific case of the Selector Results component, it was general enough that I moved it to the main `hooks/` directory. If we don't feel like this is something we are going to want to reuse I can co-locate it with the collections code.

PR #3305, opened 2022-10-04
ROX-12612,ROX-12613,ROX-12618: Collections Add and Delete workflows

keyallis:
## Description

Add workflow code for creation and deletion of Collections objects.

## Checklist

## Testing Performed

Added unit testing


central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -3,18 +3,82 @@ package datastore
...
+	// first walk adds vertices
+	err := ds.storage.Walk(ctx, func(obj *storage.ResourceCollection) error {
+		return ds.graph.AddVertexByID(obj.GetId(), obj.GetId())
+	})
+	if err != nil {
+		return err
+	}
+
+	// second walk adds edges
+	return ds.storage.Walk(ctx, func(obj *storage.ResourceCollection) error {
```
theencee, line 43: This feels inefficient because it's two walks through the data store. A simple solution would be to fetch it all and add it to an array (or use --get-all-func in gen.go to generate a GetAll func) and then traverse that to build the DAG. But there is a concern (however small of running out of memory).

So for the moment, do you mind creating a ticket and TODO to revisit optimizing this?

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -3,18 +3,82 @@ package datastore
...
+	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetId())
+	if err != nil {
+		return err
+	}
+	for _, edge := range obj.GetEmbeddedCollections() {
+		err = ds.graph.AddEdge(obj.GetId(), edge.GetId())
+		if err != nil {
+			deleteErr := ds.graph.DeleteVertex(obj.GetId())
+			if deleteErr != nil {
+				log.Errorf("Failed to delete collection, might result in bad state (%v)", deleteErr)
```
theencee, line 68: This is more of a warning, no? It's not actually going to prevent anything and you aren't deleting the collection from the datastore (so the message is incorrect as well). All it does is prevent finding cycles. A restart of the pod will fix it too.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -59,3 +123,37 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	if err != nil {
+		return err
+	}
+
+	err = ds.storage.Upsert(ctx, collection)
+	if err != nil {
+		// if we fail to upsert, update the graph
+		deleteErr := ds.deleteCollectionFromGraph(ctx, collection.GetId())
+		if deleteErr != nil {
+			log.Errorf("Failed to delete collection, might result in bad state (%v)", deleteErr)
```
theencee, line 140: This error isn't about deleting collection, it's about the local state being inconsistent. The collection itself doesn't exist in the DB so a restart would fix it.

Depending on what the id is (guid vs auto-increment id), then it might also not be an issue. Because if the id is a guid the chances are low that the next upsert will have the exact same collection id.

central/resourcecollection/datastore/datastore_impl.go
keyallis: If the vertex doesn't get deleted any following collection with the same id would not be able to be added. I think this line is more of a precaution, "proper error handling" and whatnot. I think logically the line would never be run, but if it did it would be really nice to know. According to the docs the function returns error if the id is empty (something that would cause the `AddVertexById` function call to return error earlier in the function), or unknown (something that should never happen because we know we just added the id to the graph)

central/resourcecollection/datastore/datastore_impl.go
keyallis: This is kind of the same issue as above, theoretically this issue should never occur since we know we just added the id to the graph, so removing it from the graph subsequently should never fail given the constraints of the function

keyallis:
@theencee an open question for ya, is this something that needs to utilize locks to prevent unexpected behavior between checks on the local graph and the datastore?

central/resourcecollection/datastore/datastore_impl.go
theencee: Fair enough to keep it as an error, but I think the error message is misleading to the customer who might be confused about why it mentions deleting a collection when they are trying to add.

Perhaps reworded to "Failed to remove collection <id> from internal state object" or something like that

theencee:
Yes good call on locks. For add collection to graph, you should have a RW lock on the DAG itself. Make the entire operation to add edges and vertex atomic.

central/resourcecollection/datastore/datastore_impl.go
keyallis: This has been addressed in recent commits

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -3,18 +3,120 @@ package datastore
...
-	storage  postgres.Store
+	storage  store.Store
 	indexer  index.Indexer
 	searcher search.Searcher
+
+	graphLock sync.Mutex
+	graph     *dag.DAG
+}
+
+func resetLocalGraph(ds *datastoreImpl) {
```
theencee, line 31: If this is only used for tests, you can move it to the test file since it's in the same package. Reduces risk of it accidentally getting used in the main file

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -3,18 +3,120 @@ package datastore
...
+		return nil
+	}
+
+	ds.graphLock.Lock()
+	defer ds.graphLock.Unlock()
+
+	ds.graph = dag.NewDAG()
+
+	// add ids first
+	ids, err := ds.storage.GetIDs(ctx)
```
theencee, line 52: If I understand the postgres code correctly, this will return all the ids _that can be seen by ctx_. In other words if you are logged in with a scope that can only see some collections, this won't get all collections. 

I recommend using a special context that has all access just for the init function.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -3,18 +3,120 @@ package datastore
...
+	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetId())
+	if err != nil {
+		return err
+	}
+	for _, edge := range obj.GetEmbeddedCollections() {
+		err = ds.graph.AddEdge(obj.GetId(), edge.GetId())
+		if err != nil {
+			deleteErr := ds.graph.DeleteVertex(obj.GetId())
+			if deleteErr != nil {
+				log.Errorf("Failed to delete collection, might result in bad state (%v)", deleteErr)
```
theencee, line 103: [nit] "Failed to remove collection from internal state object" (to keep consistent with other messages)?

theencee (approved):
lgtm but please consider changing the context

PR #3312, opened 2022-10-05
[ROX-12384] : Scope namespace query for evaluating service account scope permissions to enclosing cluster; Add namespace loader to cache previously queried results

charmik-redhat:
## Description

- When `serviceAccount->scopedPermissions` is queried from a cluster, the evaluation for scopedPermissions only needs the namespaces within the cluster containing the serviceAccount. So, this PR scopes the query to fetch namespaces to the enclosing cluster to avoid querying namespaces from other clusters.
- Added a namespace loader to cache already resolved Namespaces to avoid fetching all namespaces from the database for each service account queried in a single graphQL request.

## Checklist

## Testing Performed

- Added unit tests for namespace loader.
- Manually tested by deploying a multi-cluster ACS instance using the 10-sensor script from scale tests. Verified performance improvement of the `getCluster_SERVICE_ACCOUNT` graphQL query from 'Configuration management' UI.

49507739-2429-4239-879B-4BEA333549DC

5BE036D1-E433-4362-9340-DADB549BBE05

5C6C4233-10F9-4929-9D13-2C926EB53784_4_5005_c

### DIFF
--- a/central/graphql/resolvers/loaders/namespaces.go
+++ b/central/graphql/resolvers/loaders/namespaces.go
@@ -9,7 +9,6 @@ import (
 	"github.com/stackrox/rox/central/namespace/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
-	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/search"
 	"github.com/stackrox/rox/pkg/sync"
 )
@@ -17,11 +16,9 @@ import (
 var namespaceLoaderType = reflect.TypeOf(storage.NamespaceMetadata{})
 
 func init() {
-	if env.PostgresDatastoreEnabled.BooleanSetting() {
-		RegisterTypeFactory(reflect.TypeOf(storage.NamespaceMetadata{}), func() interface{} {
-			return NewNamespaceLoader(datastore.Singleton())
-		})
-	}
+	RegisterTypeFactory(reflect.TypeOf(storage.NamespaceMetadata{}), func() interface{} {
+		return NewNamespaceLoader(datastore.Singleton())
+	})
 }
 
 // NewNamespaceLoader creates a new loader for NamespaceMetaData.
@@ -105,7 +102,7 @@ func (nsldr *namespaceLoaderImpl) load(ctx context.Context, ids []string) ([]*st
 	namespaces, missing := nsldr.readAll(ids)
 	if len(missing) > 0 {
 		var err error
-		namespaces, err = nsldr.ds.GetBatch(ctx, collectMissing(ids, missing))
+		namespaces, err = nsldr.ds.GetManyNamespaces(ctx, collectMissing(ids, missing))
 		if err != nil {
 			return nil, err
 		}
--- a/central/graphql/resolvers/loaders/namespaces.go
+++ b/central/graphql/resolvers/loaders/namespaces.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stackrox/rox/central/namespace/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/search"
 	"github.com/stackrox/rox/pkg/sync"
 )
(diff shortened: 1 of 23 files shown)


central/graphql/resolvers/loaders/namespaces.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var namespaceLoaderType = reflect.TypeOf(storage.NamespaceMetadata{})
+
+func init() {
+	RegisterTypeFactory(reflect.TypeOf(storage.NamespaceMetadata{}), func() interface{} {
+		return NewNamespaceLoader(datastore.Singleton())
+	})
```
keyallis, line 21: May be worth considering changing this to behind the postgres flag, as well as it's usage

central/namespace/resolver.go
```diff
@@ -17,16 +18,28 @@ import (
...
+	if loaders.HasLoaderContext(ctx) {
+		loader, err := loaders.GetNamespaceLoader(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return loader.FromQuery(ctx, query)
+	}
+	return dataStore.SearchNamespaces(ctx, query)
+}
+
```
keyallis, line 38: I think it might be better to utilize the loader in the namespace graphql endpoint, and have that be the endpoint invoked by the service account graphql code

keyallis (changes requested):
Move loader usage to namespace graphql endpoint, put loader behind feature flag, call namespace graphql from service accounts graphql

central/graphql/resolvers/loaders/namespaces.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var namespaceLoaderType = reflect.TypeOf(storage.NamespaceMetadata{})
+
+func init() {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
```
md2119, line 20: Why is the loader behind postgres feature flag. The namespace datastore has always existed. You do not need to guard the loader registration behind the feature flag.

central/namespace/resolver.go
```diff
@@ -26,6 +26,7 @@ func ResolveAll(ctx context.Context, dataStore datastore.DataStore, deploymentDa
 
 // ResolveMetadataOnlyByQuery resolves all namespaces based on a query. This will _not_ populate volatile runtime data and that must be requested separately.
 func ResolveMetadataOnlyByQuery(ctx context.Context, q *v1.Query, dataStore datastore.DataStore) ([]*v1.Namespace, error) {
+
```
md2119, line 29: remove

central/namespace/datastore/datastore.go
```diff
@@ -189,6 +191,14 @@ func (b *datastoreImpl) GetNamespaces(ctx context.Context) ([]*storage.Namespace
 	return allowedNamespaces, nil
 }
 
+func (b *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage.NamespaceMetadata, error) {
+	namespaces, _, err := b.store.GetMany(ctx, ids)
```
md2119, line 195: Check for permission first.
```
var namespaces []*storage.NamespaceMetadata
var err error
if ok, err := namespaceSAC.ReadAllowed(ctx); err != nil {
		return nil, err
} else if !ok {
		return ds.SearchNamespaces(ctx, {namespace_query})
}
namespaces, _, err = b.store.GetMany(ctx, ids)
...
```

md2119 (changes requested):

central/namespace/datastore/datastore.go
md2119: Further explained: The above block checks for global access to `namespace` SAC resource. If not, delegate the filtering based on permissions to the searcher.

central/graphql/resolvers/namespaces.go
```diff
@@ -135,8 +135,25 @@ func (resolver *Resolver) Namespaces(ctx context.Context, args PaginatedQuery) (
...
+			return nil, err
+		}
+		namespaces = make([]*v1.Namespace, 0, len(metadataSlice))
+		for _, metadata := range metadataSlice {
+			namespaces = append(namespaces, &v1.Namespace{Metadata: metadata})
+		}
+		return resolver.wrapNamespacesWithContext(ctx, namespaces, nil)
+	}
+
+	namespaces, err = namespace.ResolveMetadataOnlyByQuery(ctx, query, resolver.NamespaceDataStore)
```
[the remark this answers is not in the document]
charmik-redhat: Had added loader usage to `namespace.ResolveMetadataOnlyByQuery` but moved it out because other functions in the file `central/namespaces/resolvers.go` are used by namespace service, which does not have initialize loaders. 

I would instead remove `ResolveMetadataOnlyByQuery` from  and just add its code to `Namespaces` graphQL resolver which is the only reference to `ResolveMetadataOnlyByQuery`.

md2119 (approved):
🚢

PR #3313, opened 2022-10-05
[ROX-12955] : Add CVE field query instead of VULNERABILITY scoping in vuln request manager when Postgres is enabled

charmik-redhat:
## Description

Scoping searches by searchCategory_VULNERABILITIES is not supported on Postgres. Also, the CVEs in vulnerability requests cannot be used as CVE IDs because Postgres datastores do not directly use CVE string as IDs. So, on postgres, vuln request manager should add a CVE conjunction query instead of scoping.

## Checklist

## Testing Performed

Manually tested by creating a deferral request until fixable, and see that deferral is handled properly post approval and does not cause any crashes on central restart. Perform the test for both postgres and rocksDB.

### DIFF
--- a/central/vulnerabilityrequest/manager/requestmgr/manager_impl.go
+++ b/central/vulnerabilityrequest/manager/requestmgr/manager_impl.go
@@ -409,14 +409,12 @@ func (m *managerImpl) getFixableDeferrals() ([]*storage.VulnerabilityRequest, er
 			// TODO: Determine if it's worth checking cvePkg.ContainsComponentBasedCVE(cve.GetTypes()) before doing this. It would involve going to the data store to read CVE data
 			// This is only necessary if somehow there ended up being a deferral on a cluster CVE. Or if it goes from an image cve to node cve
 			// TODO: Test what happens if it's a cluster cve
-			var cveScopedCtx context.Context
+			var ctx context.Context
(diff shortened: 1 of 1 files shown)


theencee (approved):
Code looks good, but please test manually with and without postgres.

Repro steps:

* View an image
* Select an CVE, Defer it. Choose "Until fixable"
* Go to risk acceptance, and accept the deferral
* Restart central and make sure it comes up

central/vulnerabilityrequest/manager/requestmgr/manager_impl.go
```diff
@@ -403,19 +404,26 @@ func (m *managerImpl) getFixableDeferrals() ([]*storage.VulnerabilityRequest, er
...
+				qb.AddExactMatches(search.CVE, cve)
+			} else {
+				cveScopedCtx = scoped.Context(allAccessCtx, scoped.Scope{
+					ID:    cve,
+					Level: v1.SearchCategory_VULNERABILITIES,
+				})
+			}
 
-			fixableQuery, err := utils.GetAffectedImagesQuery(res, search.NewQueryBuilder().AddBools(search.Fixable, true).ProtoQuery())
+			fixableQuery, err := utils.GetAffectedImagesQuery(res, qb.ProtoQuery())
```
theencee, line 423: You still need the `AddBools(search.Fixable, true)` part

PR #3314, opened 2022-10-05
Dashrews/ROX-12094 cleanup postgres after rollback to rocks

dashrews78:
## Description

In the event we upgrade from Rocks to Postgres and then downgrade back to Rocks, we could leave ourselves in a situation where we have lingering stale Postgres data on the subsequent upgrade back to Postgres.  So to alleviate that, if we have a situation where we determine we need to migrate from Rocks to Postgres, either initially or on an upgrade, we will clear out the Postgres data and start Postgres fresh.

## Checklist

## Testing Performed

Updated the unit tests as well as the clone test.  I also noticed that the clone test was not executed in Postgres tests which was an oversight from before.  So I updated the Makefile to correct that oversight.

### DIFF
--- a/migrator/clone/postgres/db_clone_manager_impl.go
+++ b/migrator/clone/postgres/db_clone_manager_impl.go
@@ -163,20 +163,24 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate(rocksVersion *migrations.Migratio
 		// so just fall through to the rest of the processing.
 		if currClone.GetMigVersion().LastPersisted.Before(rocksVersion.LastPersisted) {
 			// We want to start fresh as we are migrating from Rocks->Postgres.  So any data that exists in
-			// Postgres from a previous upgrade followed by a rollback needs to be ignored.
+			// Postgres from a previous upgrade followed by a rollback needs to be ignored.  So just drop current
+			// and let it create anew.
 			err := pgadmin.DropDB(d.sourceMap, d.adminConfig, CurrentClone)
-
-			// Create a temp clone for processing of current
-			// Seed it from an empty database because we need to run migrations from Rocks to Postgres.
-			err = pgadmin.CreateDB(d.sourceMap, d.adminConfig, pgadmin.EmptyDB, TempClone)
-			// If for some reason, we cannot create a temp clone we will need to continue to upgrade
-			// with the current.
 			if err != nil {
-				log.Errorf("Unable to create temp clone: %v", err)
-				return CurrentClone, false, nil
+				log.Errorf("Unable to drop current clone: %v", err)
+
+				// The drop of the current clone is unlikely to fail, in the event it does; we can try to
+				// process with temp.
+				err = pgadmin.DropDB(d.sourceMap, d.adminConfig, TempClone)
+				if err != nil {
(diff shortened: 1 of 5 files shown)


migrator/clone/postgres/db_clone_manager_impl.go
```diff
@@ -162,24 +162,26 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate(rocksVersion *migrations.Migratio
...
+			// We want to start fresh as we are migrating from Rocks->Postgres.  So any data that exists in
+			// Postgres from a previous upgrade followed by a rollback needs to be ignored.  So just drop current
+			// and let it create anew.
+			err := pgadmin.DropDB(d.sourceMap, d.adminConfig, CurrentClone)
+			if err != nil {
+				log.Errorf("Unable to drop current clone: %v", err)
 
-				// If for some reason, we cannot create a temp clone we will need to continue to upgrade
-				// with the current and thus no fallback.
+				// The drop of the current clone is unlikely to fail, in the event it does; we can try to
```
c-du, line 172: What could be potential reason for not being able to Drop database? I cannot think of the case that could be corrected by using temp.

migrator/clone/postgres/db_clone_manager_impl.go
```diff
@@ -162,24 +162,26 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate(rocksVersion *migrations.Migratio
...
+				// process with temp.
+				err = pgadmin.DropDB(d.sourceMap, d.adminConfig, TempClone)
```
c-du, line 174: Does temp exists? They should be been removed during Scan?

migrator/clone/postgres/db_clone_manager_impl.go
dashrews78: yeah that is probably over programming on my part.  the only ways that could fail generally are because we are in a very bad state.  It could fail if we are trying to drop the database from the same connection we are using to drop it.  But we aren't because we are using the admin config to do this.  I will clean some of this stuff up.

migrator/clone/postgres/db_clone_manager_impl.go
```diff
@@ -162,24 +162,16 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate(rocksVersion *migrations.Migratio
 		// Rocks more recently updated than Postgres so need to migrate from there.  Otherwise, Postgres is more recent
 		// so just fall through to the rest of the processing.
 		if currClone.GetMigVersion().LastPersisted.Before(rocksVersion.LastPersisted) {
-			// Rollback is not enabled, so process the Rocks -> Postgres migration on CurrentClone.
-			if !d.rollbackEnabled() {
-				return CurrentClone, true, nil
+			// We want to start fresh as we are migrating from Rocks->Postgres.  So any data that exists in
+			// Postgres from a previous upgrade followed by a rollback needs to be ignored.  So just drop current
+			// and let it create anew.
+			err := pgadmin.DropDB(d.sourceMap, d.adminConfig, CurrentClone)
```
c-du, line 168: Add an info log for this action?

c-du (approved):

PR #3315, opened 2022-10-05
Pull k8s events as yaml file also in case of test failure

msugakov:
## Description

The problem with `kubectl get events -o wide` is that the `LAST SEEN` and the `FIRST SEEN` columns have _relative_ timestamps like `4m11s`. This is very inconvenient in investigating CI failures because one needs to know the instant when `kubectl get events` was ran and do the math.

I suggest that we dump events in yaml format which preserves the original info and absolute timestamps like `2022-10-05T10:31:32Z`.

This relates to https://issues.redhat.com/browse/ROX-12963

## Checklist

None of these needed:

## Testing Performed

* None. Should just work.

### DIFF
--- a/scripts/ci/collect-service-logs.sh
+++ b/scripts/ci/collect-service-logs.sh
@@ -80,7 +80,9 @@ main() {
         done
     done
 
+    kubectl -n "${namespace}" get events -o wide >"${log_dir}/events.txt"
     kubectl -n "${namespace}" get events -o yaml >"${log_dir}/events.yaml"
+
     find "${log_dir}" -type f -size 0 -delete
 }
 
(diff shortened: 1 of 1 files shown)


msugakov:
Fun that it did not flake a single test when I actually wanted it to.

scripts/ci/collect-service-logs.sh
```diff
@@ -80,7 +80,7 @@ main() {
         done
     done
 
-    kubectl -n "${namespace}" get events -o wide >"${log_dir}/events.txt"
```
gavin-stackrox, line 83: whatabout keeping the `-o wide` version as well? (in case there is someone that is attached to that format)

gavin-stackrox (approved):

scripts/ci/collect-service-logs.sh
msugakov: Good idea, done.

PR #3319, opened 2022-10-05
Dashrews/ROX-12926 Fix query building parse issue that manifested in a broken graphql postgres tests

dashrews78:
## Description

Turns out this was not a flake.  the pods query was of the form:
"Deployment ID:  d4b04bae-9d9d-46a2-8c43-b615e07d9969"

Notice the space after the colon.  When the parser was processing the parameters and getting the value it did not strip the space from the value portion.  So when the search was done, it was trying to match the deploymentid column to:
" d4b04bae-9d9d-46a2-8c43-b615e07d9969" 
That would not find any rows because it didn't match the string exactly.  So the simple fix was to trim the space of the value portion when it is parsed out.  Feedback from those of you more familiar with the graphql processing would be greatly appreciated in the event there is a more optimal place to strip out this space.

## Checklist

## Testing Performed

Existing tests are sufficient.  The whole purpose of this ticket was because one was failing due to this issue.

### DIFF
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -86,7 +86,7 @@ func parsePair(pair string, allowEmpty bool) (key string, values string, valid b
 			spl[1] = spl[1] + WildcardString
 		}
 	}
-	return spl[0], spl[1], true
+	return spl[0], strings.TrimSpace(spl[1]), true
 }
 
 func queryFromFieldValues(field string, values []string, highlight bool) *v1.Query {
@@ -163,8 +163,9 @@ func docIDQuery(ids []string) *v1.Query {
 	})
 }
 
-//go:generate stringer -type=QueryModifier
 // QueryModifier describes the query modifiers for a specific individual query
+//
+//go:generate stringer -type=QueryModifier
 type QueryModifier int
 
 // These are the currently supported modifiers
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -164,7 +164,6 @@ func docIDQuery(ids []string) *v1.Query {
 }
 
 // QueryModifier describes the query modifiers for a specific individual query
-//
 //go:generate stringer -type=QueryModifier
 type QueryModifier int
 
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -86,7 +86,7 @@ func parsePair(pair string, allowEmpty bool) (key string, values string, valid b
 			spl[1] = spl[1] + WildcardString
 		}
 	}
-	return spl[0], strings.TrimSpace(spl[1]), true
+	return strings.TrimSpace(spl[0]), strings.TrimSpace(spl[1]), true
 }
 
 func queryFromFieldValues(field string, values []string, highlight bool) *v1.Query {
@@ -164,6 +164,7 @@ func docIDQuery(ids []string) *v1.Query {
 }
 
 // QueryModifier describes the query modifiers for a specific individual query
+//
 //go:generate stringer -type=QueryModifier
 type QueryModifier int
 
--- a/pkg/search/parser_test.go
+++ b/pkg/search/parser_test.go
@@ -51,3 +51,70 @@ func TestSplitQuery(t *testing.T) {
 		})
 	}
 }
+
+func TestParsePairs(t *testing.T) {
+	cases := []struct {
+		query      string
+		allowEmpty bool
+		pairs      []string
+		valid      bool
+	}{
+		{
+			query:      "Deployment Label:label",
+			allowEmpty: false,
+			pairs:      []string{"Deployment Label", "label"},
+			valid:      true,
+		},
+		{
+			query:      "Deployment Label:label,label",
+			allowEmpty: false,
+			pairs:      []string{"Deployment Label", "label,label"},
+			valid:      true,
+		},
+		{
+			query:      "  Deployment Label :  label+label  ",
+			allowEmpty: false,
+			pairs:      []string{"Deployment Label", "label+label"},
+			valid:      true,
+		},
+		{
+			query:      "Deployment Label:label+label",
+			allowEmpty: false,
+			pairs:      []string{"Deployment Label", "label+label"},
+			valid:      true,
+		},
+		{
+			query:      "Deployment Label",
+			allowEmpty: false,
+			pairs:      []string{"", ""},
+			valid:      false,
+		},
+		{
+			query:      "Deployment Label:",
+			allowEmpty: false,
+			pairs:      []string{"", ""},
+			valid:      false,
+		},
+		{
+			query:      "Deployment Label:",
+			allowEmpty: true,
+			pairs:      []string{"Deployment Label", WildcardString},
+			valid:      true,
+		},
+		{
+			query:      "Deployment Label:attempted-alerts-dep-6+Policy:Kubernetes Actions: Exec into Pod",
+			allowEmpty: false,
+			pairs:      []string{"Deployment Label", "attempted-alerts-dep-6+Policy:Kubernetes Actions: Exec into Pod"},
+			valid:      true,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.query, func(t *testing.T) {
+			key, value, valid := parsePair(c.query, c.allowEmpty)
+			assert.Equal(t, valid, c.valid)
+			assert.Equal(t, key, c.pairs[0])
+			assert.Equal(t, value, c.pairs[1])
+		})
+	}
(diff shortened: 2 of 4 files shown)


qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
```diff
@@ -89,7 +89,7 @@ class DeploymentEventGraphQLTest extends BaseSpecification {
 
     private final gqlService = new GraphQLService()
 
-    @Category(GraphQL)
+    @Category(BAT)
```
dashrews78, line 92: Changed this to make sure the tests were executed for the PR.  Will change that back to GraphQL after review.

janisz: I think this test should be BAT https://github.com/stackrox/stackrox/pull/3340

pkg/search/parser.go
```diff
@@ -86,7 +86,7 @@ func parsePair(pair string, allowEmpty bool) (key string, values string, valid b
 			spl[1] = spl[1] + WildcardString
 		}
 	}
-	return spl[0], spl[1], true
+	return spl[0], strings.TrimSpace(spl[1]), true
```
janisz, line 89: Could you add unit test for this?

qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
dashrews78: Fine with me.

misberner:
The change looks good to me in principle, but I'm a bit lost on the history of this change with the few things I could find out from this PR + from the associated JIRA:
- "Turns out it was not a flake" suggests that it was suspected to be a flake. That suggests the test used to pass (or was it suspected that a newly-added test that always failed was a flake?). What changed to make it break?
- The JIRA links to a postgresql test failure. Is this in any way related to postgresql, or does it also fail on non-postgres tests? (I'll look that up after posting this comment)
- The PR description mentions "familiarity with the graphql processing". I don't see anything specific to graphql in the fix. Does this not fail on non-GraphQL queries? If so, why?

misberner:
Please also make sure to [set a proper title, and to not just leave the branch name](https://github.com/stackrox/stackrox/blob/master/.github/contributing.md#creating-a-pr).

dashrews78:
> The change looks good to me in principle, but I'm a bit lost on the history of this change with the few things I could find out from this PR + from the associated JIRA:
> 
> * "Turns out it was not a flake" suggests that it was suspected to be a flake. That suggests the test used to pass (or was it suspected that a newly-added test that always failed was a flake?). What changed to make it break?
> * The JIRA links to a postgresql test failure. Is this in any way related to postgresql, or does it also fail on non-postgres tests? (I'll look that up after posting this comment)
> * The PR description mentions "familiarity with the graphql processing". I don't see anything specific to graphql in the fix. Does this not fail on non-GraphQL queries? If so, why?

@misberner The initial thought was that it was a flake because it was only happening in the nightlies and only for Postgres.  This goes through the query building which is slightly different than it was in Rocks.  That is why it failed in Postgres only.  It only failed in the nightlies because it was only executed in the nightlies because it was not marked as BAT, but I believe the original thoughts were that it was executed elsewhere.  That thought was incorrect.

Yes this one only failed when running in Postgres.

The last one is my understanding.  It was a GraphQL test and going through the query building code to get at the data.  I believe it to be GraphQL related because I think that is the only path that builds SQL queries like this.  I could be mistaken.  This could occur anywhere we are building a query that took a JSON string like:

{deploymentid: test_deployment} and eventually turned it into "where deploymentid = 'test_deployment'. So perhaps focusing on GraphQL was a bad choice of terms on my part as it is probably more accurately the Query Builder.

misberner:
I see that the code in question was last modified in 2019 and that the test in question was last modified Sep 20 with a change that looks highly unrelated. If this is something postgres-specific, this PR _might_ apply the fix at the right location from the standpoint of where it _should_ be, but not really at the "right" location from the standpoint of flattening out an inconsistency between postgresql and bleve. Or am I reading this wrong?

misberner:
My point is: if Bleve _implicitly_ strips spaces from values, and postgres _doesn't_, then this change might fix the test, but it provides no guarantee of aligning the behaviors of postgres and bleve on untrimmed values on code paths that for one reason or the other (manual query construction or something?) don't go through `parsePair`

dashrews78:
> 

@misberner And that in essence is the guidance I was asking for in the description.  It certainly makes sense to do what I did where I did it.  And I believe that should be there.  I certainly have little understanding of what bleve was doing and how we used it.  My assumption is that it was a json indexer and would have ignored that space on a match as a lot of those do.  Postgres is not going to ignore that space.

misberner:
You might know more about JSON indexers than I do. I don't see an obvious reason why using JSON representation for indexing would have to do anything with respecting or ignoring spaces in field _values_ - they are just opaque strings, unless you do word-based indexing, which I'm not sure we ever (consciously did), but which again is motivated by nothing specific to JSON.

I think the question is also the other way round: would `Deployment:foobar` match a deployment with name `" foobar"` (leading space) in Bleve? Sounds like it certainly wouldn't in postgres (disregarding the fact that deployments in k8s can't begin with or even contain spaces, but I'm sure some of our indexed fields can).

My point is, it might be safer from a parity standpoint to instead change the `v1.Query`-to-SQL-query translator for postgres to just strip leading (and maybe trailing) spaces from values. And then either verify that there remains a way to search for values with leading spaces, or ensure that this didn't exist in the Bleve world either and thus was never needed by anyone.

I do agree that where you inserted the stripping is the right place from a general standpoint, but with a feature as complex as postgres, targeting parity is paramount; fixing long-standing issues should come only after the confidence about having achieved parity/being able to tolerate divergence is sufficiently high.

dashrews78:
> My point is: if Bleve _implicitly_ strips spaces from values, and postgres _doesn't_, then this change might fix the test, but it provides no guarantee of aligning the behaviors of postgres and bleve on untrimmed values on code paths that for one reason or the other (manual query construction or something?) don't go through `parsePair`

right that is the type of insight I was looking for.  I will look around some more and see where else this may be an issue

dashrews78:
> You might know more about JSON indexers than I do. I don't see an obvious reason why using JSON representation for indexing would have to do anything with respecting or ignoring spaces in field _values_ - they are just opaque strings, unless you do word-based indexing, which I'm not sure we ever (consciously did), but which again is motivated by nothing specific to JSON.
> 
> I think the question is also the other way round: would `Deployment:foobar` match a deployment with name `" foobar"` (leading space) in Bleve? Sounds like it certainly wouldn't in postgres (disregarding the fact that deployments in k8s can't begin with or even contain spaces, but I'm sure some of our indexed fields can).
> 
> My point is, it might be safer from a parity standpoint to instead change the `v1.Query`-to-SQL-query translator for postgres to just strip leading (and maybe trailing) spaces from values. And then either verify that there remains a way to search for values with leading spaces, or ensure that this didn't exist in the Bleve world either and thus was never needed by anyone.
> 
> I do agree that where you inserted the stripping is the right place from a general standpoint, but with a feature as complex as postgres, targeting parity is paramount; fixing long-standing issues should come only after the confidence about having achieved parity/being able to tolerate divergence is sufficiently high.

So this particular query worked without spaces and with any number of spaces when run against the non-postgres version of stackrox.  i.e. deploymentid:test_deployment vs deploymentid:   test_deployment.

That comparison is what sent me down the path I went.

My experience with document based indexers is they index each word.  At least that is how elastic worked when I used it.  What Bleve does is outside my current scope of knowledge.

janisz (approved):

pkg/search/parser_test.go
```diff
@@ -51,3 +51,70 @@ func TestSplitQuery(t *testing.T) {
...
+			allowEmpty: false,
+			pairs:      []string{"", ""},
+			valid:      false,
+		},
+		{
+			query:      "Deployment Label:",
+			allowEmpty: true,
+			pairs:      []string{"Deployment Label", WildcardString},
+			valid:      true,
+		},
```
rhybrillou, line 103: I'd also be interested in the outcome of `query` "Deployments Label" with `allowEmpty` set to true.

rhybrillou (approved):

pkg/search/parser_test.go
c-du: Thanks for updating the test!

c-du (approved):

misberner (approved):

PR #3255, opened 2022-09-29
Bump K8s deps to 0.24.6

misberner:
## Description

This depends on https://github.com/stackrox/helm-operator/pull/31 and https://github.com/stackrox/helmtest , and both must be merged and the respective deps in `go.mod` be updated in order for this PR to be merged.

The `ClusterName` property in K8s object metadata was never used, so removing it doesn't hurt.

## Checklist
- [ ] Investigated and inspected CI test results
- [x] ~Unit test and regression tests added~
- [x] ~Evaluated and added CHANGELOG entry if required~
- [x] ~Determined and documented upgrade steps~
- [x] ~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

If any of these don't apply, please comment below.

## Testing Performed

- CI

PR #3320, opened 2022-10-05
[concurrency] Fix error with AsContext() result spuriously returning non-nil Err()

misberner:
## Description

The `ErrorWaitable` interface is a generalization of `context.Context`, which loosens the restrictions as to what `Err()` may return. Because it is legitimate for an `ErrorSignal` to return `Err() == nil` even when done (signaling the success of an asynchronous operation that could also have errored), the `contextWrapper` that wraps an `ErrorWaitable` to conform to the `Context` interface needs to address this fact by also mapping a `nil` `Err()` of a done signal to `context.Canceled`. However, the current implementation did this eagerly, not bothering to check whether the underlying `ErrorWaitable` was actually done.

The existing code worked fine with code that only checked `ctx.Err()` after the context indicated it was done, e.g., as in the following:
```
for {
  select {
  case <-ctx.Done():
    return ctx.Err()
  case <-someOtherChan:
    makeProgress()
  }
}
```
However, when adding a fast-exit condition:
```
for ctx.Err() == nil {
  ...
}
return ctx.Err()
```
the loop would never be entered and the function would immediately return with an error, in spite of the waitable wrapped via `AsContext()` not actually being done.

I caught this in the [attempt to upgrade k8s deps](https://github.com/stackrox/stackrox/pull/3255), observing that the `admission-control` config map would no longer get created, which was accompanied by a `context canceled` error in the sensor logs. Apparently, the authors of k8s `client-go` library added some early-exit checks to prevent making an attempt to the API server when the context was (or seemed) already cancelled at the time of invocation.

## Checklist

## Testing Performed

- Unit test

### DIFF
--- a/pkg/concurrency/context.go
+++ b/pkg/concurrency/context.go
@@ -28,9 +28,25 @@ func CancelContextOnSignal(ctx context.Context, cancel context.CancelFunc, signa
 
 type contextWrapper struct {
 	ErrorWaitable
+	isDone func() bool
+}
+
+func (w *contextWrapper) IsDone() bool {
+	if w.isDone != nil {
+		return w.isDone()
+	}
+	select {
+	case <-w.Done():
+		return true
+	default:
+		return false
+	}
 }
 
 func (w *contextWrapper) Err() error {
+	if !w.IsDone() {
+		return nil
+	}
 	err := w.ErrorWaitable.Err()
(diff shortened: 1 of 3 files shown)


misberner:
Operator test failures are flakes

dhaus67 (approved):

PR #3293, opened 2022-10-04
ROX-12934: WIP - Fix Operator reconciliation when using secrets and external DB

connorgorman:
## Description

Yeah, this definitely didn't work before and need to rework the reconciliation of the PVC and of the Central DB password

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Will add more tests

PR #3322, opened 2022-10-05
ROX-12934: Fix operator reconciliation when external DB is in use

misberner:
## Description

This PR fixes the reconciliation of the central-db persistent volume and password.

Unfortunately, the JIRA is missing a concrete description of the issue (and also doesn't cover all the existing issues). However, this PR functionally addresses the following:
- a generated password secret is deleted when central DB is disabled
- when an external DB is used, the use of a manually specified secret is enforced
- when an external DB is used, and the specified secret is not called `central-db-secret`, the contents of the user-supplied secret are mirrored into a `central-db-secret` secret (as only that is mounted into Central)
- when the `password` entry in the manually specified secret changed, the `central-db-secret` password is also updated upon reconciliation
- when there exists an unmanaged `central-db-secret`, and a password is manually specified in a different secret, reconciliation will fail if the passwords are different
- if the central DB deployment is not used, no PVC will be created
- if the central DB deployment is turned off, an existing, managed PVC will be disowned
- better error checking to prevent inconsistent settings, such as  using central-db persistence setting with an external DB

On the non-functional sides, this PR improves some naming and adds extensive unit tests for central DB password reconciler, as well as kuttl tests.

This PR is based upon https://github.com/stackrox/stackrox/pull/3293 by @connorgorman .

## Checklist

## Testing Performed

- Manual testing
- CI


operator/tests/central/basic/75-assert.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: central-db-secret
```
misberner, line 4:
```suggestion
  name: central-db-password
```

operator/tests/central/basic/76-assert.yaml
```diff
@@ -0,0 +1,31 @@
...
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Gi
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: central-db-secret
```
misberner, line 29:
```suggestion
  name: central-db-password
```

operator/tests/central/basic/76-assert.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+collectors:
+- type: pod
+  selector: app=central-db
+  tail: -1
+---
+apiVersion: apps/v1
+kind: Deployment
```
porridge, line 9: Might be good to also assert health of the central deployment in addition to central-db's health?

operator/pkg/central/values/translation/translation.go
```diff
@@ -204,11 +204,17 @@ func getCentralDBComponentValues(c *platform.CentralDBSpec) *translation.ValuesB
...
-	if c.ConnectionStringOverride != nil && c.PasswordSecret == nil {
-		cv.SetError(errors.New("if connection string override is set, then a password secret must also be set"))
-	}
-
 	if c.ConnectionStringOverride != nil {
+		if c.GetPersistence() != nil {
+			cv.SetError(errors.New("if a connection string is provided, no persistence settings must be supplied"))
+		}
+
+		// TODO: there are other settings which are ignored in external mode - should we error if those are set, too?
```
porridge, line 212: +1 for erroring loud and clear on dubious configurations

porridge (approved):

operator/tests/central/basic/75-enable-central-db-external.yaml
```diff
@@ -0,0 +1,17 @@
...
+data:
+  password: c2VjcmV0Cg== # "secret\n"
+---
+apiVersion: platform.stackrox.io/v1alpha1
+kind: Central
+metadata:
+  name: stackrox-central-services
+spec:
+  central:
+    db:
```
vladbologa, line 14: Not a code review comment, just a question: from what I understand, the `ROX_POSTGRES_DATASTORE` env var must be true for a Central to support Postgres. Are these tests executed for both cases (true and false, where RocksDB is still used) or how does this work?

operator/pkg/central/extensions/reconcile_central_db_password.go
```diff
@@ -10,57 +11,125 @@ import (
...
+	centralObj *platform.Central
+	password   string
+}
+
+func (r *reconcileCentralDBPasswordExtensionRun) readPasswordFromReferencedSecret(ctx context.Context) error {
+	if r.centralObj.Spec.Central.DB.GetPasswordSecret() == nil {
+		return errors.New("no password secret was specified")
+	}
+
+	passwordSecretName := r.centralObj.Spec.Central.DB.PasswordSecret.Name
```
SimonBaeumer, line 51:
```suggestion
	passwordSecretName := r.centralObj.Spec.Central.DB.GetPasswordSecret().Name
```

operator/pkg/central/extensions/reconcile_central_db_password.go
```diff
@@ -10,57 +11,125 @@ import (
...
+	if err := r.Client().Get(ctx, key, passwordSecret); err != nil {
+		return errors.Wrapf(err, "failed to retrieve central db password secret %q", passwordSecretName)
+	}
+
+	password, err := passwordFromSecretData(passwordSecret.Data)
+	if err != nil {
+		return errors.Wrapf(err, "reading central db password from secret %s", passwordSecretName)
+	}
+
+	r.password = password
```
SimonBaeumer, line 64: Why not returning the password and setting it from the main function? It feels odd to me because the result of that function is not returned, instead in sets state on the struct.
Setting it from the caller makes it more explicit to me.

operator/pkg/central/extensions/reconcile_central_db_password_test.go
```diff
@@ -0,0 +1,256 @@
...
+		Data: map[string][]byte{},
+	}
+
+	customPWSecretWithPW1 := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      customPWSecretName,
+			Namespace: testutils.TestNamespace,
+		},
+		Data: map[string][]byte{
+			"password": []byte(fmt.Sprintf("%s\n", pw1)),
```
SimonBaeumer, line 47: Nice job, did not thought about this test-case.

operator/pkg/central/extensions/reconcile_pvc.go
```diff
@@ -67,21 +68,33 @@ func convertDBPersistenceToPersistence(p *platform.DBPersistence) *platform.Pers
 	}
 }
 
-func getPersistenceByClaimName(central *platform.Central, claim string) *platform.Persistence {
-	switch claim {
-	case DefaultCentralPVCName:
-		return central.Spec.Central.GetPersistence()
-	case DefaultCentralDBPVCName:
-		return convertDBPersistenceToPersistence(central.Spec.Central.DB.GetPersistence())
+func getPersistenceByTarget(central *platform.Central, target PVCTarget) *platform.Persistence {
```
SimonBaeumer, line 71: Can you add a small doc string what this function is used for? I found it hard to understand, looking into the caller also did not really helped me.

SimonBaeumer (changes requested):

operator/pkg/central/values/translation/translation.go
misberner: Yeah, I think now I remember the problem, but that problem is even bigger.

The OpenShift console has the very unfortunate property of importing OpenAPI field defaults into the form _and then acting as if the user has set those field values_. See:

https://user-images.githubusercontent.com/2822367/198082234-7d0a986b-e903-44c6-828d-f1c233875a6e.mov


Currently there's no issue because Gorman didn't introduce a default value for the PVC name for the DB, but, arguably, there should be one, as we also have a `stackrox-db` default for the central persistence, and setting a default for a field is really nothing very unreasonable.

The bigger problem now is that setting a default for any of the fields means a user using OpenShift console can no longer disable central DB. OpenShift console simply does not respect the fact that (non-)nullness of fields in the CR might carry any meaning, it uses a lossy round-trip between the two representations. This, by the way, also applies in the other direction: with the existing CR structure, it is impossible to specify "use central DB, but with default settings" in the form view in OpenShift console, you have to set any of the fields (such as explicitly typing in the default `central-db` pvc name) or use YAML view.

@porridge I know you had some discussion with Gorman on the original PR re: an explicit boolean field vs the mere presence of another field (this was about external, not the DB as a whole). Putting the aspect of bool vs enum aside: what is your opinion on (a) avoiding explicit enabled/disabled fields and (b) erroring on inconsistent combinations of set fields, in light of these OpenShift console form view realities?

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -164,12 +164,9 @@ type CentralDBSpec struct {
...
 	PasswordSecret *LocalSecretReference `json:"passwordSecret,omitempty"`
 
-	// Disable database password generation. Do not use this for first-time installations in which the operator
-	// is managing Central DB as Central will have no way to connect to the database.
-	//+operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:hidden"}
-	PasswordGenerationDisabled *bool `json:"passwordGenerationDisabled,omitempty"`
-
 	// Specify a connection string that corresponds to an existing database. If set, the operator will not manage Central DB.
+	// When using this option, you must explicitly set a password secret; automatically generating a password will not
+	// be supported.
```
porridge, line 169:
```suggestion
	// When using this option, you must explicitly set a password secret; a password will not be automatically generated.
```

operator/pkg/central/values/translation/translation.go
porridge: Oh, wow. This is a game-changer. :disappointed:
I guess this means that mere presence of a struct can no longer have any meaning.
This in turn means `CentralDBEnabled()` needs to look at a new explicit toggle (I would very much prefer an enum on the `CentralComponentSpec` level rather than bools) rather than at the nilness of `.DB` :thinking:

Please also mention in a (non-user visible) comment somewhere why we don't look at struct nilness. This is probably not obvious for a non-openshift developer.

operator/pkg/central/extensions/reconcile_central_db_password_test.go
```diff
@@ -0,0 +1,256 @@
...
+					DB: &v1alpha1.CentralDBSpec{
+						PasswordGenerationDisabled: pointers.Bool(false),
+						PasswordSecret: &v1alpha1.LocalSecretReference{
+							Name: canonicalCentralDBPasswordSecretName,
+						},
+					},
+				},
+			},
+			ExpectedError: "when explicitly enabling password generation, a password secret must not be specified",
+		},
```
vladbologa, line 235:
```suggestion
```

These tests should be removed.

vladbologa (changes requested):

SimonBaeumer (approved):

vladbologa (approved):

PR #3323, opened 2022-10-05
ROX-12895: Scaffold out PatternFly Network Graph

vjwilson:
## Description

I adapted Saif's demo of the sidebar into the PF route for our new Network Graph, so that we could have a real playground to test things out. I want to bootstrap an area where we can start testing and trying things out in the app.

We could merge as-is, but I welcome suggestions even for this bare-bones version.

For my part, I have the following concerns:
1. I had to copy over a lot of ancillary files just to get this simple example to work. I didn't really think about it in looking at the original demo, or your new demos.
2. Besides the `react-topology` library itself, and `d3-polygon` which I understand, I had to add `mobx` and `mobx-react` just to get the reactive movement on the graph, and `classname` which they use for conditional class names in the styles of some their helper components. (We can use template strings instead of classnames lib, but honestly that added dep bothers me least.)
3. The demo code is in TypeScript, but they do not have strict TS rules. I had to fix, fudge, or downright ts-ignore a lot of errors just to get this code to run.
4. Getting the graph and its sidebar to span edge-to-edge in a page section required me to override some CSS vars. Seems like someone on our team had a more elegant solution for this, maybe?

## Checklist

## Testing Performed

Screen Shot 2022-10-04 at 6 18 15 PM
Screen Shot 2022-10-04 at 6 18 21 PM


ui/apps/platform/src/Containers/NetworkGraph/components/shapes/Polygon.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <polygon
+            className={className}
+            ref={dndDropRef}
+            points={points.map((p) => `${p[0]},${p[1]}`).join(' ')}
+            filter={filter}
+        />
+    );
+};
+
+export default Polygon;
```
alwayshooin, line 32: you don't have to add this file

ui/apps/platform/src/Containers/NetworkGraph/components/defaultComponentFactory.tsx
```diff
@@ -0,0 +1,48 @@
...
+    | ComponentType<{ element: GraphElement }>
+    | typeof MultiEdge
+    | typeof Group
+    | typeof GroupHull
+    | typeof GraphComponent
+    | typeof DefaultNode
+    | typeof Edge
+    | typeof undefined => {
+    switch (type) {
+        case 'multi-edge':
```
alwayshooin, line 28: you can remove this case

ui/apps/platform/src/Containers/NetworkGraph/components/defaultComponentFactory.tsx
```diff
@@ -0,0 +1,48 @@
...
+            return MultiEdge;
+        case 'group':
+            return Group;
+        case 'group-hull':
```
alwayshooin, line 32: this one too

ui/apps/platform/src/Containers/NetworkGraph/components/stylesComponentFactory.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            // @ts-ignore
+                            withSelection()(StyleNode)
+                        )
+                    )
+                )
+            );
+        case 'node-path':
+            return CustomPathNode;
+        case 'node-polygon':
+            return CustomPolygonNode;
```
alwayshooin, line 113: you can get rid of `node-path` and `node-polygon` as we're not using them in this scenario

alwayshooin (approved):
thanks for getting this out of the way, van! it _is_ a lot of copy/paste, so i left some comments on files that don't need to be here for this specific scenario to keep things simple. we can add them in later as we see fit, but i think for now it just clutters the directory imo.

PR #3331, opened 2022-10-06
ROX-12983: Add flag to disallow auto-upgrade toggle

fredrb:
Follow up from #3198 

## Description

When auto-upgrades are not allowed, the FE should show a message instead of the toggle. 

The goal of this PR is to disable the auto-upgrade toggle when in Managed Centrals.

## Checklist

I don't think it makes sense to update the readme since this will not affect users directly -- it will be a config specific to managed centrals. 

## Testing Performed

When on a managed central, the following message is displayed in the cluster management screen:

image

API response on managed centrals:
```
{"config":{"enableAutoUpgrade":false,"autoUpgradeFeature":"NOT_SUPPORTED"}}
```


ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
```diff
@@ -8,7 +8,10 @@ import {
 } from 'services/ClustersService';
 
 function AutoUpgradeToggle(): ReactElement {
-    const [autoUpgradeConfig, setAutoUpgradeConfig] = useState<AutoUpgradeConfig>({});
+    const [autoUpgradeConfig, setAutoUpgradeConfig] = useState<AutoUpgradeConfig>({
+        enableAutoUpgrade: false,
+    });
```
pedrottimark, line 13: Here is a simpler change than what I suggested on Friday:

```ts
const [autoUpgradeConfig, setAutoUpgradeConfig] = useState<AutoUpgradeConfig | null>(null);
```

Delete the following `useState` line and `finally` block.

Edit condition of `if` statement below.
```ts
    if (!autoUpgradeConfig) {
        return <></>;
    }
```

ui/apps/platform/src/services/ClustersService.ts
```diff
@@ -99,6 +99,7 @@ export function fetchClusterWithRetentionInformation(id: string): Promise<Cluste
 
 export type AutoUpgradeConfig = {
     enableAutoUpgrade?: boolean;
+    autoUpgradeAllowed?: 'NOT_ALLOWED' | 'ALLOWED';
```
pedrottimark, line 102: You can delete `?` if the response object always includes the properties.

ui/apps/platform/src/services/ClustersService.ts
```diff
@@ -114,7 +115,7 @@ export function getAutoUpgradeConfig(): Promise<AutoUpgradeConfig> {
  * Saves the cluster autoupgrade config.
  */
 export function saveAutoUpgradeConfig(config: AutoUpgradeConfig): Promise<AutoUpgradeConfig> {
```
pedrottimark, line 117: 2 problems in original frontend service function are worthwhile to fix, although unrelated to your changes.

1. The return type is inconsistent with backend service proto:

    ```proto
    rpc UpdateSensorUpgradeConfig(UpdateSensorUpgradeConfigRequest) returns (Empty) {
    ```

    Replace `Promise<AutoUpgradeConfig>` with `Promise<Empty>` a type alias we have defined.

2. The return value leaks the entire response object. Make it consistent with correct return type:

    ```ts
    return axios.post(autoUpgradeConfigUrl, wrappedObject).then((response) => response.data);
    ```

pedrottimark:
With 3 suggested changes to ui files, the frontend files fit the requirements. I will let backend reviewers approve.

ui/apps/platform/src/services/ClustersService.ts
fredrb: I don't understand (2). If we change it to return `Promise<Empty>` why would we even return something?

ui/apps/platform/src/services/ClustersService.ts
fredrb: Done

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
fredrb: Done

ui/apps/platform/src/services/ClustersService.ts
pedrottimark: Excellent question. The original code returns the entire `response` object:

https://axios-http.com/docs/res_schema

The recommended code returns just the `data` property of the response which in this case is empty object `{}` which corresponds to `Promise<Empty>` return type.

pedrottimark: In this case, calling code has only `catch` block for error and no `then` block for response, as it should not:

```ts
        saveAutoUpgradeConfig(newConfig).catch(() => {
            // reverse the optimistic update of the control in the UI
            const rollbackConfig = {
                ...autoUpgradeConfig,
                enableAutoUpgrade: previousValue,
            };
            setAutoUpgradeConfig(rollbackConfig);

            // also, re-fetch the data from the server, just in case it did update but we didn't get the network response
            fetchConfig();
        });
```

By the way, `rollbackConfig` explains the behavior you noticed when request has 500 status.

1. First change helps TypeScript compiler report error if calling code accesses properties of response data.
2. Second change actually returns empty object which has no properties, so value is consistent with type.

pedrottimark: Another way to think about it: `then((response) => response.data)` is common idiom when response to POST or PUT request returns the created or updated data. It is just as necessary when response data is empty object.

msugakov:
I observe a bug locally, after deploying your branch with `deploy/k8s/deploy-local.sh`. Did you see the same effect?

Peek 2022-10-12 16-17

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
```diff
@@ -25,11 +26,16 @@ function AutoUpgradeToggle(): ReactElement {
...
+        if (!autoUpgradeConfig) {
+            return;
+        }
+
         // @TODO, wrap this settings change in a confirmation prompt of some sort
         const previousValue = autoUpgradeConfig.enableAutoUpgrade;
         const newConfig = {
             ...autoUpgradeConfig,
             enableAutoUpgrade: !previousValue,
+            autoUpgradeAllowed: Upgradability.NOT_ALLOWED,
```
pedrottimark, line 38: @fredrb This is reason for the problem that Misha reports.

ui/apps/platform/src/services/ClustersService.ts
```diff
@@ -97,8 +97,14 @@ export function fetchClusterWithRetentionInformation(id: string): Promise<Cluste
     });
 }
 
+export enum Upgradability {
+    ALLOWED = 'ALLOWED',
+    NOT_ALLOWED = 'NOT_ALLOWED',
+}
```
pedrottimark, line 103: Although not a blocker to merge, string union type is more fluent TypeScript than enum type.

Historical note: main reason for enum type was at beginning of migration from JavaScript to TypeScript, calling code might have used an enum-like idiom using objects and properties. Because UI is making steady progress to TypeScript `'ALLOWED'` is simpler and more idiomatic.

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
fredrb: The problem is that this won't allow for `undefined` if the property definition isn't defined as `autoUpgradeAllowed?`. Should I rollback the other suggestion and allow the `autoUpgradeConfig` to have this as `undefined`?

pedrottimark: For now, replace with `Upgradability.ALLOWED` because a prerequisite to render the toggle switch, true?

After a meeting, I will take a look.

pedrottimark: Aha, notice the early return you added at the beginning of this function?

```js
        if (!autoUpgradeConfig) {
            return;
        }
```

TypeScript error because corresponding `handleChange` function in the PF file needs it too.

Bonus follow up for me is roll back looks incorrect in that function.

ui/apps/platform/src/services/ClustersService.ts
pedrottimark: I will comment here: is it possible to delete `?` from `enableAutoUpgrade` property below?

central/sensorupgradeconfig/datastore/datastore_impl.go
```diff
@@ -57,6 +57,16 @@ func (d *dataStore) UpsertSensorUpgradeConfig(ctx context.Context, sensorUpgrade
 		return errors.New("auto-upgrade not allowed on managed central")
```
msugakov, line 57: What concerns me the most is that `env.ManagedCentral.BooleanSetting()` is used in multiple files to figure out eligibility for auto-upgrades. This means we have a leaky abstraction. A slightly better version of it could be to define a function that derives `SensorAutoUpgradeFeatureStatus` based on `env.ManagedCentral.BooleanSetting()` and use it everywhere.

However, that doesn't address my other commends, and I think, the root of the problem is that you're trying to persist `auto_upgrade_allowed` attribute in a database while in fact it is automatically derived. There's no need to persist something that's automatically derived.

To make things better you could:
1. Remove `SensorAutoUpgradeFeatureStatus auto_upgrade_allowed = 2;` from `proto/storage/sensor_upgrade.proto`, i.e. not try to persist it.
2. Make `SensorAutoUpgradeFeatureStatus` a part of `GetSensorUpgradeConfigResponse`.
3. Remove any mentions of `env.ManagedCentral.BooleanSetting()` and `auto_upgrade_allowed`-derivatives from `central/sensorupgradeconfig/datastore/datastore_impl.go`
4. Compute `SensorAutoUpgradeFeatureStatus` in `central/sensorupgrade/service/service_impl.go`.

This should leave you with a hardly testable `addDefaultConfigIfEmpty` in `central/sensorupgradeconfig/datastore/singleton.go` that needs to derive `SensorAutoUpgradeFeatureStatus` from `env.ManagedCentral.BooleanSetting()`. Let's think how to simplify that one. I have a strong suspicion that that method is actually redundant and the calculation of defaults can be done elsewhere.

Another thing that will remain in question is the default value for `dataStore.autoTrigger`. I think it can be safely pushed up to be a part of `central/sensorupgrade/service/service.go`, i.e. the _service_ contract, not the _datasource_ contract. There, in `service_impl.go` the defaults can be applied even in the case when datastore doesn't hold any value.

msugakov (changes requested):
Partially checked the backend part.

What I'm interested to see as part of "testing done" is how would cluster details UI look when ACS is deployed with managed services flag but Sensor is configured to some older image.

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
pedrottimark: Three strikes and I am out! JavaScript hoists `function` to top of scope above the early return.

But `const` assignment with arrow function is where it is, so early return does indeed imply to TypeScript that `autoUpgradeConfig` is not null.

It was an intentional decision to delete TODO comment.

```ts
    if (!autoUpgradeConfig) {
        return <></>;
    }

    const toggleAutoUpgrade = () => {
        const previousValue = autoUpgradeConfig.enableAutoUpgrade;
        const newConfig = {
            ...autoUpgradeConfig,
            enableAutoUpgrade: !previousValue,
        };

        setAutoUpgradeConfig(newConfig); // optimistically set value before API call

        saveAutoUpgradeConfig(newConfig).catch(() => {
            // reverse the optimistic update of the control in the UI
            const rollbackConfig = {
                ...autoUpgradeConfig,
                enableAutoUpgrade: previousValue,
            };
            setAutoUpgradeConfig(rollbackConfig);

            // also, re-fetch the data from the server, just in case it did update but we didn't get the network response
            fetchConfig();
        });
    };
```

Change for the PF component. I added comment on line where I fixed mistake in original code.

```ts
    if (!autoUpgradeConfig) {
        return <></>;
    }

    const handleChange = (value: boolean) => {
        setIsDisabled(true);
        const newConfig = {
            ...autoUpgradeConfig,
            enableAutoUpgrade: value,
        };

        setAutoUpgradeConfig(newConfig); // optimistically set value before API call

        saveAutoUpgradeConfig(newConfig)
            .then(() => {
                setIsDisabled(false);
            })
            .catch(() => {
                // reverse the optimistic update of the control in the UI
                const rollbackConfig = {
                    ...autoUpgradeConfig,
                    enableAutoUpgrade: !value, // previous value is opposite to new value
                };
                setAutoUpgradeConfig(rollbackConfig);

                // also, re-fetch the data from the server, just in case it did update but we didn't get the network response
                fetchConfig();
            });
    };
```

You have my apology on behalf of UI team that the code was not firm foundation for you to build on.

Because so many changes, I will write draft integration tests now as a follow up for after you merge.

central/sensorupgradeconfig/datastore/datastore_impl.go
fredrb: That makes sense.

fredrb:
@msugakov @pedrottimark I've addressed all your comments. Let me know what do you guys think. 

> What I'm interested to see as part of "testing done" is how would cluster details UI look when ACS is deployed with managed services flag but Sensor is configured to some older image.

Upgrade available icon will still be shown if sensor is behind central on a manifest-based installation.
image

central/sensorupgradeconfig/datastore/singleton.go
```diff
@@ -37,7 +38,10 @@ func addDefaultConfigIfEmpty(d DataStore) error {
 	}
```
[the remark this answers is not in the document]
fredrb: Not relevant anymore after refactor.

proto/storage/sensor_upgrade.proto
```diff
@@ -5,8 +5,15 @@ option java_package = "io.stackrox.proto.storage";
 
 package storage;
 
+enum SensorAutoUpgrade {
+    ALLOWED = 0;
+    NOT_ALLOWED = 1;
+}
+
```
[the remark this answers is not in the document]
fredrb: Updated

proto/storage/sensor_upgrade.proto
```diff
@@ -5,8 +5,15 @@ option java_package = "io.stackrox.proto.storage";
...
 // SensorUpgradeConfig encapsulates configuration relevant to sensor auto-upgrades.
 message SensorUpgradeConfig {
     // Whether to automatically trigger upgrades for out-of-date sensors.
     bool enable_auto_upgrade = 1;
+    // Whether auto-upgrades can be set to true
```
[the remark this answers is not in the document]
fredrb: Not relevant after the refactor

central/sensorupgrade/service/service_impl.go
```diff
@@ -64,7 +64,13 @@ func (s *service) UpdateSensorUpgradeConfig(ctx context.Context, req *v1.UpdateS
 	if req.GetConfig() == nil {
 		return nil, errors.Wrap(errox.InvalidArgs, "need to specify a config")
 	}
-	if err := s.configDataStore.UpsertSensorUpgradeConfig(ctx, req.GetConfig()); err != nil {
+	storedConfig, err := s.configDataStore.GetSensorUpgradeConfig(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get existing config")
+	}
+	storedConfig.EnableAutoUpgrade = req.GetConfig().GetEnableAutoUpgrade()
```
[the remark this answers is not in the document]
fredrb: Added. Thanks

ui/apps/platform/src/services/ClustersService.ts
fredrb: Done

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
fredrb: Done in pair

msugakov (commented):
Only checked a part of back-end.

pedrottimark:
Fred, frontend changes are good to go. Thank you for both pair programming sessions!

Misha can approve when requested changes for backend have been completed.

central/sensorupgradeconfig/datastore/singleton.go
```diff
@@ -32,12 +32,11 @@ func addDefaultConfigIfEmpty(d DataStore) error {
...
 		return errors.Wrap(err, "failed to check initial sensor upgrade config")
 	}
-	if currentConfig != nil && (!env.ManagedCentral.BooleanSetting() || !currentConfig.GetEnableAutoUpgrade()) {
+	if currentConfig != nil {
 		return nil
 	}
 
-	// Auto upgrade is disabled by default if managed central flag is set
-	return d.UpsertSensorUpgradeConfig(ctx, upgradeConfig(!env.ManagedCentral.BooleanSetting()))
+	return d.UpsertSensorUpgradeConfig(ctx, upgradeConfig(true))
```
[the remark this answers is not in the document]
fredrb: Oh, I dropped the ball here, it definitely shouldn't be defaulted to `true`.

However, I'm a bit uneasy with setting this in the `GetSensorUpgradeConfig` thought. It still feels like it should be the behavior of the datastore startup. Another issue is that the service context comes with the user's scoped access. Which might not allow write/updates in the DB. Resulting in a code like this:

```go
func (s *service) insertDefaultSensorConfig(ctx context.Context) (*storage.SensorUpgradeConfig, error) {
	// This is needed in case the first `GET` request to SensorUpgradeConfig is done by a user
	// without permission to add sensor configs. We still want to create the default
	// config entry.
	allAccessContext := sac.WithAllAccess(ctx)
	config := defaultUpgradeConfig()
	if err := s.configDataStore.UpsertSensorUpgradeConfig(allAccessContext, config); err != nil {
		return nil, err
	}
	return config, nil
}
``` 
Pushed a commit with this. Take a look and let me know what do you think.

proto/api/v1/sensor_upgrade_service.proto
```diff
@@ -10,7 +10,11 @@ import "storage/sensor_upgrade.proto";
...
 
-message SensorToggleConfig {
+message UpdateSensorUpgradeConfigRequest {
+    storage.SensorUpgradeConfig config = 1;
+}
+
+message GetSensorUpgradeConfigResponse {
     enum SensorAutoUpgradeFeatureStatus {
         SUPPORTED = 0;
         NOT_SUPPORTED = 1;
```
msugakov, line 20:
```suggestion
        NOT_SUPPORTED = 0;
        SUPPORTED = 1;
```

central/sensorupgradeconfig/datastore/singleton.go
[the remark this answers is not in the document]
msugakov: My main point is: why persist (insert into the database) the default value at all? If all requests go through Service, the Service's `GetSensorUpgradeConfig()` can just return the `defaultUpgradeConfig()` when DataStore returns `nil`.

Same logic for `autoTrigger`, set it according to what `GetSensorUpgradeConfig()` returns upon Service startup.

If you still want to insert the default value in the database, do it through `GetSensorUpgradeConfig()` but construct all-mighty `Context` before calling that function.

central/sensorupgradeconfig/datastore/datastore.go
```diff
@@ -5,23 +5,17 @@ import (
...
 //go:generate mockgen-wrapper
 type DataStore interface {
 	GetSensorUpgradeConfig(context.Context) (*storage.SensorUpgradeConfig, error)
 	UpsertSensorUpgradeConfig(context.Context, *storage.SensorUpgradeConfig) error
-
-	AutoTriggerSetting() *concurrency.Flag
 }
 
 // New returns a new, ready-to-use, datastore.
 func New(store store.Store) (DataStore, error) {
```
msugakov, line 18: [nit] No need to declare `error` if it is always `nil`.

```suggestion
func New(store store.Store) DataStore {
```

msugakov (commented):
Did not check tests (and UI).

msugakov (commented):
Thanks! _Looks_ correct now.

central/sensorupgrade/service/service_impl.go
```diff
@@ -66,26 +71,30 @@ func (s *service) wrapToggleResponse(config *storage.SensorUpgradeConfig) *v1.Ge
...
 		return nil, err
 	}
 	if config == nil {
 		// If there's no config in the DB, return default config according to managed central flag
-		config = defaultUpgradeConfig()
+		if getAutoUpgradeFeatureStatus() == v1.GetSensorUpgradeConfigResponse_SUPPORTED {
+			return &storage.SensorUpgradeConfig{EnableAutoUpgrade: true}, nil
+		}
+		return &storage.SensorUpgradeConfig{EnableAutoUpgrade: false}, nil
+	}
```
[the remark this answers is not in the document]
fredrb: The only thing that I don't like about upserting in  the table here is that the user that originated the request might not  have write access.

We can still do the same thing we did in initialize (i.e. `sac.WithAllAccess(context.Background())`) and use a permissive context, but I'm a bit paranoid about arbitrarily setting `All` access in the request path.

central/sensorupgrade/service/service_impl.go
```diff
@@ -35,6 +40,14 @@ var (
 type service struct {
 	configDataStore datastore.DataStore
 	manager         connection.Manager
+	autoTriggerFlag concurrency.Flag
+}
+
+func (s *service) initialize() {
+	ctx := sac.WithAllAccess(context.Background())
+	defaultConfig, err := s.getOrDefaultUpgradeConfig(ctx)
+	utils.CrashOnError(err)
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgrade/service/service_impl.go
```diff
@@ -49,24 +62,62 @@ func (s *service) AuthFuncOverride(ctx context.Context, fullMethodName string) (
...
+	if req.GetConfig().GetEnableAutoUpgrade() && getAutoUpgradeFeatureStatus() == v1.GetSensorUpgradeConfigResponse_NOT_SUPPORTED {
+		return nil, errors.Wrap(errox.InvalidArgs, "auto-upgrade not supported on managed ACS")
+	}
+
 	if err := s.configDataStore.UpsertSensorUpgradeConfig(ctx, req.GetConfig()); err != nil {
 		return nil, err
 	}
+	s.autoTriggerFlag.Set(req.GetConfig().EnableAutoUpgrade)
 	return &v1.Empty{}, nil
 }
```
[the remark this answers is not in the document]
fredrb: Why? I thought that we were okay with the fact that you  could still trigger upgrades on manifest installations via the UI. We  just don't allow automatically triggering them when the versions  mismatch.

This endpoint is responsible for the "Upgrade available"  button, when auto-upgrades are disabled, but there is a new sensor  version to be installed:

image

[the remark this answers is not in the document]
msugakov: Sorry, I keep forgetting about this.

central/sensorupgrade/service/service_impl.go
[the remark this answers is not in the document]
msugakov: Well, maybe then the upsertion should be done in `initialize()` in case the value read from db is `nil`.
We're sort of back to the original approach that existed there, but not quite. The main difference is that updating the database was done there in Singleton which isn't the most testable thing. Here we'll do that in the instance initialization which _can_ be unit tested because it actually represents the unit.
If you're adventurous, give a spike and see if you'd like the resulting code.

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -1,11 +1,175 @@
...
+	for envValue, expectations := range testCases {
+		s.Run(fmt.Sprintf("ROX_MANAGED_CENTRAL=%v", envValue), func() {
+			s.dataStore.EXPECT().GetSensorUpgradeConfig(gomock.Any()).Times(1).Return(nil, nil)
+			s.dataStore.EXPECT().UpsertSensorUpgradeConfig(gomock.Any(), &UpgradeConfigMatcher{expectations.expectedAutoUpdate})
+			s.isolator.Setenv(env.ManagedCentral.EnvVar(), envValue)
+
+			instance, err := New(s.dataStore, s.manager)
+			s.NoError(err)
+
+			s.dataStore.EXPECT().GetSensorUpgradeConfig(gomock.Any()).Times(1).Return(configWith(expectations.expectedAutoUpdate), nil)
```
msugakov, line 143: Maybe between these things validate that `instance.AutoUpgradeSetting().Get()` returns `expectations.expectedAutoUpdate` because that's why we need `initialize()`.

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -1,11 +1,175 @@
...
+		},
+		"false": {
+			expectedAutoUpdate:     true,
+			expectedFeatureEnabled: v1.GetSensorUpgradeConfigResponse_SUPPORTED,
+		},
+	}
+
+	for envValue, expectations := range testCases {
+		s.Run(fmt.Sprintf("ROX_MANAGED_CENTRAL=%v", envValue), func() {
+			s.dataStore.EXPECT().GetSensorUpgradeConfig(gomock.Any()).Times(1).Return(nil, nil)
```
msugakov, line 136: Do you mind adding a test (or a case) that checks a case when `dataStore.GetSensorUpgradeConfig()` returns non-`nil`. I.e. we `initialize()` Service while DataStore **has** a value.

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -1,11 +1,175 @@
...
+
+func (m *UpgradeConfigMatcher) String() string {
+	return fmt.Sprintf("auto-upgrade enabled: %v", m.autoUpgrade)
+}
+
+func (s *SensorUpgradeServiceTestSuite) TestAuthzWorks() {
+	s.dataStore.EXPECT().GetSensorUpgradeConfig(gomock.Any()).Times(1).Return(configWith(true), nil)
+	serviceInstance, err := New(s.dataStore, s.manager)
+	s.NoError(err)
+	testutils.AssertAuthzWorks(s.T(), serviceInstance)
```
msugakov, line 174: This looks like magic to me so I trust you here. Thanks for adding such test!

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
```diff
@@ -24,7 +24,15 @@ function AutoUpgradeToggle(): ReactElement {
         fetchConfig();
     }, []);
 
-    function toggleAutoUpgrade(): void {
+    if (!autoUpgradeConfig) {
+        return <></>;
+    }
+
+    if (autoUpgradeConfig.autoUpgradeFeature === 'NOT_SUPPORTED') {
+        return <>Auto upgrade not allowed in managed central</>;
```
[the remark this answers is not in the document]
pedrottimark: Yes, thank you for looking from the outside-in with customer viewpoint.

central/sensorupgrade/service/service_impl.go
```diff
@@ -73,26 +73,38 @@ func (s *service) wrapToggleResponse(config *storage.SensorUpgradeConfig) *v1.Ge
...
+// in the DB, create one based on the default value.
+func (s *service) getOrCreateSensorUpgradeConfig(ctx context.Context) (*storage.SensorUpgradeConfig, error) {
 	config, err := s.configDataStore.GetSensorUpgradeConfig(ctx)
 	if err != nil {
 		return nil, err
 	}
 	if config == nil {
 		// If there's no config in the DB, return default config according to managed central flag
+		// and insert the value
+		var defaultConfig *storage.SensorUpgradeConfig
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgradeconfig/datastore/datastore.go
```diff
@@ -5,23 +5,17 @@ import (
...
 
 // New returns a new, ready-to-use, datastore.
-func New(store store.Store) (DataStore, error) {
+func New(store store.Store) DataStore {
 	ds := &dataStore{store: store}
-	if err := ds.initialize(); err != nil {
-		return nil, err
-	}
-	return ds, nil
+	return ds
```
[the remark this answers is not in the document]
fredrb: Done

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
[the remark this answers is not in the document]
fredrb: Done

ui/apps/platform/src/Containers/Clusters/PF/ClustersTable/AutoUpgradeToggle.tsx
```diff
@@ -30,7 +28,15 @@ function AutoUpgradeToggle(): ReactElement {
         fetchConfig();
     }, []);
 
-    function handleChange(value) {
+    if (!autoUpgradeConfig) {
+        return <></>;
+    }
+
+    if (autoUpgradeConfig.autoUpgradeFeature === 'NOT_SUPPORTED') {
+        return <>Auto upgrade not allowed in managed central</>;
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgrade/service/service_impl.go
```diff
@@ -83,16 +83,15 @@ func (s *service) getOrCreateSensorUpgradeConfig(ctx context.Context) (*storage.
 	if config == nil {
 		// If there's no config in the DB, return default config according to managed central flag
 		// and insert the value
-		var defaultConfig *storage.SensorUpgradeConfig
+		// var defaultConfig *storage.SensorUpgradeConfig
```
msugakov, line 86:
```suggestion
```

central/sensorupgrade/service/service_impl_test.go
fredrb: Done

central/sensorupgrade/service/service_impl_test.go
fredrb: Done

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
```diff
@@ -24,7 +24,15 @@ function AutoUpgradeToggle(): ReactElement {
         fetchConfig();
     }, []);
 
-    function toggleAutoUpgrade(): void {
+    if (!autoUpgradeConfig) {
+        return <></>;
+    }
+
+    if (autoUpgradeConfig.autoUpgradeFeature === 'NOT_SUPPORTED') {
```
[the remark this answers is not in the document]
fredrb: Done

ui/apps/platform/src/services/ClustersService.ts
```diff
@@ -102,6 +102,13 @@ export type AutoUpgradeConfig = {
     autoUpgradeFeature: 'SUPPORTED' | 'NOT_SUPPORTED';
 };
 
+/**
+ * Checks is auto upgrade is supported
+ */
+export function isAutoUpgradeSupported(autoUpgradeConfig: AutoUpgradeConfig) {
+    return autoUpgradeConfig.autoUpgradeFeature !== 'SUPPORTED';
+}
```
msugakov, line 110: This should be the opposite
```suggestion
export function isAutoUpgradeSupported(autoUpgradeConfig: AutoUpgradeConfig) {
    return autoUpgradeConfig.autoUpgradeFeature === 'SUPPORTED';
}
```

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
```diff
@@ -28,7 +29,7 @@ function AutoUpgradeToggle(): ReactElement {
         return <></>;
     }
 
-    if (autoUpgradeConfig.autoUpgradeFeature === 'NOT_SUPPORTED') {
+    if (isAutoUpgradeSupported(autoUpgradeConfig)) {
```
msugakov, line 32: This too
```suggestion
    if (!isAutoUpgradeSupported(autoUpgradeConfig)) {
```

ui/apps/platform/src/Containers/Clusters/PF/ClustersTable/AutoUpgradeToggle.tsx
```diff
@@ -32,7 +33,7 @@ function AutoUpgradeToggle(): ReactElement {
         return <></>;
     }
 
-    if (autoUpgradeConfig.autoUpgradeFeature === 'NOT_SUPPORTED') {
+    if (isAutoUpgradeSupported(autoUpgradeConfig)) {
```
msugakov, line 36: And this
```suggestion
    if (!isAutoUpgradeSupported(autoUpgradeConfig)) {
```

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -146,6 +161,7 @@ func (s *SensorUpgradeServiceTestSuite) Test_GetSensorUpgradeConfig_DefaultValue
 
 			s.Require().NoError(err)
 			s.Assert().Equal(expectations.expectedAutoUpdate, result.GetConfig().GetEnableAutoUpgrade())
+			s.Assert().Equal(expectations.expectedAutoUpdate, instance.AutoUpgradeSetting().Get())
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -109,13 +110,27 @@ func (s *SensorUpgradeServiceTestSuite) Test_UpdateSensorUpgradeConfig() {
...
 			}
 		})
 	}
 }
 
+func (s *SensorUpgradeServiceTestSuite) Test_GetSensorUpgradeConfig() {
+	s.isolator.Setenv(env.ManagedCentral.EnvVar(), "false")
+	s.dataStore.EXPECT().GetSensorUpgradeConfig(gomock.Any()).
+		Times(2).
+		Return(configWith(false), nil)
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -109,13 +110,27 @@ func (s *SensorUpgradeServiceTestSuite) Test_UpdateSensorUpgradeConfig() {
...
 			if testCase.upsertTimesCalled > 0 {
-				s.Require().Equal(serviceInstance.AutoUpgradeSetting().Get(),
+				s.Assert().Equal(serviceInstance.AutoUpgradeSetting().Get(),
 					testCase.req.GetConfig().GetEnableAutoUpgrade())
 			}
 		})
 	}
 }
 
+func (s *SensorUpgradeServiceTestSuite) Test_GetSensorUpgradeConfig() {
```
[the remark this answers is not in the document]
fredrb: Done

central/sensorupgrade/service/service_impl_test.go
```diff
@@ -109,13 +110,27 @@ func (s *SensorUpgradeServiceTestSuite) Test_UpdateSensorUpgradeConfig() {
...
+	serviceInstance, err := New(s.dataStore, s.manager)
+	s.NoError(err)
+
+	response, err := serviceInstance.GetSensorUpgradeConfig(context.Background(), &v1.Empty{})
+	s.NoError(err)
+
+	s.Assert().Equal(false, response.GetConfig().GetEnableAutoUpgrade())
+}
+
 func (s *SensorUpgradeServiceTestSuite) Test_GetSensorUpgradeConfig_DefaultValues() {
```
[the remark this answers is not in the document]
fredrb: Done

ui/apps/platform/src/Containers/Clusters/Components/AutoUpgradeToggle.tsx
fredrb: Done

ui/apps/platform/src/Containers/Clusters/PF/ClustersTable/AutoUpgradeToggle.tsx
fredrb: Done

msugakov (approved):
I need to trust that you checked both `ROX_MANAGED_CENTRAL` `true` and `false` (or absent) that UI works as expected and that auto-upgrader does not get triggered when `true`. If you don't feel confident in previous manual testing, please give it some time to make sure there are no regressions.

fredrb:
Prow tests look solid. Even `gke-upgrade-tests` seem to be passing. Did another manual test run and the toggle display is being controlled by the environment variable correctly.

PR #3341, opened 2022-10-06
ROX-11323 Allow the fake workloads to create services

lvalerom:
## Description

Enables the fake workloads to create services.

## Checklist

## Testing Performed

Generate a workload configMap like:

```yaml
deploymentWorkload:
- deploymentType: Deployment
  lifecycleDuration: 10m0s
  numDeployments: 2500
  numLifecycles: 0
  podWorkload:
    containerWorkload:
      numImages: 0
    lifecycleDuration: 2m0s
    numContainers: 3
    numPods: 5
    processWorkload:
      alertRate: 0.001
      processInterval: 30s
  updateInterval: 100s
networkWorkload:
  batchSize: 100
  flowInterval: 1s
nodeWorkload:
  numNodes: 1000
rbacWorkload:
  numBindings: 1000
  numRoles: 1000
  numServiceAccounts: 1000
serviceWorkload:
  numLabels: 10
  numClusterIPs: 2 
  numNodePorts: 2
  numLoadBalancers: 2
```

Start `local-sensor` using the previously mentioned configMap:

```
go run tools/local-sensor/main.go -record -with-fakeworkload=default.yaml
```

The fake workloads should create services.

### DIFF
--- a/sensor/kubernetes/fake/service.go
+++ b/sensor/kubernetes/fake/service.go
@@ -9,14 +9,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func getRandProtocol() string {
-	protocols := []string{"TCP", "UDP", "SCTP"}
-	return protocols[rand.Intn(len(protocols))]
-}
+var (
+	protocols  = [...]string{"TCP", "UDP", "SCTP"}
+	ipFamilies = [...]string{"IPv4", "IPv6"}
+)
 
-func getRandIP() string {
-	ip := generateIP()
-	return ip
+func getRandProtocol() string {
+	return protocols[rand.Intn(len(protocols))]
 }
 
 func getRandPort() uint32 {
@@ -24,7 +23,6 @@ func getRandPort() uint32 {
 }
 
 func getIPFamily() string {
-	ipFamilies := []string{"IPv4", "IPv6"}
 	return ipFamilies[rand.Intn(len(ipFamilies))]
 }
 
@@ -36,7 +34,7 @@ func getClusterIP(numLabels int) *v1.Service {
 	} else {
 		labels = createMap(numLabels)
 	}
-	clusterIP := getRandIP()
+	clusterIP := generateIP()
 	return &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      randStringWithLength(16),
@@ -71,7 +69,7 @@ func getNodePort(numLabels int) *v1.Service {
 	} else {
 		labels = createMap(numLabels)
 	}
-	clusterIP := getRandIP()
+	clusterIP := generateIP()
 	return &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      randStringWithLength(16),
@@ -107,7 +105,7 @@ func getLoadBalancer(numLabels int) *v1.Service {
 	} else {
 		labels = createMap(numLabels)
 	}
-	clusterIP := getRandIP()
+	clusterIP := generateIP()
 	internalTrafficPolicy := v1.ServiceInternalTrafficPolicyCluster
 	allocateLoadBalancerNodePorts := true
 	ipFamilyPolicy := v1.IPFamilyPolicySingleStack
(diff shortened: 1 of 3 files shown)


sensor/kubernetes/fake/service.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func getService(workload ServiceWorkload) []runtime.Object {
+	objects := make([]runtime.Object, 0, workload.NumClusterIPs+workload.NumNodePorts+workload.NumLoadBalancers)
+	for i := 0; i < workload.NumClusterIPs; i++ {
+		clusterIP := getClusterIP(workload.NumLabels)
+		objects = append(objects, clusterIP)
+	}
+	for i := 0; i < workload.NumNodePorts; i++ {
+		clusterIP := getNodePort(workload.NumLabels)
```
fredrb, line 162: Probably due to copy-paste
```suggestion
		numPort := getNodePort(workload.NumLabels)
```

sensor/kubernetes/fake/service.go
```diff
@@ -0,0 +1,170 @@
...
+		objects = append(objects, clusterIP)
+	}
+	for i := 0; i < workload.NumLoadBalancers; i++ {
+		clusterIP := getLoadBalancer(workload.NumLabels)
```
fredrb, line 166: Same here

sensor/kubernetes/fake/service.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func getRandProtocol() string {
+	protocols := []string{"TCP", "UDP", "SCTP"}
+	return protocols[rand.Intn(len(protocols))]
+}
+
+func getRandIP() string {
```
fredrb, line 17: Does this need to be a helper function? Why not call `generateIP` directly?

sensor/kubernetes/fake/service.go
```diff
@@ -0,0 +1,170 @@
...
+	ip := generateIP()
+	return ip
+}
+
+func getRandPort() uint32 {
+	return rand.Uint32() % 63556
+}
+
+func getIPFamily() string {
+	ipFamilies := []string{"IPv4", "IPv6"}
```
fredrb, line 27: Nit: this could be a `const`

sensor/kubernetes/fake/service.go
lvalerom: Fixed

sensor/kubernetes/fake/service.go
lvalerom: Fixed

sensor/kubernetes/fake/service.go
lvalerom: I've changed it to `var ipFamilies = [...]string{"IPv4", "IPv6"}`. I also added `protocols  = [...]string{"TCP", "UDP", "SCTP"}`

sensor/kubernetes/fake/service.go
```diff
@@ -0,0 +1,168 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var (
+	protocols  = [...]string{"TCP", "UDP", "SCTP"}
+	ipFamilies = [...]string{"IPv4", "IPv6"}
```
fredrb, line 14: Nice. I didn't know about this feature (`[...]`)

fredrb (approved):
LGTM from a code perspective.

I have a question, but not sure if it fits the review of this PR: From my understanding of this change, the selector labels are being generated randomly for each service, right? Doesn't it make sense to select from a pool of labels that are being used by deployments? If fake workloads are going to be used to validate that the re-sync worked, then the resources should be related to each other.

lvalerom:
> I have a question, but not sure if it fits the review of this PR: From my understanding of this change, the selector labels are being generated randomly for each service, right? Doesn't it make sense to select from a pool of labels that are being used by deployments? If fake workloads are going to be used to validate that the re-sync worked, then the resources should be related to each other.

If you specify the number of labels in the ConfigMap then the labels are not random, instead they get fixed values using this format `key-%d=value-%d`. Now that you mention it, I think creating a pool of labels from labels used by deployments could make the scenarios more realistic. I'll open a follow up PR to address this.

PR #3344, opened 2022-10-06
ROX-7273 Google SCC Integration Does Not Map Severity Properly

ksurabhi91:
PR fixes scc integration notifier to show severity correctly. The fix was tested by introducing exec policy violation which triggered scc notifier to notify SCC. See screenshot attached

### DIFF
--- a/central/notifiers/cscc/client/client.go
+++ b/central/notifiers/cscc/client/client.go
@@ -43,7 +43,7 @@ type Config struct {
 
 func (c *Config) postURL(findingID string) string {
 	return fmt.Sprintf(
-		"https://securitycenter.googleapis.com/v1beta1/%s/findings?findingId=%s",
+		"https://securitycenter.googleapis.com/v1p1beta1/%s/findings?findingId=%s",
 		c.SourceID,
 		findingID,
 	)
@@ -84,6 +84,7 @@ func (c *Config) request(finding *findings.Finding, id string) (*http.Request, e
 	if err != nil {
 		return nil, errors.Wrap(err, "marshal")
 	}
+
 	c.Logger.Debugf("Request: %s", string(b))
 
 	req, err := http.NewRequest("POST", c.postURL(id), bytes.NewReader(b))
--- a/central/notifiers/cscc/cscc.go
+++ b/central/notifiers/cscc/cscc.go
@@ -32,7 +32,7 @@ var (
 
 func init() {
 	notifiers.Add("cscc", func(notifier *storage.Notifier) (notifiers.Notifier, error) {
-		j, err := newCSCC(notifier)
+		j, err := newCSCC(notifier, clusterDatastore.Singleton())
 		return j, err
 	})
 }
@@ -42,9 +42,9 @@ type cscc struct {
 	// The Service Account is a Google JSON service account key.
 	// The GCP Organization ID is a numeric identifier for the Google Cloud Platform
 	// organization. It is required so that we can tag findings to the right org.
-	client client.Config
-	config *config
-
+	client       client.Config
+	config       *config
+	clusterStore clusterDatastore.DataStore
 	*storage.Notifier
 }
 
@@ -81,18 +81,18 @@ func (c *cscc) getAlertDescription(alert *storage.Alert) string {
 	return strings.Join(distinctSlice, " ")
 }
 
-func transformSeverity(s storage.Severity) string {
+func transformSeverity(s storage.Severity) findings.Severity {
 	switch s {
 	case storage.Severity_LOW_SEVERITY:
-		return "low"
+		return findings.Low
 	case storage.Severity_MEDIUM_SEVERITY:
-		return "medium"
+		return findings.Medium
 	case storage.Severity_HIGH_SEVERITY:
-		return "high"
+		return findings.High
 	case storage.Severity_CRITICAL_SEVERITY:
-		return "critical"
+		return findings.Critical
 	default:
-		return "info"
+		return findings.Default
 	}
 }
 
@@ -132,7 +132,7 @@ func processUUID(u string) string {
 }
 
 func (c *cscc) getCluster(id string) (*storage.Cluster, error) {
-	cluster, exists, err := clusterDatastore.Singleton().GetCluster(clusterForAlertContext, id)
+	cluster, exists, err := c.clusterStore.GetCluster(clusterForAlertContext, id)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


theencee (commented):
Change looks good. See if you can add a test and then let's pair test manually on Monday

central/notifiers/cscc/client/client.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/cloudflare/cfssl/log"
```
theencee, line 11: This is the wrong import. You need `"github.com/stackrox/rox/pkg/logging"`

theencee: Actually if you do my suggestion below you don't need to do this import at all

central/notifiers/cscc/client/client.go
```diff
@@ -42,6 +43,7 @@ type Config struct {
 }
 
 func (c *Config) postURL(findingID string) string {
+	log.Infof("ROX-7273 finding:%s source:%s", findingID, c.SourceID)
```
theencee, line 46: the log object is not initialized so this will error. But good thing is that we have a reference to it within the config object.

So change to `c.Logger.Infof`

ksurabhi91: ok sure let me change that

keyallis:
Those 2 failing tests are going to be unrelated to your changes I believe. Can be fixed with a rebase on master

central/notifiers/cscc/client/client.go
```diff
@@ -84,7 +84,6 @@ func (c *Config) request(finding *findings.Finding, id string) (*http.Request, e
 	if err != nil {
 		return nil, errors.Wrap(err, "marshal")
 	}
-	c.Logger.Debugf("Request: %s", string(b))
```
theencee, line 87: Keep this. It's useful

theencee (commented):
lgtm! Please add the unit test

central/notifiers/cscc/client/client.go
ksurabhi91: done

central/notifiers/cscc/cscc.go
```diff
@@ -158,6 +158,9 @@ func (c *cscc) Close(ctx context.Context) error {
 
 // AlertNotify takes in an alert and generates the notification
 func (c *cscc) AlertNotify(ctx context.Context, alert *storage.Alert) error {
+
+	_, _, _ = c.InitFinding(ctx, alert)
```
charmik-redhat, line 162: Use the findingID and finding returned by InitFinding here, and you can remove the code below until
`return retry.WithRetry...` as it is same as InitFinding

```
findingID, finding, err := InitFinding(ctx, alert)
  if err != nil {
      return nil, err
  }
  return retry.WithRetry(...)
```

central/notifiers/cscc/cscc.go
```diff
@@ -232,7 +234,72 @@ func (c *cscc) AlertNotify(ctx context.Context, alert *storage.Alert) error {
 	)
 }
 
-func newCSCC(protoNotifier *storage.Notifier) (*cscc, error) {
+// AlertNotify takes in an alert and generates the notification
```
charmik-redhat, line 237: Change comment to reflect `InitFinding` purpose

central/notifiers/cscc/cscc.go
```diff
@@ -232,7 +234,72 @@ func (c *cscc) AlertNotify(ctx context.Context, alert *storage.Alert) error {
...
+func (c *cscc) InitFinding(ctx context.Context, alert *storage.Alert) (string, *findings.Finding, error) {
```
charmik-redhat, line 238: Change to `initFinding` with small I in the beginning. No need to export this func because is used only in this package. The test will be able to access it because it is in the same package.

central/notifiers/cscc/cscc_test.go
```diff
@@ -0,0 +1,87 @@
...
+	cluster := &storage.Cluster{
+		Id:   "test_id",
+		Name: "test_name",
+		Status: &storage.ClusterStatus{
+			ProviderMetadata: &storage.ProviderMetadata{
+				Zone: "test_zone",
+				Provider: &storage.ProviderMetadata_Google{
+					Google: &storage.GoogleProviderMetadata{
+						Project:     "test_project",
+						ClusterName: "test_cluster",
```
charmik-redhat, line 45: nit : Although probably it doesn't matter for this test, you can keep the ClusterName here same as the cluster.Name set above, just for consistency.

central/notifiers/cscc/cscc_test.go
```diff
@@ -0,0 +1,87 @@
...
+			Id:        "myDeploymentID",
+			ClusterId: "test_id",
+		}},
+		Time: types.TimestampNow(),
+	}
+	findingID := ""
+	var finding *findings.Finding
+	var err error
+	findingID, finding, err = scc.InitFinding(context.Background(), testAlert)
+	assert.Equal(t, nil, err)
```
charmik-redhat, line 80: Use `assert.NoError` instead for error checking.

central/notifiers/cscc/cscc_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWithFakeCSCC(t *testing.T) {
+
+	const (
+		projectKey = "FJ"
+	)
+
+	sourceID := "organizations/847401270788/sources/394194234514334994"
```
charmik-redhat, line 21: Are these random numbers in the sourceID or are they for an actual source?

charmik-redhat (changes requested):

central/notifiers/cscc/cscc_test.go
theencee: Please use fake data

central/notifiers/cscc/cscc.go
ksurabhi91: done

central/notifiers/cscc/cscc.go
ksurabhi91: done

central/notifiers/cscc/cscc.go
ksurabhi91: done

central/notifiers/cscc/cscc_test.go
ksurabhi91: done

central/notifiers/cscc/cscc_test.go
ksurabhi91: done

central/notifiers/cscc/cscc.go
```diff
@@ -158,8 +158,34 @@ func (c *cscc) Close(ctx context.Context) error {
 
 // AlertNotify takes in an alert and generates the notification
 func (c *cscc) AlertNotify(ctx context.Context, alert *storage.Alert) error {
+
+	var finding *findings.Finding
+	var findingID string
+	var err error
+
+	findingID, finding, err = c.initFinding(ctx, alert)
```
charmik-redhat, line 166: Change to `findingID, finding, err := c.initFinding(ctx, alert)`.  If you use `:=` you don't need to declare the variables above.

charmik-redhat (approved):
LGTM. Just address the new comment in AlertNotify and then merge.

central/notifiers/cscc/cscc.go
```diff
@@ -169,7 +191,7 @@ func (c *cscc) AlertNotify(ctx context.Context, alert *storage.Alert) error {
 
 	cluster, err := c.getCluster(alert.GetDeployment().GetClusterId())
 	if err != nil {
-		return err
+		return findingID, nil, err
```
theencee, line 194: [nit] In this case I would still return empty string because if err != nil, then it's not meant to be used. Doesn't make a functional difference though.

central/notifiers/cscc/cscc_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/golang/mock/gomock"
+	clusterMocks "github.com/stackrox/rox/central/cluster/datastore/mocks"
+	"github.com/stackrox/rox/central/notifiers/cscc/findings"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWithFakeCSCC(t *testing.T) {
+
+	const (
```
theencee, line 17: [nit] Either just make it a var and keep it in the func or move the const outside of the function

theencee (approved):
Thanks for adding the test!

PR #3346, opened 2022-10-06
Move version store to gorm on migrator

c-du:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


migrator/version/version.go
```diff
@@ -0,0 +1,93 @@
...
+		utils.Must(errors.Wrapf(err, "failed to connect to database %s", dbName))
+	}
+	defer migGorm.Close(db)
+	SetVersionGormDB(db, updatedVersion)
+}
+
+// SetVersionGormDB - sets the version in the postgres database specified with the Gorm instance
+func SetVersionGormDB(db *gorm.DB, updatedVersion *storage.Version) {
+	ctx := sac.WithAllAccess(context.Background())
+	pkgSchema.ApplySchemaForTable(ctx, db, pkgSchema.VersionsSchema.Table)
```
dashrews78, line 71: What is the cost of this operation when the schema has already been applied?  Reason I ask is that this will get called for each migration as we loop through and process them.

c-du: It should not be much but it is always good to avoid that. Thanks.

migrator/clone/postgres/db_clone_manager_impl.go
```diff
@@ -42,6 +43,7 @@ func New(forceVersion string, adminConfig *pgxpool.Config, sourceMap map[string]
 // from disk.
 func (d *dbCloneManagerImpl) Scan() error {
 	clones := pgadmin.GetDatabaseClones(d.adminConfig)
+	ctx := sac.WithAllAccess(context.Background())
```
dashrews78, line 46: nit:  Would it make sense to make the context a var or add it to the struct?

dashrews78 (approved):
LGTM.  just one nit.

PR #3361, opened 2022-10-10
rox 13014 mock grpc server should record process info of endpoints

JoukoVirtanen:
## Description

For the processes listening on ports MVP we need to add process information to endpoints. We need integration tests for that in collector and we need the mock grpc server to record the process information that the endpoint messages have.

## Checklist
- [x] Used new grpc server in collector integration tests.

## Testing Performed

See above


erthalion:
Thanks, looks reasonable. The only question I have is whether it's going to look readable? Other fields in this output are single values, where the originator is going to be a combination of three values (exe, args, path) somehow put together.

erthalion (approved):

JoukoVirtanen:
> Thanks, looks reasonable. The only question I have is whether it's going to look readable? Other fields in this output are single values, where the originator is going to be a combination of three values (exe, args, path) somehow put together.

I have another PR which reads the output. You can find it here https://github.com/stackrox/collector/pull/870/files

I think at some point we might store these things in json.

PR #3366, opened 2022-10-10
ROX-12875: Prefix SAC indexes with the target table name

rhybrillou:
## Description

This change builds upon the work from PR https://github.com/stackrox/stackrox/pull/3238 .
Looking at the deployed database, it seems not all SAC indexes were created:
```
central_active=# \d sac_filter
             Index "public.sac_filter"
  Column   |       Type        | Key? | Definition 
-----------+-------------------+------+------------
 name      | character varying | yes  | name
 clusterid | character varying | yes  | clusterid
btree, for table "public.namespaces"

central_active=# \d process_baseline_results 
              Table "public.process_baseline_results"
    Column    |       Type        | Collation | Nullable | Default 
--------------+-------------------+-----------+----------+---------
 deploymentid | character varying |           | not null | 
 clusterid    | character varying |           |          | 
 namespace    | character varying |           |          | 
 serialized   | bytea             |           |          | 
Indexes:
    "process_baseline_results_pkey" PRIMARY KEY, btree (deploymentid)

central_active=#
```

Trying to create the SAC indexes manually ends in error:
```
central_active=# create index sac_filter on alerts using btree (clusterid, namespace);
ERROR:  relation "sac_filter" already exists
central_active=#
```

The goal of the change is to ensure unique SAC index names so that all tables holding scoping data can be properly indexed.

## Checklist

## Testing Performed

CI should be sufficient for the functional side.
Additional test is to deploy the database and to check that SAC indexes exist on all tables "listed" in the changed files.


janisz (approved):
Please remove `pkg/postgres/transactioncontext.go`

PR #3348, opened 2022-10-07
Implement basic markup structure of collection form

dvail:
## Description

Adds the main layout of the collection form page, with empty components stubbed in where they will be located once implemented.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] ~~Unit test and regression tests added~~
- [ ] ~~Evaluated and added CHANGELOG entry if required~~
- [ ] ~~Determined and documented upgrade steps~~
- [ ] ~~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed

Visit the main collections page and click the "Create collection" button. If on a large screen, the preview drawer should be open by default. If on a small screen, the preview drawer should be closed.
image
image

Test that resizing the browser window beyond the "large" threshold sets the preview drawer state to open (when the window is large) and closed (when the window is small).

Test that clicking the "Hide preview results" button closes the drawer and updates the button text. Test that clicking the button again re-opens the drawer.
image

Test the the drawer open state in the small screen view results in an "overlay" drawer, as opposed to the drawer the pushes the content to the left in the large screen view.
image

Test that the top level "Collections" link in the breadcrumbs takes you back to the main collections table page.
image

Test the visiting a page in "Create", "Clone", or "Edit" mode removes the action menu and places a "Save" and "Cancel" button in the page footer.
image

PR #3372, opened 2022-10-10
Implement core state and data flow for collections form

dvail:
## Description

This implements the data structures and related types that are needed for the main collections form.

There is a mismatch between what is a valid collection representation on the server, and the actions that the UI can support. In cases where valid data is returned from the server that cannot be handled in the UI (such as collections created directly from the CLI), we will display an error screen letting the user know that the collection cannot be viewed/edited. 

The specific mismatches are:
- The back end allows multiple top level `resourceSelector` objects in the collection response. These are logically 'AND'ed together, but the UI only supports a single `resourceSelector` object.
- The back end supports both "AND" and "OR" operators for rules attached to a `resourceSelector`. The current UI controls will only support "OR" for both name and label field rules.
- The back end can support rules for multiple fields that are related to a single entity. For example, you could apply a rule that selected Namespaces based on both "Namespace (name)" and "Namespace Label" via the API, but the UI only supports name _xor_ label when selecting Namespaces.

## Checklist

## Testing Performed

The only testable change is to test the nothing renders on the main form page when the request is in an "error" or "loading" state, and instead a white page is rendered. The rest of the changes are type-level and do not functionally change the UI at this point.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useReducer, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
     Alert,
@@ -34,18 +34,55 @@ import ConfirmationModal from 'Components/PatternFly/ConfirmationModal';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import useToasts, { Toast } from 'hooks/patternfly/useToasts';
 import { collectionsBasePath } from 'routePaths';
-import { deleteCollection, ResolvedCollectionResponse } from 'services/CollectionsService';
+import { deleteCollection } from 'services/CollectionsService';
 import { CollectionPageAction } from './collections.utils';
 import RuleSelector from './RuleSelector';
 import CollectionAttacher from './CollectionAttacher';
 import CollectionResults from './CollectionResults';
+import { Collection, ScopedResourceSelector, SelectorEntityType } from './types';
+
+type FormStateReducerAction =
+    | { type: 'setName'; name: string }
+    | { type: 'setDescription'; description: string }
+    | { type: 'setRules'; entity: SelectorEntityType; selector: ScopedResourceSelector }
+    | { type: 'attachCollection'; collectionId: string }
+    | { type: 'detachCollection'; collectionId: string };
+
+function formStateReducer(state: Collection, payload: FormStateReducerAction): Collection {
+    switch (payload.type) {
+        case 'setName':
+            return { ...state, name: payload.name };
+        case 'setDescription':
+            return { ...state, name: payload.description };
+        case 'setRules': {
+            const selectorRules = { ...state.selectorRules };
+            selectorRules[payload.entity] = payload.selector;
+            return { ...state, selectorRules };
+        }
+        case 'attachCollection':
+            return {
+                ...state,
+                embeddedCollectionIds: state.embeddedCollectionIds.concat(payload.collectionId),
+            };
+        case 'detachCollection':
+            return {
+                ...state,
+                embeddedCollectionIds: state.embeddedCollectionIds.filter(
+                    (id) => id !== payload.collectionId
+                ),
+            };
+        default:
+            // Type safe fallback to ensure we don't miss any cases
+            return ((_: never) => _)(payload);
+    }
+}
 
 export type CollectionFormProps = {
     hasWriteAccessForCollections: boolean;
     /* The user's workflow action for this collection */
     action: CollectionPageAction;
-    /* initial data used to populate the form, or `undefined` in the case of a new collection */
-    initialData: ResolvedCollectionResponse | undefined;
+    /* initial data used to populate the form */
+    initialData: Collection;
     /* Whether or not to display the collection results in an inline drawer. If false, will
     display collection results in an overlay drawer. */
     useInlineDrawer: boolean;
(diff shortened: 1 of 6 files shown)


dvail:
* master
  * **PR #3348** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3348" target="_blank">Graphite</a> 
    * **PR #3372** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3372" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3372?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -23,17 +23,53 @@ import {
...
+import { Collection, ScopedResourceSelector, SelectorEntityType } from './types';
+
+type FormStateReducerAction =
+    | { type: 'setName'; name: string }
+    | { type: 'setDescription'; description: string }
+    | { type: 'setRules'; entity: SelectorEntityType; selector: ScopedResourceSelector }
+    | { type: 'attachCollection'; collectionId: string }
+    | { type: 'detachCollection'; collectionId: string };
+
+function formStateReducer(state: Collection, payload: FormStateReducerAction): Collection {
```
dvail, line 39: This felt like a good place for a unit test, but I figured I would keep this un-exported since all of these use cases would be well tested by e2e tests once we get to that point.

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -81,6 +70,87 @@ export type ResolvedCollectionResponse = {
...
+        if (
+            !hasMultipleFieldsForEntity &&
+            !isRuleForAnnotationField &&
+            !isUnsupportedRuleOperator
+        ) {
+            collection.selectorRules[entity].rules.push(rule);
+        }
+    });
+
+    return errors.length > 0 ? new AggregateError(errors) : collection;
```
dvail, line 151: If it isn't useful to have the full list of validation errors, I can simplify this code to a generic single error and remove the `AggregateError` return type.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -23,17 +23,53 @@ import {
...
+        case 'detachCollection':
+            return {
+                ...state,
+                embeddedCollectionIds: state.embeddedCollectionIds.filter(
+                    (id) => id !== payload.collectionId
+                ),
+            };
+        default:
+            // Type safe fallback to ensure we don't miss any cases
+            return ((_: never) => _)(payload);
```
dvail, line 64: This might be useful to extract to a utility, but where to put it? And what to name it? `exhaustive()`, `unreachable()`, `cannotBeCalled()`... ?

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -34,18 +34,55 @@ import ConfirmationModal from 'Components/PatternFly/ConfirmationModal';
...
+}
 
 export type CollectionFormProps = {
     hasWriteAccessForCollections: boolean;
     /* The user's workflow action for this collection */
     action: CollectionPageAction;
-    /* initial data used to populate the form, or `undefined` in the case of a new collection */
-    initialData: ResolvedCollectionResponse | undefined;
+    /* initial data used to populate the form */
+    initialData: Collection;
```
pedrottimark, line 85: Oh, super! It becomes much easier to read property access below.

ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
```diff
@@ -22,27 +34,42 @@ function CollectionsFormPage({
...
+                action={pageAction}
+                initialData={data}
+                useInlineDrawer={isLargeScreen}
+                showBreadcrumbs
+                appendTableLinkAction={() => {
+                    /* TODO */
+                }}
+            />
+        );
+    }
```
pedrottimark, line 67: Although lint rules encourage independent `if` statements for `return` statement, what are pro and con of `if else if else if` structure for assignment, communicating to reader and TypeScript that cases are mutually exclusive? That is, if they are.

ui/apps/platform/src/services/CollectionsService.ts
pedrottimark: Although I will let this simmer in my slow cooker today, initial impression is tight coupling for service to import types from container and for service to significantly transform response from request.

pedrottimark: What would it look like if:
* service type and return value corresponds to backend behavior (and preview in network panel)
* container has limited types and pure util functions to transform data and describe errors

ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
dvail: Good call, they are not mutually exclusive, but this pattern would be a more accurate structure:
```typescript
if (error) {
} else if (data && !loading) {
} else if (loading) {
}
```

I'll update.

ui/apps/platform/src/services/CollectionsService.ts
dvail: I went back and forth on this, and originally had the transform functions right in the component file but changed it yesterday.

I'm OK with moving the transform functions and types out of the service and back into the container. I can see there being a use case where we want to get the full server response via the service function and use the result someplace else, someplace that might support the full structure beyond MVP.

pedrottimark: For example, here is a surprise. A teammate from long ago called it asymetritis.
* policies container has transformation functions from response to form state and vice versa
* frontend policies service function for POST and PUT requests also call a util function

PR #3375, opened 2022-10-10
Remove goveralls, as we no longer use it

misberner:
## Description

Title says it all.

## Checklist

## Testing Performed

- CI


janisz (approved):

janisz:
Maybe we should restart using it?

misberner:
> Maybe we should restart using it?

I think that's a discussion to be had independently. Maybe we should restart doing some coverage metrics collection. Maybe we should use goveralls for that. But we should not do either just because the dependency happens to lie around unused in the `Makefile`. Therefore I would proceed with this PR.

misberner:
Postgres test failure is not related.

PR #3379, opened 2022-10-11
ROX-13035 protobufs for processes listening on ports api

JoukoVirtanen:
## Description

Protobufs to define the processes listening on ports API.

## Checklist

## Testing Performed

See above


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   - The permission `Compliance` will deprecate the permission `ComplianceRuns`.
 
 ### Technical Changes
+- ROX-11268: An API to report which processes are listening to which ports has been added. The API is at `/v1/processes_listening_on_ports`
```
md2119, line 45: I don't believe this is a candidate for changelog

proto/api/v1/process_listening_on_port_service.proto
```diff
@@ -0,0 +1,47 @@
...
+
+message GetProcessesListeningOnPortsResponse {
+    repeated storage.ProcessListeningOnPort processes_listening_on_ports = 1;
+}
+
+message GetProcessesListeningOnPortsWithDeploymentResponse {
+    repeated ProcessListeningOnPortWithDeploymentId processes_listening_on_ports_with_deployment = 1;
+}
+
+service ProcessListeningOnPortService {
```
md2119, line 33: What is the reason to separate this from process service?

md2119 (commented):
I am not familiar with the design. Maybe it is my oversight if it was published to all engineering. Are process listening on ports stored separately? And why?  Did we consider **listening on port** as a search option?

JoukoVirtanen:
> I am not familiar with the design. Maybe it is my oversight if it was published to all engineering. Are process listening on ports stored separately? And why? Did we consider **listening on port** as a search option?

Here is the design document 

https://docs.google.com/document/d/1BHpM_W1Ty9n-kQqaMX5UGBYR9Dsc2m_srmC2nLb56ag/edit

This was mostly a customer consideration. This is also easier to do from an engineering perspective.

CHANGELOG.md
JoukoVirtanen: Okay. I can remove it. I thought that new APIs should be documented in the changelog. Is the authomatically genrated swagger document enough?

proto/api/v1/process_listening_on_port_service.proto
JoukoVirtanen: The endpoint information will be added to the process service at a future time. To best meet the criteria of the customer request in the MVP we decided to have a separate service.

CHANGELOG.md
md2119: New features are announced through release notes. Changelog is meant to point out breaking changes to existing functionality.

md2119 (commented):
Since I was not involved in the initial design discussion, I will exclude myself from approving this PR. That said, I have a couple of comments:
- The two different representation of processes in the product seems unnecessary. 
- Even if there is a strong inclination towards separating process and listening process services, I find no reason not to provide `storage.ProcessSignal` format through new APIs. The `storage.ProcessListeningOnPort` seems to be missing quite a few fields compared to `storage.ProcessSignal`. I am afraid user will have to deal with 2 endpoints. 
- The `storage.ProcessSignal` can be extended as follows:
  ```
    oneof ProcessProps {
       ListeningProcessProps listening_process_props;
    }
   ```

ovalenti:
@md2119, here are some elements that I believe made us go this way so far:

> * The two different representation of processes in the product seems unnecessary.

ProcessSignal represents a process spawn event, and the focus here is to link the endpoint to the process creating it; which is related, but not absolutely the same.

> * Even if there is a strong inclination towards separating process and listening process services, I find no reason not to provide `storage.ProcessSignal` format through new APIs. The `storage.ProcessListeningOnPort` seems to be missing quite a few fields compared to `storage.ProcessSignal`. I am afraid user will have to deal with 2 endpoints.

A slight difference could be that we report processes at the container level when ProcessSignal focuses on deployment level.

md2119:
@ovalenti,

> @md2119, here are some elements that I believe made us go this way so far:
> 
> > * The two different representation of processes in the product seems unnecessary.
> 
> ProcessSignal represents a process spawn event, and the focus here is to link the endpoint to the process creating it; which is related, but not absolutely the same.

The information captured for processes never tracked the life of the process other the spawn event. In hindsight, it seems sufficient for the use cases we wanted to cater to. That does not mean that a process cannot be enriched further. The major reason behind not tracking the life of every process was the lack of a strong use case.

> > * Even if there is a strong inclination towards separating process and listening process services, I find no reason not to provide `storage.ProcessSignal` format through new APIs. The `storage.ProcessListeningOnPort` seems to be missing quite a few fields compared to `storage.ProcessSignal`. I am afraid user will have to deal with 2 endpoints.
> 
> A slight difference could be that we report processes at the container level when ProcessSignal focuses on the deployment level.

`ProcessSignal` which is a field of `ProcessIndicator` are reported at the container level. We use this information in violations and process timeline.

```
message ProcessSignal {
    // A unique UUID for identifying the message
    // We have this here instead of at the top level
    // because we want to have each message to be
    // self contained.
    string id                         = 1;

    // ID of container associated with this process
    string container_id               = 2 [(gogoproto.moretags) = 'search:"Container ID,hidden"'];
...
}
```

JoukoVirtanen:
> @ovalenti,
> 
> > @md2119, here are some elements that I believe made us go this way so far:
> > > * The two different representation of processes in the product seems unnecessary.
> > 
> > 
> > ProcessSignal represents a process spawn event, and the focus here is to link the endpoint to the process creating it; which is related, but not absolutely the same.
> 
> The information captured for processes never tracked the life of the process other the spawn event. In hindsight, it seems sufficient for the use cases we wanted to cater to. That does not mean that a process cannot be enriched further. The major reason behind not tracking the life of every process was the lack of a strong use case.
> 
> > > * Even if there is a strong inclination towards separating process and listening process services, I find no reason not to provide `storage.ProcessSignal` format through new APIs. The `storage.ProcessListeningOnPort` seems to be missing quite a few fields compared to `storage.ProcessSignal`. I am afraid user will have to deal with 2 endpoints.
> > 
> > 
> > A slight difference could be that we report processes at the container level when ProcessSignal focuses on the deployment level.
> 
> `ProcessSignal` which is a field of `ProcessIndicator` are reported at the container level. We use this information in violations and process timeline.
> 
> ```
> message ProcessSignal {
>     // A unique UUID for identifying the message
>     // We have this here instead of at the top level
>     // because we want to have each message to be
>     // self contained.
>     string id                         = 1;
> 
>     // ID of container associated with this process
>     string container_id               = 2 [(gogoproto.moretags) = 'search:"Container ID,hidden"'];
> ...
> }
> ```

We are considering using ProcessSignal instead of ProcessIndicatorUniqueKey in the ProcessListeningOnPort protobuf definition pending discussion and further reviews.

JoukoVirtanen:
I have created a PR here that enriches the endpoints with ProcessSignal. That PR can be found here https://github.com/stackrox/stackrox/pull/3948

I have built it on top of this PR, but kept it separate, as it is just a proof of concept.

There is another PR by @ovalenti which also enriches the endpoints with more process information. I am not sure if that PR is ready to be reviewed. https://github.com/stackrox/stackrox/pull/3943

I disagree with his approach as he uses the NetworkEndpoint from network_flow for the source of processes listening on ports, and NetworkEndpoint is on the deployment level and processes listening on ports should be at the container level.

PR #3381, opened 2022-10-11
ROX-12813: Give the metrics server more time to become available after scaling up

misberner:
## Description

The ticket describes a test flake. This should be fixable through a more generous timeout.

## Checklist

## Testing Performed

- CI


misberner:
Actually, I realized that GKE upgrade tests are deterministically broken anyway, so this has to be folded into the other PR.

PR #3383, opened 2022-10-11
Dashrews/ROX-12999 Central should not crash for central-db restarts (WIP)

dashrews78:
## Description

When restarting central-db, central would crash.  It would come back up and connect fine, but the crash was unnecessary.  The Postgres retry logic was handling everything just fine, but there were locks at the datastore level that were expiring and thus causing central to crash.  In a release build central would have just held the lock and not crashed but since we have timeout locks in dev builds to help us make sure our code quality is good, we were able to see these crashes and address them.  The issue is that we were acquiring locks in some of the datastores to ensure we updated the store and the indexer together.  In Postgres we no longer need to separately update an indexer and as such the locks I've encountered have been unnecessary, so the code was updated not to use those locks in Postgres mode as they are no longer necessary.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/cluster/datastore/datastore_impl.go
+++ b/central/cluster/datastore/datastore_impl.go
@@ -403,6 +403,13 @@ func (ds *datastoreImpl) UpdateClusterHealth(ctx context.Context, id string, clu
 		return err
 	}
 
+	// Postgres does not need to update indexes and things of that nature and thus does not need to obtain a lock
+	// to perform a simple database upsert.
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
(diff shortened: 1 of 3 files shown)


dashrews78:
Closing as this is only an issue on dev builds when we test bouncing because locks timeout due to mutex_dev having a timeout.  This will require a lot of thought looking at the mutex locks at the datastore level to see if we want to release those as we retry or if we need to keep them to maintain data integrity.  Best to evaluate this as we begin to remove the RocksDB code over the course of the next few releases.

PR #3387, opened 2022-10-11
ROX-12962 PLoP sensor implementation

Molter73:
## Description

This PR adds the necessary changes to sensor in order for it to properly handle information regarding a process listening on a given port.

The basic idea is that collector will be sending some additional information on the processes that is listening on a given port in a deployment, sensor will then enrich that information with the pod id and container name and send it over to central. A new endpoint for central will be implemented in another PR, so the new data sent by collector will be stored next to the received endpoints at first and split into their own stream when sent to central.

A feature flag has been put in place to try and minimize the impact of these changes while we implement the missing features in collector and central.

## TODO
- [x] Get rid of some of the additional logging that has been added once we are done testing manually.

## Checklist

## Testing Performed

Manually checked that sending mock data into sensor pretending to be collector resulted in the processes being forwarded to central, but more testing will be needed.

### DIFF
--- a/sensor/common/networkflow/manager/manager_impl.go
+++ b/sensor/common/networkflow/manager/manager_impl.go
@@ -307,7 +307,7 @@ func (m *networkFlowManager) enrichAndSendProcesses() {
 		Time:                      types.TimestampNow(),
 	}
 
-	//metrics.IncrementTotalNetworkFlowsSentCounter(len(flowsToSend.Updated))
+	metrics.IncrementTotalProcessesSentCounter(len(processesToSend.ProcessesListeningOnPorts))
 	log.Debugf("Processes update : %v", processesToSend)
 	select {
 	case <-m.done.Done():
@@ -469,7 +469,7 @@ func (m *networkFlowManager) enrichProcessListening(ep *containerEndpoint, statu
 		if timeElapsedSinceFirstSeen > maxContainerResolutionWaitPeriod {
 			status.rotten = true
 			// Only increment metric once the connection is marked rotten
-			//flowMetrics.ContainerIDMisses.Inc()
+			flowMetrics.ContainerIDMisses.Inc()
 			log.Debugf("Unable to fetch deployment information for container %s: no deployment found", ep.containerID)
 		}
 		return
@@ -489,7 +489,6 @@ func (m *networkFlowManager) enrichProcessListening(ep *containerEndpoint, statu
 
 	log.Debugf("Enriched: %s", indicator)
 
-	// TODO: We can probably remove this check for processes
 	// Multiple endpoints from a collector can result in a single enriched endpoint,
 	// hence update the timestamp only if we have a more recent endpoint than the one we have already enriched.
 	if oldTS, found := processesListening[indicator]; !found || oldTS < status.lastSeen {
@@ -544,7 +543,7 @@ func (m *networkFlowManager) enrichProcessesListening(hostConns *hostConnections
 			delete(hostConns.endpoints, ep)
 		}
 	}
-	flowMetrics.HostEndpointsRemoved.Add(float64(prevSize - len(hostConns.endpoints)))
+	flowMetrics.HostProcessesRemoved.Add(float64(prevSize - len(hostConns.endpoints)))
 }
 
 func (m *networkFlowManager) currentEnrichedConnsAndEndpoints() (map[networkConnIndicator]timestamp.MicroTS, map[containerEndpointIndicator]timestamp.MicroTS) {
@@ -891,8 +890,6 @@ func getUpdatedContainerEndpoints(hostname string, networkInfo *sensor.NetworkCo
 			ts = timestamp.InfiniteFuture
 		}
 		updatedEndpoints[ep] = ts
-
-		log.Debugf("Received endpoint: %s", ep)
 	}
 
 	return updatedEndpoints
--- a/sensor/common/networkflow/manager/manager_impl.go
+++ b/sensor/common/networkflow/manager/manager_impl.go
@@ -166,7 +166,7 @@ func (p *processInfo) String() string {
 type containerEndpoint struct {
 	endpoint    net.NumericEndpoint
 	containerID string
-	processKey  processInfo
+	processKey  *processInfo
 }
 
 func (e *containerEndpoint) String() string {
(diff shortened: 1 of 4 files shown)


sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -144,7 +184,7 @@ func NewManager(
 		done:              concurrency.NewSignal(),
 		connectionsByHost: make(map[string]*hostConnections),
 		clusterEntities:   clusterEntities,
-		flowUpdates:       make(chan *central.MsgFromSensor),
```
JoukoVirtanen, line 147: Why is this being removed? There should be no changes to the network flow updates.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -190,18 +231,23 @@ func (m *networkFlowManager) Capabilities() []centralsensor.SensorCapability {
 }
 
 func (m *networkFlowManager) ResponsesC() <-chan *central.MsgFromSensor {
-	return m.flowUpdates
+	return m.sensorUpdates
 }
 
 func (m *networkFlowManager) enrichConnections() {
```
JoukoVirtanen, line 237: Shouldn't we have a new enrichEndpoints function instead of using enrichConnections()?

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -245,6 +291,36 @@ func (m *networkFlowManager) enrichAndSend() {
...
+		return
+	}
+
+	processesToSend := &central.ProcessListeningOnPortsUpdate{
+		ProcessesListeningOnPorts: updatedProcesses,
+		Time:                      types.TimestampNow(),
+	}
+
+	metrics.IncrementTotalProcessesSentCounter(len(processesToSend.ProcessesListeningOnPorts))
+	log.Debugf("Processes update : %v", processesToSend)
```
fredrb, line 311: Do you have an example of how does this look like in sensor debug logs? Using `%v` will print all struct's values. How useful do you think that will be for someone debugging an issue here? If you think _all_ properties should be shown, perhaps use `%+v`, which also prints the property names. This is specially good if you have string values in the struct: [example](https://go.dev/play/p/VjRjn468SeW).

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -410,6 +524,26 @@ func (m *networkFlowManager) enrichHostContainerEndpoints(hostConns *hostConnect
 	flowMetrics.HostEndpointsRemoved.Add(float64(prevSize - len(hostConns.endpoints)))
 }
 
+func (m *networkFlowManager) enrichProcessesListening(hostConns *hostConnections, processesListening map[processListeningIndicator]timestamp.MicroTS) {
+	hostConns.mutex.Lock()
```
fredrb, line 528: I wonder if this could be done with a `RWMutex` to avoid contention since you're not changing any values in the `hostConnections` object.

sensor/common/networkflow/manager/manager_impl.go
fredrb: Just notice you mentioned in the readme ` Get rid of some of the additional logging that has been added once we are done testing manually.`. 

I assume you were referencing this log statement?

fredrb (approved):
Looks good from a code perspective. I didn't test anything. Too bad we don't have any unit tests for this manager. 😞 

Comments are some suggestions. But I see you followed the same pattern that was there for the other metrics, so you should be good.

sensor/common/networkflow/manager/manager_impl.go
Molter73: It's not moved, I'm using the same channel to send the PLoP messages alongside the NetwrokFlow ones, so I simply renamed the channel in order to make it clear that this is no longer a flow exclusive channel, does that make sense?

sensor/common/networkflow/manager/manager_impl.go
Molter73: I don't really see a need to create a completely separate goroutine to send the data since:
- We are using the same data available in the existing endpoints.
- We are using the same channel to send messages.

The only real difference is in how we process the endpoint data to send over to central.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -190,18 +231,23 @@ func (m *networkFlowManager) Capabilities() []centralsensor.SensorCapability {
...
 	ticker := time.NewTicker(time.Second * 30)
+	processTicker := time.NewTicker(time.Second * 30)
 
 	for {
 		select {
 		case <-m.done.WaitC():
 			return
 		case <-ticker.C:
 			m.enrichAndSend()
+		case <-processTicker.C:
```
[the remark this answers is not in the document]
Molter73: You are right, I think my inexperience with go got the best of me here, I'll unify them both under the same ticker.

sensor/common/networkflow/manager/manager_impl.go
Molter73: Yeah, I need to go back through the entire PR and tidy up logging, this being one of the messages.

sensor/common/networkflow/manager/manager_impl.go
Molter73: I honestly just copy pasted it from the existing `enrichContainerEndpoints`, my understanding is the only 2 threads working on this are the one receiving messages from collector and this one processing them and pushing them up the chain. If that's the case, I don't think we would benefit much from using a `RWMutex` since whichever one grabs it will lock out the other one anyways.

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Okay

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Okay

sensor/common/networkflow/manager/manager_impl.go
fredrb: Yes. But from what I've seen, neither of them change the object under lock. Which means that if you used a `RWMutex` they won't lock each other for read-only operations. Maybe I'm missing something obvious?

You don't need to do anything here. I can investigate this further. This was more of a highlighting a point of improvement for us.

Molter73: It's not obvious but AFAICT, the three enrich functions here will run sequentially and the only place this mutex is used seems to be in `hostConnections.Process` which gets called from a separate routine here: https://github.com/stackrox/stackrox/blob/dafad863cffd05d21643e8ef22a498835be944e6/sensor/common/networkflow/service/service_impl.go#L137

So we should be back to no real gain using a `RWLock` until we decide to move one of the enrich functions to a separate routine.

PR #3391, opened 2022-10-11
ROX-13065: Adding skeleton of DeploymentDetails component

sachaudh:
## Description

This PR adds a DeploymentSideBar and DeploymentDetails component that you can see by clicking on a node. This is just a skeleton and the remaining work was divided into subtasks you can view here: https://issues.redhat.com/browse/ROX-12903#view-subtasks. 

## Checklist

## Testing Performed

Screen Shot 2022-10-11 at 3 04 15 PM
Screen Shot 2022-10-11 at 3 04 18 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
 import {
+    EdgeModel,
     Model,
+    NodeModel,
     SELECTION_EVENT,
     TopologySideBar,
     TopologyView,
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -99,17 +107,25 @@ const TopologyComponent = () => {
...
-                <TopologySideBar show={!!selectedId} resizable onClose={() => setSelectedId(null)}>
-                    <div style={{ height: '100%' }}>{selectedId}</div>
+                <TopologySideBar
+                    show={!!selectedObj}
+                    resizable
+                    onClose={() => setSelectedObj(null)}
+                >
+                    {selectedObj?.type === 'group' && <div>Group</div>}
+                    {selectedObj?.type === 'node' && <DeploymentSideBar />}
+                    {selectedObj?.type === 'edge' && <div>Edge</div>}
```
alwayshooin, line 122: i don't think we're going to have a separate side panel for edges from what i recall. @alanonthegit correct me if i'm wrong

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -99,17 +107,25 @@ const TopologyComponent = () => {
...
                 </TopologySideBar>
             }
-            sideBarOpen={!!selectedId}
+            sideBarOpen={!!selectedObj}
```
alwayshooin, line 125: you can have the side bar be open but not shown? that seems a little confusing to me. is that case for like a minimized side panel?

alwayshooin (approved):
looks great! just a few questions in the code for you

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: Do you mean how TopologyView has the sideBarOpen prop and the TopologySideBar has the show prop? I tested it out and seems like we need sideBarOpen but the show prop didn't seem necessary. I removed that one

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: Yea true. Let me remove that

PR #3394, opened 2022-10-12
Fix permissions in central service account to allow diagnostic bundle to retrieve pod data

dhaus67:
## Description

Currently, whe ncreating a diagnostic bundle, the central cluster will contain a `errors.txt` file with the following error message:
```
could not list pods in namespace "stackrox": pods is forbidden: User "system:serviceaccount:stackrox:central" cannot list resource "pods" in API group "" in the namespace "stackrox"
```

Before trying to retrieve any pod data (e.g. logs), we will first [check if we can list pods in the specific namespace](https://github.com/stackrox/stackrox/blob/eb7696c4b33e285969130d8ede4faf88e901ded2/pkg/k8sintrospect/collector.go#L341-L345).

However, the service account of central, which will be used, does not have the permission to list `pods` within [the stackrox-central-diagnostics 
role](https://github.com/stackrox/stackrox/blob/3d1daa92cfa7a9117753cae8aa58a5beb0bda272/image/templates/helm/stackrox-central/templates/01-central-03-diagnostics-rbac.yaml#L15-L20).

This leads to the error during checking if we can list pods, as well as the `errors.txt` file being created.

This PR fixes the listed resources within the role.

Although we previously didn't announce any changes related to the service accounts RBAC, I've added a changelog to let customers know that the role now includes one additional resource.

## Checklist

## Testing Performed

- Create a diganostic bundle
- Ensure that within `_central-cluster` folder, no `errors.txt` is found and the log data is available for each pod.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   The default SCC for sensor is now `restricted[-v2]` or `stackrox-sensor` depending on the settings.
   Both the `runAsUser` and `fsGroup` for the admission-control and sensor deployments are no longer hardcoded to 4000 on Openshift clusters
   to allow using the `restricted` and `restricted-v2` SCCs.
+- The service account "central", which is used by the central deployment, will now include `get and list` access to the pod resource in the namespace
+  where central is deployed to. This fixes an issue when generating diagnostic bundles to now correctly include all logs within the namespace of central.
 
 ## [3.72.0]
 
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #3394** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3394" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3394?utm_source=stack-comment).

misberner (approved):

dhaus67:
Test failures are infrastructure related (e.g. issues brining up the cluster), and the current change only applies to the Helm chart, so this can be ignored for this particular PR.

PR #3399, opened 2022-10-12
ROX-12834: Introduce fake nodescan generator

Maddosaurus:
## Description

This PR introduces a generator to send fake NodeScanV2 messages from Collector to Sensor.
These messages enable us to build out the full pipeline independently from implementing the NodeScanV2 logic.

* This feature is hidden behind a feature flag. Use `ROX_RHCOS_NODE_SCANNING=true` when calling `deploy.sh` to enable it.
* The interval for scans is already customizable through an env var, `ROX_NODE_SCAN_INTERVAL` that defaults to 4 hours
  * The interval can be decreased by setting this var to another value for the Collector image in the Collector DaemonSet 

## Checklist

## Testing Performed

1. Deploy (e.g. on local) with enabled feature flag
`ROX_RHCOS_NODE_SCANNING=true ./deploy/k8s/deploy-local.sh`
2. Observe the log message in the compliance container (in the compliance pod) that the scanning interval is 4 hours
3. Observe a fake Node Scan message being logged in Sensor logs once at start, then every 4 hours
~~4. Edit Collector DaemonSet and add a custom interval as env var to the collector image~~

4. Set environment variable via `kubectl`: `kubectl -n stackrox set env daemonsets/collector --containers="compliance" ROX_NODE_RESCAN_INTERVAL="20s"` 

5. Observe the Collector Pod getting reinstantiated and the collector image log message stating an interval of 15 seconds.

6. Finally, `teardown` everything and redeploy with `deploy.sh` without setting the feature flag.
Observe the collector log not stating a Node Scan Interval.

### Additional testing: Resilience
We also performed error cases:
1. Kill Sensor pod, observe Compliance container re-establishing communication and continue sending messages
2. Kill Compliance pod, observe Compliance container re-establish communication and continue sending messages


compliance/collection/main.go
```diff
@@ -181,6 +183,47 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
...
+	}
+	t := time.NewTicker(env.NodeScanInterval.DurationSetting())
+	scanner := nodescanv2.FakeNodeScanner{} // FIXME: Replace with real scanner (ROX-12971)
+
+	// send scan result once at startup, then every NodeScanInterval
+	if err := scanNode(client, &scanner); err != nil {
+		log.Errorf("error running scanNode: %v", err)
+	}
+
+	log.Infof("Node Scan interval: %v", env.NodeScanInterval.DurationSetting())
```
vikin91, line 199: I would move this log message a bit up:

```suggestion
	log.Infof("Node Scan interval: %v", env.NodeScanInterval.DurationSetting())
	
	// send scan result once at startup, then every NodeScanInterval
	if err := scanNode(client, &scanner); err != nil {
		log.Errorf("error running scanNode: %v", err)
	}
```

vikin91 (commented):
It looks really good! I was able to run it and see it working. Leaving my comments for now and will finish the review tomorrow (I want to give it a final look).

compliance/collection/main.go
```diff
@@ -201,8 +244,14 @@ func main() {
 	ctx = metadata.AppendToOutgoingContext(ctx, "rox-compliance-nodename", getNode())
 
 	stoppedSig := concurrency.NewSignal()
+
 	go manageStream(ctx, cli, &stoppedSig)
 
+	if features.RHCOSNodeScanning.Enabled() {
+		cliNode := sensor.NewComplianceServiceClient(conn)
```
vikin91, line 251: Why cannot we reuse the `cli` client from line 241?

compliance/collection/main.go
```diff
@@ -181,6 +183,47 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
 	return client, config, nil
 }
 
+func manageNodescanLoop(ctx context.Context, cli sensor.ComplianceServiceClient) {
```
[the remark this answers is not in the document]
vikin91: I see now that the `manageNodescanLoop` func duplicates a lot from `manageStream`. I see some benefits for readability and separation of code for different scenarios, but I am bit worried regarding the duplicated call to `initializeStream`.

How do you feel about merging `manageNodescanLoop` into `manageStream`? Would there be any consequences of this?

compliance/collection/main.go
```diff
@@ -181,6 +183,47 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
...
+	client, _, err := initializeStream(ctx, cli)
+	if err != nil {
+		log.Fatalf("error initializing stream to sensor: %v", err)
+	}
+	t := time.NewTicker(env.NodeScanInterval.DurationSetting())
+	scanner := nodescanv2.FakeNodeScanner{} // FIXME: Replace with real scanner (ROX-12971)
```
vikin91, line 192: This is our main dependency of that stream-handling loop - maybe we should inject `scanner` from the main function?

compliance/collection/nodescanv2/fake_nodescan.go
```diff
@@ -0,0 +1,36 @@
...
+type FakeNodeScanner struct {
+}
+
+// Scan returns a fake message in the same format a real NodeScanV2 would produce
+func (f *FakeNodeScanner) Scan(nodeName string) (*storage.NodeScanV2, error) {
+	log.Infof("Generating fake scan result message...")
+	msg := &storage.NodeScanV2{
+		NodeId:   "",
+		NodeName: "Fake Testnode",
+		ScanTime: &timestamp.Timestamp{Seconds: 42, Nanos: 24},
```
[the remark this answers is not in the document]
Maddosaurus: I thought about possibly using the fake generator for tests. If we do that, a fixed timestamp would be easier to test, but I'm open to switching to a real time for now and switching that back if we end up using this as test fixture 👍

compliance/collection/main.go
Maddosaurus: I've redesigned this area to allow reuse of the existing client

compliance/collection/main.go
```diff
@@ -181,6 +175,42 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
 	return client, config, nil
 }
 
+func manageNodeScanLoop(ctx context.Context, client sensor.ComplianceService_CommunicateClient, scanner nodescanv2.NodeScanner) {
+	t := time.NewTicker(env.NodeScanInterval.DurationSetting())
```
vikin91, line 179: I didn't check that before, but I see that `DurationSetting()` can return 0 or negative durations. The Ticker must get positive >0 duration or it will panic. I think, we need a check that would prevent us from running into problems when users provide trash input through the env variable.

compliance/collection/main.go
```diff
@@ -105,21 +107,13 @@ func startAuditLogCollection(ctx context.Context, client sensor.ComplianceServic
...
 			return
 		default:
-			client, config, err := initializeStream(ctx, cli)
-			if err != nil {
-				if ctx.Err() != nil {
-					// continue and the <-ctx.Done() path should be taken next iteration
-					continue
-				}
-				log.Fatalf("error initializing stream to sensor: %v", err)
-			}
```
vikin91, line 122: I wonder why this has been put in to the default case in the first place 🤔 Having it kept here, the code would have been called 1 or more times. Now we would call the initialization only once. Not sure whether this is a problem, but we should keep an eye on the stability of the stream.

For example: would we be able to reestablish the connection if we crash sensor while compliance is still working? That is something to check.

compliance/collection/main.go
vikin91: Confirmed that a negative value causes a crash: 

```
│ panic: non-positive interval for NewTicker                                                                                                                                                          │
│                                                                                                                                                                                                     │
│ goroutine 26 [running]:                                                                                                                                                                             │
│ time.NewTicker(0xc0007129d8?)                                                                                                                                                                       │
│     time/tick.go:24 +0x10f                                                                                                                                                                          │
│ main.manageNodeScanLoop({0x233f9a0, 0xc000592090}, {0x2345340, 0xc0005234f0}, {0x2319460, 0x369f8e8})                                                                                               │
│     github.com/stackrox/rox/compliance/collection/main.go:179 +0x6a                                                                                                                                 │
│ created by main.main                                                                                                                                                                                │
│     github.com/stackrox/rox/compliance/collection/main.go:244 +0x525                                                                                                                                │
│ Stream closed EOF for stackrox/collector-9qcdt (compliance)
```

compliance/collection/main.go
vikin91: Yes, the crash of sensor causes the connection to break and there is no self-healing. We need to fix that by reverting the change and merging the two loops by adding a new case to the for-switch loop.

compliance/collection/nodescanv2/nodescan.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
```
msugakov, line 11: Cosmetics and subjective, bug I think, `log` should go to `_impl.go` file if that exists. In this case this is `fake_nodescan.go`.
Log is an implementation detail that doesn't have to be exposed in the file where you define interfaces.

compliance/collection/nodescanv2/nodescan.go
```diff
@@ -0,0 +1,25 @@
...
+
+// NodeScanner defines an interface for V2 NodeScanning
+type NodeScanner interface {
+	Scan(nodeName string) (*storage.NodeScanV2, error)
```
msugakov, line 15: I think the node name should not be passed to the implementation that scans the node. That's the detail that should be filled in `storage.NodeScanV2` by the calling code in Compliance.

Moreover, it seems to me that `NodeScanner` should return components and notes and not care about the rest.

compliance/collection/nodescanv2/nodescan.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// NodeScan is the V2 NodeScanning implementation
+type NodeScan struct {
+}
+
+// Scan scans the current node and returns the results as storage.NodeScanV2 object
+func (n *NodeScan) Scan(nodeName string) (*storage.NodeScanV2, error) {
+	return nil, errors.New("Not implemented")
+}
```
msugakov, line 25: Something tells me that this is redundant for now. Up to you if you want to keep it though.

compliance/collection/main.go
```diff
@@ -201,7 +231,18 @@ func main() {
...
+
+	client, config, err := initializeStream(ctx, cli)
+	if err != nil {
+		log.Fatalf("error initializing stream to sensor: %v", err)
+	}
+
+	go manageStream(ctx, client, config, &stoppedSig)
+
+	if features.RHCOSNodeScanning.Enabled() {
+		scanner := nodescanv2.FakeNodeScanner{} // FIXME: Replace with real scanner (ROX-12971)
```
msugakov, line 243: [nit] There's a different format of TODO comments that our CI understands.
```suggestion
		// TODO(ROX-12971): Replace with real scanner
		scanner := nodescanv2.FakeNodeScanner{}
```

msugakov (commented):
Quick and shallow check, did not go through all code.

compliance/collection/nodescanv2/nodescan.go
Maddosaurus: Well spotted, that was a leftover from refactoring. Moved!

compliance/collection/nodescanv2/nodescan.go
Maddosaurus: The decision was influenced by 2 factors:
1. Scanners `Analyze` function that we will call in the real implementation requires the node name at call time
2. compliance main.go already defines and uses code ([here](https://github.com/stackrox/stackrox/blob/master/compliance/collection/main.go#L36)) to determine the Node name
3. 
I aimed to minimize code changes, but I could as well refactor the existing `getNode` function into the `compliance.utils` package and use it directly inside `NodeScanner.Scan()` instead of passing it from the calling main compliance loop. WDYT?

msugakov: I did not check that node name is a requirement of the real `Analyze()`. Thanks for pointing it out.

I think, it does not need to be a requirement of `Analyze` because the node name is used only as an additional logged field. I suppose eventually this field may be handy during logs analysis for filtering logs collected from all nodes.
Also, it is clear that the node name is used instead of the image layer name in logging because image layers don't exist for a node.

Overall, I don't mind to keep the node name there for now. Building something different to pass this info as some "generic logging field" parameter would result in much larger code with tiny or no benefit.
Let's keep it as-is for now and see where we end up when the real implementation will be used.

compliance/collection/main.go
Maddosaurus: Thank you for checking this! 
I have implemented a safety check to validate what is put into this variable.

compliance/collection/utils/verify_duration.go
```diff
@@ -0,0 +1,20 @@
...
+
+var (
+	log = logging.LoggerForModule()
+)
+
+// VerifyAndUpdateDuration ensures that a given duration is positive bigger than zero and returns a default otherwise
+func VerifyAndUpdateDuration(duration time.Duration) time.Duration {
+	if (duration) <= 0 {
+		log.Warn("Negative or zero duration found. Setting to 4 hours.")
+		return time.Hour * 4
```
msugakov, line 17: `time.Hour * 4` is very specific for node rescan interval, therefore this function cannot be presented as generic "util". I think you should simply define it as `GetNodeRescanInterval` in `pkg/env/node_scan.go`. Then, instead of doing this

```go
rescanInterval := complianceUtils.VerifyAndUpdateDuration(env.NodeRescanInterval.DurationSetting())
```

you can simply do this

```go
rescanInterval := env.GetNodeRescanInterval()
```

compliance/collection/main.go
msugakov: It seems to me, the code before the change will shut down Compliance when the connection to Sensor is lost.
Piotr, in your experiment did you check behavior of new or of unchanged code?

vikin91: I checked the behavior from after the change that I commented on. I also saw the fatal message, but I couldn't see compliance crashing after the sensor was gone. I guess we would need to check this on the original code.

compliance/collection/main.go
```diff
@@ -105,21 +108,13 @@ func startAuditLogCollection(ctx context.Context, client sensor.ComplianceServic
...
-			if err != nil {
-				if ctx.Err() != nil {
-					// continue and the <-ctx.Done() path should be taken next iteration
-					continue
-				}
-				log.Fatalf("error initializing stream to sensor: %v", err)
-			}
 			if err := runRecv(ctx, client, config); err != nil {
 				log.Errorf("error running recv: %v", err)
 			}
```
msugakov, line 120: @vikin91 @Maddosaurus 
Unless I read the code incorrectly, the `manageStream` function does not function the way it supposed to.
I think, its idea should be to check whether the connection got closed (`ctx.Done()`) and in this case shut down the process (`sig.Signal()`), or (if the connection is still alive) receive a message from Sensor and respond back (`runRecv()`). The problem is that `runRecv()` is a blocking function that runs an infinite loop of receive-reply logic. Once the code gets to it, it will be running forever until an error happens or unsupported message arrives.
When the code is in `runRecv()`, the shutdown logic of `manageStream` won't be executed normally.
Am I right or am I missing something?

compliance/collection/main.go
Maddosaurus: I have reverted to the original code and also implemented a similar mechanic for the NodeScan loop.
I tested this by killing the sensor pod while compliance is running and was able to verify that both loops connect to (the newly created) sensor again.

compliance/collection/main.go
vikin91: I definitely want to look at sensor and see how it is done there.

For compliance, I see that `client.Recv()` is being called on a client that was created with the context `client, err := cli.Communicate(ctx)`, so I hope that canceling the context would shutdown the client and release the lock.
(I used word "hope" as I have not verified this and do not know the internals of grpc stream).

I will do some more reading on this.

compliance/collection/main.go
```diff
@@ -108,20 +108,52 @@ func startAuditLogCollection(ctx context.Context, client sensor.ComplianceServic
...
+			if err != nil {
+				t.Reset(5 * time.Second)
+				log.Errorf("error initializing node scan stream to sensor: %v", err)
+				continue
+			}
+			if err := scanNode(client, scanner); err != nil {
+				log.Errorf("error running scanNode: %v", err)
+				continue
+			}
+			t.Reset(rescanInterval)
```
vikin91, line 152: This could be an `else` instead of the `continue`

compliance/collection/main.go
vikin91: So sensor utilizes two clever objects to manage the stream to central: a pointer `*grpcUtil.LazyClientConn` holding a lazy connection, and `centralclient.CentralConnectionFactory` that sets, monitors the connection and restarts it when failed. All communication attempts are blocking and waiting when the connection is down (no message is lost). The operations resume when the connection is back up.

We could apply this to compliance-sensor as well, but this may deserve a separate refactoring PR. Would it be okay to continue in the current shape and improve this in a separate PR @msugakov ?

compliance/collection/nodescanv2/fake_nodescan.go
```diff
@@ -0,0 +1,83 @@
...
+// Scan returns a fake message in the same format a real NodeScanV2 would produce
+func (f *FakeNodeScanner) Scan(nodeName string) (*storage.NodeScanV2, error) {
+	log.Infof("Generating fake scan result message...")
+	msg := &storage.NodeScanV2{
+		NodeId:   "",
+		NodeName: nodeName,
+		ScanTime: timestamp.TimestampNow(),
+		Components: &scannerV1.Components{
+			Namespace: "Testme OS",
+			OsComponents: []*scannerV1.OSComponent{
```
RTann, line 27: Since we only care about RHCOS at this time, I'd leave this empty/nil

compliance/collection/nodescanv2/fake_nodescan.go
```diff
@@ -0,0 +1,83 @@
...
+				},
+			},
+			RhelComponents: []*scannerV1.RHELComponent{
+				{
+					Name:      "vim-minimal",
+					Namespace: "rhel:7",
+					Version:   "2:7.4.629-6.el7.x86_64",
+					Arch:      "x86_64",
+					Module:    "FakeMod",
+					Cpes:      []string{"cpe:/a:redhat:enterprise_linux:7::baseos"},
```
RTann, line 48: Not very important, but maybe use 8 instead of 7 due to RHCOS

compliance/collection/nodescanv2/fake_nodescan.go
```diff
@@ -0,0 +1,83 @@
...
+					Name:      "sqlite-libs",
+					Namespace: "centos:8",
+					Version:   "3.26.0-6.el8",
+					AddedBy:   "FakeLayer",
+				},
+			},
+			RhelComponents: []*scannerV1.RHELComponent{
+				{
+					Name:      "vim-minimal",
+					Namespace: "rhel:7",
```
RTann, line 44: rhel:8 here as well

compliance/collection/nodescanv2/fake_nodescan.go
```diff
@@ -0,0 +1,83 @@
...
+					Version:   "2:7.4.629-6.el7.x86_64",
+					Arch:      "x86_64",
+					Module:    "FakeMod",
+					Cpes:      []string{"cpe:/a:redhat:enterprise_linux:7::baseos"},
+					AddedBy:   "FakeLayer",
+					Executables: []*scannerV1.Executable{
```
RTann, line 50: let's ignore executables as well since this is for Active Vuln mgmt, and I don't think we plan on having that in the first iteration

compliance/collection/utils/verify_duration.go
RTann: I don't think we need to worry about this because `DurationSetting()` already handles this:

```
// DurationSetting returns the Duration object represented by the environment variable
func (d *DurationSetting) DurationSetting() time.Duration {
	val := os.Getenv(d.envVar)
	if val != "" {
		dur, err := time.ParseDuration(val)
		if err == nil {
			return dur
		}
		log.Warnf("%s is not a valid environment variable for %s, using default value: %s", val, d.envVar, d.defaultDuration.String())
	}
	return d.defaultDuration
}
```

That being said, in another PR, we should probably add a `panic` in the `registerDurationSetting` if the default is not valid

RTann: https://github.com/stackrox/stackrox/pull/3558

compliance/collection/main.go
```diff
@@ -127,6 +130,30 @@ func manageStream(ctx context.Context, cli sensor.ComplianceServiceClient, sig *
...
+func manageNodeScanLoop(ctx context.Context, cli sensor.ComplianceServiceClient, scanner nodescanv2.NodeScanner, rescanInterval time.Duration) {
+	t := time.NewTicker(5 * time.Second)
+	log.Infof("Node Rescan interval: %v", rescanInterval)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			client, _, err := initializeStream(ctx, cli)
```
RTann, line 142: I wonder if we should throw this inside of `manageStream` so we can just reuse the same stream and use the same stop signals

compliance/collection/utils/verify_duration.go
RTann: ah actually `time.ParseDuration` does allow for negative durations, so `DurationSetting()` does not handle this. Perhaps we should include that in the PR above ^

compliance/collection/main.go
vikin91: We tried this and this is not trivial, because the `runRecv` blocks. I have a variation on this here https://github.com/stackrox/stackrox/pull/3548 but it also uses two loops - one for sending and one for receiving.

compliance/collection/utils/verify_duration.go
Maddosaurus: Thanks for linking the PR! We'll keep an eye on it and update our code as soon as your PR is merged.
For now, we want to get this PR merged rather quick, though.

compliance/collection/main.go
vikin91: In [a35000a](https://github.com/stackrox/stackrox/pull/3399/commits/a35000a42f261cf1a0e135f0714354e3ccf49198), we applied the idea from #3548 and now, we have separate loops for sending and receiving

compliance/collection/main.go
msugakov: The code has changed now a lot. Issues that were before are now gone but new things still don't look 100% self-explanatory to me. I think we (I?) need to confidently understand well what we do. I would not classify "understanding" as "improvement" or "optimization". "Understanding" is our core skill.

compliance/collection/main.go
```diff
@@ -154,6 +218,34 @@ func initialClientAndConfig(ctx context.Context, cli sensor.ComplianceServiceCli
 	return client, config, nil
 }
 
+// initializeSendingStream returns a client for sending the data to sensor. It ensures that the stream is open and in case not,
+// then the connection is being reopened with exp. back off
+func initializeSendingStream(ctx context.Context, cli sensor.ComplianceServiceClient) (sensor.ComplianceService_CommunicateClient, error) {
```
msugakov, line 223: Few stupid questions from me to make your life seem harder:
What's the difference between `initializeStream` and `initializeSendingStream`?
Does the stream returned by `initializeStream` not allow sending?
Can you please eliminate copy-pasted code and customize some generic function for both cases (if existence of two functions is justified)?

pkg/env/node_scan.go
```diff
@@ -0,0 +1,18 @@
+package env
+
+import "time"
+
+var (
+	nodeRescanIntervalDefault = 4 * time.Hour
+	// NodeRescanInterval will set the duration for when to scan nodes for vulnerabilities (NodeScanV2)
+	NodeRescanInterval = registerDurationSetting("ROX_NODE_RESCAN_INTERVAL", nodeRescanIntervalDefault)
```
msugakov, line 8: [nit] This setting can be unexported because its value is available through `GetNodeRescanInterval`

```suggestion
	nodeRescanInterval = registerDurationSetting("ROX_NODE_RESCAN_INTERVAL", nodeRescanIntervalDefault)
```

msugakov (commented):
Sorry, I again wasn't able to dedicate proper time on reviewing.

compliance/collection/main.go
vikin91: Let me answer those valid (not stupid) questions.

> What's the difference between `initializeStream` and `initializeSendingStream`?

One is for sending the other one is for receiving. Placing the sending code in the receiving loop is difficult as the receiver routine must block. Sensor has two separate loops for this and we applied it here separately as well.
The receiving side requires some additional steps when compared to sending (e.g. the config object, the hello message from sensor), so I found it cleaner to add sending separately and leave the receiving code untouched.
The naming of `initializeStream` could be improved to `initializeRcvStream` - would that be ok?

> Does the stream returned by `initializeStream` not allow sending?

It does allow sending. The problem is different. If the connection is down (e.g., due to sensor restart), we would need to reconnect before sending (this is what we do in `initializeSendingStream`). If we would decide to reuse the receiving client for sending then in case of the reconnection on the receiving side, the client would be invalidated. The receiving side would recreate the client in the `initializeStream` loop and the sending side would need to be notified about this to stop using the old client and start using a new one. We could share the client as it it done in sensor, but this is not trivial. 

In sensor, we also use two loops, but we share the client (using unsafe ptrs inside of `centralclient.CentralConnectionFactory`) - this is possible to apply here, but that would be a quite a big change and would make this PR much larger. We can still consider this to be applied in a follow-up if we want. 

In this PR we took a simplified approach (when compared against the solution used in sensor) and we provide two clients and two loops that take care about the reconnections.
Note that mentioning two clients here, I mean two child `sensor.ComplianceService_CommunicateClient`s while still using a single parent `sensor.ComplianceServiceClient`. While I am not fully sure what the precise overhead of this combination is, I think that we still use the same stream and the overhead is minimal. 

> Can you please eliminate copy-pasted code and customize some generic function for both cases (if existence of two functions is justified)?

Yes, this is something we can do.

pkg/env/node_scan.go
vikin91: Yep, it makes sense.

compliance/collection/main.go
vikin91: I have removed now the `initializeSendingStream` to address the code duplication issue. We will use `initializeStream` for both cases and have a followup ticket to handle it the same way as in sensor (if we identify that there are benefits coming from this approach).

Maddosaurus:
I have rebased the PR for a final time to benefit from the latest CI flake fixes.
I also put in another round of manual tests as described in the PR description to ensure that everything still works after the changes we introduced in the latest commits.

compliance/collection/main.go
msugakov: I feel like I'm not able to continue and conclude my review in a timely manner due to many other obligations. At the same time I don't want to hold you back. Please proceed without me, but please make sure you understand how things work and you're 100% confident that the original compliance isn't broken.

vikin91: I feel pretty confident, that we didn't break the logic that was there before. The only thing that we made a bit worse is the introduction of the second client that reinitializes its own connection before sending. Obviously, we need that client, and this client needs to make sure that the connection is up before sending, so let's see how large is the overhead from it.

Maybe we should spin a long-running cluster with our FF enabled to monitor the system with our changes while we work on that epic @Maddosaurus ?

compliance/collection/main.go
vikin91: Yes, sorry for adding so many changes after you started the review. I think they were needed to address some obvious drawbacks regarding sending without being sure that the connection is up.

While there are still some bits about the GRPC stream that I would like to understand better, I feel pretty confident that the code handling sending and receiving in this component follows the same principles as in other parts of our code.

vikin91:
@Maddosaurus shall we have another CI run with the new ci-labels that Misha added?

compliance/collection/main.go
Maddosaurus: @vikin91 A long running cluster would be a great idea. How about I spin one up that runs over the weekend and we merge on Monday morning, given that the cluster and the tests look good?

vikin91: I would say, we first merge ant then do the cluster :) Our changes are behind a FF, so there should be no impact to the rest of the product.

compliance/collection/main.go
```diff
@@ -105,7 +107,26 @@ func startAuditLogCollection(ctx context.Context, client sensor.ComplianceServic
...
+			return
+		case sc := <-sensorC:
+			client, _, err := initializeStream(ctx, cli)
+			if err != nil && ctx.Err() == nil {
+				// error even after retries
+				log.Fatalf("unable to establish send stream to sensor: %v", err)
+			}
+			if err := client.Send(sc); err != nil {
+				log.Errorf("failed sending nodeScanV2 to sensor: %v", err)
+			}
```
msugakov, line 124: I read up about gRPC yesterday and got a bit more dangerous ;-) although still not very educated.

And I'm definitely here to spoil the party that started (even though I haven't used the opportunity to do that earlier). It seems this code will open a stream each time there's something to send. I _think_ gRPC maintains a pool of connections on the client, plus streams get multiplexed over HTTP/2 (but only when that's used), so opening a stream doesn't mean establishing new connection, but I _think_ it still has some overhead.

What is worse, I _believe_ that we reuse the same `rpc Communicate` as the ordinary Compliance does. An effect of that might be that Sensor will push its messages over that stream too, but on Compliance side there's nothing receiving and acting on them _for this stream_. In case Sensor pushes a message to perform Compliance checks (`message TriggerRun`) through this stream, it will get lost.

From looking at Sensor code, it seems to me that more recent connection (actually stream) replaces previous in the map keyed by a node name. Therefore, I strongly suspect that what I described actually happens _if the feature flag for node scans is turned on_, i.e. compliance checks stop working properly.

I don't think it's an issue for now as the feature is gated but I think the logic should be reworked before we are ready to release.

WDYT?

vikin91: That is a very good find. We should definitely improve this before releasing - it deserves a ticket.

Applying the connection factory with shared pointer to the stream (as seen in Sesor-Central) should fix the problem that you described, but I am not saying that this is the only viable solution to it. I will open a ticket.

vikin91: A thought about this: 

> In case Sensor pushes a message to perform Compliance checks (message TriggerRun) through this stream, it will get lost.

we may close the stream immediately after sending the scan results to notify sensor that this connection should not be used anymore.

vikin91: >  As a peer in the stream you can either receive or send (or do nothing) but not receive and send concurrently.

This would be a limitation. I got the impression from Sensor<->Central that the sending and receiving works concurrently, but I didn't confirm this experimentally.

Having a separate `rpc` in the proto file would be also an option (maybe even a much cleaner one), but I do not think that this is necessary. The way we initiate the stream clients gives us possibility to send without depending on whether something was received (it is not in the final PR, but we had this variant shortly in the branch).

I am thinking how to design a manual test to confirm or deny our theories 🤔

vikin91: Created https://issues.redhat.com/browse/ROX-13383 to not forget about the issue we are discussing here.

Maddosaurus: > In case Sensor pushes a message to perform Compliance checks (message TriggerRun) through this stream, it will get lost.

A single manual test is nothing to go by, but I was able to successfully run a compliance check with the FF turned on, so it might be the message isn't lost necessarily.
But we should definitely investigate this further! Thanks for creating the ticket @vikin91

PR #3403, opened 2022-10-12
Dashrews/ROX-13063 fix network flow deadlock when we have overlapping delete calls

dashrews78:
## Description

Between removing flows when a deployment is remove and removing stale flows from the garbage collector, it is possible to get in a deadlock situation when those bulk remove operations coincide.  I considered simply wrapping it in a retry and also considered using the Postgres feature `SKIP LOCKED`.  I determined that the cleanest and best way is to avoid the deadlock situation altogether.  So I pushed the pruning of stale network flows down to the store and wrap the remove calls in a lock so only one of those is processing at time.  The store is implemented in such a way that there should be an instance per cluster.  in actuality it was creating a new version of the store all the time.  so locking on the removes did not work.  Updated the store to have one per cluster so that I could use a mutex to lock the calls to remove.

## Checklist

## Testing Performed

Mostly an exercise in moving code.  Tests exists.  But setup some long running and scale tests to verify deadlock has been avoided.

### DIFF
--- a/central/networkgraph/flow/datastore/flow_impl.go
+++ b/central/networkgraph/flow/datastore/flow_impl.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stackrox/rox/central/networkgraph/flow/datastore/internal/store"
 	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/expiringcache"
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/sac"
@@ -130,4 +131,17 @@ func (fds *flowDataStoreImpl) RemoveMatchingFlows(ctx context.Context, keyMatchF
 		return sac.ErrResourceAccessDenied
 	}
 	return fds.storage.RemoveMatchingFlows(ctx, keyMatchFn, valueMatchFn)
+}
+
+// RemoveStaleFlows - remove stale duplicate network flows
+func (fds *flowDataStoreImpl) RemoveStaleFlows(ctx context.Context) error {
+	if ok, err := networkGraphSAC.WriteAllowed(ctx); err != nil {
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return fds.storage.RemoveStaleFlows(ctx)
+	}
+	return nil
 }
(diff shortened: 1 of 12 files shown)


dashrews78:
Back to draft.  I missed something somewhere.

central/networkgraph/flow/datastore/internal/store/postgres/store.go
```diff
@@ -63,6 +65,23 @@ const (
 	SELECT nf.Props_SrcEntity_Type, nf.Props_SrcEntity_Id, nf.Props_DstEntity_Type, nf.Props_DstEntity_Id, nf.Props_DstPort, nf.Props_L4Protocol, nf.LastSeenTimestamp, nf.ClusterId
 	FROM network_flows nf ` + joinStmt +
 		`WHERE nf.Props_DstEntity_Type = 1 AND nf.Props_DstEntity_Id = $1 AND nf.ClusterId = $2`
+
+	pruneStaleNetworkFlowsStmt = `DELETE FROM network_flows a USING (
+      SELECT MAX(flow_id) as max_flow, props_srcentity_type, props_srcentity_id, props_dstentity_type, props_dstentity_id, props_dstport, props_l4protocol, clusterid
```
c-du, line 70: Even the db is not case sensitive, it is better to keep it consistent. Props_DstEntity_Type

c-du (approved):

central/networkgraph/flow/datastore/internal/store/postgres/store.go
c-du: Two nit issues.

central/networkgraph/flow/datastore/internal/store/postgres/cluster_impl.go
```diff
@@ -5,28 +5,42 @@ import (
...
+		db:        db,
+		flowStore: make(map[string]store.FlowStore),
 	}
 }
 
 type clusterStoreImpl struct {
-	db *pgxpool.Pool
+	db        *pgxpool.Pool
+	lock      sync.Mutex
+	flowStore map[string]store.FlowStore
```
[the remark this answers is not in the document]
dashrews78: This store is created on a per cluster basis, I presume to improve performance as that pre-dates me.  The methods to create and get the store were essentially the same where it would return a new store based on cluster.  So while not expensive, having multiple stores servicing the same cluster makes it difficult to prevent deadlocks because each path through the system. has access to a different copy of the store.  So using mutexes to prevent deadlocks became ineffective.  Doing it this way allows me to use a mutex to prevent deadlocks on the bulk delete calls.

central/networkgraph/flow/datastore/flow_impl.go
```diff
@@ -131,3 +132,16 @@ func (fds *flowDataStoreImpl) RemoveMatchingFlows(ctx context.Context, keyMatchF
...
 }
+
+// RemoveStaleFlows - remove stale duplicate network flows
+func (fds *flowDataStoreImpl) RemoveStaleFlows(ctx context.Context) error {
+	if ok, err := networkGraphSAC.WriteAllowed(ctx); err != nil {
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
```
md2119, line 143: Seeing this now but I feel the env check should be performed before SAC check to avoid eventual noop.

PR #3409, opened 2022-10-13
ROX-11323 Add labels pool in fake workloads

lvalerom:
## Description

Follow up PR to #3341. Creates a pool of labels to ensure label matching between resources.

## Checklist

## Testing Performed

Use a workload with `matchLabels: true` configured.

```
go run tools/local-sensor/main.go -record -with-fakeworkload=workload.yaml
```

Labels used by services and network policies should match labels used by deployments

### DIFF
--- a/sensor/kubernetes/fake/deployment.go
+++ b/sensor/kubernetes/fake/deployment.go
@@ -34,12 +34,27 @@ func createRandMap(stringSize, entries int) map[string]string {
 	return m
 }
 
+func createMap(entries int) map[string]string {
+	m := make(map[string]string, entries)
+	for i := 0; i < entries; i++ {
+		m[fmt.Sprintf("key-%d", i)] = fmt.Sprintf("value-%d", i)
+	}
+	return m
+}
+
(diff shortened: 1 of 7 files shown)


Maddosaurus (approved):
Looks good to me!
Out of interest: Do we expect any impact on performance when using non-matched labels, or was this added to create a test environment that's closer to reality?

lvalerom:
> Out of interest: Do we expect any impact on performance when using non-matched labels, or was this added to create a test environment that's closer to reality?

TL;DR: we added this to get an environment closer to reality

If we use non-matched labels is very unlikely that we could have matching resources, therefore there is no way to guarantee that certain alerts will get triggered. Before this PR we could get all resources to match with all resources. That was good in the sense that we could guarantee the alerts getting triggered but it was not very realistic, thus this change.

Regarding the performance, in a totally random environment everything is a little bit.. random 🙂 it depends on how isolated the resources are. For example, in the unlikely situation in which all resources are isolated from each other then the performance will be better because we don't need to iterate through a lot of resources to find any matching ones. I would say though that the performance should be similar in all cases.

PR #2527, opened 2022-07-26
CI: Allow a range of CVE count from GCR

gavin-stackrox:
## Description

Be less tied to GCRs current state.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

CI is sufficient

PR #3415, opened 2022-10-13
CI: Increase a range of CVE count to avoid flakes

rukletsov:
## Description

Looks the expected CVE range introduced in https://github.com/stackrox/stackrox/pull/2527 is not enough: I've seen "Verify Image Registry+Scanner Integrations" test fail today on most flavours with 181 CVEs.

Increasing the range seems harmless since a reasonable ballpark very likely means a successful test. 0 or 100500 CVE would indicate that something is wrong.

## Checklist

## Testing Performed

CI run.


rukletsov:
`go-postgres-tests` failure can't be related and won't be investigated here because this PR aims to fix a consistent CI failure.

theencee (approved):

gavin-stackrox (approved):

rukletsov:
Looks like the issue is resolved, merging.

PR #3416, opened 2022-10-13
Dashrews/ROX-13072 migrator cleanup of rocks causes bad symlink and crashing migrator

dashrews78:
## Description

When we determined that we no longer need RocksDB, the migrator will decomission it and removed current and .db-*.  The problem is that since the PVC hasn't been fully cleaned up yet, central adds some stuff back to the PVC and doesn't create things in the expected way.  This causes an issue where migrator will not be able to recover because the symlink of current is pointing to something invalid.  This is evident if you try to rollback but forget to add the forceRollbackVersion to the config map.  You should be able to recover from that by editing the config map.  But when we remove the directory and links we get ourselves in a position we cannot recover from..  So the simplest and safest path for now is to simply  remove the migration_version.yaml and rename the database current points to, to be an obviously unsupported database.  This ensures we have nothing in there to interfere with Postgres and allows us to recover from situations like this.  When we finally get all the stuff off the PVC then we can safely remove the directory.

## Checklist

## Testing Performed

Deploy a central
Upgrade central to a new version
Rollback central
Notice we are in a crash back because forceRollbackVersion is not set
Set forceRollbackVersion in the config map
Notice central stops crashing.

### DIFF
--- a/migrator/clone/rocksdb/db_clone_manager_impl.go
+++ b/migrator/clone/rocksdb/db_clone_manager_impl.go
@@ -104,7 +104,7 @@ func (d *dbCloneManagerImpl) Scan() error {
 		}
 	}
 
-	// Remove unknown clones that is not in use
+	// Remove unknown clones that are not in use
 	for _, r := range d.cloneMap {
 		clonesToRemove.Remove(r.GetDirName())
 	}
--- a/migrator/clone/rocksdb/db_clone_manager_impl.go
+++ b/migrator/clone/rocksdb/db_clone_manager_impl.go
@@ -341,15 +341,9 @@ func (d *dbCloneManagerImpl) DecommissionRocksDB() {
 		} else if exists {
 			// TODO(ROX-9882): While there is still stuff on the PVC there is an expectation that current
 			// exists in a certain state.  So until we can remove the rest of the components on the PVC, it
-			// is safer to clear the contents.  This will keep the link and directory in place to ensure
-			// things behave in a safe manner on restarts.
+			// is safer to put the contents in an obviously deprecated location.
 			if k == CurrentClone {
-				dir, err := os.ReadDir(dirPath)
-				if err != nil {
-					log.Error(err)
-					continue
-				}
-				// Move the data to the decommissioned rocks directory and point current there.
+				// Get the location Current points to
 				linkTo, err := fileutils.ResolveIfSymlink(dirPath)
 				if err != nil {
 					log.Error(err)
@@ -372,6 +366,7 @@ func (d *dbCloneManagerImpl) DecommissionRocksDB() {
 				}
 				// Remove clone symbolic link only, if exists.
 				_ = os.Remove(d.getPath(CurrentClone))
+				// Set the new symbolic link
 				if err := fileutils.AtomicSymlink(decommissionedCurrent, d.getPath(CurrentClone)); err != nil {
 					log.Error(err)
 					continue
--- a/migrator/clone/rocksdb/db_clone_manager_impl.go
+++ b/migrator/clone/rocksdb/db_clone_manager_impl.go
@@ -3,6 +3,7 @@ package rocksdb
 import (
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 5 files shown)


rhybrillou (approved):

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -327,27 +328,46 @@ func (d *dbCloneManagerImpl) GetDirName(cloneName string) string {
 func (d *dbCloneManagerImpl) DecommissionRocksDB() {
 	log.Info("DecommissionRocksDB")
 	for k := range d.cloneMap {
-		path := d.getPath(k)
-		if exists, err := fileutils.Exists(path); err != nil {
+		dirPath := d.getPath(k)
```
c-du, line 331: Decomission rocksdb should only happens when we have at least 1 postgres versions before and are attempting to upgrade to another. Why changing this could help in the case you described?

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -3,6 +3,7 @@ package rocksdb
 import (
```
c-du, line 3: What are the extra stuff added?

So while we upgrade to Postgres, if we keep previous valid which could be used for rollback and a fake current which could be at the version right before 73 DB, it looks like a valid clone set to old versions.

dashrews78: Answering this in the other question.

migrator/clone/rocksdb/db_clone_manager_impl.go
dashrews78: Because there are things like bolt and bleve and telemetry that still reference current and at least started by central.  In fact central puts a new Rocks in there.  We discussed this a little when I first put this method in but unwinding all that other stuff was going to become a bigger challenge than it looked on the surface.  So we decided we would take out the rest of the stuff when ROX-9882 is done.  At that time we can really remove current, .db-init, and whatever else is there.

We don't have to worry about being confused when we see the rocksdb clone because when this is executed the migration_version.yaml is removed.  So even though we will see a current rocksdb, its version will be 0 and thusly ignored in favor of Postgres.

dashrews78: I'd like to go through everything and make sure we stop referencing these things in central now, but I don't think we have the time for that.

c-du: Old versions add new rocks if it cannot find a valid current clone. In the scenario you described in test steps, it only upgrade to one version of Postgres, that means Central PVC has a valid set of databases including bolt, bleve and rocks. 

The scenario here is after Postgres migrated twice, if the databases are decomissioned, rolling back to an old rocksdb version would create new databases. Of course that is not supported but we still need to give it a way to recover :).

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -327,27 +328,46 @@ func (d *dbCloneManagerImpl) GetDirName(cloneName string) string {
...
+			// is safer to clear the contents.  This will keep the link and directory in place to ensure
+			// things behave in a safe manner on restarts.
+			if k == CurrentClone {
+				dir, err := os.ReadDir(dirPath)
+				if err != nil {
+					log.Error(err)
+					continue
+				}
+				for _, d := range dir {
+					log.Infof("Removing %q", d.Name())
```
c-du, line 347: If we remove all the content here, the old version would fail scan() while reading versions.
It would be better if we can carry some information to old versions telling people/support that what happens and how to address this condition.
For example, we can remove all copies like you did before and then add a new copy of link and target dir as 
current -> .central.internal.database.decommissioned.roll.back.not.supported.... and we can communicate this signature to support.
That could make us happier.

dashrews78: Failing scan wouldn't be an issue in Postgres mode as we don't fail migrator for a failed Rocks scan.  I think the safest way based on what we've been seeing is to keep whatever lingers in current except for migration_version.yaml. 
 We could remove/rename that to ensure we never consider this to be a valid DB to migrate from and change what current points to as you suggest.

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -327,27 +334,61 @@ func (d *dbCloneManagerImpl) GetDirName(cloneName string) string {
...
+					log.Error(err)
+					continue
+				}
+				if linkTo == d.getPath(decommissionedCurrent) {
+					log.Info("we have already moved to the RocksDB current to be decomissioned")
+					continue
+				}
+				// Remove the migrations versions to ensure we do not try to use this data for migrations
+				migrationVersionPath := filepath.Join(dirPath, migrations.MigrationVersionFile)
+				if err = os.RemoveAll(migrationVersionPath); err != nil {
```
c-du, line 358: You can remove the entire directory here. migrationVersionPath is only the yaml file.

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -327,27 +334,61 @@ func (d *dbCloneManagerImpl) GetDirName(cloneName string) string {
...
+				// Move directory: not following link, do not overwrite
+				cmd := exec.Command("mv", linkTo, d.getPath(decommissionedCurrent))
+				if output, err := cmd.CombinedOutput(); err != nil {
+					log.Error(errors.Wrapf(err, "failed to copy current db %s", output))
+					continue
+				}
+				// Remove clone symbolic link only, if exists.
+				_ = os.Remove(d.getPath(CurrentClone))
+				// Set the new symbolic link
+				if err := fileutils.AtomicSymlink(decommissionedCurrent, d.getPath(CurrentClone)); err != nil {
```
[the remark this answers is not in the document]
c-du: One suggestion, the desired sequence is:
1. create a decommissionedCurrent dir.
2. Fix symbolic link to decommissionedCurrent.
3. Remove the dirPath
This way we have a clear cut off time which is after step 2 and ensure idempotency if anything went wrong during this process.

migrator/clone/rocksdb/db_clone_manager_impl.go
```diff
@@ -327,27 +334,66 @@ func (d *dbCloneManagerImpl) GetDirName(cloneName string) string {
...
+				if linkTo == d.getPath(decommissionedCurrent) {
+					log.Info("we have already moved to the RocksDB current to be decomissioned")
+					continue
+				}
+				decommissionExists, err := fileutils.Exists(d.getPath(decommissionedCurrent))
+				if err != nil {
+					log.Error(err)
+					continue
+				}
+				if !decommissionExists {
```
c-du, line 362: nit: We can use MkdirAll to make sure the directory is there. It does not throw error if it is there already so you do not have to check if it exists.

c-du (approved):
With one comment.

PR #3417, opened 2022-10-13
ROX-9156: Secured Cluster Services Helm Chart: make sure to not reference non-existing secrets

mtesseract:
## Description

Within the Secured Cluster Services Helm Chart we are letting service accounts reference image pull secrets irregardless of their existence on the cluster. This is benign but creating noisy error log messages. See https://issues.redhat.com/browse/ROX-9156 for details.

This PR consists of two parts. The first one does two things:
1. it uses the `safeLookup` function for checking if the default secret names (contained in `$defaultSecretNames`) actually exist before adding it to the list of to-be-referenced image pull secrets.
2. it removes the unconditional addition of `$secretResourceName` to the list of to-be-referenced image pull secrets.

The second part takes care of fixing the helmtest based tests for the chart. In order to be able to verify the new behaviour (and to keep the changes to the helmtests conscise) we need a way to simulate a functioning `safeLookup`. Hence I introduced new internal configuration settings for the Helm chart:
```
meta:
  useFakeLookup: # bool
   fakeLookupResources: # map with resource object names as keys (in the form `apiVersion/kind/name`)
```

With this in place the helmtest tests can configure the test environment to contain certain objects (e.g. a `stackrox` secret) which then causes the respective references to be added to the service accounts. The affected tests had to be reorganized somewhat so that one part is executed where these fake resources are in scope, and a different branch is executed where these fake resources are absent.

**For obtaining a clearer overview of what changed in the tests I recommend to use the diff split view with spaces hidden.**

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/image/templates/helm/shared/templates/_image-pull-secrets.tpl
+++ b/image/templates/helm/shared/templates/_image-pull-secrets.tpl
@@ -49,7 +49,15 @@
   {{ include "srox.fail" $msg }}
 {{ end }}
 
-{{ $imagePullSecretNames = concat (append $imagePullSecretNames $secretResourceName) $defaultSecretNames | uniq | sortAlpha }}
+{{ range $defaultSecretName := $defaultSecretNames }}
+  {{ $secret := dict }}
+  {{ include "srox.safeLookup" (list $ $secret "v1" "Secret" $namespace $defaultSecretName) }}
+  {{ if $secret.result }}
+    {{ $imagePullSecretNames = append $imagePullSecretNames $defaultSecretName }}
+  {{ end }}
+{{ end }}
+
+{{ $imagePullSecretNames = $imagePullSecretNames | uniq | sortAlpha }}
 {{ $_ := set $imagePullSecrets "_names" $imagePullSecretNames }}
 {{ $_ := set $imagePullSecrets "_creds" $imagePullCreds }}
 
--- a/image/templates/helm/stackrox-secured-cluster/templates/_lookup.tpl
+++ b/image/templates/helm/stackrox-secured-cluster/templates/_lookup.tpl
@@ -20,6 +20,14 @@
   {{ end }}
   {{ include "srox._doLookup" . }}
   {{ $_ := set $out "reliable" $._rox._state.lookupWorks }}
+{{ else if $._rox.meta.useFakeLookup }}
+  {{ $apiVersion := index . 2 }}
+  {{ $kind := index . 3 }}
+  {{ $resource := index . 4 }}
+  {{ $resourceRef := printf "%s/%s/%s" $apiVersion $kind $resource }}
+  {{ $fakeObj := get $._rox.meta.fakeLookupResources $resourceRef }}
+  {{ $_ := set $out "result" $fakeObj }}
+  {{ $_ := set $out "reliable" true }}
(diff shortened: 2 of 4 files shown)


misberner (changes requested):
The `fakeLookupResources` solution is nifty .. but I'm not sure if I like it.

IMHO, this would conceptually fit in much better at the helm test level. In Helm test, we already configure which server version the chart sees, and which resource types are available (which in effect means: which CRD objects exist on the server). It would only seem natural to also allow specifying the presence of objects. I can imagine that injecting this into Helm can be challenging (IIRC we _don't_ use the Helm action client for that reason - it's pretty messy), but if it were possible, `fake.NewSimpleClientset` from `k8s.io/client-go/kubernetes/fake` would be a great fit for this.

Another concern I have is: this can break existing users. If somebody relies on these default secrets to be included, and is, for example, using a CD pipeline like ArgoCD in which lookup isn't available, things will suddenly break. I don't know how likely or not that is, but it is a risk, and typically we have sought to avoid cases like this.

I guess a middle ground would be to _omit_ image pull secrets only if `lookup` is reliable. That then leads to another error: we don't have `lookup` in the operator, in order to make the chart rendering a function of the translated values only (modulo the revision counter). Therefore, even if the default secrets are present and the operator is used, they would not be inserted. We could write a custom reconciler to detect and insert them into the translated values, but.. meh.

porridge:
I took a look, and I think the only conclusion I can make is that I still don't understand how this was nor is supposed to work. For example:

## what is the role of the `configureImagePullSecrets` "function"?

I can see it:
- populates the `_dockerAuths` field of the passed object (which is then used by the templates to create the `stackrox`, `secured-cluster-services-collector` and `secured-cluster-services-main` secrets).
- it also populates the `_names` field, which is subsequently used in the templates to fill in the `imagePullSecrets` structs of the various `ServiceAccount` resources created by the charts
- however it also seems to populate the `_creds` field, which AFAICT is completely unused?
- are there other (side) effects that I missed?

## What is the semantic difference between $secretResourceName and $defaultSecretNames?

Looking at the call sites, the `stackrox` name is present in most cases, except for the non-scanner-related call in secured cluster chart, where the `secured-cluster-services-main` name is passed as the `$secretResourceName`. AFAICT in the old implementation, all these were joined into a single list, contrary to the documentation which claimed that `$secured-cluster-services-main` is only used if `.username` is set

## Are producing or consuming the secrets, and which ones?

As mentioned above, the `stackrox` secret is being created by the charts. And yet your change checks to see if it's present, which seems to create a catch-22 situation, at least for the initial installation. I.e. the secret does not exist initially, so the function omits its name in `_names`, which ends up empty, therefore `_dockerAuths` ends up empty, leading to no secret being created at all? Just a hypothesis, but the artifacts produced by the failing e2e tests seem to support. In the operator one is hard to see due to https://issues.redhat.com/browse/ROX-12440 but the other ones seem consistent with this hypothesis.


To sum up, to my small mind this just seems like a mess, so I'd suggest moving towards less magic rather than more - fix the logic to match the documented behavior, and add a warning to release notes if need be.

porridge:
Superseded by https://github.com/stackrox/stackrox/pull/9052

PR #3419, opened 2022-10-13
Use upstream proto

janisz:
## Description

Let's remove private fork and bump to latest release.

## Testing Performed

CI


misberner:
This won't work as we depend on custom features

janisz:
@misberner what features?

janisz:
Oh, I see we are not on master but on `-rox` branch. Why we are using 1.2.2 then and not 1.3.2?

PR #3423, opened 2022-10-13
Remove the last use of genny-style codegeneration by moving sliceutils to generics

misberner:
## Description

Title says it all. Bye-bye, genny, you've served us well, but your time with us has come to an end.

## Checklist

## Testing Performed

- CI
- Adapted unit tests


pkg/sliceutils/concat.go
```diff
@@ -0,0 +1,15 @@
+package sliceutils
+
+// Concat concatenates slices, returning a slice with newly allocated backing storage of the exact
+// size.
+func Concat[T any](slices ...[]T) []T {
+	length := 0
```
RTann, line 6: nit: I think usually `var length int` is preferred in our codebase

pkg/sliceutils/diff.go
```diff
@@ -0,0 +1,43 @@
+package sliceutils
+
+// Diff returns, given two slices a and b sorted according to lessFunc, a slice of the elements occurring in a and b
+// only, respectively.
+func Diff[T any](a, b []T, lessFunc func(a, b T) bool) (aOnly, bOnly []T) {
+	i, j := 0, 0
```
RTann, line 6: similarly, here. I feel like `var i, j int` is usually preferred in our codebase

pkg/sliceutils/diff.go
```diff
@@ -0,0 +1,43 @@
...
+	return
+}
+
+// Difference returns the array of elements in the first slice that aren't in the second slice
+func Difference[T comparable](slice1, slice2 []T) []T {
+	if len(slice1) == 0 || len(slice2) == 0 {
+		return slice1
+	}
+
+	blockedElems := make(map[T]struct{}, len(slice2))
```
RTann, line 31: do we want to use our fancy, new, generic `set.Set`?

pkg/sliceutils/diff.go
```diff
@@ -0,0 +1,43 @@
...
+			j++
+		}
+	}
+
+	aOnly = append(aOnly, a[i:]...)
+	bOnly = append(bOnly, b[j:]...)
+	return
+}
+
+// Difference returns the array of elements in the first slice that aren't in the second slice
```
RTann, line 25: nit: returns the slice?

pkg/sliceutils/equal.go
```diff
@@ -0,0 +1,14 @@
+package sliceutils
+
+// Equal checks if the two given slices are equal.
+func Equal[T comparable](a, b []T) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i, aElem := range a {
```
RTann, line 8: maybe just 
```
for i := range a {
        if a[i] != b[i] { ... }
}
```

pkg/sliceutils/equal.go
```diff
@@ -0,0 +1,14 @@
+package sliceutils
+
+// Equal checks if the two given slices are equal.
+func Equal[T comparable](a, b []T) bool {
```
RTann, line 4: Another function does `slice1` and `slice2`. Maybe we should converge on a standard?

sensor/common/externalsrcs/handler.go
```diff
@@ -156,8 +156,8 @@ func normalizeNetworkList(listProto *sensor.IPNetworkList) {
 }
 
 func networkListsEqual(a, b *sensor.IPNetworkList) bool {
-	return sliceutils.ByteEqual(a.GetIpv4Networks(), b.GetIpv4Networks()) &&
-		sliceutils.ByteEqual(a.GetIpv6Networks(), b.GetIpv6Networks())
+	return bytes.Equal(a.GetIpv4Networks(), b.GetIpv4Networks()) &&
```
RTann, line 159: kind of interesting `[]byte` and `string` are semantically the same, but `[]byte` is not `comparable`. It makes sense, but Go already special cases some things (like allowing `==` for slices when comparing to `nil`), so I'm surprised it didn't special case `==` for `[]byte`

go.mod
```diff
@@ -130,6 +129,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require golang.org/x/exp v0.0.0-20220823124025-807a23277127
```
RTann, line 132: how come this is separate?

pkg/sliceutils/diff.go
```diff
@@ -0,0 +1,43 @@
+package sliceutils
+
+// Diff returns, given two slices a and b sorted according to lessFunc, a slice of the elements occurring in a and b
+// only, respectively.
+func Diff[T any](a, b []T, lessFunc func(a, b T) bool) (aOnly, bOnly []T) {
```
RTann, line 5: `Diff` vs `Difference` can be kind of confusing to differentiate without reading the doc

RTann (commented):
a couple of nits otherwise LGTM. Maybe next step is to put these in a separate repository so these can be used without depending on the `stackrox` repo? 😄

RTann (approved):
forgot to press the approve button

go.mod
```diff
@@ -277,6 +278,7 @@ require (
 	github.com/mattn/go-isatty v0.0.16 // indirect
 	github.com/mattn/go-runewidth v0.0.13 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
+	github.com/mauricelam/genny v0.0.0-20190320071652-0800202903e5 // indirect
```
0x656b694d, line 281: Hello genny.

0x656b694d (approved):
Nice

Makefile
```diff
@@ -143,11 +143,6 @@ $(MOCKGEN_BIN): deps
 	@echo "+ $@"
 	go install github.com/golang/mock/mockgen
 
-GENNY_BIN := $(GOBIN)/genny
```
dhaus67, line 146: There are still some references in the Makefile for `GENNY_BIN` within `go-generated-srcs` in line 314.

dhaus67 (changes requested):

sensor/common/externalsrcs/handler.go
misberner: Well, they're not really semantically the same: a `string` is immutable whereas a `[]byte` is mutable. They are convertible into each other, but that's about it. A `[]byte` is a reference type and a `string` is not, and I think the expectation generally is that `==` on a reference type should check reference equality (or be forbidden), not referenced object equality.

Comparison `== nil` is a different story altogether; a bunch of otherwise incomparable types (`func`s etc.) are comparable against `nil`. There's a [proposal](https://github.com/golang/go/issues/53656) to add a nullable constraint for types that are comparable against `nil`, but not necessarily against each other.

pkg/sliceutils/concat.go
misberner: done

pkg/sliceutils/diff.go
misberner: done

pkg/sliceutils/diff.go
misberner: No, I don't think so, mostly because I think it isn't obvious that `sliceutils` is less fundamental than `set`. We can only add an import in one direction, and I consider `sliceutils` as lower-level.

pkg/sliceutils/diff.go
misberner: done

pkg/sliceutils/equal.go
misberner: done

pkg/sliceutils/equal.go
misberner: I used `slice1` and `slice2` here and also in `Difference`.

go.mod
misberner: no idea, merged it into the main `require` block

pkg/sliceutils/diff.go
misberner: Yeah, I renamed `Difference` to `Without`, which reads quite naturally IMHO:
```
tlsConf.NextProtos = sliceutils.Without(tlsConf.NextProtos, []string{"h2", alpn.PureGRPCALPNString})
```

Makefile
misberner: thx, fixed

go.mod
misberner: 👋

dhaus67 (approved):

PR #3429, opened 2022-10-14
Add more CI labels to dependabot PRs

msugakov:
## Description

Add more CI labels to dependabot PRs so that our CI runs more comprehensive tests for dependencies updates.
Because PRs like #3426 may break some things that won't be visible in just BAT tests or come up on other *KS platforms.

I don't know if docker updates are still working, but will not investigate this time.

## Checklist

Won't do any of these:

## Testing Performed

* CI will do the testing.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -32,7 +32,6 @@ updates:
       - "ci-eks-tests"
       - "ci-no-api-tests"
       - "ci-no-ui-tests"
-      - "ci-openshift-4-tests"
       - "dependencies"
     reviewers:
       - "stackrox/backend-dep-updaters"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yaml
```diff
@@ -81,6 +86,12 @@ updates:
       interval: 'weekly'
       day: 'wednesday'
     open-pull-requests-limit: 1
+    labels:
```
gavin-stackrox, line 89: The only `ci-` label that has any impact in this list now is `ci-all-qa-tests`. To get say the EKS flavor to run a comment of `/test eks-qa-e2e-tests` would be required. (AKS is not operational at the moment.) Another reliable flavor to add would be `/test openshift-penultimate-qa-e2e-tests` if dependabot can comment like that.
I will go through the list of labels and remove the `ci-` flavors that no longer have any function.

gavin-stackrox (changes requested):

.github/dependabot.yaml
msugakov: I cleaned up labels but wasn't able to spot a way to instruct dependabot to also write comments on PR. Checked this https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

gavin-stackrox (approved):

PR #3433, opened 2022-10-14
[ROX-12945] Remove not required inner joins

mtodor:
## Description

We have identified that, in some cases, we have 2 inner joins between 3 tables. But one table is joined over the same column to both other tables.

i.e.
```
t1 INNER JOIN t2 ON t1.id = t2.some_column
t2 INNER JOIN t3 ON t2.some_column = t3.id
```
In this case, t2 is not essential and slows down the query because index scans or other similar actions are performed.

With this PR, the join chain will be changed, and t2 will be removed. We will have:
```
t1 INNER JOIN t3 ON t1.id = t3.id
```

## Checklist

## Testing Performed

1. Added unit tests
2. Additional manual testing in Stackrox UI
3. Testing with performance testing scripts provided in another PR: https://github.com/stackrox/stackrox/blob/a737674586819f088a5ee8a8954395771b0a0365/test_sac_perf.py


mtodor:
* master
  * **PR #3433** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3433" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3433?utm_source=stack-comment).

pkg/search/postgres/joins.go
```diff
@@ -185,5 +190,77 @@ func getJoinsAndFields(src *walker.Schema, q *v1.Query) ([]innerJoin, map[string
...
+// a INNER JOIN b ON a.id = b.same_column
+// b INNER JOIN c ON b.same_column = c.id
+// If table b is not used in any other way, we can remove it from the join chain.
+// Outcome: a INNER JOIN c ON a.id = c.id
+func (j *joinTreeNode) removeUnnecessaryRelations(requiredFields map[string]searchFieldMetadata) {
+	if j == nil {
+		return
+	}
+
+	requiredTables := make(map[string]struct{})
```
janisz, line 209: I think we have a string set in our codebase

pkg/search/postgres/joins.go
```diff
@@ -185,5 +190,77 @@ func getJoinsAndFields(src *walker.Schema, q *v1.Query) ([]innerJoin, map[string
...
+	for child, columnPairs := range j.children {
+		child.removeUnnecessaryRelations(requiredFields)
+
+		if _, isRequired := requiredTables[child.currNode.Table]; isRequired {
+			rootChildren[child] = columnPairs
+
+			continue
+		}
+
+		childColumns := make(map[string]string)
```
janisz, line 224:
```suggestion
		childColumns := make(map[string]string, len(columnPairs))
```

pkg/search/postgres/joins.go
```diff
@@ -185,5 +190,77 @@ func getJoinsAndFields(src *walker.Schema, q *v1.Query) ([]innerJoin, map[string
...
+
+		childChildren := make(map[*joinTreeNode][]walker.ColumnNamePair)
+		for childChild, childColumnPairs := range child.children {
+			if len(columnPairs) != len(childColumnPairs) {
+				childChildren[childChild] = childColumnPairs
+
+				continue
+			}
+
+			rootColumnPairs := make([]walker.ColumnNamePair, 0)
```
janisz, line 237: microoptimisation
```suggestion
			rootColumnPairs := make([]walker.ColumnNamePair, len(childColumnPairs))
```

pkg/search/postgres/joins_test.go
```diff
@@ -0,0 +1,613 @@
+//go:build sql_integration
+// +build sql_integration
```
janisz, line 2: Why do we need this?

pkg/search/postgres/joins_test.go
```diff
@@ -0,0 +1,613 @@
...
+			// We have to sort before using DeepEqual. Otherwise, results could defer.
+			// It's sufficient to sort by joined tables.
+			sort.SliceStable(expectedInnerJoins, func(i, j int) bool {
+				return expectedInnerJoins[i].rightTable+expectedInnerJoins[i].leftTable < expectedInnerJoins[j].rightTable+expectedInnerJoins[j].leftTable
+			})
+			sort.SliceStable(innerJoins, func(i, j int) bool {
+				return innerJoins[i].rightTable+innerJoins[i].leftTable < innerJoins[j].rightTable+innerJoins[j].leftTable
+			})
+
+			assert.True(t, reflect.DeepEqual(expectedInnerJoins, innerJoins))
```
janisz, line 610:
```suggestion
			assert.EqualValues(t, expectedInnerJoins, innerJoins)
```

pkg/search/postgres/joins_test.go
```diff
@@ -0,0 +1,613 @@
...
+
+func getTestData() map[string]testSet {
+	data := map[string]testSet{
+		"nil join tree": {
+			Root:         nil,
+			ExpectedRoot: nil,
+		},
+		"single table": {
+			Root: &joinTreeNode{
+				currNode: &walker.Schema{
```
janisz, line 29: How about create helper function to create objects so this will be easier to read?
```go
func t(id string) *walker.Schema {
	return &walker.Schema{
		Table: id,
	}
}
```

janisz:
```go
func node(tid string) *joinTreeNode {
	return &joinTreeNode{
		currNode: t(tid),
	}
}

func t(id string) *walker.Schema {
	return &walker.Schema{
		Table: "t" + id,
	}
}

func pairs(this, other string) []walker.ColumnNamePair {
	return []walker.ColumnNamePair{{
		ColumnNameInThisSchema:  this,
		ColumnNameInOtherSchema: other,
	}}
}

func pair(this, other string) walker.ColumnNamePair {
	return walker.ColumnNamePair{
		ColumnNameInThisSchema:  this,
		ColumnNameInOtherSchema: other,
	}
}
```
If you use above functions then code will be much easier to read
e.g.
```go
			ExpectedRoot: &joinTreeNode{
				currNode: t("1"),
				children: map[*joinTreeNode][]walker.ColumnNamePair{
					{
						currNode: t("2"),
						children: map[*joinTreeNode][]walker.ColumnNamePair{
							node("3"): pairs("t2_c1", "t3_c1"),
						},
					}: pairs("t1_c1", "t2_c1"),
				},
			},
```

pkg/search/postgres/joins.go
mtodor: This could break logic. Because I'm using `len(rootColumnPairs)` afterward to check that all columns are processed.
I'll change it to `0, len(childColumnPairs)` - because it will never grow bigger.

pkg/search/postgres/joins_test.go
mtodor: Tbh. I'm not sure. I have seen that tag used in other tests in this package. I assume it's related to group `go-postgres-unit-tests`.

pkg/search/postgres/joins_test.go
mtodor: We still have to sort. The reason for that is that "children" inside joinTreeNode is a map where a pointer is used as a key. Because of that order in a loop (range) can be different.

I have updated the comment and changed it to `assert.EqualValues` - it gives better output.

pkg/search/postgres/joins_test.go
janisz: But you are not calling external database. I think we can remove it.

janisz (changes requested):

janisz (approved):

PR #3436, opened 2022-10-14
ROX-12858: Evaluate usages of query.Data in logs and remove on releases to avoid logging secrets

theencee:
## Description

See title for details.

Additionally add a check that will be run in PR and merge CI jobs to check that no one logs query.Data. It might be overkill, but it won't hurt.

## Checklist

## Testing Performed

Tested manually on a debug build and saw that logs were generated. To test on release, will need to merge first and wait for nightly build.

### DIFF
--- a/scripts/check-log-query-data.sh
+++ b/scripts/check-log-query-data.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+regex="log\.(Error|Debug|Info|Warn|Panic|Fatal)f*\(.*(query.Data).*\)"
+
+tmpfile="$(mktemp)"
+function delete_tmpfile() {
+	rm -f "$tmpfile"
+}
+trap delete_tmpfile EXIT
+
+git grep -inE "$regex" >"$tmpfile"
+
+if [[ ! -s "$tmpfile" ]]; then
+	exit 0
+fi
+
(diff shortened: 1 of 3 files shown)


scripts/ci/jobs/check-log-query-data.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
theencee, line 2: This won't run right now. There's a PR that still needs to made in openshift/release

dashrews78 (approved):

scripts/check-log-query-data.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+regex="log\.(Error|Debug|Info|Warn|Panic|Fatal)f*\(.*(query.Data).*\)"
```
rhybrillou, line 3: Note: this will not catch cases where the log text is long and query.Data is on a second line to comply with style line length. This will nevertheless limit the risk of logging sensitive query data.

rhybrillou (approved):

scripts/check-log-query-data.sh
theencee: Good catch. Given that this is not urgent, I'll address it separately

PR #3437, opened 2022-10-14
CI: Explain the slack/log disconnect

gavin-stackrox:
## Description

The fact that the slack message appears before the log is ready can be confusing. A longer term enhancement might be to send these with a delay: https://api.slack.com/messaging/scheduling.

## Checklist

## Testing Performed

CI is sufficient


gavin-stackrox:
On second thoughts, the delay, at least in the case we discussed today was a bug (#3440), so I'll fix the bug and if delayed logs reoccurs we can revisit the :turtle: message.

PR #3438, opened 2022-10-14
ROX-11931: Convert junit failure artifacts to Slack attachments

BradLugo:
## Description

Converts Junit artifacts that contain failures and failure messages to Slack message attachments.

Related to https://github.com/stackrox/junit-parse/pull/1

## Testing Performed

- Test CI: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3438/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1588220418928414720 and successful Slack messages:https://srox.slack.com/archives/C740BUV25/p1667500352946239
- Tested with various artifact directories from failed Prow jobs


scripts/ci/lib.sh
```diff
@@ -1368,6 +1373,22 @@ EOT
...
+junit2slack() {
+    info "Converting junit failures to slack attachments"
+
+    if [[ -z "${ARTIFACT_DIR}" ]]; then
+        info "Warning: junit2slack() requires an ARTIFACT_DIR"
+        return
+    fi
+
+    local junit_file_names
+    junit_file_names=$(find "${ARTIFACT_DIR}" -type f -name 'junit-*.xml' | xargs)
```
gavin-stackrox, line 1385: this might need some extended bash globbing to get to junit xml in sub dirs. e.g. with this one:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-local-roxctl-tests/1582096454128766976
there is a report.xml in 
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-local-roxctl-tests/1582096454128766976/artifacts/merge-local-roxctl-tests/stackrox-initial/artifacts/junit-reports/
I don't think that `find` will find it.

scripts/ci/lib.sh
```diff
@@ -1290,6 +1290,11 @@ __EOM__
 }
 '
 
+    local slack_attachments = junit2slack
+    if [[ "${slack_attachments}" != "" ]]; then
```
gavin-stackrox, line 1294:
```suggestion
    if [[ -n "${slack_attachments}" ]]; then
```

gavin-stackrox (commented):
In general, LGTM!, let me know when you are ready for a final review.

scripts/ci/lib.sh
BradLugo: This actually brings up an interesting point (along with your suggestion on how to test these changes); at least from what it seems in the failed job [#1582162936795762688](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-local-roxctl-tests/1582162936795762688), spyglass only looks at `junit_operator.xml`. In fact, when I tried to parse `merge-local-roxctl-tests/stackrox-initial/artifacts/junit-reports/report.xml`, it threw an error which leads me to believe we're not properly formatting some of these reports. That said, I haven't dug into that error yet. It could actually be the case that spyglass isn't surfacing useful error messages.

So then, should we only look at `report.xml` if present for now? Or spend the time to debug these (seemingly) malformed junit reports and try to parse all XML files? Interested in your take on this.

gavin-stackrox: > So then, should we only look at report.xml if present for now?

Different jobs/tests name the JUnit differently. Sometimes single files, sometimes directories of XML (or maybe a directory with a single file). There is a wrapper script `store_test_results` to give you an idea of what you are dealing with.

[#1582162936795762688](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-local-roxctl-tests/1582162936795762688) is a special but interesting case. The prow/spyglass Go code must be hitting the same error that you are and that is why it is not displayed in the UI. For errors we should do the same, silently drop them from the slack message. But it would be good to look at fixing the XML when it is dropped and/or fixing the parser in slack2junit and/or prow.

There are quite a few cases where you'll see prow/spyglass parse the Junit we (stackrox tests) create OK: 
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-gke-qa-e2e-tests/1582357585149825024
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-master-merge-openshift-penultimate-qa-e2e-tests/1580969317434920960

BradLugo: Awesome, appreciate the feedback. I'll take a look at what's causing the malformed XML after this PR is merged.

---

Going back to the bash glob:
> this might need some extended bash globbing to get to junit xml in sub dirs

I thought `find` would do this automatically. From the POSIX manual:
> The find utility shall recursively descend the directory hierarchy from each file specified by path

gavin-stackrox: not with GNU find 4.9.0:
```
$ mkdir -p ttt/junit-crud
$ touch ttt/junit-crud/report.xml
$ find ttt -type f -name 'junit-*.xml'
$
```

gavin-stackrox: `**` will work though and does not seem to require `shopt extglob` (at least for my shell which is closer to the CI OS then say MacOS).
```
$ x=ttt/**/*.xml
$ echo $x
ttt/junit-crud/report-2.xml ttt/junit-crud/report.xml
$
```

BradLugo:
@gavin-stackrox It seems that the unit test file is causing the CI to break. It seems to be breaking from the `| go list` pipe in the `go-unit-tests` make target. I think it's looking for a Go module from the master branch, but it's not there yet. Any thoughts or ideas on how to resolve this? I can ask the other engineers as well

scripts/ci/junit2slack/main.go
```diff
@@ -0,0 +1,152 @@
+package main
+
+import (
```
RTann, line 3: the imports are out of order, which I think `style-checks` caught. You can run `make style` or `quickstyle`, if you have the https://github.com/stackrox/workflow scripts

scripts/ci/junit2slack/main.go
```diff
@@ -0,0 +1,152 @@
...
+
+			junitSuites, err := junit.Parse(data)
+			if err != nil {
+				log.Printf("error while parsing junit suites in %s: %s", fileName, err)
+				continue
+			}
+			junitFiles = append(junitFiles, junitSuites)
+
+		} else if errors.Is(err, os.ErrNotExist) {
+			log.Printf("%s doesn't exist: %s", fileName, err)
```
RTann, line 38: one annoying thing about `Printf` if it does not add `\n`, so we might want to add that ourselves

scripts/ci/junit2slack/main.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+
+				titleTextBlock := slack.NewTextBlockObject("plain_text", title, false, false)
+				titleSectionBlock := slack.NewSectionBlock(titleTextBlock, nil, nil)
+				failedTestsBlocks = append(failedTestsBlocks, titleSectionBlock)
+
+				failureMessage := result.Failure.Message
+
+				// If there's no failure message, we'll use a different message (this shouldn't be the usual case)
+				if len(failureMessage) <= 0 {
```
RTann, line 90: you can just check `== 0`

scripts/ci/junit2slack/main.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	b, err := json.Marshal(slackMsg)
+	if err != nil {
+		log.Printf("error while marshaling Slack message to json: %s", err)
+	}
+	fmt.Println(string(b))
+}
+
+func convertJunitToSlack(junitFiles []*junit.Suites) []slack.Attachment {
```
RTann, line 59: you can also do `junitFiles ...*junit.Suites`. This way when you test with just one file, you can just do `convertJunitToSlack(suites)` where `suites` is of type `*junit.Suites`

scripts/ci/junit2slack/main.go
```diff
@@ -0,0 +1,152 @@
...
+				// We've reached the message limit we want. We need to break out of all the loops
+				if len(attachments) <= 3 {
+					goto pushFinalSlackAttachments
+				}
+			}
+		}
+	}
+
+pushFinalSlackAttachments:
+	if failedTestsBlocks == nil || len(failedTestsBlocks) <= 0 {
```
RTann, line 135: you can just check `len(failedTestsBlocks) == 0`

scripts/ci/junit2slack/main.go
BradLugo: Looks like each `Logger` method will end up using [Logger.Output](https://pkg.go.dev/log#Logger.Output) which says:

> A newline is appended if the last character of s is not already a newline.

That being said, I'm down to change this if we do something similar elsewhere in the repo

scripts/ci/junit2slack/main_test.go
```diff
@@ -0,0 +1,32 @@
...
...CVE@eca16279, CSVTest$CVE@e5511974, CSVTest$CVE@e55f30f5, CSVTest$CVE@32589016, CSVTest$CVE@29b612f9, CSVTest$CVE@74f75429, CSVTest$CVE@54f6a8bb, CSVTest$CVE@f5e08943, CSVTest$CVE@a58c86e5, CSVTest$CVE@6edadf49, CSVTest$CVE@3e277299, CSVTest$CVE@3c8a1e51, CSVTest$CVE@4c136df1, CSVTest$CVE@49def44, CSVTest$CVE@4ac06c5, CSVTest$CVE@4ba1e46, CSVTest$CVE@34ba2dae, CSVTest$CVE@65fe59ad, CSVTest$CVE@69ae8070, CSVTest$CVE@6ae48586, CSVTest$CVE@6af29d07, CSVTest$CVE@6b00b488, CSVTest$CVE@8de64d7d, CSVTest$CVE@cb1352ff, CSVTest$CVE@cc8fcd9a, CSVTest$CVE@30ed073a, CSVTest$CVE@d17ed81f, CSVTest$CVE@d9c09db5, CSVTest$CVE@d9ceb536, CSVTest$CVE@6aa11951, CSVTest$CVE@68c3e72f, CSVTest$CVE@98615211, CSVTest$CVE@39df66b1, CSVTest$CVE@f2bf931a, CSVTest$CVE@27c1992f, CSVTest$CVE@c5c47329, CSVTest$CVE@acf47c90, CSVTest$CVE@d576e4c5, CSVTest$CVE@9a12044e, CSVTest$CVE@9a4a6252, CSVTest$CVE@9a5879d3, CSVTest$CVE@9a669154, CSVTest$CVE@9a74a8d5, CSVTest$CVE@bf8072f6, CSVTest$CVE@bf8e8a77, CSVTest$CVE@bf9ca1f8, CSVTest$CVE@3fa5bdd8, CSVTest$CVE@71b61e0c, CSVTest$CVE@3fdde12d, CSVTest$CVE@3febf8ae, CSVTest$CVE@1a2a6ff7, CSVTest$CVE@a4a453df, CSVTest$CVE@4bc5f8f8, CSVTest$CVE@5423ed90, CSVTest$CVE@579"}}]}]`)
```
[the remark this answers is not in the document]
BradLugo: When I tried embedding the json file, it would add an extra `.` character to the end causing the unit test to fail. I'm not quite sure what was going on there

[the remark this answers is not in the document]
BradLugo: Turns out it was a newline character

scripts/ci/junit2slack/main.go
misberner: Generally I would prefer `failureMessage == ""` in Go. This is the most direct and has no performance disadvantages over `len(...) == 0`

BradLugo:
Created a new repo to host the go project here: https://github.com/stackrox/junit-parse
Related PR: https://github.com/stackrox/junit-parse/pull/1

scripts/ci/lib.sh
```diff
@@ -1409,6 +1414,26 @@ EOT
     hub-comment -type build -template-file "$tmpfile"
 }
 
+get_junit_parse_cli() {
+    # TODO !! Need to remove this commit hash before merging
+    go install github.com/stackrox/junit-parse@blugo/4e2c784
```
gavin-stackrox, line 1419: Installing in CI can often lead to flakiness. It would be best to install this in the CI image to avoid the `go install` (stackrox/rox-ci-image:stackrox-test). However getting that image updated in CI is a major pain so I would rely on a future update to get the binary present. In the meantime you can `command -v` to check if it is present before installing. Something like: https://github.com/stackrox/stackrox/blob/d0cc2801cdd5373f6256d77aebe69f8cd83ae0c9/tests/upgrade/run.sh#L105

scripts/ci/lib.sh
```diff
@@ -1409,6 +1414,26 @@ EOT
...
+junit2slack() {
+    info "Converting junit failures to slack attachments"
+
+    if [[ -z "${ARTIFACT_DIR}" ]]; then
+        info "Warning: junit2slack() requires an ARTIFACT_DIR"
+        return
+    fi
+
+    local junit_file_names="$ARTIFACT_DIR/**.*.xml"
+    pushd "$SCRIPTS_ROOT/tools/junit2slack" || return
```
gavin-stackrox, line 1431: Is this needed anymore? And I expect the dir is no longer present now that you have moved the go code.

scripts/ci/lib.sh
```diff
@@ -1409,6 +1414,26 @@ EOT
...
+    get_junit_parse_cli
+    junit-parse "$junit_file_names"
```
gavin-stackrox, line 1433: I'd suggest ` || return` following both of these. If there are any errors getting the tool or parsing the binary the plain slack message should still be sent.

scripts/ci/lib.sh
```diff
@@ -1409,6 +1414,26 @@ EOT
...
+    popd
```
gavin-stackrox, line 1434: also delete?

gavin-stackrox (changes requested):

scripts/ci/lib.sh
BradLugo: Circling back to this since I ended up going with `find`; it looks like your example is looking for `junit-*.xml` but the name of the file is `report.xml`.

scripts/ci/lib.sh
```diff
@@ -1296,6 +1296,43 @@ __EOM__
...
+    ]
+  }
+]
+'
+    if [[ -n "${ARTIFACT_DIR}" ]]; then
+        if ! command -v junit-parse >/dev/null 2>&1; then
+            get_junit_parse_cli || exitstatus="$?"
+            if [[ "$exitstatus" == "0" ]]; then
+                local junit_file_names
+                junit_file_names=($(find "${ARTIFACT_DIR}" -type f -name '*.xml' | xargs)) || true
```
gavin-stackrox, line 1322: 👍 to whatever works, we can iterate on it.

gavin-stackrox (commented):
I'm ready to approve once you are done and all the forced failure & test changes are removed.

BradLugo:
Sorry about that - I thought I had squared all that away before re-requesting your review. Should be good to go now

scripts/ci/lib.sh
```diff
@@ -1299,6 +1299,39 @@ __EOM__
...
+        }
+      }
+    ]
+  }
+]
+'
+    if [[ -n "${ARTIFACT_DIR}" ]]; then
+        if ! command -v junit-parse >/dev/null 2>&1; then
+            get_junit_parse_cli || exitstatus="$?"
+            if [[ "$exitstatus" == "0" ]]; then
```
gavin-stackrox, line 1323: I think to handle the case when junit-parse is already in the test image:
```suggestion
            get_junit_parse_cli || true
        fi
        if command -v junit-parse >/dev/null 2>&1; then
```
We'll make a best effort try to get junit-parse and also handle when it is already present and not exit/error in either case.

gavin-stackrox (changes requested):

gavin-stackrox (approved):
LGTM!

BradLugo:
Thanks for all the help everyone!

PR #3406, opened 2022-10-12
ROX-13002: Fix upgrade test hang

gavin-stackrox:
## Description

With OpenShift CI `ci-operator` any dangling process will cause the test job to hang.

## Checklist
- [x] Investigated and inspected CI test results

## Testing Performed

CI is sufficient. `gke-upgrade-tests` fails as expected due to ROX-13001 but not with a hang and timeout. Instead it posts a JUnit failure. https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3406/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1580354374029283328

PR #3440, opened 2022-10-14
CI: Cleanup dangling processes

gavin-stackrox:
## Description

This PR kills processes that tests have left running. These 'dangling' processes cause ci-operator to hang. Which causes timeouts and delayed access to logs.

For example, in the case of a recent gke-upgrade-tests failure when the test is complete there is a sleep left running. This can be seen from the openshift CI console and logs:
```
$ ps -e -O ppid
    PID    PPID S TTY          TIME COMMAND
      1       0 S ?        00:00:01 /tmp/entrypoint-wrapper/entrypoint-wrapper /tools/entrypoint
     20       1 S ?        00:00:00 /tools/entrypoint
   1151       1 S ?        00:00:00 /usr/bin/coreutils --coreutils-prog-shebang=sleep /usr/bin/sleep 3600
   2228       1 Z ?        00:00:00 [kubectl] <defunct>
```
This sleep is from the GKE token refresh process (typically it is 900 but changed to 3600 here to force the issue). The refresh process is terminated by the cluster teardown() but leaves the sleep running. That is easy enough to fix (and is on this PR by trapping the SIGTERM) but it would be useful to have a general purpose process cleaner on exit to avoid playing whackamole with these problems (e.g. another recent case: https://github.com/stackrox/stackrox/pull/3406).

## Checklist

## Testing Performed

CI is sufficient. In the first three test cases we expect the test to fail as it still has the `--license` bug.
- [x] general purpose cleaner kills the sleep and terminates the test in a timely manner. [result](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3440/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1581115615148511232). 
- [x] killing the sleep when the wait is interrupted also should do the same. [result](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3440/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1581316679466487808)
- [x] with both. [result](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3440/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1581332802316537856)
- [x] rebase for a green run

### DIFF
--- a/.openshift-ci/clusters.py
+++ b/.openshift-ci/clusters.py
@@ -77,6 +77,7 @@ class GKECluster:
             time.sleep(60)
 
         if self.refresh_token_cmd is not None:
+            print("Terminating GKE token refresh")
             try:
                 popen_graceful_kill(self.refresh_token_cmd)
             except Exception as err:
--- a/image/fetch-stackrox-data.sh
+++ b/image/fetch-stackrox-data.sh
@@ -22,4 +22,3 @@ fetch_stackrox_data() {
 }
 
 fetch_stackrox_data
-
--- a/image/fetch-stackrox-data.sh
+++ b/image/fetch-stackrox-data.sh
@@ -22,3 +22,4 @@ fetch_stackrox_data() {
 }
 
 fetch_stackrox_data
+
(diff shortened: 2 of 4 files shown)


gavin-stackrox:
@RTann you probably want this for Scanner

RTann (approved):

PR #3444, opened 2022-10-16
Implementation for RuleSelector collection component

dvail:
## Description

This adds the RuleSelector component to the Collection Form. The RuleSelector is the main component used to configure rules against clusters, namespaces, and deployments that result in a collection.

The top level component has three states which causes conditional rendering of secondary components:
- All entities selected -> (no component)
- Select by entity name -> ByNameSelector
- Select by entity label -> ByLabelSelector

The "All entities selected" option functions as a wildcard rule, selecting all entities in the chosen scope.

The ByNameSelector allows a user to filter by entities matching a list of name string values. An entity will be selected if _any_ of the values of this component match.

The ByLabelSelector allows a user to filter by entities containing a matching label. This component allows the user to specify multiple label _values_ for a given key, as well as multiple label keys (with their own list of values). Multiple label keys will be "AND'ed" together when the rule is evaluated.

## Review notes

This is a sizeable PR since the RuleSelector drives the most complicated pieces of the collection structure.

1. The types of the [server response](https://github.com/stackrox/stackrox/pull/3444/files#diff-c483251ef43c9e6ca98e960f9f1a180cb75520211b66347542e8999c4d7b21e3R16) and the [client side representation](https://github.com/stackrox/stackrox/pull/3444/files#diff-2902d0161d4d4122f4b8cee5ed1fbb11d17e575d8d1644a4cd2d0e3afd9b32d5R4) of a collection have been better separated for type safety and cleaner code when working with the data.
2. The translation from server response to client side data is still handled in the [parser](https://github.com/stackrox/stackrox/pull/3444/files#diff-a359f374cc4ae1bd66ea2ed3705cd8dcd41a622b902b294849c13487510de697R34), which received minor updates to support the new changes.
3. The top level reducer was discarded in favor of [Formik](https://github.com/stackrox/stackrox/pull/3444/files#diff-d2c9ca91f4296b6bb0e9282f7247b7f07c1408766275178d70782bb211242aa1R125), with [yup validation](https://github.com/stackrox/stackrox/pull/3444/files#diff-d2c9ca91f4296b6bb0e9282f7247b7f07c1408766275178d70782bb211242aa1R64).
4. The RuleSelector component was moved into its own subdirectory and broken into smaller components for the sake of readability. The intention is that only the[ top level RuleSelector component](https://github.com/stackrox/stackrox/pull/3444/files#diff-d9bab0b8183d717fd284d4063ce8cd084845a1dd4c9446403e0b59509d265d21) will be imported for direct use.
5. For the CollectionForm, it may be easier to read with "Show whitespace" disabled, due to nesting changes in the markup.
6. Other comments inline.

### Follow ups
- Implement autocomplete for rule dropdowns
- Show top level validation errors. Right now the extent of validation is marking the individual fields visually with an "error" icon.
- Create a ticket for validation error when no rules are selected for any of the entity types. This needs to be done after the AttachCollections component is in place, since a value in either is valid for the collection form.
- Fix unmounted component error and re-enable test.

## Checklist

## Testing Performed

Visit the collection page and click the "Create collection" button. A form with three rule selectors, in the "all entities" state should be visible.
image

In one of the RuleSelectors, change the top level dropdown to select "by name". The ByNameSelector should appear with a single blank value, in an invalid state (empty).
image

Enter a value into the typeahead component and select the "Create" option when presented. The value should persist and the validation error should disappear.
image

Click the "Add value" button multiple times. A single additional blank field should be added. Additional fields cannot be added if any of the currently visible fields are blank.
image
image

Click the delete icon next to a field to remove this value. Removal of all fields should place the component back into the "All entities" state.
image
image
image

Repeat the above process, but instead of deleting one-by-one, manually choose the "All entities" state. This should clear the user-entered options and return the component to its default state.

In one of the RuleSelectors, change the component to the ByLabel state. This should present a set of controls with a blank label key and blank label value dropdown, both in the invalid state.
image

Entering a value into the dropdowns will clear the invalid states.
image
image

Note that the values must conform to valid k8s label key/values. Invalid characters, lengths, etc. will place the fields into an invalid state.
image

Clicking the "Add value" button will allow the addition of multiple values under a single key, with a maximum of a single blank row allowed.
image
image

Clicking the "Add label rule" button will allow the addition of an entirely new label rule. Like the "Add value" button, an additional rule will only be added if all keys and labels in existing rules are non-empty.
image
image

Label values can be deleted one by one by clicking the trash icon. If all values for a single rule are deleted, the entire rule will be deleted as well. If all rules are deleted, the component will revert to the "All entities" state.
image
image
image
image

image
image


dvail:
* master
  * **PR #3444** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3444" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3444?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -116,15 +122,30 @@ function CollectionForm({
...
+            embeddedCollectionIds: yup.array(yup.string()),
+            selectorRules: yup.object().shape({
+                Deployment: yupSelectorRuleObject(),
+                Namespace: yupSelectorRuleObject(),
+                Cluster: yupSelectorRuleObject(),
+            }),
+        }),
+    });
+
+    // eslint-disable-next-line no-console
```
dvail, line 140: - [ ] TODO, delete this before merge

ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
```diff
@@ -59,7 +59,7 @@ function CollectionsFormPage({
                 {/* TODO - Handle UI for parse errors */}
             </>
         );
-    } else if (loading && !initialData) {
+    } else if (loading) {
```
dvail, line 62: We need to hold off on rendering the form until loading is complete, even if we have a default value for `initialData`, since Formik will use this default data as state and ignore the request result once completed.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+            resourceSelector = newSelector;
+        }
+
+        render(<DeploymentRuleSelector defaultSelector={resourceSelector} onChange={onChange} />);
+
+        expect(await screen.findByText('All deployments')).toBeInTheDocument();
+    });
+
+    // eslint-disable-next-line jest/no-commented-out-tests
+    /*
```
dvail, line 45: There is a `console.error` due to a state update on an unmounted component that occurs when:

1. The collection form is visited with initial data (view/edit/clone states)
2. You click any value dropdown in one of the ByNameSelector or ByLabelSelector components and edit the value.
3. You click anywhere else on the page.

This appears to only happen the first time this occurs in _any_ of these components per page load. I've spent some time on this but have not been able to figure the issue out, short of doing a binary search of the page to limit this to a minimal reproduction. It seems related to updating state via Formik, because rendering an additional `<AutoCompleteSelector>` component with its own state management does not produce the error.

I left this test in the PR, commented out, since the logic passes when the global Spy is disabled. As I follow up I can work at this more, find a way to disable the global Spy, or delete this test entirely if we want to rely only on Cypress tests for this feature.

The stack trace ends in the PF `<Select>` component, but without a usable line number.

This doesn't appear to impact UI functionality, but causes tests to fail due to [our code forcing tests that call console.error to fail](https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/setupTests.js#L17).

ui/apps/platform/src/Containers/Collections/types.ts
```diff
@@ -1,46 +1,76 @@
...
-/**
- * This type extracts the `fieldName` property from the individual rules and groups them
- * using a single `field` property due to the UI only supporting a single field per entity
- * type in collection rules.
- */
-export type ScopedResourceSelector = {
-    field: SupportedSelectorField;
-    rules: Omit<DisjunctionSelectorRule, 'fieldName'>[];
-} | null;
+export type ByNameResourceSelector = {
```
dvail, line 50: Note the difference between `rule` and `rules` in these two types. The UX MVP only allows a single "rule" when selecting by name (i.e. the entity matches any of a list of values), while the MVP supports multiple rules when selecting by label (i.e. the entity matches _all_ of a list of label key/values, with unique label keys being grouped together and matching on _any_ of their respective values.)

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.test.tsx
dvail: Aha, this is because of the `key` prop on the outer `<Flex>` component. The `value` of the the rule is used as the `key`, which is changed when editing. This causes the entire component to unmount and a new one to mount in its place, which is throwing the error above.

There isn't currently a stable key for the items that are being iterated over, so I'll look for a workaround.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -116,15 +122,30 @@ function CollectionForm({
     const [isDeleting, setIsDeleting] = useState(false);
     const { toasts, addToast, removeToast } = useToasts();
 
+    const { values, isValid, errors, setFieldValue } = useFormik({
+        initialValues: initialData,
+        onSubmit: () => {},
+        validationSchema: yup.object({
+            name: yup.string().trim().required(),
+            description: yup.string().required(),
```
pedrottimark, line 130: Is `description` required? For example, it is optional with `yup.string()` in Access Control forms.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -268,44 +294,80 @@ function CollectionForm({
...
+                                <Flex
+                                    className="pf-u-background-color-100 pf-u-p-lg"
+                                    direction={{ default: 'column' }}
+                                    spaceItems={{ default: 'spaceItemsMd' }}
+                                >
+                                    <Title className="pf-u-mb-xs" headingLevel="h2">
+                                        Add new collection rules
+                                    </Title>
+                                    <p>
+                                        Target deployments via selector rules. Regular expressions
```
pedrottimark, line 316: Off the subject of this pull request, here is a plain alternative to consider:

Select collections of deployments via rules. You can use regular expressions (RE2).

* Target is more often a noun than a verb. For example, online dictionary, thesaurus, and translate.
* The distinction between selector rules and fields is not clear.

dvail: Alan has suggested "Select deployments via rules. You can use regular expressions (RE2 syntax).", which removes the usage of "collections" in the definition of what a collection is, and is more explicit about RE2 _syntax_.

He also mentioned linking the text to the RE2 docs page, or having more info on tooltip. My first thought was to do the former, but I couldn't find a single other instance of an external link in the app. Is this an intentional choice, possibly due to offline environments?

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -268,44 +284,80 @@ function CollectionForm({
...
+                                    </Title>
+                                    <p>
+                                        Select deployments via rules. You can use regular
+                                        expressions (RE2 syntax).
+                                    </p>
+                                    <Divider className="pf-u-mb-lg" component="div" />
+                                    <RuleSelector
+                                        entityType="Deployment"
+                                        scopedResourceSelector={values.selectorRules.Deployment}
+                                        handleChange={onRuleSelectorChange}
```
pedrottimark, line 313: Possible follow up to find a more consistent naming convention:
* `selectorRules` is property name in the response, correct?
* `scopedResourceSelect` prop name
* `onRuleSelectorChange` callback function name

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -268,44 +284,80 @@ function CollectionForm({
...
-                                </CardBody>
-                            </Card>
-                        </Flex>
-                        {action.type !== 'view' && (
-                            <div className="pf-u-p-lg pf-u-py-md">
-                                <>
+                                </div>
+                            </Flex>
+                            {action.type !== 'view' && (
+                                <div className="pf-u-p-lg pf-u-py-md">
```
pedrottimark, line 350: As a follow up, can you see if `ActionGroup` element meets the need with fewer `className` props?

https://www.patternfly.org/v4/components/form#actiongroup

ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
```diff
@@ -0,0 +1,207 @@
...
+    validationErrors: FormikErrors<ByLabelResourceSelector> | undefined;
+};
+
+function ByLabelSelector({
+    entityType,
+    scopedResourceSelector,
+    handleChange,
+    validationErrors,
+}: ByLabelSelectorProps) {
+    function onChangeLabelKey(resourceSelector: ByLabelResourceSelector, ruleIndex) {
```
pedrottimark, line 34: A possible follow up is simple naming convention to make it even clearer to future readers that these are higher-order functions which return callback functions.

ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
```diff
@@ -0,0 +1,207 @@
...
+                                            '2px dashed var(--pf-global--Color--light-300)',
+                                        flex: '1 1 0',
+                                    }}
+                                />
+                            </Flex>
+                        )}
+
+                        <Flex>
+                            <Flex className="pf-u-flex-grow-1">
+                                <FormGroup
```
pedrottimark, line 122: Similar to another comment, we might need `fieldId` prop for `FormGroup` and `id` prop for `AutoCompleteSelect` element for both accessibility and integration tests. Also, what about additional values after the first.

ui/apps/platform/src/Containers/Collections/parser.ts
```diff
@@ -67,14 +80,36 @@ export function parseCollection(data: CollectionResponse): Collection | Aggregat
...
+
+        if (isByLabelSelector(selector)) {
+            const firstValue = rule.values[0]?.value;
+
+            if (firstValue && firstValue.includes(LABEL_SEPARATOR)) {
+                const key = firstValue.split(LABEL_SEPARATOR)[0] ?? '';
+                selector.rules.push({
+                    operator: 'OR',
+                    key,
+                    values: rule.values.map(({ value }) => value.split('=')[1] ?? ''),
```
pedrottimark, line 107: Remind me, is there, will there, or even can there be form validation to limit valid characters.

Just today, we saw an intentional fuzzing attempt to include split character in a value.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+            className="pf-u-p-lg"
+            style={{ border: '1px solid var(--pf-global--BorderColor--100' }}
+        >
+            <Select
+                toggleAriaLabel={`Select ${pluralEntity.toLowerCase()} by name or label`}
+                className={`${selection === 'All' ? '' : 'pf-u-mb-lg'}`}
+                isOpen={isOpen}
+                onToggle={onToggle}
+                selections={selection}
+                onSelect={onRuleOptionSelect}
```
pedrottimark, line 86: Forgive me for asking instead of testing whether form state update is correct if user selects a type of rule, enters values, and then changes selection to different type of rule? That is, it replaces the old with the new values instead of having a combination of both.

pedrottimark (approved):
Let’s go for it. Some comments for your consideration in the future.

pedrottimark:
Dave, one future looking request: can you please go through the steps that you followed to take the pictures to verify responsive behavior at a window width of your best judgement?

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
dvail: Hmm good catch here, I knew something felt off. What I was calling `selectorRules` in the FE type property is probably more similar to `resourceSelectors` in the BE response. The difference is the FE type is more restrictive, and pulls the `fieldName` up to the `resourceSelector` level when the BE places it on the individual `selectorRules`.

I'll update this to be:

- `resourceSelectors` in the BE response & FE data object
- `scopedResourceSelector` prop name
- `onResourceSelectorChange` callback function name

ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
dvail: I think this is a leftover from an earlier refactor. I'll see if I can get rid of the HOF in favor of a plain callback, like the other event handlers.

ui/apps/platform/src/Containers/Collections/parser.ts
dvail: Initially I was using the label helper methods on user input to ensure these were valid k8s labels, but then realized that these form fields need to support RE2 syntax.

Either way, I'll make a note to follow up with the BE team and implement something more robust here.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
dvail: Yes, updating the top level drop down for rule type will clear out and replace previous values. The MVP UX is intentionally limited so that users are only able to pick one of "all entities", "by name", or "by label" for each entity type.

dvail:
> Dave, one future looking request: can you please go through the steps that you followed to take the pictures to verify responsive behavior at a window width of your best judgement?

Fortunately PF handles the responsive styles really well out of the box. There were definitely a couple of odd places I missed though. I've updated the screenshots in the description and will push a fix in a few.

PR #3447, opened 2022-10-17
ROX-13112: Fix a race condition in setting the sensor hash

misberner:
## Description

The recent change of setting the sensor hash led to a race condition in the deduper (see https://issues.redhat.com/browse/ROX-13112). Another disadvantage is that it is non-obvious that the hash is computed and populated on all paths before sending out a deployment event. To improve this, set the hash in the same place where other mutations to the deployment object happen.
 
## Checklist

## Testing Performed

- CI

### DIFF
--- a/sensor/common/detector/deduper.go
+++ b/sensor/common/detector/deduper.go
@@ -1,7 +1,6 @@
 package detector
 
 import (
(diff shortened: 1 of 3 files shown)


lvalerom (approved):
LGTM

PR #3452, opened 2022-10-17
Use generics in RunGetQueryForSchema

janisz:
## Description

Look at 
- [ ] `pkg/search/postgres/common.go`
- [ ] `tools/generate-helpers/pg-table-bindings/store.go.tpl`

## Testing Performed

CI

### DIFF
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -721,7 +721,7 @@ type unmarshaler[T any] interface {
 }
 
 // RunGetQueryForSchema executes a request for just the search against the database
-func RunGetQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) ([]byte, error) {
+func RunGetQueryForSchema[T any, PT unmarshaler[T]](ctx context.Context, schema *walker.Schema, q *v1.Query, db *pgxpool.Pool) (*T, error) {
 	query, err := standardizeQueryAndPopulatePath(q, schema, GET)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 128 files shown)


pkg/search/postgres/common.go
```diff
@@ -735,6 +735,35 @@ func RunGetQueryForSchema(ctx context.Context, schema *walker.Schema, q *v1.Quer
...
+		if err := row.Scan(&data); err != nil {
+			return nil, err
+		}
+		msg := PT(new(T))
+		if err := msg.Unmarshal(data); err != nil {
+			return nil, err
+		}
+
+		return (*T)(msg), nil
+	})
```
misberner, line 764:
```suggestion
	data, err := RunGetQueryForSchema(ctx, schema, q, db)
	if err != nil {
		retur nil, err
	}
	msg := PT(new(T))
	if err := msg.Unmarshal(data); err != nil {
		return nil, err
	}
	
	return (*T)(msg), nil
```
to reduce code duplication?

misberner (changes requested):
The slice/`GetMany` variant has a performance advantage. This is really only a cosmetic change, no? I'm not against it, I just doubt how effective this is given that at least in this PR you only change generated code. If this code will also be used in handwritten code, I think it's reasonable.

pkg/search/postgres/common.go
janisz: Good point, I extracted unamrshal and scanRows methods so and fixed handwritten code

md2119 (approved):

janisz:
Merging as failures are known issue https://issues.redhat.com/browse/ROX-13242

PR #3456, opened 2022-10-17
Jv update collector version processes listening on ports

JoukoVirtanen:
## Description

Updates the collector version so that it reports processes listening on ports. This is behind a feature flag which is disabled by default.

## Checklist

## Testing Performed

CI is sufficient


sbostick (approved):
LGTM

PR #3460, opened 2022-10-17
ROX-12763: Forward claims from IdP to Rox token

ivan-degtiarenko:
## Description

This PR goal is to make it possible to add custom claims from IdP token to Rox token. This is API-only feature, and it will not be exposed in the UI.

This is done by mapping IdP token claims to Rox token attributes.

It's possible to:
* map nested into struct array and primitive(string and bool) claims
* map several claims into one attribute(will be combined in one array)

Examples of possible mappings:
* `something.realm_access.roles` -> `groups` where roles is array of primitives(string, bool, integer)
*  `a.b.c` -> `anyname` where `c` is primitive

It's impossible to:
* map struct claims
* map specific index in nested array
* process float and integer number claim values

Examples of impossible mappings:
* `a.b.c.` where `c` is JSON object(for example,  `"c": { "d": "e" }`
* `something.realm_access.roles.0` where roles is array
* `a.b.c.` where `c` can take integer/float value( `{ "c": 12.13 }`

## Checklist

## Testing Performed

1. Unit test
2. Manual test: create auth provider, see claims added in "test login" page


pkg/auth/authproviders/backend_factory.go
```diff
@@ -8,7 +8,7 @@ import (
 // BackendFactory is responsible for creating Backends.
 type BackendFactory interface {
 	// CreateBackend creates a new backend instance for the given auth provider, using the specified configuration.
-	CreateBackend(ctx context.Context, id string, uiEndpoints []string, config map[string]string) (Backend, error)
+	CreateBackend(ctx context.Context, id string, uiEndpoints []string, config map[string]string, mappings map[string]string) (Backend, error)
```
janisz, line 11: Why mappings are not part of config?

janisz, line 11: How about creating a struct for this params? Maybe a custom type instead of string?

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -629,9 +653,69 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
-	if p.httpClient != nil {
-		return oidc.ClientContext(ctx, p.httpClient)
+type claimExtractor interface {
+	Claims(v interface{}) error
+}
+
+func extractCustomClaims(externalUserClaim *tokens.ExternalUserClaim, mappings map[string]string, claimExtractor claimExtractor) error {
+	claims := make(map[string]interface{}, 0)
+	if err := claimExtractor.Claims(&claims); err != nil {
+		return errors.Wrap(err, "failed to construct Go map")
```
janisz, line 663:
```suggestion
		return errors.Wrap(err, "failed to extract claims")
```

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -629,9 +653,69 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
 	return claim
 }
 
-func (p *backendImpl) injectHTTPClient(ctx context.Context) context.Context {
-	if p.httpClient != nil {
-		return oidc.ClientContext(ctx, p.httpClient)
+type claimExtractor interface {
```
janisz, line 656: Looks like below code is not tested

janisz (commented):
I do not have much context for this change

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -629,9 +653,69 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+		}
+	}
+	return nil
+}
+
+func addClaimToUserClaims(externalUserClaim *tokens.ExternalUserClaim, attributeName string, claimValue interface{}) error {
+	switch v := claimValue.(type) {
+	case []interface{}:
+		for i, arrayVal := range v {
+			_, isArray := arrayVal.([]interface{})
```
mtodor, line 681: In the PR description stays that we support nested arrays, but I didn't see that functionality. By nested arrays, I would assume the mapping of:

with path: `arr`
```
{
  "arr": [
    ["one", "two"]
  ]
}
```

or with path `arr.num`:
```
{
  "arr": [
    { "num": "one" }
  ]
}
```

What is meant by nested array? Or have I missed something in the code? 😊

ivan-degtiarenko: To be honest I didn't figure out right vocabulary yet, so it can be confusing what is possible and what isn't. I'll add examples to the PR description.

pkg/auth/authproviders/oidc/extract_claims_test.go
```diff
@@ -0,0 +1,64 @@
...
+	claimExtractor := &mockClaimExtractor{
+		claims: map[string]interface{}{
+			"realm_access": map[string]interface{}{
+				"roles": rolesList,
+			},
+			"a": map[string]interface{}{
+				"b": "a-b-value",
+			},
+			"is_internal":       true,
+			"is_email_verified": false,
```
mtodor, line 45: I would suggest adding test cases for some error situations:
- a case for `Unsupported claim type` - i.e. map is not supported. Probably path `a` would cover that.
- a case for `expected next value to be of map type`. I think path `a.b.c` would cover that.

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -630,9 +649,67 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+				if err := addClaimToUserClaims(externalUserClaim, attributeName, arrayVal); err != nil {
+					return errors.Wrapf(err, "failed to add %d element of %+v", i, v)
+				}
+			} else {
+				return errors.Errorf("Unsupported claim type %T with value %+v", arrayVal, arrayVal)
+			}
+		}
+	case string:
+		externalUserClaim.Attributes[attributeName] = append(externalUserClaim.Attributes[attributeName], v)
+	case bool:
```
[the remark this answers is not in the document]
ivan-degtiarenko: We removed it after a discussion with Alex. I updated the description.

mtodor (approved):
Nice work!

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -630,9 +649,66 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+		}
+		if i == len(claimPath)-1 {
+			return nextVal, nil
+		}
+		currentNode, ok = nextVal.(map[string]interface{})
+		if !ok {
+			return nil, errors.Errorf("expected next value to be of map type but got %T", nextVal)
+		}
+	}
+	return nil, nil
```
rukletsov, line 713: This looks weird. Is it because we are never supposed to reach this line? If so, I'd suggest to:
* either change the loop control flow to return `nextVal, nil` upon completion
* or add a warning log line between after the loop saying unexpected instructions are called and return an error in L713.

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -630,9 +649,66 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+			_, isNestedStruct := arrayVal.(map[string]interface{})
+			if isArray || isNestedStruct {
+				return errors.Errorf("Unsupported claim type %T with value %+v", arrayVal, arrayVal)
+			}
+			if err := addClaimToUserClaims(externalUserClaim, attributeName, arrayVal); err != nil {
+				return errors.Wrapf(err, "failed to add %d element of %+v", i, v)
+			}
+		}
+	case string:
+		externalUserClaim.Attributes[attributeName] = append(externalUserClaim.Attributes[attributeName], v)
```
rukletsov, line 688: Maybe extract `externalUserClaim.Attributes[attributeName]` for readability? E.g., `targetClaimValues := externalUserClaim.Attributes[attributeName]`.

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -630,9 +649,66 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+}
+
+func addClaimToUserClaims(externalUserClaim *tokens.ExternalUserClaim, attributeName string, claimValue interface{}) error {
+	switch v := claimValue.(type) {
+	case []interface{}:
+		for i, arrayVal := range v {
+			_, isArray := arrayVal.([]interface{})
+			_, isNestedStruct := arrayVal.(map[string]interface{})
+			if isArray || isNestedStruct {
+				return errors.Errorf("Unsupported claim type %T with value %+v", arrayVal, arrayVal)
```
rukletsov, line 681:
```suggestion
				return errors.Errorf("unsupported claim type %T with value %+v", arrayVal, arrayVal)
```
We strive to standardize on lower case error messages.

pkg/auth/authproviders/oidc/backend_impl.go
```diff
@@ -630,9 +649,66 @@ func userInfoToExternalClaims(userInfo *userInfoType) *tokens.ExternalUserClaim
...
+			if err := addClaimToUserClaims(externalUserClaim, attributeName, arrayVal); err != nil {
+				return errors.Wrapf(err, "failed to add %d element of %+v", i, v)
+			}
+		}
+	case string:
+		externalUserClaim.Attributes[attributeName] = append(externalUserClaim.Attributes[attributeName], v)
+	case bool:
+		externalUserClaim.Attributes[attributeName] = append(externalUserClaim.Attributes[attributeName], strconv.FormatBool(v))
+	default:
+		return errors.Errorf("Unsupported claim type %T with value %+v", claimValue, claimValue)
```
rukletsov, line 692:
```suggestion
		return errors.Errorf("unsupported claim type %T with value %+v", claimValue, claimValue)
```

proto/storage/auth_provider.proto
```diff
@@ -37,4 +37,9 @@ message AuthProvider {
     repeated RequiredAttribute required_attributes = 11;
 
     Traits traits = 12;
+
+    // Specifies claims from IdP token that will be copied to Rox token attributes.
+    // For example, we can specify nested array claim "realm_access.roles" to be mapped to "groups" attribute.
+    // Currently we only support this feature only for OIDC auth provider.
```
rukletsov, line 43: This proto file is the only documentation we expose for this feature. Hence I strongly suggest:
* Format it for readability, both in text and in API reference (you can use [this](https://github.com/stackrox/stackrox/blob/a99e49ee0bb165573a04c41af76c7018323af4d2/proto/api/v1/role_service.proto#L201-L228) for inspiration and see its rendering issues [here](https://staging.demo.stackrox.com/main/apidocs#tag/RoleService/operation/RoleService_ComputeEffectiveAccessScope)
* Give examples of what's possible and what's not, what you do in the PR description
* Explicitly say what both `strings` mean, i.e., the first is the dot separated path to the claim in the IdP token and the second is the key name under `Attributes` in the Rox token.

pkg/auth/authproviders/backend_factory.go
```diff
@@ -8,7 +8,7 @@ import (
 // BackendFactory is responsible for creating Backends.
 type BackendFactory interface {
 	// CreateBackend creates a new backend instance for the given auth provider, using the specified configuration.
```
rukletsov, line 10: Please explain what `mappings` mean here. The comment will likely duplicate some info from the comment you will write in .proto and that's great. Specifically, I suggest to mention the intent (forwarding specific claims from IdP to Rox token) and the format (key is path and value is attribute name).

Alternatively, you can create a type, e.g., `IdPToRoxClaimMapping`, where all the comments above will go. Up to you.

pkg/auth/authproviders/oidc/extract_claims_test.go
```diff
@@ -0,0 +1,109 @@
...
+		for k, v := range e.claims {
+			(*u)[k] = v
+		}
+		return nil
+	default:
+		return errors.Errorf("unsupported type %T", input)
+	}
+}
+
+func TestExtractCustomClaims(t *testing.T) {
```
rukletsov, line 27: Thanks!

pkg/auth/authproviders/oidc/backend_impl.go
ivan-degtiarenko: Simple extraction does not work(I put `targetClaimValues := externalUserClaim.Attributes[attributeName]` and attribute slices are not filled correctly in `map_claims_test.go`).

I think that happens because `append` returns a new slice if the underlying array has to grow to accomodate the new element, thus we cannot guarantee that `targetClaimValues` and `externalUserClaim.Attributes[attributeName]` point to the same slice after `append` operation.

rukletsov: I see.

rukletsov (approved):

PR #3461, opened 2022-10-17
Convert test containers from x86 arch to multi-arch

kcrane:
## Description

Containers used during testing are x86 and will fail when testing ppc64le or s390x.  Instead use a container repo with multi-arch versions of these containers.

### DIFF
--- a/qa-tests-backend/build.gradle
+++ b/qa-tests-backend/build.gradle
@@ -143,7 +143,7 @@ tasks.withType(Test) {
 }
 
 test {
-    timeout = Duration.ofHours(3)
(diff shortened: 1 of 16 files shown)


porridge:
cc @stackrox/collector-team

porridge:
This seems to be against `master` tip as of August, I think you'll need to rebase for the builds/tests to succeed.

JoukoVirtanen:
> This seems to be against `master` tip as of August, I think you'll need to rebase for the builds/tests to succeed.

Yes. @kcrane please rebase.

PR #3462, opened 2022-10-17
ROX-12615: Collections DryRun workflow

keyallis:
## Description

Implement the collection dryrun workflow. Store name as value in the DAG to check for duplicates before trying to upsert into database, and add test to cover that.

## Checklist

## Testing Performed

Added test coverage.


central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -119,6 +132,33 @@ func (ds *datastoreImpl) deleteCollectionFromGraph(id string) error {
...
+		if err != nil {
+			ret = err
+			break
+		}
+	}
+
+	err = ds.graph.DeleteVertex(obj.GetId())
+	if err != nil && ret == nil {
+		ret = err
+	}
```
md2119, line 157: This can potentially result in a lock timeout. The chances of that happening increase significantly as the graph size increases.
Have you experimented with the `Copy()` operation performance? That is create a copy of DAG and work on the copy and set it to nil.

central/resourcecollection/datastore/datastore.go
```diff
@@ -30,7 +30,7 @@ type DataStore interface {
 
 	AddCollection(ctx context.Context, collection *storage.ResourceCollection) error
 	DeleteCollection(ctx context.Context, id string) error
-	// AddCollectionDryRun(ctx context.Context, collection *storage.ResourceCollection) error TODO ROX-12615
+	DryRunCollection(ctx context.Context, collection *storage.ResourceCollection) error
```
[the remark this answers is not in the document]
keyallis: As discussed, I think the deployment matching will be a separate function for the service layer to invoke

central/resourcecollection/datastore/datastore_impl.go
keyallis: These are good ideas and I've captured them for future work: https://issues.redhat.com/browse/ROX-13228

I think it would be good to continue working through functionality and getting up the service layer and getting baseline workflow functionality down before spending too much time on optimizing

keyallis:
Rebased on master to resolve CI failures

central/resourcecollection/datastore/datastore.go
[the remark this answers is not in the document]
md2119: Okay, let's rename the function to `DryRunAddCollection` to clarify.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -54,16 +54,29 @@ func (ds *datastoreImpl) initGraph() error {
 			sac.AccessModeScopeKeys(storage.Access_READ_ACCESS),
 			sac.ResourceScopeKeys(resources.WorkflowAdministration)))
```
md2119, line 55: This can just be `sac.WithAllAccess(context.Background())` since we know we are triggering the access and for what purposes.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -54,16 +54,29 @@ func (ds *datastoreImpl) initGraph() error {
...
+		} else {
+			objs, _, err = ds.storage.GetMany(ctx, ids[i:])
+		}
 		if err != nil {
 			return err
 		}
+		for _, obj := range objs {
+			err = ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
+			if err != nil {
+				return err
```
md2119, line 77: Add context to the return `errors.Wrapf(err, "building collections graph")`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -182,6 +222,10 @@ func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.
 	return err
 }
 
+func (ds *datastoreImpl) DryRunCollection(_ context.Context, collection *storage.ResourceCollection) error {
+	return ds.dryRunCollectionInGraph(collection)
```
md2119, line 226: Collection names are unique. Therefore, you have to validate the name uniqueness in the dry-run. You can hold a set of stored names to perform the name uniqueness validation.

md2119 (changes requested):

central/resourcecollection/datastore/datastore_impl.go
md2119: Fair enough

central/resourcecollection/datastore/datastore_impl.go
md2119, line 226: You must perform a permission check on `Workflow Administrator` first.

md2119: Further explained: the dry run operation does not reach the postgres store which handles permission checks.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -119,6 +132,33 @@ func (ds *datastoreImpl) deleteCollectionFromGraph(id string) error {
...
+	defer ds.graphLock.Unlock()
+
+	var ret error
+
+	// we essentially just add the obj to the graph and delete it before returning
+	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
+	if err != nil {
+		return err
+	}
+	for _, edge := range obj.GetEmbeddedCollections() {
```
md2119, line 146: nit: `edge` -> `peer`

central/resourcecollection/datastore/datastore.go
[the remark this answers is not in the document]
keyallis: addressed

central/resourcecollection/datastore/datastore_impl.go
keyallis: addressed

central/resourcecollection/datastore/datastore_impl.go
keyallis: addressed

central/resourcecollection/datastore/datastore_impl.go
keyallis: There are already unit tests included in this PR for this. The way this gets handled is that the DAG library checks for uniqueness of keys and values, the name field is stored as the value. In the future if we need to certify the uniqueness of more fields we can define a struct to store all the relevant ones

keyallis: addressed and added a variation of the comment

central/resourcecollection/datastore/datastore_impl.go
keyallis: addressed

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -182,6 +222,15 @@ func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.
 	return err
 }
 
+func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	// check for access since dryrun flow doesn't actually hit the postgres layer
+	if ok, err := workflowSAC.WriteAllowed(ctx); err != nil || ok != true {
```
md2119: `ok != true` -> `!ok`

md2119 (approved):
Minor nit otherwise LGTM.

Separately, noticed that the graph is init through accessors rather than datastore init. Hence, you require `initGraphOnce()`. Move the graph init inside datastore init i.e. inside `New(...)`.
 
I would refrain from panicking Central if there are issue initializing the graph, although the graph and it's correction is key for any Collection workflow. I recommend logging instead. 

You can further enhance the user experience by having a datastore field `initErr` which you can set with the error encountered during graph init, check for it in every workflow, and return it if set. Therefore, all user interactions with the UI actually inform user about the issue.

central/resourcecollection/datastore/datastore_impl.go
md2119: Got it!

central/resourcecollection/datastore/datastore_impl.go
keyallis: That came up in style-checks

PR #3463, opened 2022-10-17
Consolidate set uses

misberner:
## Description

Use our `pkg/set` package everywhere, getting rid of two usages of a third-party set library (nothing wrong with that library, but our `pkg/set` is not going away as we rely on the interface - a.o.t. allowing us to `range` over the set directly - way too much).

(Actually, one thing that bugs me about the library. The sets are by default thread-safe. I think that's a fallacy - except for in rare circumstances, you don't get concurrency-correct code by simply using thread-safe data structures, the entire algorithm has to be designed around the possibility of concurrent insertions/updates).

## Checklist

## Testing Performed

- CI

### DIFF
--- a/central/policy/service/validator.go
+++ b/central/policy/service/validator.go
@@ -6,7 +6,6 @@ import (
 	"regexp"
 	"strings"
 
-	mapset "github.com/deckarep/golang-set"
 	"github.com/pkg/errors"
 	notifierDataStore "github.com/stackrox/rox/central/notifier/datastore"
 	"github.com/stackrox/rox/generated/storage"
@@ -18,6 +17,7 @@ import (
 	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/policies"
 	"github.com/stackrox/rox/pkg/scopecomp"
+	"github.com/stackrox/rox/pkg/set"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


central/policy/service/validator.go
```diff
@@ -218,14 +218,15 @@ func (s *policyValidator) getCaps(policy *storage.Policy, capsTypes string) []*s
 }
 
 func (s *policyValidator) validateCapabilities(policy *storage.Policy) error {
-	set := mapset.NewSet()
+	values := set.NewSet[string]()
 	for _, s := range s.getCaps(policy, fieldnames.AddCaps) {
```
RTann, line 222: want to have `s.getCaps` return a `set.Set[*storage.PolicyValue]` and do `(*set).Intersect`?

pkg/secret/convert/convert.go
```diff
@@ -1,18 +1,17 @@
 package convert
 
 import (
-	mapset "github.com/deckarep/golang-set"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/set"
 )
 
 // SecretToSecretList converts a secret to list secret
 func SecretToSecretList(s *storage.Secret) *storage.ListSecret {
```
RTann, line 9: maybe for another PR, but I feel like `SecretToListSecret` makes more sense

RTann (approved):

central/policy/service/validator.go
misberner: I thought about this (or rather, doing something directly set-based), but the downside is that the error message then becomes non-deterministic, or requires sorting. The chosen approach of de-duplicating but relying on the existing ordering seems more elegant to me.

PR #3475, opened 2022-10-18
ROX-13185: Adding deployments table to namespace side panel

sachaudh:
## Description

Adding the namespace side panel with the deployments table. This is an iterative addition for the deployments tab of the namespace side panel

Additional sub-tasks for the deployments tab of the namespace side panel can be found here: https://issues.redhat.com/browse/ROX-12907

## Checklist

## Screenshots

Screen Shot 2022-10-18 at 2 00 21 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/namespace/NamespaceSideBar.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/namespace/NamespaceSideBar.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import {
+    Badge,
+    Flex,
+    FlexItem,
+    Tab,
+    TabContent,
+    Tabs,
+    TabTitleText,
+    Text,
+    TextContent,
+    TextVariants,
+} from '@patternfly/react-core';
+
+import useTabs from 'hooks/patternfly/useTabs';
+import NamespaceDeployments from './NamespaceDeployments';
+
+function NamespaceSideBar() {
+    const { activeKeyTab, onSelectTab } = useTabs({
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/namespace/NamespaceSideBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+                        <Text component={TextVariants.h1} className="pf-u-font-size-xl">
+                            stackrox
+                        </Text>
+                    </TextContent>
+                    <TextContent>
+                        <Text
+                            component={TextVariants.h2}
+                            className="pf-u-font-size-sm pf-u-color-200"
+                        >
+                            in &quot;remote&quot;
```
vjwilson, line 40: @alanonthegit 
The way we have to code these quotes reminded me to ask:

Should we be emphasizing the k8s hierarchy (cluster or cluster/namespace) with something besides quotes?
Like **bold** and/or *italic*

Quotes looks a little like an afterthought, and if we internalize our app, neither locals use different symbols for quote marks, so the meaning is more obscure.

vjwilson (approved):
Good iterative progress.

Non-blocking question for Alan.

ui/apps/platform/src/Containers/NetworkGraph/namespace/NamespaceSideBar.tsx
sachaudh: On that note, is it even necessary to have this if we have the breadcrumbs up top?

PR #3491, opened 2022-10-19
Add fix for unstable component key and fill out tests

dvail:
## Description

This adds a fix for unstable key values in the collections form. This follows the guidelines linked in the React docs [here](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern#update-exception-from-the-rule).

This solution effectively is using the array index as the key, but appended to a stable, unique value. The downside with this approach is that you need to manually call the `invalidateIndexKeys` function whenever one of the rules from the link above are violated. This resets the stable value, wiping the keys for your array. This sacrifices the performance boost React gets from iterating over a list in favor of ensuring the list will render correctly.

This PR also re-enables the broken test caused by the error, and adds an equivalent test for the ByLabelSelector component.

An [alternative approach](https://github.com/stackrox/stackrox/pull/3492) to this is to add arbitrary unique ids to each item client side. This is more verbose and adds cruft to the data model, but doesn't require us to manually invalidate keys and is a more type-safe approach.

## Checklist

## Testing Performed

Manual testing according to #3444

UI unit tests are passing

### DIFF
--- a/ui/apps/platform/src/hooks/useIndexKey.tsx
+++ b/ui/apps/platform/src/hooks/useIndexKey.tsx
@@ -27,7 +27,11 @@ type UseIndexKeyReturn = {
  * rendering.
  */
 export default function useIndexKey(): UseIndexKeyReturn {
-    const [prefix, setPrefix] = useState(globalPrefix);
+    const [prefix, setPrefix] = useState(() => {
+        // Increment on initial call to hook, to ensure no two components start from the same index
+        globalPrefix += 1;
+        return globalPrefix;
+    });
     const keyFor = useCallback((index: number) => `${prefix}-useIndexKey-${index}`, [prefix]);
     const invalidateIndexKeys = useCallback(() => {
         globalPrefix += 1;
--- a/ui/apps/platform/src/hooks/useIndexKey.tsx
+++ b/ui/apps/platform/src/hooks/useIndexKey.tsx
@@ -1,0 +1,38 @@
+import { useState, useCallback } from 'react';
+
+// Using a single value at the module scope allows multiple components to use this hook
+// without generating conflicting keys.
+let globalPrefix = 0;
+
+type UseIndexKeyReturn = {
+    /**
+     * Generate a stable key for an array index
+     */
+    keyFor: (index: number) => string;
+    /**
+     * Invalidates all keys in use by this hook, causing React to rerender the entirety
+     * of the list, but keeping render results predictable and correct.
+     */
+    invalidateIndexKeys: () => void;
+};
+
+/**
+ * Hook that allows the usage of `index` values for React component keys in -specific- situations.
+ * This should only be used when the following are true:
+ * 1. There is no stable key to be used instead.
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #3444** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3444" target="_blank">Graphite</a> 
    * **PR #3491** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3491" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3491?utm_source=stack-comment).

ui/apps/platform/src/hooks/useIndexKey.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * 1. There is no stable key to be used instead.
+ * 2. The ordering of the items in the list never changes.
+ *
+ * For the second case, this hook can be used but the `invalidateKeys` function must be called whenever
+ * the list is rearranged, or an item is deleted. This will eschew performance in favor of predictable
+ * rendering.
+ */
+export default function useIndexKey(): UseIndexKeyReturn {
+    const [prefix, setPrefix] = useState(globalPrefix);
+    const keyFor = useCallback((index: number) => `${prefix}ReactIndexKey-${index}`, [prefix]);
```
pedrottimark, line 35: Although `key` is only in React DevTools not HTML inspector, here is a stronger hint to consider:

`${prefix}-useIndexKey-${index}`

dvail:
1. Each `map` can use the same `keyFor()` from a single hook call, so long as the mapped elements are not siblings. I can't think of any place this is the case right now, but I suppose it is possible if you have two arrays of similar objects that you want to display in a single list.

```javascript
    // This is fine
    const { keyFor } = useIndexKey();
    const fruit = ['apples', 'oranges', 'kiwi'];
    const veggies = ['kale', 'cabbage', 'carrot'];

    return (
        <>
            <ul>
                {fruit.map((f, i) => (
                    <li key={keyFor(i)}>{f}</li>
                ))}
            </ul>
            <ul>
                {veggies.map((v, i) => (
                    <li key={keyFor(i)}>{v}</li>
                ))}
            </ul>
        </>
    );


    // This could result in key conflicts and would need a separate `useIndexKey()` call
    const { keyFor } = useIndexKey();
    const fruit = ['apples', 'oranges', 'kiwi'];
    const veggies = ['kale', 'cabbage', 'carrot'];

    return (
        <>
            <ul>
                {fruit.map((f, i) => (
                    <li key={keyFor(i)}>{f}</li>
                ))}
                {veggies.map((v, i) => (
                    <li key={keyFor(i)}>{v}</li>
                ))}
            </ul>
        </>
    );
```

2. `invalidateIndexKeys()` should be called any time you mutate the array and cause an item to change its index, e.g. deleting or swapping an element. If you are just appending to the list, that should be OK to not call `invalidateIndexKeys`

PR #3497, opened 2022-10-19
ROX-13134: Make sidebar conditional on URL params

vjwilson:
## Description

Quick scaffolding of selecting graph nodes and triggering sidebar with URL change

## Checklist

## Testing Performed

Only node clicks trigger for now (will add deep-link refresh functionality later)

Deployment
Screen Shot 2022-10-19 at 2 49 38 PM

Namespace
Screen Shot 2022-10-19 at 2 49 42 PM

CIDR Block
Screen Shot 2022-10-19 at 2 49 45 PM

External Entities
Screen Shot 2022-10-19 at 2 49 50 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,9 +1,8 @@
-import React, { useState } from 'react';
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+import React from 'react';
 import { useHistory } from 'react-router-dom';
 import {
-    EdgeModel,
     Model,
-    NodeModel,
     SELECTION_EVENT,
     TopologySideBar,
     TopologyView,
@@ -24,6 +23,18 @@ import ExternalEntitiesSideBar from './external/ExternalEntitiesSideBar';
 
 import './Topology.css';
 
+// TODO: move these type defs to a central location
+export const UrlDetailType = {
+    NAMESPACE: 'namespace',
+    DEPLOYMENT: 'deployment',
+    CIDR_BLOCK: 'cidr',
+    EXTERNAL_ENTITIES: 'internet',
+    EXTERNAL: 'external',
+} as const;
+export type UrlDetailTypeKey = keyof typeof UrlDetailType;
+export type UrlDetailTypeValue = typeof UrlDetailType[UrlDetailTypeKey];
+
+// TODO: replace this dummy data with real parsed graph data
 const model: Model = {
     graph: {
         id: 'g1',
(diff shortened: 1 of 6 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -28,76 +32,167 @@ const model: Model = {
...
+        {
+            id: 'e4',
+            type: 'edge',
+            source: 'afa12424-bde3-4313-b810-bb463cbe8f90',
+            target: 'e337f873-64d8-46be-84ed-2a0c38c75fac',
         },
     ],
 };
 
+function findEntityById(graphModel, id) {
```
sachaudh, line 152: What do you think about defining these functions in separate modules? That way we can also create a test file for each of them

sachaudh: I would imagine we'll eventually have different types of helper functions for the graph:
* Finding nodes/edges using some condition (id, name, namespace, etc.)
* Filtering nodes/edges by some condition (connected to a specific node, etc.)

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -28,76 +32,167 @@ const model: Model = {
...
+    let entity = null;
+    entity = graphModel.nodes?.find((node) => node.id === id);
+    if (!entity) {
+        entity = graphModel.groups?.find((group) => group.id === id);
+    }
+    return entity;
+}
+
+function getUrlParamsForEntity(selectEntity): [string, string] {
```
sachaudh, line 161: We should probably add a type for the `selectedEntity`. I think this might require us to make a custom type for our `Model`. The `data` type for `Model` is `any`. If we make our own custom type for `Model`, we can specify the structure for our `data` field

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -28,76 +32,167 @@ const model: Model = {
...
     const controller = useVisualizationController();
 
     React.useEffect(() => {
         function onSelect(ids: string[]) {
             const newSelectedId = ids?.[0] || null;
             // check if selected id is for a node
-            const newSelectedNode = model.nodes?.find((node) => node.id === newSelectedId);
-            if (newSelectedNode) {
-                setSelectedObj(newSelectedNode);
+            // const newSelectedEntity = model.nodes?.find((node) => node.id === newSelectedId);
```
sachaudh, line 186: Looks like we'll have to do this a bit differently because of this case:
1. Select a deployment
2. Refresh the page (the URL should include that deployment id)
3. It won't show the side panel or the deployment selected

I would expect that the side panel would be open to that deployment

sachaudh: Also, maybe we shouldn't keep the `selectedObj` state because we'll have an issue with synchronizing data. We can infer the selected object using the URL state

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -7,6 +8,9 @@ import NetworkGraph from './NetworkGraph';
 import './NetworkGraphPage.css';
 
 function NetworkGraphPage() {
+    const { detailType, detailId } = useParams();
```
sachaudh, line 11: Seems like we're not doing anything with the `detailType` and `detailId`. Maybe that's intentional and you were going to do some iterative work. I think to solve the issues highlighted above, we could have a custom hook that does the following:

1. Uses the `useParam` hook to get the `type` and `id` of a selected object (deployment, namespace, etc.)
2. Provides a function that can be used in the `onSelect` of the `NetworkGraph` component to update the URL

We can then pass the `type` and `id` as props to the `NetworkGraph` component and it can be used to determine what's selected

```
example:

const [selectedObject, setSelectedObject] = useNetworkObjectSelection();
...
return (
    ...
    <NetworkGraph selectedType={selectedObject?.type} selectedId={selectedObject?.id} />
    ...
)
```

sachaudh (approved):
Approving to unblock you

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
vjwilson: I'm reluctant to do that as part of this PR, because balloons the scope. I think that belongs more as a task along with @alwayshooin 's data mapping.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
vjwilson: You're absolutely right--I should have noted in my description that is the end-goal, of which this is just a piece, so that we can work on related sections without getting merge conflicts.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
vjwilson: Will definitely pull these out into a separate file. I just started with them here while I was exploring options.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
vjwilson: Using those params will be in part 2, _URL Network Graph_: Return to the URL_ ;)

PR #3499, opened 2022-10-19
[DRAFT] Testing multi-arch images for non-x86 platform support

sbostick:
https://docs.google.com/document/d/1B2wx5548-juts5NnDPlf9EotwmuonuJCbNZxT_P7-aQ/edit?usp=sharing


gavin-stackrox:
@sbostick you will need to take this PR out of draft to get the images built so that the test can find them.

gavin-stackrox:
> Moving to review-ready so the images get built. Thanks, @gavin-stackrox.

I use the PR view to try to make sense of what OpenShift CI is doing: https://prow.ci.openshift.org/pr-history/?org=stackrox&repo=stackrox&pr=3499

/test gke-upgrade-tests
^^ should get one running so it finds the images when they are done and starts to test.

scripts/ci/jobs/gke_upgrade_tests.py
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env -S python3 -u
 
+# TODO(sbostick): just tracking relevant files
+
 """
 Run the upgrade test in a GKE cluster
 """
```
gavin-stackrox, line 7: You probably need:
```
os.environ["ORCHESTRATOR_FLAVOR"] = "k8s"
os.environ["SENSOR_HELM_DEPLOY"] = "true"
```
at line 15. `k8s` because central is not openshift 🤞

sbostick: Got it, trying now...

tests/upgrade/run.sh
```diff
@@ -54,17 +54,21 @@ test_upgrade() {
     wait_for_api
     setup_client_TLS_certs
 
+    # TODO(sbostick): try installing to secondary "secured cluster"
     info "Deploying sensor"
-    "$TEST_ROOT/$DEPLOY_DIR/sensor.sh"
-    validate_sensor_bundle_via_upgrader "$TEST_ROOT/$DEPLOY_DIR"
+    "$TEST_ROOT/$DEPLOY_DIR/sensor.sh"  # <---- override as "deploy/openshift"?
```
gavin-stackrox, line 59: This is going to be the rub. The API endpoint and other configuration for sensor<->central will be wrong. I would put some hold logic in here and work out of CI to get the power/z cluster configured as the 'remote' cluster in the GKE cluster that gets configured for test the CI run. 
My goto hold for openshift CI:
```
    touch /tmp/hold
    while [[ -e /tmp/hold ]]; do
        info "Holding this job for debug"
        sleep 60
    done

```

gavin-stackrox: when you are ready to run the test you can remove the hold through `oc rsh` https://docs.engineering.redhat.com/display/StackRox/Tips+for+using+OpenShift+CI

PR #3509, opened 2022-10-20
ROX-13163: Rewrite frontend group service in TypeScript

pedrottimark:
## Description

Increase type safety with compile-time checks.

Counting down from double to single digits: 10 - 1 = 9 JavaScript files out of 36 total WhateverService files.

### Review questions

- [x] Which properties of `Group` type are optional?
- [x] Which properties of `Traits` type are optional?
- [x] Action to delete rule group is superseded by update groups?
- [ ] Recommend integration tests?

### Changed files

Here is a detailed description of changes, because type changes implied some functional changes, which are difficult to test:

1. Edit src/Containers/AccessControl/AuthProviders/RuleGroups.tsx
    * Import `Group` type from types file instead of service file.

2. Edit src/Containers/AccessControl/AuthProviders/authProviders.utils.ts
    * Import `Group` type from types file instead of service file.

3. Edit src/Containers/AccessControl/AuthProviders/Roles.tsx
    * Import `Group` type from types file instead of service file.

4. Edit src/Containers/AccessControl/AuthProviders/RolesList.tsx
    * Import `Group` type from types file instead of service file.

5. Edit src/reducers/groups.js
    * Delete unused type and action function to delete a rule group.

6. Edit src/sagas/groupSagas.js
    * Delete incorrect `result?.response || []` which was intended to mitigate access control failures. However, the value is an object `{ groups: [] }` instead of an array. Replace with `result.response` because request status 403 goes to `catch` block.
    * Replace `defaultGroup?.response` with `defaultGroup` because of change to simplify return value of `getDefaultGroup` service function.
    * Delete functions for unused action to delete a rule group.

7. Edit src/services/AuthService/AuthService.ts
    * Import types to be parallel with backend file structure:
        https://github.com/stackrox/stackrox/blob/master/proto/storage/group.proto
        https://github.com/stackrox/stackrox/blob/master/proto/storage/traits.proto
    * **Review question**: correct or incorrect that `Group` type had optional properties?
        ```js
        export type Group = {
            roleName: string;
            props: {
                authProviderId: string;
                key?: string;
                value?: string;
                id?: string;
                traits?: Traits;
            };
        };
        ```
    * Although linter and compiler are happy with the changes, this file is most difficult for me to read and understand because it “enriches” the `AuthProvider` response type with bonus `groups` and `defaultRole` properties.

8. Rewrite src/services/GroupsService.js in TypeScript (too bad, so sad that so many indentation changes cause GitHub to display as deleted and added files instead of renamed and changed)
    * Import `Group` and `Empty` types.
    * Add return type for `fetchGroups` function. To limit the number of changes because it is called in groupSagas.js and Roles.tsx files, I did not simplify the return type like the other functions. Even though it confuses me every time to read `response` property name when the value is the data property of the response :(
    * Add `Promise<Empty>` return type for `updateOrAddGroup` function to be consistent with endpoint and add `then((response) => response.data)` to make it true. Although it might look unintuitive at first, the promise chain method is as necessary for empty as non-empty response. Without it, the function leaks the entire response object!
    * Add simpler `Promise<Group | undefined>` return type for `getDefaultGroup` function because called only in groupSagas.js file.
    * Delete unused `deleteRuleGroup` function, whose only call was added in rox 772 and deleted in rox 885.

9. Edit src/types/group.proto.ts
    * Import `Traits` type.
    * Add `traits` property to `Group` type.
    * **Review question**: Please verify: optional property? possible `null` value?

10. Add src/types/traits.proto.ts
    * **Review question**: Please verify: optional properties? possible `null` values?

## Checklist
- [ ] Edited integration tests TODO pending code review and manual testing

## Testing Performed


ui/apps/platform/src/services/GroupsService.ts
```diff
@@ -0,0 +1,53 @@
...
+export function fetchGroups(): Promise<{ response: { groups: Group[] } }> {
+    return axios.get<{ groups: Group[] }>(url).then((response) => ({
+        response: response.data,
+    }));
+}
+
+/**
+ * Update/Add a group rule.
+ */
+export function updateOrAddGroup({ oldGroups, newGroups }: UpdateGroupsArg): Promise<Empty> {
```
dhaus67, line 21: nit: the weird thing with this `groupsbatch` API is that it can be used for adding, updating, or removing groups. It would be best to reflect that in the function name, since now it only implies adding / updating.

ui/apps/platform/src/types/group.proto.ts
```diff
@@ -20,4 +22,6 @@ export type GroupProperties = {
     authProviderId: string;
     key: string;
     value: string;
```
dhaus67, line 24: For the group properties, both `key` and `value` fields can be treated as optional. Only `id` and `authProviderId` are required.

For a bit of background and edge cases, [here's the reference to the validation within the datastore](https://github.com/stackrox/stackrox/blob/master/central/group/datastore/validate.go).

ui/apps/platform/src/types/traits.proto.ts
```diff
@@ -0,0 +1,21 @@
...
+    // modifying operations except object removal with force bit on.
+    //
+    // Be careful when changing the state of this field. For example, modifying an
+    // object from ALLOW_MUTATE to ALLOW_MUTATE_FORCED is allowed but will prohibit any further
+    // changes to it, including modifying it back to ALLOW_MUTATE.
+    mutabilityMode?: TraitsMutabilityMode;
+
+    // EXPERIMENTAL.
+    // visibility allows to specify whether the object should be visible for certain APIs.
+    visibility?: TraitsVisibility;
```
dhaus67, line 16: Both properties can be treated as optional.

dhaus67 (approved):
Only minor comments to answer the questions you have posted within the PR description.

Regarding the integration cases, I have couple of them in mind which we could implement together in a pair-session if you'd like:

1. A group without any traits object set should be modifiable.
2. A group with the `traits.mutabilityMode == ALLOW_MUTATE` should be modifiable.
3. A group with the `traits.mutabilityMode == ALLOW_MUTATE_FORCE` should _not_ be modifiable.
4. A group with the `traits.mutabilityMode == ALLOW_MUTATE_FORCE` should _not_ be modifiable, and should stay as-is, irrespective of other changes made to the auth provider (e.g. adding other groups, changing the default minimum access role).

As a note regarding the visibility trait: currently, this is only meant for the basic auth provider in managed service context, to filter it out from the API response. So, I don't think it makes sense to add integration tests specifically for this here.

pedrottimark:
Superseded by a more recent contribution.

PR #3511, opened 2022-10-20
ROX-12844: Prefactor proto changes to allow multiple image names

dhaus67:
## Description

This is the first prefactor in a series of PRs within the context of [ROX-12844](https://issues.redhat.com/browse/ROX-12844).

The underlying issue is the following:
Currently, we use an image's ID as the unique identifier of an image (which happens to be the digest of the compressed manifest). This has its upside, such as we do not have duplication on entries of images within the database and subsequently not scan multiple images with the same contents.

However, there's also some downside to this: repository digests _can_ be the same accross different registries and repositories.
Meaning, the image within `docker.io/nginx:1.23` and the image `quay.io/nginx:1.23` will have the same repository digest.
This leads us to only having a single entry of the image within central, which would be the first image that will be detected and upserted. This also means we lose the image name information of the second image.
This does not impact scanning, since we assume the image has the same content. However, this does affect properties that are bound to the location of the image, such as image signatures.
Here, the signatures will only be fetched and verified based on the image name, which will not include _all_ images with the _same_ digest coming from different locations (i.e. repositories / registries).
The consequence of this is having false-positives / false-negatives, based on whether the initially added image has signatures or does not have signatures.

This series of PRs aims to fix this short-coming by doing the following:
- Allow an image with _the same digest_ to have _multiple_ image names associated with it, from different repositories / registries.
- Ensure that, for signature detection and verification, _all_ potentially available image names are used and reflected within the image signature verification results.
- Fix the image signature policy to account for the newly added property on the image signature verification result and ensure that the image in the deployment spec matches with any of the verified image references.
- Ensure that sensor will properly upsert additional image names within the image enrichment pipeline.
- Ensure that central will reprocess the additional image names within the image enrichment pipeline.
- Create a migration for the proto related changes.

The first PR will only include proto changes. The proto changes will include:
- Creating a new field `names` for the `Image` message, which is a repeated `ImageName` message field. This will later on include all `ImageName`s that are associated with this particular image ID.
- Creating a new field `verified_image_refernces` for the `ImageSignatureVerificationResult` message, which is a repeated string field. This will later on include all full names of images that are covered by this particular image signature verification result.

## Checklist

## Testing Performed
- only proto changes, so no unit tests adjusted as of now.


dhaus67:
* master
  * **PR #3511** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3511" target="_blank">Graphite</a>  👈
    * **PR #3512** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3512" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3511?utm_source=stack-comment).

misberner (changes requested):

md2119 (changes requested):
@dhaus67 Do you have a design published for the changes you plan to make? It is not clear from the PR description the impact that the changes may have on existing workflow. Especially migration for proto changes worries me, but probably because no details are included in the description.

dhaus67:
@md2119 I've added doc in the description, would be great if you could take a look whether this makes things clearer.

proto/storage/image.proto
```diff
@@ -104,6 +108,9 @@ message ImageSignatureVerificationResult {
     Status status                               = 3;
     // description is set in the case of an error with the specific error's message. Otherwise, this will not be set.
     string description                          = 4 ;
+    // The full image name that the signature references. Since images can come from different locations, this will
+    // ensure we can verify the matching location as well.
+    repeated string verified_image_references = 5;
```
md2119, line 113: Can we move this field inside `ImageSignatureVerificationData`?

md2119 (commented):
Thanks for sharing the design document. According to the doc, there are no plans yet to replace the existing `name` field nor to make any changes to workflows that depend on the `name` field including UI. Keep me posted if anything changes on that front.

proto/storage/image.proto
dhaus67: Makes sense.

I'll change the type from `repeated string` to something else, as I imagine currently we'd like to have the coupling between verified image names and the signature integration ID.

Reason for this is that we currently in the policy have the user select a signature integration ID with which they expect the image's signature to be verified with.
Imagining we have multiple signature integrations, and each verifies different image names, we would lose this relationship there and potentially not create an alert for a specific image name and a specific signature integration ID.

proto/storage/image.proto
```diff
@@ -79,6 +83,12 @@ message ImageScan {
 
 message ImageSignatureVerificationData {
     repeated ImageSignatureVerificationResult results = 1;
+    // The full image names that are verified (meaning, having a successful signature verification) keyed by the
```
md2119, line 86: I think the comment makes the most sense above the `verified_image_references` field.

proto/storage/image.proto
```diff
@@ -79,6 +83,12 @@ message ImageScan {
...
+    // signature they were verified with.
+    message ImageReference {
+        repeated string names = 1;
```
md2119, line 89: What else could `ImageReference` contain in the future?

proto/storage/image.proto
```diff
@@ -79,6 +83,12 @@ message ImageScan {
...
+    }
+    map<string, ImageReference> verified_image_references     = 2;
```
md2119, line 91: I think searching over a `map<string, object>` is unsupported, and is tricky to support. If there is a remote chance for these fields to be searchable in future, I recommend changing to array.
```
message ImageReference {
  string image_signature_integration_id;
  repeated string image_name;
}
repeated ImageReference verified_image_references;
```

md2119 (commented):
I think the you can streamline the structure as follows which reduces some redundancy of stored data:

```
message ImageSignatureVerificationResult {
    google.protobuf.Timestamp verification_time = 1;
    // verifier_id correlates to the ID of the signature integration used to verify the signature.
    string verifier_id                          = 2;
    repeated string verified_image_names;
...
}
```

md2119 (approved):
LGTM. Thanks for doing the iterations!

proto/storage/image.proto
```diff
@@ -9,10 +9,14 @@ import "gogoproto/gogo.proto";
 
 package storage;
 
-// Next Tag: 18
+// Next Tag: 19
 message Image {
     string                          id                          = 4 [(gogoproto.moretags) = 'search:"Image Sha,store,hidden" sql:"pk"'];
     ImageName                       name                        = 1;
```
ivan-degtiarenko, line 15: Do we want to put `deprecated` on this field? Or not yet?

dhaus67: I'd say not yet.

PR #3514, opened 2022-10-21
ROX-13201: Create dedicated config for db settings

c-du:
## Description
Separate the external db config from central-config. So that we can create central-db bundle.

## Checklist
- ~[ ]~ Unit test and regression tests added
- ~[ ]~ Evaluated and added CHANGELOG entry if required
- ~[ ]~ Determined and documented upgrade steps
- ~[ ]~ Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

Deploy local and ci test


janisz (changes requested):
Could you add a unit test for this change?

PR #3515, opened 2022-10-21
ROX-12844: Refactor image enricher with multiple image names

dhaus67:
## Description

This is the third PR in a series of PRs to enable the image enricher to respect multiple image names when fetching signatures.

For additional context of the series of PRs, see the parent's PR context as well as [ROX-12844](https://issues.redhat.com/browse/ROX-12844).

To achieve this, the following has been done:
- Refactor image enricher's `enrichWithSignature` to ensure additional image names are respected.
- Adopt the existing interfaces for fetching signatures to include a image reference from which the signature should be fetched.

## Checklist

## Testing Performed

- see CI, unit tests adjusted.

### DIFF
--- a/pkg/images/enricher/enricher_impl.go
+++ b/pkg/images/enricher/enricher_impl.go
@@ -24,6 +24,7 @@ import (
 	scannerTypes "github.com/stackrox/rox/pkg/scanners/types"
 	"github.com/stackrox/rox/pkg/signatures"
 	"github.com/stackrox/rox/pkg/sync"
+	"github.com/stackrox/rox/pkg/urlfmt"
 	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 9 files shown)


dhaus67:
* master
  * **PR #3511** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3511" target="_blank">Graphite</a> 
    * **PR #3512** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3512" target="_blank">Graphite</a> 
      * **PR #3515** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3515" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3515?utm_source=stack-comment).

pkg/images/enricher/enricher_impl.go
```diff
@@ -711,11 +715,13 @@ func filterRegistriesBySource(requestSource *RequestSource, registries []registr
...
-	image *storage.Image) ([]registryTypes.ImageRegistry, error) {
-	var matchingRegistries []registryTypes.ImageRegistry
+	image *storage.Image) (map[string][]registryTypes.ImageRegistry, error) {
+	matchingRegistries := make(map[string][]registryTypes.ImageRegistry, len(image.GetNames().GetNames()))
 	for _, registry := range registries {
-		if registry.Match(image.GetName()) {
-			matchingRegistries = append(matchingRegistries, registry)
+		for _, name := range image.GetNames().GetNames() {
+			if registry.Match(image.GetName()) {
+				matchingRegistries[name.GetFullName()] = append(matchingRegistries[name.GetFullName()], registry)
```
misberner, line 723: This seems disadvantageous. If you have multiple images referring to the same registry, you will return the registry multiple times, and make multiple calls to the same registry for different names. This seems to be unavoidable from the standpoint of the cosign library (you can only fetch signatures for a single reference), but I guess it could be argued from a retry budget standpoint that retry budget for transient errors should be per registry, not per reference.

~Therefore, grouping by registry seems to make more sense to me.~ I wrote this before my above comments that registries should be sorted. Not sure what would be the right approach here. However, it seems unnecessarily complex to return a `map[string][]Registry`. Couldn't you just have left this function as-is except for replacing the `image` parameter with a `name` parameter, and then just call it in every iteration of the outer loop in l.608?

pkg/images/enricher/enricher_impl.go
```diff
@@ -605,27 +605,30 @@ func (e *enricherImpl) enrichWithSignature(ctx context.Context, enrichmentContex
...
-		}
+	for imageName, matchingRegs := range matchingRegistries {
+		for _, registry := range matchingRegs {
+			// FetchImageSignaturesWithRetries will try fetching of signatures with retries.
+			sigs, err := signatures.FetchImageSignaturesWithRetries(ctx, e.signatureFetcher, img, imageName, registry)
+			fetchedSignatures = append(fetchedSignatures, sigs...)
+			// Skip other matching registries if we have a successful fetch of signatures for the respective image name,
+			// irrespective of whether signatures were found or not.
+			// Retrying this for other registries won't change the fact that signatures are available or not for this
+			// particular image name. Note that we still will fetch image signatures for _all_ other image
```
misberner, line 616: Comment seems to end abruptly/miss a "names" in the end?

pkg/images/enricher/enricher_impl.go
```diff
@@ -605,27 +605,30 @@ func (e *enricherImpl) enrichWithSignature(ctx context.Context, enrichmentContex
...
-			break
-		}
+	for imageName, matchingRegs := range matchingRegistries {
+		for _, registry := range matchingRegs {
+			// FetchImageSignaturesWithRetries will try fetching of signatures with retries.
+			sigs, err := signatures.FetchImageSignaturesWithRetries(ctx, e.signatureFetcher, img, imageName, registry)
+			fetchedSignatures = append(fetchedSignatures, sigs...)
+			// Skip other matching registries if we have a successful fetch of signatures for the respective image name,
+			// irrespective of whether signatures were found or not.
+			// Retrying this for other registries won't change the fact that signatures are available or not for this
```
misberner, line 615: It's not quite clear to me how one image name could have multiple matching _registries_. I can imagine that there can be multiple matching _integrations_, i.e., a global `quay.io` with some default credentials, and a more specific `quay.io/rhacs-eng` with credentials with access to this org. This would however imply that the list of registries/integrations per image name should be sorted, no?

pkg/images/enricher/enricher_impl.go
```diff
@@ -711,11 +715,13 @@ func filterRegistriesBySource(requestSource *RequestSource, registries []registr
...
 func getMatchingRegistries(registries []registryTypes.ImageRegistry,
-	image *storage.Image) ([]registryTypes.ImageRegistry, error) {
-	var matchingRegistries []registryTypes.ImageRegistry
+	image *storage.Image) (map[string][]registryTypes.ImageRegistry, error) {
+	matchingRegistries := make(map[string][]registryTypes.ImageRegistry, len(image.GetNames().GetNames()))
 	for _, registry := range registries {
-		if registry.Match(image.GetName()) {
-			matchingRegistries = append(matchingRegistries, registry)
+		for _, name := range image.GetNames().GetNames() {
+			if registry.Match(image.GetName()) {
```
misberner, line 722: Don't you mean
```suggestion
			if registry.Match(name.GetName()) {
```
here?

misberner (changes requested):

pkg/images/enricher/enricher_impl.go
dhaus67: I agree that it's unfortunate regarding having to make multiple calls to the same registry for the different image names.

I've change it up a bit to follow your suggestion with calling getMatchingRegistries within the for-loop, with minor caveats:

before trying to do anything, we first validate whether at least one registry exists (irrespective of the image name) with a new helper - which will error out with the same conditions as previously for getMatchingRegistries.
within the for-loop, iff for one particular image name no matching integrations are given, we are going to log it and skip fetching signatures for that particular image name - I don't think this mandates that we error out completely of the image enrichment process. Say, we would have the case where an image is having the same digest in a OpenShift internal registry and a public registry, we would always error out enriching that particular image in the central context.

pkg/images/enricher/enricher_impl.go
dhaus67: Yeah, its currently being used quite interchangeably within the image enricher the meaning of "registry" and "image integration" - the latter is meant here which also matches exactly the example you have given.
I've renamed the existing occurrences of e.g. functions / variables and sorted the matching image integrations by their registry hostname.

md2119 (approved):

PR #174, opened 2021-12-21
ROX-8392: Make use of the injected trusted CA bundle

parametalol:
## Description

_Cluster Network Operator_ merges the user-provided and system CA certificates into a single bundle and injects the merged bundle into a labelled ConfigMap.

The PR adds the injected CA certificates bundle to the trusted cert pool.

There could be implemented a test in the authprovider backend so that in the case of existence of the mounted bundle to not include the system bundle to the pool, but hardly that will bring any benefits over the suggested simplified code.

Related update to scanner: https://github.com/stackrox/scanner/pull/604

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- [ ] ~Evaluated and added CHANGELOG entry if required~
- [ ] ~Determined and documented upgrade steps~

If any of these don't apply, please comment below.

## Testing Performed

```bash
$ oc get cm | grep trusted
trusted-ca                 1      8m4s
trusted-ca-sensor          1      7m37s

$ oc get cm trusted-ca -o yaml
apiVersion: v1
data:
  ca-bundle.crt: |
    # ACCVRAIZ1
    -----BEGIN CERTIFICATE-----
    MIIH0zCCBbugAwIBAgIIXsO3pkN/pOAwDQYJKoZIhvcNAQEFBQAwQjESMBAGA1UE
...
```

PR #3518, opened 2022-10-21
ROX-8392: Remove unused files.

porridge:
## Description

It seems these files were added in https://github.com/stackrox/stackrox/pull/174 by mistake.
They are not actually useful, because there is no assertion to check their effect.

## Checklist

## Testing Performed

Relying on CI.


0x656b694d (approved):
Thank you.

PR #3522, opened 2022-10-21
Dashrews/ROX-13100 process indicator pruning performance batches

dashrews78:
## Description

`store.go.tpl` and `store_test.go.tpl` are really the only changes.  Everything else was generated from those.

In some performance testing I would notice many occurrences of 
`delete from process_indicators where process_indicators.Id = ANY($1::text[])` timing out during pruning.  I took a look and tried to do this work with SQL, but that too would run into timeouts from time to time due to the volume of data in process indicators.  So the simplest path for now is to batch the deletes so they are more likely to succeed.  Later we may want to investigate utilizing partitions on some of these larger tables to help with the pruning jobs.

## Checklist

## Testing Performed

Lots of scale test type testing.  Set it up like this:
```
stackrox $ ./scale/dev/cluster.sh some-test-cluster

stackrox $ MAIN_IMAGE_TAG=some_image_tag ROX_POSTGRES_DATASTORE=true ./scale/dev/launch_central.sh
# Edit central config to extend the limit to 16 CPUs and 64Gi RAM
$ kubectl edit -n stackrox deploy/central
# Configure postgres compatible UI and postgres query tracing
$ kubectl -n stackrox set env deploy/central ROX_FRONTEND_VM_UPDATES=true
$ kubectl -n stackrox set env deploy/central ROX_POSTGRES_QUERY_TRACER=true
$ kubectl -n stackrox set env deploy/central ROX_POSTGRES_QUERY_TRACER_GRAPHQL_THRESHOLD=50ms
$ kubectl -n stackrox set env deploy/central ROX_POSTGRES_QUERY_TRACER_QUERY_THRESHOLD=1ms 

stackrox $ pushd scale/dev;  MAIN_IMAGE_TAG=some_image_tag ./run-many.sh 10-sensors 10; popd
```
Ran for a couple of hours and then check the central-db log to see if there were timeout errors.  Followed that up with a bounce of central to ensure it reprocessed everything and checked the logs of central-db again for timeout errors.  

Also added some UT.

### DIFF
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -55,7 +55,7 @@ const (
         // using copyFrom, we may not even want to batch.  It would probably be simpler
         // to deal with failures if we just sent it all.  Something to think about as we
         // proceed and move into more e2e and larger performance testing
-        batchSize = 5000
+        batchSize = 10000
 
         cursorBatchSize = 50
 
--- a/tools/generate-helpers/pg-table-bindings/store.go.tpl
+++ b/tools/generate-helpers/pg-table-bindings/store.go.tpl
@@ -58,6 +58,10 @@ const (
         batchSize = 10000
 
         cursorBatchSize = 50
+
+    {{- if not .JoinTable }}
+        deleteBatchSize = 25000
+    {{- end }}
 )
 
 var (
@@ -893,12 +897,32 @@ func (s *storeImpl) DeleteMany(ctx context.Context, ids []{{$singlePK.Type}}) er
     {{- end }}
     {{- end }}{{/* if not $inMigration */}}
 
-    q := search.ConjunctionQuery(
-    sacQueryFilter,
-        search.NewQueryBuilder().AddDocIDs(ids...).ProtoQuery(),
-    )
+    // Batch the deletes
+    batchSize := deleteBatchSize
+    for {
+        if len(ids) == 0 {
+            break
+        }
 
-    return postgres.RunDeleteRequestForSchema(ctx, schema, q, s.db)
+        if len(ids) < batchSize {
+            batchSize = len(ids)
+        }
+
+        idBatch := ids[0:batchSize]
+        q := search.ConjunctionQuery(
+        sacQueryFilter,
+            search.NewQueryBuilder().AddDocIDs(idBatch...).ProtoQuery(),
+        )
+
+        if err := postgres.RunDeleteRequestForSchema(ctx, schema, q, s.db); err != nil {
+            return err
+        }
+
+        // Move the slice forward to start the next batch
+        ids = ids[batchSize:]
+    }
+
+    return nil
 }
 {{- end }}
 {{- end }}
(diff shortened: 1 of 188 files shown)


tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -893,12 +897,32 @@ func (s *storeImpl) DeleteMany(ctx context.Context, ids []{{$singlePK.Type}}) er
...
+        if len(ids) == 0 {
+            break
+        }
+
+        if len(ids) < batchSize {
+            batchSize = len(ids)
+        }
 
-    return postgres.RunDeleteRequestForSchema(ctx, schema, q, s.db)
+        idBatch := ids[0:batchSize]
```
md2119, line 911: nit: `ids[:batchSize]`

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -893,12 +897,32 @@ func (s *storeImpl) DeleteMany(ctx context.Context, ids []{{$singlePK.Type}}) er
...
+        q := search.ConjunctionQuery(
+        sacQueryFilter,
+            search.NewQueryBuilder().AddDocIDs(idBatch...).ProtoQuery(),
+        )
+
+        if err := postgres.RunDeleteRequestForSchema(ctx, schema, q, s.db); err != nil {
+            return err
```
md2119, line 918: Since the `DeleteMany` operation is split into individual transactions, it would be nice to log as well as return in error how many entries were actually deleted.

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -893,12 +897,34 @@ func (s *storeImpl) DeleteMany(ctx context.Context, ids []{{$singlePK.Type}}) er
     {{- end }}
     {{- end }}{{/* if not $inMigration */}}
 
-    q := search.ConjunctionQuery(
-    sacQueryFilter,
-        search.NewQueryBuilder().AddDocIDs(ids...).ProtoQuery(),
-    )
+    // Batch the deletes
+    batchSize := deleteBatchSize
```
c-du, line 901: This is shadowing global const, rename it?

md2119 (approved):

tools/generate-helpers/pg-table-bindings/store.go.tpl
```diff
@@ -893,12 +897,34 @@ func (s *storeImpl) DeleteMany(ctx context.Context, ids []{{$singlePK.Type}}) er
...
+        }
+
+        idBatch := ids[:batchSize]
+        q := search.ConjunctionQuery(
+        sacQueryFilter,
+            search.NewQueryBuilder().AddDocIDs(idBatch...).ProtoQuery(),
+        )
+
+        if err := postgres.RunDeleteRequestForSchema(ctx, schema, q, s.db); err != nil {
+            log.Errorf("unable to delete all the records.  Successfully deleted %d out of %d", numRecordsToDelete - len(ids), numRecordsToDelete)
```
c-du, line 919: You can wrap the error first and then log  and return it. So you do not need to have the same message here twice.

tools/generate-helpers/pg-table-bindings/store_test.go.tpl
```diff
@@ -125,13 +125,15 @@ func (s *{{$namePrefix}}StoreSuite) TestStore() {
 	{{- end }}
 
 	var {{$name}}s []*{{.Type}}
+	var {{$name}}Ids []string
```
c-du, line 128: {{$name}}IDs

c-du (approved):

PR #3527, opened 2022-10-21
Dashrews/rox 12398 backup test eof (WIP)

dashrews78:
## Description

This seems like a timing issue.  When the steps before the backup were running, an action was performed that would cause central to restart.  That was on purpose and controlled.  That can leave us with a timing issue if that is still cycling through as the backup/restore test is completed.  So I added a `wait_for_api` in the restore test.  That should help in these cases and cause no harm.  However, I cannot get the test to fail in the same way and it doesn't seem to happen particularly often.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -443,6 +443,9 @@ db_backup_and_restore_test() {
     require_environment "API_ENDPOINT"
     require_environment "ROX_PASSWORD"
 
+    # Ensure central is ready for requests after any previous tests
+    wait_for_api
+
     local output_dir="$1"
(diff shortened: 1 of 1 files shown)


dashrews78:
@gavin-stackrox can you take a look at osd-aws-qa-e2e-tests, seems Shane's departure has it in a bad state.

gavin-stackrox:
(sorry, @dashrews78 I missed your comment from 12 days ago!)

gavin-stackrox (approved):
Seems reasonable to me.

PR #3528, opened 2022-10-21
ROX-12614: Collection Update Workflow

keyallis:
## Description

Implement Collection Update workflow and add unit tests to cover the expected functionality.

## Checklist

## Testing Performed

Added unit tests.

### DIFF
--- a/central/resourcecollection/datastore/datastore_impl.go
+++ b/central/resourcecollection/datastore/datastore_impl.go
@@ -106,6 +106,10 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
 	ds.graphLock.Lock()
 	defer ds.graphLock.Unlock()
 
+	if obj == nil {
+		return errors.New("passed collection must not be nil")
+	}
+
 	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
 	if err != nil {
 		return err
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
 	return nil
 }
 
+// graphEdgeUpdates stores added and removed edges from the DAG graph
+type graphEdgeUpdates struct {
+	srcId   string
+	added   []string
+	removed []string
+}
+
+func (ds *datastoreImpl) updateCollectionInGraph(obj *storage.ResourceCollection) (*graphEdgeUpdates, error) {
+	ds.initGraphOnce()
+	ds.graphLock.Lock()
+	defer ds.graphLock.Unlock()
+
+	if obj == nil {
+		return nil, errors.New("passed collection must not be nil")
+	}
+
+	// return object struct
+	ret := &graphEdgeUpdates{
+		obj.GetId(),
+		make([]string, 0),
+		make([]string, 0),
+	}
+
+	// get edges for the current object
+	curEdges, err := ds.graph.GetDescendants(obj.GetId())
+	if err != nil {
+		return nil, errors.Wrap(err, "update attempt on unknown collection")
+	}
+
+	for _, edge := range obj.GetEmbeddedCollections() {
+		_, present := curEdges[edge.GetId()]
+		if !present {
+			ret.added = append(ret.added, edge.GetId())
+		} else {
+			// if we visit an edge, set the value to nil
+			curEdges[edge.GetId()] = nil
+		}
+	}
+
+	// if a value is nil we visited it, if it's not nil we want to remove that edge
+	for key, value := range curEdges {
+		if value != nil {
+			ret.removed = append(ret.removed, key)
+		}
+	}
+
+	// since adding can cause cycles but removing never can, we remove first
+	// we iterate back to front so that if we fail a deletion we can remove the index from our returned list of deletions
+	for idx := len(ret.removed) - 1; idx >= 0; idx-- {
+		deleteErr := ds.graph.DeleteEdge(obj.GetId(), ret.removed[idx])
+		if deleteErr != nil {
+			ret.removed = append(ret.removed[:idx], ret.removed[idx+1:]...)
+			log.Errorf("Failed to remove collection edge from internal state object (%v)", deleteErr)
+		}
+	}
+
+	// add new edges
+	for idx, addId := range ret.added {
+		err = ds.graph.AddEdge(obj.GetId(), addId)
+		if err != nil {
+
+			// make note of where we stopped adding edges
+			ret.added = ret.added[:idx]
+
+			ds.undoEdgeUpdatesInGraph(ret)
+			return nil, err
+		}
+	}
+	return ret, nil
+}
+
+func (ds *datastoreImpl) undoEdgeUpdatesInGraphWithLock(updates *graphEdgeUpdates) {
+	ds.graphLock.Lock()
+	defer ds.graphLock.Unlock()
+
+	ds.undoEdgeUpdatesInGraph(updates)
+}
+
+func (ds *datastoreImpl) undoEdgeUpdatesInGraph(updates *graphEdgeUpdates) {
+
+	// ensure lock is acquired before calling this function
+	if ds.graphLock.TryLock() {
+		log.Errorf("function called without lock")
+		ds.graphLock.Unlock()
+		return
+	}
+
+	if updates == nil {
+		log.Infof("passed updates object was nil")
+		return
+	}
+
+	// remove added edges first since removing can never cause cycles
+	for _, added := range updates.added {
+		deleteErr := ds.graph.DeleteEdge(updates.srcId, added)
+		if deleteErr != nil {
+			log.Infof("tried to delete an edge that didn't exist (%v)", deleteErr)
+		}
+	}
+
+	// then restore edges that were removed
+	for _, removed := range updates.removed {
+		restoreErr := ds.graph.AddEdge(updates.srcId, removed)
+		if restoreErr != nil {
+			log.Errorf("Failed to restore collection edge in internal state object (%v)", restoreErr)
+		}
+	}
+}
+
 func (ds *datastoreImpl) deleteCollectionFromGraph(id string) error {
 	ds.initGraphOnce()
 	ds.graphLock.Lock()
@@ -135,6 +248,10 @@ func (ds *datastoreImpl) dryRunAddCollectionInGraph(obj *storage.ResourceCollect
 	ds.initGraphOnce()
 	ds.graphLock.Lock()
 	defer ds.graphLock.Unlock()
+
+	if obj == nil {
+		return errors.New("passed collection must not be nil")
+	}
 
 	var ret error
 
@@ -203,6 +320,9 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
 }
 
 func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	if collection == nil {
+		return errors.New("passed collection must not be nil")
+	}
 
 	// add to graph first to detect any cycles
 	err := ds.addCollectionToGraph(collection)
@@ -228,7 +348,30 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
 		return err
 	}
 
+	if collection == nil {
+		return errors.New("passed collection must not be nil")
+	}
+
 	return ds.dryRunAddCollectionInGraph(collection)
+}
+
+func (ds *datastoreImpl) UpdateCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	if collection == nil {
+		return errors.New("passed collection must not be nil")
+	}
+
+	// update graph first to detect cycles
+	edgeUpdates, err := ds.updateCollectionInGraph(collection)
+	if err != nil {
+		return err
+	}
+
+	err = ds.storage.Upsert(ctx, collection)
+	if err != nil {
+		// if we fail to upsert, try to restore the graph
+		ds.undoEdgeUpdatesInGraphWithLock(edgeUpdates)
+	}
+	return err
 }
 
 func (ds *datastoreImpl) DeleteCollection(ctx context.Context, id string) error {
--- a/central/resourcecollection/datastore/datastore_impl.go
+++ b/central/resourcecollection/datastore/datastore_impl.go
@@ -2,7 +2,9 @@ package datastore
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/google/uuid"
 	"github.com/heimdalr/dag"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/resourcecollection/datastore/index"
@@ -14,6 +16,7 @@ import (
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/sac"
 	pkgSearch "github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/rox/pkg/sync"
 	"github.com/stackrox/rox/pkg/utils"
 )
@@ -31,6 +34,15 @@ type datastoreImpl struct {
 
 	graphLock sync.Mutex
 	graph     *dag.DAG
+	names     set.Set[string]
+}
+
+type graphEntry struct {
+	id string
+}
+
+func (ge graphEntry) ID() string {
+	return ge.id
 }
 
 func (ds *datastoreImpl) initGraphOnce() {
@@ -47,10 +59,9 @@ func (ds *datastoreImpl) initGraph() error {
 		return nil
 	}
 
-	ds.graphLock.Lock()
-	defer ds.graphLock.Unlock()
+	// build graph object
+	graph := dag.NewDAG()
 
-	ds.graph = dag.NewDAG()
 	ctx := sac.WithAllAccess(context.Background())
 
 	// get ids first
@@ -71,7 +82,17 @@ func (ds *datastoreImpl) initGraph() error {
 			return errors.Wrap(err, "building collection graph")
 		}
 		for _, obj := range objs {
-			err = ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
+
+			// track names
+			if ds.names.Contains(obj.GetName()) {
+				return errors.New(fmt.Sprintf("encountered duplicate name building collection graph (%s)", obj.GetName()))
+			}
+			ds.names.Add(obj.GetName())
+
+			// add object
+			_, err = graph.AddVertex(graphEntry{
+				id: obj.GetId(),
+			})
 			if err != nil {
 				return errors.Wrap(err, "building collection graph")
 			}
@@ -80,42 +101,56 @@ func (ds *datastoreImpl) initGraph() error {
 
 	// then add edges by batches
 	for i := 0; i < len(ids); i += initBatchSize {
-		var objs []*storage.ResourceCollection
+		var parents []*storage.ResourceCollection
 		if i+initBatchSize < len(ids) {
-			objs, _, err = ds.storage.GetMany(ctx, ids[i:i+initBatchSize])
+			parents, _, err = ds.storage.GetMany(ctx, ids[i:i+initBatchSize])
 		} else {
-			objs, _, err = ds.storage.GetMany(ctx, ids[i:])
+			parents, _, err = ds.storage.GetMany(ctx, ids[i:])
 		}
 		if err != nil {
 			return errors.Wrap(err, "building collection graph")
 		}
-		for _, obj := range objs {
-			for _, peer := range obj.GetEmbeddedCollections() {
-				err = ds.graph.AddEdge(obj.GetId(), peer.GetId())
+		for _, parent := range parents {
+			for _, child := range parent.GetEmbeddedCollections() {
+				err = graph.AddEdge(parent.GetId(), child.GetId())
 				if err != nil {
 					return errors.Wrap(err, "building collection graph")
 				}
 			}
 		}
 	}
+
+	// set the graph object
+	ds.graph = graph
+
 	return nil
 }
 
 func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) error {
 	ds.initGraphOnce()
+
+	var err error
+
+	// input validation
+	if err = verifyCollectionObjectNotEmpty(obj); err != nil {
+		return err
+	}
+	if obj.GetId() != "" {
+		return errors.New("invalid argument, added collection must not have a pre-set `id`")
+	}
+
 	ds.graphLock.Lock()
 	defer ds.graphLock.Unlock()
 
-	if obj == nil {
-		return errors.New("passed collection must not be nil")
-	}
-
-	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
+	obj.Id, err = ds.graph.AddVertex(graphEntry{
+		id: uuid.New().String(),
+	})
 	if err != nil {
 		return err
 	}
-	for _, peer := range obj.GetEmbeddedCollections() {
-		err = ds.graph.AddEdge(obj.GetId(), peer.GetId())
+
+	for _, child := range obj.GetEmbeddedCollections() {
+		err = ds.graph.AddEdge(obj.GetId(), child.GetId())
 		if err != nil {
 			deleteErr := ds.graph.DeleteVertex(obj.GetId())
 			if deleteErr != nil {
@@ -129,18 +164,27 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
 
 // graphEdgeUpdates stores added and removed edges from the DAG graph
 type graphEdgeUpdates struct {
-	srcID   string
-	added   []string
-	removed []string
+	parentID           string
+	addedChildrenIDs   []string
+	removedChildrenIDs []string
 }
 
 func (ds *datastoreImpl) updateCollectionInGraph(obj *storage.ResourceCollection) (*graphEdgeUpdates, error) {
-	ds.initGraphOnce()
-	ds.graphLock.Lock()
-	defer ds.graphLock.Unlock()
+	return ds.updateCollectionInGraphWorkflow(obj, false)
+}
 
-	if obj == nil {
-		return nil, errors.New("passed collection must not be nil")
+func (ds *datastoreImpl) dryRunUpdateCollectionInGraph(obj *storage.ResourceCollection) error {
+	_, err := ds.updateCollectionInGraphWorkflow(obj, true)
+	return err
+}
+
+func (ds *datastoreImpl) updateCollectionInGraphWorkflow(obj *storage.ResourceCollection, dryrun bool) (*graphEdgeUpdates, error) {
+	ds.initGraphOnce()
+
+	var err error
+
+	if err = verifyCollectionObjectNotEmpty(obj); err != nil {
+		return nil, err
 	}
 
 	// return object struct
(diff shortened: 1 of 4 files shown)


central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+	if obj == nil {
+		return nil, errors.New("passed collection must not be nil")
+	}
+
+	// return object struct
+	ret := &graphEdgeUpdates{
+		obj.GetId(),
+		make([]string, 0),
+		make([]string, 0),
+	}
```
md2119, line 151: Do this before acquiring the lock

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+
+	// get edges for the current object
+	curEdges, err := ds.graph.GetDescendants(obj.GetId())
+	if err != nil {
+		return nil, errors.Wrap(err, "update attempt on unknown collection")
```
md2119, line 156: The error message could be clearer. Maybe `failed to read collection  {id x} from the graph`?

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+	}
+
+	for _, edge := range obj.GetEmbeddedCollections() {
+		_, present := curEdges[edge.GetId()]
+		if !present {
+			ret.added = append(ret.added, edge.GetId())
+		} else {
+			// if we visit an edge, set the value to nil
+			curEdges[edge.GetId()] = nil
```
md2119, line 165: `delete(currEdges, edge.GetId()`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+	for key, value := range curEdges {
+		if value != nil {
+			ret.removed = append(ret.removed, key)
+		}
+	}
+
+	// since adding can cause cycles but removing never can, we remove first
+	// we iterate back to front so that if we fail a deletion we can remove the index from our returned list of deletions
+	for idx := len(ret.removed) - 1; idx >= 0; idx-- {
+		deleteErr := ds.graph.DeleteEdge(obj.GetId(), ret.removed[idx])
```
md2119, line 179: If this fails it means that the update operation is unsuccessful, and subsequent operations will have an undesired outcome. Therefore, return err here.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+			ret.removed = append(ret.removed[:idx], ret.removed[idx+1:]...)
+			log.Errorf("Failed to remove collection edge from internal state object (%v)", deleteErr)
+		}
+	}
+
+	// add new edges
+	for idx, addID := range ret.added {
+		err = ds.graph.AddEdge(obj.GetId(), addID)
+		if err != nil {
+
```
md2119, line 190: nit : rm newline

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+		}
+	}
+
+	// since adding can cause cycles but removing never can, we remove first
+	// we iterate back to front so that if we fail a deletion we can remove the index from our returned list of deletions
+	for idx := len(ret.removed) - 1; idx >= 0; idx-- {
+		deleteErr := ds.graph.DeleteEdge(obj.GetId(), ret.removed[idx])
+		if deleteErr != nil {
+			ret.removed = append(ret.removed[:idx], ret.removed[idx+1:]...)
+			log.Errorf("Failed to remove collection edge from internal state object (%v)", deleteErr)
```
md2119, line 182: undo changes

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,6 +127,115 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+	if ds.graphLock.TryLock() {
+		log.Errorf("function called without lock")
+		ds.graphLock.Unlock()
+		return
+	}
+
+	if updates == nil {
+		log.Infof("passed updates object was nil")
+		return
+	}
```
md2119, line 220: do this before acquiring lock

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -136,6 +249,10 @@ func (ds *datastoreImpl) dryRunAddCollectionInGraph(obj *storage.ResourceCollect
 	ds.graphLock.Lock()
 	defer ds.graphLock.Unlock()
 
+	if obj == nil {
```
md2119, line 252: before lock

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -228,9 +348,32 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
...
+		return errors.New("passed collection must not be nil")
+	}
+
 	return ds.dryRunAddCollectionInGraph(collection)
 }
 
+func (ds *datastoreImpl) UpdateCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	if collection == nil {
+		return errors.New("passed collection must not be nil")
+	}
```
md2119, line 361: factor this out into `validateCollectionIsNonEmpty(...)` and replace other use instances.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -228,9 +348,32 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
...
+
+	// update graph first to detect cycles
+	edgeUpdates, err := ds.updateCollectionInGraph(collection)
+	if err != nil {
+		return err
+	}
```
md2119, line 367: not for now but this is essentially a variation to update dryrun. Dryrun is test changes + revert changes instead of commit.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -228,9 +348,32 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
...
+
+	err = ds.storage.Upsert(ctx, collection)
+	if err != nil {
+		// if we fail to upsert, try to restore the graph
+		ds.undoEdgeUpdatesInGraphWithLock(edgeUpdates)
```
md2119, line 372: This could error as well. Capture error and return. This makes another case for copy operation because once the graph is bad state there is no way to recover other than Central restart.
```
a. Obtain lock
b. Create a graph copy.
c. Do all the changes on copy.
d. If successful, `ds.graph = clonedGraph`. 
e. Release lock.
```

Note this workflow is slightly different than dry run workflow when it comes to locking.

md2119 (changes requested):

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -129,7 +129,7 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
 
 // graphEdgeUpdates stores added and removed edges from the DAG graph
 type graphEdgeUpdates struct {
-	srcId   string
+	srcID   string
```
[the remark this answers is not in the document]
keyallis: That's a good idea, much clearer to understand that naming schema

central/resourcecollection/datastore/datastore_impl.go
keyallis: I think the context of the id failed to read will be provided by the DAG layer and since this is an error wrap I just want it to provide the context of being a failed update attempt

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,38 +101,56 @@ func (ds *datastoreImpl) initGraph() error {
...
 	ds.initGraphOnce()
+
+	var err error
+
+	// input validation
+	if err = verifyCollectionObjectNotEmpty(obj); err != nil {
+		return err
+	}
+	if obj.GetId() != "" {
+		return errors.New("invalid argument, added collection must not have a pre-set `id`")
```
md2119, line 139: `added collection` -> `new collection`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,40 +162,177 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
+	}
+
+	// get current children edges for the object
+	curChildren, err := graphCopy.GetChildren(obj.GetId())
+	if err != nil {
+		return nil, errors.Wrap(err, "update attempt on unknown collection")
+	}
+
+	// determine additions
+	for _, newChildren := range obj.GetEmbeddedCollections() {
```
md2119, line 216: nit: `newChild`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -231,16 +429,72 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
...
+
+	// resolve object to check if the name was changed
+	obj, ok, err := ds.storage.Get(ctx, collection.GetId())
+	if err != nil || !ok {
+		return errors.Wrap(err, "failed to resolve collection being updated")
+	}
+	if obj.GetName() != collection.GetName() && ds.names.Contains(collection.GetName()) {
+		return fmt.Errorf("collection name already in use (%s)", collection.GetName())
+	}
+
```
md2119, line 445: Hold the lock here. Holding the lock inside `updateCollectionInGraphWorkflow` means once the graph is updated, the lock is released and the next op—add, update, delete—can continue even before we have made state changes to db. Plus, it will lead to error when you undo below.

With that, you can change the signature of `updateCollectionInGraphWorkflow` to `dryRunUpsertCollectionInGraph(obj *storage.ResourceCollection) (*dag.DAG, error)`. You set `ds.graph = graphCopy` if the db operation succeeds. I believe with this change, `undoEdgeUpdatesInGraph` is no longer needed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -231,16 +429,72 @@ func (ds *datastoreImpl) DryRunAddCollection(ctx context.Context, collection *st
...
+	err = ds.storage.Delete(ctx, id)
 	if err != nil {
 		return err
 	}
-	deleteErr := ds.deleteCollectionFromGraph(id)
-	if deleteErr != nil {
-		log.Errorf("Failed to remove collection from internal state object (%v)", deleteErr)
+
+	// update tracking collections
+	ds.names.Remove(obj.GetName())
```
md2119, line 486: you need to make ops on set thread safe. Therefore, lock above.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -203,26 +379,48 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
 
 func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	if collection == nil {
+		return errors.New("passed collection must not be nil")
+	}
+	if collection.GetId() != "" {
+		return fmt.Errorf("added collections must not have an `id` preset (%s)", collection.GetId())
+	}
+	if collection.GetName() == "" || ds.names.Contains(collection.GetName()) {
+		return fmt.Errorf("added collections must have non-empty, unique `name` values (%s)", collection.GetName())
```
md2119, line 389: nit: `added collections` -> `collections`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -123,40 +162,177 @@ func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) e
...
-	err = ds.graph.DeleteVertex(obj.GetId())
-	if err != nil && ret == nil {
-		ret = err
+	deleteErr := ds.graph.DeleteVertex(dryrunID)
+	if err != nil {
+		return err
 	}
-
-	return ret
+	return deleteErr
```
md2119, line 335: This is pretty much the same as `updateCollectionInGraphWorkflow` except that there are always no edges to delete for new collections.

md2119 (changes requested):

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -203,26 +379,48 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
 }
 
 func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
```
charmik-redhat, line 381: nit : Since it sets the Id too, would it be better for the func to return the id as well?
 
Yes, it sets it on the pointer to obj and it would be updated for the caller too, but the caller is ideally not looking at this implementation before calling `AddCollection` and just looking at datastore interface. Another way could be to add a comment on the interface that says id would be set in place.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,38 +101,56 @@ func (ds *datastoreImpl) initGraph() error {
...
 	for i := 0; i < len(ids); i += initBatchSize {
-		var objs []*storage.ResourceCollection
+		var parents []*storage.ResourceCollection
 		if i+initBatchSize < len(ids) {
-			objs, _, err = ds.storage.GetMany(ctx, ids[i:i+initBatchSize])
+			parents, _, err = ds.storage.GetMany(ctx, ids[i:i+initBatchSize])
 		} else {
-			objs, _, err = ds.storage.GetMany(ctx, ids[i:])
+			parents, _, err = ds.storage.GetMany(ctx, ids[i:])
 		}
```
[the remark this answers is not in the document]
keyallis: initBatchSize is a value defined in the package, I don't think we want to reset this value? It would be a const if it was in c

central/resourcecollection/datastore/datastore_impl.go
keyallis: Since the collection passed into the function passes by reference the Id field just gets set on the object. The way it's currently been refactored is that the Id will only get set upon the operation being successful, I'll go ahead and add some unit tests to ensure we follow that expected functionality.

I'll be sure to add a note on the interface for that. As far as whether we should also return Id I'm unsure, I'd think it's not necessary since we should be returning nil as the object in the service layer upon error anyways.

central/resourcecollection/datastore/datastore_impl.go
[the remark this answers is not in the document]
md2119: Why is it global?

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 
-	err := ds.graph.AddVertexByID(obj.GetId(), obj.GetName())
+	var err error
+
+	// input validation
+	if err = verifyCollectionObjectNotEmpty(obj); err != nil {
+		return nil, err
+	}
+	if obj.GetId() != "" {
+		return nil, errors.New("invalid argument, added collection must not have a pre-set `id`")
```
md2119, line 136: nit 1: `new collection`
nit 2: use `errox.InvalidArgs`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 	}
+
+	// set the graph object
+	ds.graph = graph
+
 	return nil
 }
 
-func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) error {
+func (ds *datastoreImpl) addCollectionToGraph(obj *storage.ResourceCollection) (*dag.DAG, error) {
```
md2119, line 126: nit 1: rename func to `addCollectionToGraphNoLock`
nit 2: add a comment that this function operates on the a copy of graph

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
+	graph, err := ds.graph.Copy()
 	if err != nil {
-		return err
+		return nil, err
+	}
+
+	// get current children edges for the object
+	curChildren, err := graph.GetChildren(obj.GetId())
+	if err != nil {
+		return nil, errors.Wrap(err, "update attempt on unknown collection")
```
md2119, line 187: nit: -> `could not update collection {id}`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 	}
-	return nil
+
+	// set id for the object and return the new graph
+	obj.Id = id
+	return graph, nil
 }
 
-func (ds *datastoreImpl) deleteCollectionFromGraph(id string) error {
+func (ds *datastoreImpl) updateCollectionInGraph(obj *storage.ResourceCollection) (*dag.DAG, error) {
```
md2119, line 166: nit 1: `updateCollectionInGraphNoLock`
nit 2: add a comment to clarify that this function operates on copy.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
+func (ds *datastoreImpl) deleteCollectionFromGraph(id string) error {
+	ds.initGraphOnce()
+
+	// this function covers removal of the vertex and any edges to or from other vertices, it is also threadsafe
+	return ds.graph.DeleteVertex(id)
 }
 
 func (ds *datastoreImpl) Search(ctx context.Context, q *v1.Query) ([]pkgSearch.Result, error) {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
```
md2119, line 234: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 // Count returns the number of search results from the query
 func (ds *datastoreImpl) Count(ctx context.Context, q *v1.Query) (int, error) {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
+
 	return ds.searcher.Count(ctx, q)
 }
 
 func (ds *datastoreImpl) SearchCollections(ctx context.Context, q *v1.Query) ([]*storage.ResourceCollection, error) {
+	ds.lock.RLock()
```
md2119, line 248: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 func (ds *datastoreImpl) Search(ctx context.Context, q *v1.Query) ([]pkgSearch.Result, error) {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
+
 	return ds.searcher.Search(ctx, q)
 }
 
 // Count returns the number of search results from the query
 func (ds *datastoreImpl) Count(ctx context.Context, q *v1.Query) (int, error) {
+	ds.lock.RLock()
```
md2119, line 241: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
 
 func (ds *datastoreImpl) SearchCollections(ctx context.Context, q *v1.Query) ([]*storage.ResourceCollection, error) {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
+
 	return ds.searcher.SearchCollections(ctx, q)
 }
 
 func (ds *datastoreImpl) SearchResults(ctx context.Context, q *v1.Query) ([]*v1.SearchResult, error) {
+	ds.lock.RLock()
```
md2119, line 255: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -80,103 +98,170 @@ func (ds *datastoreImpl) initGraph() error {
...
+	defer ds.lock.RUnlock()
+
 	return ds.searcher.SearchResults(ctx, q)
 }
 
 func (ds *datastoreImpl) Get(ctx context.Context, id string) (*storage.ResourceCollection, bool, error) {
+	ds.lock.RLock()
```
md2119, line 262: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -186,6 +271,9 @@ func (ds *datastoreImpl) Get(ctx context.Context, id string) (*storage.ResourceC
 }
 
 func (ds *datastoreImpl) Exists(ctx context.Context, id string) (bool, error) {
+	ds.lock.RLock()
```
md2119, line 274: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -194,6 +282,9 @@ func (ds *datastoreImpl) Exists(ctx context.Context, id string) (bool, error) {
 }
 
 func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage.ResourceCollection, error) {
```
md2119, line 284: nit: `GetMany`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
 
-func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+func (ds *datastoreImpl) addCollectionWorkflow(ctx context.Context, collection *storage.ResourceCollection, dryrun bool) error {
+
+	// sanity checks
+	if err := verifyCollectionObjectNotEmpty(collection); err != nil {
+		return err
+	}
+	if collection.GetId() != "" {
+		return fmt.Errorf("added collections must not have an `id` preset (%s)", collection.GetId())
```
md2119, line 303: nit 1: `errors.Errorf`
nit 2: `new collection`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	}
+
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
 
-	// add to graph first to detect any cycles
-	err := ds.addCollectionToGraph(collection)
+	// verify that the name is not already in use
+	if collection.GetName() == "" || ds.names.Contains(collection.GetName()) {
+		return fmt.Errorf("collections must have non-empty, unique `name` values (%s)", collection.GetName())
```
md2119, line 311: nit: `errors.Errorf`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -194,6 +282,9 @@ func (ds *datastoreImpl) Exists(ctx context.Context, id string) (bool, error) {
 }
 
 func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage.ResourceCollection, error) {
+	ds.lock.RLock()
```
md2119, line 285: locking is not required because neither the graph nor the name set is accessed.

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
-	return err
+
+	// we've succeeded, now set all the values
+	ds.names.Add(collection.GetName())
+	ds.graph = graph
+	return nil
+}
+
+func (ds *datastoreImpl) AddCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	return ds.addCollectionWorkflow(ctx, collection, false)
```
md2119, line 339: check sac

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	// resolve object to check if the name was changed
+	obj, ok, err := ds.storage.Get(ctx, collection.GetId())
+	if err != nil || !ok {
+		return errors.Wrap(err, "failed to resolve collection being updated")
+	}
+	if obj.GetName() != collection.GetName() && ds.names.Contains(collection.GetName()) {
+		return fmt.Errorf("collection name already in use (%s)", collection.GetName())
```
md2119, line 371: nit 1: `errors.Errorf`
nit 2: `collection name %q already in use`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	err = ds.storage.Upsert(ctx, collection)
+	if err != nil {
+		return err
+	}
+
+	// success, we now update objects
+	if obj.GetName() != collection.GetName() {
+		ds.names.Remove(obj.GetName())
+		ds.names.Add(collection.GetName())
+	}
```
md2119, line 395: Add returns a bool to indicate if the arg is new or existing.
```
if ds.names.Add(collection.GetName()) {
 ds.names.Remove(obj.GetName())
}
```

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	if collection.GetId() == "" {
+		return errors.New("update must be called on an existing collection")
+	}
+
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	// resolve object to check if the name was changed
+	obj, ok, err := ds.storage.Get(ctx, collection.GetId())
```
md2119, line 366: nit: `obj` -> `storedCollection`

central/resourcecollection/datastore/datastore_impl.go
```diff
@@ -202,45 +293,144 @@ func (ds *datastoreImpl) GetBatch(ctx context.Context, ids []string) ([]*storage
...
+	if obj.GetName() != collection.GetName() {
+		ds.names.Remove(obj.GetName())
+		ds.names.Add(collection.GetName())
+	}
+	ds.graph = graph
+	return nil
+}
+
+func (ds *datastoreImpl) UpdateCollection(ctx context.Context, collection *storage.ResourceCollection) error {
+	return ds.updateCollectionWorkflow(ctx, collection, false)
```
md2119, line 401: sac check

md2119 (commented):
Much simplified with the use of `Copy()`. Thanks for doing the iteration.

central/resourcecollection/datastore/datastore_impl.go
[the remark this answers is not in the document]
keyallis: It's a configurable value defined at the top of the package. Is this not an idiomatic way of handling the idea of configurable values in go? We define a number of objects we want to handle as a batch size because of the concern that we might hit a point where querying all will consume too much memory. If in the future we decide we want to raise this number we only have to change the value in one place. If I move this value under a `const()` block would that make it better?

central/resourcecollection/datastore/datastore_impl.go
keyallis: None of the dag functions utilize locking now, seems irrelevant to make a note of it since the locking is expected at the layer above. There also wouldn't ever be a reason to lock this function now that it's refactored to use copy since the copy operation is threadsafe and we never touch the top level graph again in the function after calling copy

central/resourcecollection/datastore/datastore_impl.go
keyallis: Do we not have to worry about concurrency at the postgres level?

central/resourcecollection/datastore/datastore_impl.go
keyallis: This operation will not succeed unless it passes a call to postgres' `Upsert`, which will do the sac check for us

central/resourcecollection/datastore/datastore_impl.go
keyallis: Same comment as on the AddCollection function

central/resourcecollection/datastore/datastore_impl.go
md2119: I see no harm in bringing in clarity since the expectation is not at all clear RN. Refer to some examples (grep `NoLock` on codebase.

central/resourcecollection/datastore/datastore_impl.go
md2119: `SELECT` queries are implicitly wrapped in a transaction and work on the last committed state.

central/resourcecollection/datastore/datastore_impl.go
md2119: The pre-req steps, which are blocking in nature, are executed un-necessarily before the postgres check happens. I strongly recommend the workflows be such that they avoid blocking when possible.

central/resourcecollection/datastore/datastore_impl.go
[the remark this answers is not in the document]
md2119: Fair enough, although this var is meant only for one time use, that is during init, unlike `batchSize` constant in other places in code which are used across multiple ops. 

Defining `initBatchSize` communicates more clearly.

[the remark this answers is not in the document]
keyallis: I went ahead and made it `graphInitBatchSize` just for extra clarity

md2119 (approved):

PR #3532, opened 2022-10-24
ROX-13226: Improve handling of NVD CVEs

mtodor:
## Description

This PR improves the handling of loaded NVD CVEs. It covers missing `nil` checks.

Schema comes from our fork of `facebookincubator/nvdtools`. They are using the following schema to generate Go structures: https://csrc.nist.gov/schema/nvd/feed/1.0/nvd_cve_feed_json_1.0.schema 

A lot of fields that we are using are not required. That's why I went the route of checking in the code that fields relevant to us exist.

I have also added a unit test that checks many combinations of missing data or empty arrays. Please read the comment in the test file for more information.

**Changes**

1. Remove not used function and test for it (`NvdCVEToProtoCVE `)
2. Add new `nil` checks
3. Add unit test with a created list of potential data corruptions

**Considerations**

We can also add a panic handler. The best place would be to ensure that `NvdCVEToEmbeddedCVE` does not panic because we have 3 places where this function is called.

It would look something like this:
```
func NvdCVEToEmbeddedCVE(nvdCVE *schema.NVDCVEFeedJSON10DefCVEItem, ct CVEType) (cve *storage.EmbeddedVulnerability, retErr error) {
	defer func() {
		if r := recover(); r != nil {
			cve = nil
			retErr = errors.Errorf("panic running NvdCVEToEmbeddedCVE for type %d: %v", ct, r)
		}
	}()
```

## Checklist

## Testing Performed

1. CI pipeline
4. Added unit test


mtodor:
* master
  * **PR #3532** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3532" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3532?utm_source=stack-comment).

misberner (approved):
The changes look good, but just to be on the safe side, can you combine this with `safe.RunE` or so?

mtodor:
@misberner I'll merge this PR as it is and open a new one with the panic handler.

PR #3533, opened 2022-10-24
ROX-12835: Add support for NodeScanV2 to Sensor

vikin91:
## Description

This PR enriches sensor with support to handling the NodeScanV2 messages. 
In the current scope, the messages should be forwarded to Central without any manipulation (passthrough).
This PR also adds an empty Central pipeline that receives the NodeScanV2 messages from sensor and prints a log line that the message was received.

All these parts should work only when the feature-flag `ROX_RHCOS_NODE_SCANNING=true` is enabled.

## Checklist

## Testing Performed

- CI
- Run it locally in Colima and saw the message propagate from Compliance through Sensor to Central (requires #3399)


Maddosaurus (approved):
I've discussed this code with Piotr in a call already, so I am happy to approve. These are a nitpick and a question that I had re-reading the code

sensor/kubernetes/sensor/sensor.go
```diff
@@ -91,6 +91,7 @@ func CreateSensor(cfg *CreateOptions) (*sensor.Sensor, error) {
 	auditLogEventsInput := make(chan *sensorInternal.AuditEvents)
 	auditLogCollectionManager := compliance.NewAuditLogCollectionManager()
 
+	indicators := make(chan *central.MsgFromSensor)
```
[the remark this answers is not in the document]
vikin91: I think I had a variant of a solution that was using this channel. But this is a leftover from it and it can be moved back to where it was.

vikin91:
* master
  * **PR #3533** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3533" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3533?utm_source=stack-comment).

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func (c *nodeScanHandlerImpl) run() {
+	defer c.stoppedC.Signal()
+	for {
+		select {
+		case <-c.stopC.Done():
+			c.stoppedC.SignalWithError(c.stopC.Err())
+
```
msugakov, line 53: I think I understand that the intent of this code is to arrange a graceful shutdown.
`c.stopC` is when the shutdown was requested. `c.stoppedC` is when the handler actually shut down.

I understand that you likely copied the pattern found in `command_handler_impl.go`, but I think the pattern is wrong.

This specific fragment will just make `for` loop repeat quickly eating up CPU until `c.nodeScans` channel is closed.

I believe the implementation should be fixed here and propagated to other handlers.

I find https://github.com/stackrox/stackrox/blob/master/sensor/common/sensor/central_sender_impl.go to be more clean. Hopefully it can serve as a reference implementation.

A unit test for this handler will be very handy to demonstrate and validate its behavior.

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,81 @@
...
+			// Do something with the scan if needed, e.g. attach NodeID
+			c.sendScan(scan)
+		}
+	}
+}
+
+func (c *nodeScanHandlerImpl) sendScan(scan *storage.NodeScanV2) {
+	select {
+	case <-c.stoppedC.Done():
+		log.Errorf("failed to send update: %s", proto.MarshalTextString(scan))
```
msugakov, line 68: This will log the entire `scan` message which can be quite big. Maybe it is enough to just log that a node scan for specific node name will not be sent.

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+			// Do something with the scan if needed, e.g. attach NodeID
+			c.sendScan(scan)
+		}
+	}
+}
+
+func (c *nodeScanHandlerImpl) sendScan(scan *storage.NodeScanV2) {
+	select {
+	case <-c.stoppedC.Done():
```
msugakov, line 67: The idea of this case is not very clear to me. When the stop intent is registered, we eat up messages from `c.nodeScans` but drop them on the floor instead of sending to `c.toCentral`. Is that your idea? Would it make sense to just not read messages from `c.nodeScans` at all instead of dropping them?

sensor/common/compliance/service.go
```diff
@@ -21,13 +22,15 @@ type Service interface {
 
 	Output() chan *compliance.ComplianceReturn
 	AuditEvents() chan *sensor.AuditEvents
+	NodeScansV2() <-chan *storage.NodeScanV2
```
msugakov, line 25: Maybe just `NodeScans`?
```suggestion
	NodeScans() <-chan *storage.NodeScanV2
```

central/sensor/service/pipeline/nodescansv2/pipeline.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/stackrox/rox/pkg/metrics"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
+
+// Template design pattern. We define control flow here and defer logic to subclasses.
+//////////////////////////////////////////////////////////////////////////////////////
+
```
msugakov, line 22: This comment does not make anything clear-er, for me. Unless you see a value in it, I suggest removing it.
```suggestion
```

sensor/common/compliance/service.go
Maddosaurus: In regards to V2: You're absolutely right about V1 not being a real NodeScan ([src](https://github.com/stackrox/stackrox/pull/3533#discussion_r1010714715)), but as the existing feature is called NodeScan in every piece of currently used code, it might be very confusing to introduce a completely different thing under the same name, right?

vikin91: I will make a proposal to clear up the naming and their meaning, as I see that it makes even more problems when navigating though the code of Central and Scanner.

msugakov: I don't mind keeping V2. In this case I'd vote to have V2 more consistently. There are some spots in this PR where it could be (re-)added.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: > I find https://github.com/stackrox/stackrox/blob/master/sensor/common/sensor/central_sender_impl.go to be more clean. Hopefully it can serve as a reference implementation.

I would love to follow up on this, as I see there several issues that makes the concurrent code difficult to follow. These are not no-goes, but some patterns that I learnt are violated. Examples:
- `forwardResponses` writes to a channel `to` that is passed as an external parameter and never closes the channel to signal that there will no more messages.
- `componentMsgsC` is never closed, so why to have the code that handles the situation when the channel is closed?
- `forwardResponses` stops working as soon as `for !s.stopC.IsDone() {` - this is the same situation as here, we drop messages on the floor instead of forwarding them, as we do not have any guarantee that the stream will still be up. 

I am not saying that the code is not working, but I have several rules that I follow when working with concurrency:
- the goroutine that writes to a channel is responsible for creating the channel and closing it,
- only the writer knows that there will be no more messages sent to a channel. If that happens, the channel should be closed to inform all readers that there will be no more messages, 
- there is exactly one writer goroutine for a channel, but there may be many readers.

Having said that, I am curious to learn why do you think that the code `sensor/common/sensor/central_sender_impl.go` is cleaner.

> This specific fragment will just make for loop repeat quickly eating up CPU until c.nodeScans channel is closed.

Yes, this is an issue and should be addressed. I guess adding `return` should do the job, but I will check that.

> A unit test for this handler will be very handy to demonstrate and validate its behavior.

Agreed! Will try to produce some tests for this.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: This was a copy-paste, but I think that we do not send the message because it may happen that the stream will be  already closed when the context is canceled. 
I think the risk here is that the `c.toCentral` channel may not have any readers attached when `c.stoppedC.Done()` happens. This could lead to the `c.toCentral <- &central.MsgFromSensor{` blocking despite the context has been already canceled.

I think this deserves an experiment to be sure, WDYT?

pkg/centralsensor/caps_list.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// ScopedImageIntegrations identifies the capability to have image integrations with sources from image pull secrets
 	ScopedImageIntegrations SensorCapability = "ScopedImageIntegrations"
+
+	// NodeScanningV2 identifies the capability to handle compliance messages of type NodeScanV2
+	NodeScanningV2 SensorCapability = "NodeScanningV2"
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (c *nodeScanHandlerImpl) Stop(err error) {
+	c.stopC.SignalWithError(err)
+}
+
+func (c *nodeScanHandlerImpl) Stopped() concurrency.ReadOnlyErrorSignal {
+	return &c.stoppedC
+}
+
+func (c *nodeScanHandlerImpl) ProcessMessage(msg *central.MsgToSensor) error {
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,81 @@
...
+		select {
+		case <-c.stopC.Done():
+			c.stoppedC.SignalWithError(c.stopC.Err())
+
+		case scan, ok := <-c.nodeScans:
+			if !ok {
+				c.stoppedC.SignalWithError(errors.New("channel receiving node scans v2 is closed"))
+				return
+			}
+			// Do something with the scan if needed, e.g. attach NodeID
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_impl.go
vikin91: Done

central/sensor/service/pipeline/nodescansv2/pipeline.go
vikin91: Done

sensor/common/compliance/nodescan_handler_impl.go
msugakov: Your notes about channels are likely right. I did not check how channels are handled but I definitely saw that some of them aren't closed.

I brought `central_sender_impl.go` because it more accurately deals with `c.stopC` and `c.stoppedC` than this code. Particularly, there should be only one place that sets `c.stoppedC` - the deferred hook. Everything else like error conditions or closed channels should just initiate the shut down by setting `c.stopC`. Signalling that the shutdown _completed_ (which `c.stoppedC` is supposed to do) in the middle of the function seems wrong to me.

sensor/common/compliance/nodescan_handler_impl.go
msugakov: I guess my question is more about what should we expect from `c.nodeScans`. Depending on that we can make decisions what to do with `c.toCentral`.

Also, I wasn't sure if we should rather handle `c.stopC.Done()` instead of `c.stoppedC.Done()` here.

vikin91: I will look into `stopC` and `stoppedC` (including your [other comment](https://github.com/stackrox/stackrox/pull/3533#discussion_r1011511017)).

`c.nodeScans` is created in the compliance service `compliance/service_impl.go` and it receives all messages of type `*sensor.MsgFromCompliance_NodeScanV2` directly from the stream (`msg, err := server.Recv()`).
Unfortunately, `c.nodeScans` is never closed, as we do not have a function that would be called when the service is shutdown. We do not want to close the channel when the stream is closed, because the stream can be reopened. We should close the channel on sensor shutdown, but the API of `type Service interface` does not have any method that can be called to signal this.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: This should be addressed in [2077d2a](https://github.com/stackrox/stackrox/pull/3533/commits/2077d2a1e49c6e17960db9da061492231098b338)

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I played a bit with stopping in the test that I added in [ab64a00](https://github.com/stackrox/stackrox/pull/3533/commits/ab64a0004f037dbfaeade9b899dc58b3270b4781). This allowed me to simplify the `sendScan`, so that we are not dropping anything. 

I will give it a test in a local cluster, but the unit test that I added looks good!

vikin91: > I will give it a test in a local cluster, but the unit test that I added looks good!

I have discovered no issues.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+	numReceivedAtCentral int
+	cancel               context.CancelFunc
+	ctx                  context.Context
+	toCentral            chan *central.MsgFromSensor
+}
+
+func (s *NodeScanHandlerTestSuite) SetupTest() {
+	s.mu = &sync.Mutex{}
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
msugakov, line 91: Curious: why locking mutex in `SetupTest()`?

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+// TestStopHandler goal is to stop handler while there are still some messages to process
+// in the channel passed into NewNodeScanHandler.
+// We expect that premature stop of the handler produces no race condition or goroutine leak.
+// Exec with: go test -race -count=1 -v -run ^TestNodeScanHandler$ ./sensor/common/compliance
+func (s *NodeScanHandlerTestSuite) TestStopHandler() {
+	defer s.cancel()
+	numScans := 1000
+	// nodeScansProducerThatStops stops the handler after producing the first message.
+	p := newNodeScansProducerThatStops(numScans)
```
msugakov, line 142: It seems to me that `nodeScansProducerThatStops` can be simply inlined thus reducing mental overhead.

```go
func (s *NodeScanHandlerTestSuite) TestStopHandler() {
	defer s.cancel()
	numScans := 1000

	nodeScans := make(chan *storage.NodeScanV2)
	defer close(nodeScans)

	h := NewNodeScanHandler(nodeScans)

	go func(numMsgs int) {
		for i := 0; i < numMsgs; i++ {
			nodeScans <- fakeNodeScanV2(fmt.Sprintf("Bobby_%d", i))
			if i == 0 {
				h.Stop(nil)
			}
		}
	}(numScans)
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+	s.startFakeCentral()
+	s.startForwardingToCentral(h.ResponsesC())
+
+	err := h.Start()
+	assert.NoError(s.T(), err)
+	<-h.Stopped().Done()
+	assert.True(s.T(), h.Stopped().IsDone())
+
+	// give central a chance to receive something before we lock s.mu
+	<-time.After(50 * time.Millisecond)
```
msugakov, line 154: This seems brittle. What if after a bad refactoring (in the future) you give it more time and it receives all 1000 messages? What if 50 milliseconds are not enough to receive even 1 message on some systems?
I have no concrete ideas for improvement but it feels the test tries to do too much.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+				s.numReceivedAtCentral++
+				s.mu.Unlock()
+			}
+		}
+	}(s.ctx)
+}
+
+// TestStopHandler goal is to stop handler while there are still some messages to process
+// in the channel passed into NewNodeScanHandler.
+// We expect that premature stop of the handler produces no race condition or goroutine leak.
```
msugakov, line 136: How come no goroutine leaks if the one started by `nodeScansProducerThatStops.start()` should hang and leak after sending the first few items over its channel?

msugakov (approved):
Nothing blocking the PR, mostly nitpicks in attempt to understand the existing design and how what you introduce fits in it. I'm ok if this change gets merged as-is and you address my comments in a follow-up.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Only to prevent race in accessing `numReceivedAtCentral`. I will try to optimize it out, as the mutex can hide some other potential races.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Correct. I fixed that in a bit less flaky (but not ideal) way.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Correct, it leaked. This should be fixed now (checked with https://github.com/uber-go/goleak, which I cannot include as the test logger used here leaks one goroutine)

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
+	// This component doesn't actually process or handle any messages sent from Central to Sensor (yet).
+	// It uses the sensor component so that the lifecycle (start, stop) can be handled when Sensor starts up.
+	return nil
+}
+
+func (c *nodeScanHandlerImpl) run() {
+	defer c.stoppedC.SignalWithError(c.stopC.Err())
```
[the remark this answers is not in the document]
vikin91: Got it covered!

I am only not fully sure which goroutine is meant with _"eliminate one forwarding goroutine"_ 🤔

sensor/common/compliance/nodescan_handler_test.go
vikin91: I played more with the leaks and the `goleak` from Uber. I was able to eliminate all leaks in the code (test & feat). I hope it is okay to add https://github.com/uber-go/goleak as a new dependency.

vikin91:
Rebasing against master

sensor/common/compliance/service.go
vikin91: Follow-up to the naming discussion: https://github.com/stackrox/stackrox/pull/3755

sensor/common/compliance/nodescan_handler.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/sensor/common"
+)
+
+// NodeScanHandler is responsible for handling the arriving NodeScanV2 messages, processing then, and sending them to central
+type NodeScanHandler interface {
+	Stopped() concurrency.ReadOnlyErrorSignal
```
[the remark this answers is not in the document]
vikin91: It is gone now.

sensor/common/compliance/nodescan_handler_test.go
vikin91: The mutex and the counter var `numReceivedAtCentral` are gone now.

pkg/metrics/resources.go
```diff
@@ -1,6 +1,7 @@
 package metrics
 
 // Resource represents the resource that we want to time.
+//
```
RTann, line 4: I think this is a go1.19 thing. Since we are updating the file anyway, might as well keep it, but I have also had the displeasure of dealing with this when trying to style this repo

sensor/common/compliance/service_impl.go
```diff
@@ -143,7 +145,9 @@ func (s *serviceImpl) Communicate(server sensor.ComplianceService_CommunicateSer
 			s.auditEvents <- t.AuditEvents
 			s.auditLogCollectionManager.AuditMessagesChan() <- msg
 		case *sensor.MsgFromCompliance_NodeScanV2:
-			log.Infof("NodeScanV2 message received: %v", msg)
```
RTann, line 146: Up to you, but want to keep this as a Debug log if that'll help development?

sensor/common/compliance/service.go
```diff
@@ -21,13 +22,15 @@ type Service interface {
 
 	Output() chan *compliance.ComplianceReturn
 	AuditEvents() chan *sensor.AuditEvents
+	NodeScans() <-chan *storage.NodeScanV2
```
RTann, line 25: curious, do we ever send to the other channels? why is this the only receive-only channel?

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/goleak"
+)
+
+func assertNoGoroutineLeaks(t *testing.T) {
+	goleak.VerifyNone(t,
+		// Ignore a known leak: https://github.com/DataDog/dd-trace-go/issues/1469
+		goleak.IgnoreTopFunction("github.com/golang/glog.(*loggingT).flushDaemon"),
```
RTann, line 19: uh oh. is this a bad leak? I see it's unresolved, but I wonder how much this affects us

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+	err := h.Start()
+	assert.NoError(s.T(), err)
+}
+
+func (s *NodeScanHandlerTestSuite) TestRestartHandler() {
+	nodeScans := make(chan *storage.NodeScanV2)
+	defer close(nodeScans)
+	h := NewNodeScanHandler(nodeScans)
+
+	assert.NoError(s.T(), h.Start())
```
RTann, line 128: since we have a `suite`, maybe we should use `s.Assert().NoError(h.Start())`? Doesn't really matter tbh, though

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+		}
+	}()
+}
+
+// TestStopHandler goal is to stop handler while there are still some messages to process
+// in the channel passed into NewNodeScanHandler.
+// We expect that premature stop of the handler produces no race condition or goroutine leak.
+// Exec with: go test -race -count=1 -v -run ^TestNodeScanHandler$ ./sensor/common/compliance
+func (s *NodeScanHandlerTestSuite) TestStopHandler() {
```
RTann, line 90: How is this run in CI?

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return errors.New("stopped handlers cannot be restarted")
+}
+
+func (c *nodeScanHandlerImpl) Stop(err error) {
+	c.stopC.SignalWithError(err)
+}
+
+func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
+	// This component doesn't actually process or handle any messages sent from Central to Sensor (yet).
```
RTann, line 41: Does this imply there is a plan to add this (I'm imagining Central requesting a scan periodically as part of reprocessing or user forcing it)?

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *nodeScanHandlerImpl) Capabilities() []centralsensor.SensorCapability {
+	return []centralsensor.SensorCapability{centralsensor.NodeScanningCap}
+}
+
+func (c *nodeScanHandlerImpl) ResponsesC() <-chan *central.MsgFromSensor {
+	return c.toCentral
+}
+
+func (c *nodeScanHandlerImpl) Start() error {
```
RTann, line 27: Do we want to allow multiple `Start` calls to happen?

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,82 @@
...
+				return
+			}
+			// TODO(ROX-12943): Do something with the scan, e.g., attach NodeID
+			c.sendScan(scan)
+		}
+	}
+}
+
+func (c *nodeScanHandlerImpl) sendScan(scan *storage.NodeScanV2) {
+	if scan != nil {
```
RTann, line 66: up to you, but maybe we can do something like the following to remove all the indentation:

```
if scan == nil {
        return
}
...
```

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,82 @@
...
+		case c.toCentral <- &central.MsgFromSensor{
+			Msg: &central.MsgFromSensor_Event{
+				Event: &central.SensorEvent{
+					Resource: &central.SensorEvent_NodeScanV2{
+						NodeScanV2: scan,
+					},
+				},
+			},
+		}:
+			return
```
RTann, line 79: do we need the `return` here and for the other case?

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/stackrox/rox/pkg/centralsensor"
+	"github.com/stackrox/rox/pkg/concurrency"
+)
+
+type nodeScanHandlerImpl struct {
+	nodeScans <-chan *storage.NodeScanV2
+	toCentral chan *central.MsgFromSensor
+
+	stopC    concurrency.ErrorSignal
+	stoppedC concurrency.ErrorSignal
```
RTann, line 16: what is this used for?

pkg/metrics/resources.go
vikin91: Yeah, I saw that and reverted to go 1.18

sensor/common/compliance/service_impl.go
vikin91: We will work more on the sensor corner, so there will be a chance to add some more meaningful debugging

sensor/common/compliance/service.go
vikin91: I am following a pattern where I explicitly identify writers and readers by using general channel in the first case and the read-only in the second case. Here, this helps us to discover early potential errors when the reader would try to write to a channel.

sensor/common/compliance/nodescan_handler_test.go
vikin91: I didn't dig deeper, but there is some info on the link added. I think the leak is in a dependency of a dependency  of a dependency and it may be tricky to get rid of.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Oh, sorry, you are right, this would look nicer. I spotted your review first after merging, but I will apply the suggestions in the next PR in the stack (https://github.com/stackrox/stackrox/pull/3755) or in a separate PR.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Good question. I need to check whether this is included in the race tests.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: This is added to implement the handler interface. It does not meant that we will support anything arriving from central, but we do have such possibility of we want to :)

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I investigated and this will cause a panic due to closing a channel twice (only when the feature-flag is enabled). I will provide a fix and a test for this.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: This seems to be a leftover.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: No, no need for them. I will remove them.

PR #3534, opened 2022-10-24
ROX-12948 Test_DeploymentExposure: provided port is already allocated fix

lvalerom:
## Description

Waiting for the resources to get Deleted is not enough, k8s reports that the resource has been deleted but on creation sometimes we still get the same error. Adding retries on creation helped a lot but it's still not enough. We ended up using different ports in each test to avoid this error. 

## Checklist

## Testing Performed

Manual testing and CI

```
go test -race -count=1 github.com/stackrox/rox/sensor/tests/resource/service
```


sensor/tests/resource/service/service_test.go
```diff
@@ -200,27 +207,48 @@ func (s *DeploymentExposureSuite) Test_NodePortPermutation() {
...
+		})
 }
 
-func (s *DeploymentExposureSuite) Test_LoadBalancerPermutation() {
-	s.testContext.RunWithResourcesPermutation(
-		[]resource.YamlTestFile{
-			NginxDeployment,
-			NginxServiceLoadBalancer,
-		}, "LoadBalancer", func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
+func (s *DeploymentExposureSuite) Test_NodePortPermutation() {
```
fredrb, line 224: Is there a way to write this as a table test? Maybe with a test case struct like: 
```go
type testCase struct {
    port int
    orderedResources []resource.YamlTestFile
}
```

fredrb: I think it would also make sense to write a comment on why we need to use different port numbers. Like you wrote in the PR description.

sensor/tests/resource/service/yaml/nginx-service-node-port-30007.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: nginx-svc-node-port
+  name: nginx-svc-node-port-30007
```
fredrb, line 4: This is very annoying. Maybe not for this PR if this generates some CI failures, but it would be great to have a template for this other than having to copy the file content.

Or, we could also offer the possibility to create resources using Go structs as opposed to yaml files.

sensor/tests/resource/helper.go
```diff
@@ -54,6 +54,18 @@ type YamlTestFile struct {
 	File string
 }
 
+// awaitedResources slice of resources that need to be awaited
```
fredrb, line 57: nit: `awaited` seems like the wrong word here. Maybe `waited` or `required` would fit better what you mean.

sensor/tests/resource/helper.go
```diff
@@ -54,6 +54,18 @@ type YamlTestFile struct {
...
+var awaitedResources = []string{"Service"}
+
+func isAwaitedResource(kind string) bool {
```
fredrb, line 60: nit: Just like the comment above, I think this one can be renamed to `shouldWaitForResource(...)`. It reads better in the code.

fredrb (changes requested):
Unless I misunderstood something critical, I think we can simplify the backoff logic to avoid the logic (and parameters) in the `execWithRetry` function.

The rest are suggestions to improve the code.

sensor/tests/resource/helper.go
```diff
@@ -402,8 +414,14 @@ func (c *TestContext) ApplyFile(ctx context.Context, ns string, file YamlTestFil
 		return nil, err
 	}
 
-	if err := c.r.Create(ctx, obj); err != nil {
-		return nil, err
+	if isAwaitedResource(file.Kind) {
+		if err := execWithRetry(5*time.Hour, createResourceFunc(ctx, c.r, obj)); err != nil {
```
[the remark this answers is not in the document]
lvalerom: Oops... that was supposed to be 5 minutes 😅

sensor/tests/resource/service/service_test.go
lvalerom: Done.

sensor/tests/resource/service/yaml/nginx-service-node-port-30007.yaml
lvalerom: My thoughts exactly, but I think we should do that in a different PR. I wanted to time box this fix to focus in other tasks.

sensor/tests/resource/helper.go
lvalerom: I went with `required` for some reason `waited` doesn't sound good to me

sensor/tests/resource/helper.go
lvalerom: Same as above

sensor/tests/resource/helper.go
```diff
@@ -413,10 +431,49 @@ func (c *TestContext) ApplyFile(ctx context.Context, ns string, file YamlTestFil
...
+	return func(sleepTime time.Duration) error {
+		if err := r.Create(ctx, obj); err != nil {
+			time.Sleep(sleepTime)
+			return err
+		}
+		return nil
+	}
+}
+
+func execWithRetry(timeout time.Duration, fn waitFunc) error {
```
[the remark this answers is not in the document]
lvalerom: 1. Makes sense, I was supposed to rename the function to `execFunc` or something like that. 
2. Good catch

I didn't know about that backoff library. Everything makes sense; I've changed the code to use it, but I kept the parameters to the `execWithRetry` function just in case we want to use it with different timeouts and intervals. Let me know what you think.

sensor/tests/resource/helper.go
lvalerom: Done

fredrb (approved):
Sorry @lvalerom, I completely forgot about coming back to this one! Looks good now, thanks for making the changes.

PR #3536, opened 2022-10-24
CI: Move the upgrade test to 3.69.x

gavin-stackrox:
## Description

Per title. Upgrading from 3.63.x is not very relevant. This also fixes the upgrade-test as it was the `scanImage()` in `3.63.x` Groovy tests that failed most often (once the OOMs were fixed). ~Also excluded a policy in the default policy comparison due to ROX-13265.~

## Checklist

## Testing Performed

CI is sufficient.
~/test gke-upgrade-tests~ Don't need this now that openshift/release is configured to test on changes to the test 🎉

### DIFF
--- a/tests/upgrade/run.sh
+++ b/tests/upgrade/run.sh
@@ -371,9 +371,10 @@ test_upgrade_paths() {
 
     local log_output_dir="$1"
 
-    EARLIER_SHA="9f82d2713cfec4b5c876d8dc0149f6d9cd70d349"
-    EARLIER_TAG="3.63.x-163-g2c4fe1563c"
+    EARLIER_SHA="83963e020fe08e61c9d368a42e4d23bfc22ba8f2"
+    EARLIER_TAG="3.70.x-1-g83963e020f"
     FORCE_ROLLBACK_VERSION="$EARLIER_TAG"
+    export FORCE_ROLLBACK_VERSION
 
     cd "$REPO_FOR_TIME_TRAVEL"
     git checkout "$EARLIER_SHA"
@@ -389,14 +390,14 @@ test_upgrade_paths() {
     kubectl -n stackrox set image deploy/central "central=$REGISTRY/main:$(make --quiet tag)"
     wait_for_api
 
-    validate_upgrade "00-3-63-x-to-current" "central upgrade to 3.63.x -> current" "268c98c6-e983-4f4e-95d2-9793cebddfd7"
+    validate_upgrade "00-3-70-x-to-current" "central upgrade to 3.70.x -> current" "268c98c6-e983-4f4e-95d2-9793cebddfd7"
 
     force_rollback
     wait_for_api
 
     cd "$REPO_FOR_TIME_TRAVEL"
 
-    validate_upgrade "01-current-back-to-3-63-x" "forced rollback to 3.63.x from current" "268c98c6-e983-4f4e-95d2-9793cebddfd7"
+    validate_upgrade "01-current-back-to-3-70-x" "forced rollback to 3.70.x from current" "268c98c6-e983-4f4e-95d2-9793cebddfd7"
 
     cd "$TEST_ROOT"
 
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/UpgradesTest.groovy
```diff
@@ -320,7 +325,9 @@ class UpgradesTest extends BaseSpecification {
         "Upgraded policies should match the default policies in code"
         upgradedPolicies.forEach {
             def defaultPolicy = defaultPolicies[it.id]
-            assert it == defaultPolicy
+            if (!BROKEN_POLICIES.contains(it.id)) {
+                assert it == defaultPolicy
+            }
```
janisz, line 330: How about

```groovy
        upgradedPolicies
                .findAll { !BROKEN_POLICIES.contains(it.id) }
                .forEach {
                    def defaultPolicy = defaultPolicies[it.id]
                    assert it == defaultPolicy
                }             
```

or even 
```groovy
        "Upgraded policies should match the default policies in code"
        def upgradedPoliciesFiltered = upgradedPolicies.findAll { !BROKEN_POLICIES.contains(it.id) }.toSet()
        def defaultPoliciesFiltered = defaultPolicies.values().findAll { !BROKEN_POLICIES.contains(it.id) }.toSet()
        assert upgradedPoliciesFiltered == defaultPoliciesFiltered
```
or maybe don't use map at all
```groovy
        def defaultPoliciesFiltered = defaultPolicies.values().collect { policy ->
            def builder = PolicyOuterClass.Policy.newBuilder(policy)
            if (policy.exclusionsList != null || !policy.exclusionsList.isEmpty()) {
                builder.clearExclusions().addAllExclusions(
                        policy.exclusionsList.sort(false) { it.name }
                )
            }
            builder.build()
        }.findAll { notBrokenPolicy(it) }

        then:
        "All default policies must still exist"
        assert upgradedPolicies.size() >= defaultPolicies.size()

        and:
        "Upgraded policies should match the default policies in code"
        def upgradedPoliciesFiltered = upgradedPolicies.findAll { notBrokenPolicy(it) }.toSet()
        assert upgradedPoliciesFiltered == defaultPoliciesFiltered
    }
    
    boolean notBrokenPolicy(PolicyOuterClass.Policy p) {
        !BROKEN_POLICIES.contains(p.id)
    }
```

qa-tests-backend/src/test/groovy/UpgradesTest.groovy
```diff
@@ -43,6 +43,11 @@ class UpgradesTest extends BaseSpecification {
             }
         }"""
 
+    private static final List<String> BROKEN_POLICIES = [
```
theencee, line 46: the idea behind `KnownPolicyDiffs` is that it wouldn't fail if the specific sections were broken but it would fail if any other fields get broken.

For example, in this policy currently we know remediation and rationale are missing in default so we want to ignore that in the tests.

But let's say a migration was made to this policy that changed the _criteria_ but not for default policies (or vice versa). This would ignore that, but the other way using KnownPolicyDiffs wouldn't. 

Minor point, and I don't want to block since this is expected to be fixed by 73, but still worth pointing out.

theencee (commented):
We still have customers on 69 and 68 even though we don't "officially" support it. I think just to be safe, it's better we test upgrades from 69 if possible.

qa-tests-backend/src/test/groovy/UpgradesTest.groovy
gavin-stackrox: The fix for [ROX-13265](https://issues.redhat.com/browse/ROX-13265) should also remove this brute force `BROKEN_POLICIES` fix.

gavin-stackrox:
> We still have customers on 69 and 68 even though we don't "officially" support it. I think just to be safe, it's better we test upgrades from 69 if possible.

Ok. I will try 69.

gavin-stackrox:
Using 69 means we can also avoid the horrible BROKEN_POLICIES hack 🎉

theencee (approved):

PR #3537, opened 2022-10-24
Dashrews/ROX-13253 wait for central-db to come back after bounce and allow FATAL connection lost error

dashrews78:
## Description

2 issues here:
1.  After bouncing central-db we need to make sure it is back up before we continue with the test
2. When central-db is bounced central will often report FATAL errors it receives from PGX about the connection being down.  Things like `FATAL: terminating connection due to administrator command (SQLSTATE 57P01)`.  These errors need to be allowed in this test, however, we probably shouldn't blindly allow them to ensure we don't miss issues in other tests.  So that string is added to the allow list for this test only.

## Checklist

## Testing Performed

This is a test.  The testing is running it.

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -439,6 +439,45 @@ setup_automation_flavor_e2e_cluster() {
     fi
 }
 
+# When working as expected it takes less than one minute for the API server to
+# reach ready. Often times out on OSD. If this call fails in CI we need to
+# identify the source of pull/scheduling latency, request throttling, etc.
+# I tried increasing the timeout from 5m to 20m for OSD but it did not help.
+wait_for_central_db() {
+    info "Waiting for Central DB to start"
+
+    start_time="$(date '+%s')"
+    max_seconds=300
+
+    while true; do
+        central_db_json="$(kubectl -n stackrox get deploy/central-db -o json)"
+        replicas="$(jq '.status.replicas' <<<"$central_db_json")"
+        ready_replicas="$(jq '.status.readyReplicas' <<<"$central_db_json")"
+        curr_time="$(date '+%s')"
+        elapsed_seconds=$(( curr_time - start_time ))
+
+        # Ready case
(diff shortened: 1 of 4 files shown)


tests/e2e/lib.sh
```diff
@@ -439,6 +439,45 @@ setup_automation_flavor_e2e_cluster() {
     fi
 }
 
+# When working as expected it takes less than one minute for the API server to
+# reach ready. Often times out on OSD. If this call fails in CI we need to
+# identify the source of pull/scheduling latency, request throttling, etc.
+# I tried increasing the timeout from 5m to 20m for OSD but it did not help.
```
c-du, line 445: I see you set the wait timeout to be 5 mins. Do you know why did OSD take so long to reinstate? How long does it usually take?

dashrews78: @c-du I literally copied the top half of wait_for_api where it checks to make sure the deployment is running.  That includes the copying of the comment.  So I don't know.

tests/e2e/lib.sh
```diff
@@ -439,6 +439,45 @@ setup_automation_flavor_e2e_cluster() {
...
+    while true; do
+        central_db_json="$(kubectl -n stackrox get deploy/central-db -o json)"
+        replicas="$(jq '.status.replicas' <<<"$central_db_json")"
+        ready_replicas="$(jq '.status.readyReplicas' <<<"$central_db_json")"
+        curr_time="$(date '+%s')"
+        elapsed_seconds=$(( curr_time - start_time ))
+
+        # Ready case
+        if [[ "$replicas" == 1 && "$ready_replicas" == 1 ]]; then
+            sleep 30
```
janisz, line 461: Why do we need this if replicas are ready?

dashrews78: @janisz I mirrored (actually copied) this from the existing wait_for_api.  So I assumed that was there for reasons.  I could take it out if everyone likes, but I don't think it does any harm.

c-du (approved):

PR #3538, opened 2022-10-24
Make gRPC service structs forward-compatible

misberner:
## Description

gRPC code generation generates `Unimplemented...` structs that return an unimplemented status code for all methods. This struct should be embedded in all gRPC service implementations. While this is not currently enforced, future gRPC code generator versions will make this mandatory (this behavior can be disabled, but it is not recommended; see [here](https://github.com/grpc/grpc-go/issues/3669) for a lengthy discussion on this matter with the clear verdict that gRPC authors see embedding of `Unimplemented..` as something that should always have been required).

Alas, this makes service type assertions ineffective, but that seems to be the pill to swallow.

## Checklist

## Testing Performed

- CI

### DIFF
--- a/central/alert/service/service_impl.go
+++ b/central/alert/service/service_impl.go
@@ -75,6 +75,8 @@ var (
 
 // serviceImpl is a thin facade over a domain layer that handles CRUD use cases on Alert objects from API clients.
 type serviceImpl struct {
+	v1.UnimplementedAlertServiceServer
+
 	dataStore         datastore.DataStore
 	notifier          notifierProcessor.Processor
 	baselines         baselineDatastore.DataStore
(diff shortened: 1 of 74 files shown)


md2119 (commented):
How do make sure none of the services are missed? Is there a way to unit test it?

misberner:
> How do make sure none of the services are missed? Is there a way to unit test it?

We could write a unit test or linter. But because a soon-to-be-upgraded-to version of the proto library will enforce this via the type system, this seems unnecessary.

md2119 (approved):

janisz (approved):

misberner:
Test failure is not related.

PR #3540, opened 2022-10-24
ROX-13231: use passed context when non-postgres

keyallis:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/graphql/resolvers/cluster_vulnerabilities.go
+++ b/central/graphql/resolvers/cluster_vulnerabilities.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/graph-gophers/graphql-go"
+	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/graphql/resolvers/loaders"
 	"github.com/stackrox/rox/central/metrics"
 	v1 "github.com/stackrox/rox/generated/api/v1"
@@ -313,6 +314,12 @@ func withOpenShiftTypeFiltering(q string) string {
 }
 
 func (resolver *clusterCVEResolver) clusterVulnerabilityScopeContext(ctx context.Context) context.Context {
+	if ctx == nil {
+		err := utils.Should(errors.New("argument 'ctx' is nil"))
+		if err != nil {
+			log.Error(err)
+		}
+	}
 	if resolver.ctx == nil {
 		resolver.ctx = ctx
 	}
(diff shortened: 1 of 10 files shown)


central/graphql/resolvers/cluster_vulnerabilities.go
```diff
@@ -312,10 +312,9 @@ func withOpenShiftTypeFiltering(q string) string {
 		search.NewQueryBuilder().AddExactMatches(search.CVEType, storage.CVE_OPENSHIFT_CVE.String()).Query())
 }
 
-func (resolver *clusterCVEResolver) clusterVulnerabilityScopeContext() context.Context {
+func (resolver *clusterCVEResolver) clusterVulnerabilityScopeContext(ctx context.Context) context.Context {
 	if resolver.ctx == nil {
-		log.Errorf("attempted to scope context on nil")
```
md2119, line 317: 😱

central/graphql/resolvers/nodes.go
```diff
@@ -543,24 +518,23 @@ func (resolver *nodeResolver) PlottedNodeVulnerabilities(ctx context.Context, ar
...
+	res.ctx = resolver.nodeScopeContext(ctx)
 	return res, nil
 }
 
 func (resolver *nodeResolver) UnusedVarSink(_ context.Context, _ RawQuery) *int32 {
 	return nil
 }
 
-func (resolver *nodeResolver) nodeScopeContext() context.Context {
+func (resolver *nodeResolver) nodeScopeContext(ctx context.Context) context.Context {
```
md2119, line 532: Context of incoming requests is not expected to be nil. Therefore, for this and other scope context builders, check `ctx` is not nil and panic on dev build. That way we can catch if incorrect context is passed around. Remember that  `resolver.ctx` is set to `ctx` from request.

```
if ctx == nil {
  utils.Should("argument 'ctx' is nil")
}
```

md2119 (changes requested):
One comment on nil check on `ctx` from request, otherwise lgtm.

central/graphql/resolvers/cluster_vulnerabilities.go
keyallis: Do we want to retain these messages? They don't seem necessary or at the very least shouldn't be errors

md2119 (approved):

PR #3541, opened 2022-10-24
ROX-13123, ROX-13124: Redirect to clusters page when no secured clusters detected

vjwilson:
## Description

After much discussion, we have landed on the look in this Jira comment
https://issues.redhat.com/browse/ROX-12931?focusedCommentId=21174780&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21174780

Simple alert and language, with a link that opens a doc page which will be updated with more appropriate instructions when the 3.73 docs are released.

## Checklist
- [x] Investigated and inspected CI test results (postgres ui-e2e tests passed earlier, now stuck in certificate error)

## Testing Performed

* Added basic cypress e2e smoke test
* Manual testing

(overrode comparison to show redirect and Empty State of clusters)
Screen Shot 2022-11-04 at 12 37 19 PM

And here is the link to the docs page
Screen Shot 2022-11-02 at 10 43 27 AM


ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
```diff
@@ -0,0 +1,62 @@
...
+                        </Text>
+                    </TextContent>
+                </FlexItem>
+                <FlexItem>
+                    <Button
+                        variant={ButtonVariant.primary}
+                        component="a"
+                        target="_blank"
+                        rel="noopener noreferrer nofollow"
+                        href="https://docs.openshift.com/acs/3.72/installing/install-ocp-operator.html#adding-a-new-cluster-to-rhacs"
```
dvail, line 52: Is this URL something we can determine dynamically, or do we just need to make a note to check this before each release?

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
```diff
@@ -51,6 +63,17 @@ function MainPage(): ReactElement {
     const { isFeatureFlagEnabled, isLoadingFeatureFlags } = useFeatureFlags();
     const { hasReadAccess, hasReadWriteAccess, isLoadingPermissions } = usePermissions();
 
+    // Check for clusters under management
+    // if none, and user can admin Clusters, redirect to clusters section
+    // (only applicable in Cloud Services version)
```
dvail, line 68: By "only applicable in Cloud Services version", does that mean that this situation is only ever possible in the cloud services version? Is there any way an on-prem install could return no clusters?

Related, is the ability to view clusters something that is controlled by SAC? Is it possible a user could have insufficient permissions to view clusters but still be able to do useful work, and this would cause them to get stuck on the clusters page?

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
```diff
@@ -51,6 +63,17 @@ function MainPage(): ReactElement {
...
+    useQuery<ClusterCountResponse>(CLUSTER_COUNT, {
```
dvail, line 69: Another thought, I wonder if it would be worth adding some text either in the UI or on the docs page that tells the user they need to refresh the ACS UI once a cluster is added. Since Apollo caches the GQL query results, this call will continue to return `0` every time the user navigates to a new page, even after a cluster is added.

dvail (approved):
I had a couple of thoughts/questions, but they might not be applicable to this PR directly.

ui/apps/platform/cypress/integration/clusters/clusters.test.js
```diff
@@ -48,6 +49,24 @@ describe('Clusters page', () => {
             });
         });
     });
+
+    describe('when no secured clusters are added yet (only applies to Cloud Service)', () => {
+        it('should should redirect to the Clusters page', () => {
+            visitDashboardWithNoClusters();
+
+            cy.url().should('contain', `${clustersUrl}`);
```
pedrottimark, line 57: Please rewrite assertion according to majority pattern (118 versus 3 occurrences) and move into helper function because semantically related to heading assertion there, which is repeated here.

Also removes need to import url here. By the way, it can be exact match, true?

```js
cy.location('pathname').should('eq', clustersUrl);
```

ui/apps/platform/cypress/integration/clusters/clusters.test.js
```diff
@@ -48,6 +49,24 @@ describe('Clusters page', () => {
...
+
+            cy.get(selectors.clustersListHeading);
+
+            cy.get(
+                'p:contains("You have successfully deployed a Red Hat Advanced Cluster Security platform.")'
+            );
+
+            cy.get('h1:contains("Configure the clusters you want to secure.")');
```
pedrottimark, line 65: Accessibility: how about `h2` element here, because **Clusters** page already has `h1` element?

ui/apps/platform/cypress/integration/clusters/clusters.test.js
```diff
@@ -48,6 +49,24 @@ describe('Clusters page', () => {
             });
         });
     });
+
+    describe('when no secured clusters are added yet (only applies to Cloud Service)', () => {
```
pedrottimark, line 53: Please add test to a new file like clustersCloudService.test.js

```js
describe('Clusters Cloud Service', () => {
    it('should redirect from main dashboard when no secured clusters', () => {
```

ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
```diff
@@ -0,0 +1,62 @@
...
+                <p className="pf-u-font-weight-normal">
+                    You have successfully deployed a Red Hat Advanced Cluster Security platform. Now
+                    you can configure the clusters you want to secure.
+                </p>
+            </Alert>
+            <Flex
+                alignItems={{ default: 'alignItemsCenter' }}
+                justifyContent={{ default: 'justifyContentCenter' }}
+                className="pf-u-text-align-center"
+                direction={{ default: 'column' }}
```
pedrottimark, line 29: Is this equivalent to `Bullseye` layout element?

https://www.patternfly.org/v4/layouts/bullseye/design-guidelines

> Use a **Bullseye** layout to center content, both vertically and horizontally within a container.

ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
lmaynard88: We plan to redirect to a page in the 3.73 doc set once we publish the 3.73 docs. So it's not dynamic. We may want a different approach in future.

ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
```diff
@@ -0,0 +1,62 @@
...
+            >
+                <FlexItem className="pf-u-mb-0">
+                    <img src={SecurityFullColorPink} alt="" style={{ height: '430px' }} />
+                </FlexItem>
+                <FlexItem className="pf-u-w-66">
+                    <TextContent className="pf-u-mb-xl">
+                        <Text component={TextVariants.h1} data-testid="congratulations">
```
pedrottimark, line 36: 1. Although people debate number of `h1` elements on a page for SEO, accessibility day of learning resources included live interview of screen reader user exploring a page for the first time, which included extensive use of heading hierarchy and comments about anything that diverged from semantic hierarchy. In that case skips in levels. It seems likely that multiple `h1` would have earned a comment.
2. is `data-testid` prop left over? Test asserts on element and content.

ui/apps/platform/src/Containers/Clusters/ClustersTablePanel.js
```diff
@@ -263,23 +271,28 @@ function ClustersTablePanel({ selectedClusterId, setSelectedClusterId, searchOpt
...
-                            toggleRow={toggleCluster}
-                            toggleSelectAll={toggleAllClusters}
-                            selection={checkedClusterIds}
-                            selectedRowId={selectedClusterId}
-                            noDataText="No clusters to show."
-                            minRows={20}
-                            pageSize={pageSize}
-                        />
-                    </div>
+                    {(!fetchingClusters || pollingCount > 0) && currentClusters.length <= 0 && (
```
pedrottimark, line 274: To follow up excellent improvement to add `fetchingClusters` and move to a pattern which supports error message if that becomes necessary, something like `{content}` here and if-else-if-else for assignment makes it clearer to render spinner by default (else in potential future, error message) else if no clusters `AddClusterPrompt` else `CheckboxTable` element.

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
```diff
@@ -51,6 +63,17 @@ function MainPage(): ReactElement {
     const { isFeatureFlagEnabled, isLoadingFeatureFlags } = useFeatureFlags();
     const { hasReadAccess, hasReadWriteAccess, isLoadingPermissions } = usePermissions();
 
+    // Check for clusters under management
+    // if none, and user can admin Clusters, redirect to clusters section
```
pedrottimark, line 67: The phrase **and use can admin Clusters** implies a potential timing problem, because the following `if` statement waits until response from /v1/mypermissions request.

pedrottimark (approved):
Approved with some comments for your consideration. Let’s be careful.

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
vjwilson: ^ @lmaynard88 
Is Dave's suggestion something you can add to the docs?

lmaynard88:
Minor nits for future consideration (if too late now): Change future to present tense. "...new clusters are listed here" instead of "will be."

Also, aren't all clusters listed, not just new ones? So you can add, and at a later time add more, and they all appear here, whether new or not. Should it say "secured clusters are listed here" instead of "new" clusters?

LGTM otherwise.

ui/apps/platform/cypress/integration/clusters/clusters.test.js
vjwilson: I intentionally took all assertions out of the helper function, because this is a one-off test, where the helper function is mainly just to abstract the API mocks and waiting. 
(I almost moved everything into test, instead of keeping the mocking in a helper function, but in the end made the subjective call that the setup distracted from the assertions, so that's where I drew the line in this special case.

ui/apps/platform/cypress/integration/clusters/clusters.test.js
vjwilson: Good point, but tricky because the important information is that "you aren't monitoring any clusters yet".
I'll see if I can make it a semantic h2, but with h1 styling.

ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
vjwilson: I tried a Bullseye at first, but required style overrides to get the spacing between the image, text, and button correct. Rather than hack in overrides, I chose to simplify. (The larger UX team recommended this look to Zhenpeng as a variation the

ui/apps/platform/src/Containers/Clusters/ClustersTablePanel.js
vjwilson: I started to improve the non-empty flow, but stopped myself, so as not to set a bad example through "scope creep".

pedrottimark: The complicated conditions imply non-obvious behavior change to render neither when fetching.

How about the following, which keeps render neither until after first fetch, but simplifies the conditions:
* It is hard to see how array length can become less than zero.
* If length is non-zero (or if you prefer, greater than zero) it seems redundant to require not fetching.

```js
{(!fetchingClusters || pollingCount > 0) && currentClusters.length === 0 && (
    <AddClusterPrompt />
)}
{currentClusters.length !== 0 && (
    <div data-testid="clusters-table" className="h-full w-full">
```

pedrottimark: After release 3.73 we can circle back to question by Dave whether no clusters unambiguously implies Cloud Service.

For the moment, no support for cluster auto-upgrade implies Cloud Service. Because it is not impossible that auto-upgrade might become supported in the future, we can follow up about how to get unambiguous indication. Besides this initial conditional rendering, it seems relevant information to render in Clusters and System Health.

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/services/ClustersService.ts#L100-L110

ui/apps/platform/src/Containers/Clusters/AddClusterPrompt.tsx
msugakov: Why plan if redirects work already? https://github.com/stackrox/stackrox/pull/3799

vjwilson: @msugakov The redirect that Lynne is referring to, is a 301 redirect of that doc link to a new improved documentation page about installation, which would not exist until 3.73 is released. The redirect in the documentation means we will not have to change this hard-coded link in the product itself.

PR #3546, opened 2022-10-25
ROX-13320 Process Listening On Port: Implementation of data store, service, e2e tests

erthalion:
## Description

Implement data store needed to persist and query PLOP objects, and corresponding internal API pipeline to actually process it.

## Checklist

## Testing Performed

Manual testing of the full pipeline with the Collector sending PLOP events to the Sensor, all the way up to the Central.


central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,147 @@
...
+	*storage.ProcessListeningOnPort, error,
+) {
+
+	portProcessList, err := ds.storage.GetPLOPForDeployment(ctx, deploymentId)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(portProcessList) > 0 {
+		return portProcessList[0], nil
```
JoukoVirtanen, line 143: Why return only the first element?

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,652 @@
...
+	data, err := postgres.RunGetQueryForSchema(ctx, schema, q, s.db)
+	if err != nil {
+		return nil, false, pgutils.ErrNilIfNoRows(err)
+	}
+
+	var msg storage.ProcessListeningOnPortStorage
+	if err := proto.Unmarshal(data, &msg); err != nil {
+		return nil, false, err
+	}
+	return &msg, true, nil
```
JoukoVirtanen, line 346:
```suggestion
	return data, true, nil
```

JoukoVirtanen, line 346:
```suggestion
	return data, true, nil
```

JoukoVirtanen, line 346:
```suggestion
	return data, true, nil
```

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,652 @@
...
+	sacQueryFilter, err = sac.BuildClusterNamespaceLevelSACQueryFilter(scopeTree)
+	if err != nil {
+		return nil, nil, err
+	}
+	q := search.ConjunctionQuery(
+		sacQueryFilter,
+		search.NewQueryBuilder().AddDocIDs(ids...).ProtoQuery(),
+	)
+
+	rows, err := postgres.RunGetManyQueryForSchema(ctx, schema, q, s.db)
```
JoukoVirtanen, line 458:
```suggestion
	rows, err := postgres.RunGetManyQueryForSchema[storage.ProcessListeningOnPortStorage](ctx, schema, q, s.db)
```

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,654 @@
...
+	// XXX: Verify the query plan to make sure needed indexes are in use.
+	getByNamespaceAndDeploymentStmt =
+		"SELECT plop.id, plop.serialized, " +
+		"proc.podid, proc.containername, proc.signal_name, " +
+		"proc.signal_args, proc.signal_execfilepath " +
+		"FROM process_listening_on_ports plop " +
+		"JOIN process_indicators proc " +
+		"ON plop.processindicatorid = proc.id " +
+		"WHERE proc.namespace = $1 AND proc.deploymentid = $2"
+)
```
JoukoVirtanen, line 57:
```suggestion
        getByNamespaceStmt =
                "SELECT plop.id, plop.serialized, proc.deploymentid, " +
                "proc.podid, proc.containername, proc.signal_name, " +
                "proc.signal_args, proc.signal_execfilepath " +
                "FROM process_listening_on_ports plop " +
                "JOIN process_indicators proc " +
                "ON plop.processindicatorid = proc.id " +
                "WHERE proc.namespace = $1"
)
```

JoukoVirtanen, line 57:
```suggestion
        getByNamespaceStmt =
                "SELECT plop.id, plop.serialized, proc.deploymentid, " +
                "proc.podid, proc.containername, proc.signal_name, " +
                "proc.signal_args, proc.signal_execfilepath " +
                "FROM process_listening_on_ports plop " +
                "JOIN process_indicators proc " +
                "ON plop.processindicatorid = proc.id " +
                "WHERE proc.namespace = $1"

)
```

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,654 @@
...
+	if err != nil {
+		return nil, pgutils.ErrNilIfNoRows(err)
+	}
+	defer rows.Close()
+
+	plops, err := s.readRows(rows)
+
+	return plops, err
+}
+
```
JoukoVirtanen, line 605:
```suggestion

// Manually written function to get PLOP joined with ProcessIndicators
func (s *storeImpl) GetProcessListeningOnPortForNamespace(
        ctx context.Context,
        namespace string,
) ([]*v1.ProcessListeningOnPortWithDeploymentId, error) {
        // XXX: Use SetPostgresOperationDurationTime with a proper generated Metric

        return pgutils.Retry2(func() ([]*v1.ProcessListeningOnPortWithDeploymentId, error) {
                return s.retryableGetPLOPForNamespace(ctx, namespace)
        })
}

func (s *storeImpl) retryableGetPLOPForNamespace(
        ctx context.Context,
        namespace string,
) ([]*v1.ProcessListeningOnPortWithDeploymentId, error) {
        var rows pgx.Rows
        var err error

        rows, err = s.db.Query(ctx, getByNamespaceStmt, namespace)

        if err != nil {
                return nil, pgutils.ErrNilIfNoRows(err)
        }
        defer rows.Close()

        plops, err := s.readRowsNamespace(rows)

        return plops, err
}

```

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,654 @@
...
+				ProcessExecFilePath: signalExecFilePath,
+			},
+		}
+
+		plops = append(plops, plop)
+	}
+
+	log.Debugf("Read returned %+v plops", len(plops))
+	return plops, nil
+}
```
JoukoVirtanen, line 654:
```suggestion
}

// Manual converting of raw data from SQL query to ProcessListeningOnPortWithDeploymentId
func (s *storeImpl) readRowsNamespace(
        rows pgx.Rows,
) ([]*v1.ProcessListeningOnPortWithDeploymentId, error) {
        plopsMap := make(map[string][]*storage.ProcessListeningOnPort)

        for rows.Next() {
                var id string
                var serialized []byte
                var deploymentId string
                var podId string
                var containerName string
                var signalName string
                var signalArgs string
                var signalExecFilePath string

                // We're getting ProcessIndicator directly from the SQL query, PLOP
                // parts have to be extra deserialized.
                if err := rows.Scan(
                        &id, &serialized, &deploymentId,
                        &podId, &containerName,
                        &signalName, &signalArgs, &signalExecFilePath); err != nil {
                        return nil, pgutils.ErrNilIfNoRows(err)
                }

                var msg storage.ProcessListeningOnPortStorage
                if err := proto.Unmarshal(serialized, &msg); err != nil {
                        return nil, err
                }

                plop := &storage.ProcessListeningOnPort{
                        Port: msg.Port,
                        Protocol: msg.Protocol,
                        CloseTimestamp: msg.CloseTimestamp,
                        Process: &storage.ProcessIndicatorUniqueKey {
                                PodId: podId,
                                ContainerName: containerName,
                                ProcessName: signalName,
                                ProcessArgs: signalArgs,
                                ProcessExecFilePath: signalExecFilePath,
                        },
                }

                plopsMap[deploymentId] = append(plopsMap[deploymentId], plop)
                                                                                                                                                                                                 

        plops := make([]*v1.ProcessListeningOnPortWithDeploymentId, 0)

        for k, v := range plopsMap {
                plop := &v1.ProcessListeningOnPortWithDeploymentId{
                        DeploymentId:                   k,
                        ProcessesListeningOnPorts:      v,
                }
                plops = append(plops, plop)
        }

        log.Debugf("Read returned %+v plops", len(plops))
        return plops, nil
}
```

dashrews78:
I may be wrong, but I think we should be able to use the searcher query builder functionality to get the PLOP data based on namespace and/or deployment of the process_indicators.  Similar to how you get the process_indicators in `AddProcessListeningOnPort`.  We would need to add a searcher for PLOP.  Then we should be able to utilize that to join with process_indicators and get the data you are looking for and eliminate the need to make the store custom.  process_indicators and images have examples of searchers that my be useful as examples.  Additionally the grandparent tests in the searcher query builder itself may provide some example information.  `pkg/search/postgres/graph_queries_test.go`.  I suspect/hope the query builder should generate the query very similarly to the one you hand generated in these particular cases.

proto/storage/process_listening_on_port.proto
```diff
@@ -15,3 +17,17 @@ message ProcessListeningOnPort {
     ProcessIndicatorUniqueKey process = 3;
     google.protobuf.Timestamp close_timestamp = 4;
 }
+
+message ProcessListeningOnPortStorage {
+    // Ideally it has to be GENERATED ALWAYS AS IDENTITY, which will make it a
```
dashrews78, line 22: If we do wind up going with a custom store, we could probably justify a custom schema to create a PK how you would like.

central/processindicator/store/postgres/gen.go
```diff
@@ -1,3 +0,0 @@
-package postgres
-
-//go:generate pg-table-bindings-wrapper --type=storage.ProcessIndicator --search-category PROCESS_INDICATORS --references storage.Deployment
```
dashrews78, line 3: add back in with the --schema-only flag

central/processlisteningonport/datastore/datastore.go
```diff
@@ -0,0 +1,43 @@
...
+//go:generate mockgen-wrapper
+type DataStore interface {
+	AddProcessListeningOnPort(context.Context, ...*storage.ProcessListeningOnPort) error
+	GetProcessListeningOnPort(
+		ctx context.Context,
+		opts GetOptions,
+	) (map[string][]*storage.ProcessListeningOnPort, error)
+}
+
+// New creates a data store object to access the database. Since some
```
dashrews78, line 34: subtle distinction.  This includes the PLOP store but the process indicator datastore.  comment states using the process indicator store directly.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/logging"
+	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/uuid"
+)
+
+type datastoreImpl struct {
+	storage        postgres.Store
+	indicatorStore processIndicatorStore.DataStore
```
dashrews78, line 22: nit:  to be consistent and clear we should probably call this `indicatorDataStore`

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
+
+	// First query all needed process indicators in one go
+	for _, val := range portProcesses {
+		lookups = append(lookups,
+			search.NewQueryBuilder().
+				AddExactMatches(search.ContainerName, val.Process.ContainerName).
```
dashrews78, line 65: is there any chance `val.Process` can be nil?  May want to protect ourselves for that case.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,166 @@
...
+			indicatorID = indicator.GetId()
+			log.Debugf("Got indicator %s: %+v", indicatorID, indicator)
+		} else {
+			// XXX: Create a metric for this
+			log.Warnf("Found no matching indicators for %s", key)
+		}
+
+		plopObjects[i] = &storage.ProcessListeningOnPortStorage{
+			// XXX, ResignatingFacepalm: Use regular GENERATE ALWAYS AS
+			// IDENTITY, which would require changes in store generator
```
dashrews78, line 120: if we are going to keep the store custom, no need to restrict ourselves to this ID pattern if another pattern makes sense.  IMO

central/processlisteningonport/datastore/singletone.go
```diff
@@ -0,0 +1,26 @@
+package datastore
```
dashrews78, line 1: file name spelling.  should be `singleton.go`

central/processlisteningonport/datastore/singletone.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var (
+	once sync.Once
+
+	store DataStore
+)
+
+func initialize() {
+	plopStorage := postgres.New(globaldb.GetPostgres())
```
dashrews78, line 17: nit:  i would differentiate the names to make it clear that one is the store and one references the datastore

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if processesListeningOnPorts == nil {
+		log.Debug("In processlisteningonport service query return nil")
+		result := &v1.GetProcessesListeningOnPortsWithDeploymentResponse{
+			ProcessesListeningOnPortsWithDeployment: make([]*v1.ProcessListeningOnPortWithDeploymentId, 0),
+		}
+		return result, nil
+	}
+
+	result := make([]*v1.ProcessListeningOnPortWithDeploymentId, 0)
```
dashrews78, line 70: should probably make it for `len(processesListeningOnPorts)`

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if processesListeningOnPorts == nil {
+		log.Debug("In processlisteningonport service query return nil")
+		result := &v1.GetProcessesListeningOnPortsResponse{
+			ProcessesListeningOnPorts: make([]*storage.ProcessListeningOnPort, 0),
+		}
+		return result, nil
+	}
+
+	result := make([]*storage.ProcessListeningOnPort, 0)
```
dashrews78, line 115: should probably make it for len(processesListeningOnPorts)

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// AuthFuncOverride specifies the auth criteria for this API.
+func (s *serviceImpl) AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error) {
+	//// XXX: An anonymous access only for development
+	//  return ctx, allow.Anonymous().Authorized(ctx, fullMethodName)
+	return ctx, authorizer.Authorized(ctx, fullMethodName)
+}
+
+func (s *serviceImpl) GetProcessesListeningOnPortsByNamespace(ctx context.Context, req *v1.GetProcessesListeningOnPortsByNamespaceRequest) (*v1.GetProcessesListeningOnPortsWithDeploymentResponse, error) {
```
dashrews78, line 49: Do we really need 2 endpoints or could we just use one and allow the deployment to be optional?  Just thinking out loud.  Also these 2 methods do very similar things so there may be some consolidation we can do and move some of it to functions.

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,806 @@
+// Code generated by pg-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: This line needs adjusted if we keep it custom

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,806 @@
...
+
+const (
+	baseTable = "process_listening_on_ports"
+
+	batchAfter = 100
+
+	// using copyFrom, we may not even want to batch.  It would probably be simpler
+	// to deal with failures if we just sent it all.  Something to think about as we
+	// proceed and move into more e2e and larger performance testing
+	batchSize = 10000
```
dashrews78, line 41: Based on what all this thing is doing, I would probably reduce this and the deleteBatchSize.

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,806 @@
...
+func (s *storeImpl) AckKeysIndexed(ctx context.Context, keys ...string) error {
+	return nil
+}
+
+// GetKeysToIndex returns the keys that need to be indexed
+func (s *storeImpl) GetKeysToIndex(ctx context.Context) ([]string, error) {
+	return nil, nil
+}
+
+// Manually written function to get PLOP joined with ProcessIndicators
```
dashrews78, line 605: nit:  i would move these above the Stubs section for clarity

central/role/resources/list.go
```diff
@@ -107,6 +107,11 @@ var (
 	// To-be-deprecated in 3.75 with ROX-12750 (deprecation notice in 3.73).
 	ServiceIdentity = newDeprecatedResourceMetadata("ServiceIdentity",
 		permissions.GlobalScope, Administration)
+	SignatureIntegration = newDeprecatedResourceMetadata("SignatureIntegration",
```
dashrews78, line 110: I'm curious why these are in the deprecated section?

pkg/postgres/schema/process_indicators.go
```diff
@@ -1,5 +1,8 @@
 // Code generated by pg-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: This line will need to be adjusted to reflect it is now custom.

pkg/postgres/schema/process_indicators.go
```diff
@@ -45,16 +48,20 @@ const (
 )
 
 // ProcessIndicators holds the Gorm model for Postgres table `process_indicators`.
+// XXX: unique_indicator index is specified manually, as it has to cover
```
dashrews78, line 51: We already talked about it, but for posterity, we should update the generators to allow for this case.  Though possibly as follow on work.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,147 @@
...
+	err = ds.storage.UpsertMany(ctx, plopObjects)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (ds *datastoreImpl) GetProcessListeningOnPortForDeployment(
+	ctx context.Context,
```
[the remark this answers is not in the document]
erthalion: It's being covered by `GetOptions`, although we may want to get rid of the namespace argument altogetherm

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,652 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	q := search.ConjunctionQuery(
+		sacQueryFilter,
+		search.NewQueryBuilder().AddDocIDs(id).ProtoQuery(),
+	)
+
+	data, err := postgres.RunGetQueryForSchema(ctx, schema, q, s.db)
```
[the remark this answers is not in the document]
erthalion: Covered by regenerating on the new golang version.

central/processlisteningonport/store/postgres/store.go
erthalion: Incorporated in to the PR.

proto/storage/process_listening_on_port.proto
erthalion: Yep, that definitely makes sense, and PKs are in fast internal details -- ideally the app doesn't need to know that much about nor even generate them.

central/processindicator/store/postgres/gen.go
erthalion: For process indicators it's actually not going to work, as the custom modifications are in the schema. Will do that for plop objects.

central/processlisteningonport/datastore/datastore.go
erthalion: Good point. What is the common practice to distinguish both, are "store" and "datastore" different enough?

central/processlisteningonport/datastore/datastore_impl.go
erthalion: Probably no, but it never hurts to check.

central/processlisteningonport/datastore/datastore_impl.go
erthalion: It was harder to do without UUID support, when the feature was started. Now should be easier, but to limit the scope a bit I would add this as a follow up step (already got quite a list).

central/processlisteningonport/service/service_impl.go
erthalion: @JoukoVirtanen what do you think?

central/processlisteningonport/store/postgres/store.go
erthalion: Can you elaborate pls, why reduce?

central/role/resources/list.go
erthalion: The feature was branched off before a lot of deprecated SAC were dropped I believe. Need to ask about this, but you're right, probably makes sense to keep it within the new hierarchy.

central/processlisteningonport/datastore/datastore.go
dashrews78: yes generally using store vs datastore in the name is sufficient.  From time to time I've even seen datastore as DS.  but I think I prefer it spelled out.

central/processlisteningonport/datastore/datastore_impl.go
dashrews78: the only point of caution on that, is that if we are going to change the ID we should do so before it gets deployed for people to use so we don't have to write a migration.  so follow up is fine as long as that happens before it gets deployed, IMO

erthalion:
> I may be wrong, but I think we should be able to use the searcher query builder functionality to get the PLOP data based on namespace and/or deployment of the process_indicators. Similar to how you get the process_indicators in AddProcessListeningOnPort. We would need to add a searcher for PLOP. Then we should be able to utilize that to join with process_indicators and get the data you are looking for and eliminate the need to make the store custom. process_indicators and images have examples of searchers that my be useful as examples. Additionally the grandparent tests in the searcher query builder itself may provide some example information. pkg/search/postgres/graph_queries_test.go. I suspect/hope the query builder should generate the query very similarly to the one you hand generated in these particular cases.

Thanks. For the records, after some discussions about this topic here is what we've got:

* The query builder can potentially create join queries, but due to the legacy they will search for the join target dynamically. This adds unnecessary small performance overhead, and makes it harder to figure out what's going on. Couple of examples for using query builder in this way are `pkg/search/postgres/graph_queries_test.go` and `central/image/datastore/datastore_impl_postgres_test.go`.
* The query builder doesn't support group by, which is a part of the current custom implementation.
* Another custom modification required for the PR is schema changes for ProcessIndicator (an index spanning multiple columns), which is not supported by the store generator.
* To carry the custom changes in the store it would be easier to use `--schema-only` flag in `gen.go`.

Most of those topics are planned to address as a follow-up steps, with an idea to have everything generated as usual. From this side I'm even thinking about introducing a new feature into the storage generator: an option to extend the generated storage interface with some custom interface and implementation, that do not have to be living in the generated files. This will make such transitions (generated + custom for time being) easier.

central/processlisteningonport/store/postgres/store.go
dashrews78: Just something to be mindful of when you get around to running some scale and performance type tests.  It may actually be a more likely problem for indicators than PLOP due to the additional multi-column unique index slowing down writes.

central/processlisteningonport/datastore/datastore_impl.go
erthalion: This change ideally would require only a schema migration, not the data migration, so maybe this will make things easier. The plan I have is to change the pk column and add a default value to generate random UUID.

dashrews78: Sadly, changing the type of the PK would not be seamlessly handled by Gorm.  So there would be some work to do there.

central/role/resources/list.go
erthalion: Looks like I was mistaken about the meaning of this construction, the advice I've got was to use `DeploymentExtension` resource as other similar objects.

central/processlisteningonport/service/service_impl.go
JoukoVirtanen: We will go with one endpoint. I will soon create a PR to get rid of specifying the namespace.

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,150 @@
...
+	size := len(processesListeningOnPorts)
+	result := make([]*storage.ProcessListeningOnPort, size)
+
+	// Storage returns map DeploymentID -> PLOP. Just in case verify that
+	// deployment id matches.
+	for k, v := range processesListeningOnPorts {
+		if k != deployment {
+			log.Warnf("Requested deployment %s, got %s. Skipping", deployment, k)
+		} else {
+			result = append(result, v...)
```
JoukoVirtanen, line 143: Similar comment as above.

PR #3547, opened 2022-10-25
Fix shell check linter for deploy/common/deploy.sh

lavanyajain:
## Description

Fix shell check linter for deploy/common/deploy.sh as a part of this defect - Fix shell check linter for deploy/common/deploy.sh

## Checklist

## Testing Performed

$ shellcheck myscript
No issues detected!

### DIFF
--- a/deploy/common/deploy.sh
+++ b/deploy/common/deploy.sh
@@ -45,7 +45,8 @@ if [[ -z "${SCANNER_IMAGE}" ]]; then
 fi
 export SCANNER_DB_IMAGE="${SCANNER_DB_IMAGE:-}"
 if [[ -z "${SCANNER_DB_IMAGE}" ]]; then
-  export SCANNER_DB_IMAGE="$DEFAULT_IMAGE_REGISTRY/scanner-db:$(cat "$(git rev-parse --show-toplevel)/SCANNER_VERSION")"
+  SCANNER_DB_IMAGE="$DEFAULT_IMAGE_REGISTRY/scanner-db:$(cat "$(git rev-parse --show-toplevel)/SCANNER_VERSION")"
+  export SCANNER_DB_IMAGE
 fi
 echo "StackRox scanner image set to $SCANNER_IMAGE"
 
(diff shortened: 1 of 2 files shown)


deploy/common/deploy.sh
```diff
@@ -150,7 +155,7 @@ function get_cluster_zip {
       exit 1
     fi
 
-    ID="$(cat "${TMP}" | jq -r .cluster.id)"
+    ID="$(command < "${TMP}" | jq -r .cluster.id)"
```
dastgirp, line 158: This would fail.

I created sample file "test.json", with following content:
```json
{
        "cluster": {
                "id": 100
        }
}
```
and performed following test:

- `cat test.json | jq -r .cluster.id`
Result: `100`
- command < "${TMP}" | jq -r .cluster.id
No Output
- `jq -r .cluster.id test.json`
Result: `100`

It's better to convert to jq command with filename as argument, instead of cat.

lavanyajain: @dastgirp Modified all commands with cat and JQ to command JQ with filename as argument

lavanyajain:
@janisz @tommartensen @gavin-stackrox - Requesting for PR review. Thanks :)

lavanyajain:
@janisz @tommartensen @gavin-stackrox - Requesting for PR review. Thanks :)

deploy/common/deploy.sh
```diff
@@ -85,19 +86,23 @@ function wait_for_central {
 
     echo -n "Waiting for Central to respond."
     set +e
-    local start_time="$(date '+%s')"
+    start_time="$(date '+%s')"
+    local start_time
     local deadline=$((start_time + 10*60))  # 10 minutes
-    until $(curl_central --output /dev/null --silent --fail "https://$LOCAL_API_ENDPOINT/v1/ping"); do
```
gavin-stackrox, line 90: I think the correct fix for this is:
`until curl_central --output /dev/null --silent --fail "https://$LOCAL_API_ENDPOINT/v1/ping"; do`
i.e. remove the `$(...)`. The loop will then be terminated when `curl_central` is successful.

gavin-stackrox (changes requested):
Thanks!

lavanyajain:
@janisz @tommartensen @gavin-stackrox - Requesting for PR review. Thanks :)

janisz (approved):

deploy/common/deploy.sh
```diff
@@ -243,7 +245,8 @@ function setup_license() {
     echo "Injecting license ..."
     [[ -f "$license_file" ]] || { echo "License file $license_file not found!" ; return 1 ; }
 
-    local tmp="$(mktemp)"
+    tmp="$(mktemp)"
```
dastgirp, line 248: This change will convert "tmp" to global variable. (Before this change, it was a local variable). Is this intended change?

lavanyajain: @dastgirp Makes sense to update tmp to be a local variable. Similarly refined and validated all other variable definitions to retain its original state.

lavanyajain:
[janisz](https://github.com/janisz) [dastgirp](https://github.com/dastgirp) Can we merge this PR? I changes look okay

janisz:
@gavin-stackrox PTAL

gavin-stackrox (approved):

lavanyajain:
@gavin-stackrox @janisz I see both of you approved this PR. Can you help me in merging this?

PR #3550, opened 2022-10-25
ROX-13125: Show error if CVE CSV download in non-Postgres env

vjwilson:
## Description

This is an MVP solution to letting the user know they now have to choose a CVE type (IMAGE_CVE, K8S_CVE, ISTIO_CVE, NODE_CVE, and OPENSHIFT_CVE) in a non-Postgres environment.

**This needs to go into 3.73 because Postgres will not be turned on by default for that version.**

## Checklist

## Testing Performed

Manual testing

Go to the CVE list page and try to download a CSV without setting a CVE_TYPE filter.
Screen Shot 2022-10-25 at 11 23 07 AM

### DIFF
--- a/ui/apps/platform/src/Components/ExportButton.js
+++ b/ui/apps/platform/src/Components/ExportButton.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import * as Icon from 'react-feather';
 import onClickOutside from 'react-onclickoutside';
+import { toast } from 'react-toastify';
 
 import downloadCSV from 'services/CSVDownloadService';
 import WorkflowPDFExportButton from 'Components/WorkflowPDFExportButton';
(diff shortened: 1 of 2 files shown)


pedrottimark:
Before I approve, can y’all point me to where is the query parameter mentioned in the issue?

Has there been code changes since most recently updated staging and staging-db? Same requests and no error:
* staging request on cves page: /api/vm/export/csv?query=&pagination.offset=0&pagination.limit=0&pagination.sortOption.field=CVSS&pagination.sortOption.reversed=true
* staging-db request on node-cves page: /api/export/csv/node/cve?query=&pagination.offset=0&pagination.limit=0&pagination.sortOption.field=CVSS&pagination.sortOption.reversed=true

Charmik and/or Van can you take a look at staging response to verify if multiple lines mess up the csv results:
staging-cves-csv-response

vjwilson:
> Before I approve, can y’all point me to where is the query parameter mentioned in the issue?
The query params for the filter get added in the exported utility function `getCvesAsCsv` in VunerabilitiesService.js
https://github.com/stackrox/stackrox/blob/c0e4213012d73522c7f41124d9572ee5d8f0779a/ui/apps/platform/src/services/VulnerabilitiesService.js#L90-L105

---
> Has there been code changes since most recently updated staging and staging-db? Same requests and no error:
This fails now in master, unless there is a CVE_TYPE filter. (see screenshot below)

staging must still be at a version before it became a requirement. (I think it's at least a week behind master.)

It's only a requirement for non-Postgres environments, so doesn't apply to staging-db, whose purpose is to "dog-food" Postgres. (Postgres environments will not throw the error.)

---
> Charmik and/or Van can you take a look at staging response to verify if multiple lines mess up the csv results: 

I can't comment on the format of the CSV; that comes as-is from the API. All the ones I successfully download have opened up cleanly in VSCode and Numbers.

Screen Shot 2022-10-25 at 2 53 49 PM

bradr5 (approved):

charmik-redhat (approved):
lgtm

PR #3500, opened 2022-10-19
Impl name and description fields for collection form.

dvail:
## Description

Adds name and description fields to the collection form.

## Checklist
- [ ] Investigated and inspected CI test results

If any of these don't apply, please comment below.

## Testing Performed

Visit the collection page and click "Create collection" to view the empty name and description fields. For validation, the name field requires only a non-empty string and the description field is left unrestricted. The validation state will appear onBlur for the name field (in the future it will appear onSubmit as well, if the user skips the field entirely.)
image
image

Visit the collection page and click the name of an existing collection. The name and description fields should be populated with the correct data. Changing the value of the name field will update the title at the top of the page.
image
image
image

Mobile view:
﻿image

PR #3556, opened 2022-10-25
Add field IDs to name and label selector components

dvail:
## Description

This adds `fieldId` (`for` attribute) to form labels and `id` to form input fields.

Since most of the form labels have multiple inputs there are two approaches we can take here:

1. Associate the form label with the _last_ field input, dynamically, so that we get a clean `for`<->`id` mapping between labels and inputs to enable focusing the input element when a label is clicked. This might be somewhat convenient for the user at the expense of ignoring all fields that are not that last field associated with the input. I'm also not sure if this is semantically valid, since the label would only be tied to a single field (in reality is the label references multiple fields). This is even more of a problem since the label key and value labels reference multiple inputs across multiple rules.

2. Assign the form labels and field inputs their own unique ids that make sense given the context. This approach appends an index to each label for clear separation. We lose the ability to click a label to focus an input in all cases except for "Label key", but we gain clear programmatic control of accessing fields in the form.

Name rules will have:
```
# single name rule
for=Entity-name-value
id=Entity-name-value-0
id=Entity-name-value-1
id=Entity-name-value-2
...
```
And label rules will have:
```
# first label rule
for=Entity-label-key-0
id=Entity-label-key-0
for=Entity-label-value-0
id=Entity-label-value-0-0
id=Entity-label-value-0-1

# second label rule
for=Entity-label-key-1
id=Entity-label-key-1
for=Entity-label-value-1
id=Entity-label-value-1-0
id=Entity-label-value-1-1
...
```

I'm honestly not sure which is better for accessibility. The second option doesn't tie form fields and labels together at all, which may or may not be worse than the first option that ties one, but not all, fields to a label.

Another note is that the PatternFly `<Select>` does not give us full control over what the underlying `<input>` element's id is. The component is hard coded to use the passed `toggleId` and append the text `-select-typeahead` to it. The rendered HTML looks like the following when `Namespace-label-key-0` is passed as the `toggleId`.
```html
<div class="pf-c-select__toggle pf-m-typeahead">
  <div class="pf-c-select__toggle-wrapper">
    <input class="pf-c-form-control pf-c-select__toggle-typeahead" id="Namespace-label-key-0-select-typeahead" aria-label="" placeholder="" type="text" autocomplete="off" value="kubernetes.io/metadata.name">
  </div>

  <button id="Namespace-label-key-0" aria-labelledby=" Namespace-label-key-0" aria-expanded="false" aria-haspopup="listbox" type="button" class="pf-c-button pf-c-select__toggle-button pf-m-plain" aria-label="Options menu" tabindex="-1">
    <svg fill="currentColor" height="1em" width="1em" viewBox="0 0 320 512" aria-hidden="true" role="img" class="pf-c-select__toggle-arrow" style="vertical-align: -0.125em;"><path d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"></path></svg>
  </button>
</div>
```

## Checklist

## Testing Performed

Manual viewing of HTML output.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
+++ b/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
@@ -3,6 +3,7 @@ import { Select, ValidatedOptions } from '@patternfly/react-core';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 
 export type AutoCompleteSelectProps = {
+    id: string;
     selectedOption: string;
     className?: string;
     typeAheadAriaLabel?: string;
@@ -12,6 +13,7 @@ export type AutoCompleteSelectProps = {
 
 /* TODO Implement autocompletion */
 export function AutoCompleteSelect({
+    id,
     selectedOption,
     className = '',
     typeAheadAriaLabel,
@@ -28,6 +30,7 @@ export function AutoCompleteSelect({
     return (
         <>
             <Select
+                toggleId={id}
                 validated={validated}
                 typeAheadAriaLabel={typeAheadAriaLabel}
                 className={className}
--- a/ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
+++ b/ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
@@ -127,7 +127,7 @@ function ByLabelSelector({
                                 >
                                     <AutoCompleteSelect
                                         id={`${entityType}-label-key-${ruleIndex}`}
-                                        typeAheadAriaLabel={`Select a value for the ${entityType.toLowerCase()} label key`}
+                                        typeAheadAriaLabel={`Select label key for ${entityType} rule ${ruleIndex} of ${scopedResourceSelector.rules.length}`}
                                         selectedOption={rule.key}
                                         onChange={(fieldValue: string) =>
                                             onChangeLabelKey(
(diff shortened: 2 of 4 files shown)


dvail:
* master
  * **PR #3444** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3444" target="_blank">Graphite</a> 
    * **PR #3556** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3556" target="_blank">Graphite</a>  👈
    * **PR #3500** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3500" target="_blank">Graphite</a> 
    * **PR #3491** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3491" target="_blank">Graphite</a> 
    * **PR #3492** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3492" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3556?utm_source=stack-comment).

dvail:
The thought was that if we could only tie the label click behavior to a single input, that the last input in the group would be the one most likely to be focused by the user. Just a hunch though, and may not be accurate. TBH given the multiple input elements per label, I don't think the user is going to be clicking on the label very frequently anyway.

bradr5 (approved):

pedrottimark (approved):
Thank you.

For your info, integration test helpers a include lite imitation of testing-library which:
1. gets `label` element by its text
2. gets `input` or `select` element by its `id` from `for` attribute value

Collection form tests will give some simulation of pro and con for accessibility.

dvail:
@pedrottimark When you put it that way, I'm fairly confident that even with this change we won't be able to access these input elements without doing our own string concatenation of the `id` values based on the `for` value of the labels. Even if we were to use the full API set of testing-library we couldn't get these reliably.

What might make more sense is to pass this information as part of the `typeAheadAriaLabel`. This value currently reads "Select a value for the <entity> name" for name values but can be easily changed to something like "Select value <index> of <total> for the <entity> name", e.g. "Select value 2 of 3 for the deployment name" and similarly "Select value 2 of 3 for label <key|value> for deployment rule 2".

I know we could use the testing-library `getByLabelText()` to retrieve these elements in a semantic way, but does that help us in integration tests at all?

pedrottimark:
Dave, thank you for investigating other accessibility props. What you describe sounds more useful to screen reader users, pardon pun. Forgive me, if I was not clear enough in the original comment which lead to this pull request. Although I have not checked with aXe DevTools for a while, these changes have probably reduced the number of reported issues.

PR #3558, opened 2022-10-25
validate duration settings

RTann:
## Description

* Panic upon registering duration setting with invalid default value.
* Only allow for users to set durations > 0. Otherwise, they do not make much sense.

## Checklist

## Testing Performed

CI


misberner (changes requested):
Don't we ever use `0` to indicate "never"? I think I found this at least for the notification debounce setting, possibly others as well.

misberner (changes requested):
See other PR, we use a default of 0 for the alert notification debounce setting to indicate no debounce

RTann:
> See other PR, we use a default of 0 for the alert notification debounce setting to indicate no debounce

@misberner which other PR?

RTann:
> See other PR, we use a default of 0 for the alert notification debounce setting to indicate no debounce

I do see we use 0 for `ROX_ALERT_RENOTIF_DEBOUNCE_DURATION` and `ROX_RECORDER_DURATION` (which I don't think is used anymore?). That being said, allowing 0 makes sense, but I do feel there are times where 0 should not be allowed. I wonder if I should add an option to allow/disallow 0. What do you think?

RTann:
PR to remove `ROX_RECORDER_DURATION` https://github.com/stackrox/stackrox/pull/3559

misberner:
> @misberner which other PR?

I got two separate email threads because you renamed the PR, so I just assumed it was a different PR :) was on my phone

misberner:
> I wonder if I should add an option to allow/disallow 0. What do you think?

Hmm, personally I would think that disallowing `0` durations should happen at the level of the logic that consumes the duration setting. But I guess you could make the case that a `0` duration that is not intercepted can be very dangerous when it leads to code running in a loop with no delay when there should be a delay between iterations, and thus consuming insane amounts of CPU. Therefore, yeah, I think it makes sense to prescribe that a `0` duration has to be allowed explicitly.

misberner (approved):

PR #3575, opened 2022-10-26
ROX-12976: fix for 3.71.x: cherry-pick finish-release to avoid auto-trigger on release publish

tommartensen:
## Description

Cherry-picking the finish-release workflows from the master branch.
This is to remove the trigger on a new release, which previously triggered the finish-release workflow from the release branch after the release engineer had created a new release. Though having no side effects, the Slack message it did send lead to some confusion.

The variables workflow may not be strictly necessary to be updated, but I did it as it is a direct dependency of the finish-release workflow. The common.sh and other scripts are taken from `master`, so we don't need to update them to the release branch.

## Checklist

## Testing Performed

No testing required.

### DIFF
--- a/.github/workflows/finish-release.yml
+++ b/.github/workflows/finish-release.yml
@@ -1,30 +1,133 @@
 name: Finish Release
 on:
-  release:
-    types: [released]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Release version (A.B.C[-N])
+        required: true
+        default: 0.0.0-test
+        type: string
+      dry-run:
+        description: Dry-run
+        required: false
+        default: true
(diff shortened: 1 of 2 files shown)


tommartensen:
> Shouldn't we just delete the file?

Would also work, I think updating the workflow is more future facing 😄

0x656b694d:
I thought the idea was to never execute this workflow from release branch :thinking:

vikin91 (approved):
👍 (I didn't check the code in details)

tommartensen:
> I thought the idea was to never execute this workflow from release branch 🤔

The idea is only to remove the auto trigger.

PR #3577, opened 2022-10-26
ROX-13300: Remove certexpiry request from generic wait in integration tests

pedrottimark:
### Test failure

* 1584938569003175936 from master build for 3535 on 2022-10-25

Compliance dashboard page should show the same amount of nodes between the Dashboard and List Page

> Timed out retrying after 10000ms: `cy.wait()` timed out waiting `10000ms` for the 1st request to the route: `credentialexpiry_CENTRAL`. No request ever occurred.

credentialexpiry

This is the fourth test in compliance/complianceDashboard.test.js

### Analysis

Compliance dashboard has made its requests before responses to credentialexpiry requests. Because the former are several and slow, they might even delay the latter.

Because `MainPage` renders pair of `CredentialExpiryBanner` elements which make the requests, they are in parallel with page specific requests, not in sequence as prerequisites. Therefore, my bad to include them in `requestConfigGeneric` for visit helpers.

### Solution

Because certexpiration tests independently mock the response to the relevant request without staticResponseMap argument, just delete the route matcher properties.

## Changed files

1. Edit cypress/constants/apiEndpoints
    * Add `availableAuthProviders` property.

2. Edit cypress/helpers/visit.js
    * Add key constants as single source of truth for keys in optional staticResponseMap argument.
    * Add availableAuthProviders and loginAuthProviders route matcher properties. Brad, these requests seem relevant to auth tests. Although on second thought, those tests mock them in a specialized context.
    * Replace credentialexpiray route matcher properties with comment.
    * Rename `visitWithPermissions` as `visitWithStaticResponseForPermissions` function and also `permissionsStaticResponse` with `staticResponseForPermissions` to prevent misunderstanding that argument is a permissions object.

3. Edit cypress/integration/accessControl/accessControlAuthProviders.test.js
    * Ditto bullet point above.

### Residue

1. Now that we know credentialexpiry requests are test-specific instead of generic, improve certexpiration tests in general and test which mocks permissions in specific.

## Checklist
- [x] Edited integration test constants and helpers

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/constants/apiEndpoints.js
+++ b/ui/apps/platform/cypress/constants/apiEndpoints.js
@@ -80,6 +80,7 @@ export const images = {
 };
 
 export const auth = {
+    availableAuthProviders: '/v1/availableAuthProviders',
     loginAuthProviders: '/v1/login/authproviders',
     authProviders: '/v1/authProviders',
     authStatus: '/v1/auth/status',
--- a/ui/apps/platform/cypress/helpers/visit.js
+++ b/ui/apps/platform/cypress/helpers/visit.js
@@ -2,33 +2,45 @@ import * as api from '../constants/apiEndpoints';
 
 import { interceptRequests, waitForResponses } from './request';
 
+// Single source of truth for keys in optional staticResponseMap argument.
+export const availableAuthProvidersAlias = 'availableAuthProviders';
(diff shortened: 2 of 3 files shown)


bradr5 (approved):
Looks good 👍

PR #3583, opened 2022-10-26
ROX-12909: wire up active graph to PF topology library data model

alwayshooin:
getting the new PF topology graph to be hooked up to the network flow graph endpoint. the cluster fetch business logic and epoch refetching logic will probably need to be improved as we come around to it again.

image

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useParams } from 'react-router-dom';
 import { PageSection, Title, Flex, FlexItem } from '@patternfly/react-core';
 
+import { fetchNetworkFlowGraph } from 'services/NetworkService';
 import PageTitle from 'Components/PageTitle';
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -1,14 +1,45 @@
...
+
+    useEffect(() => {
+        fetchClustersAsArray()
+            .then((response) => {
+                setClusters(response);
+            })
+            .catch(() => {
+                // TODO
+            });
+    }, []);
```
vjwilson, line 29: I had a draft PR, just to test out cluster select, that was using @sachaudh 's suggestion of a custom hook for this.
https://github.com/stackrox/stackrox/pull/3122/files#diff-3117d6b9291b82097bae968b466a3a827737a027595aefde5378d1b2d1a59fca

Once this PR of yours is merged, I will "merge" your data flow changes with the select logic, in a fresh branch, so everybody can share the cluster-fetching logic.

Kum-ba-ya.

vjwilson (approved):
All the overloaded uses of the word "model" make my head hurt, but onward and downward, as Dante would say.

(If you don't get the reference, you probably don't want to know.)

PR #3584, opened 2022-10-26
Remove references to k8s/cve-list.json and k8s/checksum

RTann:
## Description

These directories are no longer used for Kubernetes vulnerabilities, so this PR removes all references of them.

This PR also replaces the CVE JSON and checksum for unit tests with something more relevant.

## Checklist

## Testing Performed

CI


central/cve/fetcher/manager_impl.go
```diff
@@ -48,11 +48,11 @@ func (m *orchestratorIstioCVEManagerImpl) initialize() {
 		m.mgrMode = online
 	}
 
-	if err := copyCVEsFromPreloadedToPersistentDirIfAbsent(utils.Istio); err != nil {
-		log.Errorf("could not copy preloaded istio CVE files to persistent volume %q: %v", path.Join(persistentCVEsPath, commonCveDir, istioCVEsDir), err)
+	if err := copyCVEsFromPreloadedToPersistentDirIfAbsent(); err != nil {
```
misberner, line 51: Can we take a step back and ask _why_ we're actually storing this on the persistent volume? This was a major contributor to the crashes requiring manual intervention. Is the CVE data not inserted into the database in our own format anyway? If the answer is no, I wonder if it wouldn't be better - if we want to store it on the PV in some JSON format or so - to do so only post conversion.

misberner (approved):
Looks good for what the PR intends to do, I just wonder if the Istio handling can be improved further.

central/cve/fetcher/manager_impl.go
RTann: I'm not sure the original reasoning behind all this, but I know we have plans to have this completely removed by 3.74, so I don't think we should bother with trying to update this from disk storage to a PV at this time

PR #3585, opened 2022-10-26
Add nil checks and unit test for NVDCVEToEmbeddedCVE

RTann:
## Description

This method does not account for a bunch of nils, so this PR adds some checks to ensure we can survive nil values.

## Checklist

## Testing Performed

CI


central/cve/converter/utils/convert_utils_test.go
```diff
@@ -1,242 +1,51 @@
...
-		assert.Nil(t, err)
-		assert.Equal(t, actualVul, expectedVul)
+func TestNVDCVEToEmbeddedCVE(t *testing.T) {
+	data, err := os.ReadFile("testdata/cve-list.json")
+	require.NoError(t, err)
+	var cveEntries []*schema.NVDCVEFeedJSON10DefCVEItem
+	require.NoError(t, json.Unmarshal(data, &cveEntries))
+
+	for _, cveEntry := range cveEntries {
+		t.Run(cveEntry.CVE.CVEDataMeta.ID, func(t *testing.T) {
```
misberner, line 32:
```suggestion
		cveEntry := cveEntry
		t.Run(cveEntry.CVE.CVEDataMeta.ID, func(t *testing.T) {
```
to de-alias

central/cve/converter/utils/convert_utils_test.go
```diff
@@ -1,242 +1,51 @@
...
+			embeddedCVE, err := NVDCVEToEmbeddedCVE(cveEntry, Istio)
+			assert.NoError(t, err)
+
+			assert.Equal(t, cveEntry.CVE.CVEDataMeta.ID, embeddedCVE.Cve)
+			assert.True(t, embeddedCVE.CvssV2 != nil || embeddedCVE.CvssV3 != nil)
+			if embeddedCVE.CvssV3 != nil {
```
misberner, line 38: It looks like you are testing the internal consistency iof `embeddedCVE` here. I wonder if it would not be better to also tie this to some of the `cveEntry` data. Something like
```suggestion
			if cveEntry.Impact != nil && cveEntry.Impact.BaseMetricsV3 != nil {
				assert.NotNil(t, embeddedCVE.CvssV3)
```
etc.

misberner (approved):
Looks fine, but I think the unit test could be improved to cover more of the actual transformation, rather than internal consistency of the result.

central/cve/converter/utils/convert_utils_test.go
RTann: Good catch, I forgot this is a slice of pointers

central/cve/converter/utils/convert_utils_test.go
RTann: Good call

mtodor:
@RTann Unfortunately, I have provided PR that fixes missing `nil` checks. It's already merged: https://github.com/stackrox/stackrox/pull/3532

Now that will make a lot of conflicts with this PR. 😞 

What I have added in my PR is some kind of "fuzz" testing, where the function is called for any missing field (or path), to ensure that it does not panic anymore.

RTann:
@house-d use https://github.com/stackrox/stackrox/pull/3532, though I think that PR is missing something. I will followup in another PR

PR #2150, opened 2022-06-22
ROX-11349: Add rationale and guidance text according to product feedback

fredrb:
## Description

Adds text fields updated according to product's suggestion in [ROX-11110](https://issues.redhat.com/browse/ROX-11110). 

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] ~~Unit test and regression tests added~~
- [x] Evaluated and added CHANGELOG entry if required
- [ ] ~~Determined and documented upgrade steps~~
- [ ] ~~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

PR #3597, opened 2022-10-27
ROX-13265: Fix default policy "Deployments should have at least one ingress Network Policy"

fredrb:
## Description

Another PR (#2150) introduced changes to the default policy "Deployments should have at least one ingress Network Policy", but it didn't update the actual default policy JSON. This led new installations to not have the new fields added. 

This PR:
1. Adds the fields to the default policy
2. Creates another migration exactly like the original for customers that did a fresh installation after it

## Checklist


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   to allow using the `restricted` and `restricted-v2` SCCs.
 - The service account "central", which is used by the central deployment, will now include `get` and `list` access to the following resources in the namespace where central is deployed to:
   `pods`, `events`, and `namespaces`. This fixes an issue when generating diagnostic bundles to now correctly include all relevant information within the namespace of central.
+- ROX-13265: Fix missing rationale and remediation texts for default policy "Deployments should have at least one ingress Network Policy" on new installations
```
md2119, line 39: The changelog indicates `on new installations` but there is a migration being added through this PR. Migrations are only needed to make changes to existing installations.

md2119 (changes requested):

md2119:
> @c-du - does there need to be an equivalent migration for postgres? Or since postgres doesn't release until 73, there's no reason to do so?

I believe the `m`'s execute before the `n`'s, thus when policy store is being migrated to postgres, the rocksdb store already has the required changes, and nothing more is required to do.

CHANGELOG.md
theencee: I would just remove "on new installations". This fix will add it for new **and** any older installations that missed it last time.

fredrb: Done

md2119 (approved):

PR #3599, opened 2022-10-27
Implement collection attacher component.

dvail:
## Description

Adds a UI component that allows users to attach and detach "embedded" collections from a collection configuration.

When the initial collection data is loaded, a secondary request will fetch metadata for all of the embedded collections of the initial collection. These embedded collections will make up the attached list in the widget, and are not paginated. 

Once this data is loaded and the form is rendered, the collection attacher component will attempt to fetch a page of detached collections (everything else in the system). Since this secondary request may include collections that are already part of the attached list, we need to filter the results client side. If this filtering results in less than half a page load, we fetch an additional page worth of data. The idea being we want to prevent situations where "View more" returns a full page sometimes, and then very few another time (possibly even zero), which could lead to looking like buggy behavior.

When the user enters a search term, the locally cached list of detached collections is cleared out and the page counter is reset. This will require refetching the detached collections, now filtered by search term. This could cause some wasted requests, but the alternative of keeping the cache and filtering client side had larger problems. If we don't clear cache and all data has not yet been loaded, there would be two ways to handle the follow up "view more" requests:

1. Pass the search term along with the request. This would have good performance, but would leave holes in the local cache data if the user then removed the search term. We would need to clear the local cache anyway or be unable to present the full data set to the user.
2. _Don't_ pass the search term along with the request, and filter the results client side. This would build up the local cache cleanly and show the user the full, accurate set of data. The downside here is that a search term with few, or zero, matches would create a string of requests that load all of the collections from the server. Worst case would be multiple sequential requests with a loading spinner that eventually resulted in no new data.

I've documented some potential workarounds and optimizations for this in the code that we can do as follow ups if needed.

## Checklist

## Testing Performed

Visit the collections page, and select a collection that has embedded collection data:
image

Clicking the "Attach" button should remove items from the bottom list and place them in the top list:
image
image

Clicking the "Detach" button should move items from the top list to the bottom list:
image
image

Entering a search value should filter the top and bottom lists by collection name, case insensitive:
image
Clearing the filter should return the list to its previous state:
image

**To demo the "View more" behavior, I manually changed the `pageSize` and minimum threshold to be `1`.**

If there are more results on the server, a "View more" button should be displayed. 
image

Clicking this button should load another page of results:
image
image

Continuously clicking this button will load more and more results from the server.

Note that when we get to the fifth item, two network requests are sent instead of one. This is because the first request contains only "Notable deployments", which is already present in the attached list. Since this would result in no new items in the detached collections list, another request is sent.
image

When all results have been loaded by the server, the view more button disappears.
image

Test that layout is usable on small screens:
image

### DIFF
--- a/ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
+++ b/ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
@@ -79,6 +79,8 @@ function BacklogTable<Item>({
                                         variant="link"
                                         onClick={() => listAction(item)}
                                         icon={actionIcon}
+                                        className="pf-u-text-nowrap"
+                                        isInline
                                     >
                                         {buttonText}
                                     </Button>
--- a/ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
+++ b/ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
@@ -1,0 +1,166 @@
+import React, { Key, ReactNode } from 'react';
+import {
+    Badge,
+    Button,
+    EmptyState,
+    EmptyStateIcon,
+    EmptyStateVariant,
+    Flex,
+    FormGroup,
+    Title,
+} from '@patternfly/react-core';
+import { CubesIcon, MinusCircleIcon, PlusCircleIcon } from '@patternfly/react-icons';
+
+import {
+    BaseCellProps,
+    TableComposable,
+    TableVariant,
+    Tbody,
+    Td,
+    Tr,
+} from '@patternfly/react-table';
+
+type BacklogTableProps<Item> = {
+    type: 'selected' | 'deselected';
+    label: string | undefined;
+    items: Item[];
+    listAction: (item: Item) => void;
+    rowKey: (item: Item) => Key;
+    cells: {
+        name: string;
+        render: (item: Item) => ReactNode;
+        width?: BaseCellProps['width'];
+    }[];
+    buttonText: string;
+};
+
+function BacklogTable<Item>({
+    type,
+    label,
+    items,
+    listAction,
+    rowKey,
+    cells,
+    buttonText,
+}: BacklogTableProps<Item>) {
+    const actionIcon =
+        type === 'selected' ? (
+            <MinusCircleIcon color="var(--pf-global--danger-color--200)" />
+        ) : (
+            <PlusCircleIcon color="var(--pf-global--primary-color--100)" />
+        );
+    return (
+        <FormGroup
+            label={
+                <>
+                    {label}
+                    <Badge className="pf-u-ml-sm" isRead>
+                        {items.length}
+                    </Badge>
+                </>
+            }
+        >
+            {items.length > 0 ? (
+                <TableComposable aria-label={label} variant={TableVariant.compact}>
+                    <Tbody>
+                        {items.map((item) => (
+                            <Tr key={rowKey(item)}>
+                                {cells.map(({ name, width, render }) => (
+                                    <Td key={name} dataLabel={name} width={width}>
+                                        {render(item)}
+                                    </Td>
+                                ))}
+                                <Td width={10} dataLabel="Item action">
+                                    <Button
+                                        variant="link"
+                                        onClick={() => listAction(item)}
+                                        icon={actionIcon}
+                                    >
+                                        {buttonText}
+                                    </Button>
+                                </Td>
+                            </Tr>
+                        ))}
+                    </Tbody>
+                </TableComposable>
+            ) : (
+                <EmptyState variant={EmptyStateVariant.xs}>
+                    <EmptyStateIcon icon={CubesIcon} />
+                    <Title headingLevel="h4">No items remaining</Title>
+                </EmptyState>
+            )}
+        </FormGroup>
+    );
+}
+
+export type BacklogListSelectorProps<Item> = {
+    selectedOptions: Item[];
+    deselectedOptions: Item[];
+    onSelectItem: (item: Item) => void;
+    onDeselectItem: (item: Item) => void;
+    onSelectionChange?: (selected: Item[], deselected: Item[]) => void;
+    rowKey: (item: Item) => Key;
+    cells: {
+        name: string;
+        render: (item: Item) => ReactNode;
+    }[];
+    selectedLabel?: string;
+    deselectedLabel?: string;
+    selectButtonText?: string;
+    deselectButtonText?: string;
+};
+
+function BacklogListSelector<Item>({
+    selectedOptions,
+    deselectedOptions,
+    onSelectItem,
+    onDeselectItem,
+    onSelectionChange = () => {},
+    rowKey,
+    cells,
+    selectedLabel = 'Selected items',
+    deselectedLabel = 'Deselected items',
+    selectButtonText = 'Add',
+    deselectButtonText = 'Remove',
+}: BacklogListSelectorProps<Item>) {
+    function onSelect(item: Item) {
+        onSelectItem(item);
+        onSelectionChange(
+            selectedOptions.concat(item),
+            deselectedOptions.filter((option) => option !== item)
+        );
+    }
+
+    function onDeselect(item: Item) {
+        onDeselectItem(item);
+        onSelectionChange(
+            selectedOptions.filter((option) => option !== item),
+            deselectedOptions.concat(item)
+        );
+    }
+
+    return (
+        <Flex direction={{ default: 'column' }} spaceItems={{ default: 'spaceItemsXl' }}>
+            <BacklogTable
+                type="selected"
+                label={selectedLabel}
+                items={selectedOptions}
+                listAction={onDeselect}
+                buttonText={deselectButtonText}
+                rowKey={rowKey}
+                cells={cells}
+            />
+            <BacklogTable
+                type="deselected"
+                label={deselectedLabel}
+                items={deselectedOptions}
+                listAction={onSelect}
+                rowKey={rowKey}
+                buttonText={selectButtonText}
+                cells={cells}
+            />
+        </Flex>
+    );
+}
+
+export default BacklogListSelector;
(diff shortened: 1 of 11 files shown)


dvail:
* master
  * **PR #3599** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3599" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3599?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -85,7 +85,7 @@ function CollectionsTable({
     const onSearchInputChange = useMemo(
         () =>
             debounce(
-                (value: string) => setSearchFilter({ Collection: value }),
+                (value: string) => setSearchFilter({ 'Collection Name': value }),
```
dvail, line 88: This is an unrelated change, but not worth a separate PR.

The BE team is using `Collection Name` instead of `Collection` for this search label.

ui/apps/platform/src/utils/searchUtils.ts
```diff
@@ -185,10 +185,13 @@ export function flattenFilterValue<UndefinedFallback>(
 export function getListQueryParams(
     searchFilter: SearchFilter,
     sortOption: ApiSortOption,
-    page: number,
-    pageSize: number
+    page?: number,
```
dvail, line 188: Most backend APIs can support un-paginated requests, so we don't need to enforce that these values are set.

In this specific use case, we need to request all attached collections by id without pagination.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
```diff
@@ -1,7 +1,95 @@
...
+                    isInline
+                    title="There was an error loading more collections"
+                />
+            )}
+            {hasMore && (
+                <Button
+                    className="pf-u-align-self-flex-start"
+                    variant="secondary"
+                    onClick={() => fetchMore(search)}
+                    isLoading={fetchMoreLoading}
```
pedrottimark, line 86: Most boolean state variables have `is` or occasionally `has` prefix:

```js
isLoading={isFetchMoreLoading}
```

An grammatical alternative in this case which I mention, not necessarily recommend: `isFetchingMore`

ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
```diff
@@ -0,0 +1,166 @@
...
+                <>
+                    {label}
+                    <Badge className="pf-u-ml-sm" isRead>
+                        {items.length}
+                    </Badge>
+                </>
+            }
+        >
+            {items.length > 0 ? (
+                <TableComposable aria-label={label} variant={TableVariant.compact}>
```
pedrottimark, line 64: Can you inspect table cells to see why centered alignment in responsive view?

Here is an example from /main/access-control/permission-sets for comparison:
Permission_sets

ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
```diff
@@ -0,0 +1,166 @@
...
+            <PlusCircleIcon color="var(--pf-global--primary-color--100)" />
+        );
+    return (
+        <FormGroup
+            label={
+                <>
+                    {label}
+                    <Badge className="pf-u-ml-sm" isRead>
+                        {items.length}
+                    </Badge>
```
pedrottimark, line 59: Thoughts about badge, which I was happy to see after designs had moved away from it for a while:
* Attached collections: Does search filter apply to this table? Length seems meaningful only if not filtered.
* Detached collections: Length seems meaningful only if `!hasMore` and maybe only without search filter.

ui/apps/platform/src/Containers/Collections/types.ts
```diff
@@ -72,6 +72,8 @@ export function isByLabelSelector(
     return isByLabelField(selector.field);
 }
 
+export type CollectionSlim = { id: string; name: string; description: string };
```
dvail, line 75: Note to self - we can just use the full type since we are querying the server for the full `Collection` object here.
- [ ] Replace `CollectionSlim` with `Collection`

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
dvail: I think the grammatical alternative reads better in this case, updated the prop name.

ui/apps/platform/src/Components/PatternFly/BacklogListSelector.tsx
dvail: Agreed that ht badge has limited utility when the full dataset is not present, and the one case where it would be meaningful (unfiltered attached collections) is likely going to be a small enough number that the badge is fairly unimportant.

I left the ability to show the badge in the underlying component though, since if we get reuse out of it there is the possibility we use it against data that is complete and unfilterable.

ui/apps/platform/src/Containers/Collections/types.ts
dvail: Instead of `Collection`, I just passed `CollectionResponse[]` directly since the parsing of the response into a stricter data type is not needed for this piece of the UI.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
```diff
@@ -1,7 +1,95 @@
...
+    );
+
+    const selectedOptions = attached.filter(compareNameLowercase(search));
+    const deselectedOptions = detached.filter(compareNameLowercase(search));
+
+    return (
+        <Flex direction={{ default: 'column' }} spaceItems={{ default: 'spaceItemsXl' }}>
+            <SearchInput
+                aria-label="Search by name"
+                placeholder="Search by name"
```
pedrottimark, line 57: Although I see **Search** Google and Gmail, **Filter** seems more common for placeholder in UI.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
```diff
@@ -1,7 +1,95 @@
...
+                value={search}
+                onChange={onSearchInputChange}
+            />
+            <BacklogListSelector
+                selectedOptions={selectedOptions}
+                deselectedOptions={deselectedOptions}
+                onSelectItem={({ id }) => attach(id)}
+                onDeselectItem={({ id }) => detach(id)}
```
pedrottimark, line 65: An observation, not a change request, that change in terminology from attach/detach to select/deselect is what I would expect for a reusable component, what do you think are pro and con:
* pro: how likely to reuse in a context where application level is other than attach/detach
* con: slight substitution by reader and possible effect on Find in Files

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
```diff
@@ -1,7 +1,95 @@
...
+                onSelectionChange={onSelectionChange}
+                rowKey={({ id }) => id}
+                cells={selectorListCells}
+                selectedLabel="Attached collections"
+                deselectedLabel="Detached collections"
+                selectButtonText="Attach"
+                deselectButtonText="Detach"
+            />
+            {fetchMoreError && (
```
pedrottimark, line 74: First impression gave me a much needed smile: fetch more error, please ;)

pedrottimark (approved):
Approving and am available as needed if there is anything more to discuss.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
```diff
@@ -1,7 +1,95 @@
...
+import { CollectionResponse } from 'services/CollectionsService';
+import useEmbeddedCollections from './hooks/useEmbeddedCollections';
+
+const selectorListCells = [
+    {
+        name: 'Name',
+        render: ({ name }) => (
+            <Button variant="link" className="pf-u-pl-0" isInline>
+                {name}
+            </Button>
```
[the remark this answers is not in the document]
dvail: Yes that is exactly correct. In a follow up I'll likely move the `CollectionAttacher.tsx#selectorListCells` up to `CollectionFormPage` and pass through as props `CollectionFormPage -> CollectionForm -> CollectionAttacher` to accommodate the two use cases for the main page table and the modal table.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
dvail: Fair, there are multiple occurrences of **Filter** in similar contexts but this looks to be the first case where we use **Search**. This is kind of a hybrid case, as we are both searching and filtering the two lists with a single input, but I'll change for the sake of consistency.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
dvail: Hmm this is a good question. My gut feeling is that the default terminology for using this in other contexts would be "select/deselect". This component is similar to the [dual list selector PF](https://www.patternfly.org/v4/components/dual-list-selector) component which uses terms that are fairly close. I think the attach/detach are specific, but not exclusive, to the semantics behind Collections.

That said it doesn't feel great having a disconnect between the props and the only use case. (The `useSelectToggle()` hook suffers from this mismatch as well.) I'm OK with changing the terminology or leaving as-is. Do you think people are more likely to Find in Files or Ctrl+Click to navigate around this code in the future?

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
dvail: Hahah now I can't read it any other way! Seriously though, do you think there is enough chance for confusion that we should change this to `fetchingError`?

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
pedrottimark: Your choice. You caught me: I often use Find in Files by mental muscle memory from less capable editors.

ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
pedrottimark: It just hit me funny at that moment. In first reading a few days ago, it seemed nicely consistent.

PR #3600, opened 2022-10-27
[ROX-12622] : Add create collection service

charmik-redhat:
## Description

Added REST endpoint for creating a new collection.

## Checklist

## Testing Performed

Unit tests added.

### DIFF
--- a/central/resourcecollection/service/service_impl.go
+++ b/central/resourcecollection/service/service_impl.go
@@ -98,14 +98,10 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
 // CreateCollection creates a new collection from the given request
 func (s *serviceImpl) CreateCollection(ctx context.Context, request *v1.CreateCollectionRequest) (*v1.CreateCollectionResponse, error) {
 	if !features.ObjectCollections.Enabled() {
-		return nil, nil
+		return nil, errors.New("Resource collections is not enabled")
 	}
 
-	if request.GetName() == "" {
-		return nil, errors.Wrap(errox.InvalidArgs, "Collection name should not be empty")
-	}
-
-	collection, err := collectionRequestToCollection(ctx, request, true)
+	collection, err := collectionRequestToCollection(ctx, request, true, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -118,10 +114,18 @@ func (s *serviceImpl) CreateCollection(ctx context.Context, request *v1.CreateCo
 	return &v1.CreateCollectionResponse{Collection: collection}, nil
 }
 
-func collectionRequestToCollection(ctx context.Context, request collectionRequest, isCreate bool) (*storage.ResourceCollection, error) {
-	user := utils.UserFromContext(ctx)
-	if user == nil {
-		return nil, errors.New("User identity not provided")
+func collectionRequestToCollection(ctx context.Context, request collectionRequest, isCreate bool, getID func() string) (*storage.ResourceCollection, error) {
+	if request.GetName() == "" {
+		return nil, errors.Wrap(errox.InvalidArgs, "Collection name should not be empty")
+	}
+
+	slimUser := utils.UserFromContext(ctx)
+	if slimUser == nil {
+		return nil, errors.New("Could not determine user identity from provided context")
+	}
+
+	if len(request.GetResourceSelectors()) == 0 {
+		return nil, errors.Wrap(errox.InvalidArgs, "No resource selectors were provided")
 	}
 
 	timeNow := protoconv.ConvertTimeToTimestamp(time.Now())
@@ -130,13 +134,15 @@ func collectionRequestToCollection(ctx context.Context, request collectionReques
 		Name:              request.GetName(),
 		Description:       request.GetDescription(),
 		LastUpdated:       timeNow,
-		UpdatedBy:         user,
+		UpdatedBy:         slimUser,
 		ResourceSelectors: request.GetResourceSelectors(),
 	}
 
 	if isCreate {
-		collection.CreatedBy = user
+		collection.CreatedBy = slimUser
 		collection.CreatedAt = timeNow
+	} else if getID != nil {
+		collection.Id = getID()
 	}
 
 	if len(request.GetEmbeddedCollectionIds()) > 0 {
--- a/central/resourcecollection/service/service_impl.go
+++ b/central/resourcecollection/service/service_impl.go
@@ -2,19 +2,25 @@ package service
 
 import (
 	"context"
+	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/resourcecollection/datastore"
 	"github.com/stackrox/rox/central/role/resources"
 	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/permissions"
 	"github.com/stackrox/rox/pkg/errox"
 	"github.com/stackrox/rox/pkg/features"
+	"github.com/stackrox/rox/pkg/grpc/authn"
 	"github.com/stackrox/rox/pkg/grpc/authz"
 	"github.com/stackrox/rox/pkg/grpc/authz/or"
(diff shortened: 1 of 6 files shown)


keyallis (changes requested):
The service layer should be pretty slim, left some review comments of things you can trim down on since the functionality will be in the datastore layer

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +89,64 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+
+func extractUserIdentity(ctx context.Context) *storage.SlimUser {
+	ctxIdentity := authn.IdentityFromContextOrNil(ctx)
+	if ctxIdentity == nil {
+		return nil
+	}
+
+	return &storage.SlimUser{
+		Id:   ctxIdentity.UID(),
+		Name: ctxIdentity.FullName(),
```
charmik-redhat, line 150: @keyallis @md2119 
Identity  has FullName and FriendlyName among other things. Is it correct to set the name to FullName here?

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +89,64 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+		return nil, errors.Wrap(errox.AlreadyExists, "A collection with that name already exists")
+	}
+
+	creator := extractUserIdentity(ctx)
+	if creator == nil {
+		return nil, errors.New("User identity not provided")
+	}
+
+	collection := &storage.ResourceCollection{
+		Id:                uuid.NewV4().String(),
```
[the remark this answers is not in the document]
charmik-redhat: Don't see that in datastore's `AddCollection` implementation. Is that something currently in review? Or I can add this line to datastore level.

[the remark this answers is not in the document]
keyallis: #3528

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,58 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+
+	timeNow := protoconv.ConvertTimeToTimestamp(time.Now())
+
+	collection := &storage.ResourceCollection{
+		Name:              request.GetName(),
+		Description:       request.GetDescription(),
+		LastUpdated:       timeNow,
+		UpdatedBy:         user,
+		ResourceSelectors: request.GetResourceSelectors(),
+	}
```
keyallis, line 135: You might have to do some digging to figure out a clean way of doing this (maybe [generics](https://go.dev/doc/tutorial/generics)!) but for future work we need to also include the `Id` field population in the object creation for Update requests

keyallis (approved):
Just one comment on future work but otherwise LGTM

central/resourcecollection/service/service_impl.go
charmik-redhat: Ah yes, I will add an `idGetter func() string` to the arguments, and update can then pass `func() {return  request.GetId() } ` as that func.

About generics, I thought to do it using generics, but since we are calling functions `GetName`, `GetDescription`, etc on the passed obj, if I call them on generic type R, compilation would fail. The solution for that is to constrain the generic type using an interface. In this case the constraining interface would look like

```
type collectionRequest interface {
    *v1.CreateCollectionRequest | *v1.DryRunCollectionRequest | *v1.UpdateCollectionRequest
}

func [R collectionRequest] collectionRequestToCollection(ctx context, request R, isCreate bool) {...}
```

However, this fails to compile too since go 1.18 doesn't have a good support for union of structs within a constraining interface. Since  *v1.CreateCollectionRequest and others are structs and not interfaces themselves, `GetName` are yet not accessible from collectionRequest.

The workaround for this is to then add those methods to the constraining interface as well, which makes the interface look like
```
type collectionRequest interface {
	*v1.CreateCollectionRequest

	GetName() string
	GetDescription() string
	GetResourceSelectors() []*storage.ResourceSelector
	GetEmbeddedCollectionIds() []string
}
```

This works, but then I still had to put those methods in the interface. so I just avoided the generics altogether for this case.

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,58 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
 		Deployments: nil,
 	}, nil
 }
+
+// CreateCollection creates a new collection from the given request
+func (s *serviceImpl) CreateCollection(ctx context.Context, request *v1.CreateCollectionRequest) (*v1.CreateCollectionResponse, error) {
+	if !features.ObjectCollections.Enabled() {
+		return nil, nil
```
md2119, line 101: return error

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,58 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+		return nil, err
+	}
+
+	return &v1.CreateCollectionResponse{Collection: collection}, nil
+}
+
+func collectionRequestToCollection(ctx context.Context, request collectionRequest, isCreate bool) (*storage.ResourceCollection, error) {
+	user := utils.UserFromContext(ctx)
+	if user == nil {
+		return nil, errors.New("User identity not provided")
```
md2119, line 124: because it implied from the context `could not determine user identity` seems more appropriate

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,58 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+	}
+
+	timeNow := protoconv.ConvertTimeToTimestamp(time.Now())
+
+	collection := &storage.ResourceCollection{
+		Name:              request.GetName(),
+		Description:       request.GetDescription(),
+		LastUpdated:       timeNow,
+		UpdatedBy:         user,
+		ResourceSelectors: request.GetResourceSelectors(),
```
md2119, line 134: valid resource selectors are non-empty

md2119 (approved):
Some minor comments, otherwise lgtm.

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,58 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+
+// CreateCollection creates a new collection from the given request
+func (s *serviceImpl) CreateCollection(ctx context.Context, request *v1.CreateCollectionRequest) (*v1.CreateCollectionResponse, error) {
+	if !features.ObjectCollections.Enabled() {
+		return nil, nil
+	}
+
+	if request.GetName() == "" {
+		return nil, errors.Wrap(errox.InvalidArgs, "Collection name should not be empty")
+	}
```
md2119, line 106: move this into `collectionRequestToCollection(...)`

central/resourcecollection/service/service_impl.go
charmik-redhat: Added an error on empty resource selectors.

central/resourcecollection/service/service_impl.go
```diff
@@ -83,3 +94,64 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+		UpdatedBy:         slimUser,
+		ResourceSelectors: request.GetResourceSelectors(),
+	}
+
+	if isCreate {
+		collection.CreatedBy = slimUser
+		collection.CreatedAt = timeNow
+	} else if getID != nil {
+		collection.Id = getID()
+	}
```
md2119, line 146: implying `getID` wrt `isCreate` is confusing. Strongly recommend dropping this check and explicitly basing the condition on create and update workflow.

charmik-redhat: Ah, I merged this PR before noticing the comment. I will add it in my Delete collections service PR. Should I change to something like below?
```
if isCreate {...}
else {
    if getID == nil {
        return nil, err
    }
    collection.Id = getID()
}
```

central/resourcecollection/service/service_impl.go
charmik-redhat: @keyallis

PR #3603, opened 2022-10-27
Improve helper functions for credentialexpiry and systemconfig integration tests

pedrottimark:
## Description

Practice for collections and follow up improvements to visit helpers in #3577

### Objectives

* Encapsulate page and endpoint addresses in component/container specific helpers file.
* Encapsulate selectors which are needed only in specific helpers file.
* Call relevant generic functions from specific functions.

### Review

Cypress data structures:
* `RouteMatcher`: https://docs.cypress.io/api/commands/intercept#Matching-with-RouteMatcher
    For example, `{ method: 'GET', url: '/v1/collections' }` however, it needed `?query=*` query string to match!
* `StaticResponse`: https://docs.cypress.io/api/commands/intercept#StaticResponse-objects
    For example, response body: `{ body: { collections: […] } }`
    For example, fixture file: `{ fixture: 'collections/collectionsWhatever.json' }`

Here are conventions-in-progress for integration tests:
* `whateverAlias`
    for REST endpoint: delete **/v1/** and usually omit search query (prepend with method if other than GET)
    for GraphQL endpoints: delete **/api/graphql?opname=**
* `routeMatcherMap` object properties have alias for key and RouteMatcher object for value 
* `staticResponseMap` object properties have alias for key and StaticResponse object for value

### Changed files

1. Edit cypress/constants/SystemConfigPage.js
    * Delete `systemConfigUrl` because encapsulated in helpers file.
    * Delete `saveButton` because encapsulated in helpers file.

2. Edit cypress/constants/apiEndpoints.js
    * Delete `certExpiry` and `certGen` because encapsulated in credentialExpiry helpers file.
    * Delete `system` because `config` and `configPublic` are encapsulated in systemConfig and visit helpers files, respectively.

3. Add cypress/helpers/credentialExpiry.js
    * Add 1 generic `visitSystemConfigurationWithCredentialExpiryBanner` and 2 component-specific functions.
    * Add 1 generic `interactAndWaitForCertificateDownload` and 2 component-specific functions.

4. Add cypress/helpers/systemConfig.js
    * Add visit helper functions.
    * Add save helper function.

5. Edit cypress/helpers/visit.js
    * Replace `api.system.configPublic` with `'/v1/config/public'` because it is the only reference to the endpoint.

6. Rename cypress/integration/certexpiration.test.js as credentialExpiry.test.js
    * Align test file names with request endpoint and React component name.
    * Replace main dashboard which has several slower requests with System Configuration page which has one faster request.
    * Replace `mockCertExpiryAndVisitHomepage` function with visit helper functions.
    * Replace intercept and wait for permissions with optional `staticResponseForPermissions` argument.
    * Replace incomplete mock data with fixture.
    * Replace intercept and wait for certificate download with interact helper functions.

7. Edit cypress/integration/systemconfig.test.js
    * Replace 3 local functions with import from helpers file.
    * Leave the rest of the functions alone, because out of scope for the objectives.
    * Add a test of conditional rendering for read only permission.

## Checklist

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/integration/credentialExpiry.test.js
+++ b/ui/apps/platform/cypress/integration/credentialExpiry.test.js
@@ -1,41 +1,34 @@
 import dateFns from 'date-fns';
 import withAuth from '../helpers/basicAuth';
-import * as api from '../constants/apiEndpoints';
 import { selectors } from '../constants/CertExpiration';
-import { visitMainDashboard } from '../helpers/main';
+import {
+    interactAndWaitForCentralCertificateDownload,
+    interactAndWaitForScannerCertificateDownload,
+    renderCentralCredentialExpiryBanner,
+    renderScannerCredentialExpiryBanner,
+} from '../helpers/credentialExpiry';
 
-describe('Cert Expiration Banner', () => {
+describe('Credential Expiry', () => {
     withAuth();
 
-    const mockCertExpiryAndVisitHomepage = (endpoint, expiry) => {
-        cy.intercept('GET', endpoint, {
-            body: { expiry },
-        }).as('certExpiry');
-        visitMainDashboard();
-        cy.wait('@certExpiry');
-    };
-
-    describe('Central', () => {
+    describe('for Central', () => {
         it('should not display banner if central cert is expiring more than 14 days later', () => {
             const expiry = dateFns.addHours(dateFns.addDays(new Date(), 15), 1);
-            mockCertExpiryAndVisitHomepage(api.certExpiry.central, expiry);
+
+            renderCentralCredentialExpiryBanner(expiry);
 
             cy.get(selectors.centralCertExpiryBanner).should('not.exist');
         });
 
         it('should display banner without download button if user does not have the required permission', () => {
-            cy.intercept('GET', api.permissions.mypermissions, {
-                body: {
-                    globalAccess: 'READ_ACCESS',
-                    resourceToAccess: {
-                        VulnerabilityManagementRequests: 'READ_ACCESS',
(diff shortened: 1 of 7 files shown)


ui/apps/platform/cypress/integration/credentialExpiry.test.js
```diff
@@ -1,41 +1,34 @@
...
-        cy.wait('@certExpiry');
-    };
-
-    describe('Central', () => {
+    describe('for central', () => {
         it('should not display banner if central cert is expiring more than 14 days later', () => {
             const expiry = dateFns.addHours(dateFns.addDays(new Date(), 15), 1);
-            mockCertExpiryAndVisitHomepage(api.certExpiry.central, expiry);
+
+            renderCentralCredentialExpiryBanner(expiry);
```
dvail, line 18: One thought on naming. At first look, I thought this function might have manipulated the DOM in some way due to the `render` prefix but it turns out it uses the standard `intercept/wait` cycle to mock the request and cause the banner to render. Any reason not to go with something like `visitWithCentralCredentialExpiryBanner`?

dvail (approved):
LGTM, just one comment about a naming convention that was confusing to me at first.

bradr5 (approved):

ui/apps/platform/cypress/integration/credentialExpiry.test.js
pedrottimark: Good call.

Because visit functions without direct object in name have path as first argument, I used a more verbose alternative `visitSystemConfigurationWithCentralCredentialExpiryBanner` although which page is an implementation detail.

PR #3605, opened 2022-10-27
ROX-13318: Fix skipTLSVerify property on Generic Webhook integration form

vjwilson:
## Description

Casing is hard. Fixes a casing typo in one of the fields in the Generic Webhook integration objection.

## Checklist

## Testing Performed

Manual testing
Screen Shot 2022-10-27 at 5 14 21 PM

### DIFF
--- a/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/GenericWebhookIntegrationForm.tsx
+++ b/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/GenericWebhookIntegrationForm.tsx
@@ -31,7 +31,7 @@ import FormLabelGroup from '../FormLabelGroup';
 export type GenericWebhookIntegration = {
     generic: {
         endpoint: string;
-        skipTlsVerify: boolean;
+        skipTLSVerify: boolean;
         auditLoggingEnabled: boolean;
         caCert: string;
         username: string;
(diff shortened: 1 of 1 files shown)


pedrottimark (approved):
Yeah, complicated :(
1. The left shift key did not know what the right shift key was doing:
    * Generic `skipTLSVerify` see https://github.com/stackrox/stackrox/blob/master/proto/storage/notifier.proto#L110
    * Syslog `skipTlsVerify` see https://github.com/stackrox/stackrox/blob/master/proto/storage/notifier.proto#L141-L142
2. Too bad, so sad. Incorrect local type instead of import correct frontend proto type:
    https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/types/notifier.proto.ts#L88

bradr5 (approved):

vjwilson:
> Yeah, complicated :(
> 
> 1. The left shift key did not know what the right shift key was doing:
>    
>    * Generic `skipTLSVerify` see https://github.com/stackrox/stackrox/blob/master/proto/storage/notifier.proto#L110
>    * Syslog `skipTlsVerify` see https://github.com/stackrox/stackrox/blob/master/proto/storage/notifier.proto#L141-L142
> 2. Too bad, so sad. Incorrect local type instead of import correct frontend proto type:
>    https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/types/notifier.proto.ts#L88

I let the Core Workflows team know about #1, but I'd rather not change it that now that it's "out in the wild".

I removed the locally defined type for this integration, in favor of that import, which addresses #2 for this file.

PR #3220, opened 2022-09-28
ROX-12319: implement smoke test step with groovy test filter

jschnath:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #3611, opened 2022-10-28
ROX-12953: figure out last 4 versions of sensor automatically

jschnath:
## Description

Script to automatically determine the latest releases of the last 4 sensor versions. Still needs to be integrated into the version compatibility tests, blocked by https://github.com/stackrox/stackrox/pull/3220

## Checklist

## Testing Performed

Manually ran this successfully. Still need to run it on Prow as API calls or git commands may behave differently there.


.openshift-ci/sensorversions.py
```diff
@@ -0,0 +1,120 @@
...
+    return __getLatestTags(current_version, tags, num_versions)
+
+def main(argv):
+    latestversions = getLastSensorVersionsFromGitTagsCLI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromGitTagsAPI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromQuayAPI(argv[1], 4)
+    printversions = ""
+    for version in latestversions:
+        printversions += str(version) + " "
+    print(printversions)
```
msugakov, line 117: This can be simply
```suggestion
print(" ".join(printversions))
```

However, I think that if you plan to consume the output of this CLI, it may be more convenient to separate versions with newlines.

.openshift-ci/sensorversions.py
```diff
@@ -0,0 +1,120 @@
...
+# preferably use this to avoid API calls if possible
+def getLastSensorVersionsFromGitTagsCLI(current_version, num_versions):
+    rawtags = __cliOutputToTags(subprocess.check_output(["git", "tag", "--list"]))
+    tags = __filterGitCLITags(rawtags)
+    return __getLatestTags(current_version, tags, num_versions)
+
+def main(argv):
+    latestversions = getLastSensorVersionsFromGitTagsCLI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromGitTagsAPI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromQuayAPI(argv[1], 4)
```
msugakov, line 113: I would suggest to get rid of dead code. On one hand it is sorry to loose it but on the other hand less code - less things to maintain, simpler our life.

.openshift-ci/sensorversions.py
```diff
@@ -0,0 +1,120 @@
...
+    numTags = []
+    for t in tags:
+        numTags.append(__splitVersion(t))
+    return sorted(numTags)
+
+def __splitVersion(version):
+    digits = re.search(r"(\d+)\.(\d+)\.\D*(\d+)", version)
+    return int(digits.group(1)), int(digits.group(2)), int(digits.group(3))
+
+def __getLatestTags(current_version, tags, num_versions):
```
msugakov, line 73: [nit] I think, leading underscores only clutter the code. In a simple CLI that's not object-oriented, I _think_, it should be fine to write functions as "public", i.e. without leading `__`. No strong opinion though.

.openshift-ci/sensorversions.py
```diff
@@ -0,0 +1,120 @@
...
+    latestversions = getLastSensorVersionsFromGitTagsCLI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromGitTagsAPI(argv[1], 4)
+    #latestversions = getLastSensorVersionsFromQuayAPI(argv[1], 4)
+    printversions = ""
+    for version in latestversions:
+        printversions += str(version) + " "
+    print(printversions)
+
+if (__name__ == "__main__"):
+    main(sys.argv)
```
msugakov, line 120: I'm missing the code which calls this `.openshift-ci/sensorversions.py` file. Do you plan to add it in the same PR?

msugakov (commented):
Did a shallow read. Will gladly re-review once you clean up unused code and/or integrate the script in whatever script that orchestrates compatibility tests.

.openshift-ci/sensorversions.py
jschnath: This was just for manual testing of the functions, I'll remove the main function and prints later

.openshift-ci/sensorversions.py
jschnath: Yeah, we probably won't be needing any of the API calls since git tag --list works

.openshift-ci/sensorversions.py
jschnath: I learned to do it this way as good form since those functions are not supposed to be called from the outside, I'll change it if that's our common practice though.

.openshift-ci/sensorversions.py
jschnath: Yes, I plan to add it once I get https://github.com/stackrox/stackrox/pull/3220 merged, it's just one line anyways.

jschnath: Which is hopefully today, I've just solved the latest prow issue and it's in a run.

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+    numTags = []
+    for t in tags:
+        numTags.append(splitVersion(t))
+    return sorted(numTags)
+
+def splitVersion(version):
+    digits = re.search(r"(\d+)\.(\d+)\.\D*(\d+)", version)
+    return int(digits.group(1)), int(digits.group(2)), int(digits.group(3))
+
+def getLatestTags(tags, num_versions):
```
fredrb, line 35: I think this can be simplified. You don't need to convert it to numbers and then back to strings. `sorted` will work just fine in numbered strings. 

Here's what I suggest you do:
1. Make a dictionary all versions based on their major versions: `{ 3.69: [3.69.0, 3.69.1 ...], 3.70: [ ... ], ...`
2. Sort the dictionary keys and get the top 4
3. Sort each bucket and get top 1

Here's a drop-in replacement that should work the same way without needing `transformTagsToNumbers`:

```python3
def getLatestTags(tags, num_versions):
    top_patch_version = defaultdict(int)
    for t in tags:
        [major, minor, patch] = t.split('.')
        k = '.'.join([major, minor])
        top_patch_version[k] = max(top_patch_version[k], int(patch))
    top_major_versions = sorted(list(top_patch_version.keys()), reverse=True)[:num_versions]
    return [t + '.' + str(top_patch_version[t]) for t in top_major_versions]
```

I quickly tested with your script and it _seems_ to work. But please, if you use the function above make sure it works for all the cases you thought of. 

This is just a suggestion to simplify the code.

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+    numericaltags = transformTagsToNumbers(tags)
+    _,y,_ = splitVersion(makeImageTag())
```
fredrb, line 37: Is `makeImageTag` command necessary here? Is there a scenario where the result of this command will be _lower_ than the latest release? This will be frequently run on PRs, which are on the latest major version anyway. 

The only case where this will be lower is when we're building patches for older releases. But in this case we might end up testing sensor versions that are already outside the support window. E.g.: patching 3.69.x and running compatibility tests for `3.68.x`, `3.67.x` and `3.66.x`. 

I'd prefer to simply run the tests with the latest 4 released versions of sensor, regardless of the tag from `make image`.

scripts/ci/jobs/gke_version_compatibility_tests.py
```diff
@@ -13,7 +14,7 @@
 # don't use postgres
 os.environ["ROX_POSTGRES_DATASTORE"] = "false"
 
-versions=["3.71.0", "3.70.0", "3.69.0"]
+versions=getLastSensorVersionsFromGitTagsCLI(4)
```
fredrb, line 17: I think this function name is too verbose. I suggest renaming it to `get_lastest_main_git_tags`

 1. Tags are for the main image. Not specific to sensor
 2. I think `latest` is a better word here. Because they are the _newest_ released tags.
 3. If it comes from the CLI or not is a redundant information and it doesn't really fit in the function name IMO. 
 4. You're

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+
+"""
+Returns the latest patches of the last 4 major versions
```
msugakov, line 4: Either `def getLastSensorVersionsFromGitTagsCLI(num_versions):` should loose `num_versions` parameter or the description here should be modified.
Would be even better if you can restore the `main` function that simply calls and prints `getLastSensorVersionsFromGitTagsCLI` output. This way anyone will be able to run this script as a stand-alone executable and see how it works and what it produces.

scripts/ci/jobs/gke_version_compatibility_tests.py
msugakov: Another suggestion: `get_latest_release_tags` or `get_latest_release_versions`.

Also, please don't forget to rename `sensor_versions.py` file accordingly.

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+            break
+        if tags[1] < ycurr:
+            ycurr = tags[1]
+        if tags[1] == ycurr:
+            latestversions.append(str(tags[0]) + "." + str(tags[1]) + "." + str(tags[2]))
+            ycurr-=1
+    return latestversions
+
+def makeImageTag():
+    return subprocess.check_output(["make", "tag"]).decode(encoding="utf-8")
```
msugakov, line 51: Whenever you call `make something` to get its output, as a best practice please apply `--quiet` and `--no-print-directory`. See discussion here https://github.com/stackrox/stackrox/pull/3051#discussion_r968840394
```suggestion
    return subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")
```

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+Returns the latest patches of the last 4 major versions
+"""
+
+import sys
+import re
+import subprocess
+
+def isReleaseTagGitCLI(version):
```
msugakov, line 11: I think mentions of `GitCLI` or just `cli` in function names are just noise. I suggest to remove them unless they are necessary and bring some meaning.

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+    return bool(re.search(r"^\d+\.\d+\.\d+$", version))
```
msugakov, line 12: Not sure about `bool()` cast, a check for `None` seems more readable. No strong opinion.
```suggestion
    return re.search(r"^\d+\.\d+\.\d+$", version) is not None
```

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+
+def filterGitCLITags(rawtags):
+    filteredtags = []
+    for t in rawtags:
+        if isReleaseTagGitCLI(t):
+            filteredtags.append(t)
+    return set(filteredtags)
```
msugakov, line 19: [nerd mode: on] A more idiomatic Python would be
```suggestion
    return set([t for t in rawtags if isReleaseTagGitCLI(t)])
```

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+
+def cliOutputToTags(stdoutput):
+    separated = stdoutput.decode(encoding="utf-8").splitlines()
+    return separated
```
msugakov, line 23: There's not much that this function does on its own. I suggest to inline it.

.openshift-ci/sensor_versions.py
```diff
@@ -0,0 +1,57 @@
...
+
+def transformTagsToNumbers(tags):
+    numTags = []
+    for t in tags:
+        numTags.append(splitVersion(t))
+    return sorted(numTags)
```
msugakov, line 29: [nerd mode: on]
```suggestion
    return sorted([splitVersion(t) for t in tags])
```

msugakov (commented):
Did not check `getLatestTags`.

.openshift-ci/sensor_versions.py
jschnath: Good idea with the dicts. Just tested it out and it works, but only if the MAIN_IMAGE_TAG is the latest, eg. this will always grab the 4 latest major sensor versions but it should grab the version corresponding to central and the previous 3. I'll adapt it accordingly.

.openshift-ci/sensor_versions.py
jschnath: I think it makes a lot of sense to test it this way, eg. ensuring that it's the last 4 major versions from current central version down that we test. We may run into unintended behavior if lets say we test for central `3.68.x` but sensor in the tests is in version `3.72.x`. And while it's true that we do not support older releases, at the time that an older version was released the compatibility for versions that were within the compatibility window at that time should be expected.

It's questionable whether these tests should / would be run on older centrals at all, but if they are I prefer the current behavior.

.openshift-ci/sensor_versions.py
jschnath: And this clears up the comment above, in that case I'll rename `sensor_versions.py` to `get_latest_release_versions.py` too

.openshift-ci/sensor_versions.py
jschnath: Good idea, I was worrying about possible side effects from calling `make tag` and had asked about them before, this should help cover edge cases.

.openshift-ci/sensor_versions.py
jschnath: Okay I'll rename it. It's basically a legacy name now since I previously had 3 different methods of getting the tags but now there is only 1 so the distinction is unnecessary.

.openshift-ci/sensor_versions.py
jschnath: Very cool, I'll take it 👍

.openshift-ci/sensor_versions.py
jschnath: Fair

.openshift-ci/get_latest_release_versions.py
```diff
@@ -0,0 +1,46 @@
...
+
+# get_latest_release_versions gets the latest patches of the last num_versions major versions via Git CLI
+def get_latest_release_versions(num_versions):
+    rawtags = subprocess.check_output(["git", "tag", "--list"]).decode(encoding="utf-8").splitlines()
+    tags = filter_tags(rawtags)
+    return get_latest_tags(tags, num_versions)
+
+def main(argv):
+    latestversions = get_latest_release_versions(int(argv[1]))
+    print("\n".join(latestversions))
```
msugakov, line 43: As a suggestion to make it a bit more self-contained and not require command-line argument by default (warning: untested)

```suggestion
def main(argv):
    n = int(argv[1]) if len(argv)>1 else 4
    latestversions = get_latest_release_versions(4)
    print(f"Last {n} versions:")
    print("\n".join(latestversions))
```

jschnath: Done

.openshift-ci/get_latest_release_versions.py
```diff
@@ -0,0 +1,46 @@
...
+    return set([t for t in rawtags if is_release_tag(t)])
+
+def make_image_tag():
+    return subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")
+
+def extract_y_from_main_image_tag(mainimagetag):
+    return int(re.search(r"\d+\.(\d+)", mainimagetag).group(1))
+
+def get_latest_tags(tags, num_versions):
+    main_image_y = extract_y_from_main_image_tag(make_image_tag())
```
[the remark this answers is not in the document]
jschnath: Fair point, done.

.openshift-ci/get_latest_release_versions.py
```diff
@@ -18,16 +18,17 @@ def filter_tags(rawtags):
 def make_image_tag():
     return subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")
 
-def extract_y_from_main_image_tag(mainimagetag):
-    return int(re.search(r"\d+\.(\d+)", mainimagetag).group(1))
+def extract_x_y_from_main_image_tag(mainimagetag):
+    x_y = re.search(r"(\d+)\.(\d+)", mainimagetag)
+    return int(x_y.group(1)), int(x_y.group(2))
 
 def get_latest_tags(tags, num_versions):
```
msugakov, line 25: After looking at this, I think that the function can be split in two: the first one finds the latest patch for each `major.minor`. The second function will query `make tag` and take the last `N` versions just before the tag.

.openshift-ci/get_latest_release_versions.py
```diff
@@ -0,0 +1,49 @@
...
+    return get_latest_tags(tags, num_versions)
+
+def main(argv):
+    n = int(argv[1]) if len(argv)>1 else 4
+    latestversions = get_latest_release_versions(n)
+    print(f"Last {n} versions:")
+    print("\n".join(latestversions))
+
+if (__name__ == "__main__"):
+    main(sys.argv)
```
msugakov, line 49: Please don't forget to make this file executable. What I currently see is
image

msugakov (approved):
LGTM. Feel free to submit after addressing my final comments and making sure it's also clear from Fred's side.

.openshift-ci/get_latest_release_versions.py
jschnath: Okay, done. I didn't know you could call python scripts directly like that, I had been calling it using `python3 .openshift-ci/get_latest_release_versions.py`, I learned something new again.

.openshift-ci/get_latest_release_versions.py
```diff
@@ -18,16 +18,17 @@ def filter_tags(rawtags):
 def make_image_tag():
     return subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")
 
-def extract_y_from_main_image_tag(mainimagetag):
-    return int(re.search(r"\d+\.(\d+)", mainimagetag).group(1))
+def extract_x_y_from_main_image_tag(mainimagetag):
+    x_y = re.search(r"(\d+)\.(\d+)", mainimagetag)
```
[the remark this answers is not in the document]
jschnath: Fair

.openshift-ci/get_latest_release_versions.py
jschnath: Done

.openshift-ci/get_latest_release_versions.py
msugakov: That's a feature of *nix first of all. Whenever it sees that the file starts with `#!` (it's called shebang) it uses the following string as a command to run, and the rest of the file becomes input to that command.

jschnath:
The gke-version-compatibility-tests are passing and I've just double-checked the logs, it did automatically find the correct versions:
```
INFO: Tue Nov 15 10:08:02 UTC 2022: Starting test (sensor compatibility test 3.72.2)
INFO: Tue Nov 15 10:35:50 UTC 2022: Starting test (sensor compatibility test 3.71.2)
INFO: Tue Nov 15 11:02:11 UTC 2022: Starting test (sensor compatibility test 3.70.2)
INFO: Tue Nov 15 11:28:17 UTC 2022: Starting test (sensor compatibility test 3.69.2)
```

fredrb (approved):

scripts/ci/jobs/gke_version_compatibility_tests.py
```diff
@@ -4,6 +4,7 @@
 Run version compatibility tests
 """
 import os
+import get_latest_release_versions
```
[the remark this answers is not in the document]
jschnath: Done

.openshift-ci/get_latest_release_versions.py
```diff
@@ -0,0 +1,53 @@
...
+        [major, minor, patch] = t.split('.')
+        k = '.'.join([major, minor])
+        top_patch_version[k] = max(top_patch_version[k], int(patch))
+    top_major_versions = sorted(list(top_patch_version.keys()), reverse=True)
+    return [t + '.' + str(top_patch_version[t]) for t in top_major_versions]
+
+def make_image_tag():
+    return subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")
+
+def extract_x_y_from_main_image_tag(mainimagetag):
```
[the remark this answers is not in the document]
jschnath: Done

PR #3620, opened 2022-10-28
Scanner 2.26.2

RTann:
## Checklist

## Testing Performed

CI


RTann:
Why are tests skipped?

RTann:
> @RTann maybe the PR needs to be against master? Or if there is a closed PR to master, then it needs to be tagged to this milestone.
> 
> For the 3.72.1 release, this PR to master was cherry-picked by automation when the milestone was closed and the PR was found closed (cut-rc action: https://github.com/stackrox/stackrox/blob/master/.github/workflows/cut-rc.yml#L180): #3424

The version of Scanner on master is for the 73 release, and is not compatible with the 72 release. Updating the version on master will require us to revert another file, which will then all need to be reverted back once we cut this release. Making this PR based on the release branch allows us to avoid those complications.

There precedent for making PRs against the release branch directly, if it's necessary (for example, see https://github.com/stackrox/stackrox/pull/3596 which was merged and is also part of the same Milestone)

RTann:
> Thanks! I'm not sure why OSCI isn't running on this or that PR. I hadn't looked before for OSCI on PR's like this to a release branch. The config for this release is here: https://github.com/openshift/release/blob/e4978df7abc56e88a0f9280de4ef1a2a95d3e38b/ci-operator/config/stackrox/stackrox/stackrox-stackrox-release-3.72.yaml#L390 -> https://github.com/openshift/release/pull/32143/files#diff-cd82043cb1b2c7242c71e63307df514741a8d666f3587e1a722302706fc6c108R6-R7

I believe it's because `postsubmit: true` is set, so it'll only run upon merge. I don't think direct merges were expected

vikin91 (approved):

PR #3624, opened 2022-10-28
remove GetNodeRescanInterval(), as it is no longer needed

RTann:
## Description

Duration environment variables are now validated upon registration and retrieval. See https://github.com/stackrox/stackrox/blob/master/pkg/env/durationsetting.go#L46 for more information

## Checklist

## Testing Performed

CI


vikin91:
@RTann, we added this function, because providing 0 or negative rescan interval causes a panic. The user is an actor who  configures this, so the input validation is something that may make sense.

RTann:
> @RTann, we added this function, because providing 0 or negative rescan interval causes a panic. The user is an actor who configures this, so the input validation is something that may make sense.

I probably should have added this to the description, but this is already validated as-is. I'll do it now: https://github.com/stackrox/stackrox/blob/master/pkg/env/durationsetting.go#L46

vikin91:
Could you give it a quick test with 
`kubectl -n stackrox set env daemonsets/collector --containers="compliance" ROX_NODE_RESCAN_INTERVAL="0s"` 
or
`kubectl -n stackrox set env daemonsets/collector --containers="compliance" ROX_NODE_RESCAN_INTERVAL="-5s"`

vikin91:
> > @RTann, we added this function, because providing 0 or negative rescan interval causes a panic. The user is an actor who configures this, so the input validation is something that may make sense.
> 
> I probably should have added this to the description, but this is already validated as-is. I'll do it now: https://github.com/stackrox/stackrox/blob/master/pkg/env/durationsetting.go#L46

Oh, I see now, this looks good!

vikin91 (approved):
Cool, thanks for optimizing-out the redundant code :)

PR #3635, opened 2022-10-31
ROX-12380: Scope subjects to cluster role for Most Cluster Admin Roles widget

vjwilson:
## Description

Per Charmik in the Jira request:
> The query is used by "Users with most cluster admin roles" widget and queries all clusters and all subjects within each cluster. With postgres is enabled, the widget call a query with "Cluster Role:true" filter added to subjects because for all clusterAdmin subjects, their RoleBindings would have ClusterRole set to true. The additional filter can improve the turnaround time of the query in many cases.

## Checklist

## Testing Performed

Gives the same results before and after the optimization.

Screen Shot 2022-10-31 at 11 39 06 AM

### DIFF
--- a/ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
+++ b/ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
@@ -11,10 +11,10 @@ import Lollipop from 'Components/visuals/Lollipop';
 import ReactRouterPropTypes from 'react-router-prop-types';
 
 const QUERY = gql`
-    query usersWithClusterAdminRoles {
+    query usersWithClusterAdminRoles($query: String) {
         clusters {
             id
-            subjects {
+            subjects(query: $query) {
                 id
                 name
                 clusterAdmin
@@ -75,7 +75,12 @@ const UsersWithMostClusterAdminRoles = ({ match, location }) => {
     }
 
     return (
-        <Query query={QUERY}>
+        <Query
+            query={QUERY}
+            variables={{
+                query: 'Cluster Role:true',
+            }}
+        >
             {({ loading, data, networkStatus }) => {
                 let contents = <Loader />;
                 let viewAllLink;


vjwilson:
@charmik-redhat Does this change need to be scoped, so that it only happens in a Postgres environment, or can it exist in a non-Postgres environment, too?

bradr5 (approved):

charmik-redhat:
> 

@vjwilson I have not tested the extra filter with Postgres disabled. Tagging @md2119 to ask if boolean matching queries need any pre-processing in non-postures environments.

ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
```diff
@@ -11,10 +11,10 @@ import Lollipop from 'Components/visuals/Lollipop';
 import ReactRouterPropTypes from 'react-router-prop-types';
 
 const QUERY = gql`
-    query usersWithClusterAdminRoles {
+    query usersWithClusterAdminRoles($query: String) {
         clusters {
```
md2119, line 15: Why are we querying clusters?

ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
```diff
@@ -75,7 +75,12 @@ const UsersWithMostClusterAdminRoles = ({ match, location }) => {
     }
 
     return (
-        <Query query={QUERY}>
+        <Query
+            query={QUERY}
+            variables={{
+                query: 'Cluster Role:true',
```
md2119, line 81: This will not work yet. If you open the `Users and Groups` and focus on the search bar, you will notice that this is not in the search options for subjects. 
For this to work, the search options for the search category must be expanded to include `Cluster Role`.

md2119 (commented):
In addition to the code comments, note that filtering on `Cluster Role` only alleviates some pressure performance. The frontend check to determine cluster-admin role based on `clusterAdmin` should stay.

ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
charmik-redhat: I think this will work only for the 'Users with most cluster admin roles' widget as the query is passed as a string const and not using the result of search options to extract the `Cluster Role` field. But I agree, this would not work on `Users and Groups` page because that will show available filters based on the response of searchOptions. I will create  a separate PR to add the search option while this one can still be merged.

ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js
vjwilson: @md2119 
We loop through each cluster, and then the subjects within each cluster, to build a map of users with `clusterAdmin` permission, and how many.

The code for that is further down in this file,
https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/ConfigManagement/Dashboard/widgets/UsersWithMostClusterAdminRoles.js#L32-L55

This was written by Assaf over 3 years ago; if there is now a more efficient way to get those counts to display in the widget, let the UI know.

md2119: It seems not needed but let's leave the code as is because it could be brittle.

vjwilson:
@charmik-redhat @md2119 

Is this OK to merge yet, or not?

If not, do you want to add the BE changes to this PR, so that it an atomic change?

charmik-redhat (commented):
Since this change is relevant to widget's performance with Postgres, I think we can merge it after 3.73 to avoid any breaking change to 3.73.

charmik-redhat:
> @charmik-redhat @md2119
> 
> Is this OK to merge yet, or not?
> 
> If not, do you want to add the BE changes to this PR, so that it an atomic change?

@vjwilson The BE change is in PR [#3682](https://github.com/stackrox/stackrox/pull/3682) but I think we can merge that and then this PR after 3.73

md2119:
> @charmik-redhat @md2119
> 
> Is this OK to merge yet, or not?
> 
> If not, do you want to add the BE changes to this PR, so that it an atomic change?

Hold this PR until 3.73 is released and the backend PR is merged.

md2119:
@vjwilson might want to update the JIRA with `3.74` fix version.

vjwilson:
@charmik-redhat 
Now that release 3.73 has been cut, can I merge this FE change?

cc: @md2119

charmik-redhat:
> @charmik-redhat Now that release 3.73 has been cut, can I merge this FE change?
> 
> cc: @md2119

@vjwilson Yes, it should be good to merge now.

PR #3636, opened 2022-10-31
Implement read-only view of individual collection page

dvail:
## Description

This implements the read-only "view" mode of individual collection pages.

Form controls will be disabled and non-interactive, extraneous UI components (trash buttons) will not be visible, and the collection attacher component will be replaced with a read-only alternative display.

## Checklist

## Testing Performed

Visit the collection page and select a collection from the table. The browser will navigate to the read only view for that collection with the constraints in the above description. 
Clicking on disabled UI controls will have no effect. Tab-navigation will skip over the disabled controls.
image
image

When viewing a collection with no embedded collections in read-only mode, an empty state component will appear in place of the table:
image

Mobile layout views:
image
image
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
@@ -20,6 +20,9 @@ import {
     DropdownItem,
     DropdownSeparator,
     DropdownToggle,
+    EmptyState,
+    EmptyStateIcon,
+    EmptyStateVariant,
     Flex,
     FlexItem,
     Form,
@@ -28,8 +31,9 @@ import {
     Text,
     TextInput,
     Title,
+    Truncate,
 } from '@patternfly/react-core';
-import { CaretDownIcon } from '@patternfly/react-icons';
+import { CaretDownIcon, CubesIcon } from '@patternfly/react-icons';
 import { useFormik } from 'formik';
 import * as yup from 'yup';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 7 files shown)


dvail:
* master
    * **PR #3636** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3636" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3636?utm_source=stack-comment).

ui/apps/platform/src/css/trumps.css
```diff
@@ -80,6 +80,11 @@ button.pf-c-toggle-group__button:disabled {
     border-width: 1px !important;
 }
 
+/* Un-override the above override when the <Select> component is disabled */
+.pf-c-select__toggle.pf-m-disabled::before {
```
dvail, line 84: Before this change:
image

After this change:
image

Disabled `<Select>` styles on the PatternFly site:
image

dvail: Also note the subtle color difference between the top and bottom dropdowns. This is due to an `opacity: 0.5;` rule from the `ui-components` package that is applying to `<button>` elements. The top `<Select>` uses a button as the main element while the bottom uses a `<div>` due to the difference PF options in use.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -46,6 +51,34 @@ import CollectionAttacher from './CollectionAttacher';
...
+                ))}
+            </Tbody>
+        </TableComposable>
+    ) : (
+        <EmptyState variant={EmptyStateVariant.xs}>
+            <EmptyStateIcon icon={CubesIcon} />
+            <Title headingLevel="h4">
+                There are no other collections attached to this collection
+            </Title>
+        </EmptyState>
```
pedrottimark, line 78: This is consistent with many other empty state titles. However, a question for PatternFly Developer Office Hours is pro and con for accessibility of `h4` element versus `p` element with the identical style.

In a video for a past day of learning with theme of accessibility, the screen reader user used heading structure of page quite a bit to form a mental model of the information hierarchy (and commented on any skips in heading levels). This page has intuitive hierarchy for `h1` and `h2` but not so much for this heading.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -386,14 +425,29 @@ function CollectionForm({
...
-                                    </Title>
-                                    <p>Extend this collection by attaching other sets.</p>
-                                    <CollectionAttacher
-                                        initialEmbeddedCollections={initialEmbeddedCollections}
-                                        onSelectionChange={onEmbeddedCollectionsChange}
-                                    />
+                                    {isReadOnly ? (
+                                        <>
+                                            <Title className="pf-u-mb-xs" headingLevel="h2">
+                                                Attached collections
```
pedrottimark, line 431: Bravo! Does **Add new collection rules** deserve similar treatment?

pedrottimark: On the other side of the coin, consistent **Attached collections** heading is better for docs team seems clearer (to me, at least).

pedrottimark (commented):
How did we ever survive without conditional rendering via React?

2 comments for your consideration.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
dvail: True I agree, especially with the screen reader comment since this bit of text doesn't really communicate hierarchy. The page has `h1` and `h2`, but there isn't anything semantically that makes this a `h3`, `h4`, etc. I'll update to a `<p>` with styles.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
dvail: Good point. How about "Attached collections" in both modes, but conditional rendering of "Extend this collection by attaching other sets." which only applies in the edit modes? I think that would be a good call to action in combination with the "Attach" and "Detach" buttons below.

I'll take a look at the collection rules text too.

pedrottimark (approved):

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
pedrottimark: Yes, indeed.

PR #3637, opened 2022-10-31
Add collection request generating function

dvail:
## Description

This PR adds a function to convert the FE `Collection` object to the expected BE request format `CollectionRequest`. This also cleans up some cruft in the type system and naming that surfaced during the conversion.

- [Type predicate functions were replaced with a discriminator field](https://github.com/stackrox/stackrox/pull/3637/files#diff-2902d0161d4d4122f4b8cee5ed1fbb11d17e575d8d1644a4cd2d0e3afd9b32d5R58) when dealing with the three resource selector types. (All resources, By name, By label). Among other improvements, this allows us to [exhaustively check each condition](https://github.com/stackrox/stackrox/pull/3637/files#diff-6171f3d9ba3e6fc4bd227b6a964231394f0f5de37019c0c413956e87ee88d547R146) to help ensure that when new selector types (By annotation) are added that cases are not missed.
- The [front end `Collection` type](https://github.com/stackrox/stackrox/pull/3637/files#diff-2902d0161d4d4122f4b8cee5ed1fbb11d17e575d8d1644a4cd2d0e3afd9b32d5R81) renames `resourceSelectors` to `resourceSelector`, since there is one a single one on the FE type, and renames `embeddedCollections` to `embeddedCollectionIds` since we are not dealing with full object instances.

## Checklist

## Testing Performed

Unit tests added for new `generateResponse` function and existing `parseCollection` function.


dvail:
* master
      * **PR #3637** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3637" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3637?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/converter.ts
```diff
@@ -0,0 +1,177 @@
...
+            Namespace: { type: 'All' },
+            Cluster: { type: 'All' },
+        },
+    };
+
+    const errors: string[] = [];
+
+    if (data.resourceSelectors.length > 1) {
+        errors.push(
+            `Multiple 'ResourceSelectors' were found for this collection. Only a single resource selector is supported in the UI. Further validation errors will only apply to the first resource selector in the response.`
```
pedrottimark, line 49: Can you ask docs team to suggest a synonym for **in the UI**

ui/apps/platform/src/Containers/Collections/converter.ts
```diff
@@ -0,0 +1,177 @@
...
+                break;
+            default:
+                ensureExhaustive(selector);
+        }
+    });
+
+    return {
+        name: collection.name,
+        description: collection.description,
+        embeddedCollectionIds: collection.embeddedCollectionIds,
```
pedrottimark, line 174: Can you double check it the key is `embeddedCollectionIds` or `embeddedCollections`

https://github.com/stackrox/stackrox/blob/master/proto/storage/resource_collection.proto#L28

pedrottimark (approved):
Thank you for writing unit tests! A few comments.

Two observations here because independent of these changes:
1. Do you want to include `useRestQuery` as discussion item for next team meeting with goal to move from src/Containers/Dashboard/hooks to src/hooks folder?
2. What do you think are pro and con to distinguish which types correspond to payloads and responses versus which types are for rendering in form? That is, alternatives:
    * CollectionsService.ts import `SelectorField` from container types file?
    * first 8 lines from container types file move to CollectionsService.ts file?

ui/apps/platform/src/Containers/Collections/converter.ts
dvail: I think the internal proto is different from what they will be sending/accepting at the service layer. I was basing the naming on the original BE design doc as well as following along with some of the PRs that get announced on the epic channel. It looks like for this type of request [they are going](https://github.com/stackrox/stackrox/pull/3600/files#diff-2abd114e43fc8626980140aecb4464f03014c29f5ca1fdf3c4ef88a7196caa7dR30) with the `ids` version.

dvail:
>     1. Do you want to include `useRestQuery` as discussion item for next team meeting with goal to move from src/Containers/Dashboard/hooks to src/hooks folder?
> 
>     2. What do you think are pro and con to distinguish which types correspond to payloads and responses versus which types are for rendering in form? That is, alternatives:
>        
>        * CollectionsService.ts import `SelectorField` from container types file?
>        * first 8 lines from container types file move to CollectionsService.ts file?

1. Agreed, that is on my personal clean up list to move that hook up. Better yet might be to reevaluate React-Query as the standard way to handle these. I know Saif used it for the ACSCS work so maybe I can get him to throw his opinion into the mix.
2. This I'm not too sure on. There are some types that are clearly request/response oriented, and some that are clearly "in-container" oriented, but I'm not really sure where the best place to put the common bits is. Moving them into a top level `src/types/collections` directory seems neutral, but then splits types into three unrelated locations. I'd almost be inclined to weaken the types in the service file `e.g.   SelectorField -> string` since we aren't doing runtime type validation anyway and the service layer types are more of a scout's honor that they are what they say they are. We have to run type guards on the response data in the parser anyway, so weakening the types might let us keep these decoupled without any loss of safety.

pedrottimark:
Dave, what you suggest sounds good to me:
* Edit CollectionsService.ts: `fieldName: string` consistent with https://github.com/stackrox/stackrox/blob/master/proto/storage/resource_collection.proto#L37-L46
* Leave Collections/types.ts as source of truth for container

PR #3638, opened 2022-10-31
ROX-13377: Network Graph selection for namespace doesn't work

sachaudh:
## Description

Essentially, the `findEntityById` function wasn't working quite right. The model didn't have a `groups` field. We only have `graph`, `nodes,` and `edges`. I also pulled out the URL-related logic up to the parent component since the NetworkGraph component doesn't need to know all that.

## Checklist

## Testing Performed

Screen Shot 2022-10-31 at 11 26 05 AM
Screen Shot 2022-10-31 at 11 26 09 AM
Screen Shot 2022-10-31 at 11 26 12 AM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 import React from 'react';
-import { useHistory } from 'react-router-dom';
 import {
     Model,
     SELECTION_EVENT,
@@ -13,9 +12,9 @@ import {
     Visualization,
     VisualizationProvider,
     VisualizationSurface,
+    NodeModel,
 } from '@patternfly/react-topology';
 
-import { networkBasePathPF } from 'routePaths';
 import stylesComponentFactory from './components/stylesComponentFactory';
 import defaultLayoutFactory from './layouts/defaultLayoutFactory';
 import defaultComponentFactory from './components/defaultComponentFactory';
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -163,32 +162,8 @@ const model: Model = {
...
-    }
-    let entity;
-    entity = graphModel.nodes?.find((node: { id: string }) => node.id === id);
-    if (!entity) {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        entity = graphModel.groups?.find((group: { id: string }) => group.id === id);
-    }
+function findEntityById(graphModel: Model, id: string): NodeModel | undefined {
+    const entity = graphModel.nodes?.find((node: { id: string }) => node.id === id);
```
sachaudh, line 166: We're only dealing with nodes, so we can just do the following

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -205,13 +180,14 @@ function getUrlParamsForEntity(selectEntity: {
...
+    closeSidebar: () => void;
+    onSelectNode: (type, id) => void;
 };
 
 export type TopologyComponentProps = NetworkGraphProps;
 
-const TopologyComponent = ({ detailType, detailId }: TopologyComponentProps) => {
-    const selectedEntity = detailId && findEntityById(model, detailId, detailType);
-    const history = useHistory();
+const TopologyComponent = ({ detailId, closeSidebar, onSelectNode }: TopologyComponentProps) => {
```
sachaudh, line 189: @alwayshooin @vjwilson  I'm actually not too fond of the prefix `detail`. I think that it's a bit confusing. What do you all think about `nodeType` and `nodeId` instead? Or if we eventually do want to make edges selectable, then we can do `objectType` and `objectId` or `elementType` and `elementId`. Open to other suggestions too

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -224,13 +200,7 @@ const TopologyComponent = ({ detailType, detailId }: TopologyComponentProps) =>
...
             const [newDetailType, newDetailId] = getUrlParamsForEntity(newSelectedEntity);
 
-            // if found, and it's not the logical grouping of all external sources, then trigger URL update
-            if (newSelectedEntity && newSelectedEntity?.data?.entityType !== 'EXTERNAL') {
-                history.push(`${networkBasePathPF}/${newDetailType}/${newDetailId}`);
-            } else {
-                // otherwise, return to the graph-only state
-                history.push(`${networkBasePathPF}`);
-            }
+            onSelectNode(newDetailType, newDetailId);
```
sachaudh, line 203: I was thinking we don't need to do the URL manipulation on this level. We could have the NetworkGraphPage get/set the URL data and NetworkGraph can get/set the topology model data

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
alwayshooin: i think `nodeType` and `nodeId` works! btw, you can select a node from the graph via the controller by saying `controller.getNodeById`. not sure if that simplifies things or if you've explored that but just wanted to let u know in case u werent aware

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
vjwilson: We're not setting the URL to _reflect_ the data that NetworkGraph component decides to display separately. The data should only change _after_ we update the URL. (Think of the URL as Redux, and a URL update as an "action".)

With that said, I think this change would be necessary once we add the filtering, so we should go ahead and move the URL handling up.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
vjwilson: Either I imagined it, or the demo app had a separate "groups" list. 😕

vjwilson (approved):
Let's leave the "detail" prefix for now, because it reflect the result we want, instead of an internal implementation detail.

There still seems to be a flash (rerender) when clicking on the "external" group of CIDR blocks/external entities. We may need to re-introduce the "type of node" check to short-circuit that at some point, but this is good to get us un-blocked.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: @alwayshooin oh wow I didn't know that. Maybe we can try to use that function instead of making a custom one

sachaudh: @vjwilson I think `node` or `object` is pretty much what we've called the selectable parts of our network graph regardless of whatever library we've used, so I don't think the terminology is specific to this library. I think in any graph, it's pretty common to have `nodes` and `edges`. If we ever decide to ditch the graph aspect of this view and go with something else, sure, then maybe saying `nodeType` and `nodeId` might not work anymore. I think for the foreseeable future it does make sense, but if it's a big issue we can also just do `type` and `id`

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: Oh, yea you're right. I didn't go all the way. Let me make a few extra changes then

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
alwayshooin: groups are technically nodes (appended to the end of the nodes array), so that's where you might've gotten confused @vjwilson

PR #3641, opened 2022-11-01
ROX-13380: Conditional rendering edges for deployments and namespaces

alwayshooin:
as the title states -- the default graph should render no edges 

image

clicking on a deployment should render only edges related to said deployment

image

closing the side panel (unselecting the deployment) should clear the edges

image

clicking on a namespace should render only edges related to the child deployments within said namespace

image

also -- the graph should not flicker/rerender on clicking on nodes now :)

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 import React from 'react';
+import { useHistory, useParams } from 'react-router-dom';
 import {
     Model,
     SELECTION_EVENT,
@@ -10,11 +11,12 @@ import {
     TopologyControlBar,
     useVisualizationController,
     Visualization,
-    VisualizationProvider,
     VisualizationSurface,
+    VisualizationProvider,
     NodeModel,
 } from '@patternfly/react-topology';
 
+import { networkBasePathPF } from 'routePaths';
 import stylesComponentFactory from './components/stylesComponentFactory';
 import defaultLayoutFactory from './layouts/defaultLayoutFactory';
 import defaultComponentFactory from './components/defaultComponentFactory';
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -77,19 +77,49 @@ function getUrlParamsForEntity(selectEntity: {
...
+}
+
+function setVisibleEdges(edges) {
+    edges.forEach((edge) => {
+        edge.setVisible(true);
+    });
+}
+
+const TopologyComponent = ({ model }: TopologyComponentProps) => {
+    const { detailType, detailId } = useParams();
```
sachaudh, line 98: Wouldn't the fix still work if `useParams` was in the parent component? I think it makes sense to keep `useParams` in the parent component since we don't want this component to start doing too many things. Maybe you can try testing that theory out and seeing if it anything happens

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -77,19 +77,49 @@ function getUrlParamsForEntity(selectEntity: {
...
-    detailType?: UrlDetailTypeValue;
-    detailId?: string;
     model: Model;
 };
 
 export type TopologyComponentProps = NetworkGraphProps;
 
-const TopologyComponent = ({ detailType, detailId, model }: TopologyComponentProps) => {
+function getNodeEdges(selectedNode) {
+    const egressEdges = selectedNode.getSourceEdges();
```
sachaudh, line 86: The UI seems to break and it's being caused by this `getSourceEdges`. What is that function?

Screen Shot 2022-11-01 at 8 36 37 AM

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
alwayshooin: okay turns out we do need to keep it at this level to prevent rerenders of the parent component on each node select/unselect. it was still flickering when i moved it out and post rebase from moving `history` out as well. when i moved them both back in as it was in this PR initially, the flickering went away

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -77,19 +77,49 @@ function getUrlParamsForEntity(selectEntity: {
...
+
+const TopologyComponent = ({ model }: TopologyComponentProps) => {
+    const { detailType, detailId } = useParams();
     const selectedEntity = detailId && findEntityById(model, detailId, detailType);
     const history = useHistory();
-
     const controller = useVisualizationController();
 
+    if (detailId) {
+        const selectedNode = controller.getNodeById(detailId);
```
vjwilson, line 104: I notice that if I click a node, and then while that first note is still selected, I click on another node, the URL updates, and the second nodes shows edges, but the first node's edges also still show. 

Nuanced issue, because we don't want to complete deselect the first node, and cause the sidepanel to close--just to re-open immediately--but we still need to `setVisible(false)` the previous edges.

(I wonder if this would help the memory-leak issue.)

alwayshooin:
i'll address the performance issues in a separate ticket and merge this one in while we're waiting for the PF team to get back to us on the conditional rendering issue.

vjwilson (approved):

alwayshooin:
merging bc we have no e2e tests that deal with the new PF network graph

PR #3646, opened 2022-11-01
ROX-11829: Diagnostic bundles generated contain info about central's environment if flag is set

theencee:
## Description

Exclude any information that is specific to central and central's environment from diagnostic bundle and debug dump. Of course that wouldn't enough if central was part of a secured cluster, but in managed services we know that won't be the case. See the [ticket](https://issues.redhat.com/browse/ROX-11829) for way more info.

Additionally conditionally remove the RBAC required to get diagnostic info from central's service account if it's a managed service central.

Lastly, the setting is controlled via an _undocumented_ independent environment variable that anyone can set and get the same behavior (with the caveat of the secured cluster issue as mentioned previously).

I decided to use a preexisting setting in the helm chart, `._rox.env.managedServices` to force the environment variable to be set. This avoids having to coordinate with fleet manager. Still unsure about if that's the right approach.

## Checklist

## Testing Performed

Manually tested

### DIFF
--- a/central/debug/service/diagnostics.go
+++ b/central/debug/service/diagnostics.go
@@ -276,8 +276,10 @@ func (s *serviceImpl) getK8sDiagnostics(ctx context.Context, zipWriter *zip.Writ
 	// Pull data from the central cluster.
 	// TODO: It would be nice if we could add a flag to a sensor connection that indicates whether this sensor is
 	// running colocated with central, so we could skip this.
-	wg.Add(1)
-	go pullCentralClusterDiagnostics(subCtx, filesC, &wg, opts.since)
+	if opts.withCentral {
+		wg.Add(1)
+		go pullCentralClusterDiagnostics(subCtx, filesC, &wg, opts.since)
+	}
 
 	go func() {
 		select {
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -32,6 +32,7 @@ import (
 	"github.com/stackrox/rox/pkg/auth/authproviders"
 	"github.com/stackrox/rox/pkg/auth/permissions"
 	"github.com/stackrox/rox/pkg/buildinfo"
+	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/errox"
 	grpcPkg "github.com/stackrox/rox/pkg/grpc"
 	"github.com/stackrox/rox/pkg/grpc/authz"
@@ -484,6 +485,7 @@ type debugDumpOptions struct {
 	withLogImbue      bool
 	withAccessControl bool
 	withNotifiers     bool
+	withCentral       bool
 	clusters          []string
 	since             time.Time
 }
@@ -499,33 +501,35 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 		return
 	}
 
-	if err := zipPrometheusMetrics(zipWriter, "metrics-1"); err != nil {
-		log.Error(err)
-	}
-
-	if err := getMemory(zipWriter); err != nil {
-		log.Error(err)
-	}
-
-	if err := getGoroutines(zipWriter); err != nil {
-		log.Error(err)
-	}
-
-	if err := getBlock(zipWriter); err != nil {
-		log.Error(err)
-	}
-
-	if err := getMutex(zipWriter); err != nil {
-		log.Error(err)
-	}
-
-	if opts.withCPUProfile {
-		if err := getCPU(ctx, zipWriter, cpuProfileDuration); err != nil {
+	if opts.withCentral {
+		if err := zipPrometheusMetrics(zipWriter, "metrics-1"); err != nil {
 			log.Error(err)
 		}
 
-		if err := zipPrometheusMetrics(zipWriter, "metrics-2"); err != nil {
+		if err := getMemory(zipWriter); err != nil {
 			log.Error(err)
+		}
+
+		if err := getGoroutines(zipWriter); err != nil {
+			log.Error(err)
+		}
+
+		if err := getBlock(zipWriter); err != nil {
+			log.Error(err)
+		}
+
+		if err := getMutex(zipWriter); err != nil {
+			log.Error(err)
+		}
+
+		if opts.withCPUProfile {
+			if err := getCPU(ctx, zipWriter, cpuProfileDuration); err != nil {
+				log.Error(err)
+			}
+
+			if err := zipPrometheusMetrics(zipWriter, "metrics-2"); err != nil {
+				log.Error(err)
+			}
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


theencee:
postgres upgrade test failure is unrelated

central/debug/service/service.go
```diff
@@ -540,7 +544,7 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 	}
 
 	if s.telemetryGatherer != nil && opts.telemetryMode > 0 {
-		telemetryData := s.telemetryGatherer.Gather(ctx, opts.telemetryMode >= 2)
+		telemetryData := s.telemetryGatherer.Gather(ctx, opts.telemetryMode >= 2, opts.withCentral)
```
misberner, line 547: I cannot comment on there, but you also need to disable ll.566-576 (new) when `opts.withCentral` is false.

central/debug/service/service.go
```diff
@@ -625,6 +630,11 @@ func (s *serviceImpl) getDebugDump(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if env.DisableCentralDiagnostics.BooleanSetting() {
+		// Regardless of query, don't generate logs for managed central
+		opts.logs = noLogs
```
misberner, line 635: You would also need to set `opts.withLogImbue` to `false` per what is said in the ticket, but see my comments above - I think it would be better to push that check down one level to where we actually attempt to collect logs.

central/debug/service/service.go
```diff
@@ -651,6 +661,7 @@ func (s *serviceImpl) getDiagnosticDump(w http.ResponseWriter, r *http.Request)
 		withCPUProfile:    false,
 		withLogImbue:      true,
 		withAccessControl: true,
+		withCentral:       !env.DisableCentralDiagnostics.BooleanSetting(),
```
misberner, line 664: No strong feelings either way, but IMHO, this is not a per-request property but rather a question of what mode the debug service runs in. And therefore I think this flag should be a field of `serviceImpl` and be passed in the `service.New(...)` function at its invocation in `main()`.

misberner (changes requested):

central/debug/service/service.go
theencee: I thought about it originally but I moved it to the debugDumpOptions since this service covers a lot more than just debug dump/diag bundle and this option doesn't apply to all. Just to these two. But also no strong feelings myself either, so I can switch it if you have a suggestion for a good property name?

rukletsov (approved):

central/debug/service/service.go
misberner: Naming is always hard, I would've suggested something like `mayAccessCentralEnv` or so, but maybe it's the simplest to just keep it.

misberner (approved):

pkg/env/central_diagnostics.go
```diff
@@ -0,0 +1,6 @@
+package env
+
+var (
+	// DisableCentralDiagnostics is set to true to signal that diagnostic bundles or dumps should not contain any information about central or the environment that central runs in.
+	DisableCentralDiagnostics = RegisterBooleanSetting("ROX_DISABLE_CENTRAL_DIAGNOSTICS", false)
```
SimonBaeumer, line 5: Can we avoid the double negation here, instead using `ROX_ENABLE_CENTRAL_DIAGNOSTICS`?

central/debug/service/diagnostics.go
```diff
@@ -276,8 +276,10 @@ func (s *serviceImpl) getK8sDiagnostics(ctx context.Context, zipWriter *zip.Writ
 	// Pull data from the central cluster.
 	// TODO: It would be nice if we could add a flag to a sensor connection that indicates whether this sensor is
 	// running colocated with central, so we could skip this.
-	wg.Add(1)
-	go pullCentralClusterDiagnostics(subCtx, filesC, &wg, opts.since)
+	if opts.withCentral {
+		wg.Add(1)
+		go pullCentralClusterDiagnostics(subCtx, filesC, &wg, opts.since)
```
SimonBaeumer, line 281: For my own understanding: does this function collect kubernetes manifests for the diagnostic bundle? 
The code in L281 fetches Central resources, the above fetches secured cluster k8s resources from all connected sensors?

theencee: > does this function collect kubernetes manifests for the diagnostic bundle?
Yes for stackrox namespace. Manifests, logs and events. Also the root ca configmap.

> The code in L281 fetches Central resources, the above fetches secured cluster k8s resources from all connected sensors?

Correct. This is done in addition to the secured clusters in case central is deployed without sensor.

That's why the caveat exists that if central has sensor this itself isn't enough to hide all central data. For ACSMS, since central doesn't have sensor, it works out.

pkg/env/central_diagnostics.go
theencee: I decided against it because by default diagnostics always includes central env. This is turning it off, because that's non-default. Also I copied from `ROX_DISABLE_AUTOGENERATED_REGISTRIES` :p 

I don't feel too strongly though so I can change it if you do feel strongly on it.

dhaus67 (approved):

pkg/env/central_diagnostics.go
SimonBaeumer: > Also I copied from ROX_DISABLE_AUTOGENERATED_REGISTRIES :p

I would not speak myself free that I never introduced one 😆 

> I don't feel too strongly though so I can change it if you do feel strongly on it.

Please, I would appreciate that we try to avoid the negative inverted settings.

central/debug/service/diagnostics.go
SimonBaeumer: Thank you @theencee!

pkg/env/central_diagnostics.go
SimonBaeumer: @rukletsov My requested change would imply to change the default to `true` to not introduce breaking changes or limit functions in existing environments. Does this change your vote? 🤔

md2119 (approved):

pkg/env/central_diagnostics.go
SimonBaeumer: @rukletsov My requested change would imply to change the default to `true` to not introduce breaking changes or limit functions in existing environments. Does this change your vote? 🤔

theencee: Yea if the default is true that it's backwards compatible as a missing env var means it's included by default. The only change would be to set it to _false_ for managed services specifically

charmik-redhat (approved):

pkg/env/central_diagnostics.go
rukletsov: As long as the default is "enable", I'm good : ).

PR #3647, opened 2022-11-01
Delete exception test from general integration tests

pedrottimark:
## Description

WIP

## Checklist
- [x] Deleted integration test

## Testing Performed

TODO


dvail (approved):
Although super high level, I still think there is value in testing that uncaught exceptions are recoverable. I wonder if there is some way we could reproduce this test in a follow up, but in a way that doesn't contaminate the logimbue-data.json file. Maybe we could piggyback off the ResizeObserver exception, or add another error to the Raven ignore list?

Just thinking of the very unlikely but non-ideal case where a regression makes it so that an exception renders the app unusable without a hard refresh.

pedrottimark:
Super suggestion.

PR #3648, opened 2022-11-01
[ROX-13030] : Add delete collection API endpoint and service implementation

charmik-redhat:
## Checklist

## Testing Performed

Added unit tests


central/resourcecollection/service/service_impl.go
```diff
@@ -95,6 +95,20 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
 	}, nil
 }
 
+// DeleteCollection deletes the collection with the given ID
+func (s *serviceImpl) DeleteCollection(ctx context.Context, request *v1.ResourceByID) (*v1.Empty, error) {
+	if !features.ObjectCollections.Enabled() {
+		return nil, nil
```
md2119, line 101: return error

central/resourcecollection/service/service_impl.go
```diff
@@ -95,6 +95,20 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
...
+	}
+	if request.GetId() == "" {
+		return nil, errors.Wrap(errox.InvalidArgs, "A collection id must be specified to delete a Collection")
```
md2119, line 104: nit: return message

central/resourcecollection/service/service_impl.go
```diff
@@ -141,7 +155,11 @@ func collectionRequestToCollection(ctx context.Context, request collectionReques
 	if isCreate {
 		collection.CreatedBy = slimUser
 		collection.CreatedAt = timeNow
-	} else if getID != nil {
+	} else {
+		// is update
+		if getID == nil || getID() == "" {
+			return nil, errors.New("[Update collection] Valid ID getter func not found")
+		}
```
md2119, line 162: I don't see a reason to for this. If it is an update, collection.Id must be set. Recommend focusing only on create the path and removing the else block entirely because this function does not address reconciliation with a stored collection for the update path.

central/resourcecollection/service/service_impl.go
```diff
@@ -141,7 +155,11 @@ func collectionRequestToCollection(ctx context.Context, request collectionReques
...
 		collection.Id = getID()
```
md2119, line 163: remove this with above suggestion.

md2119 (changes requested):

central/resourcecollection/service/service_impl.go
charmik-redhat: Removed this block and the getID func as well. Update workflow would need to make sure that the id is being set.

keyallis (approved):
LGTM

central/resourcecollection/service/service_impl.go
```diff
@@ -101,13 +101,27 @@ func (s *serviceImpl) getCollection(ctx context.Context, id string) (*v1.GetColl
 	}, nil
 }
 
+// DeleteCollection deletes the collection with the given ID
+func (s *serviceImpl) DeleteCollection(ctx context.Context, request *v1.ResourceByID) (*v1.Empty, error) {
+	if !features.ObjectCollections.Enabled() {
+		return nil, errors.New("Support for resource collections is not enabled")
+	}
+	if request.GetId() == "" {
+		return nil, errors.Wrap(errox.InvalidArgs, "A non empty collection id must be specified to delete a collection")
```
md2119, line 110: nit:  `A non empty` to `non-empty`.

md2119 (approved):

PR #3649, opened 2022-11-01
ROX-13169: Adding flows table

sachaudh:
## Description

This PR adds the flows table with a basic expandable table. This is an iterative addition. You can find all the subtasks here: https://issues.redhat.com/browse/ROX-12904

## Checklist

## Testing Performed

Screen Shot 2022-11-01 at 2 09 02 PM
Screen Shot 2022-11-01 at 2 11 05 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentDetails.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentDetails.tsx
@@ -17,7 +17,7 @@ import {
     TextContent,
     TextVariants,
 } from '@patternfly/react-core';
-import { InfoCircleIcon } from '@patternfly/react-icons';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
 
 function DetailSection({ title, children }) {
     const [isExpanded, setIsExpanded] = useState(true);
@@ -61,7 +61,7 @@ function DeploymentDetails() {
                                             <Label
                                                 variant="outline"
                                                 color="red"
-                                                icon={<InfoCircleIcon />}
+                                                icon={<ExclamationCircleIcon />}
                                             >
                                                 2 external flows
                                             </Label>
@@ -70,7 +70,7 @@ function DeploymentDetails() {
                                             <Label
                                                 variant="outline"
                                                 color="gold"
-                                                icon={<InfoCircleIcon />}
+                                                icon={<ExclamationCircleIcon />}
                                             >
                                                 3 internal flows
                                             </Label>
@@ -89,7 +89,7 @@ function DeploymentDetails() {
                                             <Label
                                                 variant="outline"
                                                 color="gold"
-                                                icon={<InfoCircleIcon />}
+                                                icon={<ExclamationCircleIcon />}
                                             >
                                                 0 egress, allowing 325 flows
                                             </Label>
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
@@ -1,0 +1,3 @@
+.pf-c-table__toggle {
+    padding: 0 !important
+}
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
@@ -1,0 +1,239 @@
+import React from 'react';
+import {
+    Flex,
+    FlexItem,
+    Stack,
+    StackItem,
+    Text,
+    TextContent,
+    TextVariants,
+    Toolbar,
+    ToolbarContent,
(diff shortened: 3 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentDetails.tsx
```diff
@@ -17,7 +17,7 @@ import {
     TextContent,
     TextVariants,
 } from '@patternfly/react-core';
-import { InfoCircleIcon } from '@patternfly/react-icons';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
```
sachaudh, line 20: Oops I used the wrong icon in the Details page

ui/apps/platform/src/Containers/NetworkGraph/deployment/deploymentFlows.css
```diff
@@ -0,0 +1,3 @@
+.pf-c-table__toggle {
```
sachaudh, line 1: I only needed to do this tweak so the toggle doesn't take up so much space with padding

sachaudh: Open to suggestions on how to organize CSS files. I put it on the same level as the component that uses it and same naming structure

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
```diff
@@ -0,0 +1,3 @@
+.pf-c-table__toggle {
+    padding: 0 !important
+}
```
alwayshooin, line 3: oh is there extra padding otherwise?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
```diff
@@ -0,0 +1,252 @@
...
+    namespace: string;
+    direction: string;
+    port: string;
+    protocol: string;
+    isAnomalous: boolean;
+}
+
+interface Flow extends FlowBase {
+    children: FlowBase[];
+}
```
alwayshooin, line 39: i wonder if it would be useful to have these in a separate `types.ts` file in this directory for easy access

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
```diff
@@ -0,0 +1,252 @@
...
+                                                            </TextContent>
+                                                        </div>
+                                                    )}
+                                                    {row.type === 'External' && (
+                                                        <div>
+                                                            <TextContent>
+                                                                <Text
+                                                                    component={TextVariants.small}
+                                                                >
+                                                                    {`${row.children.length} active flows`}
```
alwayshooin, line 194: maybe you can do the conditional texts here instead of around the `<div><TextContent><Text component={TextVariants.small}>`

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
```diff
@@ -0,0 +1,252 @@
...
+                    <TableComposable aria-label="Deployment flow" variant="compact">
+                        <Thead>
+                            <Tr>
+                                <Th width={10} />
+                                <Th width={40}>{columnNames.entity}</Th>
+                                <Th width={20}>{columnNames.direction}</Th>
+                                <Th width={30}>{columnNames.portAndProtocol}</Th>
+                            </Tr>
+                        </Thead>
+                        {flows.map((row, rowIndex) => {
```
alwayshooin, line 155: what happens when there is an empty state? is an empty state possible here?

alwayshooin (approved):
looks good overall! just a few questions/comments in the code

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
sachaudh: Yea this is the difference

Screenshot 2022-11-04 at 9 19 13 AM
Screenshot 2022-11-04 at 9 19 21 AM

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
sachaudh: I wouldn't expect this type to be used between many components, but it might be useful when we get to the baselines tab. For now, I'll keep it here, and if required we can put it in its own type file.

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
sachaudh: I guess it'll just be empty. I'll have to add an empty state. @alanonthegit what should the empty state for the flows and baselines tab look like?

PR #3495, opened 2022-10-19
ROX-12989: Refactor listeners output interface

fredrb:
## Description

Refactor handlers to use a queue to publish events and detector requests. 

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #3653, opened 2022-11-02
ROX-12503 implement sensor's pipeline component

lvalerom:
## Description

Introducing a new sensor component that will wrap the k8s events management. This PR tackles all the refactoring on the handlers to use a queue to publish events and detector requests.

## Checklist

## Testing Performed

* CI
* Manually executing `sensor-integration-test` and `go-unit-tests`
* comparing benchmarks ([see](https://github.com/stackrox/stackrox/pull/3653#issuecomment-1300680692))
* long-running cluster with the following workload:
```yaml
deploymentWorkload:
- deploymentType: Deployment
  lifecycleDuration: 30m0s
  numDeployments: 1000
  numLifecycles: 0
  numLabels: 10
  randomLabels: true
  podWorkload:
    containerWorkload:
      numImages: 0
    lifecycleDuration: 10m0s
    numContainers: 3
    numPods: 5
    processWorkload:
      alertRate: 0.001
      processInterval: 1m0s
  updateInterval: 10m0s
networkPolicyWorkload:
- lifecycleDuration: 30m0s
  numNetworkPolicies: 500
  numLifecycles: 0
  updateInterval: 10m0s
networkWorkload:
  batchSize: 100
  flowInterval: 1s
nodeWorkload:
  numNodes: 1000
rbacWorkload:
  numBindings: 1000
  numRoles: 1000
  numServiceAccounts: 1000
serviceWorkload:
  numClusterIPs: 334
  numNodePorts: 333
  numLoadBalancers: 333
matchLabels: true
numNamespaces: 10
```


fredrb:
For the record if anyone finds this PR in the future. This was first implemented in pair in this PR: #3495. All initial review comments are there.

fredrb:
I ran the code in this branch against the benchmark in #3656 and compared the results with the benchmark shown in the PR:

```
% benchstat ./old.result.txt new-pipeline.result.txt
name          old time/op  new time/op  delta
_Pipeline-12   297ms ± 7%   267ms ± 8%  -9.88%  (p=0.032 n=5+5)
```

The pipeline here seems to be 9.8% faster. I think it's negligible since we're only considering deployments. But at least we know it's not significantly worse.

lvalerom:
> I ran the code in this branch against the benchmark in #3656 and compared the results shown in the description:
> 
> ```
> % benchstat ./old.result.txt new-pipeline.result.txt
> name          old time/op  new time/op  delta
> _Pipeline-12   297ms ± 7%   267ms ± 8%  -9.88%  (p=0.032 n=5+5)
> ```
> 
> The pipeline here seems to be 9.8% faster. I think it's negligible since we're only considering deployments. But at least we know it's not significantly worse.

That's great! Deployments are the resources that should take more time/resources to process so I think still good seeing that it's slightly better 🎉

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -0,0 +1,52 @@
...
+	// This property should be removed in the future and only the
+	// DetectionObject should be sent
+	CompatibilityDetectionDeployment []CompatibilityDetectionMessage
+	// CompatibilityReprocessDeployments is also used for compatibility reasons with Network Policy handlers
+	// in the future this will not be needed as the dependencies are taken care by the resolvers
+	CompatibilityReprocessDeployments []string
+}
+
+// WrapOutputMessage wraps the SensorEvents, CompatibilityDetectionMessages, and the CompatibilityReprocessDeployments into a ResourceEvent message
+func WrapOutputMessage(sensorMessages []*central.SensorEvent, detectionDeployment []CompatibilityDetectionMessage, reprocessDeploymentsIds []string) *ResourceEvent {
```
fredrb, line 32: nit: This isn't wrapping anything, this is more like a constructor for `ResourceEvent`. 

Maybe let's just call it `NewResourceEvent`

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -0,0 +1,52 @@
...
+	return &ResourceEvent{
+		ForwardMessages:                   sensorMessages,
+		CompatibilityDetectionDeployment:  detectionDeployment,
+		CompatibilityReprocessDeployments: reprocessDeploymentsIds,
+	}
+}
+
+// MergeOutputMessages merges two ResourceEvents
+func MergeOutputMessages(dest, src *ResourceEvent) *ResourceEvent {
```
fredrb, line 41: I think this name is also obsolete. In the original code we called the object `OutputMessage`, but it's now `ResourceEvent`. Let's rename this to `MergeResourceMessages`

sensor/kubernetes/eventpipeline/output/mocks/output.go
```diff
@@ -0,0 +1,35 @@
+package mocks
```
fredrb, line 1: Shouldn't this be generated by gomock?

sensor/kubernetes/eventpipeline/component/component.go
```diff
@@ -0,0 +1,16 @@
...
+
+// PipelineComponent components that constitute the eventPipeline
+type PipelineComponent interface {
+	Start() error
+	Stop(error)
+	ResponsesC() <-chan *central.MsgFromSensor
+}
+
+// OutputQueue component that redirects Resource Events and Alerts to the output channel
+type OutputQueue interface {
```
fredrb, line 13: Add
```
// go:generate mockgen-wrapper
```
And run the code generation step (I think `make go-generated-srcs`?) to generate the gomock implementation of this other than implementing it by hand.

sensor/kubernetes/eventpipeline/output/output_impl.go
```diff
@@ -0,0 +1,70 @@
...
+func (q *outputQueueImpl) Start() error {
+	go q.runOutputQueue()
+	return nil
+}
+
+// Stop stops the outputQueueImpl component
+func (q *outputQueueImpl) Stop(_ error) {
+	defer close(q.innerQueue)
+	defer close(q.forwardQueue)
+	q.stopSig.Signal()
```
fredrb, line 42: I think there is some odd behavior here. `q.stopSig` is a pointer passed in the constructor. It makes sense to use it if `output` fails unexpectedly and we want to propagate the error up. However, calling it when it gets closed conflicts with the parent `Close()` call. Which does the same thing:

```go
// Stop implements common.SensorComponent
func (p *eventPipeline) Stop(err error) {
	defer close(p.eventsC)
	for _, c := range p.components {
		c.Stop(err)
	}
	p.stopSig.Signal()
}
```

This is confusing because `stopSig.Signal()` will be called more than once (pipeline and output) and if the behavior is copied, it will be called in other sub-components as well.

sensor/kubernetes/eventpipeline/output/output_impl.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// runOutputQueue reads messages from the inner queue, forwards them to the forwardQueue channel
+// and sends the deployments (if needed) to Detector
+func (q *outputQueueImpl) runOutputQueue() {
+	for {
+		select {
+		case <-q.stopSig.Done():
+			return
+		case msg, more := <-q.innerQueue:
```
fredrb, line 52: Related to the message above, it seems redundant to have both a `case` for `stropSig` being triggered and handling closed channels. 

Maybe a simpler way to code this is to drop the `select` block entirely, since we can rely on `Stop()` being called and channels being closed.

```go
func (q *outputQueueImpl) runOutputQueue() {
	for {
                msg, more := <-q.innerQueue
                if !more {
                       return
                }
                // ...
        }
}
```
Reading on closed channels will return `(nil, true)` if there are no more values buffered. 

For now we can do without the signal and I think the simplification in the code is justified. I'm fine to introduce it back again in case we want to propagate output errors.

sensor/kubernetes/eventpipeline/pipeline_impl.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
+
+type eventPipeline struct {
+	output     component.OutputQueue
+	components []component.PipelineComponent
```
fredrb, line 17: I don't see the advantage of having `components` slice. The only benefit is being able to iterate on them on `Stop()`. But since there won't be many, listing them and calling directly seems fine to me. IMO it makes it less clear to me what components are there, because I saw both a slice of `component.PipelineComponent` and a `component.OutputQueue` one. 

I'll leave it up to you if you want to keep it this way. I don't have strong arguments against it.

sensor/kubernetes/listener/resources/networkpolicy_test.go
```diff
@@ -400,12 +396,13 @@ func (suite *NetworkPolicyDispatcherSuite) Test_ProcessEvent() {
 				deleteMock.Times(0)
 			}
 			events := suite.dispatcher.ProcessEvent(c.netpol, c.oldNetpol, c.action)
-			reprocessDeploymentMock.Times(1)
+			deps := set.NewStringSet()
```
fredrb, line 399: nit: Maybe a `require.NotNil(events)` would be good here?

fredrb (changes requested):
Nice job on fixing the tests! Here are some comments focused on the code structure. 

After from this review, I think we're almost there to get this merged:
- [x] Similar benchmarks for processing deployments
- [x] CI tests green
- [ ] Fake workloads + long running cluster behaving normally
- [x] Run E2E tests on OpenShift/OSD clusters (since there are a lot of OpenShift specific resources being listened to, I think we should run them too to make sure nothing broke).

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: done

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: done

sensor/kubernetes/eventpipeline/output/mocks/output.go
lvalerom: changed

sensor/kubernetes/eventpipeline/component/component.go
lvalerom: done

sensor/kubernetes/eventpipeline/output/output_impl.go
lvalerom: That was a mistake on my part. The idea is to have independent signals for each component. Similar to how it’s done for `SensorComponents`. This way each component is in charge of stopping itself. Now I see your point of wanting to propagate an error in a sub-component up. I think if we want to do that we should use a different signal (e.g. `errorSignal`) and have the parent component listen to the error signal in its main loop. WDYT?

sensor/kubernetes/eventpipeline/output/output_impl.go
lvalerom: The only problem that I see there is: If we call `Stop()` and there are 10 events buffered in the channel, we won’t stop until all of them are processed (if I'm not mistaken). Wouldn’t it make sense to just stop processing at the time Stop is called instead of processing the 10 events and then stopping?

sensor/kubernetes/eventpipeline/pipeline_impl.go
lvalerom: That was probably over architecting on my side. Once the components are created, the `eventPipeline` only needs to know about the output to read it; therefore, we can create the PipelineComponent abstraction and just iterate through it when we want to start or stop. But you are right, there are only three components so I think I’ll just change it to call the components directly.

sensor/kubernetes/eventpipeline/pipeline_impl.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	go p.forwardMessages()
+	return nil
+}
+
+// Stop implements common.SensorComponent
+func (p *eventPipeline) Stop(err error) {
+	defer close(p.eventsC)
+	for _, c := range p.components {
+		c.Stop(err)
```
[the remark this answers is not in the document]
lvalerom: Again, probably over architecting 😅  but the idea behind that is: If the parent component stops due to an error, it calls to stop the rest of the sub-components and if some fail onStop, we could wrap the errors so we can keep track of them. But I think a better way to do something like that would be returning an error on Stop.

sensor/kubernetes/listener/resources/networkpolicy_test.go
lvalerom: added

sensor/kubernetes/eventpipeline/output/output.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/sensor/common/detector"
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component"
+)
+
+var (
+	boundedQueueSize = 100
```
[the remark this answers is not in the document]
lvalerom: Added `EventPipelineOutputQueueSize` variable 🙂

sensor/kubernetes/eventpipeline/output/output_impl.go
fredrb: Yes, you're correct. That's a fair point. I think the `Stop` and `ErrorSignal` path can distinguish the graceful stop and crashing.

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// CompatibilityDetectionMessage should be used by old handlers
+// it's here for retrocompatibility reasons.
+type CompatibilityDetectionMessage struct {
+	// Object The deployment object that needs to be processed by the detector
```
fredrb, line 11:
```suggestion
	// Object references the object that needs to be processed by the detector
```

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -0,0 +1,52 @@
...
+	Object *storage.Deployment
+	// Action the event action (CREATE_RESOURCE, REMOVE_RESOURCE, UPDATE_RESOURCE, and SYNC_RESOURCE)
```
fredrb, line 13:
```suggestion
	// Action of an event (CREATE_RESOURCE, REMOVE_RESOURCE, UPDATE_RESOURCE, and SYNC_RESOURCE)
```

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ResourceEvent message used by the event pipeline's components
+type ResourceEvent struct {
+	// ForwardMessages messages generated by the handlers that need to be forwarded to central
+	ForwardMessages []*central.SensorEvent
+	// CompatibilityDetectionDeployment should be used by old handlers
+	// and it's here for retrocompatibility reasons.
+	// This property should be removed in the future and only the
+	// DetectionObject should be sent
```
fredrb, line 24: I think this comment was written based on the first iteration of this. 
```suggestion
	// deployment references should be sent
```

sensor/kubernetes/eventpipeline/output/output.go
```diff
@@ -0,0 +1,19 @@
+package output
+
+import (
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/sensor/common/detector"
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component"
+)
+
+// New Creates a new Queue component
```
fredrb, line 9:
```suggestion
// New instantiates a an output Queue component
```

sensor/kubernetes/eventpipeline/output/output_impl.go
```diff
@@ -0,0 +1,63 @@
...
+	log = logging.LoggerForModule()
+)
+
+type outputQueueImpl struct {
+	innerQueue   chan *component.ResourceEvent
+	forwardQueue chan *central.MsgFromSensor
+	detector     detector.Detector
+}
+
+// Send sends a ResourceEvent message to the inner queue
```
fredrb, line 20: The function name is part of the godoc text
```suggestion
// Send a ResourceEvent message to the inner queue
```

sensor/kubernetes/eventpipeline/output/output_impl.go
```diff
@@ -0,0 +1,63 @@
...
+func (q *outputQueueImpl) Send(msg *component.ResourceEvent) {
+	q.innerQueue <- msg
+}
+
+// ResponsesC returns the MsgFromSensor channel
+func (q *outputQueueImpl) ResponsesC() <-chan *central.MsgFromSensor {
+	return q.forwardQueue
+}
+
+// Start starts the outputQueueImpl component
```
fredrb, line 30:
```suggestion
// Start the outputQueueImpl component
```

sensor/kubernetes/eventpipeline/output/output_impl.go
```diff
@@ -0,0 +1,63 @@
...
+func (q *outputQueueImpl) Start() error {
+	go q.runOutputQueue()
+	return nil
+}
+
+// Stop stops the outputQueueImpl component
```
fredrb, line 36:
```suggestion
// Stop the outputQueueImpl component
```

sensor/kubernetes/eventpipeline/pipeline.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/sensor/common"
+	"github.com/stackrox/rox/sensor/common/config"
+	"github.com/stackrox/rox/sensor/common/detector"
+	"github.com/stackrox/rox/sensor/kubernetes/client"
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/output"
+	"github.com/stackrox/rox/sensor/kubernetes/listener"
+)
+
+// New Creates a new eventPipeline component
```
fredrb, line 18:
```suggestion
// New instantiates the eventPipeline component
```

fredrb (approved):
Just some nitpick on the godoc comments

PR #3656, opened 2022-11-02
Add benchmark to sensor message processing

fredrb:
## Description

Adds benchmark tests for sensor event ingestion. The idea is to record how long it takes for a list of deployment events to go through the pipeline. Since the deployment's handler is synchronous, they will be processed one by one. 

We won't see much difference at first when running this against the the new version of the pipeline in #3653, but at least it might give a good indicator that things are not *much* worse. 

## Checklist

## Testing Performed

```
% LOGLEVEL="panic" go test -bench=. -run=^# -benchtime=30s -count=5 ./sensor/tests/pipeline | tee pipeline.results.txt
goos: darwin
goarch: amd64
pkg: github.com/stackrox/rox/sensor/tests/pipeline
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Benchmark_Pipeline-12    	     133	 296588557 ns/op
Benchmark_Pipeline-12    	     100	 311108985 ns/op
Benchmark_Pipeline-12    	     124	 284232498 ns/op
Benchmark_Pipeline-12    	     120	 313928439 ns/op
Benchmark_Pipeline-12    	     123	 277223344 ns/op
PASS
ok  	github.com/stackrox/rox/sensor/tests/pipeline	304.525s
```

Tests seem to be consistent enough (only 7% difference between samples):
```
% benchstat ./pipeline.results.txt
name          time/op
_Pipeline-12  297ms ± 7%
```


lvalerom (approved):
LGTM. I've created (as we discussed) the follow up task [ROX-13414](https://issues.redhat.com/browse/ROX-13414) to move the initialization functions used here and the integration test to a common package.

PR #3657, opened 2022-11-02
Rename variable ScannerGRPCEndpoint to ScannerSlimGRPCEndpoint

SimonBaeumer:
## Description

Rename scanner env variable to better differentiate that this variable is used by Sensor for Slim Scanners.
I was confused in PR https://github.com/stackrox/stackrox/pull/3469 because this variable is set differently from Central's endpoint configuration.

## Checklist
- [ ] 
- [ ] ~~Documented user facing changes (create PR based on [openshift/openshift-docs]~~(https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

 - CI

### DIFF
--- a/pkg/env/sensor.go
+++ b/pkg/env/sensor.go
@@ -13,9 +13,9 @@ var (
 	// SensorEndpoint is used to communicate the sensor endpoint to other services in the same cluster.
 	SensorEndpoint = RegisterSetting("ROX_SENSOR_ENDPOINT", WithDefault("sensor.stackrox.svc:443"))
 
-	// ScannerGRPCEndpoint is used to communicate the scanner endpoint to other services in the same cluster.
+	// ScannerSlimGRPCEndpoint is used to communicate the scanner endpoint to other services in the same cluster.
 	// This is typically used for Sensor to communicate with a local Scanner-slim's gRPC server.
-	ScannerGRPCEndpoint = RegisterSetting("ROX_SCANNER_GRPC_ENDPOINT", WithDefault("scanner.stackrox.svc:8443"))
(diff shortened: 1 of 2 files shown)


pkg/env/sensor.go
```diff
@@ -12,9 +12,9 @@ var (
 	// SensorEndpoint is used to communicate the sensor endpoint to other services in the same cluster.
 	SensorEndpoint = RegisterSetting("ROX_SENSOR_ENDPOINT", WithDefault("sensor.stackrox.svc:443"))
 
-	// ScannerGRPCEndpoint is used to communicate the scanner endpoint to other services in the same cluster.
+	// ScannerSlimGRPCEndpoint is used to communicate the scanner endpoint to other services in the same cluster.
 	// This is typically used for Sensor to communicate with a local Scanner-slim's gRPC server.
-	ScannerGRPCEndpoint = RegisterSetting("ROX_SCANNER_GRPC_ENDPOINT", WithDefault("scanner.stackrox.svc:8443"))
+	ScannerSlimGRPCEndpoint = RegisterSetting("ROX_SCANNER_GRPC_ENDPOINT", WithDefault("scanner.stackrox.svc:8443"))
```
RTann, line 17: Typically used for this, but it's possible Sensor communicates with a full-Scanner (when this is the Central cluster). I'll check the other PR because I'm curious where the confusion came from

SimonBaeumer: The confusion was there because I search for scanner endpoints and found this env variable. Without looking at the concrete package path you can't see whether the variable belongs to sensor or central.

Renaming the variable to slim helps distinguishing when the env var is read. I do agree that it is not 100% technically correct in that case.

PR #3661, opened 2022-11-02
ROX-11826: Disable kernel support package uploads for managed central

theencee:
## Description

This is only used for those running in offline mode and thus is unnecessary in managed central. Uses a new env var, `ROX_DISABLE_KERNEL_PACKAGE_UPLOAD`, which is enabled by default for managed central (via helm).

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/main.go
+++ b/central/main.go
@@ -672,16 +672,20 @@ func customRoutes() (customRoutes []routes.CustomRoute) {
 
 	if env.PostgresDatastoreEnabled.BooleanSetting() {
 		customRoutes = append(customRoutes, routes.CustomRoute{
-			Route:         "/db/backup",
-			Authorizer:    dbAuthz.DBReadAccessAuthorizer(),
-			ServerHandler: httputil.HandlerNotImplementedIfSettingDisabled(!env.ManagedCentral.BooleanSetting(), globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), false)),
-			Compression:   true,
+			Route:      "/db/backup",
+			Authorizer: dbAuthz.DBReadAccessAuthorizer(),
+			ServerHandler: utils.IfThenElse[http.Handler](
+				env.ManagedCentral.BooleanSetting(), httputil.NotImplementedHandler("api is not supported in a managed central environment."),
+				globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), false)),
+			Compression: true,
 		})
 		customRoutes = append(customRoutes, routes.CustomRoute{
-			Route:         "/api/extensions/backup",
-			Authorizer:    user.WithRole(role.Admin),
-			ServerHandler: httputil.HandlerNotImplementedIfSettingDisabled(!env.ManagedCentral.BooleanSetting(), globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), true)),
-			Compression:   true,
+			Route:      "/api/extensions/backup",
+			Authorizer: user.WithRole(role.Admin),
+			ServerHandler: utils.IfThenElse[http.Handler](
+				env.ManagedCentral.BooleanSetting(), httputil.NotImplementedHandler("api is not supported in a managed central environment."),
+				globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), true)),
+			Compression: true,
 		})
 		customRoutes = append(customRoutes, routes.CustomRoute{
 			Route:         "/api/export/csv/node/cve",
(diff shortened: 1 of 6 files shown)


central/main.go
```diff
@@ -673,13 +674,13 @@ func customRoutes() (customRoutes []routes.CustomRoute) {
 		customRoutes = append(customRoutes, routes.CustomRoute{
 			Route:         "/db/backup",
 			Authorizer:    dbAuthz.DBReadAccessAuthorizer(),
-			ServerHandler: notImplementedOnManagedServices(globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), false)),
+			ServerHandler: httputil.HandlerNotImplementedIfSettingDisabled(!env.ManagedCentral.BooleanSetting(), globaldbHandlers.BackupDB(nil, nil, globaldb.GetPostgres(), false)),
```
misberner, line 677: I don't have a good solution to that problem, but I find it kind of annoying that using the handler to convey a "not implemented" means the entire rest of the pipeline still runs - authn (probably desired), audit (maybe okay), authz (meh - it's odd to say "not authorized" on an unimplemented API). I wondered if it wouldn't be better to only conditionally append to `customRoutes`, but then we'd just get a 404 instead of a more meaningful not implemented message..

pkg/httputil/handler.go
```diff
@@ -13,3 +15,16 @@ func WrapHandlerFunc(handlerFn func(req *http.Request) error) http.HandlerFunc {
 		}
 	})
 }
+
+// HandlerNotImplementedIfSettingDisabled takes a handler func and returns a handler func that will error out if
+// specified boolean is disabled, otherwise it will return the given handler func
+func HandlerNotImplementedIfSettingDisabled(enabled bool, fn http.Handler) http.Handler {
```
misberner, line 21: This function seems oddly specific for a utility package.
1. Why is it relevant that the `bool` parameter corresponds to a _setting_?
2. What if one wanted to customize the error message?

I think a more elegant way to solve this might be by introducing a general-purpose ITE (if-then-else)/ternary operator function, e.g., in `pkg/utils`, like:
```go
func ITE[T any](cond bool, a, b T) T {
  if cond {
    return a
  }
  return b
}
```
and then also have something like a `func NotImplementedHandler(msg string) http.Handler`. You can then invoke this as, e.g.,
```go
utils.ITE[http.Handler](env.DisableKernelPackageUpload.BooleanSetting(), httputil.NotImplementedHandler("..."), http.HandlerFunc(s.handleProbeUpload))
```

misberner (approved):
I find the function odd and not worthy of a utility function, if that makes sense (I would propose to move it to somewhere under `central/` at least, as we won't need it elsewhere), but I want to remain conscious of the overengineering risk so no strong demands to change the approach

rukletsov (approved):
Modulo Malte's comments.

pkg/env/kernel_package_upload.go
```diff
@@ -0,0 +1,6 @@
+package env
+
+var (
+	// DisableKernelPackageUpload is set to true to signal that kernel support package uploads should be disabled.
+	DisableKernelPackageUpload = RegisterBooleanSetting("ROX_DISABLE_KERNEL_PACKAGE_UPLOAD", false)
```
SimonBaeumer, line 5: Can we invert this setting to a positive indication like `ROX_ENABLE_KERNEL_PACKAGE_UPLOAD` to be easier to follow the code paths (and reducing inversions in if-cases)?

central/probeupload/service/service_impl.go
```diff
@@ -79,7 +80,7 @@ func (s *service) CustomRoutes() []routes.CustomRoute {
 		{
 			Route:         "/api/extensions/probeupload",
 			Authorizer:    user.With(permissions.Modify(resources.ProbeUpload)),
-			ServerHandler: http.HandlerFunc(s.handleProbeUpload),
+			ServerHandler: httputil.HandlerNotImplementedIfSettingDisabled(!env.DisableKernelPackageUpload.BooleanSetting(), http.HandlerFunc(s.handleProbeUpload)),
```
SimonBaeumer, line 83: Three negations here make the code hard to understand.
My suggestion:

 - Rename function to `httputil.DisableHandler(bool, func)`, a true disables the handler. Or `EnableHandler` which reacts on a positive setting.
 - Changing the env var to a positive `EnableKernelPackageUpload`

```suggestion
			ServerHandler: httputil.HandlerEnabled(env.EnableKernelPackageUpload.BooleanSetting(), http.HandlerFunc(s.handleProbeUpload)),
```

SimonBaeumer (changes requested):

md2119 (approved):

pkg/httputil/handler.go
theencee: I totally forgot we support generics now. Love this idea so I'll switch to this

central/probeupload/service/service_impl.go
theencee: I'm using the IfThenElse and with the switch to `EnableKernelPackageUpload` it'll become true enables it and false returns not implemented.

central/main.go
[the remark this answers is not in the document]
theencee: I switched to a If-Then-Else construct which won't support this, but does it make it easy to read for this (more common) use case

theencee (commented):
had to rebase to fix conflicts

SimonBaeumer (approved):

PR #3663, opened 2022-11-02
ROX-13357: Service/Deployment store refactor

fredrb:
## Description

Refactor services and deployment stores so they expose relationship behavior publicly. This will be needed for disabling sensor re-sync. 

Most of the code in this PR is just moved around and refactored to expose properties/interfaces that were private. 

**Motivation: Why are we moving the code?**
1. Deployment dependencies will be resolved outside of the deployment handler. In order to provide the needed dependencies externally (hence the new method `BuildDeploymentWithDependencies(id string, dependencies Dependencies) (*storage.Deployment, error)`), sibling components need to be able to access service and RBAC stores to build the dependency object. 
2. The `resource` package is really big. Part of this refactor is also aimed to reduce the scope of responsibility of this package. Pulling store implementation and dependency resolution to other packages as a long-term goal will make `resource` easier to maintain, as it does one thing: parse k8s resources. 
3. Resource wrappers (like service and deployment wrapper) couple the parsing logic in the stored object. This PR splits the two for service wrapper, where the wrapper struct only has the properties and the parsing logic is moved to `kubernetes/convert/service` package.

I've started moving deployment related wrapper functions as an exercise, but the change became way too big and difficult to make sense of. This is a minimum change necessary to allow the re-sync work to continue. 

## Checklist


sensor/common/store/service/servicewrapper/wrap.go
```diff
@@ -0,0 +1,121 @@
+package servicewrapper
```
fredrb, line 1: Note to reviewers: this **isn't** new code. This was moved from `listener/resources/service.go`. Some struct/func names were changed.

lvalerom (commented):
Quick question before continuing the review: Wouldn't make sense to move the `DeploymentStore` and the `deploymentWrap` out of the `resources` package too?

Edit: Looking more into it. I saw that simply moving the structures will create a cyclic dependency with the `updatePortExposureFromStore` function of the `deploymentWrap`. But I think it could be solve by just extracting that function from the `deploymentWrap`:

```golang
func UpdatePortExposureFromStore(w *deployment.DeploymentWrap, store ServiceStore) {
	svcs := store.GetMatchingServicesWithRoutes(w.Namespace, w.PodLabels)
	w.UpdatePortExposureFromServices(svcs...)
}
```

Or simply or simply changing the call in the `deploymentHandler`:

```golang
svcs := store.GetMatchingServicesWithRoutes(w.Namespace, w.PodLabels)
deploymentWrap.UpdatePortExposureFromServices(svcs...)
```

WDYT?

fredrb:
> Quick question before continuing the review: Wouldn't make sense to move the DeploymentStore and the deploymentWrap out of the resources package too?

The "problem" is that this requires exposing the `deploymentWrap` publicly as well. Because the deployment handler uses the wrapper extensively. 

It's not a real problem though, and maybe is justified now that the service wraps are also exposes publicly. 

I'll look into it.

fredrb:
Closing in favor of #3851

PR #3664, opened 2022-11-02
Don't use protobuf values (instead of pointers) in function and method signatures

misberner:
## Description

Copying a protobuf message will result in a vet error in a [future version of the protobuf library](https://gist.github.com/misberner/de1f70db96defcf562da3b8956c8151c), so create a linter to prohibit that, and fix all existing usages.

## Checklist

## Testing Performed

- CI
- Ran `make roxvet` locally until no further errors were reported

### DIFF
--- a/tools/roxvet/analyzers/protoptrs/analyzer.go
+++ b/tools/roxvet/analyzers/protoptrs/analyzer.go
@@ -6,6 +6,7 @@ import (
 	"go/types"
 	"regexp"
 
+	"github.com/stackrox/rox/tools/roxvet/common"
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/inspect"
 	"golang.org/x/tools/go/ast/inspector"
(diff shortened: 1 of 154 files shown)


misberner:
Rebased this on top of https://github.com/stackrox/stackrox/pull/3665 and reverted the easyjson change as I figured introducing a fork might be controversial.

janisz (approved):

tools/roxvet/analyzers/protoptrs/analyzer.go
```diff
@@ -0,0 +1,78 @@
...
+	Run:      run,
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	inspectResult := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
+	nodeFilter := []ast.Node{
+		(*ast.FuncType)(nil),
+		(*ast.StructType)(nil),
+	}
+	common.FilteredPreorder(inspectResult, common.Not(common.IsGeneratedFile), nodeFilter, func(n ast.Node) {
```
dhaus67, line 39: out of curiosity: do we also get `ast.FuncType/StructType` from the go template file in `generate-helpers/pg-table-bindings/migration.go.tpl`? Since you fixed the postgres migrations and some of them _should_ be detected as generated files.

If that's the case, this is quite neat to get it from the template files!

dhaus67 (approved):

tools/roxvet/analyzers/protoptrs/analyzer.go
misberner: I wish it worked that way, but unfortunately no, vet-tools simply look at actual Go code, not the templates. Those fixes were applied manually by me to the templates, after my IDE warned me that the file I was about to edit (based on the `roxvet` finding) was generated.

misberner:
The failure is unrelated to this PR and discussed here: https://srox.slack.com/archives/C030MTTE1S5/p1667496808272139

PR #3665, opened 2022-11-02
Support file-filtered traversal in roxvet

misberner:
## Description

Excluding certain file types (generated files, test files) in `roxvet` is a common desire. Therefore, add a generic mechanism to do so.

The refactoring uncovered a bug in the `uncheckedifassign` analyzer, where an if with assignment inside the then-block of an if with assignment wasn't checked properly, which exhibited that the `uncheckedifassign` analyzer was missing logic for inc/dec statements.

## Checklist

## Testing Performed

- CI
- Ran `make roxvet` locally


tools/roxvet/analyzers/uncheckedifassign/analyzer.go
```diff
@@ -60,31 +59,21 @@ func checkAnyObjectUsedIn(expr ast.Expr, objs map[*ast.Object]struct{}) bool {
...
-		if file, _ := n.(*ast.File); file != nil {
-			if len(file.Comments) > 0 && len(file.Comments[0].List) > 0 {
-				return !generatedFileCommentRegex.MatchString(file.Comments[0].List[0].Text)
-			}
-		}
-
-		ifStmt, _ := n.(*ast.IfStmt)
-		if ifStmt == nil || ifStmt.Init == nil {
-			return false
+	common.FilteredPreorder(inspectResult, common.Not(common.IsGeneratedFile), nodeFilter, func(n ast.Node) {
```
RTann, line 67: why switch this to preorder?

tools/roxvet/analyzers/uncheckedifassign/analyzer.go
```diff
@@ -24,18 +20,21 @@ var Analyzer = &analysis.Analyzer{
...
+	switch initAssign := ifStmt.Init.(type) {
+	case *ast.AssignStmt:
+		for _, expr := range initAssign.Lhs {
+			idExpr, _ := expr.(*ast.Ident)
+			if idExpr == nil || idExpr.Obj == nil || idExpr.Obj.Kind != ast.Var {
+				continue
+			}
+			objs[idExpr.Obj] = struct{}{}
+		}
+	case *ast.IncDecStmt:
```
RTann, line 33: is this `x += 1` or `x++`?

tools/roxvet/common/file_filters.go
```diff
@@ -0,0 +1,41 @@
...
+	if firstComment.Pos() > f.Package {
+		// comment is not at beginning of file, before the package keyword
+		return false
+	}
+
+	return generatedCodeRegex.MatchString(firstComment.Text)
+}
+
+// IsTestFile checks if the given file is a test file.
+// Note: rather than going off of a _test.go suffix, we check if the file imports the "testing" package.
```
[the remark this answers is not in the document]
misberner: I think so. Excluding testing files from vet checks is motivated by the assumption that code not being used in production does not need to adhere to as strict standards as production code. If a file imports `testing`, it should not contain code used in production.

tools/roxvet/common/traverse.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"go/ast"
+
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+// FilteredPreorder calls `inspector.Preorder(nodeTypes, fn)`, but filters out all files that do not pass the given
+// fileFilter.
+func FilteredPreorder(inspector *inspector.Inspector, fileFilter FileFilter, nodeTypes []ast.Node, fn func(n ast.Node)) {
+	effTypes := nodeTypes[:len(nodeTypes):len(nodeTypes)]
```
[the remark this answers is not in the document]
misberner: `eff` = `effective`. Setting the capacity ensures that the call to `append` will allocate a new slice. Too subtle?

tools/roxvet/analyzers/uncheckedifassign/analyzer.go
misberner: The only reason to use `Nodes` here was to allow preventing the traversal of generated `*ast.File` nodes. Doing so in a preorder traversal is possible, but much less intuitively so than in a DFS traversal. However, given that we now have a generic utility method, there's no harm in implementing the file skipping logic once and then just using it here. And ultimately, if we can use `Preorder`, we should, because it's simpler (as is apparent from the callback function signature).

tools/roxvet/analyzers/uncheckedifassign/analyzer.go
misberner: This is `x++`. `x += 1` is actually an `*ast.AssignStmt`

RTann: I don't think I've ever seen this used in an `if`. Do we have any examples?

tools/roxvet/common/traverse.go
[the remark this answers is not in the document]
RTann: I had to look up the meaning of the second `:` because I haven't really seen it. Why do we want to ensure a new backing array is allocated? To make `nodeTypes` effectively read-only? Does it have to be read-only?

tools/roxvet/common/traverse.go
```diff
@@ -0,0 +1,62 @@
...
+	hadFile := hasFile(nodeTypes)
+	if !hadFile {
+		effTypes = append(effTypes, (*ast.File)(nil))
+	}
+	skip := false
```
RTann, line 17: I won't block this PR, but I've been told `var skip bool` is preferred, so I usually tell others this, too. However, this isn't a core part of the product, so I don't mind as much

tools/roxvet/common/traverse.go
```diff
@@ -0,0 +1,62 @@
...
+	inspector.Preorder(effTypes, func(n ast.Node) {
+		if f, ok := n.(*ast.File); ok {
+			skip = !fileFilter(f)
```
RTann, line 20: Just for my own understanding, we come across some file which we do not care about. `skip` will tell us to ignore this file plus any other child file (I'm assuming that just means skipping a directory, sub directories, and any files in that tree). `skip` will be reset to `false` once we exit that tree and look at a tree we do care about. Is this right?

tools/roxvet/common/traverse.go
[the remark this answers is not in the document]
misberner: It's just common practice to ensure that parameters are not modified, unless this is made explicit in the function signature.

Additionally, `append` will not modify the visible value of `nodeTypes`. But if somebody passed something like `allNodeTypes[:len(allNodeTypes)-1]`, the last element might get overwritten. This is very subtle and a nightmare to debug, so I'm usually extra careful with `append`.

tools/roxvet/analyzers/uncheckedifassign/analyzer.go
misberner: Yeah, for example `pkg/networkgraph/tree/radix.go:160`. I saw this failing on `make roxvet` with the earlier changes.

tools/roxvet/common/traverse.go
misberner: Good point, will change this

tools/roxvet/common/traverse.go
misberner: Not quite, in the AST representation, there are no "child files". All the children of the `*ast.File` are the AST elements contained in the `.go` file, such as imports, `var (...)` blocks, function declarations etc. In the preorder traversal, first the `*ast.File` node will be emitted, followed by all its children. Therefore, everything that comes after an `*ast.File` that is not itself an `*ast.File` belongs to that file, another `*ast.File` node signals the beginning of a new file.

tools/roxvet/common/traverse.go
misberner:
```suggestion
	var skip bool
```

janisz (commented):
How about adding unit tests for this?

dhaus67 (approved):
I wish I would understand the `ast` package more to give you a better review.

I've looked at the `roxvet/common/file_filters.go` and the `common/traverse.go` and those, TTBOMK, look good to me. So, approving for those.

misberner:
> How about adding unit tests for this?

Good callout. Added, and fixed some logic issues.

misberner:
Merging since all roxvet-related stages have succeeded.

PR #3671, opened 2022-11-03
ROX-13226: Add panic handler for Istio updateCVEs

mtodor:
## Description

This PR adds a panic handler in Istio `updateCVEs` function.

**Some considerations**

I have looked at call hierarchy. The problematic function `NvdCVEToEmbeddedCVE` is always called from `updateCVEs -> NvdCVEsToEmbeddedCVEs -> NvdCVEToEmbeddedCVE`. But `updateCVEs` is called from 3 different functionalities:
- online mode - when the file is downloaded
- offline mode - when the file is uploaded by the user
- init - when the application starts. I'm not 100% sure, but I think that in this case, the file is already embedded in a docker image.

Going upper in the call hierarchy would mean that we should cover all 3 call paths, and we already have proper error handling there. So, I think it's sufficient to throw an error from `updateCVEs` to the caller.

## Checklist

## Testing Performed

1. Tested with added unit test

### DIFF
--- a/central/cve/fetcher/istio.go
+++ b/central/cve/fetcher/istio.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/facebookincubator/nvdtools/cvefeed/nvd/schema"
-	"github.com/pkg/errors"
 	"github.com/stackrox/k8s-istio-cve-pusher/nvd"
 	clusterDataStore "github.com/stackrox/rox/central/cluster/datastore"
 	clusterCVEEdgeDataStore "github.com/stackrox/rox/central/clustercveedge/datastore"
@@ -20,6 +19,7 @@ import (
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/httputil"
+	"github.com/stackrox/rox/pkg/safe"
 	"github.com/stackrox/rox/pkg/sync"
 	"github.com/stackrox/rox/pkg/urlfmt"
 )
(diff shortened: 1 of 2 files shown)


mtodor:
* master
  * **PR #3671** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3671" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3671?utm_source=stack-comment).

central/cve/fetcher/istio.go
```diff
@@ -67,7 +68,13 @@ func (m *istioCVEManager) setCVEs(cves []*storage.EmbeddedVulnerability, nvdCVEs
 	m.embeddedCVEs = cves
 }
 
-func (m *istioCVEManager) updateCVEs(newCVEs []*schema.NVDCVEFeedJSON10DefCVEItem) error {
+func (m *istioCVEManager) updateCVEs(newCVEs []*schema.NVDCVEFeedJSON10DefCVEItem) (retErr error) {
+	defer func() {
```
misberner, line 72: Why not simply wrap a call to `updateCVEs` in `safe.RunE` from `github.com/stackrox/rox/pkg/safe`?

central/cve/fetcher/istio.go
```diff
@@ -67,7 +68,13 @@ func (m *istioCVEManager) setCVEs(cves []*storage.EmbeddedVulnerability, nvdCVEs
...
+		if r := recover(); r != nil {
```
misberner, line 73: Note that this does not catch a `panic(nil)` (and doing so is possible, but definitely a bit of a pain). `safe.RunE` handles this correctly. (I wouldn't be saying this if I hadn't once wasted hours hunting a mysterious API failure with no trace in the logs but still completely bogus results, until I discovered that a `panic(nil)` was just swallowed because of code like this)

misberner (changes requested):

central/cve/fetcher/istio_test.go
```diff
@@ -0,0 +1,28 @@
...
+	newCVEs := make([]*schema.NVDCVEFeedJSON10DefCVEItem, 0)
+
+	var m *istioCVEManager
+	if buildinfo.ReleaseBuild {
+		// Panic should happen because m is nil, and we use
+		// several struct fields in called functions.
+		err := m.updateCVEs(newCVEs)
+
+		assert.Error(t, err)
+		assert.True(t, strings.HasPrefix(err.Error(), "caught panic"), "Error should be returned by panic handler.")
```
RTann, line 22: not a big deal, but I think you can also do `assert.ErrorContains(t, err, "caught panic")`

RTann (approved):

central/cve/fetcher/istio_test.go
mtodor: Tnx for the suggestion. I'll leave it as it is to ensure that the last `RunE` handled panic.

mtodor:
@RTann Tnx for the review!

PR #3672, opened 2022-11-03
Generalize User-Agent setup

0x656b694d:
## Description

Follow-up on ROX-12987 (`roxctl` User-Agent), set the header in other binaries.

## Checklist

## Testing Performed

The agents are seen on amplitude charts and were seen in the debug log.

```
pkg/telemetry/marketing/amplitude: 2022/11/04 14:24:51.660065 amplitude.go:86: Info: Tracking event API Call with map[Code:0 Path:/v1.ImageService/ScanImageInternal User-Agent:sensor-k8s/3.72.x-409-gc158c71d8e-dirty (linux; amd64) grpc-go/1.50.1]
```


central/main.go
```diff
@@ -239,6 +240,8 @@ func main() {
 		return
 	}
 
+	clientconn.SetUserAgent("central")
```
janisz, line 243: does it make sense to put more data in user agent like version and platform like Mozilla do https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox

janisz (approved):

pkg/clientconn/useragent.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var userAgent string
+
+func init() {
+	SetUserAgent("stackrox")
+}
+
+// SetUserAgent formats and sets a value for the User-Agent for the process.
+func SetUserAgent(agent string) {
```
janisz, line 19: Could you add a test so we can see how user agent will look like?

dhaus67, line 19: Instead of relying on the processes to call `SetUserAgent`, we could do something similar to our singleton's for datastore's, meaning we use `once.Do()` within `GetUserAgent` to set the user agent once and afterwards just return it.

This way we don't have the dependency on the consumer on making sure to have called `SetUserAgent` beforehand. Otherwise, I would expect a comment for `GetUserAgent` stating explicitly that `SetUserAgent` has to be called beforehand, otherwise it will return an empty string.

sensor/kubernetes/main.go
```diff
@@ -46,6 +47,7 @@ func main() {
 	} else {
 		sharedClientInterface = client.MustCreateInterface()
 	}
+	clientconn.SetUserAgent("sensor-k8s")
```
dhaus67, line 50: Not simply just sensor? I don't know historically why there's this `kubernetes` directory, but the deployed component itself will be called `sensor`, so might as well just use it?

pkg/clientconn/useragent.go
0x656b694d: `GetUserAgent()` is called by packages which have no idea about the client binary, so the value has to be set on the process initialization. Default value is set to `stackrox`. I'll add a comment.

PR #3675, opened 2022-11-03
Move integration tests for page title from general to specific containers

pedrottimark:
### That was then and this is now

* It made sense to add the batch of tests for award-winning hackathon project.
* It was efficient to find assertions affected by product branding.

Now general test file is out of sight and out of mind:
* To follow patterns from existing containers for new containers.
* To add or update assertions if page titles change in UI.

Tentative pattern:
* List page: include title assertion in test with table column heading assertions.
* Form page: include title assertion in test of breadcrumbs, headings, and so on.
* Dashboard page: if it has only widget tests, then add a separate test for title assertion.

### Residue

1. Dashboard: Add integration test file.
2. Policy Management: Investigate inconsistencies in heading and titles.
3. System Health PatternFly: component does not render System Health title.

## Checklist

## Testing Performed


dvail (approved):
:shipit: LGTM

PR #3676, opened 2022-11-03
Add autocomplete for name selector dropdowns

dvail:
## Description

This adds the ability to autocomplete text input on the dropdowns for _by name_ rule selectors only.

### Side note

As the collection form grows, there is an increasing amount of prop drilling going on. This PR passes the `collection` value from `CollectionForm -> RuleSelector -> [ByNameSelector, ByLabelSelector]` and `RuleSelector -> [ByNameSelector, ByLabelSelector] -> AutocompleteSelector` components now all need access to the `entityType`. I tried playing around with storing these values in a React context at the RuleSelector level, but wasn't convinced it was any cleaner. I also took a quick look at how to make these more composable, but that seemed to just lift the verbosity up for no apparent gain. Maybe the components are just inherently coupled, but I'm open to suggestions on how to make this component tree easier to use.

## Checklist

## Testing Performed

Go to a collection page and select "Deployment with name matching" from the rule dropdown. Focus the text input box for the name value and a list of suggestions should appear.
image

Begin typing text into the input. A short delay after the last keystroke the items should update to only include values that match the entered text. Additionally, the last option should be the ability to "Create", or use, the value of the entered text directly.
image

Clicking on one of the text suggestions should make the value of the input box match the item that was selected.
image

Alternatively, clicking the "Create" option will keep the user entered value of the text in the dropdown.
image

The same behavior should occur for both Namespace and Cluster resources. These suggestions should have the correct resource icons to match the resource types:
image
image

At this point, label key and value dropdowns should not attempt to autocomplete text. (This is TBD whether or not we will implement this for MVP.)
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
+++ b/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
@@ -1,6 +1,16 @@
-import React from 'react';
-import { Select, ValidatedOptions } from '@patternfly/react-core';
+import React, { ReactElement, useCallback, useMemo, useState } from 'react';
+import {
+    debounce,
+    Select,
+    SelectOption,
+    SelectOptionProps,
+    ValidatedOptions,
+} from '@patternfly/react-core';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import useRestQuery from 'Containers/Dashboard/hooks/useRestQuery';
+import { CancellableRequest } from 'services/cancellationUtils';
+import ResourceIcon from 'Components/PatternFly/ResourceIcon';
+import { SelectorEntityType } from '../types';
 
 export type AutoCompleteSelectProps = {
     id: string;
@@ -10,9 +20,29 @@ export type AutoCompleteSelectProps = {
     onChange: (value: string) => void;
     validated: ValidatedOptions;
     isDisabled: boolean;
+    autocompleteProvider?: (search: string) => CancellableRequest<string[]>;
 };
 
-/* TODO Implement autocompletion */
+function ResourceSelectOption({ option, entityType }) {
+    return (
+        <SelectOption
+            value={option}
+            component={(props) => (
+                <div {...props}>
+                    <ResourceIcon kind={entityType} />
+                    {option}
+                </div>
+            )}
+        />
+    );
+}
+
+function getOptions(data: string[] | undefined): ReactElement[] | undefined {
+    return data?.map((option) => (
+        <ResourceSelectOption key={option} option={option} entityType="Deployment" />
+    ));
+}
+
 export function AutoCompleteSelect({
     id,
     selectedOption,
@@ -21,13 +51,34 @@ export function AutoCompleteSelect({
     onChange,
     validated,
     isDisabled,
+    autocompleteProvider,
 }: AutoCompleteSelectProps) {
     const { isOpen, onToggle, closeSelect } = useSelectToggle();
+    const [typeahead, setTypeahead] = useState(selectedOption);
+
+    const autocompleteCallback = useCallback(() => {
+        const shouldMakeRequest = isOpen && autocompleteProvider;
+        if (shouldMakeRequest) {
+            return autocompleteProvider(typeahead);
+        }
+        return {
+            request: Promise.resolve([]),
+            cancel: () => {},
+        };
+    }, [isOpen, autocompleteProvider, typeahead]);
+
+    const { data } = useRestQuery(autocompleteCallback);
 
     function onSelect(_, value) {
         onChange(value);
         closeSelect();
     }
+
+    // Debounce the autocomplete requests to not overload the backend
+    const updateTypeahead = useMemo(
+        () => debounce((value: string) => setTypeahead(value), 800),
+        []
+    );
 
     return (
         <>
@@ -39,11 +90,15 @@ export function AutoCompleteSelect({
                 variant="typeahead"
                 isCreatable
                 isOpen={isOpen}
+                onFilter={() => getOptions(data)}
                 onToggle={onToggle}
+                onTypeaheadInputChanged={updateTypeahead}
                 selections={selectedOption}
                 onSelect={onSelect}
                 isDisabled={isDisabled}
-            />
+            >
+                {getOptions(data)}
+            </Select>
         </>
     );
 }
--- a/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
+++ b/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
@@ -1,11 +1,5 @@
-import React, { ComponentType, ReactElement, useCallback, useMemo, useState } from 'react';
-import {
-    debounce,
-    Select,
-    SelectOption,
-    SelectOptionProps,
-    ValidatedOptions,
-} from '@patternfly/react-core';
+import React, { ReactElement, ReactNode, useCallback, useMemo, useState } from 'react';
+import { debounce, Select, SelectOption, ValidatedOptions } from '@patternfly/react-core';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import useRestQuery from 'Containers/Dashboard/hooks/useRestQuery';
 import { CancellableRequest } from 'services/cancellationUtils';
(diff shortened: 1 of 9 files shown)


dvail:
* master
        * **PR #3676** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3676" target="_blank">Graphite</a>  👈
        * **PR #3644** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3644" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3676?utm_source=stack-comment).

ui/apps/platform/src/Containers/Dashboard/hooks/useRestQuery.ts
```diff
@@ -19,6 +19,7 @@ export default function useRestQuery<ReturnType>(
         const { request, cancel } = cancellableRequestFn();
 
         setError(undefined);
+        setLoading(true);
```
dvail, line 22: 🙀

bradr5:
Would be nice if there was a way to easily compare what was in `parser.ts` to `converter.ts` from within here

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
```diff
@@ -49,12 +47,14 @@ function RuleSelector({
         }
 
         const selectorMap: Record<RuleSelectorOption, ScopedResourceSelector> = {
-            All: {},
+            All: { type: 'All' },
             ByName: {
+                type: 'ByName',
```
bradr5, line 52: Not asking you to change anything but was curious of your opinion on having these 3 types (All, ByName, ByLabel) defined as constants?

bradr5:
I can see why you avoided context here, the benefit doesn't feel like it would outweigh the complexity here.

bradr5 (approved):

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
```diff
@@ -10,9 +14,33 @@ export type AutoCompleteSelectProps = {
     onChange: (value: string) => void;
     validated: ValidatedOptions;
     isDisabled: boolean;
+    autocompleteProvider?: (search: string) => CancellableRequest<string[]>;
+    entityType: SelectorEntityType;
```
pedrottimark, line 18: If the only reason for `entityType` prop is for `ResourceIcon` element, then instead of context versus props, the question is cost in code complexity versus benefit in user experience. Besides hierarchical organization deployments, namespaces, clusters, and form element labels, it seems likely that naming conventions for the 3 entity types are quite different.

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -216,13 +216,13 @@ export function getCollectionAutoComplete(
...
 ): CancellableRequest<string[]> {
-    const params = qs.stringify(
-        { resourceSelectors, searchCategory, searchLabel },
-        { arrayFormat: 'repeat' }
-    );
     return makeCancellableAxiosRequest((signal) =>
         axios
-            .get<{ values: string[] }>(`${collectionsAutocompleteUrl}?${params}`, { signal })
+            .post<{ values: string[] }>(
+                `${collectionsAutocompleteUrl}`,
```
pedrottimark, line 222: Not a blocker to merge, does change in params let you replace template literal with variable?

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
pedrottimark: On the positive side of the coin, bravo for composition via `autocompleteProvider` prop.

ui/apps/platform/src/Containers/Collections/RuleSelector/ByNameSelector.tsx
```diff
@@ -77,6 +93,8 @@ function ByNameSelector({
                                     : ValidatedOptions.default
                             }
                             isDisabled={isDisabled}
+                            autocompleteProvider={onAutocomplete}
```
pedrottimark, line 96: A question for understanding, not necessarily a request for change. The difference between prop name and prop value function name.
* Leftover from initial draft of code?
* Intended to communicate meaning to reader?

pedrottimark (approved):

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
dvail: This is something I go back and forth on, but lately I've been just writing the string directly in cases where the code is well-typed enough that a mistyped string would result in a compile time error. This might be worth bringing up during one of the weekly calls though to see if the team has a strong preference one way or another.

My initial feeling is the the pros/cons for using a constant here are:
**Pros** 
- changing the property name only requires a change in one place
- easier to Ctrl+Click to see where the relevant constants are coming from

**Cons**
- more verbose (subjective)
- changing the expected value _doesn't_ result in compile errors at each place the string is used. I could see an argument for this being a trail of breadcrumbs to get eyes on relevant code during a refactor.

dvail: Ah - and didn't realize there were too many changes in the parser for git to detect it as a rename instead of an entirely new file... I can swap the name back if we think it is important to retain that history at this point.

ui/apps/platform/src/services/CollectionsService.ts
dvail: +1 Hmm it seems like this is something prettier would automatically fix. I'll push an update.

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
dvail: > If the only reason for entityType prop is for ResourceIcon element, then instead of context versus props, the question is cost in code complexity versus benefit in user experience. Besides hierarchical organization deployments, namespaces, clusters, and form element labels, it seems likely that naming conventions for the 3 entity types are quite different.

I'm not quite sure I understand.

ui/apps/platform/src/Containers/Collections/RuleSelector/ByNameSelector.tsx
dvail: A little bit of both, but probably more of the former. This was something in the back of my head but since you called it out I think it is worth fixing now. I'll update all to `autocompleteProvider`, since `onAutocomplete` implies the function is a callback, and this is being passed to a declarative data fetching API.

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
pedrottimark: * The component did not need `entityType` prop to fetch the data, because encapsulated in `autocompleteProvider` function, correct?
* Unless my eyes missed something in the changed lines, the reason to add `entityType` prop is to render different resource icon?

Maybe I misunderstood the comment the the description about prop drilling;

> now all need access to the `entityType`

If this is an example which motivated the comment, then my counter question is instead of how to deal with increasing coupling between components, whether this change is pushing beyond a limit of how much coupling is too costly.

It looks like the the component was potentially reusable to become independent of container.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
pedrottimark: TypeScript string enumeration or discriminated union seem fluent as literal strings.

pedrottimark: If we felt the need to encapsulate, then Redux action creator **functions** is a precedent to adapt.

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
dvail: Ah ok got it, yeah good point. Early on I was thinking the `RuleSelector` component would have been a good candidate to extract to `Components/PatternFly` if it had uses outside of this container. I think a lot of the tight coupling exists across everything under the `./RuleSelector` directory since they were originally all one component that I split apart for readability.

I think we can make this reusable again if we need to by pulling the autocomplete function up, and maybe making the error pieces less Formik specific.

pedrottimark: What caught my eye is asymmetry between `entityType` encapsulated for data but not for rendering.

If such specific rendering is impossible to avoid, then a more symmetrical solution is prop for option rendering callback function.

PR #3678, opened 2022-11-03
ROX-13170: Adding table row actions for flows table

sachaudh:
## Description

This PR adds the table row actions for the flows table. This is an iterative addition. You can find all the subtasks here: https://issues.redhat.com/browse/ROX-12904

## Checklist

### Add flow to baseline
Screenshot 2022-11-03 at 10 13 11 AM

### Mark flow as anomalous
Screenshot 2022-11-03 at 10 13 15 AM


vjwilson (approved):
(I think you put the same screenshot in for both states, but code looks right.)

sachaudh:
> (I think you put the same screenshot in for both states, but code looks right.)

oops

PR #3679, opened 2022-11-03
Dashrews/ROX-13082 UUID searcher and common updates to set allow use of postgres UUID PR 1 of 4

dashrews78:
## Description

These are the updates to the searcher to allow for it to handle columns that are UUIDs in postgres but still strings in go since the protobufs don't handle UUID.  Testing will be taken care of it the subsequent PRs and I will continue to add tests.  I want to start getting some reviews.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/postgres/pgutils/utils.go
+++ b/pkg/postgres/pgutils/utils.go
@@ -63,7 +63,7 @@ func NilOrTime(t *types.Timestamp) *time.Time {
 	return &ts
 }
 
-// NilOrUUID allows for a proto timestamp to be stored a timestamp type in Postgres
+// NilOrUUID allows for a proto string to be stored as a UUID type in Postgres
 func NilOrUUID(value string) *uuid.UUID {
 	if value == "" {
 		return nil
--- a/pkg/postgres/pgutils/utils.go
+++ b/pkg/postgres/pgutils/utils.go
@@ -13,6 +13,7 @@ import (
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/postgres"
 	"github.com/stackrox/rox/pkg/utils"
+	"github.com/stackrox/rox/pkg/uuid"
 	"gorm.io/gorm"
 	"gorm.io/gorm/schema"
 )
@@ -60,6 +61,18 @@ func NilOrTime(t *types.Timestamp) *time.Time {
 		return nil
 	}
 	return &ts
+}
+
+// NilOrUUID allows for a proto timestamp to be stored a timestamp type in Postgres
+func NilOrUUID(value string) *uuid.UUID {
+	if value == "" {
+		return nil
+	}
+	id, err := uuid.FromString(value)
+	if err != nil {
+		return nil
+	}
+	return &id
 }
 
 // CreateTableFromModel executes input create statement using the input connection.
--- a/pkg/search/postgres/query/uuid_query.go
+++ b/pkg/search/postgres/query/uuid_query.go
@@ -44,7 +44,7 @@ func newUUIDQueryWhereClause(columnName string, value string, queryModifiers ...
 			Query:  fmt.Sprintf("%s = $$", columnName),
 			Values: []interface{}{uuidVal},
 			equivalentGoFunc: func(foundValue interface{}) bool {
-				return strings.HasPrefix(foundValue.(string), value)
+				return strings.EqualFold(foundValue.(string), value)
 			},
 		}, nil
 	}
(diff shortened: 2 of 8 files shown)


rhybrillou (approved):

pkg/postgres/pgutils/utils.go
```diff
@@ -62,6 +63,18 @@ func NilOrTime(t *types.Timestamp) *time.Time {
 	return &ts
 }
 
+// NilOrUUID allows for a proto timestamp to be stored a timestamp type in Postgres
```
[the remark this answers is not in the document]
janisz: I'm not sure if `nil` is good to have in our system. How about 0 uuid?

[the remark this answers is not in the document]
dashrews78: @janisz I went with nil because I wanted a postgres error if we tried to stick an empty UUID in there.  If we go with a 0 uuid, then we have to add logic to ensure we aren't using a 0 uuid in a required field.  We also have the case of nullable columns where we may in fact want a null.  So allowing for a null to go into those columns seemed simpler.  That was my thought process.  I'll gladly discuss further if you think that is necessary.

pkg/search/query_builder.go
```diff
@@ -263,6 +263,14 @@ func (qb *QueryBuilder) AddExactMatches(k FieldLabel, values ...string) *QueryBu
 	return qb
 }
 
+// AddExactMatchesUUID adds a key value pair to the query
+func (qb *QueryBuilder) AddExactMatchesUUID(k FieldLabel, values ...string) *QueryBuilder {
```
c-du, line 267: If there is no different between this and the AddExactMatches, you can just call it internally, just to have a name difference.

pkg/search/postgres/query/uuid_query.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(queryModifiers) == 0 {
+		uuidVal, err := uuid.FromString(value)
+		if err != nil {
+			return WhereClause{}, errors.Wrapf(err, "value %q in search query must be valid UUID", value)
+		}
+		return WhereClause{
+			Query:  fmt.Sprintf("%s = $$", columnName),
+			Values: []interface{}{uuidVal},
+			equivalentGoFunc: func(foundValue interface{}) bool {
+				return strings.HasPrefix(foundValue.(string), value)
```
c-du, line 47: Why using HasPrefix? It does not sound right for uuid to have a partial match.

pkg/search/postgres/query/uuid_query.go
```diff
@@ -0,0 +1,109 @@
...
+			uuidVal, err := uuid.FromString(value)
+			if err != nil {
+				return WhereClause{}, errors.Wrapf(err, "value %q in search query must be valid UUID", value)
+			}
+			return WhereClause{
+				Query:  fmt.Sprintf("%s != $$", columnName),
+				Values: []interface{}{uuidVal},
+				equivalentGoFunc: func(foundValue interface{}) bool {
+					foundVal := strings.ToLower(foundValue.(string))
+					return !strings.HasPrefix(foundVal, value)
```
c-du, line 68: Dido.

c-du:
Overall looks good to me. Left some comments.

pkg/search/query_builder.go
dashrews78: That is a left over artifact from earlier work.  It wasn't being used at all.  So removed.

c-du (approved):

pkg/search/postgres/query/uuid_query.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	switch queryModifiers[0] {
+	case pkgSearch.Regex:
+		re, err := regexp.Compile(value)
+		if err != nil {
+			return WhereClause{}, fmt.Errorf("invalid regexp %s: %w", value, err)
+		}
+
+		if value != "*" {
```
md2119, line 81: I might be missing some context but why only `*` is supported and not other regexes, for example, `abc.*`?

dashrews78: @md2119 I didn't think that made a lot of sense for a UUID.  What would be the use case for that since UUIDs are random?  There is no logical grouping of say all UUIDs that start with `abc*`.  If we need to cover that, I can certainly add it.

md2119: This is a breaking change given that we currently support any regex on the string fields which are planned to be converted uuid. At the very least we should call it out in the docs.

dashrews78: @md2119 got it.  Probably easy enough to just support.  I will take a look at that capability in with PR https://github.com/stackrox/stackrox/pull/3698

PR #3680, opened 2022-11-03
Cut rc query update, cloud service and allow missing release

davdhacs:
## Description

* Allow 4XX fails on jira query for fixRelease issues.
* Change to jira component name "ACS Cloud Service"

## Checklist

## Testing Performed

Tested script against 3.72.2 and existing releases. For existing components and fixVersion releases, the request completes (curl exit code 0) and for missing fixVersion it return a 400 http error (curl exit code 22), but the script does not return as failed.


.github/workflows/scripts/check-jira-issues.sh
```diff
@@ -8,42 +8,53 @@ set -euo pipefail
...
 
 get_issues() {
+    local status=0
     curl --fail -sSL --get --data-urlencode "jql=$1" \
         -H "Authorization: Bearer $JIRA_TOKEN" \
         -H "Accept: application/json" \
-        "https://issues.redhat.com/rest/api/2/search"
+        "https://issues.redhat.com/rest/api/2/search" || status="$?"
+    if [[ "$status" == 22 ]]; then  # curl man page: "22     HTTP  page  not retrieved. .. HTTP error code being 400 or above."
+        gh_log warning "4XX http error code" >&2
```
tommartensen, line 45: This would be masking the failure to find the version on Jira as it is not hard-stopping the pipeline (think release engineer has a typo and cut-rc for 3.89 instead of 3.78).

PR #3685, opened 2022-11-03
Fix AddCluster error with cluster exists

c-du:
## Description
Fix an issue blocking non-groovy test.
```
roxctl --insecure-skip-tls-verify -e 35.188.132.52:443 -p omJnRzC1JyvbsX5Zjz5B1wE84 sensor generate openshift --name os-istio-test-cluster --continue-if-exists --output-dir /tmp/tmp.b78Sol7JBW
[OK] No DestinationRule will be created
[FAIL] Error invoking command
Captured output was:
ERROR:	error creating cluster: rpc error: code = Internal desc = ERROR: duplicate key value violates unique constraint "clusters_name_key" (SQLSTATE 23505)
```
Link: https://srox.slack.com/archives/C030MTTE1S5/p1667498696672839

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.](https://srox.slack.com/archives/C030MTTE1S5/p1667496808272139)

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/cluster/datastore/datastore_impl.go
+++ b/central/cluster/datastore/datastore_impl.go
@@ -314,7 +314,7 @@ func (ds *datastoreImpl) AddCluster(ctx context.Context, cluster *storage.Cluste
 	}
 	_, found := ds.nameToIDCache.Get(cluster.Name)
 	if found {
-		return "", errox.AlreadyExists.Newf("cannot add cluster %s, it already exists", cluster.GetName())
+		return "", errox.AlreadyExists.Newf("the cluster with name %s exists, cannot re-add it", cluster.GetName())
 	}
 
 	ds.lock.Lock()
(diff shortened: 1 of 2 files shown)


central/cluster/datastore/datastore_impl.go
```diff
@@ -312,6 +312,10 @@ func (ds *datastoreImpl) AddCluster(ctx context.Context, cluster *storage.Cluste
 	if err := checkWriteSac(ctx, cluster.GetId()); err != nil {
 		return "", err
 	}
+	_, found := ds.nameToIDCache.Get(cluster.Name)
+	if found {
+		return "", errox.AlreadyExists.Newf("cannot add cluster %s, it already exists", cluster.GetName())
```
md2119, line 317: I think the message could be clearer because It is the cluster name that is conflicting and not the cluster config. maybe something like `cluster name must be unique. Cannot add cluster {name} because a cluster {stored cluster name} ({id}) already exists`

md2119 (approved):
one comment on the messaging.

central/cluster/datastore/datastore_impl.go
c-du: Updated the message.

PR #3688, opened 2022-11-03
ROX-13068: Use real data for deployment details

sachaudh:
## Description

This PR adds real data for the deployment details page. This is an iterative addition. You can find all the subtasks here:
https://issues.redhat.com/browse/ROX-12903

NOTE: I decided to break this down further. I will add real data for the network policy section and port config section of the details page in a separate PR

## Checklist

## Testing Performed

Screenshot 2022-11-03 at 4 02 15 PM
Screenshot 2022-11-03 at 4 02 18 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -26,6 +26,7 @@ import CidrBlockSideBar from './cidr/CidrBlockSideBar';
 import ExternalEntitiesSideBar from './external/ExternalEntitiesSideBar';
 
 import './Topology.css';
+import { getNodeById } from './utils/networkGraphUtils';
 
 // TODO: move these type defs to a central location
 export const UrlDetailType = {
@@ -37,11 +38,6 @@ export const UrlDetailType = {
 } as const;
 export type UrlDetailTypeKey = keyof typeof UrlDetailType;
 export type UrlDetailTypeValue = typeof UrlDetailType[UrlDetailTypeKey];
-
-function findEntityById(graphModel: Model, id: string | undefined): NodeModel | undefined {
-    const entity = graphModel.nodes?.find((node: { id: string }) => node.id === id);
-    return entity;
-}
 
 function getUrlParamsForEntity(selectedEntity: NodeModel): [UrlDetailTypeValue, string] {
     const detailType = UrlDetailType[selectedEntity.data.type];
@@ -95,7 +91,7 @@ function setEdges(controller, detailId) {
 const TopologyComponent = ({ model }: TopologyComponentProps) => {
     const history = useHistory();
     const { detailId } = useParams();
-    const selectedEntity = detailId && findEntityById(model, detailId);
+    const selectedEntity = detailId && getNodeById(model, detailId);
     const controller = useVisualizationController();
 
     // to prevent error where graph hasn't initialized yet
@@ -109,7 +105,7 @@ const TopologyComponent = ({ model }: TopologyComponentProps) => {
 
     function onSelect(ids: string[]) {
         const newSelectedId = ids?.[0] || '';
-        const newSelectedEntity = findEntityById(model, newSelectedId);
+        const newSelectedEntity = getNodeById(model, newSelectedId);
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         if (newSelectedEntity) {
@@ -132,6 +128,7 @@ const TopologyComponent = ({ model }: TopologyComponentProps) => {
         return () => {
             controller.removeEventListener(SELECTION_EVENT, onSelect);
         };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [controller, model]);
 
     const selectedIds = selectedEntity ? [selectedEntity.id] : [];
@@ -144,7 +141,11 @@ const TopologyComponent = ({ model }: TopologyComponentProps) => {
                         <NamespaceSideBar />
                     )}
                     {selectedEntity && selectedEntity?.data?.type === 'DEPLOYMENT' && (
-                        <DeploymentSideBar />
+                        <DeploymentSideBar
+                            deploymentId={selectedEntity.id}
+                            nodes={model?.nodes || []}
+                            edges={model?.edges || []}
+                        />
                     )}
                     {selectedEntity && selectedEntity?.data?.type === 'CIDR_BLOCK' && (
                         <CidrBlockSideBar />
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
@@ -1,8 +1,12 @@
 import React from 'react';
 import {
+    Alert,
+    AlertVariant,
     Badge,
+    Bullseye,
     Flex,
     FlexItem,
+    Spinner,
     Stack,
     StackItem,
     Tab,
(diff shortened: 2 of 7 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -98,7 +98,7 @@ const TopologyComponent = ({
                         <NamespaceSideBar />
                     )}
                     {selectedEntity && selectedEntity?.data?.type === 'DEPLOYMENT' && (
-                        <DeploymentSideBar />
+                        <DeploymentSideBar deploymentId={selectedEntity.id} model={model} />
```
alwayshooin, line 101: why not pass the actual entity data through here instead of passing through the entire data model and the relevant id?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -15,15 +15,44 @@ import {
...
 import DeploymentFlows from './DeploymentFlows';
+import {
+    getListenPorts,
+    getNumExternalFlows,
+    getNumInternalFlows,
+} from '../utils/networkGraphUtils';
+
+type DeploymentSideBarProps = {
+    deploymentId: string;
+    model: Model;
```
alwayshooin, line 32: do we need the whole model here or can we just pass through the `NodeModel` of this deployment instead?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -34,15 +63,15 @@ function DeploymentSideBar() {
                     <FlexItem>
                         <TextContent>
                             <Text component={TextVariants.h1} className="pf-u-font-size-xl">
-                                visa-processor
+                                {deployment?.name}
```
alwayshooin, line 66: shouldn't we have this all the time via the model?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -34,15 +63,15 @@ function DeploymentSideBar() {
...
                             </Text>
                         </TextContent>
                         <TextContent>
                             <Text
                                 component={TextVariants.h2}
                                 className="pf-u-font-size-sm pf-u-color-200"
                             >
-                                in &quot;production / naples&quot;
+                                in &quot;{deployment?.clusterName} / {deployment?.namespace}&quot;
```
alwayshooin, line 74: same here as above

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+                return [...acc, curr.id];
+            }
+            return acc;
+        }, [] as string[]) || [];
+    return externalNodeIds;
+}
+
+export function getNodeById(model: Model, nodeId: string): NodeModel | undefined {
+    return model.nodes?.find((node) => node.id === nodeId);
+}
```
alwayshooin, line 20: idt we need this because we can use `controller.getNodeById(id)`

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+        }, 0) || 0;
+    return numExternalFlows;
+}
+
+/* deployment helper functions */
+
+export function getListenPorts(model: Model, deploymentId: string): ListenPort[] {
+    const deployment = model.nodes?.find((node) => {
+        return node.id === deploymentId;
+    });
```
alwayshooin, line 59: i think you can use `controller.getNodeById(id)` here too

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+            if (
+                (edge.source === deploymentId && externalNodeIds.includes(edge.target || '')) ||
+                (edge.target === deploymentId && externalNodeIds.includes(edge.source || ''))
+            ) {
+                return acc + 1;
+            }
+            return acc;
+        }, 0) || 0;
+    return numExternalFlows;
+}
```
alwayshooin, line 52: alternatively for both the internal and external flows here we could store this in the `data` for the node when we construct the model, or maybe create an edge map and use `selectedNode.getSourceEdges()` and `selectedNode.getTargetEdges()` from the `NodeModel` to get the edges of the node and look up whether an edge is external or not. not sure what would be the best option here, just throwing out some other suggestions!

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
```diff
@@ -0,0 +1,64 @@
+import { Model, NodeModel } from '@patternfly/react-topology';
```
alwayshooin, line 1: should we move the other `utils.tsx` file to this directory as well and rename it to be more specific like `modelUtils.tsx` or something?

ui/apps/platform/src/Containers/NetworkGraph/utils.tsx
```diff
@@ -29,15 +29,19 @@ export function transformData(nodes: Node[]): Model {
...
             type: 'node',
             width: 75,
             height: 75,
             label: getLabel(entity),
-            data: entity,
+            // @TODO: create a consistent data structure for "data" between all nodes
+            data: {
+                ...entity,
+                policyIds,
+            },
```
alwayshooin, line 44: this is definitely something we should discuss further

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: I decided to pass the whole model because we need access to various data in this tab:  deployment id, policy ids (from node), and edges from/to deployment. We could pass just what we need, but I thought it would be better to give the model now, and then if we want to refactor and make things more optimal, we can do that separately

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
sachaudh: The `deployment` value actually comes from the API to get the deployment details so that's why it could be undefined until the API returns a response. You're right though, we could find the deployment name through the selected node in the model, but I thought it made sense to use the `deployment` value. We can get around this by having a loading state and empty state

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
sachaudh: Ah but then I'd have to pass the `controller` or use the hook (`useVisualizationController`). I'd have to rethink the flow of data here. I might have to have the `DeploymentSideBar` use the `controller` and prepare the data to be passed to `DeploymentDetails` in a different way. Are there more helper functions that the `controller` offers?

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
sachaudh: I'm hesitant to add it to the node because I think having derived data from a single source is a safer option. If we had filters that hide certain edges or modify the graph in some way, wouldn't that number not be accurate anymore if we go by what we see visually? 

That's also worth noting. Is it better to:
1) precompute the value and save it whenever the model changes or
2) compute it only when the user goes to that view

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
sachaudh: Yea we can do that

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
alwayshooin: oh i see. yeah i think it makes sense to use both the graph deployment data and the deployment API response since we're using both for the side panel (at least for the flows tab) anyways

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
alwayshooin: thats true. we should see if it makes more or less sense to pass the controller like that around. somewhat, you can check the API doc i made (https://docs.google.com/document/d/1j0P1cqOJ69VWDOlpemSejlkL_ya99suSp1IvgqojeGA/edit#heading=h.v6s7p22e33i8) -- there are also additional functions for each node/edge element defined in the library too

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.ts
alwayshooin: hm i suppose compute on when the user goes to the view is probably more efficient/performant overall at scale. i guess we can always optimize later too

alwayshooin (approved):

PR #3689, opened 2022-11-04
[ROX-13403] : Fix node -> topVuln sub resolver bug when node cves is empty

charmik-redhat:
## Description

Issue happens when postgres is disabled

When there are no node cves, the topVuln sub resolver passes a nil query to cveDatastore, which results in datastore returning all cves including image and platform cves. The fix avoids querying cveDatastore and returns topVuln value as nil when there are no cves in a node.

## Checklist
- [ ] Investigated and inspected CI test results~

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/graphql/resolvers/node_vulnerabilities_test.go
+++ b/central/graphql/resolvers/node_vulnerabilities_test.go
@@ -1,6 +1,3 @@
-//go:build sql_integration
-// +build sql_integration
-
 package resolvers
 
 import (
@@ -18,7 +15,7 @@ import (
 	nodeCVESearch "github.com/stackrox/rox/central/cve/node/datastore/search"
 	nodeCVEPostgres "github.com/stackrox/rox/central/cve/node/datastore/store/postgres"
 	"github.com/stackrox/rox/central/graphql/resolvers/loaders"
-	nodeDackboxDataStore "github.com/stackrox/rox/central/node/datastore/dackbox/datastore"
+	nodeDS "github.com/stackrox/rox/central/node/datastore/dackbox/datastore"
 	nodeGlobalDataStore "github.com/stackrox/rox/central/node/datastore/dackbox/globaldatastore"
 	nodeSearch "github.com/stackrox/rox/central/node/datastore/search"
 	nodePostgres "github.com/stackrox/rox/central/node/datastore/store/postgres"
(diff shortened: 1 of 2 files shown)


md2119 (changes requested):
Apply the same changes to `getTopNodeCVEV1Query` and similarly for images

charmik-redhat:
> Apply the same changes to `getTopNodeCVEV1Query` and similarly for images

Currently the check in `getTopNodeCVEV1Query ` is
```
if resolver.data.GetSetTopCvss() == nil {
    return nil, nil
}
```
I can change the check to `if resolver.data.GetTopCvss() == 0` instead. But I am not sure if there exist a CVE that has Cvss score 0. If there is such a CVE, then we would want to return that CVE and not nil. The check if `resolver.data.GetSetTopCvss() == nil` does seem more accurate in that case. So I am keeping `getTopNodeCVEV1Query ` the same and at the caller level, just adding an extra check for nil query.

Also, I do not think we need to make any changes to images since topVuln resolver in images does not call a `getTopImageCVEQuery` function. They directly return nil cveResolver  when `GetSetTopCvss() == nil` .

md2119:
> > Apply the same changes to `getTopNodeCVEV1Query` and similarly for images
> 
> Currently the check in `getTopNodeCVEV1Query ` is
> 
> ```
> if resolver.data.GetSetTopCvss() == nil {
>     return nil, nil
> }
> ```
> 
> I can change the check to `if resolver.data.GetTopCvss() == 0` instead. But I am not sure if there exist a CVE that has Cvss score 0. If there is such a CVE, then we would want to return that CVE and not nil. The check if `resolver.data.GetSetTopCvss() == nil` does seem more accurate in that case. So I am keeping `getTopNodeCVEV1Query ` the same and at the caller level, just adding an extra check for nil query.
> 
> Also, I do not think we need to make any changes to images since topVuln resolver in images does not call a `getTopImageCVEQuery` function. They directly return nil cveResolver when `GetSetTopCvss() == nil` .



> Member

Yes, CVEs may have a score of 0. I am confused by the flow. Evaluation of the `args RawQuery` must happen after knowing the fact that there are indeed a non-zero number of cves. Recommend moving the following check out of the function `topVulnV2`.
```
if resolver.data.GetSetTopCvss() == nil {
		return nil, nil
	}
```

md2119 (approved):

PR #3695, opened 2022-11-04
ROX-13356: Use deployment ID reference when processing deployments

fredrb:
Depends on #3653 

## Description

This will be used by a future component that can find deployments that need processing based on a `DeploymentReference` implementation. 

## Checklist
- [ ] Run tests locally with feature flag enabled

## Testing Performed

This is hidden behind a feature flag, and for now it's not supposed to work as is, because nothing consumes the `DeploymentReference` property.

### DIFF
--- a/sensor/common/store/resolver/resolver.go
+++ b/sensor/common/store/resolver/resolver.go
@@ -1,0 +1,12 @@
+package resolver
+
+import "github.com/stackrox/rox/sensor/common/store"
+
+type DeploymentReference func(store store.DeploymentStore) []string
+
+// ResolveDeploymentIds is an identify function that simply returns a list of deployment ids passed
+func ResolveDeploymentIds(ids ...string) DeploymentReference {
+	return func(_ store.DeploymentStore) []string {
+		return ids
+	}
+}
--- a/sensor/common/store/resolver/resolver.go
+++ b/sensor/common/store/resolver/resolver.go
@@ -2,6 +2,7 @@ package resolver
 
 import "github.com/stackrox/rox/sensor/common/store"
 
+// DeploymentReference generates a list of deployment IDs that need to be updated given the deployment store.
 type DeploymentReference func(store store.DeploymentStore) []string
 
 // ResolveDeploymentIds is an identify function that simply returns a list of deployment ids passed
--- a/sensor/kubernetes/listener/resources/deployments.go
+++ b/sensor/kubernetes/listener/resources/deployments.go
@@ -195,7 +195,7 @@ func (d *deploymentHandler) processWithType(obj, oldObj interface{}, action cent
 			// if resource is being removed, we can create the remove message here without related resources
 			events = component.MergeResourceEvents(events, component.NewResourceEvent([]*central.SensorEvent{deploymentWrap.toEvent(action)}, nil, nil))
 		} else {
-			// If re-sync is disabled, we need don't process deployment relationships here. We pass a deployment
+			// If re-sync is disabled, we don't need to process deployment relationships here. We pass a deployment
 			// references up the chain, which will be used to trigger the actual deployment event and detection.
 			events = component.MergeResourceEvents(events,
 				component.NewDeploymentRefEvent(resolver.ResolveDeploymentIds(deploymentWrap.GetId())))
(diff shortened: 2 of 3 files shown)


sensor/kubernetes/listener/resources/deployments.go
```diff
@@ -180,19 +177,42 @@ func (d *deploymentHandler) processWithType(obj, oldObj interface{}, action cent
...
-			Object: deploymentWrap.GetDeployment(),
-			Action: action,
-		},
-	}, nil)
 
-	events = component.MergeResourceEvents(events, outputMessage)
+	if features.ResyncDisabled.Enabled() {
+		if action == central.ResourceAction_REMOVE_RESOURCE {
+			// if resource is being removed, we can create the remove message here without related resources
+			events = component.MergeResourceEvents(events, component.NewResourceEvent([]*central.SensorEvent{deploymentWrap.toEvent(action)}, nil, nil))
```
lvalerom, line 186: I think we need to append the ID here too. The detector needs to process the `remove events` in order to remove them from the deduper and send the empty alerts message to mark all previous alerts as stale.

sensor/common/store/resolver/resolver.go
```diff
@@ -0,0 +1,12 @@
+package resolver
+
+import "github.com/stackrox/rox/sensor/common/store"
+
+type DeploymentReference func(store store.DeploymentStore) []string
+
+// ResolveDeploymentIds is an identify function that simply returns a list of deployment ids passed
+func ResolveDeploymentIds(ids ...string) DeploymentReference {
```
lvalerom, line 8: I think we need to add the `action` somewhere. I'm not sure if here would be a good idea or maybe as a field in the `ResourceEvent`. The reason why is because the detector relies on the action to do different things and if I understood the code correctly we will send a `ResourceEvent` with nothing but the `DeploymentReference` to the detector, right?

lvalerom (changes requested):
Overall I think it looks good. I'm just blocking because I think we need to pass the `action` to the detector as well.

sensor/common/store/resolver/resolver.go
fredrb: Addressed this in a follow-up PR. If you don't mind, I'd keep it there only: https://github.com/stackrox/stackrox/pull/3853

sensor/kubernetes/listener/resources/deployments.go
fredrb: Fair enough. But I think we should use the `CompatibilityDetectionMessage` property for this. 

If we pass the deployment reference, it's a problem if the deployment was deleted from the internal store at the end of the handler. Then the resolver component won't be able to trigger the detection for a deployment that no longer exists.  

I think a definite solution for this should do one of the following:

1) When it's a remove event, copy the deployment object (doesn't matter if there are no related resources), send the entire object to the pipeline. This object is forwarded to the detection and ignored by the dependency resolution step. 
2) The detection package receives "delete" operations based on IDs. Not sure how possible is that, but if it's just removing from the deduper and sending an empty alert, it should be ok. 

My proposal is to do (1) for now, using the compatibility property. If we stick to it, this property will no longer be there only for compatibility reasons since we have a legitimate reason to send a copy of the object over the pipeline. 

If you're ok with (1), I'll update the code.

fredrb: By the way, there should be a test that fails if the empty alerts are not sent for a deployment that was deleted. 😞

sensor/kubernetes/listener/resources/deployments.go
```diff
@@ -180,19 +177,52 @@ func (d *deploymentHandler) processWithType(obj, oldObj interface{}, action cent
...
+			events = component.MergeResourceEvents(events, component.NewResourceEvent(nil, []component.CompatibilityDetectionMessage{
+				{
+					Object: deploymentWrap.GetDeployment(),
+					Action: action,
+				},
+			}, nil))
+			// if resource is being removed, we can create the remove message here without related resources
+			events = component.MergeResourceEvents(events, component.NewResourceEvent([]*central.SensorEvent{deploymentWrap.toEvent(action)}, nil, nil))
+		} else {
+			// If re-sync is disabled, we need don't process deployment relationships here. We pass a deployment
```
lvalerom, line 198:
```suggestion
			// If re-sync is disabled, we don't need to process deployment relationships here. We pass a deployment
```
[nit]

lvalerom (approved):
LGTM 🚀

PR #3699, opened 2022-11-04
ROX-13368: Skip failing nongroovy tests

janisz:
## Description

Let's skip failing non groovy tests on postgres and enable them later.

## Tests

To make if faster I created a test branch with PG enabled on nongroovy test
https://github.com/stackrox/stackrox/pull/3700


c-du (approved):
Please file jiras to track the fix for each test.

PR #3700, opened 2022-11-04
WIP: Rox 13368/skip failing test on pg enable tests

janisz:
This PR is only to run nongroovy test on postgres


janisz:
https://github.com/stackrox/stackrox/pull/3699 was merged

PR #3702, opened 2022-11-04
cut.rc and upgrade-test to use existing quay secrets

davdhacs:
Also used in build.yaml: https://github.com/stackrox/stackrox/blob/master/.github/workflows/build.yaml#L157
Confirmed working in run log:
https://github.com/stackrox/stackrox/actions/runs/3230995702/jobs/5290193019#step:13:2


gavin-stackrox (approved):
LGTM!

PR #3703, opened 2022-11-04
ROX-13427: Fix network graph side panel height issue

sachaudh:
## Description

This PR fixes the height issue of the network graph side panel. The tab content will now scroll instead of the whole section.

## Checklist

## Testing Performed

https://user-images.githubusercontent.com/4805485/200031732-b4553f3d-0f41-4905-901b-fda6c19cbd1e.mov

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
@@ -4,3 +4,9 @@
     --pf-c-page__main-section--PaddingLeft: 0;
     --pf-c-page__main-section--PaddingRight: 0;
 }
+
+.pf-topology-resizable-side-bar {
+    /* we use the page header min height twice to account for: page header + nav header */ 
+    /* @TODO: Consider a more future-proof way of doing this. See https://github.com/stackrox/stackrox/pull/3703#discussion_r1014292340 */
+    height: calc(100vh - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight));
+}
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
@@ -3,6 +3,8 @@ import {
     Badge,
     Flex,
     FlexItem,
+    Stack,
+    StackItem,
     Tab,
     TabContent,
     Tabs,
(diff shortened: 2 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -22,28 +24,30 @@ function DeploymentSideBar() {
...
-                            component={TextVariants.h2}
-                            className="pf-u-font-size-sm pf-u-color-200"
-                        >
-                            in &quot;production / naples&quot;
-                        </Text>
-                    </TextContent>
-                </FlexItem>
-            </Flex>
-            <FlexItem flex={{ default: 'flex_1' }}>
+        <Stack>
```
sachaudh, line 27: Instead of `Flex`, it seemed simpler to use a `Stack`

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
```diff
@@ -4,3 +4,8 @@
     --pf-c-page__main-section--PaddingLeft: 0;
     --pf-c-page__main-section--PaddingRight: 0;
 }
+
+.pf-topology-resizable-side-bar {
+    /* 155.2px is the height of the page header + nav header */ 
+    height: calc(100vh - 155.2px);
```
sachaudh, line 10: Open to other suggestions, but seemed like this helped with the scrolling issue

vjwilson: I wonder if we could use the PF CCS variables in the calculation for future-proofing?

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
```diff
@@ -4,3 +4,8 @@
...
+}
```
vjwilson, line 11: You may want to turn on **Insert Final Newline** in your editor.

Screen Shot 2022-11-04 at 1 29 24 PM

vjwilson (approved):

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
sachaudh: Looks like we can use `var(--pf-c-page__header--MinHeight)` and multiply that by two since we have two levels of headers (nav header + page header). I did notice, however, that it did this


https://user-images.githubusercontent.com/4805485/200041529-b92edf7b-18e1-49f9-b29e-ba14924f88d9.mov

sachaudh: Maybe I can add a TODO for considering a different approach

PR #3712, opened 2022-11-04
Add more logs to DefaultPoliciesTest for debugging "Apache struts : CVE-2017-5638" CI flake

charmik-redhat:
## Description

Added warnings when struts component or CVE-2017-5638 is not present in the scan. Also, the helper function used to get the image scan and log it did not wait to consume the complete output. The PR fixes that too.

## Checklist

## Testing Performed

### DIFF
--- a/qa-tests-backend/src/main/groovy/util/Helpers.groovy
+++ b/qa-tests-backend/src/main/groovy/util/Helpers.groovy
@@ -144,7 +144,7 @@ class Helpers {
             FileWriter sout = new FileWriter(output)
             StringBuilder serr = new StringBuilder()
 
-            proc.consumeProcessOutput(sout, serr)
+            proc.waitForProcessOutput(sout, serr)
             proc.waitFor()
 
             if (proc.exitValue() != 0) {
--- a/qa-tests-backend/src/main/groovy/util/Helpers.groovy
+++ b/qa-tests-backend/src/main/groovy/util/Helpers.groovy
@@ -139,7 +139,7 @@ class Helpers {
             Path imageScans = Paths.get(Constants.FAILURE_DEBUG_DIR).resolve("image-scans")
             new File(imageScans.toAbsolutePath().toString()).mkdirs()
 
-            Process proc = "./scripts/ci/roxctl.sh image scan -i ${image}".execute(null, new File(".."))
+            Process proc = "./scripts/ci/roxctl.sh image scan -i ${image} -a -f".execute(null, new File(".."))
             String output = imageScans.resolve(saveName).toAbsolutePath()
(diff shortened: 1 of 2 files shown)


qa-tests-backend/src/main/groovy/util/Helpers.groovy
```diff
@@ -139,12 +139,12 @@ class Helpers {
             Path imageScans = Paths.get(Constants.FAILURE_DEBUG_DIR).resolve("image-scans")
             new File(imageScans.toAbsolutePath().toString()).mkdirs()
 
-            Process proc = "./scripts/ci/roxctl.sh image scan -i ${image}".execute(null, new File(".."))
+            Process proc = "./scripts/ci/roxctl.sh image scan -i ${image} -a -f".execute(null, new File(".."))
```
theencee, line 142: `-f` forces scanner to rescan whereas without it you'll get whatever was the last scan result. Assuming the image exists in central db, otherwise it'll just get a new scan.

So wouldn't getting what was in central last be more helpful for debugging?

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -187,6 +187,13 @@ class DefaultPoliciesTest extends BaseSpecification {
             log.info "Temporarily enabled policy '${policyName}'"
             policyEnabled = true
         }
+        // debugging to see if the test fails due to incomplete scan
```
theencee, line 190: Add the ticket with a "TODO" comment so this can be found quickly.

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -187,6 +187,13 @@ class DefaultPoliciesTest extends BaseSpecification {
...
+        if (policyName == "Apache Struts: CVE-2017-5638") {
+            def image = ImageService.scanImage("library/nginx:1.10", true)
+            if (!hasApacheStrutsVuln(image)) {
+                log.warn("[Apache struts] CVE-2017-5638 is absent from image scan")
```
theencee, line 194: IMO you should add the JIRA to the log messages to make grepping for it in the OSCI logs much easier.

md2119 (approved):
Second with Nakul on adding tracker info to the logs. Address Nakul's comments before merging.

qa-tests-backend/src/main/groovy/util/Helpers.groovy
charmik-redhat: Yeah, that's right. Since alerts would depend on what was present in the central, the debug info should also print the same. Removed -f option

PR #3715, opened 2022-11-05
CI: Fill the gaps for https://testgrid.k8s.io/

gavin-stackrox:
## Description

Saving Junit XML only on failure results in 'stale' tests from testgrids point if view. This PR changes the deployment breadcrumb check to report on success and failure. It also adds an earlier check for the cluster and that images are available. Both of these tend to be common earlier failure causes that are currently identify as a Stackrox deployment failure.

## Checklist

## Testing Performed

- [x] Happy path: 3x steps in passing tests - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3715/pull-ci-stackrox-stackrox-master-gke-qa-e2e-tests/1588993722295521280
- [x] Provision fails: GKEs fail with provisioning highlighted - rephrased as a pass/fail test statement - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3715/pull-ci-stackrox-stackrox-master-gke-upgrade-tests/1589034424001892352
- [x] Imaging fails: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3715/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests/1589049211557515264
- [x] Deployment fails: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3715/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests/1589070005767507968

### DIFF
--- a/.openshift-ci/runners.py
+++ b/.openshift-ci/runners.py
@@ -4,6 +4,7 @@
 Common test run patterns
 """
 
+import os
 from datetime import datetime
 from clusters import NullCluster
 from pre_tests import NullPreTest
@@ -35,6 +36,7 @@ class ClusterTestSetsRunner:
             self.log_event("About to provision")
             self.cluster.provision()
             self.log_event("provisioned")
+            self.set_provisioned_state()
         except Exception as err:
             self.log_event("ERROR: provision failed")
             hold = err
@@ -101,8 +103,10 @@ class ClusterTestSetsRunner:
                 hold = err
             try:
                 self.log_event("About to run post test", test_set)
-                test_set["post_test"].run(test_outputs=test_set["test"].test_outputs,
-                                          test_results=test_set["test"].test_results)
+                test_set["post_test"].run(
+                    test_outputs=test_set["test"].test_outputs,
+                    test_results=test_set["test"].test_results,
+                )
                 self.log_event("post test completed", test_set)
             except Exception as err:
                 self.log_event("ERROR: post test failed", test_set)
@@ -121,6 +125,9 @@ class ClusterTestSetsRunner:
         print(marker)
         print(f"{marker} {time}: {msg}")
         print(marker)
+
+    def set_provisioned_state(self):
+        os.system("tests/e2e/lib.sh set_provisioned_state")
 
 
 # pylint: disable=too-many-arguments
--- a/.openshift-ci/runners.py
+++ b/.openshift-ci/runners.py
@@ -4,7 +4,7 @@
 Common test run patterns
 """
 
-import os
+import subprocess
 from datetime import datetime
 from clusters import NullCluster
 from pre_tests import NullPreTest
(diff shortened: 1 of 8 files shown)


.openshift-ci/runners.py
```diff
@@ -122,6 +126,9 @@ def log_event(self, msg, test_set=None):
         print(f"{marker} {time}: {msg}")
         print(marker)
 
+    def set_provisioned_state(self):
+        os.system("tests/e2e/lib.sh set_provisioned_state")
```
porridge, line 130: Is the usage of `os.system` on purpose due to the [signal handling differences](https://docs.python.org/3/library/subprocess.html#replacing-os-system)?
If not, please change to `subprocess.call` or perhaps better yet, `subprocess.check_call`.

scripts/ci/lib.sh
```diff
@@ -1348,6 +1351,27 @@ __EOM__
     }
 }
 
+save_junit_success() {
+    if [[ "$#" -ne 2 ]]; then
+        die "missing args. usage: save_junit_success <class> <description>"
+    fi
+
+    if [[ -z "${ARTIFACT_DIR}" ]]; then
+        info "Warning: save_junit_success() requires an ARTIFACT_DIR"
```
porridge, line 1360:
```suggestion
        info "Warning: save_junit_success() requires the \$ARTIFACT_DIR variable to be set"
```

scripts/ci/test_state.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# State files used to track test progress across script invocations
```
porridge, line 3:
```suggestion
# State files used to track test progress across script invocations.
```

scripts/ci/test_state.sh
```diff
@@ -0,0 +1,7 @@
...
+
+export STATE_CLUSTER_PROVISIONED="/tmp/state_cluster_provisioned"
+export STATE_IMAGES_AVAILABLE="/tmp/state_images_available"
+export STATE_DEPLOYED="/tmp/state_deployed"
```
porridge, line 7: Perhaps prefix the filenames with something like `stackrox_ci_`?

porridge (approved):

.openshift-ci/runners.py
```diff
@@ -122,6 +126,9 @@ def log_event(self, msg, test_set=None):
         print(f"{marker} {time}: {msg}")
         print(marker)
 
+    def set_provisioned_state(self):
+        subprocess.check_call("tests/e2e/lib.sh set_provisioned_state", shell=True)
```
BradLugo, line 130: Not going to block on this nit but should we use `common.popen_graceful_kill` here?

BradLugo (approved):
Everything else LGTM

.openshift-ci/runners.py
gavin-stackrox: I don't think it is appropriate in this case i.e. just touching a state file

PR #3681, opened 2022-11-03
Dashrews/ROX-13082 UUID generator templates PR 2 of 4

dashrews78:
## Description
This one builds upon https://github.com/stackrox/stackrox/pull/3679
The focus of this PR is the postgres generation templates.  These changes are slight adjustments to allow for using the UUIDs in Postgres.  Due to this, the test template has an update to the test to fix an occurrence of a hard code string to be a valid hard coded UUID string.  Added a new test proto and updated a couple test protos to utilize the UUID type.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
~- [ ] Evaluated and added CHANGELOG entry if required~
~- [ ] Determined and documented upgrade steps~
~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs]~(https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Added some UUIDs to some of the test protos to ensure some generated stores (and their tests) utilized those

PR #3725, opened 2022-11-07
ROX-13326: Use UUID columns for roles, permissionsets and simpleaccessscopes (PermissionSet and SimpleAccessScope identifiers) in postgres mode.

rhybrillou:
## Description

The relation between roles and the associated permission sets as well as access scopes used to be done relying on decorated UUIDs (generated UUIDs prefixed with human-readable strings).
These object are candidate to conversion to UUIDs.

Some of the UUID base work is needed for this change to work (PRs https://github.com/stackrox/stackrox/pull/3679 and https://github.com/stackrox/stackrox/pull/3681 ).

PR stack:
https://github.com/stackrox/stackrox/pull/3802 (server support code)
+- https://github.com/stackrox/stackrox/pull/3725 (data migration)
... +- https://github.com/stackrox/stackrox/pull/3818 (UI code)

## Checklist

## Testing Performed

CI is a good first step.

Manual testing performed:
- deploy central in non-postgres mode.
- check the existing roles, permission sets and access scopes.
- create a sample permission set, a sample access scopes, and roles covering the possible combinations of default and created permission sets and access scopes.
- open the existing roles, permission sets and access scopes and keep trace of their identifiers.
- deploy postgres central-db.
- switch central to postgres mode.
- monitor for migration steps. These should include migration 148 to 149, 157 to 158 and 163 to 164 (currently latest rocks migration sequence number is 111, and the altered migrations are postgres 37 to 38, 46 to 47 and 52 to 53).
- open the UI
- check the new identifiers. They should all be UUIDs. The identifiers that were prefixed UUIDs before the migration should be the same UUID without prefix.
- check that the inter-object relations are kept (roles referencing permission sets and access scopes as well as referenced permission sets and access scopes).
- connect to the postgres DB
- check the table datamodel (id columns for permission_sets and simple_access_scopes should be UUIDs).

One observation during the manual test was that the `Unrestricted` access scope page failed to open. A pull request has been opened in order to fix the UI part as a consequence (https://github.com/stackrox/stackrox/pull/3818).


central/role/datastore/singleton.go
```diff
@@ -67,23 +67,34 @@ func Singleton() DataStore {
...
 }
 
+func (attributes *roleAttributes) getID() string {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return attributes.postgresID
+	}
+	return rolePkg.EnsureValidPermissionSetID(attributes.idSuffix)
+}
+
 var defaultRoles = map[string]roleAttributes{
```
dhaus67, line 82: nit: might be worthwhile, like we do with the tags in our proto message definition, to create a comment with `// Next ID: xxxxx` so other people have it easier in adding a new default role.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -31,20 +46,70 @@ var (
...
+		"scopemanager":          "ffffffff-ffff-fff4-f5ff-fffffffffffb",
+		"sensorcreator":         "ffffffff-ffff-fff4-f5ff-fffffffffffa",
+		"vulnmgmtapprover":      "ffffffff-ffff-fff4-f5ff-fffffffffff9",
+		"vulnmgmtrequester":     "ffffffff-ffff-fff4-f5ff-fffffffffff8",
+		"vulnreporter":          "ffffffff-ffff-fff4-f5ff-fffffffffff7",
+	}
 )
 
+func convertAccessScopeID(permissionSetID string) string {
+	identifierSuffix := strings.TrimPrefix(permissionSetID, accessScopeIDPrefix)
```
dhaus67, line 91: I was initially confused with the name, it probably should be the following:
```suggestion
func convertAccessScopeID(accessScopeID string) string {
	identifierSuffix := strings.TrimPrefix(accessScopeID, accessScopeIDPrefix)
```

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -67,10 +132,129 @@ func move(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacy.Store) e
...
+	if len(permissionSets) > 0 {
+		if err = store.UpsertMany(ctx, permissionSets); err != nil {
+			log.WriteToStderrf("failed to persist permission_sets to store %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+func getRoleAccessScopeID(obj *storage.Role) (string, error) {
```
dhaus67, line 180: nit: you're not required to do these changes, for me personally I'd like to be more specific about _which_ object we are dealing with. Applies to other functions as well (`getRolePermissionSetID`)
```suggestion
func getRoleAccessScopeID(role *storage.Role) (string, error) {
```

pkg/postgres/pgutils/utils.go
```diff
@@ -62,6 +63,18 @@ func NilOrTime(t *types.Timestamp) *time.Time {
 	return &ts
 }
 
+// NilOrUUID allows for a proto timestamp to be stored a timestamp type in Postgres
```
dhaus67, line 66:
```suggestion
// NilOrUUID allows for a string to be stored as a UUID type in Postgres.
```

pkg/search/postgres/common.go
```diff
@@ -133,7 +133,11 @@ func (q *query) getPortionBeforeFromClause() string {
 		var primaryKeyPaths []string
 		// Always select the primary keys first.
 		for _, pk := range q.Schema.PrimaryKeys() {
-			primaryKeyPaths = append(primaryKeyPaths, qualifyColumn(pk.Schema.Table, pk.ColumnName))
+			var cast string
+			if pk.SQLType == "uuid" {
```
dhaus67, line 137: How about re-using `postgres/walker/types.go` here instead of using the string representation?
I haven't looked too deep into it, but I saw that you already import `postgres/walker`, so I would assume that we do not end up in a circular dependency. 
However, if it's not possible, we should create an unexported const for this within this file.

pkg/search/postgres/common.go
```diff
@@ -423,8 +437,12 @@ func compileQueryToPostgres(schema *walker.Schema, q *v1.Query, queryFields map[
 	case *v1.Query_BaseQuery:
 		switch subBQ := q.GetBaseQuery().Query.(type) {
 		case *v1.BaseQuery_DocIdQuery:
+			cast := "::text[]"
+			if schema.ID().SQLType == "uuid" {
```
dhaus67, line 441: Why are we not mapping it to `text[]` here in case of `uuid`?

pkg/search/postgres/query/uuid_query.go
```diff
@@ -0,0 +1,108 @@
...
+			Query:  fmt.Sprintf("%s = $$", columnName),
+			Values: []interface{}{uuidVal},
+			equivalentGoFunc: func(foundValue interface{}) bool {
+				return strings.EqualFold(foundValue.(string), value)
+			},
+		}, nil
+	}
+	if queryModifiers[0] == pkgSearch.AtLeastOne {
+		panic("I dont think this is used")
+	}
```
dhaus67, line 53: Should we handle this more graceful instead of this panic and return an error here?

I'm not sure whether this is a pattern within the postgres code, I certainly know it's not for the rest of the stackrox code 😛

pkg/search/postgres/query/uuid_query.go
```diff
@@ -0,0 +1,108 @@
...
+		return WhereClause{
+			Query:  fmt.Sprintf("%s %s= $$", columnName, negationString),
+			Values: []interface{}{uuidVal},
+			equivalentGoFunc: func(foundValue interface{}) bool {
+				return strings.EqualFold(foundValue.(string), value) != negated
+			},
+		}, nil
+	}
+	err := fmt.Errorf("unknown query modifier: %s", queryModifiers[0])
+	utils.Should(err)
```
dhaus67, line 106: Since we already use the switch case, let's use this within a default case.
```suggestion
        default:
	        err := fmt.Errorf("unknown query modifier: %s", queryModifiers[0])
	        utils.Should(err)
	        return WhereClause{}, err
```

dhaus67: I recently learned that we should use `utils.Should` only in case or programatic errors that are not tied to user input. In this case, I'm not sure we should use it (since its bound on the query modifier parameter from the consumer of the function) - so it might be just sufficient enough to log within the default case and return an error here instead of using `utils.Should`.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
rhybrillou: Nice catch. Thanks.

central/role/default.go
```diff
@@ -57,12 +58,26 @@ var (
...
-		Id:          EnsureValidAccessScopeID("unrestricted"),
+		Id:          getAccessScopeIncludeAllID(),
 		Name:        "Unrestricted",
 		Description: "Access to all clusters and namespaces",
 	}
 )
 
+func getAccessScopeExcludeAllID() string {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return "ffffffff-ffff-fff4-f5ff-fffffffffffe"
```
dashrews78, line 69: nit:  Probably worth putting these as constants somewhere since we use this string here and in singleton.go

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/legacypermissionsets/rocksdb_plugin.go
```diff
@@ -0,0 +1,59 @@
+// Code generated by rocksdb-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: Since you copied this file, it is no longer generated.  We should probably update the comment.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/legacyroles/rocksdb_plugin.go
```diff
@@ -0,0 +1,56 @@
+// Code generated by rocksdb-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: Since you copied this file, it is no longer generated.  We should probably update the comment.

migrator/migrations/n_37_to_n_38_postgres_permission_sets/migration.go
```diff
@@ -1,20 +1,19 @@
-// Code generated by pg-bindings generator. DO NOT EDIT.
```
c-du, line 1: Please remove the postgres and legacy directory here since it is not in use.

migrator/migrations/n_46_to_n_47_postgres_roles/migration.go
```diff
@@ -1,6 +1,8 @@
-// Code generated by pg-bindings generator. DO NOT EDIT.
```
c-du, line 1: A reminder to remove legacy and postgres directory here.

c-du (approved):
Left a few reminders. LGTM for migration part.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -31,20 +46,70 @@ var (
...
 
+func convertAccessScopeID(accessScopeID string) string {
+	identifierSuffix := strings.TrimPrefix(accessScopeID, accessScopeIDPrefix)
+	replacement, found := accessScopeIDMapping[identifierSuffix]
+	if found {
+		return replacement
+	}
+	_, err := uuid.FromString(identifierSuffix)
+	if err != nil {
+		generatedID := uuid.NewV4().String()
```
mtodor, line 98: Can we use `GenerateAccessScopeID` here?

Do we need to keep old UUIDs? If not, then we can remove all this `uuid.FromString(identifierSuffix)` and always generate new V4 UUIDs.

The same can be applied to `convertPermissionSetID`.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -31,20 +46,70 @@ var (
...
+		"vulnreporter":          "ffffffff-ffff-fff4-f5ff-fffffffffff7",
+	}
 )
 
+func convertAccessScopeID(accessScopeID string) string {
+	identifierSuffix := strings.TrimPrefix(accessScopeID, accessScopeIDPrefix)
+	replacement, found := accessScopeIDMapping[identifierSuffix]
+	if found {
+		return replacement
+	}
```
mtodor, line 95: nitpick: let's use one-line if.
```suggestion
	if replacement, found := accessScopeIDMapping[identifierSuffix]; found {
		return replacement
	}
```

The same can be applied to `convertPermissionSetID`.

janisz (commented):
Could it be splited into two? Migration and code?

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration_test.go
```diff
@@ -86,3 +92,63 @@ func (s *postgresMigrationSuite) TestSimpleAccessScopeMigration() {
...
+func (s *postgresMigrationSuite) TestPermissionSetMigration() {
+	newStore := pgPermissionSetStore.New(s.postgresDB.Pool)
+	legacyStore, err := legacypermissionsets.New(s.legacyDB)
+	s.NoError(err)
+
+	// Prepare data and write to legacy DB
+	var permissionSets []*storage.PermissionSet
+	for i := 0; i < 200; i++ {
+		permissionSet := &storage.PermissionSet{}
+		s.NoError(testutils.FullInit(permissionSet, testutils.UniqueInitializer(), testutils.JSONFieldsFilter))
```
mtodor, line 105: As I see, `testutils.FullInit` will generate an ID with UUID V4. This will always test best case scenario. Maybe we should add in tests a few permission sets with IDs:
- just string values
- prefixed UUIDs
- special types, like `analyst`, etc.

This suggestion can be also applied to access scopes.

pkg/sac/testconsts/sac_consts.go
```diff
@@ -2,11 +2,10 @@ package testconsts
 
 // Clusters and Namespaces for scoped access control tests
 const (
-	Cluster1     = "aaaaaaaa-bbbb-4011-0000-111111111111"
```
mtodor, line 5: I'm trying to understand what is the reason behind changing values from UUIDs to a simple string.

mtodor (commented):
Nice work!

I agree with Tomek. Maybe it would be good to split PR so that it's easier to review.

pkg/sac/testconsts/sac_consts.go
dashrews78: @mtodor I think he needs a rebase since I pushed my changes yesterday.  That would solve this diff and probably clean up the review in general.

central/role/default.go
```diff
@@ -57,12 +58,26 @@ var (
...
+	}
+	return EnsureValidAccessScopeID("denyall")
+}
+
+func getAccessScopeIncludeAllID() string {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return "ffffffff-ffff-fff4-f5ff-ffffffffffff"
```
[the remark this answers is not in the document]
rhybrillou: The identifier constants are extracted in specific files. I'll check if the files can be re-used in the migrator code.

central/role/datastore/singleton.go
```diff
@@ -67,23 +67,37 @@ func Singleton() DataStore {
...
+	return rolePkg.EnsureValidPermissionSetID(attributes.idSuffix)
+}
+
+// The postgresID field is there for UUID support of permission set ID fields in postgres mode.
+// The values are UUIDs taken in descending order from ffffffff-ffff-fff4-f5ff-ffffffffffff
+// Next ID: ffffffff-ffff-fff4-f5ff-fffffffffff6
 var defaultRoles = map[string]roleAttributes{
 	rolePkg.Admin: {
 		idSuffix:           "admin",
+		postgresID:         "ffffffff-ffff-fff4-f5ff-ffffffffffff",
```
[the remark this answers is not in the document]
rhybrillou: If I read the spec correctly, there are only 6 bits that distinguish v4 UUIDs from random 128-bit strings, and they should be set accordingly here.
Now, v4 is supposedly for random UUIDs, while the IDs used for the default permissions and scopes are fixed.

IMHO, the IDs for the default permission sets and access scopes should have fixed values so that the role references to these do not get broken.

central/role/validate.go
```diff
@@ -27,11 +28,27 @@ const (
...
 func GeneratePermissionSetID() string {
-	return permissionSetIDPrefix + uuid.NewV4().String()
+	generatedIDSuffix := uuid.NewV4().String()
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return generatedIDSuffix
+	}
+	return permissionSetIDPrefix + generatedIDSuffix
+}
+
+func parseOrGenerateUUID(id string) string {
```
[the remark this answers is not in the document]
rhybrillou: `isValidID` is a valuable suggestion.
`GenerateID` is more delicate to implement (the generator code for rocksDB needs both prefix and ID again, and would be duplicating a lot of the previous function code).

migrator/migrations/n_37_to_n_38_postgres_permission_sets/migration.go
```diff
@@ -29,48 +28,22 @@ var (
...
-	if err != nil {
-		return err
-	}
-	if len(permissionSets) > 0 {
-		if err = store.UpsertMany(ctx, permissionSets); err != nil {
-			log.WriteToStderrf("failed to persist permission_sets to store %v", err)
-			return err
-		}
-	}
+	// The data migration code was moved to the simpleaccessscope migrator.
```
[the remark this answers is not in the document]
rhybrillou: All the migration code (schema creation included) is now moved to the n_52_to_n_53 migrator.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
rhybrillou: Technically, it would be doable. There still would have to be an exception for the default permission sets and access scopes, which get overwritten at each start from central.
I suppose the number of access scopes, permission sets and roles should stay reasonably low, but I'd still would prefer to avoid adding resource overhead for the migration. I would also like to keep the adjustment effort as low as possible for the case customers would have stored some identifiers in scripts and would have to adjust the list after the migration.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -67,10 +132,129 @@ func move(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacy.Store) e
...
+	_, accessIDParseErr := uuid.FromString(roleAccessScopeID)
+	if accessIDParseErr != nil {
+		log.WriteToStderrf("failed to convert role to postgres format, bad access scope ID. Role [%s], error %v", role.GetName(), accessIDParseErr)
+		return "", accessIDParseErr
+	}
+	return roleAccessScopeID, nil
+}
+
+func getRolePermissionSetID(role *storage.Role) (string, error) {
+	rolePemissionSetID := strings.TrimPrefix(role.GetPermissionSetId(), permissionSetIDPrefix)
```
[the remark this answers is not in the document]
rhybrillou: Nice catch, thank you.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -67,10 +132,129 @@ func move(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacy.Store) e
...
+	replacement, found := permissionsetIDMapping[rolePemissionSetID]
+	if found {
```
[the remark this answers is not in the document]
rhybrillou: Very good point. That would then require to keep track of all the migrated IDs for permission sets and access scopes. Then it would be possible to log any role that has references pointing to nothing.

janisz (approved):

central/role/accessscope_ids.go
```diff
@@ -1,6 +1,8 @@
 package role
 
-// postgres ID for default sccess scopes
+// Postgres IDs for access scopes
+// The values are UUIDs taken in descending order from ffffffff-ffff-fff4-f5ff-ffffffffffff
+// Next ID: ffffffff-ffff-fff4-f5ff-fffffffffffd
 const (
 	denyAllAccessScopeID      = "ffffffff-ffff-fff4-f5ff-fffffffffffe"
```
dashrews78, line 7: nit:  would it make sense to order these in descending order like was done with permission sets?

dashrews78 (approved):

PR #3726, opened 2022-11-07
Fix release tag for pre-release github action job

BradLugo:
## Description

Fixes release tag for the pre-release GitHub action job

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -518,7 +518,11 @@ jobs:
 
   patch-long-running-cluster:
     name: Patch long-running cluster
-    needs: [variables, properties, create-long-running-cluster]
+    needs: [variables, properties]
+    if: >-
+      github.event.inputs.dry-run != 'true' &&
+      needs.variables.outputs.rc != '1' &&
+      github.event.inputs.create-long-cluster != 'false'
     runs-on: ubuntu-latest
     env:
       NAME: gke-long-running-${{ needs.variables.outputs.milestone }}
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -529,7 +529,6 @@ jobs:
       TAG: ${{needs.variables.outputs.milestone}}
       KUBECONFIG: artifacts/kubeconfig
       INFRA_TOKEN: ${{secrets.INFRA_TOKEN}}
-      USE_GKE_GCLOUD_AUTH_PLUGIN: "True"
     steps:
       - uses: stackrox/actions/infra/install-infractl@v1
       - name: Test readiness
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -525,7 +525,7 @@ jobs:
       github.event.inputs.create-long-cluster != 'false'
     runs-on: ubuntu-latest
     env:
-      NAME: gke-long-running-${{ needs.variables.outputs.milestone }}
+      NAME: gke-long-running-${{ needs.variables.outputs.named-release-patch }}-rc-1
       TAG: ${{needs.variables.outputs.milestone}}
       KUBECONFIG: artifacts/kubeconfig
       INFRA_TOKEN: ${{secrets.INFRA_TOKEN}}
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -529,6 +529,7 @@ jobs:
       TAG: ${{needs.variables.outputs.milestone}}
       KUBECONFIG: artifacts/kubeconfig
       INFRA_TOKEN: ${{secrets.INFRA_TOKEN}}
+      USE_GKE_GCLOUD_AUTH_PLUGIN: "True"
     steps:
       - uses: stackrox/actions/infra/install-infractl@v1
       - name: Test readiness
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -267,7 +267,7 @@ jobs:
           set -uo pipefail
           gh api -H "$ACCEPT_RAW" "${{env.script_url}}" | bash -s -- \
             create-release \
-            "${{ needs.variables.outputs.named-release-patch }}" \
+            "${{ needs.variables.outputs.milestone }}" \
             "true"
 
       - run: |
--- a/.github/workflows/scripts/create-release.sh
+++ b/.github/workflows/scripts/create-release.sh
@@ -23,7 +23,8 @@ create_release_notes() {
       -H "Accept: application/vnd.github.v3.raw" \
       "/repos/${GITHUB_REPOSITORY}/contents/CHANGELOG.md?ref=${TAG}"
     )"
-    ESCAPED_VERSION="${TAG//./\.}"
+    VERSION_WITHOUT_RC="${TAG/-rc.[0-9]*/}"
+    ESCAPED_VERSION="${VERSION_WITHOUT_RC//./\.}"
     OUTPUT="$(echo "$CHANGELOG" | sed -n "/^## \[$ESCAPED_VERSION]$/,/^## \[/p" | sed '1d;$d')"
     if [ "$(wc -m <<< "$OUTPUT" | awk '{ print $1 }')" -gt 150000 ]; then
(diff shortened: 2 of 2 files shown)


tommartensen (changes requested):
Thanks for fixing - please review https://github.com/stackrox/stackrox/pull/3734 and merge it into your changes, it fixes finding the versions in the CHANGELOG for minor Y releases. 

Could you provide some context for your changes to the `patch-long-running-cluster` changes? As far as we understood, having `create-long-running-cluster` as a prerequisite to `patch-long-running-cluster` was preferably to repeating the condition. The auth plugin is necessary to avoid a deprecation on GKE/kubectl. 
The NAME should also been fine as it was, since we replace `X.Y.Z-rc.D` with `X-Y-Z-rc-X` with `${NAME//./-}`, and not hard-coding the milestone to the first release candidate would allow to reuse the long-running cluster for the second or third milestone with fewer changes.

vikin91 (commented):
I am missing context here - what is this fix about? What is happening without it and what will happen when we have it? How this has been tested?

Without these info, only people who know the code by heart, like @tommartensen & @0x656b694d  can properly review :)

.github/workflows/cut-rc.yml
```diff
@@ -518,14 +518,17 @@ jobs:
...
-    needs: [variables, properties, create-long-running-cluster]
+    needs: [variables, properties]
+    if: >-
+      github.event.inputs.dry-run != 'true' &&
+      needs.variables.outputs.rc != '1' &&
+      github.event.inputs.create-long-cluster != 'false'
     runs-on: ubuntu-latest
     env:
-      NAME: gke-long-running-${{ needs.variables.outputs.milestone }}
+      NAME: gke-long-running-${{ needs.variables.outputs.named-release-patch }}-rc-1
```
0x656b694d, line 528: Why this change? The code in the steps below replaces periods with dashes in place.

BradLugo:
Sorry y'all, I should've added context to the description. This was more of a temporary fix for the 3.72.2-rc.2 release than it was something to actually get merged in, which might become clear after answering some questions here. Before we get into it, it's important to note that Cut 3.72.2-rc.1 didn't run properly to begin with: Cut 3.72.2-rc.1

> As far as we understood, having create-long-running-cluster as a prerequisite to patch-long-running-cluster was preferably to repeating the condition.
> ...
> The NAME should also been fine as it was, since we replace X.Y.Z-rc.D with X-Y-Z-rc-X with ${NAME//./-}, and not hard-coding the milestone to the first release candidate would allow to reuse the long-running cluster for the second or third milestone with fewer changes.

Some discussion might be needed to resolve this problem, but in the current implementation, `create-long-running-cluster` only runs if `needs.variables.outputs.rc == '1'` (i.e., only on the first release candidate). Since the `create-long-running-cluster` is a prerequisite for `patch-long-running-cluster`, we noticed that `patch-long-running-cluster` wasn't running for 3.72.2-rc.2: https://github.com/stackrox/stackrox/actions/runs/3412566998 (both `patch-long-running-cluster` is skipped here). It might be ideal if there's a way to check if `create-long-running-cluster` was run on any of the named patch versions.

What @house-d and I were discussing yesterday was maybe the ideal solution is to create a long-running cluster named without the release candidate suffix (i.e., just the named release patch:`gke-long-running-${{ needs.variables.outputs.named-release-patch }}`) - perhaps refresh the lifespan after a release candidate is cut. We'd like y'all's input in this as well since y'all definitely know more about this process than we do 🙂 

That said, I'd be willing to believe something obvious is going over my head and that I'm thinking about the long-running cluster workflow all wrong. Please let me know if that's the case.

> The auth plugin is necessary to avoid a deprecation on GKE/kubectl.

When I was trying to gather the evidence to support the change reference here, I couldn't find what I thought was a failed job involving this (I might've incorrectly assumed the job here was relevant, which it doesn't seem actually be: https://github.com/stackrox/stackrox/actions/runs/3388210706/jobs/5643902473). This might not be a problem after all.

---

Hopefully, this explanation helps. I don't think these changes should be merged, but the changes' intention should be applied after we're done with this release. I'm also down to hop in a Slack huddle or meeting to discuss these changes or their intention more. Again, apologizes for not adding context. I should've added this info after we got it working, if not from the very beginning

BradLugo:
@house-d feel free to add anything I missed

tommartensen:
> Some discussion might be needed to resolve this problem, but in the current implementation, create-long-running-cluster only runs if needs.variables.outputs.rc == '1' (i.e., only on the first release candidate). Since the create-long-running-cluster is a prerequisite for patch-long-running-cluster, we noticed that patch-long-running-cluster wasn't running for 3.72.2-rc.2: https://github.com/stackrox/stackrox/actions/runs/3412566998 (both patch-long-running-cluster is skipped here). It might be ideal if there's a way to check if create-long-running-cluster was run on any of the named patch versions.

> What @house-d and I were discussing yesterday was maybe the ideal solution is to create a long-running cluster named without the release candidate suffix (i.e., just the named release patch:gke-long-running-${{ needs.variables.outputs.named-release-patch }}) - perhaps refresh the lifespan after a release candidate is cut. We'd like y'all's input in this as well since y'all definitely know more about this process than we do 🙂

In my understanding, if the provisioning of the long-running cluster (be it in creating or patching phase) failed during RC.1, we should re-run the workflow for RC.1 with manual dispatch. The workflows are idempotent. It is intended to create the long-running cluster only on RC.1. Regarding refreshing the lifespan, I think it might be a good idea to do during cut-rc, especially if a release takes longer than expected. Could you open a new issue in this Epic: https://issues.redhat.com/browse/ROX-10657?

Thanks for your explanations, I will try to root cause the gke-auth-plugin issue.

tommartensen:
> When I was trying to gather the evidence to support the change reference here, I couldn't find what I thought was a failed job involving this (I might've incorrectly assumed the job here was relevant, which it doesn't seem actually be: https://github.com/stackrox/stackrox/actions/runs/3388210706/jobs/5643902473). This might not be a problem after all.

The job you linked uses an outdated cut-rc workflow definition (without the installation of the GKE auth plugin), if you ran it with manual dispatch + selecting workflow from master instead of re-running it on the GHA UI, it would have worked :) 
Therefore, could you please revert the removal of the `USE_GKE_GCLOUD_AUTH_PLUGIN` env var.

BradLugo:
> > Some discussion might be needed to resolve this problem, but in the current implementation, create-long-running-cluster only runs if needs.variables.outputs.rc == '1' (i.e., only on the first release candidate). Since the create-long-running-cluster is a prerequisite for patch-long-running-cluster, we noticed that patch-long-running-cluster wasn't running for 3.72.2-rc.2: [stackrox/stackrox/actions/runs/3412566998](https://github.com/stackrox/stackrox/actions/runs/3412566998) (both patch-long-running-cluster is skipped here). It might be ideal if there's a way to check if create-long-running-cluster was run on any of the named patch versions.
> 
> > What @house-d and I were discussing yesterday was maybe the ideal solution is to create a long-running cluster named without the release candidate suffix (i.e., just the named release patch:gke-long-running-${{ needs.variables.outputs.named-release-patch }}) - perhaps refresh the lifespan after a release candidate is cut. We'd like y'all's input in this as well since y'all definitely know more about this process than we do 🙂
> 
> In my understanding, if the provisioning of the long-running cluster (be it in creating or patching phase) failed during RC.1, we should re-run the workflow for RC.1 with manual dispatch. The workflows are idempotent. It is intended to create the long-running cluster only on RC.1.

Ok, I think I understand. Are you saying that if the Cut 3.72.2-rc.1 workflow had been successful, the `patch-long-running-cluster` job from Cut 3.72.2-rc.2 would have not been skipped?

If the workflows are idempotent, I would expect (perhaps incorrectly) that if `create-long-running-cluster` was successful in the Cut 3.72.2-rc.1 workflow, ' patch-long-running-cluster` could run on any Cut 3.72.2-rc.X workflow. I could also be conflating the idempotency of the _workflow_ with the idempotency of each _job_. My general understanding of idempotency could be totally off too, so in any case, please correct me if I'm wrong. 

---
re: the gke-plugin, yeah I think I was barking up the wrong tree with my changes there. Thanks for taking a look at that 🙂

tommartensen:
@BradLugo This PR is superseded by https://github.com/stackrox/stackrox/pull/3877.

PR #3744, opened 2022-11-08
Refactor collection form page for better composition

dvail:
## Description (1 of 3)

This is the first of three PRs that aims to restructure the Collection form page to achieve two goals:
1. Allow better composition of components, so that the collection form can be used in places other than the form page (e.g. in a modal)
2. Decouple the validation of the incoming collection data from the rendering of the results

**These PRs aim to result in no functional changes and are only a restructuring to allow more flexibility for upcoming requirements.**

The first goal is accomplished with the following changes:
- Remove the `showBreadcrumbs` boolean prop and allow passing the [full header contents](https://github.com/stackrox/stackrox/pull/3744/files#diff-f8273c0635bf4dddbe59e1573266a1d33e49482717841f29ed6cc9dfbf708aacR201) from the parent container to the CollectionForm. This allows us to customize, or omit, any content that appears above the form inputs. This is needed for the in-page modal as well as the modal in the VulnMgmt reporting page.
- [Move the action callback handlers](https://github.com/stackrox/stackrox/pull/3744/files#diff-f8273c0635bf4dddbe59e1573266a1d33e49482717841f29ed6cc9dfbf708aacR149) (edit/clone/save/delete) from the CollectionForm to the parent CollectionFormPage. This allows us to decouple the editing of the data in the form from what happens when the data changes. e.g. The collection page and the modals will have different behavior.

The second goal is accomplished with the following changes:
- Extract the form itself from the PatternFly Drawer. This allow us to better handle the contents of the `DrawerContents` and the `DrawerPanel`. This is done in a [follow up PR](https://github.com/stackrox/stackrox/pull/3745) to attempt to preserve history as much as possible.
- Perform the [parsing of the back end collection response](https://github.com/stackrox/stackrox/pull/3744/files#diff-d2c9ca91f4296b6bb0e9282f7247b7f07c1408766275178d70782bb211242aa1R128) at the Drawer level, instead of at the `CollectionFormPage` level. This allows us to [display an error](https://github.com/stackrox/stackrox/pull/3744/files#diff-d2c9ca91f4296b6bb0e9282f7247b7f07c1408766275178d70782bb211242aa1R179) in place of the form if the UI cannot handle the response, while still allowing us to render the `CollectionResults`. This is needed because although the UI can only handle a subset of the collection rules, it would still be possible (and useful) to still display the results for that collection.

### Rough visual overview

This chart shows the flow of data and the conditional rendering __before__ this change.

- The form actions were performed inside the `CollectionForm` component
- The entirety of the rendering of the `CollectionForm` was handled by the component itself
- The conditional rendering for unsupported collection responses was performed in the `CollectionFormPage` component, rendering either the entire form contents and results, or nothing at all.

```mermaid
flowchart LR
    id2-->id3
    id4-->id7-->id8
    subgraph CollectionFormPage.tsx
        id1(request collection from server)-->id2(parse collection)
        id3{is collection valid?}
        id3-->id3-y((Yes))   
        id3-->id3-n((No))   
        id3-n-->id6(render error information)    
        style id6 fill:#f99 
        id3-y-->id5
        subgraph CollectionForm.tsx
            subgraph Drawer Content
                id3-y-->id4(render form)       
                style id4 fill:#9f9 
                id7(Header, Breadcrumbs)
                id8(Form)
            end
            subgraph Drawer Panel
                id5(render collection results)
                style id5 fill:#9f9 
            end
            id8-->|on form action|id9
            id9(Handle Actions)-->id10(on submit, generate server response)
            id9-->id11(on action: handle edit, clone, delete)
        end
    end
```

This chart shows the flow of data and the conditional rendering __after__ this change.

- The form actions are now performed as callbacks inside the parent `CollectionFormPage` component
- The rendering of the header contents of the `CollectionForm` is handled by the parent component
- The conditional rendering for unsupported collection responses s performed in the `CollectionForm` component, allowing us to render the form and the collection results independently of a parsing error.

```mermaid
flowchart LR
    subgraph CollectionFormPage.tsx
        id1(request collection from server)
        id2(pass header JSX as prop)
        id8-->|on form action|id10
        id10(Handle Actions)-->id12(on submit, generate server response)
        id10-->id11(on action: handle edit, clone, delete)
        id1-->id2
        id1-->id3
        id2-.->id7
        subgraph CollectionFormDrawer.tsx
            id3(Render Form Drawer)
            id3-->id4
            id3-->|Unsupported collections can still render results|id5
            subgraph Drawer Panel
                id5(render collection results)
                style id5 fill:#9f9 
            end
            subgraph Drawer Content
                subgraph CollectionForm.tsx
                    id7(render passed header prop)
                    id8(Form)
                end
                id4-->id4-y((Yes))   
                id4{is collection valid?}
                id4-y-->id9(render form)       
                style id9 fill:#9f9 
                id9-->id7
                id9-->id8
                id4-->id4-n((No))   
                id4-n-->id6(render error information)    
                style id6 fill:#f99 
            end
        end
    end
```

This chart shows the plans in a future PR to render a collection form in a modal, independent of the parent container. In this case the parent is the `CollectionFormPage`, but in the future this can be a separate page that renders the modal (e.g. the VulnMgmt reporting page).

```mermaid
flowchart TB
    subgraph CollectionFormPage.tsx
        id1(request collection from server)
        id2(pass header JSX as prop)
        id3-->|on form action|id10
        id10(Handle Form Actions)
        id1-->id2
        id1-->id3
        id2-.->id3
        subgraph CollectionFormDrawer.tsx
            id3(Render Form and Results)
            id3b(On collection link click)-->idb1
        end
    end

    idb1(Open modal with another collection)
    idb1-->ida1

    subgraph CollectionFormModal.tsx
        ida1(request collection from server)
        ida2(pass header JSX as prop)
        ida1-->ida2
        ida1-->ida3
        ida2-.->ida3
        subgraph CollectionFormDrawer
            ida11(On collection link click)-->ida10(Links open new tab)
            ida3(Render Form and Results)
        end
    end
```

## Checklist

## Testing Performed

Manual testing of all form functionality:
- Create, edit, clone and delete. (Both from the collections table and from the actions menu within the form page)
- Add name selectors
- Add label selectors
- Empty selector validation
- Ability to show/hide results panel
- Results panel behavior responds to smaller screen sizes

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
@@ -14,6 +14,7 @@ import {
     DropdownToggle,
     Flex,
     FlexItem,
+    PageSection,
     Title,
 } from '@patternfly/react-core';
 import { useMediaQuery } from 'react-responsive';
@@ -289,7 +290,7 @@ function CollectionsFormPage({
     }
 
     return (
-        <>
+        <PageSection className="pf-u-h-100" padding={{ default: 'noPadding' }}>
             {content}
             <AlertGroup isToast isLiveRegion>
                 {toasts.map(({ key, variant, title, children }) => (
@@ -321,7 +322,7 @@ function CollectionsFormPage({
             >
                 Are you sure you want to delete this collection?
             </ConfirmationModal>
-        </>
+        </PageSection>
     );
 }
 
--- a/ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
@@ -1,17 +1,43 @@
-import React, { ReactElement, useCallback } from 'react';
-import { PageSection } from '@patternfly/react-core';
+import React, { ReactElement, useCallback, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import {
+    Alert,
+    AlertActionCloseButton,
+    AlertGroup,
+    Breadcrumb,
+    BreadcrumbItem,
+    Button,
+    Divider,
+    Dropdown,
+    DropdownItem,
+    DropdownSeparator,
+    DropdownToggle,
+    Flex,
+    FlexItem,
+    Title,
+} from '@patternfly/react-core';
 import { useMediaQuery } from 'react-responsive';
 
 import useRestQuery from 'Containers/Dashboard/hooks/useRestQuery';
 import {
     CollectionResponse,
+    createCollection,
+    deleteCollection,
     getCollection,
     listCollections,
     ResolvedCollectionResponse,
+    updateCollection,
 } from 'services/CollectionsService';
+import { CaretDownIcon } from '@patternfly/react-icons';
+import BreadcrumbItemLink from 'Components/BreadcrumbItemLink';
+import { collectionsBasePath } from 'routePaths';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import ConfirmationModal from 'Components/PatternFly/ConfirmationModal';
+import useToasts from 'hooks/patternfly/useToasts';
+import { values } from 'lodash';
 import { CollectionPageAction } from './collections.utils';
 import CollectionForm from './CollectionForm';
-import { parseCollection } from './converter';
+import { generateRequest } from './converter';
 import { Collection } from './types';
 
 export type CollectionsFormPageProps = {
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
```diff
@@ -1,39 +1,64 @@
...
     pageAction: CollectionPageAction;
 };
 
-const noopRequest = {
-    request: Promise.resolve(undefined),
-    cancel: () => {},
-};
-
-const defaultCollectionData: Collection = {
+const defaultCollectionData: Omit<CollectionResponse, 'id'> = {
```
pedrottimark, line 48: You keep teaching me more TypeScript. What are pro and con versus `id: ''` property?

dvail: The main benefit is that this is explicitly saying we don't care that the `CollectionResponse` object is missing the `id` property (and in fact this forbids it from happening). Setting an empty value of `id: ''` would allow mixing of objects in the code that either have a valid id or have an empty string. To prevent bugs when we wanted to access the id we would need to add `if (id === '') {}` runtime checks, while using `Omit` enforces this at compile time.

ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
```diff
@@ -26,10 +27,14 @@ function CollectionsPage() {
 
     if (validPageActionProp) {
         return (
-            <CollectionsFormPage
-                hasWriteAccessForCollections={hasWriteAccessForCollections}
-                pageAction={validPageActionProp}
-            />
+            <>
+                <PageSection className="pf-u-h-100" padding={{ default: 'noPadding' }}>
```
pedrottimark, line 31: Comparing to `CollectionsTablePage` below brings to mind a question: what is the reason to factor `PageSection` element out of `CollectionsFormPage`?

pedrottimark (approved):
Thank you for breaking the refactor into steps with clear explanation. 2 questions.

ui/apps/platform/src/Containers/Collections/CollectionsPage.tsx
dvail: Admittedly this is probably a detail lost in the shuffle. I agree it makes more sense to be in `CollectionFormPage` to match the usage of `CollectionsTablePage`.

bradr5 (approved):

PR #3745, opened 2022-11-08
Extract collection form from drawer wrapper layout

dvail:
## Description (2 of 3)

This is the second step to the refactor described in https://github.com/stackrox/stackrox/pull/3744, which decouples the `CollectionForm` from the PatternFly Drawer layout used to display it. Viewing the code changes with Github's "Hide whitespace" option enabled makes it more apparent how code was move around in this change.

This is another non-functional change that only aims to restructure this code.

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
@@ -16,7 +16,7 @@ import {
 } from '@patternfly/react-core';
 import { CubesIcon } from '@patternfly/react-icons';
 import { TableComposable, TableVariant, Tbody, Tr, Td } from '@patternfly/react-table';
-import { Formik, FormikHelpers } from 'formik';
+import { useFormik } from 'formik';
 import * as yup from 'yup';
 
 import { collectionsBasePath } from 'routePaths';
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #3744** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3744" target="_blank">Graphite</a> 
    * **PR #3745** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3745" target="_blank">Graphite</a>  👈
      * **PR #3746** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3746" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3745?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -152,255 +125,172 @@ function CollectionForm({
...
-                            </DrawerHead>
-                            <DrawerPanelBody className="pf-u-h-100" style={{ overflow: 'auto' }}>
-                                <CollectionResults />
-                            </DrawerPanelBody>
-                        </DrawerPanelContent>
-                    }
-                >
-                    <DrawerContentBody className="pf-u-background-color-100 pf-u-display-flex pf-u-flex-direction-column">
-                        {headerContent}
-                        {initialData instanceof AggregateError ? (
```
pedrottimark, line 179: Ternary expression caused change from `useFormik` hook to `Formik` element and render prop pattern.

After this refactoring step, what are pro and con for return of the hook. It sounds like a movie ;)

ui/apps/platform/src/Containers/Collections/CollectionFormDrawer.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const initialData = parseCollection(collectionData.collection);
+    const initialEmbeddedCollections = collectionData.embeddedCollections;
+
+    useEffect(() => {
+        toggleDrawer(useInlineDrawer);
+    }, [toggleDrawer, useInlineDrawer]);
+
+    return (
+        <>
+            <Drawer isExpanded={isDrawerOpen} isInline={useInlineDrawer}>
```
pedrottimark, line 61: Seeing both boolean props raises pro and con of `use` versus `is` for prefix of second state variable.

ui/apps/platform/src/Containers/Collections/CollectionFormDrawer.tsx
```diff
@@ -0,0 +1,105 @@
...
+                <DrawerContent
+                    panelContent={
+                        <DrawerPanelContent
+                            style={{
+                                borderLeft: 'var(--pf-global--BorderColor--100) 1px solid',
+                            }}
+                        >
+                            <DrawerHead>
+                                <Title headingLevel="h2">Collection results</Title>
+                                <Text>See a live preview of current matches.</Text>
```
pedrottimark, line 71: A question came to mind about **live preview** instead of simply **preview**
* Intended meaning: when you change the collection in the form, you see change in current matches.
* Unintended meaning: when cloud environment changes, you see live changes to current matches.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
dvail: Haha, so this was a change that I've been thinking of for a while and decided to make since this code was already in flux.

Two main reasons:
- The [documentation on the Formik site](https://formik.org/docs/api/useFormik#use-cases-for-useformik) is pretty insistent that this is not the correct use case for this hook.
- I was testing usage of some of the other Formik components (`<Field>`, `<FieldArray>`), but none of them work since the Formik context is not available when using this hook. Additionally I think it might be cleaner to access some of the form values/errors in child components via the `useFormikContext` hook.

ui/apps/platform/src/Containers/Collections/CollectionFormDrawer.tsx
dvail: Fair point. A `use` prefix is especially bad here given the association with React.

pedrottimark (approved):
Refactoring for the win! 2 comments, as usual.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
pedrottimark: Thank you for explaining.

* Maintainer wrote that when render prop pattern was at its height, but React Router is example of comparable package which moved to hooks without looking back.
* Yeah, Linda needed context for policies to compose container form elements. A word of caution is to render Formik-specific elements only if package-specific abstraction adds enough value to balance need for team to understand it versus Formik state plus PatternFly form elements which we must all know well.

ui/apps/platform/src/Containers/Collections/CollectionFormDrawer.tsx
pedrottimark: Ha ha, I missed that connotation.

PR #3746, opened 2022-11-08
Extract collection fetching code into hook

dvail:
## Description (3 of 3)

This extracts the collection and embedded collection fetching logic into a reusable hook, which will be needed later when we add an independent modal for displaying the collection form.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

PR #3747, opened 2022-11-08
Add ability to view embedded collections in a pop up modal

dvail:
## Description

This adds the ability to view the collection form and results in a modal. The collection name links in the "Attached collections" section now open this modal allowing the user to get a quick view of this nested collection. When viewing a collection in the modal, clicking the link in the "Attached collections" table yet again will instead open a new tab to view that collection.

Additionally, the drawer panel close "X" button is removed when the drawer is displayed inline. This is to avoid conflicting with the same "X" close button in the modal that appears in nearly the same location, as well as the fact that we have a dedicated hide/show button next to the panel. The exception to this is when the drawer panel overlays the main content, since without the "X" button there would be no way to close the panel.

## Checklist

## Testing Performed

Navigate to a collection page that has attached collections.
image

Clicking on one of the collections in the table will open a modal that displays the information of the collection in read only mode.
image

From within the modal, clicking an embedded collection will open a new tab for that collection in "view" mode.
image

At the top of the modal, clicking the "Edit collection" button will open a new tab for the modal collection in "edit" mode.
image

Small screen views:
image
image
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
@@ -1,10 +1,11 @@
 import React, { ReactElement } from 'react';
-import { Button, Divider, Flex, FlexItem, Modal, Title } from '@patternfly/react-core';
+import { Button, Divider, Flex, FlexItem, Modal, Title, Truncate } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { useMediaQuery } from 'react-responsive';
 
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import { collectionsBasePath } from 'routePaths';
+import { CollectionResponse } from 'services/CollectionsService';
 import useCollection from './hooks/useCollection';
 import CollectionFormDrawer from './CollectionFormDrawer';
 
@@ -13,6 +14,28 @@ export type CollectionsFormModalProps = {
     collectionId: string;
     onClose: () => void;
 };
+const collectionTableCells = [
+    {
+        name: 'Name',
+        render: ({ id, name }: CollectionResponse) => (
+            <Button
+                variant="link"
+                component="a"
+                isInline
+                href={`${collectionsBasePath}/${id}?action=edit`}
+                target="_blank"
+                rel="noopener noreferrer"
+                icon={<ExternalLinkAltIcon />}
+            >
+                {name}
+            </Button>
+        ),
+    },
+    {
+        name: 'Description',
+        render: ({ description }) => <Truncate content={description} />,
+    },
+];
 
 function CollectionsFormModal({
     hasWriteAccessForCollections,
@@ -92,10 +115,7 @@ function CollectionsFormModal({
                     toggleDrawer={toggleDrawer}
                     // Since the form cannot be submitted, stub this out with an empty promise
                     onSubmit={() => Promise.resolve()}
-                    appendTableLinkAction={(id) => {
-                        const url = `${window.location.origin}${collectionsBasePath}/${id}`;
-                        window.open(url, '_blank', 'noopener noreferrer');
-                    }}
+                    collectionTableCells={collectionTableCells}
                 />
             </Modal>
         );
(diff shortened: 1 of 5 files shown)


dvail:
* master
  * **PR #3744** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3744" target="_blank">Graphite</a> 
    * **PR #3745** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3745" target="_blank">Graphite</a> 
      * **PR #3746** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3746" target="_blank">Graphite</a> 
        * **PR #3747** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3747" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3747?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    collectionData={data}
+                    isInlineDrawer={isLargeScreen}
+                    isDrawerOpen={isDrawerOpen}
+                    toggleDrawer={toggleDrawer}
+                    // Since the form cannot be submitted, stub this out with an empty promise
+                    onSubmit={() => Promise.resolve()}
+                    appendTableLinkAction={(id) => {
+                        const url = `${window.location.origin}${collectionsBasePath}/${id}`;
+                        window.open(url, '_blank', 'noopener noreferrer');
+                    }}
```
pedrottimark, line 98: First a simple question about naming. Is there any other use of `appendTableLinkAction` other than as value for `onItemClick` prop? My mind pauses to connect that thought with append in the prop name here.

pedrottimark: Second, a question about a potential alternative which I feel worth to explore without necessarily recommending. In this case, the result looks like a link, but you do not see address at lower left of browser window if you point at it, because `onClick` prop:

```tsx
<Button
    variant="link"
    className="pf-u-pl-0"
    isInline
    onClick={() => onItemClick(id)}
>
```

I am interested in any pro and con for accessibility from PatternFly team when genuine link has `onClick` handler instead of `href` prop, like the following:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/MitreAttackVectors/MitreAttackLink.tsx#L12-L23

If, and only if, feedback suggested that there is benefit to outweigh the cost, would it make sense to change prop value from callback function to `Button` element which encapsulates `href` or `onClick` prop.

pedrottimark (approved):
You are making quite a collection of contributions, pardon pun. 2 questions.

ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
dvail: I agree, the `appendTableLinkAction` is a bit of an artifact from some earlier plans that have evolved with design changes. Additionally, the read-only table that replaces the CollectionAttacher has some duplication with the cells that are rendered so I should be able to consolidate if we are passing the components down.

TBH I don't think there are any pros to using an onclick/window.open in this case instead of a link button. The main pro was to avoid some indirection and prop drilling, but since we are passing the callback just as deeply as we would be the components, I don't think it outweighs the cons of no link preview and programmatic navigation in this case.

dvail: Since we are passing components instead of callbacks, I was able to add the "external link" icon to the table cells, much like the ones in the header. It better conveys what the link will do, but do you think this is too much?

image

pedrottimark: Because my mind uses ”one of these things is not like the others” principle, I would notice that **Edit collection** has icon but collection names do not. Therefore, I think it is ”just right” to mix metaphors.

Two observations which you are free to choose according to house special for today: take it, or leave it ;-)
* Although icon at the left of collection name seems superior for alignment in the table, icon at the right feels more fluent for **Edit collection** link.
* The first picture under **Description** both wraps name and truncates description of the first collection. Although every additional style risks trouble in a responsive layout, what do you think are pro and con of no wrap for name?

dvail: I'm neutral on the first point, so a positive and a neutral == positive, I'll move the icon to the right.

For the second point, I like the idea but this leaves us open to very long user input breaking the layout of the page due to the nowrap. I've felt the table layouts need a second look though, so I'll add this to my list of follow ups to take a closer look at.

PR #3750, opened 2022-11-09
ROX-13432: Cleaning up unused code copied/pasted from topology demo

alwayshooin:
Cleaning up the copy/paste code for the PF Topology network graph. You shouldn't see the edge connector arrows (to create new edges) and the rest of the graph should look/behave the same as it did before. I only removed the parts of the code that are not using and are not going to be using as part of MVP.

image


vjwilson (approved):
/lgtm

alwayshooin:
merging bc the tests are unrelated to the changes

PR #3756, opened 2022-11-09
ROX-13466: Fix deletion of groups with empty properties

dhaus67:
## Description

For additional context, see [ROX-13446](https://issues.redhat.com/browse/ROX-13446) as well.

During startup and the initialization of the group datastore, we attempt to delete groups with empty properties.

This has been there historically for quite some time ([the initial PR that introduced this](https://github.com/stackrox/rox/pull/1659) + [the associated ticket ROX-1698](https://issues.redhat.com/browse/ROX-1698)).

Now, we added validation with the move from the tuple ID key for groups (using its propreties) to a single UUID.
This leads to an issue when trying to remove groups with empty properties during startup.
Since the properties will be empty, the validation which is done within the `Remove` call, will fail (since the validation is checking whether auth provider ID is set etc.).

Within this PR, a separate function has been introduced at the datastore layer that will:
- retrieve all groups with empty properties.
- delete those groups, skipping validation and other mutability trait verification.

## Checklist
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into 
[rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed

- see unit tests added running within CI and passing.

### DIFF
--- a/central/group/datastore/datastore.go
+++ b/central/group/datastore/datastore.go
@@ -21,6 +21,7 @@ type DataStore interface {
 	Mutate(ctx context.Context, remove, update, add []*storage.Group, force bool) error
 	Remove(ctx context.Context, props *storage.GroupProperties, force bool) error
 	RemoveAllWithAuthProviderID(ctx context.Context, authProviderID string, force bool) error
+	RemoveAllWithEmptyProperties(ctx context.Context) error
 }
 
 // New returns a new DataStore instance.
--- a/central/group/datastore/datastore_impl.go
+++ b/central/group/datastore/datastore_impl.go
@@ -4,11 +4,11 @@ import (
 	"context"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/group/datastore/internal/store"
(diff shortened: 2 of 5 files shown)


dhaus67:
* master
  * **PR #3756** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3756" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3756?utm_source=stack-comment).

janisz (approved):

ivan-degtiarenko (changes requested):
From the initial look at it, I don't see why we can't change the direction here compared to the one chosen here https://issues.redhat.com/browse/ROX-1698 and make it a migration. That would allow us to avoid introducing and supporting `RemoveAllWithEmptyProperties` which doesn't have any other potential applications.

ivan-degtiarenko (approved):
As discussed offline, migration will potentially be done in a separate PR - this PR addresses only empty `authProviderId`, `key` and `value` cases. This PR doesn't address the `nil` props case.

LGTM 🚢

PR #3755, opened 2022-11-09
Rename NodeScanV2 to NodeInventory

vikin91:
## Description

Scope: This PR renames `storage.NodeScanV2` to `storage.NodeInventory`.

Why?
This is done to avoid confusion between the word `scan` being understood as:

1. A result returned by the Scanner (containing list of vulnerabilities)
2. A result of Compliance searching files on a disk.

We hereby propose to call the (1) a scan, whereas the (2) a node inventory.

## Checklist

- [x] Investigated and inspected CI test results

## Do not apply

- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

Reason: This is a pure rename refactor

## Testing Performed

- [x] CI
- [x] Manual tests consisting of starting a local cluster and looking in the
  logs of Sensor, Central, and Compliance

PR #3757, opened 2022-11-09
ROX-13722: Move NodeInventory into Node

vikin91:
## Description

This PR changes the scope in which the `storage.NodeInventory` is treated as a standalone proto message.

Before this PR, `NodeInventory` was used as a standalone message between: `Compliance -> Sensor -> Central`.
After this PR is merged, it is: `Compliance -> Sensor`.

Sensor is responsible for connecting the `NodeInventory` with `Node` (goal of https://github.com/stackrox/stackrox/pull/3892) and sending `Node` further to central.

The newly added `NodeInventory` field will not be saved into the DB as a separate field or table, but it will be included in the serialized data in the DB.
`NodeInventory` is meant to be used in the request to Scanner to scan the packages included in it.
The reply from Scanner will be converted in Central into a `storage.NodeScan` that will
be inserted into the DB the same way as the existing v1 node scan data (i.e., without any need for further refactoring as `NodeScan` message is already being used).

Currently, the API between Scanner and Central is unable to handle the
`NodeInventory` - support for this will be added in separate PRs: 
- https://github.com/stackrox/stackrox/pull/3940 (WIP)
- https://github.com/stackrox/scanner/pull/1004

This PR handles the proto part for [ROX-12943](https://issues.redhat.com/browse/ROX-12943).
The proper implementation of ROX-12943 will be done in a follow-up - see  https://github.com/stackrox/stackrox/pull/3892.

⚠️ See also the prefactor PR: #4392 that was created from this PR by extracting the non-risky parts. This PR will be rebased when the prefactor is merged. 

## Checklist

## Checks that do not apply

    - we will cover the tests in [ROX-12943](https://issues.redhat.com/browse/ROX-12943) as this PR only changes the proto structure and defines no extra behaviors.

## Testing Performed

- [x] Regenerated code with `make proto-generated-srcs generated-srcs` and checked `central/graphql/resolvers/generated.go` for errors
- [ ] CI
- [x] Manual test consisting deploying a cluster locally, looking at the
  Central logs and seeing that the `Node` with `NodeInventory` arrives.

### Query Central HTTP API

The goal was to query the API to see how the data is provided by Central. 
The inventory scanning was disabled in this test (for each inventory item, scanner returned 0 vulns).
Query: `GET https://localhost:8443/v1/nodes/{clusterID}/{nodeID}`

#### Node contains v1 and v2 data (scanner scans v1 only)

<details>
  <summary>Call result</summary>

```json
{
	"id": "bf5bf7d4-2d77-4194-9ab5-570848c55777",
	"name": "colima-fake",
	"taints": [],
	"clusterId": "4b3f14e0-b7ce-461d-8a36-b8698ff51f15",
	"clusterName": "remote",
	"labels": {
		"fakeLabelsK": "fakeLabelsV"
	},
	"annotations": {
		"fakeAnnotationsK": "fakeAnnotationsV"
	},
	"joinedAt": "2023-01-06T15:33:45.324727501Z",
	"internalIpAddresses": [
		"192.168.255.254"
	],
	"externalIpAddresses": [
		"10.10.255.254"
	],
	"containerRuntimeVersion": "docker://20.10.18",
	"containerRuntime": {
		"type": "DOCKER_CONTAINER_RUNTIME",
		"version": "20.10.18"
	},
	"kernelVersion": "99.19.16",
	"operatingSystem": "Alpine Linux v3.16",
	"osImage": "RedHat CoreOS v99.66.33",
	"kubeletVersion": "v1.25.0+k3s1",
	"kubeProxyVersion": "v1.25.0+k3s1",
	"lastUpdated": "2023-01-06T15:33:45.561847530Z",
	"k8sUpdated": "2023-01-06T15:33:45.325838631Z",
	"nodeInventory": {
		"nodeName": "colima",
		"scanTime": "2023-01-06T15:33:45.324727501Z",
		"components": {
			"namespace": "Testme OS",
			"rhelComponents": [
				{
					"id": "0",
					"name": "vim-minimal",
					"namespace": "rhel:8",
					"version": "2:7.4.629-6.el8.x86_64",
					"arch": "x86_64",
					"module": "FakeMod",
					"cpes": [
						"cpe:/a:redhat:enterprise_linux:8::baseos"
					],
					"addedBy": "FakeLayer",
					"executables": []
				},
				{
					"id": "0",
					"name": "libsolv",
					"namespace": "rhel:8",
					"version": "0.7.7-1.el8.x86_64",
					"arch": "x86_64",
					"module": "FakeMod",
					"cpes": [],
					"addedBy": "FakeLayer",
					"executables": []
				}
			]
		},
		"notes": [
			"LANGUAGE_CVES_UNAVAILABLE"
		]
	},
...
}
```

</details>

#### Node contains v1 data only

<details>
  <summary>Call result</summary>

```json
{
	"id": "f723ae15-334c-4c1e-a38a-453f27eba5f1",
	"name": "colima",
	"taints": [],
	"clusterId": "4b3f14e0-b7ce-461d-8a36-b8698ff51f15",
	"clusterName": "remote",
	"labels": {...},
	"annotations": {...},
	"joinedAt": "2023-01-06T09:29:58Z",
	"internalIpAddresses": [
		"192.168.5.15"
	],
	"externalIpAddresses": [],
	"containerRuntimeVersion": "docker://20.10.20",
	"containerRuntime": {
		"type": "DOCKER_CONTAINER_RUNTIME",
		"version": "20.10.20"
	},
	"kernelVersion": "5.15.82-0-virt",
	"operatingSystem": "linux",
	"osImage": "Alpine Linux v3.16",
	"kubeletVersion": "v1.25.4+k3s1",
	"kubeProxyVersion": "v1.25.4+k3s1",
	"lastUpdated": "2023-01-06T14:16:11.495989645Z",
	"k8sUpdated": "2023-01-06T14:15:30.578707270Z",
	"nodeInventory": null,
...
}
```
</details>

### Query Central GraphQL API

`POST https://localhost:8443/api/graphql`

<details>
  <summary>gQL querries</summary>

```
query nodev1v2 {
	 node(id: "bf5bf7d4-2d77-4194-9ab5-570848c55777") {
		...fields
	}
}

query nodev1 {
	node(id: "f723ae15-334c-4c1e-a38a-453f27eba5f1") {
		...fields
	}
}

fragment fields on Node {
   # arbitrary selected fields including
   nodeInventory{...} # input to scanner
   scan{...} # output from scanner
}
```
</details>

#### Node contains v1 and v2 data (scanner scans v1 only)

<details>
  <summary>Call result</summary>

```json
{
	"data": {
		"node": {
			"id": "bf5bf7d4-2d77-4194-9ab5-570848c55777",
			"name": "colima-fake",
			"nodeInventory": {
				"nodeName": "colima",
				"scanTime": "2023-01-06T16:11:25.317593632Z",
				"notes": [
					"LANGUAGE_CVES_UNAVAILABLE"
				],
				"components": {
					"rhelComponents": [
						{
							"id": 0,
							"namespace": "rhel:8",
							"name": "vim-minimal",
							"version": "2:7.4.629-6.el8.x86_64",
							"arch": "x86_64",
							"cpes": [
								"cpe:/a:redhat:enterprise_linux:8::baseos"
							]
						},
						{
							"id": 0,
							"namespace": "rhel:8",
							"name": "libsolv",
							"version": "0.7.7-1.el8.x86_64",
							"arch": "x86_64",
							"cpes": []
						}
					]
				}
			},
			"scan": {
				"scanTime": "2023-01-06T16:11:25.511172079Z",
				"nodeComponents": [
					{
						"id": "docker#20.10.18#",
						"name": "docker",
						"nodeVulnerabilityCount": 0
					},
					{
						"id": "kernel#99.19.16#",
						"name": "kernel",
						"nodeVulnerabilityCount": 13
					},
					{
						"id": "kubelet#v1.25.0+k3s1#",
						"name": "kubelet",
						"nodeVulnerabilityCount": 0
					},
					{
						"id": "kube-proxy#v1.25.0+k3s1#",
						"name": "kube-proxy",
						"nodeVulnerabilityCount": 0
					}
				]
			}
		}
	}
}
```
</details>

#### Node contains v1 data only

<details>
  <summary>Call result</summary>

```json
{
	"data": {
		"node": {
			"id": "f723ae15-334c-4c1e-a38a-453f27eba5f1",
			"name": "colima",
			"nodeInventory": null,
			"scan": {
				"scanTime": "2023-01-06T14:16:11.4924593Z",
				"nodeComponents": [
					{
						"id": "docker#20.10.20#",
						"name": "docker",
						"nodeVulnerabilityCount": 0
					},
					{
						"id": "kernel#5.15.82-0-virt#",
						"name": "kernel",
						"nodeVulnerabilityCount": 181
					},
					{
						"id": "kubelet#v1.25.4+k3s1#",
						"name": "kubelet",
						"nodeVulnerabilityCount": 0
					},
					{
						"id": "kube-proxy#v1.25.4+k3s1#",
						"name": "kube-proxy",
						"nodeVulnerabilityCount": 0
					}
				]
			}
		}
	}
}
```
</details>


vikin91:
* master
  * **PR #3533** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3533" target="_blank">Graphite</a> 
    * **PR #3755** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3755" target="_blank">Graphite</a> 
      * **PR #3757** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3757" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3757?utm_source=stack-comment).

proto/storage/node.proto
```diff
@@ -49,6 +49,7 @@ message Node {
     // Time we received an update from Kubernetes.
     google.protobuf.Timestamp k8s_updated  = 26 [(gogoproto.moretags) = 'sensorhash:"ignore"'];
 
+    NodeInventory node_inventory           = 28 [(gogoproto.moretags) = 'policy:"Node Inventory" sql:"-"'];
```
vikin91, line 52: Add a comment that this will not be stored.
Check that this is not stored in the Node table.

proto/storage/node.proto
```diff
@@ -86,11 +89,28 @@ message NodeScan {
 }
 
 message NodeInventory {
-    string                         node_id    = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk,type(uuid)"'];
-    string                         node_name  = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
-    google.protobuf.Timestamp      scan_time  = 3 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
-    scannerV1.Components           components = 4;
-    repeated scannerV1.Note        notes      = 5;
+    string                         node_id    = 1;
```
md2119, line 92: `NodeInventory` is a field within `Node` which has `id` and `name`. Why are we adding id and name again?

vikin91: `NodeInventory` is generated in Compliance and has only node name at that stage. Once it arrives to Sensor, we compare node name against all other node names known to Sensor to connect these two by attaching `NodeInventory` to `Node`. 

You are right, node ID is not required for `NodeInventory` anymore. I will remove it here.

vikin91: We were thinking about separating `NodeInventory` from `Node`, but this would cause a lot of extra work. Instead, we think that `NodeInventory` does not make a lot of sense if it is detached from the `Node`.

Moreover, the information about `NodeInventory` will be duplicated in the DB - the same information will be available in the `NodeScan` field that is returned from Scanner (but it will include all information about the vulnerabilities).

The flow of messages goes like this: `Sensor -> Node+NodeInventory -> Central Pipeline -> NodeInventory + 4 fields of Node -> Scanner -> NodeScan -> Central -> Node+NodeScan -> DB`. 
I think there is no need to store  `Node+NodeInventory+NodeScan` as the information would be duplicated. What do you think?

Alternatively, we may add `NodeInventory` to the `Node` in the DB, but I would need some support for this from DB experts. How this would need to be done if we would like to go that way?

central/node/datastore/dackbox/datastore/datastore_sac_test.go
```diff
@@ -262,9 +262,11 @@ func (s *nodeDatastoreSACSuite) TestGetNode() {
 				s.True(found)
 				s.NotNil(fetchedNode)
 
-				// Priority can have updated value, and we want to ignore it.
-				fetchedNode.Priority = s.testNodes[nodeID].Priority
-				s.Equal(*s.testNodes[nodeID], *fetchedNode)
+				if fetchedNode != nil {
```
vikin91, line 265: This prevents potential panic if `fetchedNode` is nil.

sensor/common/compliance/node_inventory_handler_impl.go
```diff
@@ -82,28 +84,63 @@ func (c *nodeInventoryHandlerImpl) run() <-chan *central.MsgFromSensor {
...
+				Resource: &central.SensorEvent_Node{
+					Node: node,
 				},
 			},
 		},
 	}:
 	}
 }
+
+func (c *nodeInventoryHandlerImpl) fakeNode(inventory *storage.NodeInventory) *storage.Node {
```
vikin91, line 121: This function is temporary and will be replaced in the follow-up PR. I used it here for manually testing as it allows us to send messages to Central and see the fake node appear in the UI.

vikin91:
@rhybrillou @keyallis @md2119 I would appreciate your review of the parts related to: the proto annotations, the graphQL code generation, and the influence of those on the database.

If you have any additional tests cases that can be added for the GraphQL & DB part, then please suggest them.

We do in Maple synchronous reviews of selected PRs in a video calls - if you prefer this form of review, then please let me know and I will organize a meeting.

central/graphql/resolvers/gen/main.go
```diff
@@ -84,6 +84,10 @@ var (
 				ParentType: reflect.TypeOf(storage.ImageScan{}),
 				FieldName:  "Components",
 			},
+			{
+				ParentType: reflect.TypeOf(storage.Node{}),
+				FieldName:  "Components",
+			},
```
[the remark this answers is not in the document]
md2119: The graphQL auto-code generator (and most of the code generators we have) works on reflection. It indicates the gen code to skip generating code for  `Components` field for all occurrences of `storage.Node` in unwrap.

proto/internalapi/central/sensor_events.proto
```diff
@@ -68,7 +68,6 @@ message SensorEvent {
         storage.NamespaceMetadata      namespace          = 6;
         storage.Secret                 secret             = 7;
         storage.Node                   node               = 9;
-        storage.NodeInventory          node_inventory     = 25;
```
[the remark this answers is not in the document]
vikin91: Thanks for this hint! In this case we are inside of the `oneof ` block, and apparently the keyword `reserved` cannot be used. I checked https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility_issues and we should be safe as long as we remove the code referring to this field (which we do by removing central/sensor/service/pipeline/nodescansv2/pipeline.go).

So I think, there is no way to block the number 25 from being reused in this case. Let me know if you have different opinion.

proto/storage/node.proto
```diff
@@ -86,11 +89,17 @@ message NodeScan {
 }
 
 message NodeInventory {
-    string                         node_id    = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk,type(uuid)"'];
-    string                         node_name  = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
-    google.protobuf.Timestamp      scan_time  = 3 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
-    scannerV1.Components           components = 4;
-    repeated scannerV1.Note        notes      = 5;
```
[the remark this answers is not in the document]
vikin91: This is correct, the `NodeInventory` is new and we allow ourselves a bit more changes. But I followed your suggestion and kept the reserved numbers.

proto/storage/node.proto
```diff
@@ -86,10 +89,18 @@ message NodeScan {
 }
 
 message NodeInventory {
-    string                         node_id    = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk,type(uuid)"'];
-    string                         node_name  = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
-    google.protobuf.Timestamp      scan_time  = 3 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
-    scannerV1.Components           components = 4;
+    reserved 1; // was node_id; now node_name is used to define the relation between node_inventory with a node
```
Maddosaurus, line 92: Question: We removed the `node_id` here, because `NodeInventory` is now always sent as part of a `Node` message, which already contains the `ID`, correct?

vikin91: Yes. If we would keep it here, the info would be duplicated

proto/storage/node.proto
vikin91: @md2119 @keyallis could I ask you to give this PR another look?

We still keep the proto changes as last time, but the `NodeInventory` is now present in the Database in a serialized form. 
I was wrong last time saying that it is not stored in the DB - what I meant was that there will be no extra tables or fields created in the DB for NodeInventory. I corrected the PR description to be more precise.

This PR blocks a major part of the work on RHCOS epic and I would love to move forward with it.

central/node/datastore/dackbox/datastore/datastore_sac_test.go
Maddosaurus: For me to understand: `s.NotNil()` will not stop execution of the test when the subject is nil, correct?

central/graphql/generator/codegen/utils.go
```diff
@@ -41,6 +41,12 @@ func plural(s string) string {
 }
 
 func importedName(p reflect.Type) string {
+	// Handle collision of "v1":
+	// "github.com/stackrox/rox/generated/api/v1"
+	// "github.com/stackrox/scanner/generated/scanner/api/v1"
+	if strings.HasSuffix(p.PkgPath(), "scanner/generated/scanner/api/v1") {
+		return fmt.Sprintf("scannerV1.%s", p.Name())
```
Maddosaurus, line 48: Could you elaborate a bit how this is used?
I tracked it back to codegen.go, but I think I'm missing a bit of knowledge about graphql and/or our code generation involved here.

vikin91: This is used in `make generated-srcs` in one of the steps that can be executed isolated by:  

```
go run central/graphql/resolvers/gen/main.go
mv generated.go central/graphql/resolvers/generated.go
```

When you run this, then the generated code will refer to some libraries using `v1.FuncName`. Before this PR, the `v1` was `github.com/stackrox/rox/generated/api/v1` but now we start using also `github.com/stackrox/scanner/generated/scanner/api/v1` so using the `v1`  would be ambiguous.

Thus, I proposed that all imports ending with  `scanner/generated/scanner/api/v1` shall be addressed by a named import `scannerV1`, while `github.com/stackrox/rox/generated/api/v1` would remain the same.


By introducing `message Components {}` locally inside `NodeInventory` (Instead of using `scannerV1.Components components` I minimalized the number of conflicts between `v1` and `scannerv1`, but that change still has influence on the generated code.

If we would keep `scannerV1.Components components` in the `NodeInventory` in `node.proto`, we would be forced to solve multiple other conflicts concerning functions named idetically but having different types of parameters. In that case, the change in `[codegen/utils.go](https://github.com/stackrox/stackrox/pull/3757/files/3dbf09e0d79e1392a2d4fd6733d26770966b4681#diff-47afd3b752f5e9f8f4eb339e51277b30015312ff03684e7977dc66d411e05e3f)` would not be sufficient.

central/node/datastore/dackbox/datastore/datastore_sac_test.go
vikin91: Correct, `NodNil` will do the assertion, print a message and then continue. The continuation in this case would lead to a panic.
It is not so bad as this is only a test code that would panic, but in this case it was an clear bug.

Maddosaurus:
Besides the Style errors you're getting, this looks good to me!
I'll re-review once these are sorted out for the protobuf :)

vikin91:
> Besides the Style errors you're getting, this looks good to me! I'll re-review once these are sorted out for the protobuf :)

The style errors are wrong in this case - I made the changes to avoid potential compatibility issues and the `protolock` started complaining. The solution here would be to overwrite the protolock file and force-ignore the errors, but I wanted to consult with @rhybrillou first.

proto/storage/node.proto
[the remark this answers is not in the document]
vikin91: I dig into the protoloc (the binary that checks proto backwards compatibility) and there are two options that we can do:

1. Overwrite the proto.lock file forcing any incompatibilities to be ignored - should be no harm as there are 0 clusters running old version of Compliance but new version of Sensor (both are using the same image). This would  allow us to get rid of the `reserved` fields and write any definition of NodeInventory we want.
2. Keep the strict backwards compatibility and reserve the numbers 1 and 4 but also names `node_id` and `components` ([why names also matter for protolock's author](https://github.com/nilslice/protolock/issues/52#issuecomment-431084318)). This would require us to rename the field `components` and start reserving the names additionally to the numbers.

In the given situation, I tend to choose the first option. If there are any opinions against this, I am fine to revisit this choice. (FYI: @rhybrillou )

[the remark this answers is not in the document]
rhybrillou: Considering `NodeInventory` is being added to `Node` here, I don't expect any `storage.NodeInventory` with the old format present in the store or on the communication channels.
Should there be the slightest risk of our processes exploding in flight while decoding objects encoded with the old format, I'd go for the more conservative approach.
In the current state of things, I think the easier first option is OK. Please make sure the `proto.lock` file is also updated accordingly.

proto/storage/node.proto
```diff
@@ -86,11 +87,47 @@ message NodeScan {
...
 message NodeInventory {
-    string                         node_id    = 1 [(gogoproto.moretags) = 'search:"Node ID,store" sql:"pk,type(uuid)"'];
-    string                         node_name  = 2 [(gogoproto.moretags) = 'search:"Node,store"'];
-    google.protobuf.Timestamp      scan_time  = 3 [(gogoproto.moretags) = 'search:"Node Scan Time,store"'];
-    scannerV1.Components           components = 4;
-    repeated scannerV1.Note        notes      = 5;
+    string                         node_name  = 1;
+    google.protobuf.Timestamp      scan_time  = 2;
+
+    // Components represents a subset of the scannerV1.Components proto message containing only fields required for RHCOS node scanning
```
vikin91, line 93: Put also comment in Scanner proto that this field is related to scannerV1.Components and should be updated in pair.

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -108,6 +106,26 @@ func convertAndDedupRHELComponents(rc *database.RHELv2Components) []*scannerV1.R
 	return maps.Values(convertedComponents)
 }
 
-func makeComponentKey(component *scannerV1.RHELComponent) string {
+func convertExecutables(exe []*scannerV1.Executable) []*storage.NodeInventory_Components_RHELComponent_Executable {
```
vikin91, line 109: Make sure to use getters here to avoid nils

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -92,7 +90,7 @@ func convertAndDedupRHELComponents(rc *database.RHELv2Components) []*scannerV1.R
 			Arch:        rhelc.Arch,
 			Module:      rhelc.Module,
 			Cpes:        rc.CPEs,
-			Executables: rhelc.Executables,
+			Executables: convertExecutables(rhelc.Executables),
```
vikin91, line 93: Make sure to use getters here to avoid nils

proto/internalapi/central/sensor_events.proto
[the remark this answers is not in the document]
vikin91: Try reserving 25 outside of `oneof`

vikin91:
Less risky parts of this PR have been moved into a separate prefactor PR: https://github.com/stackrox/stackrox/pull/4392

proto/storage/node.proto
```diff
@@ -86,11 +87,47 @@ message NodeScan {
...
+            message Executable {
+                string                      path              = 1;
+                message FeatureNameVersion {
+                    string name    = 1;
+                    string version = 2;
+                }
+                repeated FeatureNameVersion required_features = 2;
+            }
+            repeated Executable executables = 9;
+        }
```
jvdm, line 118: As part of [ROX-13794](https://issues.redhat.com/browse/ROX-13794), we will not map CPEs in the node "invetorization" (aka. scanning). Instead, we will return the content sets found in the O.S. installation. The CPE mapping will be performed by Central Scanner when matching vulnerabilities. Because of that, we need to replace `cpe` with `content_sets`. Two options:

1. Replace the `cpe` field in `RHELComponent` with `content_sets`.
2. Remove the `cpe` field and add a `content_sets` field to `Component`.

Option 2 exists because today, we cannot determine precisely which content set each package came from, so we assume the package might have come from any of the content sets specified in the O.S. installation. This may be possible in the long future, so we might just use individual fields just as we do with CPEs, especially considering the size is neglectable.

CC @vikin91

jvdm: See related PR: https://github.com/stackrox/scanner/pull/1053

vikin91: Thanks! As #4392 is a prefactor to this PR, the change will be addressed in #4392 (option 2) and then this PR will be rebased.

vikin91:
Converting this to draft, as this PR will be replaced by https://github.com/stackrox/stackrox/pull/4430 (or its descendants)

vikin91:
Closing in favor of https://github.com/stackrox/stackrox/pull/4483 and its descendants

PR #3758, opened 2022-11-09
Remove os.Std* from roxctl/central

janisz:
## Description

We have misconfigured linter so it shipped too much not only `central` but also `roxctl/central`.
This PR fixes some issues and enable linting on `roxctl/central`.

## Testing Performed
CI


dhaus67 (approved):

c-du (approved):

.golangci.yml
```diff
@@ -15,7 +15,12 @@ issues:
     - path: "(central|compliance|integration-tests|local|migrator|operator|pkg|sensor|tests|tools|scale)/"
       linters:
         - wrapcheck
+    - path: "(central/graphql/schema/print|compliance|integration-tests|local|migrator|operator|pkg|sensor|tests|tools|scale)/"
```
md2119, line 18: I think we should not exclude sensor. If there are `fmt.Print`, we should change them to `errors.Info/Warn...`.  Nonetheless, it can come in a follow-up PR.

md2119 (approved):

PR #3763, opened 2022-11-09
ROX-13464 adding flows dropdown in NG

alwayshooin:
also adding placeholder buttons for CIDR blocks/netpol simulator

image
image

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/FlowsSelect.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/FlowsSelect.tsx
@@ -1,0 +1,46 @@
+import React, { useState } from 'react';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+export type FlowsState = 'active' | 'extraneous';
+type FlowsSelectProps = {
+    flowsState: FlowsState;
+    setFlowsState: (state) => void;
+};
+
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/FlowsSelect.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from 'react';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+export type FlowsState = 'active' | 'extraneous';
```
sachaudh, line 4: So I was a little confused since we use the term `flows` for the extraneous flows and `traffic` for the active traffic. So I think `flows` might not be the best wording for this. I talked to Alan and I think he was good with the term `EdgeState`. Could we change the `Flows` prefix to that?

sachaudh (changes requested):

ui/apps/platform/src/Containers/NetworkGraph/FlowsSelect.tsx
sachaudh: He says `EdgeType` would also work

sachaudh (approved):
lgtm

PR #3764, opened 2022-11-09
cleanup image digest utilities

RTann:
## Description

I noticed a few things:

* There were two utility functions which did the same thing: `GetImageID` and `GetSHA`
    * I decided to converge on `GetSHA` to, hopefully, make it clear the `(*storage.Image).Id` is the image's SHA digest
* SHA 512 digests were not supported by `ExtractImageDigest`
    * See https://github.com/opencontainers/image-spec/blob/main/descriptor.md#registered-algorithms. SHA 512 is a valid image digest algorithm
* `NewDigest` defaults to SHA 256
    * I didn't want to break anything, so I documented this is the behavior, so it's clear

## Checklist

## Testing Performed

CI


pkg/images/types/digest.go
```diff
@@ -32,7 +34,7 @@ func (d *Digest) Algorithm() string {
 	if d == nil {
 		return ""
 	}
-	return d.algorithm + ":" + d.hash
```
md2119, line 35: What!

md2119 (approved):

pkg/images/types/digest.go
RTann: looks like this is never used, so it was never caught before :(

PR #3769, opened 2022-11-10
Freeze 3.73 schema in migrator

c-du:
## Description

This PR wont merge before 3.73 branching out.
Freeze 3.73 schema in migrator.
I did not remove the auto generator because I want to keep that in a clean PR dedicated for stopping generators so we can conveniently re-use that PR in case we want to make massive change.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


tools/generate-helpers/pg-table-bindings/migration.go.tpl
```diff
@@ -19,7 +19,8 @@ import (
 	pgStore "github.com/stackrox/rox/migrator/migrations/{{.Migration.Dir}}/postgres"
 	"github.com/stackrox/rox/migrator/types"
 	pkgMigrations "github.com/stackrox/rox/pkg/migrations"
-	pkgSchema "github.com/stackrox/rox/pkg/postgres/schema"
+	pkgSchema "github.com/stackrox/rox/migrator/migrations/frozenschema/v73"
```
dashrews78, line 22: I think we should call this frozenSchema to make it stand out.

dashrews78:
a README or something might be useful to explain the reasoning for separating the schema as well as how migrations work going forward.  Things like when we lock the migration schema, how we work migrations for the next release.  How that impacts the generated code.

c-du:
> a README or something might be useful to explain the reasoning for separating the schema as well as how migrations work going forward. Things like when we lock the migration schema, how we work migrations for the next release. How that impacts the generated code.

Nice idea. Will do

dashrews78 (approved):
LGTM

migrator/migrations/frozenschema/v73/groups.go
```diff
@@ -0,0 +1,35 @@
...
+	CreateTableGroupsStmt = &postgres.CreateStmts{
+		GormModel: (*Groups)(nil),
+		Children:  []*postgres.CreateStmts{},
+	}
+
+	// GroupsSchema is the go schema for table `groups`.
+	GroupsSchema = func() *walker.Schema {
+		schema := walker.Walk(reflect.TypeOf((*storage.Group)(nil)), "groups")
+		return schema
+	}()
```
md2119, line 24: The schema seems different. The registration to global table is missing:
```	
GroupsSchema = func() *walker.Schema {
		schema := GetSchemaForTable("groups")
		if schema != nil {
			return schema
		}
		schema = walker.Walk(reflect.TypeOf((*storage.Group)(nil)), "groups")
		RegisterTable(schema, CreateTableGroupsStmt)
		return schema
	}()

```

md2119 (changes requested):
The frozen schema seems to be differ than the ones in `pkg/postgres/schema`.

migrator/migrations/frozenschema/v73/groups.go
c-du: That is on purpose. The dependency is pre-defined and the tables are not to be reused.

md2119: Okay. What I am worried about is that event if the migrator and its schema is frozen, one can still make changes to the schema/proto which actually will go unhandled. I think we should add a check which scan `pkg/schema/` dir for any changes. If the changes are desirable, the file is added to a whitelist. WDYT?

c-du: The migration should just have the schema at 3.73. It should not contain future changes to schema. That is what we expect.

md2119 (approved):
Approving because I do not want to block the work, but I feel a required check that fails the CI would be better for maintenance than dropping schema into new pkg.

PR #3776, opened 2022-11-10
Remove bulk delete from collections table

dvail:
## Description

This removes the "Bulk delete" feature of the collections table in order to maintain simplicity in the user's workflow.

## Checklist

## Testing Performed

Navigate to the collection page and ensure the checkbox table cells and "Bulk action" dropdown are no longer present when logged in as a user with read-write collection permissions.
image

Perform a single delete of a collection by using the action menu on the right side of the table.
image

Deleting the collection should succeed and refresh the table:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
@@ -4,28 +4,21 @@ import {
     Bullseye,
     Button,
     ButtonVariant,
-    Dropdown,
-    DropdownItem,
-    DropdownToggle,
     Pagination,
     SearchInput,
     Text,
     Toolbar,
     ToolbarContent,
     ToolbarItem,
-    ToolbarItemVariant,
     Truncate,
 } from '@patternfly/react-core';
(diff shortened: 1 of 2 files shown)


dvail:
* master
  * **PR #3776** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3776" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3776?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionsTable.tsx
```diff
@@ -91,29 +81,18 @@ function CollectionsTable({
         [setSearchFilter]
     );
 
-    function onConfirmDeleteCollection() {
+    function onConfirmDeleteCollection(collection: CollectionResponse) {
```
dvail, line 84: By passing the `collection` here instead of reading from state, we can take advantage of the fact that the type has already been narrow to be non-null.

pedrottimark (approved):
Who will delete the deleter? You did.

PR #3777, opened 2022-11-10
Validate that label keys are valid k8s labels and ensure correct key splitting

dvail:
## Description

This hardens the FE implementation of label rules with the following changes:

1. Label keys no longer accept any non-empty string. Keys must only contain valid k8s label characters.
2. When parsing label values, we now take into account the possibility that the delimiter (`=`) can be found in the value, by only splitting on the first occurance of this character.

## Checklist

## Testing Performed

Unit tests added.

In the collection form, test that valid values are accepted by formik:
image
image
image

Test that invalid values are not accepted:
image
image
image


pedrottimark (approved):
Thank you for writing unit tests.

PR #3780, opened 2022-11-11
ROX-13168: Add traffic filters

sachaudh:
## Description

I'm creating a new component called `AdvancedFlowsFilter` which will contain a dropdown for filtering by deployment flows, directionality, protocols, ports

Screenshot 2022-11-11 at 10 43 03 AM

**Reference Mocks:** https://www.sketch.com/s/9d082d82-3515-40f2-875a-1b322a02bace/a/zxLMR2W

This draft PR is being sent out to get thoughts on the following:

1. **What do we think about having this new component to manage the filters?** It will take care of organizing the filter data into a data structure so that the parent component can access it in a reasonable format
2. **Where should this new component live?** I placed it in a new directory called `flows`. I considered creating a `DeploymentFlows` directory and placing `DeploymentFlows.tsx`, `DeploymentFlows.css`, and `AdvancedFlowsFilters` within that directory, but wasn't sure if nesting too much would be too much. Putting it within the `flows` directory seems reasonable since we seem to be separating components by sub-features so far

**More things I need to do:**
1. Add tests for the helper functions used to organize the filter data

## Checklist

## Testing Performed

Screenshot 2022-11-11 at 10 43 46 AM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
@@ -5,3 +5,7 @@
 tbody tr.pf-c-table__expandable-row td.pf-c-table__check label {
     padding: 0;
 }
+
+.advanced-flows-filters-select .pf-c-select__menu {
+    min-width: 300px;
+}
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {
+    Divider,
     DropdownItem,
     Flex,
     FlexItem,
@@ -25,8 +26,11 @@ import {
 } from '@patternfly/react-table';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 
-import './DeploymentFlows.css';
 import BulkActionsDropdown from 'Components/PatternFly/BulkActionsDropdown';
+import AdvancedFlowsFilter, { defaultAdvancedFlowsFilters } from '../flows/AdvancedFlowsFilter';
+
+import './DeploymentFlows.css';
+import { AdvancedFlowsFilterType } from '../flows/types';
 
 interface FlowBase {
     id: string;
@@ -118,16 +122,19 @@ const flows: Flow[] = [
 ];
 
 function DeploymentFlow() {
-    // derived values
+    // component state
+    const [advancedFilters, setAdvancedFilters] = React.useState<AdvancedFlowsFilterType>(
+        defaultAdvancedFlowsFilters
+    );
+    const initialExpandedRows = flows.filter((row) => !!row.children.length).map((row) => row.id); // Default to all expanded
+    const [expandedRows, setExpandedRows] = React.useState<string[]>(initialExpandedRows);
+    const [selectedRows, setSelectedRows] = React.useState<string[]>([]);
+
+    // derived data
     const totalFlows = flows.reduce((acc, curr) => {
         // if there are no children then it counts as 1 flow
         return acc + (curr.children.length ? curr.children.length : 1);
     }, 0);
-
-    // component state
-    const initialExpandedRows = flows.filter((row) => !!row.children.length).map((row) => row.id); // Default to all expanded
-    const [expandedRows, setExpandedRows] = React.useState<string[]>(initialExpandedRows);
-    const [selectedRows, setSelectedRows] = React.useState<string[]>([]);
 
     // getter functions
     const isRowExpanded = (row: Flow) => expandedRows.includes(row.id);
@@ -169,6 +176,18 @@ function DeploymentFlow() {
     return (
         <div className="pf-u-h-100 pf-u-p-md">
             <Stack hasGutter>
+                <StackItem>
+                    <Flex>
+                        <FlexItem flex={{ default: 'flex_1' }} />
+                        <FlexItem>
+                            <AdvancedFlowsFilter
+                                filters={advancedFilters}
+                                setFilters={setAdvancedFilters}
+                            />
+                        </FlexItem>
+                    </Flex>
+                </StackItem>
+                <Divider component="hr" />
                 <StackItem>
                     <Toolbar>
                         <ToolbarContent>
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
@@ -25,6 +25,7 @@ import {
     Tr,
 } from '@patternfly/react-table';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
+import { uniq } from 'lodash';
 
 import BulkActionsDropdown from 'Components/PatternFly/BulkActionsDropdown';
 import AdvancedFlowsFilter, { defaultAdvancedFlowsFilters } from '../flows/AdvancedFlowsFilter';
@@ -92,7 +93,7 @@ const flows: Flow[] = [
         entity: 'Deployment 1',
         namespace: 'naples',
         direction: 'Ingress',
-        port: 'Many',
+        port: '9000',
         protocol: 'TCP',
         isAnomalous: true,
         children: [],
@@ -103,7 +104,7 @@ const flows: Flow[] = [
         entity: 'Deployment 2',
         namespace: 'naples',
         direction: 'Ingress',
-        port: 'Many',
+        port: '8080',
         protocol: 'UDP',
         isAnomalous: false,
         children: [],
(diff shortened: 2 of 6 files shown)


ui/apps/platform/src/Containers/NetworkGraph/flows/FlowsFilterToolbar.tsx
```diff
@@ -0,0 +1,148 @@
...
+                            variant={SelectVariant.typeaheadMulti}
+                            aria-label="Select one or more days"
+                            onToggle={onTogglePortsSelect}
+                            onSelect={onSelectPorts}
+                            selections={filters.ports}
+                            isOpen={isPortsSelectOpen}
+                            placeholderText="Select ports"
+                            isCreatable
+                            createText="Select port"
+                        />
```
sachaudh, line 140: I guess I could actually pass all the ports to this component and display them as an option here

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.css
```diff
@@ -5,3 +5,7 @@
 tbody tr.pf-c-table__expandable-row td.pf-c-table__check label {
     padding: 0;
 }
+
+.advanced-flows-filters-select .pf-c-select__menu {
```
sachaudh, line 9: If I don't set a min-width, the dropdown is too narrow and the ports select doesn't fit so nicely

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
```diff
@@ -169,6 +178,18 @@ function DeploymentFlow() {
     return (
         <div className="pf-u-h-100 pf-u-p-md">
             <Stack hasGutter>
+                <StackItem>
+                    <Flex>
+                        <FlexItem flex={{ default: 'flex_1' }} />
```
vjwilson, line 183: What does this empty flex item do?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
```diff
@@ -169,6 +178,18 @@ function DeploymentFlow() {
...
+                        <FlexItem>
+                            <AdvancedFlowsFilter
+                                filters={advancedFilters}
+                                setFilters={setAdvancedFilters}
+                            />
```
vjwilson, line 188: Would you describe how you envision using these filters to filter the data?
It seems this PR is mainly for getting the presentation of the filter dropdown working, but it would be nice to know how you see it fitting into the data flow.

Doesn't need to be a comment in the code (unless you think it deserves a permanent code comment even when implemented). A PR comment, or adding it to the PR description, would be fine.

ui/apps/platform/src/Containers/NetworkGraph/flows/AdvancedFlowsFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+// which the Select component expects
+function filtersToSelections(filters: AdvancedFlowsFilterType) {
+    return Object.values(filters).reduce((acc, curr) => {
+        return [...acc, ...Object.values(curr)];
+    }, []);
+}
+
+// This function is used to convert the selections array of strings, which the Select component
+// expects, into our filters data structure
+function selectionsToFilters(selections: string[]) {
```
vjwilson, line 41: Might be cleaner to pull these two functions out into a helper file, especially if you think they will eventually be reused in other filters in the new network graph.

Factoring out would allow for quick yet comprehensive unit tests of all filter type combinations, too, and allow fewer e2e tests for same code coverage.

ui/apps/platform/src/Containers/NetworkGraph/flows/AdvancedFlowsFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+            <SelectGroup label="Flow directionality">
+                <SelectOption value="ingress">Ingress (inbound)</SelectOption>
+                <SelectOption value="egress">Egress (outbound)</SelectOption>
+            </SelectGroup>
+            <SelectGroup label="Protocols">
+                <SelectOption value="TCP">TCP</SelectOption>
+                <SelectOption value="UDP">UDP</SelectOption>
+            </SelectGroup>
+            <SelectGroup label="Ports">
+                <Select
```
vjwilson, line 132: This is a bit of an odd UX flow. Are we planning to later "pre-seed" the list of ports, based on what's available? Otherwise, every port will be a new "createable" entry--in which case, is there a PF text-input-with-chips component that would be clearer?

@alanonthegit The UX team may have already tried different solutions and found this was the best available, but I figure it's worth asking.

ui/apps/platform/src/Containers/NetworkGraph/flows/AdvancedFlowsFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    className="pf-u-px-md"
+                    variant={SelectVariant.typeaheadMulti}
+                    aria-label="Select one or more days"
```
vjwilson, line 135: The copy-paste gremlin is messing with the aircraft engine again.

```suggestion
                    aria-label="Select one or more ports"
```

vjwilson (commented):
Overall approach looks good. Left a few questions and suggestions.

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
sachaudh: It's just a placeholder for the eventual search input that I'll add soon

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentFlows.tsx
sachaudh: The component will keep track of these `advanced` filters we will use to filter the flows data. We also have a search input we'll use to filter by entity name. Using both of these values, we can have some sort of util function that takes these filters and the flows data, and just return the flows that match those filters. That will happen within the `DeploymentFlows` component

ui/apps/platform/src/Containers/NetworkGraph/flows/AdvancedFlowsFilter.tsx
sachaudh: Yea, I was thinking about that after I sent out the PR (https://github.com/stackrox/stackrox/pull/3780#discussion_r1019775427). I can actually pass all unique ports as select options. I think that would help make the UX feel less odd

sachaudh:
Hello @alwayshooin @vjwilson, just wanted to get this one reviewed soon

ui/apps/platform/src/Containers/NetworkGraph/flows/advancedFlowsFilterUtils.ts
```diff
@@ -0,0 +1,42 @@
+import { AdvancedFlowsFilterType, FilterValue } from './types';
+
+function isValidPort(port: string): boolean {
+    const portNum = parseInt(port, 10);
+    if (Number.isNaN(portNum)) {
+        return false;
+    }
+    return portNum >= 1 && portNum <= 65535;
```
alwayshooin, line 8: just curious, where are you getting these numbers? is it possible to get non-valid port numbers from the API resopnse?

ui/apps/platform/src/Containers/NetworkGraph/flows/advancedFlowsFilterUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+            filters.flows.push(selection);
+        } else if (selection === 'ingress' || selection === 'egress') {
+            filters.directionality.push(selection);
+        } else if (selection === 'TCP' || selection === 'UDP') {
+            filters.protocols.push(selection);
+        } else if (isValidPort(selection)) {
+            filters.ports.push(selection);
+        } else {
+            // do nothing
+        }
```
alwayshooin, line 39: why not use a switch statement here and push to filters.ports as default? an empty else clause seems odd to see

alwayshooin (approved):
looks fine to me, just some q in the code

ui/apps/platform/src/Containers/NetworkGraph/flows/advancedFlowsFilterUtils.ts
sachaudh: I'm assuming we will always get valid ports, but this is more to distinguish what is a port value in that `selections` array

ui/apps/platform/src/Containers/NetworkGraph/flows/advancedFlowsFilterUtils.ts
sachaudh: We can but I'd like to do it like this in case we add something else to the filters. If we don't distinguish each filter, we might accidentally think something is a port that shouldn't be

PR #3787, opened 2022-11-11
[ROX-13500] Perform type check for V1 CronJob

md2119:
## Description

This PR fixes Central panic when deployment YAML check is run for V1 CronJob API due to insufficient type check and nil check.  V1 Beta1 CronJob API was deprecated in k8s 1.21 and was promoted to V1. 

## Checklist

## Testing Performed
Unit Test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - The service account "central", which is used by the central deployment, will now include `get` and `list` access to the following resources in the namespace where central is deployed to:
   `pods`, `events`, and `namespaces`. This fixes an issue when generating diagnostic bundles to now correctly include all relevant information within the namespace of central.
 - ROX-13265: Fix missing rationale and remediation texts for default policy "Deployments should have at least one ingress Network Policy"
-- ROX-13500: Previously, deployment YAML check on V1 CronJob workload would cause Central to panic. This has is now fixed.
+- ROX-13500: Previously, deployment YAML check on V1 CronJob workload would cause Central to panic. This is now fixed.
 
 ## [3.72.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - The service account "central", which is used by the central deployment, will now include `get` and `list` access to the following resources in the namespace where central is deployed to:
   `pods`, `events`, and `namespaces`. This fixes an issue when generating diagnostic bundles to now correctly include all relevant information within the namespace of central.
 - ROX-13265: Fix missing rationale and remediation texts for default policy "Deployments should have at least one ingress Network Policy"
+- ROX-13500: Previously, deployment YAML check on V1 CronJob workload would cause Central to panic. This has is now fixed.
 
 ## [3.72.0]
 
--- a/central/detection/service/service_impl_test.go
+++ b/central/detection/service/service_impl_test.go
@@ -530,7 +530,7 @@ spec:
   adminPortalCredentialsRef:
     name: asecretname
 `
-const exampleYaml = `
+const cronYaml = `
 apiVersion: batch/v1
 kind: CronJob
 metadata:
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - The service account "central", which is used by the central deployment, will now include `get` and `list` access to the following resources in the namespace where central is deployed to:
   `pods`, `events`, and `namespaces`. This fixes an issue when generating diagnostic bundles to now correctly include all relevant information within the namespace of central.
 - ROX-13265: Fix missing rationale and remediation texts for default policy "Deployments should have at least one ingress Network Policy"
+- ROX-13500: Previously, deployment YAML check on V1 CronJob workload would cause Central to panic. This has is now fixed.
```
RTann, line 51: I don't think this is needed, since the ticket already talks about this

central/detection/service/service_impl_test.go
```diff
@@ -530,6 +530,27 @@ spec:
   adminPortalCredentialsRef:
     name: asecretname
 `
+const exampleYaml = `
```
RTann, line 533: nit: maybe rename to cronYaml to make it clear what this is an example of

pkg/protoconv/resources/resources.go
```diff
@@ -237,8 +241,11 @@ func (w *DeploymentWrap) populateFields(obj interface{}) {
 		// types do. So, we need to directly access the Pod's Spec field,
 		// instead of looking for it inside a PodTemplate.
 		podSpec = o.Spec
+	// batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+.
 	case *batchV1beta1.CronJob:
 		podSpec = o.Spec.JobTemplate.Spec.Template.Spec
+	case *batchV1.CronJob:
```
RTann, line 247: since these are the same, should we consider doing

```
case *batchV1beta1.CronJob, *batchV1.CronJob:
        podSpec = o.Spec.JobTemplate.Spec.Template.Spec
```

instead?

pkg/protoconv/resources/resources.go
```diff
@@ -122,6 +123,9 @@ func newWrap(meta metav1.Object, kind, clusterID, registryOverride string) *Depl
 // SpecToPodTemplateSpec turns a top level spec into a podTemplateSpec
 func SpecToPodTemplateSpec(spec reflect.Value) (v1.PodTemplateSpec, error) {
 	templateInterface := spec.FieldByName("Template")
+	if !doesFieldExist(spec) {
```
RTann, line 126: Do you mean `!doesFieldExist(templateInterface)`? Otherwise, this should be above the `templateInterface := spec.FieldByName("Template")` line

CHANGELOG.md
md2119: RH JIRA is not externally exposed. This just makes it a lot easier for docs team based on past experience. :)

pkg/protoconv/resources/resources.go
md2119: No, Go needs to know the exact type that is being upon so case list does not work. I tried.

pkg/protoconv/resources/resources.go
md2119: Ah, good catch.

CHANGELOG.md
dhaus67, line 51:
```suggestion
- ROX-13500: Previously, deployment YAML check on V1 CronJob workload would cause Central to panic. This is now fixed.
```

central/detection/service/service_impl_test.go
```diff
@@ -543,6 +564,9 @@ func TestParseList_Success(t *testing.T) {
 
 	_, _, err = getObjectsFromYAML(openshiftDeploymentConfigYaml)
 	require.NoError(t, err)
+
+	_, _, err = getObjectsFromYAML(cronYaml)
```
dhaus67, line 568: nit: could rewrite this to a loop instead of copy-paste, but don't think it's critical.
```go
yamlToTest := []string{listYAML, openshiftDeploymentConfigYaml, multiYaml, openshiftDeploymentConfigYaml, cronYaml}

for _, yaml := range yamlToTest {
        _, _, err := getObjectsFromYAML(yaml)
        require.NoError(t, err)
} 
```

dhaus67 (approved):

RTann (approved):

PR #3789, opened 2022-11-11
Add migration for groups with invalid values

dhaus67:
## Description

Within the initial PR to migrate the groups bucket from a composite key created by the `storage.GroupProperties` set on the group to a UUID, groups that had an empty role name have been missed during the migration.

Those have surfaced recently after looking at a ticket ([ROX-13446](https://issues.redhat.com/browse/ROX-13446)), where the groups bucket contained two groups with properties set to `nil` and an empty role name.

Irrespective of how the groups got there in the first place, the first migration missed to add an ID to groups with a) either no role name set or b) no properties being set.

After adding validation, groups that either do not reference a role name or an auth provider are considered as invalid.

This PR creates a migration that specifically removes those groups.

## Checklist
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into 
[rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed

- see unit tests added for the migration.

Additionally, the following tests were done manually to ensure the migration worked as expected:
1. Create a central which inserts invalid data on startup (I used a custom build that added those improper groups with nil properties and empty role name).
2. You should observe the following output when querying the groups endpoint:
```
roxcurl  /v1/groups
{"groups":[{"props":null,"roleName":""},{"props":{"authProviderId":"177332fb-c6ae-4062-ab60-54b42cdeb024","key":"","value":""},"roleName":""}
```
3. Upgrade to this version.
4. Query the groups endpoint and verify that the previously invalid entries are not available anymore.
```
roxcurl  /v1/groups
{}
```

### DIFF
--- a/migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
+++ b/migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
@@ -1,0 +1,55 @@
+package binenc
+
+import (
+	"bytes"
+	"encoding/binary"
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+// WriteBytesList takes a list of byte slices and writes them in encoded form to a writer.
+func WriteBytesList(w io.Writer, byteSlices ...[]byte) (int, error) {
+	var total int
+	for _, byteSlice := range byteSlices {
+		n, err := WriteUVarInt(w, uint64(len(byteSlice)))
+		total += n
+		if err != nil {
+			return total, err
+		}
+
+		n, err = w.Write(byteSlice)
+		total += n
+		if err != nil {
+			return total, err
+		}
+	}
+	return total, nil
+}
+
+// EncodeBytesList takes a list of byte slices and encodes them into a single byte slice.
+func EncodeBytesList(byteSlices ...[]byte) []byte {
+	var buf bytes.Buffer
+	_, _ = WriteBytesList(&buf, byteSlices...)
+	return buf.Bytes()
+}
+
+// DecodeBytesList takes a byte buffer encoded via EncodeBytesList or WriteBytesList and decodes it into a list of byte
+// slices.
+func DecodeBytesList(buf []byte) ([][]byte, error) {
+	var result [][]byte
+
+	for len(buf) > 0 {
+		sliceLen, l := binary.Uvarint(buf)
+		if l <= 0 {
+			return nil, errors.New("invalid varint in buffer")
+		}
+		buf = buf[l:]
+		if sliceLen > uint64(len(buf)) {
+			return nil, errors.Errorf("encountered varint %d which is larger than the remaining buffer size (%d)", sliceLen, len(buf))
+		}
+		result = append(result, buf[:sliceLen])
+		buf = buf[sliceLen:]
+	}
+	return result, nil
+}
--- a/migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
+++ b/migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
@@ -1,0 +1,99 @@
+package m111tom112
+
+import (
+	"bytes"
+
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/migrator/migrations"
+	"github.com/stackrox/rox/migrator/types"
+	bolt "go.etcd.io/bbolt"
+)
+
+// groupStoredByCompositeKey is a helper struct which contains the group as well as the composite key.
+type groupStoredByCompositeKey struct {
+	grp          *storage.Group
+	compositeKey []byte
+}
+
+var (
+	bucketName = []byte("groups2")
+
+	emptyPropertiesCompositeKey = serializePropsKey(nil)
+
+	migration = types.Migration{
+		StartingSeqNum: 111,
+		VersionAfter:   &storage.Version{SeqNum: 112},
+		Run: func(databases *types.Databases) error {
+			return removeGroupsWithInvalidValues(databases.BoltDB)
+		},
+	}
+)
+
+func init() {
+	migrations.MustRegisterMigration(migration)
+}
+
+func removeGroupsWithInvalidValues(db *bolt.DB) error {
+	groupsWithInvalidValues, err := fetchGroupsToRemove(db)
+	if err != nil {
+		return errors.Wrap(err, "error fetching groups to remove")
+	}
+
+	if err := removeGroupsStoredByCompositeKey(db, groupsWithInvalidValues); err != nil {
+		return errors.Wrap(err, "error removing groups with invalid values")
+	}
+
+	return nil
+}
+
+func fetchGroupsToRemove(db *bolt.DB) (groupsStoredByCompositeKey []groupStoredByCompositeKey, err error) {
+	err = db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(bucketName)
+		// Pre-req: Migrating a non-existent bucket should not fail.
+		if bucket == nil {
+			return nil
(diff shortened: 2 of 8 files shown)


dhaus67:
* master
  * **PR #3789** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3789" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3789?utm_source=stack-comment).

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
```diff
@@ -0,0 +1,99 @@
...
+		bucket := tx.Bucket(bucketName)
+		// Pre-req: Migrating a non-existent bucket should not fail.
+		if bucket == nil {
+			return nil
+		}
+		return bucket.ForEach(func(k, v []byte) error {
+			// In a prior migration, the groups bucket was migrated from groups being stored by a composite key to groups
+			// being stored by a UUID.
+			// Within that migration, groups that had an empty role name were skipped during migration.
+			// This lead to bucket entries where the properties and role name was both empty, thus making it impossible
```
ivan-degtiarenko, line 61: Should we also delete groups with empty role names here as a pre-caution?

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
```diff
@@ -0,0 +1,99 @@
...
+		bucket := tx.Bucket(bucketName)
+		// Pre-req: Migrating a non-existent bucket should not fail.
+		if bucket == nil {
+			return nil
+		}
+
+		for i := range groupStoredByCompositeKeys {
+			compositeKey := groupStoredByCompositeKeys[i].compositeKey
+
+			// 1. Remove the value stored behind the composite key, since the migrated group is now successfully stored.
```
ivan-degtiarenko, line 91: There's no `2`

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
```diff
@@ -0,0 +1,99 @@
...
+					return err
+				}
+
+				groupsStoredByCompositeKey = append(groupsStoredByCompositeKey,
+					groupStoredByCompositeKey{grp: grp, compositeKey: k})
+			}
+			return nil
+		})
+	})
+	return groupsStoredByCompositeKey, err
```
ivan-degtiarenko, line 77: let's explicitly return `nil` instead of `err` here

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
```diff
@@ -0,0 +1,99 @@
...
+
+func removeGroupsStoredByCompositeKey(db *bolt.DB, groupStoredByCompositeKeys []groupStoredByCompositeKey) error {
+	return db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(bucketName)
+		// Pre-req: Migrating a non-existent bucket should not fail.
+		if bucket == nil {
+			return nil
+		}
+
+		for i := range groupStoredByCompositeKeys {
```
ivan-degtiarenko, line 88: What if we use `for _, group := range groupStoredByCompositeKeys`? Then we don't need to address array value by index within the loop

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
dhaus67: We do delete groups with empty role names as long as they are still stored keyed by the composite key.

I don't think we need to remove groups with empty role names when they are stored using a unique ID - currently it would be possible to delete those via API directly (I also don't expect those to be there, but then again neither did I with the current situation). Still, I'd say let's not do this and add additional complexity to the migration.

ivan-degtiarenko: For me, empty role name = invalid group = should be deleted. This corresponds with the goal of this migration as I see it - fix our previous mistake and return the groups bucket to the valid state. Maybe I understood it wrong and we delete somewhere groups with an empty role, but definitely not during startup:
https://github.com/stackrox/stackrox/blob/6113b92dce180cd7856b1917569fb3fd18786037/central/group/datastore/singleton.go#L22-L27

I won't fixate on this though. If you feel strongly about not adding this to migration right now, let's put it on hold - for me, it makes sense to do it now due to our release cadence.

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration_test.go
```diff
@@ -0,0 +1,151 @@
...
+			oldValue:    true,
+		},
+		{
+			storedGroup: expectedGroups["r4"],
+			newGroup:    expectedGroups["r4"],
+			oldValue:    true,
+		},
+	}
+
+	// 1. Buckets don't exist should succeed still
```
ivan-degtiarenko, line 91: super-nit

```suggestion
	// 1.  Migration should succeed if the bucket does not exist
```

migrator/migrations/m_111_to_m_112_groups_invalid_values/migration.go
```diff
@@ -0,0 +1,105 @@
...
+	return db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(bucketName)
+		// Pre-req: Migrating a non-existent bucket should not fail.
+		if bucket == nil {
+			return nil
+		}
+
+		var deleteGroupErrs *multierror.Error
+		for _, group := range groupStoredByCompositeKeys {
+			compositeKey := group.compositeKey
```
ivan-degtiarenko, line 94: nit: can be inlined

ivan-degtiarenko (approved):
LGTM modulo couple of nits

migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
```diff
@@ -0,0 +1,55 @@
...
+		total += n
+		if err != nil {
+			return total, err
+		}
+	}
+	return total, nil
+}
+
+// EncodeBytesList takes a list of byte slices and encodes them into a single byte slice.
+func EncodeBytesList(byteSlices ...[]byte) []byte {
```
janisz, line 31: Who is using it?

migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
```diff
@@ -0,0 +1,55 @@
...
+	var buf bytes.Buffer
+	_, _ = WriteBytesList(&buf, byteSlices...)
+	return buf.Bytes()
+}
+
+// DecodeBytesList takes a byte buffer encoded via EncodeBytesList or WriteBytesList and decodes it into a list of byte
+// slices.
+func DecodeBytesList(buf []byte) ([][]byte, error) {
```
janisz, line 39: Who is using it?

migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
dhaus67: I followed the same examples for copying stackrox code into the migrations - which is copying the package as is and not modifying it, even though some functions may be unused.

migrator/migrations/m_111_to_m_112_groups_invalid_values/binenc/byte_slice_list.go
dhaus67: Same as above.

PR #3793, opened 2022-11-11
[ROX-13441][POSTGRES] Propagate context correctly in retries

md2119:
## Description

Noticed that the context canceled on Postgres operations. Consequently, search operations seem to be failing and causing central log spam:
```
pkg/search/postgres: 2022/11/10 20:57:23.714605 common.go:628: Error: Query issue: select alerts.Id::text from alerts where (alerts.Deployment_Id = $1 and ((alerts.State = $2) or (alerts.State = $3))) order by alerts.Time desc [4169b3e6-2d50-4243-be15-0faa239e5ed1 0 3]: context canceled
```
This seems to have originated from incorrect context propagation in Postgres operation retries.

## Checklist
- ~Unit test and regression tests added~

## Testing Performed
non-trivial. Will monitor the CI on master commits.

### DIFF
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/pkg/buildinfo"
+	"github.com/stackrox/rox/pkg/contextutil"
 	"github.com/stackrox/rox/pkg/devbuild"
 	"github.com/stackrox/rox/pkg/errox"
 	"github.com/stackrox/rox/pkg/logging"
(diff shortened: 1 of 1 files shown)


dashrews78 (approved):

c-du (approved):

dhaus67 (approved):

pkg/search/postgres/common.go
```diff
@@ -581,6 +582,7 @@ func RunSearchRequest(ctx context.Context, category v1.SearchCategory, q *v1.Que
 	schema := mapping.GetTableFromCategory(category)
 
 	return pgutils.Retry2(func() ([]searchPkg.Result, error) {
+		ctx := contextutil.WithValuesFrom(context.Background(), ctx)
```
janisz, line 585: I don't understand this :(
Why we need to use `context.Background` here? How do we propagate timeouts or cancelation then?

PR #3797, opened 2022-11-14
ROX-12814: Disable PolicyFieldsTest on openshift.

porridge:
## Description

Scanner OOMs pretty reliably on this test on osd jobs for over 2 weeks.
Disabling for now according to [our policy](https://srox.slack.com/archives/CELUQKESC/p1667906225265369).

## Checklist
- [ ] Investigated and inspected CI test results to make sure this test was skipped

## Testing Performed

Relying on CI.
/test all

### DIFF
--- a/qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
@@ -23,7 +23,7 @@ import spock.lang.Shared
 import spock.lang.Unroll
 import util.Env
 
-// TODO(ROX-12814): Scanner OOMs on this test in some Openshift jobs.
+// Scanner OOMs on this test in some Openshift jobs. See ROX-12814.
 @IgnoreIf({ Env.mustGetOrchestratorType() == OrchestratorTypes.OPENSHIFT })
 class PolicyFieldsTest extends BaseSpecification {
 
--- a/qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
@@ -18,10 +18,13 @@ import org.junit.Assume
 import org.junit.experimental.categories.Category
 import services.AlertService
 import services.PolicyService
+import spock.lang.IgnoreIf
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
```diff
@@ -18,10 +18,13 @@ import org.junit.Assume
 import org.junit.experimental.categories.Category
 import services.AlertService
 import services.PolicyService
+import spock.lang.IgnoreIf
 import spock.lang.Shared
 import spock.lang.Unroll
 import util.Env
 
+// TODO(ROX-12814): Scanner OOMs on this test in some Openshift jobs.
+@IgnoreIf({ Env.mustGetOrchestratorType() == OrchestratorTypes.OPENSHIFT })
```
gavin-stackrox, line 27: What about only disabling for OSD GCP? Other openshifts do not have this OOM IIRC?
```suggestion
@IgnoreIf({ Env.CI_JOBNAME.contains("osd-gcp") })
```

gavin-stackrox (changes requested):

qa-tests-backend/src/test/groovy/PolicyFieldsTest.groovy
porridge: There are mentions of failures on `rosa` and `openshift-oldest` too in the ticket :-/

gavin-stackrox (approved):
Sorry! I missed those in the hidden comments!

janisz (approved):

porridge:
Looks like the set of tests that run on per-submit is smaller than the set run on nightlies.
While this test is explicitly mentioned [in this nightly `openshift-newest-qa` run](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-stackrox-nightlies-openshift-newest-qa-e2e-tests/1592022230034812928) for example, it is not mentioned at all in the corresponding run for this PR (rather than mentioned as skipped).
So I'm going to merge this, since checking tomorrow's nightly is probably the least effort way to check that this change worked as intended.

PR #3800, opened 2022-11-14
ROX-13540: Allow Sensor Component `NodeScanHandler` to start max. once 

vikin91:
## Description

This is a follow-up PR to address review comments from #3533.
The original PR was merged by accident without addressing optional but
good suggestions.

This PR changes the NodeScan component to start maximally once.

It includes:
- ~🗑️ Removing `stoppedC` leftover~ (readded in d7353aed4c076ee0b9e3a9ad492ab84d28a93433)
- 🐛 Handling a panic when two or more consecutive Start() calls happen without a Stop() call in-between
- 🧪 Adding a regression test for the two or more consecutive Start() calls
- 🐛 Fixing previously undiscovered race condition
- 💄 Removing one level of indentation in `sendScan`
- 💄 Using `s.Assert()` instead `assert.`
- ~✨ (Extra) Making the component restartable~ (removed in d7353aed4c076ee0b9e3a9ad492ab84d28a93433)
- Re-added `Stopped()` so that the tests can wait for the handler to finish work. 

### Extra

After the first [comment about not closing the `toCentral` channel](https://github.com/stackrox/stackrox/pull/3800/files#r1023454649), I decided that the only clean solution to that challenge it to make the component restartable. I added the restartability in https://github.com/stackrox/stackrox/pull/3800/commits/be4bb7cde40d2548bc71fa84d6fd1a3b158d3752.

## Checklist

### Not applies

## Testing Performed

- [x] CI
- [x] `go test -race -count=1 -v -run ^TestNodeScanHandler$
  ./sensor/common/compliance`
- [x] `go test -race -count=12 -v -run ^TestNodeScanHandler$
  ./sensor/common/compliance` (more executions at once)
- [x] Manual deployment of cluster locally, enabling the FF (`ROX_RHCOS_NODE_SCANNING=true)`, and
  confirming that the messages are sent to Sensor and Central. Also
killing sensor pod and making sure that it gets restarted and
self-heals.

### DIFF
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -21,6 +21,5 @@ func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 		stopC:     concurrency.NewErrorSignal(),
 		lock:      &sync.Mutex{},
 		stoppedC:  concurrency.NewErrorSignal(),
-		numStarts: 0,
 	}
 }
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -1,6 +1,7 @@
 package compliance
 
 import (
+	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/sensor/common"
@@ -15,9 +16,7 @@ type NodeScanHandler interface {
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
 		nodeScans: ch,
-		toCentral: nil,
-
-		stopC:    concurrency.NewErrorSignal(),
-		stoppedC: concurrency.NewErrorSignal(),
+		toCentral: make(chan *central.MsgFromSensor),
+		stopC:     concurrency.NewErrorSignal(),
 	}
 }
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -1,7 +1,6 @@
 package compliance
 
 import (
-	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/pkg/sync"
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -20,8 +20,8 @@ func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 		nodeScans: ch,
 		toCentral: make(chan *central.MsgFromSensor),
 		stopC:     concurrency.NewErrorSignal(),
-		startedC:  concurrency.NewSignal(),
 		lock:      &sync.Mutex{},
 		stoppedC:  concurrency.NewErrorSignal(),
+		numStarts: 0,
 	}
 }
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -1,7 +1,6 @@
 package compliance
 
 import (
-	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/sensor/common"
@@ -16,7 +15,7 @@ type NodeScanHandler interface {
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
 		nodeScans: ch,
-		toCentral: make(chan *central.MsgFromSensor),
+		toCentral: nil,
 
 		stopC:    concurrency.NewErrorSignal(),
 		stoppedC: concurrency.NewErrorSignal(),
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -17,11 +17,11 @@ type NodeScanHandler interface {
 // NewNodeScanHandler returns a new instance of a NodeScanHandler
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
-		nodeScans:    ch,
-		toCentral:    make(chan *central.MsgFromSensor),
-		stopC:        concurrency.NewErrorSignal(),
-		startedC:     concurrency.NewSignal(),
-		lock:         &sync.Mutex{},
-		runFinishedC: concurrency.NewErrorSignal(),
+		nodeScans: ch,
+		toCentral: make(chan *central.MsgFromSensor),
+		stopC:     concurrency.NewErrorSignal(),
+		startedC:  concurrency.NewSignal(),
+		lock:      &sync.Mutex{},
+		stoppedC:  concurrency.NewErrorSignal(),
 	}
 }
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -18,7 +18,7 @@ type NodeScanHandler interface {
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
 		nodeScans: ch,
-		toCentral: make(chan *central.MsgFromSensor),
+		toCentral: nil,
 		stopC:     concurrency.NewErrorSignal(),
 		lock:      &sync.Mutex{},
 		stoppedC:  concurrency.NewErrorSignal(),
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -21,7 +21,7 @@ func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 		toCentral:    make(chan *central.MsgFromSensor),
 		stopC:        concurrency.NewErrorSignal(),
 		startedC:     concurrency.NewSignal(),
-		toCentralMux: &sync.Mutex{},
+		lock:         &sync.Mutex{},
 		runFinishedC: concurrency.NewErrorSignal(),
 	}
 }
--- a/sensor/common/compliance/nodescan_handler.go
+++ b/sensor/common/compliance/nodescan_handler.go
@@ -4,19 +4,24 @@ import (
 	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
 	"github.com/stackrox/rox/sensor/common"
 )
 
 // NodeScanHandler is responsible for handling the arriving NodeScanV2 messages, processing then, and sending them to central
 type NodeScanHandler interface {
 	common.SensorComponent
+	Stopped() concurrency.ReadOnlyErrorSignal
 }
 
 // NewNodeScanHandler returns a new instance of a NodeScanHandler
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
-		nodeScans: ch,
-		toCentral: make(chan *central.MsgFromSensor),
-		stopC:     concurrency.NewErrorSignal(),
+		nodeScans:    ch,
+		toCentral:    make(chan *central.MsgFromSensor),
+		stopC:        concurrency.NewErrorSignal(),
+		startedC:     concurrency.NewSignal(),
+		toCentralMux: &sync.Mutex{},
+		runFinishedC: concurrency.NewErrorSignal(),
 	}
 }
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -64,7 +64,9 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
 	toC := make(chan *central.MsgFromSensor)
 	go func() {
-		defer c.stoppedC.SignalWithError(c.stopC.Err())
+		defer func() {
+			c.stoppedC.SignalWithError(c.stopC.Err())
+		}()
 		defer close(toC)
 		for !c.stopC.IsDone() {
 			select {
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -11,7 +11,10 @@ import (
 	"github.com/stackrox/rox/pkg/sync"
 )
 
-var errStartMoreThanOnce = errors.New("unable to start the component more than once")
+var (
+	errInputChanClosed   = errors.New("channel receiving node scans v2 is closed")
+	errStartMoreThanOnce = errors.New("unable to start the component more than once")
+)
 
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
@@ -75,7 +78,7 @@ func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
 				return
 			case scan, ok := <-c.nodeScans:
 				if !ok {
-					c.stopC.SignalWithError(errors.New("channel receiving node scans v2 is closed"))
+					c.stopC.SignalWithError(errInputChanClosed)
 					return
 				}
 				// TODO(ROX-12943): Do something with the scan, e.g., attach NodeID
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -64,20 +64,19 @@ func (c *nodeScanHandlerImpl) run() {
 }
 
 func (c *nodeScanHandlerImpl) sendScan(scan *storage.NodeScanV2) {
-	if scan != nil {
-		select {
-		case <-c.stopC.Done():
-			return
-		case c.toCentral <- &central.MsgFromSensor{
-			Msg: &central.MsgFromSensor_Event{
-				Event: &central.SensorEvent{
-					Resource: &central.SensorEvent_NodeScanV2{
-						NodeScanV2: scan,
-					},
+	if scan == nil {
+		return
+	}
+	select {
+	case <-c.stopC.Done():
+	case c.toCentral <- &central.MsgFromSensor{
+		Msg: &central.MsgFromSensor_Event{
+			Event: &central.SensorEvent{
+				Resource: &central.SensorEvent_NodeScanV2{
+					NodeScanV2: scan,
 				},
 			},
-		}:
-			return
-		}
+		},
+	}:
 	}
 }
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -16,13 +16,13 @@ type nodeScanHandlerImpl struct {
 	lock *sync.Mutex
 
 	stopC concurrency.ErrorSignal
-	// runFinishedC is signaled when the goroutine inside of run() finishes
-	runFinishedC concurrency.ErrorSignal
-	startedC     concurrency.Signal
+	// stoppedC is signaled when the goroutine inside of run() finishes
+	stoppedC concurrency.ErrorSignal
+	startedC concurrency.Signal
 }
 
 func (c *nodeScanHandlerImpl) Stopped() concurrency.ReadOnlyErrorSignal {
-	return &c.runFinishedC
+	return &c.stoppedC
 }
 
 func (c *nodeScanHandlerImpl) Capabilities() []centralsensor.SensorCapability {
@@ -64,7 +64,7 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
 	toC := make(chan *central.MsgFromSensor)
 	go func() {
-		defer c.runFinishedC.Signal()
+		defer c.stoppedC.Signal()
 		defer close(toC)
 		for !c.stopC.IsDone() {
 			select {
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -12,8 +12,7 @@ type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
 	toCentral chan *central.MsgFromSensor
 
-	stopC    concurrency.ErrorSignal
-	stoppedC concurrency.ErrorSignal
+	stopC concurrency.ErrorSignal
 }
 
 func (c *nodeScanHandlerImpl) Capabilities() []centralsensor.SensorCapability {
@@ -44,10 +43,6 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 }
 
 func (c *nodeScanHandlerImpl) run() {
-	c.toCentral = make(chan *central.MsgFromSensor)
-	defer c.stoppedC.SignalWithError(c.stopC.Err())
-	defer close(c.toCentral)
-
 	for !c.stopC.IsDone() {
 		select {
 		case <-c.stopC.Done():
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -6,13 +6,23 @@ import (
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/centralsensor"
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
 )
 
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
-	toCentral chan *central.MsgFromSensor
+	toCentral <-chan *central.MsgFromSensor
+	// toCentralMux prevents the race condition between Start() [writer] and ResponsesC() [reader]
+	toCentralMux *sync.Mutex
 
 	stopC concurrency.ErrorSignal
+	// runFinishedC is signaled when the goroutine inside of run() finishes
+	runFinishedC concurrency.ErrorSignal
+	startedC     concurrency.Signal
+}
+
+func (c *nodeScanHandlerImpl) Stopped() concurrency.ReadOnlyErrorSignal {
+	return &c.runFinishedC
 }
 
 func (c *nodeScanHandlerImpl) Capabilities() []centralsensor.SensorCapability {
@@ -20,20 +30,27 @@ func (c *nodeScanHandlerImpl) Capabilities() []centralsensor.SensorCapability {
 }
 
 func (c *nodeScanHandlerImpl) ResponsesC() <-chan *central.MsgFromSensor {
+	c.toCentralMux.Lock()
+	defer c.toCentralMux.Unlock()
 	return c.toCentral
 }
 
 func (c *nodeScanHandlerImpl) Start() error {
-	// c.run closes chan toCentral on exit, so we do not want to close twice in case of a restart
-	if !c.stopC.IsDone() {
-		go c.run()
-		return nil
+	c.toCentralMux.Lock()
+	defer c.toCentralMux.Unlock()
+	if c.startedC.IsDone() {
+		return errors.New("running handler must be stopped before it can be started again")
 	}
-	return errors.New("stopped handlers cannot be restarted")
+	c.startedC.Signal()
+	if !c.stopC.IsDone() {
+		c.toCentral = c.run()
+	}
+	return nil
 }
 
 func (c *nodeScanHandlerImpl) Stop(err error) {
 	c.stopC.SignalWithError(err)
+	c.startedC.Reset()
 }
 
 func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
@@ -42,29 +59,37 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 	return nil
 }
 
-func (c *nodeScanHandlerImpl) run() {
-	for !c.stopC.IsDone() {
-		select {
-		case <-c.stopC.Done():
-			return
-		case scan, ok := <-c.nodeScans:
-			if !ok {
-				c.stopC.SignalWithError(errors.New("channel receiving node scans v2 is closed"))
+// run handles the messages from Compliance and forwards them to Central
+// This is the only goroutine that writes into the toCentral channel, thus it is responsible for creating and closing that chan
+func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
+	toC := make(chan *central.MsgFromSensor)
+	go func() {
+		defer c.runFinishedC.Signal()
+		defer close(toC)
+		for !c.stopC.IsDone() {
+			select {
+			case <-c.stopC.Done():
 				return
+			case scan, ok := <-c.nodeScans:
+				if !ok {
+					c.stopC.SignalWithError(errors.New("channel receiving node scans v2 is closed"))
+					return
+				}
+				// TODO(ROX-12943): Do something with the scan, e.g., attach NodeID
+				c.sendScan(toC, scan)
 			}
-			// TODO(ROX-12943): Do something with the scan, e.g., attach NodeID
-			c.sendScan(scan)
 		}
-	}
+	}()
+	return toC
 }
 
-func (c *nodeScanHandlerImpl) sendScan(scan *storage.NodeScanV2) {
+func (c *nodeScanHandlerImpl) sendScan(toC chan *central.MsgFromSensor, scan *storage.NodeScanV2) {
 	if scan == nil {
 		return
 	}
 	select {
 	case <-c.stopC.Done():
-	case c.toCentral <- &central.MsgFromSensor{
+	case toC <- &central.MsgFromSensor{
 		Msg: &central.MsgFromSensor_Event{
 			Event: &central.SensorEvent{
 				Resource: &central.SensorEvent_NodeScanV2{
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -1,6 +1,8 @@
 package compliance
 
 import (
+	"sync/atomic"
+
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
@@ -12,13 +14,13 @@ import (
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
 	toCentral <-chan *central.MsgFromSensor
-	// lock prevents the race condition between Start() [writer] and ResponsesC() [reader]
-	lock *sync.Mutex
 
-	stopC concurrency.ErrorSignal
+	// lock prevents the race condition between Start() [writer] and ResponsesC() [reader]
+	lock      *sync.Mutex
+	numStarts uint32
+	stopC     concurrency.ErrorSignal
 	// stoppedC is signaled when the goroutine inside of run() finishes
 	stoppedC concurrency.ErrorSignal
-	startedC concurrency.Signal
 }
 
 func (c *nodeScanHandlerImpl) Stopped() concurrency.ReadOnlyErrorSignal {
@@ -36,21 +38,17 @@ func (c *nodeScanHandlerImpl) ResponsesC() <-chan *central.MsgFromSensor {
 }
 
 func (c *nodeScanHandlerImpl) Start() error {
+	if !atomic.CompareAndSwapUint32(&c.numStarts, 0, 1) {
+		return errors.New("unable to start - component reached the maximum number of starts: 1")
+	}
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	if c.startedC.IsDone() {
-		return errors.New("running handler must be stopped before it can be started again")
-	}
-	c.startedC.Signal()
-	if !c.stopC.IsDone() {
-		c.toCentral = c.run()
-	}
+	c.toCentral = c.run()
 	return nil
 }
 
 func (c *nodeScanHandlerImpl) Stop(err error) {
 	c.stopC.SignalWithError(err)
-	c.startedC.Reset()
 }
 
 func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
@@ -64,7 +62,7 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
 	toC := make(chan *central.MsgFromSensor)
 	go func() {
-		defer c.stoppedC.Signal()
+		defer c.stoppedC.SignalWithError(c.stopC.Err())
 		defer close(toC)
 		for !c.stopC.IsDone() {
 			select {
--- a/sensor/common/compliance/nodescan_handler_impl.go
+++ b/sensor/common/compliance/nodescan_handler_impl.go
@@ -12,8 +12,8 @@ import (
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
 	toCentral <-chan *central.MsgFromSensor
-	// toCentralMux prevents the race condition between Start() [writer] and ResponsesC() [reader]
-	toCentralMux *sync.Mutex
+	// lock prevents the race condition between Start() [writer] and ResponsesC() [reader]
+	lock *sync.Mutex
 
 	stopC concurrency.ErrorSignal
 	// runFinishedC is signaled when the goroutine inside of run() finishes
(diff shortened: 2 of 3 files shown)


vikin91:
* master
  * **PR #3800** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3800" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3800?utm_source=stack-comment).

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -66,70 +65,66 @@ func (s *NodeScanHandlerTestSuite) TearDownTest() {
...
 	go func() {
 		defer close(nodeScans)
-		for i := 0; i < 4; i++ {
+		for i := 0; i < 30; i++ {
 			select {
 			case <-s.ctx.Done():
 				return
-			default:
-				nodeScans <- fakeNodeScanV2("Node")
+			case nodeScans <- fakeNodeScanV2("Node"):
```
vikin91, line 96: Fix for the race: placing the blocking part in the `case` instead of `default`

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -66,70 +65,66 @@ func (s *NodeScanHandlerTestSuite) TearDownTest() {
...
+		defer close(s.toCentral)
 		for {
 			select {
 			case <-s.ctx.Done():
 				return
-			case <-s.toCentral:
+			case _, ok := <-h.ResponsesC():
+				if !ok {
+					return
+				}
```
vikin91, line 85: Simplification: Instead of forwarding to central, we now receive and discard all msgs that are meant to go to central.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -66,70 +65,66 @@ func (s *NodeScanHandlerTestSuite) TearDownTest() {
...
 	nodeScans := make(chan *storage.NodeScanV2)
-	defer close(nodeScans)
-	h := NewNodeScanHandler(nodeScans)
-
-	assert.NoError(s.T(), h.Start())
-	h.Stop(nil)
+	s.Assert().NotPanics(func() {
+		defer close(nodeScans)
+		h := NewNodeScanHandler(nodeScans)
+		s.Assert().NoError(h.Start())
```
RTann, line 112: ah I found out a `*suite` is an `*assert.Assertions`, so no need to do `s.Assert()....`. You can just do `s.NoError(...)`

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -44,9 +43,6 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
 }
 
 func (c *nodeScanHandlerImpl) run() {
-	defer c.stoppedC.SignalWithError(c.stopC.Err())
-	defer close(c.toCentral)
```
RTann, line 48: do we not want to close this anymore?

vikin91: We do want to close this, but I think this is kind of hard here. For this to happen, we would need to make the component restartable. Let me give it one more try, maybe I will get it this time :)

vikin91: Addressed in https://github.com/stackrox/stackrox/commit/3f47d78f5160018d11d026c4629931653863bcb0 and updated the PR description. PTAL

vikin91:
Inviting additional Sensor experts to review if they find a time :) @fredrb and @lvalerom

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 	"github.com/stackrox/rox/pkg/centralsensor"
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
 )
 
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
-	toCentral chan *central.MsgFromSensor
+	toCentral <-chan *central.MsgFromSensor
+	// toCentralMux prevents the race condition between Start() [writer] and ResponsesC() [reader]
```
RTann, line 15: From what I can tell, we call `Start()` for each component before eventually calling `ResponsesC()` for each (see https://github.com/stackrox/stackrox/blob/3.72.2/sensor/common/sensor/sensor.go#L210. it looks like we call `Start()` for each component and then we eventually run `ProcessMessage()` and `ResponsesC()` concurrently). It definitely does not hurt at all to be defensive, but I don't think other sensor components have bothered.

RTann: Similarly, I wonder if we should just have some flag which indicates if `Start()` has been called and only allow `ResponsesC` continue once that happens? Or we even just comment that you must call `ResponsesC` *after* `Start()` completes and leave it up to the caller

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 
 func (c *nodeScanHandlerImpl) ResponsesC() <-chan *central.MsgFromSensor {
+	c.toCentralMux.Lock()
+	defer c.toCentralMux.Unlock()
 	return c.toCentral
 }
 
 func (c *nodeScanHandlerImpl) Start() error {
-	// c.run closes chan toCentral on exit, so we do not want to close twice in case of a restart
+	c.toCentralMux.Lock()
```
RTann, line 39: The more I read this and look at other Sensor components, the more I wonder if this may be overkill and all my questions about how to use this library were not very relevant. This seems to be correct, but I also wonder if it's worth worrying so much about this since we know how these methods are called (well at least now I know because I finally looked into it). Not saying to remove it but I wanted to see what you think

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
 )
 
 type nodeScanHandlerImpl struct {
 	nodeScans <-chan *storage.NodeScanV2
-	toCentral chan *central.MsgFromSensor
+	toCentral <-chan *central.MsgFromSensor
+	// toCentralMux prevents the race condition between Start() [writer] and ResponsesC() [reader]
+	toCentralMux *sync.Mutex
```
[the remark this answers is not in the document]
RTann: also, it looks like this mutex it used for more than just `toCentral`, but rather to ensure `Start()` and `ResponseC` do are not run in parallel/concurrently. Maybe this can just be called `lock`? (Also I think mux typically implies multiplexer instead of mutex)

RTann:
havent read it all yet, but wanted to leave some comments in case i forget to come back to this in a timely manner

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 
-	stopC    concurrency.ErrorSignal
-	stoppedC concurrency.ErrorSignal
+	stopC concurrency.ErrorSignal
+	// runFinishedC is signaled when the goroutine inside of run() finishes
+	runFinishedC concurrency.ErrorSignal
```
msugakov, line 20: `stoppedC` is more established name for this thing than `runFinishedC` in Sensor's codebase. I suggest to stick to that.

That said, I recall that you agreed to delete `Stopped()` in the last PR, but restored it here. I wonder why do you need it again?
https://github.com/stackrox/stackrox/pull/3533#discussion_r1015421998

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -20,7 +18,11 @@ func assertNoGoroutineLeaks(t *testing.T) {
 	)
 }
 func TestNodeScanHandler(t *testing.T) {
```
msugakov, line 20: Nitpicks.
1. I wonder how does it pass linters without a blank line between these two functions?
```suggestion
}

func TestNodeScanHandler(t *testing.T) {
```
2. I'd suggest to put `assertNoGoroutineLeaks` closer to where it's called.

sensor/common/compliance/nodescan_handler.go
```diff
@@ -4,21 +4,24 @@ import (
...
 // NewNodeScanHandler returns a new instance of a NodeScanHandler
 func NewNodeScanHandler(ch <-chan *storage.NodeScanV2) NodeScanHandler {
 	return &nodeScanHandlerImpl{
-		nodeScans: ch,
-		toCentral: make(chan *central.MsgFromSensor),
-
-		stopC:    concurrency.NewErrorSignal(),
-		stoppedC: concurrency.NewErrorSignal(),
+		nodeScans:    ch,
+		toCentral:    make(chan *central.MsgFromSensor),
```
msugakov, line 21: There's no need to initialize `toCentral` if `nodeScanHandlerImpl.run()` would anyway create a new one.

sensor/common/compliance/nodescan_handler_impl.go
msugakov: +1 to not overkill. In my opinion, it will be sufficient to check whether a Start() was once attempted, and, if so, return an error. I don't think there's a need to make this component restartable provided that others don't take advantage of it.

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 }
 
 func (c *nodeScanHandlerImpl) Start() error {
-	// c.run closes chan toCentral on exit, so we do not want to close twice in case of a restart
+	c.toCentralMux.Lock()
+	defer c.toCentralMux.Unlock()
+	if c.startedC.IsDone() {
+		return errors.New("running handler must be stopped before it can be started again")
+	}
+	c.startedC.Signal()
```
msugakov, line 44: You can use atomic integer increment for checking whether Start() was attempted, and any values other than 1 attempt should result in the error. This way there's no need to have `startedC` signal and check-then-act race condition (`startedC` is first checked and then signalled, multiple concurrent threads can all pass the check under unlucky timing).

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -6,35 +6,51 @@ import (
...
 	if !c.stopC.IsDone() {
-		go c.run()
-		return nil
+		c.toCentral = c.run()
 	}
```
msugakov, line 47: I don't understand what does `!c.stopC.IsDone()` condition signify in the new logic.

If the component was never started before, the check passes and `run()` is called. No surprise but the check is not needed either.

If the component was started and currently running, the code wouldn't even get to this check.

If the component was started and `run()`-s inner goroutine completed, the code wouldn't even get to this check.

If the component was started and then shut down with the `Stop()` call, the condition will evaluate to `false` thus making `Start()` return without error but also without calling `run()`. I don't think that's a desired behavior.

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -43,40 +59,44 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
...
-			return
-		case scan, ok := <-c.nodeScans:
-			if !ok {
-				c.stopC.SignalWithError(errors.New("channel receiving node scans v2 is closed"))
+// run handles the messages from Compliance and forwards them to Central
+// This is the only goroutine that writes into the toCentral channel, thus it is responsible for creating and closing that chan
+func (c *nodeScanHandlerImpl) run() <-chan *central.MsgFromSensor {
+	toC := make(chan *central.MsgFromSensor)
+	go func() {
+		defer c.runFinishedC.Signal()
```
msugakov, line 67: Did you mean
```suggestion
		defer c.runFinishedC.SignalWithError(c.stopC.Err())
```
?

msugakov (commented):
Checked the main code mostly, skipped tests.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I know that some usage patterns are not used in our setup, but my tests reported race or a leak and I wanted them to be clean. Moreover, we discussed the restartablilty of sensor components in the F2F meeting and we may be soon rely on it.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I know this is an overkill a bit, but I think this investment is worth it. We may need to make all Sensor components restartable, so I thought that this may be also a good way to have the race & goroutine leak tests green.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I decided to add it again, so that the tests can wait for the component to fully stop. Waiting is important because we run a goroutine-leak tests in a deferred statement and it happened sometimes that the goroutines were analyzed before the component stopped.

I will restore the name `stoppedC`.

vikin91:
Thanks @msugakov , I will address the review on Monday!

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -43,40 +59,44 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
...
-				Event: &central.SensorEvent{
-					Resource: &central.SensorEvent_NodeScanV2{
-						NodeScanV2: scan,
-					},
+func (c *nodeScanHandlerImpl) sendScan(toC chan *central.MsgFromSensor, scan *storage.NodeScanV2) {
+	if scan == nil {
+		return
+	}
+	select {
+	case <-c.stopC.Done():
```
fredrb, line 91: If I understand this correctly, there are two reasons why `c.stopC.Done()` would have stopped:

1) The parent channel (`c.nodeScans`) is closed
2) `Stop()` was called on the component. 

I'm not sure what is the meaning of calling `Stop()` on the component -- but I assume this would be a "graceful shutdown" of sensor. If that's the case, it's likely that other components (such as the one responsible for closing `c.nodeScans`) will also shutdown and close their channels. 

With that in mind, the current code will drop the events that are already enqueued and they will be lost. I'm not sure if this is a problem or not, but if these events can't be lost, it would be better to read the channel until it returns `nil, false`, so we know that all events are sent.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +53,142 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
-			case msg, ok := <-c:
+			case _, ok := <-h.ResponsesC():
 				if !ok {
 					return
 				}
-				s.toCentral <- msg
 			}
 		}
 	}()
+	// this is a producer that stops the handler after producing the first message and then sends many (29) more messages
```
fredrb, line 89: I see that you've added a test for the exact thing I mentioned in the comment above.

I'm not sure what is the intent here. Are we expecting that if a component is stopped and then restarted, all other (29) messages in the pipeline will be read by the new instance?

If that's the case, I don't see how sensor lifecycle benefits from this. Because what we're missing is a graceful shutdown routine that can stop _all_ components and process leftover messages.

sensor/common/compliance/nodescan_handler.go
```diff
@@ -4,21 +4,24 @@ import (
...
+		stopC:        concurrency.NewErrorSignal(),
```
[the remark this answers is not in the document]
msugakov: This will not be consistent with how `stopC`/`stoppedC` are named in other files. Either we should keep the old wrong names here or we should rename all occurrences proper way. The worse thing is to have multiple patterns that mean the same thing.

sensor/common/compliance/nodescan_handler_impl.go
msugakov: I also spotted loss of events in some previous iteration on this code. I think what you suggest is correct, although achieving that will require changing the service part that receives messages from clients.

When the graceful shutdown is requested, the service should stop receiving (or ACK-ing) new messages from clients and close its internal channel (that's used for passing these messages further in Sensor's code). That should cascade throughout the system until all internal channels are closed so that the Central Sender also gets to understand that there will be nothing more to send and shuts down itself.

I'm afraid that such a change may turn quite a lot bigger than this PR. I think we should do it, but my vote for a separate PR. In this PR I suggest to not attempt fixing everything (especially deeper patterns such as `Start()` after `Stop()`), but rather just enough to make it not worse than before.

The specific message with node scan V2 (node inventory) can be dropped on the floor. By design, Compliance will resend it after 4 hours. This isn't great but it will be improved as part of the same epic by making Compliance retry sending until it receives an ACK - https://issues.redhat.com/browse/ROX-13164

sensor/common/compliance/nodescan_handler_test.go
vikin91: I added a comment to the code about what is meant by the test. My intention was to test that the context cancelation works and shuts down the component clearly even if there are still many messages waiting to be processed.

Whether we should stop now or first process everything and then stop is another kind of discussion. For that, we would need to look at how the components are stopped and what is the desired behavior in general (e.g., what if we have really many messages waiting - shall we postpone stop forever?)

sensor/common/compliance/nodescan_handler_impl.go
vikin91: I added this, but we have no guarantee that the  `c.stopC.Err()` will be something other than nil at the moment of evaluation of this statement.

vikin91:
I removed the restartabilty and simplified some parts in d7353aed4c076ee0b9e3a9ad492ab84d28a93433

sensor/common/compliance/nodescan_handler_impl.go
msugakov: :-)
So this should be something like the following?

```go
defer func() {
    c.runFinishedC.SignalWithError(c.stopC.Err())
}()
```

vikin91:
Rebased to get rid of the expired cert failure

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -1,36 +1,50 @@
...
 	return c.toCentral
 }
 
 func (c *nodeScanHandlerImpl) Start() error {
-	// c.run closes chan toCentral on exit, so we do not want to close twice in case of a restart
-	if !c.stopC.IsDone() {
-		go c.run()
-		return nil
+	if !atomic.CompareAndSwapUint32(&c.numStarts, 0, 1) {
+		return errors.New("unable to start - component reached the maximum number of starts: 1")
```
msugakov, line 42: I suggest to make the error message a bit more human-friendly

```suggestion
		return errors.New("unable to start the component more than once")
```

vikin91: Got it covered in 0cdd69cd0e1ad6887951afad8bea2d756a4c8565

sensor/common/compliance/nodescan_handler.go
vikin91: I didn't follow up on this. Last time I moved the initialization outside of the constructor I got race detection. Will check it again.

sensor/common/compliance/nodescan_handler_impl.go
vikin91: Let me add a test-case for this

sensor/common/compliance/nodescan_handler.go
vikin91: I set that to nil, however it must remain protected with the mutex to avoid race when running tests.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +48,138 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
+					return
 				}
 			}
 		}
 	}()
-	err := h.Start()
-	assert.NoError(s.T(), err)
 }
 
 func (s *NodeScanHandlerTestSuite) TestRestartHandler() {
```
msugakov, line 151: This and further tests should also make sure that the handler is stopped at the end.

jschnath (approved):
LGTM, nothing to add beyond Mishas comments

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +48,138 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
+			case nodeScans <- fakeNodeScanV2("Node"):
+				if i == 0 {
+					h.Stop(nil)
 				}
-				s.toCentral <- msg
 			}
 		}
 	}()
+	s.NoError(h.Start())
 }
```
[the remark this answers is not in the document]
vikin91: Addressed that, however the assertion `s.NoError` cannot be used here. The `err := h.Stopped().Wait()` can be either the argument passed to `h.Stop(err)` or `errInputChanClosed` if `nodeScans` closes faster than the call to `h.Stop()` is being handled.

See commit `Add waiting for Stopped() in TestStopHandler` (will add a link to the commit after I push it).

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +48,138 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
+// in the channel passed into NewNodeScanHandler.
+// We expect that premature stop of the handler produces no race condition or goroutine leak.
+// Exec with: go test -race -count=1 -v -run ^TestNodeScanHandler$ ./sensor/common/compliance
+func (s *NodeScanHandlerTestSuite) TestStopHandler() {
+	nodeScans := make(chan *storage.NodeScanV2)
+	h := NewNodeScanHandler(nodeScans)
+	s.consumeToCentral(h)
+	// This is a producer that stops the handler after producing the first message and then sends many (29) more messages
+	// The intent here it to test whether the handler can shutdown with no leaks when the context is canceled despite of
+	// multiple messages waiting in the channel to be processed
```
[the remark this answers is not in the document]
vikin91: Addressed in commit `Rephrase comment in TestStopHandler`

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +48,138 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
 	go func() {
-		defer close(s.toCentral)
-		for {
+		defer close(nodeScans)
```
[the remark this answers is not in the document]
vikin91: > Closing the channel in this goroutine is interesting and maybe asks for a dedicated test.

I am following a pattern here where the writer is responsible for closing the channel that it writes to when its writing-work is done.

> Depending on what select statement in run() sees first - closed channel, or signalled stopC, the result returned from h.Stopped() will have or not have an error.

It will have error in both cases, just the error will be different - either the error passed as an argument to `Stop()` (which can be nil), or the predefined `errInputChanClosed`. 

> For this test I suggest to move defer close(nodeScans) outside of this goroutine. That way, it will be called after the node scan handler reports that it's stopped.

Yes, that should allow us to make an `ErrorIs` assertion without the or-condition, but there is a risk of writing to the channel after closing it. I will check whether the risk can be eliminated in that particular test case.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -51,85 +48,138 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
-func (s *NodeScanHandlerTestSuite) TestStopHandler() {
-	nodeScans := make(chan *storage.NodeScanV2)
-	h := NewNodeScanHandler(nodeScans)
-	// consume all messages send toCentral
+func (s *NodeScanHandlerTestSuite) TestHandlerRegularRoutine() {
+	h := NewNodeScanHandler(s.generateTestInput())
+	s.NoError(h.Start())
+	s.consumeToCentral(h)
+	h.Stop(nil)
+}
```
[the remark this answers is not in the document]
vikin91: I changed it to

```
	errTest := errors.New("example-stop-error")
	h.Stop(errTest)
	err := h.Stopped().Wait()
	s.True(errors.Is(err, errTest) || errors.Is(err, errInputChanClosed))
```

sensor/common/compliance/nodescan_handler_test.go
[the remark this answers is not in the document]
vikin91: I checked that and not closing the channel in the goroutine triggers a data race between the test routine and the anonymous routine. I would need to add a WaitGroup to prevent this.

This is how this case would look like. It allows us to assert `s.NoError(h.Stopped().Wait())` but it clutters the code a bit more.

```go
func (s *NodeScanHandlerTestSuite) TestHandlerRegularRoutine2() {
	wg := &sync.WaitGroup{}
	wg.Add(1)
	ch := s.generateTestInputNoClose(wg)
	defer func(wg *sync.WaitGroup) {
		wg.Wait()
		close(ch)
	}(wg)
	h := NewNodeScanHandler(ch)
	s.NoError(h.Start())
	s.consumeToCentral(h)
	h.Stop(nil)
	s.NoError(h.Stopped().Wait())
}
```

Will not commit this for now and wait for your reply.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -74,11 +72,12 @@ func (s *NodeScanHandlerTestSuite) TearDownTest() {
 // Exec with: go test -race -count=1 -v -run ^TestNodeScanHandler$ ./sensor/common/compliance
 func (s *NodeScanHandlerTestSuite) TestStopHandler() {
 	nodeScans := make(chan *storage.NodeScanV2)
+	errTest := errors.New("example-stop-error")
 	h := NewNodeScanHandler(nodeScans)
 	s.consumeToCentral(h)
 	// This is a producer that stops the handler after producing the first message and then sends many (29) more messages
 	// The intent here it to test whether the handler can shutdown with no leaks when the context is canceled despite of
```
msugakov, line 79: I don't see how "context is canceled" has any meaning for the node scan handler that doesn't deal with contexts. It looks like it's best to rework this comment further.

vikin91:
One test still has issues: `go test -race -count=30 -timeout=10s -v -run ^TestNodeScanHandler/TestStopHandler$ ./sensor/common/compliance`. Will continue the next working day.

sensor/common/compliance/nodescan_handler_test.go
vikin91: Done in [c7ebe5c](https://github.com/stackrox/stackrox/pull/3800/commits/c7ebe5c3b09352a257070aa2e7eee2498a474c2b)

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+
+	errTest := errors.New("example-stop-error")
+	h.Stop(errTest)
+	s.ErrorIs(h.Stopped().Wait(), errTest)
+
+	stopAll(s.T(), producer, consumer)
+}
+
+// generateTestInputNoClose generates numToProduce messages of type NodeScanV2
+// It returns channel that must be closed
```
msugakov, line 178:
```suggestion
// It returns a channel that must be closed by the caller.
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+func (s *NodeScanHandlerTestSuite) generateTestInputNoClose(numToProduce int) (chan *storage.NodeScanV2, stoppable) {
+	input := make(chan *storage.NodeScanV2)
+	st := newStoppable()
+	// this is a producer that sends 10 nodescan messages
```
msugakov, line 182: This comment is no longer valid. I think it's just redundant.
```suggestion
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		for i := 0; i < numToProduce; i++ {
 			select {
-			case <-s.ctx.Done():
+			case <-st.stopC.Done():
 				return
-			default:
-				nodeScans <- fakeNodeScanV2("Node")
-				if i == 0 {
-					h.Stop(nil)
+			case input <- fakeNodeScanV2("Node"):
```
msugakov, line 189: Very optional
```suggestion
			case input <- fakeNodeScanV2(fmt.Sprintf("Node-%d", i)):
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+	s.NoError(h.Start())
+	consumer := consumeAndCount(h.ResponsesC(), 10)
+	s.NoError(producer.stoppedC.Wait())
+	s.NoError(consumer.stoppedC.Wait())
+
+	errTest := errors.New("example-stop-error")
+	h.Stop(errTest)
+	s.ErrorIs(h.Stopped().Wait(), errTest)
+
+	stopAll(s.T(), producer, consumer)
```
msugakov, line 158: This is redundant as the fact that `producer` and `consumer` stop is checked above.
```suggestion
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+	defer close(ch)
+	h := NewNodeScanHandler(ch)
+	s.NoError(h.Start())
+	consumer := consumeAndCount(h.ResponsesC(), 10)
+	s.NoError(producer.stoppedC.Wait())
+	s.NoError(consumer.stoppedC.Wait())
+
+	errTest := errors.New("example-stop-error")
+	h.Stop(errTest)
+	s.ErrorIs(h.Stopped().Wait(), errTest)
```
msugakov, line 156:
```suggestion
	h.Stop(nil)
	s.NoError(h.Stopped().Wait())
```
because otherwise this test is identical to `TestHandlerStoppedError`.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+
+	stopAll(s.T(), producer, consumer)
```
msugakov, line 174:
```suggestion
```
similar to the above.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
+		h.Stop(nil)
+
+		err := h.Start()
+		s.Error(err)
+		s.ErrorIs(err, errStartMoreThanOnce)
+		s.NoError(h.Stopped().Wait())
+
+		stopAll(s.T(), producer, consumer)
```
msugakov, line 236:
```suggestion
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
+
+		err := h.Start()
+		s.Error(err)
+		s.ErrorIs(err, errStartMoreThanOnce)
+		h.Stop(nil)
+		s.NoError(h.Stopped().Wait())
+
+		stopAll(s.T(), producer, consumer)
```
msugakov, line 257:
```suggestion
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		h := NewNodeScanHandler(ch)
+		s.NoError(h.Start())
+		consumer := consumeAndCount(h.ResponsesC(), 10)
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
+		h.Stop(nil)
+		h.Stop(nil)
+		s.NoError(h.Stopped().Wait())
 
+		stopAll(s.T(), producer, consumer)
```
msugakov, line 274:
```suggestion
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		s.NoError(h.Start())
+		consumer := consumeAndCount(h.ResponsesC(), 10)
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
+		h.Stop(nil)
+
+		err := h.Start()
+		s.Error(err)
+		s.ErrorIs(err, errStartMoreThanOnce)
+		s.NoError(h.Stopped().Wait())
```
msugakov, line 234: I think this line `s.NoError(h.Stopped().Wait())` needs to be moved after to `h.Stop(nil)`.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+func (s *NodeScanHandlerTestSuite) TestDoubleStartHandler() {
+	s.NotPanics(func() {
+		ch, producer := s.generateTestInputNoClose(10)
+		defer close(ch)
+		h := NewNodeScanHandler(ch)
+		s.NoError(h.Start())
+
+		consumer := consumeAndCount(h.ResponsesC(), 10)
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
```
msugakov, line 249: I think this test will be the most indicative if you move these two lines to the end of the test. This would mean that despite the second `Start()` call returns an error, the handler is able to complete its job cleanly.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+
+		err := h.Start()
+		s.Error(err)
+		s.ErrorIs(err, errStartMoreThanOnce)
```
msugakov, line 253:
```suggestion
		s.ErrorIs(h.Start(), errStartMoreThanOnce)
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +115,162 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		h := NewNodeScanHandler(ch)
+		s.NoError(h.Start())
+		consumer := consumeAndCount(h.ResponsesC(), 10)
+		s.NoError(producer.stoppedC.Wait())
+		s.NoError(consumer.stoppedC.Wait())
+		h.Stop(nil)
+
+		err := h.Start()
+		s.Error(err)
+		s.ErrorIs(err, errStartMoreThanOnce)
```
msugakov, line 233:
```suggestion
		s.ErrorIs(h.Start(), errStartMoreThanOnce)
```

sensor/common/compliance/nodescan_handler_test.go
vikin91: That change would lead to deadlock

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -3,15 +3,42 @@ package compliance
...
+}
+
+// signalAndWait sends a command to stop the stoppaple thing with an error and waits until it stops
+func (st *stoppable) signalAndWait(err error) error {
+	st.stopC.SignalWithError(err)
+	return st.stoppedC.Wait()
+}
+
 func TestNodeScanHandler(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -49,38 +73,40 @@ func fakeNodeScanV2(nodeName string) *storage.NodeScanV2 {
...
-	s.toCentral = make(chan *central.MsgFromSensor)
+func assertNoGoroutineLeaks(t *testing.T) {
+	goleak.VerifyNone(t,
+		// Ignore a known leak: https://github.com/DataDog/dd-trace-go/issues/1469
+		goleak.IgnoreTopFunction("github.com/golang/glog.(*loggingT).flushDaemon"),
+	)
 }
 
 func (s *NodeScanHandlerTestSuite) TearDownTest() {
 	defer assertNoGoroutineLeaks(s.T())
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -91,110 +137,141 @@ func (s *NodeScanHandlerTestSuite) TestStopHandler() {
...
 	}()
-	s.NoError(h.Start())
-	err := h.Stopped().Wait()
-	// if the goroutine finishes before h.Stopped() is cone, then err is errInputChanClosed
-	// otherwise it is errTest. Both are fine as a reason for stopping the handler
-	s.True(errors.Is(err, errTest) || errors.Is(err, errInputChanClosed))
+	s.NoError(consumer.stoppedC.Wait())
+	s.ErrorIs(h.Stopped().Wait(), errTest)
+
+	stopAll(s.T(), producer, consumer)
```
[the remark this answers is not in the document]
vikin91: Done

sensor/common/compliance/nodescan_handler_test.go
vikin91: I duplicated this line now, as we stop twice in this test case.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -17,14 +17,16 @@ import (
 
 // stoppable represents a gracefully stoppable thing, e.g., async process
 type stoppable struct {
-	stopC    concurrency.ErrorSignal
-	stoppedC concurrency.ErrorSignal
+	stopC     concurrency.ErrorSignal
+	stoppedC  concurrency.ErrorSignal
+	receivedC concurrency.Signal
```
msugakov: `receivedC` does not belong here as it is something very specific to what only `consumeAndCount` does.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -131,14 +133,15 @@ func (s *NodeScanHandlerTestSuite) TestStopHandler() {
 				return
 			case nodeScans <- fakeNodeScanV2("Node"):
 				if i == 0 {
+					consumer.receivedC.Wait() // avoid race between consumer receiving the first message and hadler stopping
```
msugakov: You can achieve the same without introducing `receivedC` by just calling
```go
s.NoError(consumer.stoppedC.Wait())
```
instead of `consumer.receivedC.Wait()`.

sensor/common/compliance/nodescan_handler_test.go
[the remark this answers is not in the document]
msugakov: I still see `stopAll(s.T(), producer, consumer)` in `TestStopHandler`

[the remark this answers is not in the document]
msugakov: I still see `stopAll(s.T(), producer, consumer)` in `TestStopHandler`.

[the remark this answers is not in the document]
vikin91: Sorry, I must have reverted that by accident. Fixed in [56dab50](https://github.com/stackrox/stackrox/pull/3800/commits/56dab50b63d64f8c4737a92e7c509b7e80825f96)

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +107,148 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		for i := 0; i < 30; i++ {
 			select {
-			case <-s.ctx.Done():
+			case <-producer.stopC.Done():
 				return
-			case <-s.toCentral:
+			case nodeScans <- fakeNodeScanV2("Node"):
+				if i == 0 {
+					h.Stop(errTest)
+				}
```
msugakov, line 127: Just move waiting for counting consumer inside here and it will make sure the consumer sees its 1 message.
```suggestion
				if i == 0 {
					s.NoError(consumer.stoppedC.Wait()) // This blocks until consumer receives its 1 message
					h.Stop(errTest)
				}
```

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +107,148 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
 			}
 		}
 	}()
-	s.startForwardingToCentral(h.ResponsesC())
-	// this is a producer that stops the handler after producing the first message and then sends 3 more messages
+	s.NoError(consumer.stoppedC.Wait())
+	s.ErrorIs(h.Stopped().Wait(), errTest)
+
+	stopAll(s.T(), producer)
```
msugakov, line 134: Next,
```suggestion
	s.ErrorIs(h.Stopped().Wait(), errTest)

	s.NoError(producer.signalAndWait(nil))
```
and you can delete `stopAll` as unused.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -89,47 +107,148 @@ func (s *NodeScanHandlerTestSuite) startForwardingToCentral(c <-chan *central.Ms
...
+		consumer := consumeAndCount(h.ResponsesC(), 10)
+		s.NoError(producer.stoppedC.Wait())
+
+		s.ErrorIs(h.Start(), errStartMoreThanOnce)
+		h.Stop(nil)
+		s.NoError(h.Stopped().Wait())
+
+		s.NoError(consumer.stoppedC.Wait())
+	})
+}
```
msugakov, line 239: This implementation fails test once in a while. Here's what I can suggest
```suggestion
func (s *NodeScanHandlerTestSuite) TestDoubleStartHandler() {
	s.NotPanics(func() {
		ch, producer := s.generateTestInputNoClose(10)
		defer close(ch)
		h := NewNodeScanHandler(ch)

		s.NoError(h.Start())
		s.ErrorIs(h.Start(), errStartMoreThanOnce)

		consumer := consumeAndCount(h.ResponsesC(), 10)

		s.ErrorIs(h.Start(), errStartMoreThanOnce)

		s.NoError(producer.stoppedC.Wait())
		s.NoError(consumer.stoppedC.Wait())

		h.Stop(nil)
		s.NoError(h.Stopped().Wait())
	})
}
```

sensor/common/compliance/nodescan_handler_impl.go
```diff
@@ -43,40 +64,46 @@ func (c *nodeScanHandlerImpl) ProcessMessage(_ *central.MsgToSensor) error {
...
+		defer close(toC)
+		for !c.stopC.IsDone() {
+			select {
+			case <-c.stopC.Done():
 				return
+			case scan, ok := <-c.nodeScans:
+				if !ok {
+					c.stopC.SignalWithError(errInputChanClosed)
+					return
+				}
```
msugakov, line 84: When I run tests in GoLand with coverage, they show that this code branch isn't covered. I think it's significant enough to be tested.
image

vikin91:
@RTann @fredrb @lvalerom we are done with changes (provided that CI is happy), if you want to review again, then it may be a good time for it.

PR #3801, opened 2022-11-14
Fix jira release script

janisz:
## Description

Fix jira script and use renamed ACS component.

## Testing Performed

Run script

### DIFF
--- a/scripts/release-tools/jira.sh
+++ b/scripts/release-tools/jira.sh
@@ -8,7 +8,7 @@ MENU_OPTIONS=(
   "Quit"
 )
 
-FIX_VERSION="3.72.0" # copy this from JIRA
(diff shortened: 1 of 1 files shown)


vikin91:
https://github.com/stackrox/stackrox/tree/069715f81290c3c434850af608a23ed007ac8e33/scripts/release-tools#deprecation-notice

We decided to deprecate these scripts and use GH Actions instead. For more details, see: https://srox.slack.com/archives/CELUQKESC/p1668435152850239?thread_ts=1668430547.993149&cid=CELUQKESC

0x656b694d (approved):

janisz:
@vikin91 Could you remove this scripts then?
>  The scripts are deprecated and will be removed on October 10, 2022

PR #3804, opened 2022-11-14
remove make proto-fmt

janisz:
## Description

It looks like proto-fmt is not used anywhere (especially is not called from `make style`).
When I run it locally I get list of errors. It will be great to have it enabled in CI but since it was not there for so long lets remove it.

## Testing Performed

CI


msugakov:
Looks like `proto-fmt` came with this commit in 2017 and wasn't used. https://github.com/stackrox/stackrox/commit/d34776a88584b7200dacb3f105bb52e0eb88209a
I know that Connor wanted a better/working formatter and I was looking at `clang-format` as a promising alternative, but did not get to try it.

msugakov (approved):

PR #3815, opened 2022-11-15
ROX-13523: add enabled enum to central db spec

johannes94:
## Description

This PR adds an additional property `spec.central.db.isEnabled` to the central CRD. It fixes the issue that central DB gets provisioned in openshift due to default values for the db spec. To enable the central DB one has to set `spec.central.db.isEnabled` to `Enabled`

## Checklist

## Testing Performed

```
# Tested manual with crc and running the operator locally
# Use MAIN_IMAGE_TAG do don't go through the build process of stackrox main image, since this didn't change
MAIN_IMAGE_TAG=3.72.x-nightly-20221115 make install run    

# Create namespace test
kubectl create ns test

# Create a central CR with openshift default values
cat <<EOF | kubectl apply -f -
apiVersion: platform.stackrox.io/v1alpha1
kind: Central
metadata:
  name: stackrox-central-services
  namespace: test
spec:
  central:
    db:
      persistence:
        persistentVolumeClaim:
          claimName: central-db
    persistence:
      persistentVolumeClaim:
        claimName: stackrox-db
  scanner:
    analyzer:
      scaling:
        autoScaling: Enabled
        maxReplicas: 5
        minReplicas: 1
        replicas: 1
EOF

# Make sure no central-db pods or pvc is created and stackrox-db pvc is created
kubectl get pvc -n test
kubectl get po -n test

# Verify spec.central.db.enabled=Default
kubectl get central -n test stackrox-central-services -o jsonpath={.spec.central.db.enabled}

# Delete the central cr
kubectl delete central -n test stackrox-central-services

# Create a central cr with openshift default values + spec.central.db.enabled = "Enabled"
cat <<EOF | kubectl apply -f -
apiVersion: platform.stackrox.io/v1alpha1
kind: Central
metadata:
  name: stackrox-central-services
  namespace: test
spec:
  central:
    db:
      enabled: Enabled
      persistence:
        persistentVolumeClaim:
          claimName: central-db
    persistence:
      persistentVolumeClaim:
        claimName: stackrox-db
  scanner:
    analyzer:
      scaling:
        autoScaling: Enabled
        maxReplicas: 5
        minReplicas: 1
        replicas: 1
EOF

# Verify central-db pvc and pod are created
kubectl get po -n test
kubectl get pvc -n test

# Verify spec.central.db.enabled=Enabled
kubectl get central -n test stackrox-central-services -o jsonpath={.spec.central.db.enabled}

# Cluster Cleanup
kubectl delete central -n test stackrox-central-services
kubectl delete ns test
make uninstall
```
Also test:

```
make test
```


operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -150,35 +150,50 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
 
 // CentralDBEnabled returns a bool if CentralDBSpec is not nil
```
porridge, line 151: Please update docstrings as you go. :wink:

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -150,35 +150,50 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
-	return c.DB != nil
+
+	return c.DB.Enabled == CentralDBEnabledTrue
 }
 
 // CentralDBSpec defines settings for the "central db" component.
 type CentralDBSpec struct {
+	//+kubebuilder:default=Default
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
+	Enabled CentralDBEnabled `json:"enabled"`
```
porridge, line 164: It seems there are a few related settings:
1. whether the operator should be provisioning a postgres instance
2. whether central should be using rocksdb or postgres
3. if the latter:
   - connect to the provisioned instance, or
   - to some other postgres connection string

With what you propose we would control this with 2 knobs:
- `spec.central.db.enabled`
- `spec.central.db.connectionString`

It makes sense that the operator is not provisioning a postgres if a connection string is specified. However it feels awkward that one needs to look at `connectionString` to determine (1). The result is there are a couple of places in the codebase which need to look at both `central.Spec.Central.CentralDBEnabled()` and `central.Spec.Central.DB.IsExternal()`.

How about this as a single tri-state `spec.central.storage` field instead, with the following values (exact wording TBD):
- `Embedded`: use rocksdb, don't provision a postgres
- `DB`: provision a postgres and use that; note: currently in tech preview; requires empty/null `connectionString`
- `External`: don't provision a postgres, use some other postgres connection string; note: currently in tech preview; requires non-empty `connectionString`
- `Default`: currently same as `Builtin`, in 3.74, same as `DB`

porridge:
Can you also add a step to the central kuttl test which flips this on and makes sure the central DB is ok?

operator/apis/platform/v1alpha1/central_types.go
vladbologa: It's already controlled with 2 knobs, this PR doesn't change that:
* whether `spec.central.db` is present or not
* `spec.central.db.connectionString`

I think we should not add the value `Embedded` (or `Builtin`) because we will phase this out in v3.74, as discussed with @msugakov in Slack.

And separating `DB` and `External` into 2 states might save us from looking at both `central.Spec.Central.CentralDBEnabled()` and `central.Spec.Central.DB.IsExternal()` in some places, but it also creates two possible invalid configurations.

So I'm still leaning towards how Johannes implemented it here.

johannes94: I agree, with Vlad on this. With the suggested change we would enable a user setting the type explicitly to embedded, which we intend to phase out by the next release.

operator/apis/platform/v1alpha1/central_types.go
johannes94: Did not really make sense to me to replicate the function logic in the comment, I changed it to something more generic.

johannes94:
> Can you also add a step to the central kuttl test which flips this on and makes sure the central DB is ok?

I added the enum to the existing kuttl test that test central-db creation. I also added a step to make sure central-db is not created for the default values openshift console is setting.

operator/tests/central/basic/74-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: stackrox-db
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Gi
```
janisz, line 10: nit: some files has EOL and some not

janisz (approved):

operator/config/crd/bases/platform.stackrox.io_centrals.yaml
```diff
@@ -70,6 +70,13 @@ spec:
                           set a password secret; automatically generating a password
                           will not be supported.
                         type: string
+                      enabled:
```
vladbologa, line 73: `operator/bundle/manifests/platform.stackrox.io_centrals.yaml` should be updated too, I don't know if it's done automatically or you have to do it manually.

vladbologa (changes requested):
Looks good to me, just a couple of small comments.

operator/config/crd/bases/platform.stackrox.io_centrals.yaml
johannes94: Thanks, I executed `make bundle` now, that updated the files. Hope this is sufficient.

operator/tests/central/basic/74-assert.yaml
johannes94: Thanks, CI's style check fortunately also complains about such things. Fixed it.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,60 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 		return false
 	}
-	return c.DB != nil
+
+	return *c.DB.Enabled == CentralDBEnabledTrue
 }
 
 // CentralDBSpec defines settings for the "central db" component.
 type CentralDBSpec struct {
+	//+kubebuilder:validation:Default=Default
```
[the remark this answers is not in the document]
johannes94: I added a description.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,63 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 		return false
 	}
-	return c.DB != nil
+
+	return *c.DB.Enabled == CentralDBEnabledTrue
 }
 
 // CentralDBSpec defines settings for the "central db" component.
 type CentralDBSpec struct {
+	// Specify wether central-db is enabled, Default configures central to use rocksdb.
```
vladbologa, line 162:
```suggestion
	// Specify whether central-db is enabled, Default configures central to use rocksdb.
```

vladbologa (approved):

vladbologa (approved):
lgtm

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,63 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 }
 
 // CentralDBSpec defines settings for the "central db" component.
 type CentralDBSpec struct {
+	// Specify whether central-db is enabled, Default configures central to use rocksdb.
+	// If Enabled the operator will provision the Central DB or use an external DB if a connection string is provided.
+	//+kubebuilder:validation:Default=Default
+	//+kubebuilder:default=Default
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
+	Enabled *CentralDBEnabled `json:"enabled,omitempty"`
```
msugakov, line 167: Renaming suggestion, in order to avoid confusion between attribute name `enabled` and enum value `Enabled`:
```suggestion
	IsEnabled *CentralDBEnabled `json:"isEnabled,omitempty"`
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,63 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
 	return pointer.BoolPtrDerefOr(c.AdminPasswordGenerationDisabled, false)
 }
 
-// CentralDBEnabled returns a bool if CentralDBSpec is not nil
+// CentralDBEnabled returns true if central db should be created
```
msugakov, line 151: Let's add a todo for ourselves to not forget to modify the logic in the future release.
```suggestion
// CentralDBEnabled returns true if central db should be created
// TODO(ROX-13557): modify the logic to enable Postgres DB by default.
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,63 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 	DeploymentSpec `json:",inline"`
 }
 
+// CentralDBEnabled is a type for values of spec.central.db.enabled.
+//+kubebuilder:validation:Enum=Default;Enabled
+type CentralDBEnabled string
+
+const (
+	// CentralDBEnabledDefault configures the central to use rocksdb
+	CentralDBEnabledDefault CentralDBEnabled = "Default"
```
msugakov, line 196: Let's add a todo for ourselves to not forget to modify this enum value's description in 3.74.0.

```suggestion
	// CentralDBEnabledDefault configures the central to use rocksdb
	// TODO(ROX-13557): in 3.74.0 this should say that Default enables PostgreSQL.
	CentralDBEnabledDefault CentralDBEnabled = "Default"
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -148,37 +148,63 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
+	// CentralDBEnabledTrue configures the central to use a PostgreSQL database (Technical Preview)
```
msugakov, line 197: There's no such thing as "Technical Preview", there's "Technology Preview". https://access.redhat.com/support/offerings/techpreview/
```suggestion
	// CentralDBEnabledTrue configures the central to use a PostgreSQL database (Technology Preview)
	// TODO(ROX-13557): omit mention of Technology Preview.
```

It would be really nice if you could also find+replace all other Technical Previews with Technology Preview in this file.

msugakov (commented):
Few cosmetic notes

operator/apis/platform/v1alpha1/central_types.go
johannes94: Done.

operator/apis/platform/v1alpha1/central_types.go
johannes94: Done

operator/apis/platform/v1alpha1/central_types.go
johannes94: Done.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -366,7 +368,7 @@ type ExposureLoadBalancer struct {
 	//+kubebuilder:validation:Default=false
 	//+kubebuilder:default=false
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
-	Enabled *bool `json:"enabled,omitempty"`
+	Enabled *bool `json:"isEnabled,omitempty"`
```
msugakov, line 371: I think you may not want to change `ExposureLoadBalancer`'s field name because it's something that can break backward compatibility.

msugakov (approved):
LGTM, except you should probably undo `ExposureLoadBalancer` change.

operator/apis/platform/v1alpha1/central_types.go
johannes94: Oh yes, big thank you for that I had errors in kuttl test validation and was searching what I did wrong with renaming. Here it is.

PR #3816, opened 2022-11-15
Bugfixes for collection autocomplete

dvail:
## Description

This fixes two issues with the autocomplete functionality:

1. The 'onClick' behavior was broken during the last round of refactoring, making it so selecting a value does not correctly apply it.

2. The autocomplete list will now be cleared while the user it typing.  Previously this led to clunky behavior where the list would remain populated with data that did not match the user's text until 800ms after typing was finished. (Due to the debounce). Now, when the user begins typing the dropdown options will be cleared until typing has finished.

## Checklist

## Testing Performed

Select an autocomplete input element and allow it to populate the first `#` default entries. Begin typing text into the input element and verify that the dropdown is cleared of items. When typing stops, verify that the autocomplete options are updated to match the entered text, if available.

Enter an autocomplete query (e.g. mastodon) and select the option from the list. The dropdown will close and the selected value will be entered into the input element.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
+++ b/ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
@@ -38,6 +38,8 @@ export function AutoCompleteSelect({
 }: AutoCompleteSelectProps) {
     const { isOpen, onToggle, closeSelect } = useSelectToggle();
     const [typeahead, setTypeahead] = useState(selectedOption);
+    // When isTyping is true, autocomplete results will not be displayed. This prevents
+    // a clunky UX where the dropdown results and the user text get out of sync.
     const [isTyping, setIsTyping] = useState(false);
 
     const autocompleteCallback = useCallback(() => {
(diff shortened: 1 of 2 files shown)


dvail:
* master
  * **PR #3816** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3816" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3816?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
```diff
@@ -38,14 +38,20 @@ export function AutoCompleteSelect({
 }: AutoCompleteSelectProps) {
     const { isOpen, onToggle, closeSelect } = useSelectToggle();
     const [typeahead, setTypeahead] = useState(selectedOption);
+    const [isTyping, setIsTyping] = useState(false);
```
pedrottimark, line 41: Judgment call whether to summarize description as a comment about component behavior.

pedrottimark (approved):

PR #3819, opened 2022-11-15
ROX-13582: Push marketing telemetry data to Segment

0x656b694d:
## Description

* Added Segment integration;
* Removed Amplitude integration;
* Refactored;
* Added three more fields to the `/v1/metadata` response to be used by the frontend:
  * Central ID, which is the central deployment UID;
  * Organization ID, which is to be set by fleet manager;
  * Segment Key, which is the Segment write key required for sending events.

### How to enable reporting

Append the central deployment with:
* Environment variable:
```yaml
- name: ROX_TELEMETRY_STORAGE_KEY_V1
  value: ...
```
* Optional variables:
```
ROX_TELEMETRY_ENDPOINT
ROX_TELEMETRY_FREQUENCY
```
* Optional annotations:
```yaml
"rhacs.redhat.com/telemetry-apipaths": "*" # a list of API paths like /v1/pang,/v1/pong
"rhacs.redhat.com/tenant": "..." # tenant ID provided by fleetshard
```

## Checklist

## Testing Performed
https://app.segment.com/redhat-devtools/sources/acs_instances_backend_dev/overview

Metadata response example:
```json
{"version":"3.73.x-133-g445e0297b1-dirty","buildFlavor":"development","releaseBuild":false,"licenseStatus":"VALID","userId":"8di5nQ6yIf+3P9MB9tuMIilMBCuTwrjbelBRUXZ6eGs=","telemetryEndpoint":"","storageKeyV1":"..."}
```

API Call event (notice the base64 encoded userID sha256 hash):
```json
{
  "context": {
    "Central ID": "3cdf2b20-1519-48a9-9b5c-467472421bbd",
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    }
  },
  "event": "API Call",
  "integrations": {},
  "messageId": "db789ac3-abae-47eb-a56d-74f27c2e792a",
  "originalTimestamp": "2022-11-25T17:05:34.100124907Z",
  "properties": {
    "Code": 200,
    "Path": "/api/graphql",
    "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36"
  },
  "receivedAt": "2022-11-25T17:07:44.938Z",
  "sentAt": "2022-11-25T17:07:44.102Z",
  "timestamp": "2022-11-25T17:05:34.935Z",
  "type": "track",
  "userId": "8di5nQ6yIf+3P9MB9tuMIilMBCuTwrjbelBRUXZ6eGs=",
  "writeKey": "..."
}
```


vjwilson:
@0x656b694d , I was hoping to test this image in action, but it looks like the PR build failed.

central/metadata/service/service_impl.go
```diff
@@ -48,6 +51,14 @@ func (s *serviceImpl) GetMetadata(ctx context.Context, _ *v1.Empty) (*v1.Metadat
 		ReleaseBuild:  buildinfo.ReleaseBuild,
 		LicenseStatus: v1.Metadata_VALID,
 	}
+	if marketing.Enabled() {
+		config, err := mpkg.GetDeviceConfig()
+		if err == nil {
+			metadata.CentralId = config.ID
+			metadata.OrganizationId = config.OrgID
+			metadata.SegmentKey = env.SegmentAPIKey.Setting()
+		}
```
stehessel, line 60:
```suggestion
		if config, err := mpkg.GetDeviceConfig(); err == nil {
			metadata.CentralId = config.ID
			metadata.OrganizationId = config.OrgID
			metadata.SegmentKey = env.SegmentAPIKey.Setting()
		}
```
Should we log the error here?

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,63 @@
...
+	opts := v1.GetOptions{}
+	d, err := di.Get(context.Background(), "central", opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get central deployment")
+	}
+	paths := d.GetAnnotations()[annotation]
+
+	config = &Config{
+		ID:       string(d.GetUID()),
+		OrgID:    d.GetAnnotations()[orgID],
```
stehessel, line 53: Should we fail here if `orgID` is not in the annotations?

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,63 @@
...
+	di := clientset.AppsV1().Deployments("stackrox")
+	opts := v1.GetOptions{}
+	d, err := di.Get(context.Background(), "central", opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get central deployment")
+	}
+	paths := d.GetAnnotations()[annotation]
+
+	config = &Config{
+		ID:       string(d.GetUID()),
```
stehessel, line 52: What is this ID? Is it supposed to be the instance ID generated by fleet manager?

central/telemetry/marketing/interceptor.go
```diff
@@ -0,0 +1,51 @@
...
+	"google.golang.org/grpc"
+)
+
+var ignoredPaths = set.NewFrozenSet("/v1/ping", "/v1/metadata")
+
+func track(ctx context.Context, t mpkg.Telemeter, err error, info *grpc.UnaryServerInfo, trackedPaths set.FrozenSet[string]) {
+	userAgent, userID, path, code := getRequestDetails(ctx, err, info)
+
+	// Track the API path and error code of some requests:
+	if !ignoredPaths.Contains(path) && (trackedPaths.Contains("*") || trackedPaths.Contains(path)) {
```
stehessel, line 22:
```suggestion
	if ignoredPaths.Contains(path) {
	    return
	}
	if trackedPaths.Contains("*") || trackedPaths.Contains(path) {
```

pkg/telemetry/marketing/segment/segment.go
```diff
@@ -0,0 +1,123 @@
...
+			log.Error("Cannot close Segment client: %v", err)
+		}
+	}
+}
+
+func (t *segmentTelemeter) TrackProps(event, userID string, props map[string]any) {
+	if t == nil {
+		return
+	}
+	log.Info("Tracking event ", event, " with ", props)
```
stehessel, line 114: This is logged on every gRPC call, right? Seems a bit like unnecessary log spam.

central/metadata/service/service_impl.go
0x656b694d: Today, `/v1/metadata` is called by the frontend periodically for a connection check. We don't want to log an error here, as it would probably be repetitive. The error will be printed by another user  of `GetDeviceConfig()`.
I noticed, though, that `config` could be `nil`, so I'll adapt the condition to check `config` instead.

pkg/telemetry/marketing/device.go
0x656b694d: I'm not sure that we should fail if anything goes wrong with telemetry. `orgID` might never be there for on-prem customers (if we not find a way to identify such customers).

pkg/telemetry/marketing/device.go
0x656b694d: This is central deployment UID, which essentially identifies a central deployment. Should work for both managed and wild installations.

central/telemetry/marketing/gatherer.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (m *marketing) gather() {
+	log.Debug("Starting marketing telemetry data collection.")
+	defer log.Debug("Done with marketing telemetry data collection.")
+
+	totals := make(map[string]any)
+	rs := roles.Singleton()
+
+	addTotal(totals, "PermissionSets", rs.GetAllPermissionSets)
```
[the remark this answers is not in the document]
0x656b694d: This information has not been explicitly requested yet. Could be added at any moment.

pkg/telemetry/marketing/segment/segment.go
0x656b694d: Right, needs to be removed. I used it temporarily, for debugging.

0x656b694d:
> I think in general there is test code missing. Do you plan to add unit tests later?

Yes. Having hard times to find what exactly to test the way it'd make sense.

0x656b694d:
> @0x656b694d , I was hoping to test this image in action, but it looks like the PR build failed.

@vjwilson, I'm sorry about that. There were issues with style. There's a ready to use image now.

central/telemetry/marketing/gatherer.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+}
+
+// Singleton returns the previously initialized telemeter instance.
+func Singleton() *marketing {
+	return m
+}
+
+func (m *marketing) loop() {
+	for !m.stopSig.IsDone() {
```
stehessel, line 59: Probably a go expert should take a look at this, but is this concurrency safe? To my understanding accessing a struct from inside a go routine is not safe, and value passing between different go routines should be done via channels.

0x656b694d: The values are passed by channels in the methods of this primitive from the concurrency package.
Would be nice to have a look from a go expert, indeed.

pkg/telemetry/marketing/telemeter.go
```diff
@@ -13,6 +13,7 @@ type Telemeter interface {
 type Config struct {
 	ID       string
 	OrgID    string
+	SKU      string
```
stehessel, line 16: I used the SKU as example of something could be interesting to correlate. We don't need to send the SKU in Central itself, but I think it would be worthwhile to send the tenant id as a correlation id in Central. That way, we can correlate data from fleet-manager or the console UI (like the SKU, or something else).

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,63 @@
...
+
+	di := clientset.AppsV1().Deployments(env.Namespace.Setting())
+	opts := v1.GetOptions{}
+	d, err := di.Get(context.Background(), "central", opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get central deployment")
+	}
+	paths := d.GetAnnotations()[annotation]
+
+	config = &Config{
```
vjwilson, line 52: We discussed using the same hashing algorithm on the backend and frontend to combine the central ID and Org ID to create an obfuscated Segment identifier.

I propose that the hash happens once in the API, and is passed to the FE along with this info. Doing the same thing in two different place invites problems to arise. 
https://en.wikipedia.org/wiki/Murphy%27s_law

Stephan's question below about what to do when Org ID is not available, begs a follow-up question. If API passes the hash, does central ID + "" still create a valid, unique hash to use in Segment?

vjwilson (commented):
Tested in FE.

pkg/telemetry/marketing/telemeter.go
vjwilson: What would we `SKU` work for self-hosted installs of ACS ?

central/telemetry/marketing/gatherer.go
stehessel: What I mean is that you are accessing the struct (e.g. `m.ticket` `m.telemeter`) from concurrent go routines. I think this is a potential data race, which would require either a mutex of some sort.

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,65 @@
...
+		ID:       string(d.GetUID()),
+		OrgID:    d.GetAnnotations()[orgID],
+		SKU:      d.GetAnnotations()[sku],
```
stehessel, line 55: Please remove the SKU here again, I didn't mean to capture it from within Central. However I would propose to add the Central tenant ID here, which we may use to correlate metrics similar to `OrgID`.

pkg/telemetry/marketing/telemeter.go
stehessel: We don't really have access to the SKU from self-hosted Centrals. In any case we shouldn't capture the SKU from within Central, I just used it as an example of something that we might want to correlate to using the Central tenant ID.

stehessel: Let me give another example of why I think tenant correlation is useful: At some point, someone (maybe us, maybe someone else) will provide business metrics for the ACS cloud service - for example, how much money we are making and how we are spending in infrastructure per tenant. It would be nice to be able to correlate the user behavior of a tenant with the revenue/profit it generates us. This way we would be able to improve our product based on what makes the most financial sense.

central/telemetry/marketing/gatherer.go
0x656b694d: `m.ticker` is set once before starting the loop in the method `Start()` below, then, its internal channel is being read.
`m.telemeter` is set once at the initialization and then used to call `Enqueue()` from the Segment API, which writes a message to a channel. Doesn't look like needing a mutex.

central/telemetry/marketing/interceptor.go
```diff
@@ -0,0 +1,54 @@
+package marketing
```
rukletsov, line 1: My understanding is that this file is for instrumenting API calls to gather related telemetry. Let's make it clear from naming. A couple of suggestions:
* introduce a named function that returns an interceptor `func` based on passed `telemeter` and `trackedPaths`. You can then call it something like `TelemetryAPIInterceptor(...) grpc.UnaryServerInterceptor {...}`
* call the file `api_interceptor.go`
* or move it to `telemetry/interceptors/api.go`

central/telemetry/marketing/interceptor.go
```diff
@@ -0,0 +1,54 @@
...
+	grpcError "github.com/stackrox/rox/pkg/grpc/errors"
+	"github.com/stackrox/rox/pkg/grpc/requestinfo"
+	"github.com/stackrox/rox/pkg/set"
+	mpkg "github.com/stackrox/rox/pkg/telemetry/marketing"
+	"google.golang.org/grpc"
+)
+
+var ignoredPaths = set.NewFrozenSet("/v1/ping", "/v1/metadata")
+
+func track(ctx context.Context, t mpkg.Telemeter, err error, info *grpc.UnaryServerInfo, trackedPaths set.FrozenSet[string]) {
```
rukletsov, line 18: We have `apiGatherer`. Can't it be extended to cover the use cases here?

central/telemetry/marketing/gatherer.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+
+	providerIDNames := make(map[string]string)
+	providerNames := make([]string, len(providers))
+	for _, provider := range providers {
+		providerIDNames[provider.GetId()] = provider.GetName()
+		providerNames = append(providerNames, provider.GetName())
+	}
+	totals["Auth Providers"] = providerNames
```
rukletsov, line 125: Looking at this code it feels this is datastore's business. I guess each datastore should know better how (and whether) to collect its own telemetry and should optinally expose `GatherTelemetry() (map[string]any, error)`, register itself in some global collection of "watched" datastores. Here you would then iterate over that collection, ensure keys are not overwritten, aggregate, and push to the telemetry storage backend.

pkg/telemetry/marketing/telemeter.go
```diff
@@ -0,0 +1,19 @@
+package marketing
+
+// Telemeter defines a common interface for telemetry gatherers.
+//go:generate mockgen-wrapper
+type Telemeter interface {
+	Start()
+	Stop()
+	Identify(props map[string]any)
+	TrackProps(event, userID string, props map[string]any)
```
rukletsov, line 9: Is my understanding correct that `Identify()` is for updating "static" data based on the database state like # of records of different types, etc, while `TrackProps()` is for capturing events? If so, let's maybe rename for clarity.

pkg/telemetry/marketing/telemeter.go
```diff
@@ -0,0 +1,19 @@
+package marketing
+
+// Telemeter defines a common interface for telemetry gatherers.
+//go:generate mockgen-wrapper
+type Telemeter interface {
+	Start()
+	Stop()
```
rukletsov, line 7: Could we push this to the underlying implementation and not expose here? I guess based on the telemetry storage backend the implementation should decide when to open and one to close HTTP connection, how to pool data, etc.

pkg/telemetry/marketing/telemeter.go
rukletsov: Please no unrelated junk. Centrals have no concept of SKU and vaguely of org ID. If you want to forward arbitrary data from, say, Central deployment, to telemetry store, please use a map here so things like `SKU`s are contained in a single place like `GetDeviceConfig()`.

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const annotation = "rhacs.redhat.com/telemetry-apipaths"
+const orgID = "rhacs.redhat.com/organization-id"
+const sku = "rhacs.redhat.com/sku"
+
+var config *Config
+
+// GetDeviceConfig collects the central instance telemetry configuration.
+func GetDeviceConfig() (*Config, error) {
```
rukletsov, line 23: device o_O?

pkg/telemetry/marketing/device.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/client-go/rest"
+)
+
+const annotation = "rhacs.redhat.com/telemetry-apipaths"
+const orgID = "rhacs.redhat.com/organization-id"
+const sku = "rhacs.redhat.com/sku"
+
+var config *Config
+
+// GetDeviceConfig collects the central instance telemetry configuration.
```
rukletsov, line 22: What else beyond central id we want to collect and why?

pkg/telemetry/marketing/device.go
0x656b694d: I'd avoid hashing non-user IDs, as they could be of a use in clear. I'd also avoid combining IDs to a single hash, as we might want to correlate different IDs separately.
User IDs could be hashed by central, indeed.

go.mod
```diff
@@ -358,6 +359,11 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
+require (
```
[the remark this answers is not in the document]
0x656b694d: Not me! Will merge.

proto/api/v1/metadata_service.proto
```diff
@@ -8,6 +8,13 @@ import "api/v1/empty.proto";
 
 package v1;
 
+message Marketing {
+    string user_id = 1;
+    string central_id = 2;
+    string organization_id = 3;
+    string segment_key = 4;
```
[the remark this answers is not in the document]
0x656b694d: Yes.

proto/api/v1/metadata_service.proto
```diff
@@ -8,6 +8,13 @@ import "api/v1/empty.proto";
 
 package v1;
 
+message Marketing {
```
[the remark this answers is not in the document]
0x656b694d: Metadata used to serve the license info (kind of telemetry). That's the only why, apart from it's an existing service.
The existing TelemetryService is about the old implementation, I tried to not mix the two.

proto/api/v1/metadata_service.proto
```diff
@@ -8,6 +8,13 @@ import "api/v1/empty.proto";
...
+    string user_id = 1;
```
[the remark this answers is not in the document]
0x656b694d: The plan is to hash user ID to anonymize the user, and the function will be implemented in central so that UI would use it as is.

pkg/telemetry/marketing/device.go
0x656b694d: `GetInstanceConfig`?

pkg/telemetry/marketing/segment/segment.go
```diff
@@ -0,0 +1,123 @@
...
+		traits.Set(k, v)
+	}
+	log.Info("Identifying with ", identity)
+	if err := t.client.Enqueue(identity); err != nil {
+		log.Error("Cannot enqueue Segment identity event: ", err)
+	}
+}
+
+// Init creates and initializes a Segment telemeter instance.
+func Init(config *marketing.Config) marketing.Telemeter {
```
dhaus67, line 54: nit: we typically name all functions that use `once.Do` as `Singleton()` instead of `Init()`.

pkg/grpc/server.go
```diff
@@ -230,7 +229,7 @@ func (a *apiImpl) connectToLocalEndpoint(dialCtxFunc pipeconn.DialContextFunc) (
 			return dialCtxFunc(ctx)
 		}),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxResponseMsgSize())),
-		grpc.WithUserAgent(clientconn.GetUserAgent()))
```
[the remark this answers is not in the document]
0x656b694d: Limiting an effect of a bug (ROX-13596), which will be addressed separately.

central/telemetry/marketing/marketing.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// Enabled returns true if marketing telemetery data collection is enabled.
+func Enabled() bool {
+	return segment.Enabled()
+}
+
+// Init initializes the periodic telemetry data gatherer and returns an GRPC API
+// call inteceptor. Returns nil if telemetry data collection is disabled.
+func Init() grpc.UnaryServerInterceptor {
```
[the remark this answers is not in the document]
0x656b694d: No, it doesn't make sense to call `Init()` multiple times, that would be a logic error, like calling any random apparently non-reentrant function multiple times. It's meant to be called once by `main()`, which in turn is also called once, hopefully.
The example of `gatherers/singleton.go` confuses me, as it looks to me like an over-engineering: that `Singleton()` is called exactly once from `main()`, and so there's no reason in implementing the Singleton pattern.
There are, though, other examples of singleton services, e.g. https://github.com/stackrox/stackrox/blob/53cd26d097c98edbc73618c5599f440831885c08/central/cluster/datastore/singleton.go#L91.

Nevertheless, here I don't see a use for a singleton.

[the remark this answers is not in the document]
rukletsov: What is a benefit of having a non-reentrant function here compared with the one that is no-op for every subsequent call? If there is none, than a singleton pattern used widely in our codebase is a consistent, future-proof approach.

I'm not sure I agree with the "over-engineering" argument. `sync.Once.Do()` is probably one of the simplest patterns and fits into a couple of lines **in a single file**. On the other hand one 1) would need to track code paths **in who knows what files** to ensure `Init()` is not called more than once and 2) don't forget `Init()` is non-reentrant in the future.

pkg/telemetry/marketing/device.go
0x656b694d: We need to have a correlation ID to later join data from other telemetry reporters. This is yet to be defined.

pkg/telemetry/marketing/device.go
vjwilson: Are we sure that user IDs can never be the same for two different instances of Central ?

As @rukletsov mentioned in another comment, we could send user IDs and org IDs, along with other optional useful metadata, in an attached dictionary.

rukletsov: We shall not assume uniqueness of user IDs across Centrals but that's fine as long as we use `<CentralID, UserID>` to identify universally unique ACS users.

central/telemetry/phonehome/gatherer.go
```diff
@@ -0,0 +1,158 @@
...
+	log.Debug("Starting telemetry data collection.")
+	defer log.Debug("Done with telemetry data collection.")
+
+	totals := make(map[string]any)
+	rs := roles.Singleton()
+
+	addTotal(g.ctx, totals, "PermissionSets", rs.GetAllPermissionSets)
+	addTotal(g.ctx, totals, "Roles", rs.GetAllRoles)
+	addTotal(g.ctx, totals, "Access Scopes", rs.GetAllAccessScopes)
+	addTotal(g.ctx, totals, "Signature Integrations", si.Singleton().GetAllSignatureIntegrations)
```
dhaus67, line 127: I think it would be more beneficial to have the _things we want to collect_ more or less injected within the gatherer in form of a separate abstraction, such as:
```
type gathererFunc func(ctx context.Context) (map[string]any, error)
```

Right now, if we were to arbitrarily add additional datastores we want to collect from to our targets, we would always have to touch this specific `gather()` function. I'm not sure whether it's planned in the future to do this more or less arbitrarily (e.g. defining _what_ to collect based on each version via e.g. env variables / other configuration (so that we can differentiate between information collected from on-prem vs ACSCS), but this would certainly feel like a better abstraction.

Additionally, the currently custom login w.r.t auth providers and groups should be moved to either a separate function implementing the abstraction, or even the datastore itself. I would have expected the gatherer to be just a "stupid" component without knowing specific details about the types of proto messages it is gathering (especially because an abstraction like this [already existed for the "old" telemetry](https://github.com/stackrox/stackrox/tree/master/central/telemetry/gatherers))

central/telemetry/phonehome/gatherer_test.go
```diff
@@ -0,0 +1,81 @@
...
+	suite.TearDownTestSuite
+} = (*gathererTestSuite)(nil)
+
+func (s *gathererTestSuite) SetupTest() {
+	s.mockCtrl = gomock.NewController(s.T())
+}
+
+func (s *gathererTestSuite) TearDownTest() {
+	s.mockCtrl.Finish()
+}
```
dhaus67, line 30: No need to call [mockCtrl.Finish](https://pkg.go.dev/github.com/golang/mock/gomock#Controller.Finish) if we passed `testing.T` to the `NewController` function.
You can remove this + the `suite.TearDownTestSuite` part.

central/telemetry/phonehome/interceptor.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if ri.HTTPRequest != nil && ri.HTTPRequest.URL != nil {
+		method = ri.HTTPRequest.URL.Path
+		code = grpcError.ErrToHTTPStatus(err)
+	} else if info != nil {
+		method = info.FullMethod
+		code = int(erroxGRPC.RoxErrorToGRPCCode(err))
+	} else {
+		method = "unknown"
+		code = -1
```
dhaus67, line 63: Might be good to have a comment explaining in which scenarios we would fall into this (or just simply that we do not expect this to be the case at all but have it for safety reasons).

central/telemetry/phonehome/interceptor_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *interceptorTestSuite) SetupTest() {
+	s.mockCtrl = gomock.NewController(s.T())
+	s.mockTelemeter = mocks.NewMockTelemeter(s.mockCtrl)
+}
+
+func (s *interceptorTestSuite) TearDownTest() {
+	s.mockCtrl.Finish()
+}
```
dhaus67, line 44: Same notion as above w.r.t. gomock and the finish function.

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -0,0 +1,121 @@
...
+	log.Info("Identifying with ", identity)
+	if err := t.client.Enqueue(identity); err != nil {
+		log.Error("Cannot enqueue Segment identity event: ", err)
+	}
+}
+
+// NewTelemeter creates and initializes a Segment telemeter instance.
+func NewTelemeter(userID string, identity map[string]any) *segmentTelemeter {
+	key := env.TelemetryStorageKey.Setting()
+	server := ""
```
dhaus67, line 51: I suppose this requires a `TODO` / comment that it's a placeholder value? Or should this always be empty?

pkg/telemetry/phonehome/instance_config_test.go
```diff
@@ -0,0 +1,48 @@
...
+	suite.TearDownTestSuite
+} = (*configTestSuite)(nil)
+
+func (s *configTestSuite) SetupTest() {
+	s.mockCtrl = gomock.NewController(s.T())
+}
+
+func (s *configTestSuite) TearDownTest() {
+	s.mockCtrl.Finish()
+}
```
dhaus67, line 27: Same as above w.r.t. to gomock.

pkg/telemetry/phonehome/instance_config.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/stackrox/rox/pkg/sync"
+	"github.com/stackrox/rox/pkg/version"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	apiPathsAnnotation = "rhacs.redhat.com/telemetry-apipaths"
+	csIDAnnotation     = "rhacs.redhat.com/cs-identity"
```
dhaus67, line 25: I know we currently have some labels / annotations that follow the pattern `x.stackrox.io/y`, I'm not sure whether this should be the scheme we follow or not 🤷 

@SimonBaeumer do you have any particular guidance on the pattern we use for labels / annotations of our K8S objects?

pkg/telemetry/phonehome/instance_config.go
```diff
@@ -0,0 +1,131 @@
...
+	once.Do(func() {
+		var err error
+		if config, err = getInstanceConfig(); err != nil {
+			log.Error("Failed to get telemetry configuration: ", err)
+		} else {
+			log.Info("Telemetry device ID:", config.CentralID)
+			log.Info("API path telemetry enabled for: ", config.APIPaths)
+			if config.CSProperties != nil {
+				log.Info("Cloud Services identity: ", config.CSProperties)
+			}
```
dhaus67, line 102: Since we fail open for retrieving the `csIdAnnotation` on the central deployment, might be worthwhile to print a log message in case we do not have cloud service identities, so it's easier to spot this later on.

pkg/telemetry/phonehome/instance_config.go
```diff
@@ -0,0 +1,131 @@
...
+	metadata := map[string]string{
+		"UserId":    "unauthenticated",
+		"CentralId": config.CentralID,
+	}
+	if id != nil {
+		metadata["UserId"] = HashUserID(id.UID())
+	}
+	for k, v := range config.CSProperties {
+		metadata[k] = v
+	}
```
dhaus67, line 129: What will be a part of those `CSProperties`? Should they always be returned for unauthenticated user?
(The PR description mentions something about organization ID, but this certainly shouldn't be exposed to unauthenticated users, so it might be outdated).

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (l *logWrapper) Errorf(format string, args ...any) {
+	l.internal.Errorf(format, args...)
+}
+
+func initSegment(userID string, identity map[string]any, key, server string) *segmentTelemeter {
+	segmentConfig := segment.Config{
+		Endpoint: server,
+		Interval: 5 * time.Minute,
```
dhaus67, line 70: I saw the environment variable `ROX_TELEMETRY_FREQUENCY`. Should this be reused here, or is it from the "old" telemetry implementation (so we can just remove it to avoid confusion)?

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -0,0 +1,121 @@
...
+				"Central ID": userID,
+			},
+		},
+	}
+
+	client, err := segment.NewWithConfig(key, segmentConfig)
+	if err != nil {
+		log.Error("Cannot initialize Segment client: ", err)
+		return nil
+	}
```
dhaus67, line 83: I think this might be a candidate for `utils.Should`, as we define the config programatically and "expect" it to work.

central/main.go
```diff
@@ -312,6 +313,7 @@ func startServices() {
 	pruning.Singleton().Start()
 	gatherer.Singleton().Start()
 	vulnRequestManager.Singleton().Start()
+	phonehome.GathererSingleton().Start()
```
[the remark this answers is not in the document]
0x656b694d: Even though the condition is already available as `phonehome.Enabled()`, I think it is not required to be used here thanks to the nil check.
Extracting the nil check from the methods would require all usage places to check for the environment variable. The approach of nil instance in case of disabled telemetry looks safe and transparent to me.

central/telemetry/phonehome/gatherer.go
```diff
@@ -0,0 +1,158 @@
...
+func (g *gatherer) loop() {
+	g.ticker = time.NewTicker(g.period)
+	for !g.stopSig.IsDone() {
+		select {
+		case <-g.ticker.C:
+			go g.f(g)
+		case <-g.stopSig.Done():
+			g.ticker.Stop()
+			g.ticker = nil
+			g.cancel()
```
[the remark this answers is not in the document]
0x656b694d: You suspect correctly. `cancel()` would cancel whatever the `gatherFunc` is doing within this context. This function (i.e. `gather()`) passes the context to the datastore GetAll* methods which may respect the cancellation or not.
Thanks for the nit, I discovered `concurrency.DependentContext`, which looks appropriate here.

pkg/telemetry/phonehome/instance_config.go
0x656b694d: I took this `rhacs.redhat.com` from the fleetshard code.

pkg/telemetry/phonehome/instance_config.go
0x656b694d: [outdated]

pkg/telemetry/phonehome/instance_config.go
0x656b694d: [removed]

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: The variable is from the legacy/parallel implementation, and it's not used.
Nice idea, it might be taken for this use I think.

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: I try to make the telemetry as less intrusive as possible, even for the dev builds: it should never break any flow or panic.

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: Empty is also a good value, SDK will put the default endpoint.
I'll reuse the existing `ROX_TELEMETRY_ENDPOINT` variable, just for why not, if one day we decide to use some specific endpoint.

central/telemetry/phonehome/gatherer.go
0x656b694d: Please check the updated version. I had to shuffle the code a little to avoid circular dependencies with authn package, which imports roles.

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (l *logWrapper) Errorf(format string, args ...any) {
+	l.internal.Errorf(format, args...)
+}
+
+func initSegment(userID string, identity map[string]any, key, server string) *segmentTelemeter {
+	segmentConfig := segment.Config{
+		Endpoint: server,
+		Interval: env.TelemetryFrequency.DurationSetting(),
+		Logger:   &logWrapper{internal: log},
```
dhaus67, line 48: We use the `proxy.RoundTripper` to ensure that the transport picks up all relevant proxy settings (including the `all_proxy` setting):
```suggestion
		Interval: env.TelemetryFrequency.DurationSetting(),
		Transport: proxy.RoundTripper(),
		Logger:   &logWrapper{internal: log},
```

central/authprovider/datastore/telemetry.go
```diff
@@ -0,0 +1,45 @@
...
+	providerNames := make([]string, 0, len(providers))
+	for _, provider := range providers {
+		providerIDNames[provider.GetId()] = provider.GetName()
+		providerNames = append(providerNames, provider.GetName())
+	}
+	props["Auth Providers"] = providerNames
+
+	groups, err := groupDataStore.Singleton().GetAll(ctx)
+	if err != nil {
+		return props, errors.Wrap(err, "failed to get Groups")
```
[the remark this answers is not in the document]
0x656b694d: I understand it violates the expectation of nil result in case of error. Here though we can still get the auth providers, which is more useful than nothing.

central/telemetry/phonehome/gatherer_test.go
```diff
@@ -0,0 +1,76 @@
...
+		stop.Wait()
+		gptr.Stop()
+	}()
+	s.NotNil(gptr)
+	gptr.Start()
+
+	<-gptr.ctx.Done()
+	s.ErrorIs(gptr.ctx.Err(), context.Canceled)
+
+	s.ErrorIs(gptr.ctx.Err(), context.Canceled)
```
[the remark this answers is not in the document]
0x656b694d: oops

central/telemetry/phonehome/interceptor.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+var (
+	ignoredPaths = []string{"/v1/ping", "/v1/metadata", "/static/"}
+)
+
+func track(ctx context.Context, t pkgPH.Telemeter, err error, info *grpc.UnaryServerInfo, trackedPaths set.FrozenSet[string]) {
+	userAgent, userID, path, code := getRequestDetails(ctx, err, info)
+
+	// Track the API path and error code of some requests:
```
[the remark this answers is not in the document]
0x656b694d: well, the loop below may return from the function, which is partially why only _some_ requests are tracked. So the comment makes sense before this filtering.

pkg/telemetry/phonehome/instance_config.go
```diff
@@ -0,0 +1,108 @@
...
+	if !ok {
+		paths = "*"
+	}
+
+	orchestrator := storage.ClusterType_KUBERNETES_CLUSTER.String()
+	if env.OpenshiftAPI.BooleanSetting() {
+		orchestrator = storage.ClusterType_OPENSHIFT_CLUSTER.String()
+	}
+
+	centralID := string(central.GetUID())
```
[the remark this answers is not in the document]
0x656b694d: Indeed, that sounds like a better way to get a central ID. I'll investigate.

0x656b694d:
Superseded by #4070.

PR #3820, opened 2022-11-15
Disable auto code generator for postgres migrations

c-du:
## Description

Disable auto code generator for postgres migrations. 
This one will be after the schema freeze PR and for 3.74 https://github.com/stackrox/stackrox/pull/3769
## Checklist
- [ ] Investigated and inspected CI test results~
- ~[ ] Determined and documented upgrade steps`

## Testing Performed

TODO(replace-me)
screenshots.


md2119 (commented):
Can we add a check in the generator code to errors out if a migration flag is encountered? It will ensure that nobody mistakenly adds a migration.

dashrews78:
Do we need to update the comment in the generated files or is that OK to leave?
`// Code generated by pg-bindings generator. DO NOT EDIT.`

Or do we just deal with that if we need to update something in one of those files?

c-du:
> 

Let's leave it as is. Most likely and ideally we don't need to update them anymore. They would be obsoleted in a few releases.

c-du:
> Can we add a check in the generator code to errors out if a migration flag is encountered? It will ensure that nobody mistakenly adds a migration.
In that case, it would be better to remove the migration flags in the generator.

md2119 (approved):

dashrews78 (approved):

PR #3821, opened 2022-11-15
Dashrews/ROX-13205 allow a Rocks DB backup to be restored to Postgres

dashrews78:
## Description

Updates so that a backup taken using Rocks DB can be restored but to a Postgres database.

## Checklist

## Testing Performed

Added UT test, updated the clone tests and added this case to the upgrade tests.

### DIFF
--- a/central/main.go
+++ b/central/main.go
@@ -280,7 +280,7 @@ func main() {
 
 	// Now that we verified that the DB can be loaded, remove the .backup directory
 	if err := migrations.SafeRemoveDBWithSymbolicLink(filepath.Join(migrations.DBMountPath(), migrations.GetBackupClone())); err != nil {
-		log.Errorf("Failed to remove backup DB: %v", err)
+		log.Fatalf("Failed to remove backup DB: %v", err)
 	}
 
 	// Start the prometheus metrics server
--- a/central/main.go
+++ b/central/main.go
@@ -274,13 +274,13 @@ func main() {
 		}
 		versionUtils.SetCurrentVersionPostgres(globaldb.GetPostgres())
 	} else {
-		// Now that we verified that the DB can be loaded, remove the .backup directory
-		if err := migrations.SafeRemoveDBWithSymbolicLink(filepath.Join(migrations.DBMountPath(), migrations.GetBackupClone())); err != nil {
-			log.Errorf("Failed to remove backup DB: %v", err)
-		}
-
 		// Update last associated software version on DBs.
 		migrations.SetCurrent(option.CentralOptions.DBPathBase)
+	}
+
+	// Now that we verified that the DB can be loaded, remove the .backup directory
+	if err := migrations.SafeRemoveDBWithSymbolicLink(filepath.Join(migrations.DBMountPath(), migrations.GetBackupClone())); err != nil {
+		log.Errorf("Failed to remove backup DB: %v", err)
 	}
 
 	// Start the prometheus metrics server
--- a/migrator/clone/db_clone_manager_impl.go
+++ b/migrator/clone/db_clone_manager_impl.go
@@ -1,9 +1,6 @@
 package clone
 
 import (
-	"fmt"
-	"runtime"
-
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/stackrox/rox/migrator/clone/postgres"
 	"github.com/stackrox/rox/migrator/clone/rocksdb"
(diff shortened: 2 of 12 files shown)


migrator/clone/db_clone_manager_impl.go
```diff
@@ -126,7 +136,7 @@ func (d *dbCloneManagerImpl) Persist(cloneName string, pgClone string, persistBo
 
 			// If the versions do not match, we have updated another time with Postgres,
 			// so we can no longer roll back to RocksDB.
-			if rocksVersion == nil || (currentPostgresVersion != nil && rocksVersion.MainVersion != currentPostgresVersion.MainVersion) {
+			if rocksVersion != nil && currentPostgresVersion != nil && rocksVersion.MainVersion != currentPostgresVersion.MainVersion {
```
c-du, line 139: I do not quite understand this change. If rocksVersion is nil, likely Rocksdb current does not exist. The new codes are for sure correct. But why did we decommission in this case before?

pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -337,7 +337,7 @@ func getAvailablePostgresCapacity(postgresConfig *pgxpool.Config) (int64, error)
 
 	// We should only get the header row and the row for the size of $PGDATA.  If we
 	// get more than that, then $PGDATA is not defined
-	if len(rawCapacityInfo) > 2 {
+	if len(rawCapacityInfo) < 2 {
```
c-du, line 340: Why not check it should be 2?

migrator/clone/postgres/db_clone_manager.go
```diff
@@ -21,6 +21,9 @@ const (
 
 	// TempClone - temp postgres clone
 	TempClone = "central_temp"
+
+	// RestoreFromRocksClone - special clone for restoring from RocksDB
+	RestoreFromRocksClone = "central_rocks"
```
c-du, line 26: Why do you start a new Clone name? Can we reuse RestoreClone? They shall not exist at the same time. If we enforce this rule in restore handler and anywhere else, it could be simpler in logic.

central/main.go
```diff
@@ -274,15 +274,15 @@ func main() {
...
-			log.Errorf("Failed to remove backup DB: %v", err)
-		}
-
 		// Update last associated software version on DBs.
 		migrations.SetCurrent(option.CentralOptions.DBPathBase)
 	}
 
+	// Now that we verified that the DB can be loaded, remove the .backup directory
+	if err := migrations.SafeRemoveDBWithSymbolicLink(filepath.Join(migrations.DBMountPath(), migrations.GetBackupClone())); err != nil {
+		log.Errorf("Failed to remove backup DB: %v", err)
```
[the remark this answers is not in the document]
dashrews78: I agree it should not, however, this is removing the backup clone.  That would not impact the functionality of central.  To me, error feels like the proper level.

migrator/clone/db_clone_manager_impl.go
dashrews78: @c-du The previous idea was that if rocksDB was empty we could go ahead and clean it up.  That was really an unnecessary assumption and causes issues when testing things like this.  And this logic is a little cleaner.

dashrews78: @c-du The previous idea was that if rocksDB was empty we could go ahead and clean it up.  That was really an unnecessary assumption and causes issues when testing things like this.  And this logic is a little cleaner.

central/main.go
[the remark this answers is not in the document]
c-du: If we fail to remove backup db clone here, when the central reboot, we will find there is a backup clone and we will try to restore it. This way we will lose all date updates.

migrator/clone/postgres/db_clone_manager.go
c-du: We have logs and .backup to help us differentiate this situation. The good thing for reuse is: 1) we do not need to maintain RestoreFromRocksClone, like remove the residue from earlier failed operations. 2) We do not need to make sure there is at most one, RestoreFromRocksClone or RestoreClone at any time in a system.
I feel this is easier for me. If you believe this one is better for you, please double check and make sure ^.

dashrews78: my concern is for the case where it fails in the middle and migration restarts.

postgres->postgres we have a central_restore from a dump that is complete though it may need postgres migrations.  If we fail and have a central_restore in that case, it will pick up that we have central_restore and process it.  We would have to put in logic in there to figure out if we need to find the RocksDB to migrate further.  Whereas my approach I use a separate DB, ignore it on the scan, and basically throw it away if something happens during the migration and start them over.

with your approach, we could pick up where we left off in theory.  I will think about how to move the logic to reuse it.

central/main.go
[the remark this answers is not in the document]
dashrews78: interesting.  I don't see any code that tries to restore the backup.  I could be missing it.  I will make the update though

migrator/clone/postgres/db_clone_manager.go
dashrews78: I took a look and your approach does simplify things.  I have implemented that.

c-du (approved):

PR #3824, opened 2022-11-16
Add more descriptive validation and errors to collection form

dvail:
## Description

This adds descriptive inline error messages to the UI when there are validation issues with the Collection form. This primarily handles error situations that cannot easily be done through `yup` validation, such as server response errors and errors that results in invalid _combinations_ of fields.

When an error occurs on submit, an inline alert will be placed at the top of the page and will scroll the user to the top. More descriptive errors will be placed inline as close as possible to the problem site, either inline as a PatternFly `helperTextInvalid` or as an inline alert in the relevant section.

Note: currently most of these errors occur server side and are triggered when the user attempts to save a collection. Once the dry run endpoint is implemented and used for collection results, these errors should be occurring each time the form value changes to an invalid configuration. Once this happens, we might want to distinguish between "save" errors and "dry run" errors, and not automatically scroll the page when the latter occurs since the alert should appear inline directly where the user currently is.

## Checklist

## Testing Performed

Manual testing of known error cases:

The name field is left empty:
image

A rule dropdown is left empty (these do not have inline error messages as each field is not currently wrapped in its own `FormGroup`. As a follow up I'd like to see if it makes sense semantically to wrap each in a field group if the group is not going to contain a label.)
image

Saving a collection that has the same name as another collection in the system. After this error occurs, changing the value of the name input should clear the error message.
image
image

A regex value is used that does not validate on the server. Currently the server response does not give us any indication of _which_ value is incorrect, so the error is placed at the top of the section. This is one case where it felt valuable to show the actual server error message to the user.
image

A collection is saved that contains an attached collection that forms a loop.
image
image
Detaching the offending collection will clear this error:
image

An attempt is made to save an empty collection:
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
+    Alert,
     Button,
     EmptyState,
     EmptyStateIcon,
@@ -25,6 +26,7 @@ import { CollectionPageAction } from './collections.utils';
 import RuleSelector from './RuleSelector';
 import CollectionAttacher, { CollectionAttacherProps } from './CollectionAttacher';
 import { Collection, ScopedResourceSelector, SelectorEntityType } from './types';
+import { CollectionSaveError } from './errorUtils';
 
 function AttachedCollectionTable({
     collections,
@@ -64,8 +66,12 @@ export type CollectionFormProps = {
     /* collection responses for the embedded collections of `initialData` */
     initialEmbeddedCollections: CollectionResponse[];
     onSubmit: (collection: Collection) => Promise<void>;
+    saveError?: CollectionSaveError | undefined;
+    clearSaveError?: () => void;
     /* Table cells to render for each collection in the CollectionAttacher component */
-    collectionTableCells: CollectionAttacherProps['collectionTableCells'];
+    getCollectionTableCells: (
+        collectionErrorId: string | undefined
+    ) => CollectionAttacherProps['collectionTableCells'];
     /* content to render before the main form */
     headerContent?: ReactElement;
 };
@@ -103,8 +109,10 @@ function CollectionForm({
     action,
     initialData,
     initialEmbeddedCollections,
+    saveError,
+    clearSaveError = () => {},
     onSubmit,
-    collectionTableCells,
+    getCollectionTableCells,
 }: CollectionFormProps) {
     const history = useHistory();
 
@@ -112,8 +120,7 @@ function CollectionForm({
 
     const {
         values,
-        isValid,
-        errors,
+        errors: formikErrors,
         handleChange,
         handleBlur,
         setFieldValue,
@@ -141,6 +148,19 @@ function CollectionForm({
         }),
     });
 
+    const errors = {
+        ...formikErrors,
+    };
+
+    // We can associate this type of server error to a specific field, so update the formik errors
+    if (saveError?.type === 'DuplicateName') {
+        errors.name = saveError.message;
+    }
+
+    const collectionTableCells = getCollectionTableCells(
+        saveError?.type === 'CollectionLoop' ? saveError.loopId : undefined
+    );
+
     function onCancelSave() {
         history.push({ pathname: `${collectionsBasePath}` });
     }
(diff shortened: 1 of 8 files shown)


dvail:
* master
  * **PR #3824** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3824" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3824?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionFormDrawer.tsx
```diff
@@ -33,8 +32,10 @@ export type CollectionFormDrawerProps = {
     isDrawerOpen: boolean;
     toggleDrawer: (isOpen: boolean) => void;
     headerContent?: ReactElement;
-    onSubmit: (collection: Collection) => Promise<void>;
-    collectionTableCells: CollectionFormProps['collectionTableCells'];
+    onSubmit: CollectionFormProps['onSubmit'];
```
dvail, line 35: This has been a helpful pattern when passing props several levels deep. By defining the type of a prop as the key of the child component's props, this makes it so that changes only need to occur on the lowest level component and the highest level component. I think it also helps convey that "this prop is going to be passed deeper".

pedrottimark: You keep teaching me more fluent application of TypeScript.

This pattern seems worthwhile to share at the next team meeting.

pedrottimark: Although the team agreed to export component props type with naming convention, the need has been rare until this idiom.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -141,6 +148,19 @@ function CollectionForm({
         }),
     });
 
+    const errors = {
+        ...formikErrors,
```
pedrottimark, line 152: A thought which comes to mind, not necessarily better: what is pro and con if a pure helper function at module scope has `formikErrors` and `saveError` as arguments? Early return if `saveError` is undefined, and encapsulate conditional logic for a possibly growing list of errors?

ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
```diff
@@ -108,46 +112,83 @@ function CollectionsFormPage({
...
+            if (pageAction.type === 'view') {
+                // Logically should not happen, but just in case
+                return reject(new Error('A Collection form has been submitted in read-only view'));
+            }
+            const isEmptyCollection =
+                Object.values(collection.resourceSelector).every(({ type }) => type === 'All') &&
+                collection.embeddedCollectionIds.length === 0;
+
+            if (isEmptyCollection) {
+                return reject(new Error('Cannot save an empty collection'));
```
pedrottimark, line 127: A question for understanding, not a change request: unlike some other validation errors, yup validation might in theory be able to detect empty collection?

ui/apps/platform/src/Containers/Collections/errorUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+        };
+    }
+
+    if (
+        // Error for collection save
+        /collections must have non-empty, unique `name` values/.test(rawMessage) ||
+        // Error for collection update
+        /name already in use/.test(rawMessage)
+    ) {
+        return { type: 'DuplicateName', message: 'Collection name values must be unique' };
```
pedrottimark, line 52: The `message` could be more parallel to **name is a required field**
* Although the message I quoted is yup default? Because **name** lowercase from key instead of sentence case from field label?
* How about drop **values** to get either **Collection names must be unique** if that helps understanding or **Name must be unique** for most direct relationship with label

pedrottimark (approved):

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
dvail: I think the main con is the additional memory allocation right now, which is probably negligible. Since it is just the one error being appended I felt like it was easier to follow to do this inline during render instead of calling out to a helper function. Definitely agree that if the list of errors that we want to attach to Formik validation grows that this would be cleaner to pull out into a separate function though.

ui/apps/platform/src/Containers/Collections/CollectionsFormPage.tsx
dvail: I agree that yup could do this in theory, since we can determine an empty collection 100% on the client. In practice it didn't seem readily apparent on how to accomplish this by reading through the `yup` docs and I went with this solution since the empty checks were fairly straightforward. I think part of the complication is that the validation would apply both to `values.embeddedCollectionIds` (array) and `values.resourceSelectors` (Record), the latter of which seemed to need to have errors specified for each of the `Deployment`, `Namespace`, and `Cluster` property keys.

I'll make a note to look into this again though and see if there is a clean idiomatic way to get the same behavior via `yup`.

PR #3834, opened 2022-11-16
[ROX-9864] : Sleep after creating deployment to wait for image CVEs being discovered and added to db

charmik-redhat:
## Description

It seems like the failure occurs when all image CVEs are not yet discovered and added to DB or when their cvss scores are not yet populated. This leads to different imageCVE counts between two graphQL queries, or different sort orders based on cvss score. Also, added a ToString annotation to CVE class used by the test to see better logs if test failure occurs again.

## Checklist

## Testing Performed

Not needed for a test update.


qa-tests-backend/src/test/groovy/CSVTest.groovy
```diff
@@ -306,6 +310,7 @@ class CSVTest extends BaseSpecification {
     }
 
     @EqualsAndHashCode(includeFields = true)
+    @ToString(includeFields = true)
```
clickboo, line 313: Nit: Would these cause the log to be very noisy if all cves are being logged?

charmik-redhat: That is right, I should only include id and cvss since they are relevant to debugging.

clickboo (approved):

PR #3836, opened 2022-11-17
Call interceptAndWaitForResponses function in container-specific visit functions

pedrottimark:
### Problem

Under pressure to solve timing failures in tests for various containers:

1. I replaced `cy.visit` method with generic `visit` function which intercepts requests and waits for responses according to `routeMatcherMapForAuthenticatedRoutes` object.

    * That worked well, with one exception: I incorrect includes credential expiry requests, which are not in sequence preceding, but instead in parallel with container-specific requests. Fixed in #3577

2. I wrote container-specific `visitWhatever` functions with `routeMatcherMapForWhatever` object and optional `staticResponseMap` argument.

    * However, I made a mistake in the pattern to call `visit` function with those arguments, **and then** assert page address or page header. Destination page address is an obvious prerequisite to requests and components usually render page header **before** instead of **after* responses to requests. Because the order of wait (API, API, and then DOM) contradicts the order of events, the pattern wastes the extra 4 seconds wait for DOM assertions.

    * Especially if containers or entity components take responsibility to make requests, but even if they render data from Redux store via sagas, it is possible to assert page address or page header **and then** intercept requests and wait for responses. To prevent timing problems, **alternate** API and DOM consistent with order of events:
        * API: 10 seconds wait for requests and 30 seconds wait for responses for authenticated routes in `visit` function.
        * DOM: 4 seconds wait for container-specific assertions in `visitWhatever` function.
        * API: 10 seconds wait for requests and 30 seconds wait for responses for container page in `visitWhatever` function.
        * DOM: 4 secondds wait for test-specific assertions after `visitWhatever` function call.

### Solution

TODO

## Checklist

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/constants/SystemHealth.js
+++ b/ui/apps/platform/cypress/constants/SystemHealth.js
@@ -1,7 +1,5 @@
 import scopeSelectors from '../helpers/scopeSelectors';
 
-export const systemHealthUrl = '/main/system-health';
-
 export const selectors = {
     bundle: {
         generateDiagnosticBundleButton: 'button:contains("Generate Diagnostic Bundle")',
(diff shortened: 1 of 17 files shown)


pedrottimark:
The cure was worse than the disease.

PR #3839, opened 2022-11-17
ROX-11750: SAC tests for imagecomponentedge datastore

rhybrillou:
## Description

With the migration from in-process data storage to an external postgres DB, the scoped access control data filtering logic was re-designed. Tests are being added to ensure that the SAC filtering behaviour is consistent across datastore engines.

This PR covers testing for the ImageComponentEdge storage.

## Checklist

## Testing Performed

CI is sufficient. The change is about adding tests.

When writing the tests, it was discovered that some datastores lack some elements in context for search tests to run correctly in pre-postgres mode. As a consequence, the search tests are disabled in non-postgres mode. In order to validate the baseline, the searcher code was altered to add the `graph.Context` wrapper in the common functions, then the change was reverted.

### DIFF
--- a/central/imagecomponentedge/datastore/datastore_sac_test.go
+++ b/central/imagecomponentedge/datastore/datastore_sac_test.go
@@ -1,0 +1,364 @@
+package datastore
+
+import (
+	"context"
+	"testing"
+
+	dackboxTestUtils "github.com/stackrox/rox/central/dackbox/testutils"
+	"github.com/stackrox/rox/central/role/resources"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/dackbox/edges"
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/fixtures"
+	sacTestUtils "github.com/stackrox/rox/pkg/sac/testutils"
+	"github.com/stackrox/rox/pkg/scancomponent"
+	"github.com/stackrox/rox/pkg/search"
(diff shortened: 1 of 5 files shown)


central/imagecomponentedge/datastore/datastore_sac_test.go
```diff
@@ -0,0 +1,368 @@
...
+		s.Equal(expectedCount, len(results))
+		for _, r := range results {
+			s.True(c.expectedEdgeFound[r.GetId()])
+		}
+	}
+
+}
+
+func (s *imageComponentEdgeDatastoreSACTestSuite) TestSearchRawEdges() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
```
janisz, line 344: Why do we need this test? If it's not working on PG then we will remove it once we switch to PG only mode.

rhybrillou: So far, the tests do work against postgres.
When it comes to the old dackbox, graph info is needed in the context, otherwise the searcher panics. An example of working searcher would be the `componentcveedge` one (https://github.com/stackrox/stackrox/blob/master/central/componentcveedge/search/searcher_impl.go#L62-L64). In the datastore at hand, The graph info is missing from the context.
I see two approaches here:
- change the existing dackbox searcher code to inject the graph in context and make the datastore tests work (was done to establish the test baseline, then reverted, see the second commit).
- disable the tests in pre-postgres mode (done with the revert of the above).

janisz (approved):

PR #3841, opened 2022-11-17
ROX-13468: Skip risk sort by Priority integration test

pedrottimark:
### Test failure

> Risk page without mock API should sort the Priority column

> AssertionError: expected 'notAscendingNumberValuesFromElements(24, 19)' to equal ''

prow

cypress/integration/risk.test.js

Most recent failure:
* 1593163448022208512 from master build for 3831 on 10-17

### Analysis

Although most failures have been because rows still sorted ascending after click to sort descending, this failure is because rows still sorted descending after click to sort ascending again.

This is the only sort test that fails and Risk is the only occurrence of `TableV2` element. Given that the element does not render initial table state, it might have other rendering problems which cause timing problems for the test.

## Checklist
- [x] Edited integration test

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/integration/risk/risk.test.js
+++ b/ui/apps/platform/cypress/integration/risk/risk.test.js
@@ -38,9 +38,9 @@ describe('Risk page', () => {
         });
 
         /*
-         * ROX-13468: assertSortedItems sometimes fails for sort descending.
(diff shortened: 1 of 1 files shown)


gavin-stackrox (approved):
LGTM!

PR #3847, opened 2022-11-17
ROX-12940: diagnostic bundle collection should inherit timeout from request

ivan-degtiarenko:
## Description

Currently, timeout is not considered when collecting diagnostic bundle data on the backend. That can result in the following:
1) Sensor metrics and k8s diagnostics returning non-complete even though request timeout gives enough time
2) Collection being interrupted abruptly which can result in an invalid diagnostic bundle archive

This PR addresses 1st problem. 2nd problem is addressed by warning customers in case timeout is reached, and giving them instructions on how to increase it. The idea is to persuade customers to increase timeout so that we would receive only complete diagnostic bundles when handling support cases.

**Changes in this PR:**
1) Cancellation of sensor metrics and k8s diagnostics goroutines in the backend when request context is cancelled
2) Removal of 20-second hard-coded timeout for sensor metrics and k8s diagnostics goroutines
3) Introducing 1-hour hard timeout for backend requests
4) Introducing warning to the UI and to `roxctl` to instruct the customer on how to increase the timeout in case it is reached

## Checklist
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs]~~(https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

1. Testing `roxctl`

```
degtiarenko@macbook stackrox % roxctl central debug download-diagnostics --password=<password> --endpoint=https://localhost:8000 --timeout=200ms --insecure-skip-tls-verify
INFO:   Downloading diagnostic bundle...
ERROR:  
Timeout is reached while creating diagnostic bundle. Increase 'roxctl' timeout:
'roxctl central debug download-diagnostics --timeout=400s <other parameters'
ERROR:  could not download zip: error when doing http request: Get "https://localhost:8000/api/extensions/diagnostics": context canceled
```

2. Testing UI(changed timeout constant in the UI to test that)

Screenshot 2022-11-21 at 01 47 21

### DIFF
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -555,7 +555,7 @@ func (s *serviceImpl) writeZippedDebugDumpWithTimeout(ctx context.Context, w htt
 		}
 	}
 
-	if time.Since(executionStartTime) > opts.timeout-1*time.Second {
+	if time.Since(executionStartTime) > opts.timeout {
 		writeTimeoutWarning(zipWriter)
 		return
 	}
@@ -570,7 +570,7 @@ func (s *serviceImpl) writeZippedDebugDumpWithTimeout(ctx context.Context, w htt
 		}
 	}
 
-	if time.Since(executionStartTime) > opts.timeout-1*time.Second {
+	if time.Since(executionStartTime) > opts.timeout {
 		writeTimeoutWarning(zipWriter)
 		return
 	}
@@ -733,9 +733,9 @@ func getTimeoutQueryParam(r *http.Request, opts debugDumpOptions) error {
 		if err != nil {
 			return errors.Wrapf(err, "invalid timeout value: %q\n", timeoutStr)
 		}
-		// One second is subtracted to cover time needed to establish HTTP connection between the client
-		// and the server.
-		opts.timeout = time.Duration(timeout-1) * time.Second
+		// Two seconds second is subtracted to cover time needed to establish HTTP connection between the client
+		// and the server, as well as time needed to correctly finish writing ZIP file.
+		opts.timeout = time.Duration(timeout-2) * time.Second
 	} else {
 		opts.timeout = defaultDebugDumpTimeout
 	}
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -61,10 +61,9 @@ const (
 
 	centralClusterPrefix = "_central-cluster"
 
-	metricsPullTimeout     = 20 * time.Second
-	diagnosticsPullTimeout = 20 * time.Second
-	layout                 = "2006-01-02T15:04:05.000Z"
-	logWindow              = 20 * time.Minute
+	defaultDebugDumpTimeout = 60 * time.Second
+	layout                  = "2006-01-02T15:04:05.000Z"
+	logWindow               = 20 * time.Minute
 )
 
 var (
@@ -494,6 +493,7 @@ type debugDumpOptions struct {
 	withCentral       bool
 	clusters          []string
 	since             time.Time
+	timeout           time.Duration
 }
 
 func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseWriter, filename string, opts debugDumpOptions) {
@@ -646,8 +646,14 @@ func (s *serviceImpl) getDebugDump(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	filename := time.Now().Format("stackrox_debug_2006_01_02_15_04_05.zip")
+	err := getTimeoutQueryParam(r, opts)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprint(w, err.Error())
+		return
+	}
 
+	filename := time.Now().Format("stackrox_debug_2006_01_02_15_04_05.zip")
 	s.writeZippedDebugDump(r.Context(), w, filename, opts)
 }
 
@@ -672,8 +678,28 @@ func (s *serviceImpl) getDiagnosticDump(w http.ResponseWriter, r *http.Request) 
 		fmt.Fprint(w, err.Error())
 		return
 	}
+	err = getTimeoutQueryParam(r, opts)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprint(w, err.Error())
+		return
+	}
 
 	s.writeZippedDebugDump(r.Context(), w, filename, opts)
+}
+
+func getTimeoutQueryParam(r *http.Request, opts debugDumpOptions) error {
+	timeoutStr := r.URL.Query().Get("timeout")
+	if timeoutStr != "" {
+		timeout, err := strconv.Atoi(timeoutStr)
+		if err != nil {
+			return errors.Wrapf(err, "invalid timeout value: %q\n", timeoutStr)
+		}
+		opts.timeout = time.Duration(timeout) * time.Second
+	} else {
+		opts.timeout = defaultDebugDumpTimeout
+	}
+	return nil
 }
 
 func getOptionalQueryParams(opts *debugDumpOptions, u *url.URL) error {
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -61,10 +61,11 @@ const (
 
 	centralClusterPrefix = "_central-cluster"
 
-	metricsPullTimeout     = 20 * time.Second
-	diagnosticsPullTimeout = 20 * time.Second
-	layout                 = "2006-01-02T15:04:05.000Z"
-	logWindow              = 20 * time.Minute
+	layout    = "2006-01-02T15:04:05.000Z"
+	logWindow = 20 * time.Minute
+	// This timeout is safety net to prevent request from running forever.
+	// We don't expect it to ever actually be reached. Actual timeout should be set on the client side.
+	debugDumpHardTimeout = 1 * time.Hour
 )
 
 var (
@@ -497,6 +498,8 @@ type debugDumpOptions struct {
 }
 
 func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseWriter, filename string, opts debugDumpOptions) {
+	subCtx, cancel := context.WithTimeout(ctx, debugDumpHardTimeout)
+	defer cancel()
 	w.Header().Set("Content-Type", "application/zip")
 	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", filename))
 
@@ -529,7 +532,7 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 		}
 
 		if opts.withCPUProfile {
-			if err := getCPU(ctx, zipWriter, cpuProfileDuration); err != nil {
+			if err := getCPU(subCtx, zipWriter, cpuProfileDuration); err != nil {
 				log.Error(err)
 			}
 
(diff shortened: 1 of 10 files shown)


central/debug/service/service.go
```diff
@@ -502,6 +508,16 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 
 	zipWriter := zip.NewWriter(w)
 
+	s.writeZippedDebugDumpWithTimeout(ctx, w, opts, zipWriter)
```
dhaus67, line 511: (pending on the comment I added about generally handling timeouts, i.e. setting timeout specifically vs request cancellations):

If you hit a timeout, I appreciate the warning you have added to the diagnostic bundle, but shouldn't we strive to responde with an actual error in this case?
I know we currently fail-open for other errors, but in case of a timeout, the diagnostic bundle essentially becomes quite useless, so responding with an explicit error + instruction to increase timeout personally sounds like a better behavior of the API to me.

central/debug/service/service.go
```diff
@@ -580,9 +606,21 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
...
+	log.Info("Too much time elapsed in diagnostic bundle. Quitting early as to not to hit timeout.")
+
+	w, err := zipWriter.Create(warningFilename)
+	if err != nil {
+		log.Error(errors.Wrapf(err, "unable to create zip file %q", warningFilename))
+	}
+	warningMessage := "Timeout is reached while creating diagnostic bundle. To increase timeout:\n" +
+		"1) If you use `roxctl`, use `--timeout` flag\n" +
+		"2) If you use API, use `timeout` query parameter: `/api/extensions/diagnostics?timeout=60`"
+	_, err = w.Write([]byte(warningMessage))
```
dhaus67, line 621: nit: extract this as a package variable and use multi-line strings to get rid of the \n.

central/debug/service/service.go
```diff
@@ -494,6 +495,11 @@ type debugDumpOptions struct {
 	withCentral       bool
 	clusters          []string
 	since             time.Time
+	// timeout value is sent by client. We operate under assumption that after timeout,
+	// client will drop connection, potentially resulting in incorrect diagnostic bundle.
+	// To avoid that, long-running operations(getting k8s diangostics, pulling sensor metrics) can be dropped
+	// if time is running out.
+	timeout time.Duration
```
dhaus67, line 502: I don't necessarily like setting the timeout now explicitly within a query parameter.

Initially, I would expect the timeout to come from the HTTP request settings, e.g. re-using the `http.Request.Context()` cancellation and not specifically handle timeouts within our server side logic but rather offload it to the client (like we do within roxctl by re-using `request.WithContext(context.WithTimeout(timeout))`.

Meaning, we continue to try to assemble the diagnostic bundle without _actually_ setting a specific, overarching timeout, but rather verifying whether the request's context has been cancelled or not.

You could change the condition of invoking `writeTimeoutWarning` to `ctx.Err() != nil` (however, writing _anything_ to the response writer / zip writer won't be sent to the client, so we'd change to logging the instance in this case and returning the function).

For this to work and us not ending up having blocking reads on channels when receiving e.g. messages from sensor, we would have to create some sort of internal timeout based on the longest amount of time we would be waiting for a sensor message (this would be something we'd have to find a median).

This would make the API feel more native in my opinion and we could decide to print the warning message you have already created within e.g. `roxctl` when the API call returns a `context.DeadlineExceeded` error (which is what you will get), and decide to have no explicit timeout being set within the UI (but relying on the internal timeouts to not end up in e.g. a "infinite waiting on messages from a channel" situation).

(just for posterity, the following test code highlights what happens when a HTTP request's timeout is hit on both client and server side):
```go
	server := httptest.NewServer(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
		time.Sleep(2 * time.Second)
		assert.Error(t, request.Context().Err())
		assert.Error(t, request.Context().Err(), context.Canceled)
		return
	}))
	defer server.Close()

	ctx, _ := context.WithTimeout(context.Background(), 1*time.Second)
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, server.URL, nil)
	assert.NoError(t, err)

	client := http.Client{}
	resp, err := client.Do(req)
	assert.Error(t, err)
	assert.ErrorIs(t, err, context.DeadlineExceeded)
	assert.Nil(t, resp)
        time.Sleep(2 * time.Second)
```

janisz (commented):
Could you add tests?

ivan-degtiarenko:
@janisz I changed the logic within the PR - do you think testing still makes sense? I'm not sure there are any assertions we can test

roxctl/central/debug/download_diagnostics.go
```diff
@@ -58,3 +67,10 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
 
 	return c
 }
+
+func isTimeoutError(err error) bool {
```
janisz, line 71: https://pkg.go.dev/os#IsTimeout

roxctl/central/debug/download_diagnostics.go
```diff
@@ -41,14 +44,20 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
...
 				Method:     http.MethodGet,
 				Timeout:    flags.Timeout(c),
 				BundleType: "diagnostic",
 				ExpandZip:  false,
 				OutputDir:  outputDir,
 			}, cliEnvironment.Logger())
+			if isTimeoutError(err) {
+				cliEnvironment.Logger().ErrfLn(`
+Timeout is reached while creating diagnostic bundle. Increase 'roxctl' timeout:
+'roxctl central debug download-diagnostics --timeout=400s <other parameters'`)
```
janisz, line 58: What if user already used 400s? Better use flag value times two or more
```suggestion
'roxctl central debug download-diagnostics --timeout=2*timeout from flag <other parameters'`)
```

central/debug/service/service.go
```diff
@@ -497,6 +498,8 @@ type debugDumpOptions struct {
 }
 
 func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseWriter, filename string, opts debugDumpOptions) {
+	subCtx, cancel := context.WithTimeout(ctx, debugDumpHardTimeout)
```
janisz, line 501: Are you sure there is no timeout on context passed here?

central/debug/service/service.go
```diff
@@ -61,10 +61,11 @@ const (
...
 
-	metricsPullTimeout     = 20 * time.Second
-	diagnosticsPullTimeout = 20 * time.Second
-	layout                 = "2006-01-02T15:04:05.000Z"
-	logWindow              = 20 * time.Minute
+	layout    = "2006-01-02T15:04:05.000Z"
+	logWindow = 20 * time.Minute
+	// This timeout is safety net to prevent request from running forever.
+	// We don't expect it to ever actually be reached. Actual timeout should be set on the client side.
+	debugDumpHardTimeout = 1 * time.Hour
```
janisz, line 68: When it could happen?

central/debug/service/service.go
dhaus67: This will be called by [getDiagnosticDump](https://github.com/stackrox/stackrox/blob/master/central/debug/service/service.go#L676) / [getDebugDump](https://github.com/stackrox/stackrox/blob/e14bc41579f4b1ab5e47e62dad1182edc9400bd8/central/debug/service/service.go#L651) and we will pass the HTTP request's context there - this context won't have a deadline set.

dhaus67, line 501: nit: I'd prefer something more descriptive such as `debugDumpCtx` instead of `subCtx`.

roxctl/central/debug/download_diagnostics.go
```diff
@@ -41,14 +44,20 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
...
+			err := zipdownload.GetZip(zipdownload.GetZipOptions{
 				Path:       path,
 				Method:     http.MethodGet,
 				Timeout:    flags.Timeout(c),
 				BundleType: "diagnostic",
 				ExpandZip:  false,
 				OutputDir:  outputDir,
 			}, cliEnvironment.Logger())
+			if isTimeoutError(err) {
+				cliEnvironment.Logger().ErrfLn(`
```
dhaus67, line 56: Skip the first newline (i.e. moving the ` to the next line), otherwise the error would look a bit weird IMO:
```
ERROR:
Timeout is reached while creating...
```

roxctl/central/debug/download_diagnostics.go
```diff
@@ -41,14 +44,20 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
...
+Timeout is reached while creating diagnostic bundle. Increase 'roxctl' timeout:
```
dhaus67, line 57: nit:
```suggestion
Timeout has been reached while creating diagnostic bundle. Increase 'roxctl' timeout by running the following:
```

roxctl/central/debug/download_diagnostics.go
```diff
@@ -58,3 +67,10 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
 
 	return c
 }
+
+func isTimeoutError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "net/http: request canceled") || errors.Is(err, context.DeadlineExceeded)
```
dhaus67, line 75: Instead of the string checking, you can do the following (which would mean skipping the string comparison):
```
var urlError *url.Error
return errors.As(err, &urlError) && urlError.Timeout()
```

roxctl/central/debug/download_diagnostics.go
```diff
@@ -58,3 +67,10 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
 
 	return c
 }
+
+func isTimeoutError(err error) bool {
+	if err == nil {
+		return false
+	}
```
dhaus67, line 74: If you stick with using `errors.Is/errors.As` and get rid of the string comparison, you can skip this explicit nil check here (both `errors.Is/As` will return false if the given error is nil).

ui/apps/platform/src/Containers/SystemHealth/Components/DiagnosticBundleDialogBox.tsx
```diff
@@ -117,9 +120,30 @@ const DiagnosticBundleDialogBox = (): ReactElement => {
...
+                        title="Downloading diagnostic bundle failed."
+                        variant={AlertVariant.danger}
+                        isInline
+                    >
+                        <p>
+                            <b>{error.message}</b>
+                            <br />
+                            If timeout is exceeded, repeat request with increased timeout in
+                            `roxctl` tool: `roxctl central debug download-diagnostics
+                            --timeout=400s`
```
dhaus67, line 139:
```suggestion
                            If timeout is exceeded, repeat request with increased timeout using the
                            `roxctl` command line tool: `roxctl central debug download-diagnostics
                            --timeout=400s`
```

roxctl/central/debug/download_diagnostics.go
ivan-degtiarenko: But what if user set some small timeout on the opposite, like `1s`? Then they can go through 7-8 attempts before they get right result.

Maybe we should provide some more detailed guidance on timeout in the warning. For example:
```
Timeout value used was 50 seconds, while default timeout value is 300 seconds. If your timeout value is less than a default value, use default value. If your timeout value is more or equal to default value, increase timeout value twice in size.
```
@dhaus67 thoughts?

roxctl/central/debug/download_diagnostics.go
ivan-degtiarenko: Does not work:
```
INFO:   Downloading diagnostic bundle...
ERROR:  error writing to ZIP file: net/http: request canceled (Client.Timeout or context cancellation while reading body)

```

roxctl/central/debug/download_diagnostics.go
ivan-degtiarenko: Does not work

```
INFO:   Downloading diagnostic bundle...
ERROR:  error writing to ZIP file: net/http: request canceled (Client.Timeout or context cancellation while reading body)

```

roxctl/central/debug/download_diagnostics.go
```diff
@@ -41,14 +44,20 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
...
-			return zipdownload.GetZip(zipdownload.GetZipOptions{
+			err := zipdownload.GetZip(zipdownload.GetZipOptions{
 				Path:       path,
 				Method:     http.MethodGet,
 				Timeout:    flags.Timeout(c),
 				BundleType: "diagnostic",
 				ExpandZip:  false,
 				OutputDir:  outputDir,
 			}, cliEnvironment.Logger())
+			if isTimeoutError(err) {
```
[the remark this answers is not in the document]
ivan-degtiarenko: Do you want to write a unit test for this function or mock HTTP client?

I did manual testing by using `1ms`m `100ms` and `1s` timeout value

central/debug/service/service.go
ivan-degtiarenko: This is safety net against very hypothetical situation:
1. Customers use direct API call without timeout
2. Somehow pulling sensor/kubernetes data is stuck

This is basically to prevent potential resource leaking. We don't expect to run into this timeout any time soon

roxctl/central/debug/download_diagnostics.go
ivan-degtiarenko: see below

roxctl/central/debug/download_diagnostics.go
dhaus67: Ah, I initially looked at non-http2 code.

The error returned will be `x/net/http2/httpError`, which satisfies the `net.Error` interface, so this should now do:
```
var netErr net.Error
return errors.As(err, &netErr) && netErr.Timeout()
```

roxctl/central/debug/download_diagnostics.go
dhaus67: I think it doesn't hurt to be a bit more descriptive here and mention the default timeout value.

roxctl/central/debug/download_diagnostics.go
ivan-degtiarenko: @dhaus67 nice, it works :)

ui/apps/platform/src/Containers/SystemHealth/Components/DiagnosticBundleDialogBox.tsx
```diff
@@ -117,9 +120,30 @@ const DiagnosticBundleDialogBox = (): ReactElement => {
...
+                setAlertDownload(
+                    <Alert
+                        title="Downloading diagnostic bundle failed."
+                        variant={AlertVariant.danger}
+                        isInline
+                    >
+                        <p>
+                            <b>{error.message}</b>
+                            <br />
+                            If timeout is exceeded, repeat request with increased timeout using the
```
pedrottimark, line 137: My first impression of **repeat request** was:
1. use roxctl to increase the timeout
2. repeat the request via user interface

My second impression is if timeout is exceeded when customer attempts to download via user interface, then use roxctl command line **instead** of user interface.

Can we rewrite the sentence to remove the ambiguity?

roxctl/central/debug/download_diagnostics.go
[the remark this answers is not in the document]
janisz: I think we should mock http

roxctl/central/debug/download_diagnostics.go
```diff
@@ -41,14 +44,23 @@ func downloadDiagnosticsCommand(cliEnvironment environment.Environment) *cobra.C
...
+				cliEnvironment.Logger().ErrfLn(`Timeout has been reached while creating diagnostic bundle. 
+Timeout value used was %s, while default timeout value is %s. 
+If your timeout value is less than a default value, use default value.
```
dhaus67, line 58:
```suggestion
If your timeout value is less than the default value, use the default value. 
```

dhaus67 (approved):
LGTM.

Regarding the testing parts, I'd say it would be beneficial to create a follow-up refactoring of `zipdownload.GetZip` to make it easier testable (currently, this internally creates the HTTP client + reads values from flags, which makes testing with e.g. httptest quiet hard), and once that's added add some tests here that would simulate a timeout.

pedrottimark (approved):
Approve changes in ui folder. Thank you for the improvements.

janisz (approved):

ivan-degtiarenko:
Manually tested. Unit test is introduced in a separate branch. Merging the PR

PR #3851, opened 2022-11-18
ROX-13357: store refactor

fredrb:
## Description

This is the third attempt on refactoring the in memory stores to accommodate the follow-up changes required by re-sync. 

1. The first approach, was to move store implementation to `sensor/common/store`, which meant also pulling up all wrappers and exposing them publicly. (PR #3663)
2. The second approach was about splitting the wrapper, parsing and store into different package (`sensor/common/store`, `sensor/kubernetes/convert` and `sensor/kubernetes/store`). Which made the code more complex and prone to cyclic dependencies. (draft, no PR)
3. This approach is about slightly changing the exposed APIs to only expose `storage.*` type objects and no wrappers. This way, there's no need to make wrappers and parsing logic public. (This PR)

The good thing about this approach is that it doesn't force a big design change in the stores nor the wrappers.

Additionally, rather than adding singletons for services and RBAC stores, a new entity store provider entity was created. This allows stores to be injected as dependencies to sensor components rather than access globally using a singleton. Ideally, other global stores (e.g. Deployments and Network Policies) should also move to the store provider. [This closed ticket](https://issues.redhat.com/browse/ROX-11027) was meant to solve this, but we found a workaround in the tests and the root of the problem was never fixed.

Another added bonus is that the deployment wrap functions no longer depends on service wraps. They take the storage type parameters for port exposure instead.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/sensor/common/store/types.go
+++ b/sensor/common/store/types.go
@@ -1,12 +1,17 @@
 package store
 
-import "github.com/stackrox/rox/generated/storage"
+import (
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/sensor/common/rbac"
+	"github.com/stackrox/rox/sensor/common/service"
+)
 
 // DeploymentStore provides functionality to fetch all deployments from underlying store.
 //go:generate mockgen-wrapper
 type DeploymentStore interface {
 	GetAll() []*storage.Deployment
 	Get(id string) *storage.Deployment
+	BuildDeploymentWithDependencies(id string, dependencies Dependencies) (*storage.Deployment, error)
 }
 
 // PodStore provides functionality to fetch all pods from underlying store.
@@ -34,4 +39,16 @@ type ServiceAccountStore interface {
 	Add(sa *storage.ServiceAccount)
 	Remove(sa *storage.ServiceAccount)
 	GetImagePullSecrets(namespace, name string) []string
+}
+
+// ServiceStore provides functionality to find port exposure infos from in-memory services
+//go:generate mockgen-wrapper
+type ServiceStore interface {
+	GetExposureInfos(namespace string, labels map[string]string) []map[service.PortRef][]*storage.PortConfig_ExposureInfo
+}
+
+// RBACStore provides functionality to find permission level from in-memory RBACs
+//go:generate mockgen-wrapper
+type RBACStore interface {
+	GetPermissionLevelForDeployment(deployment rbac.NamespacedServiceAccount) storage.PermissionLevel
 }
--- a/sensor/kubernetes/listener/resources/convert.go
+++ b/sensor/kubernetes/listener/resources/convert.go
@@ -20,6 +20,7 @@ import (
 	"github.com/stackrox/rox/pkg/utils"
 	"github.com/stackrox/rox/pkg/uuid"
 	"github.com/stackrox/rox/sensor/common/registry"
+	"github.com/stackrox/rox/sensor/common/service"
 	"github.com/stackrox/rox/sensor/kubernetes/listener/resources/references"
 	"github.com/stackrox/rox/sensor/kubernetes/orchestratornamespaces"
 	"github.com/stackrox/rox/sensor/kubernetes/selector"
@@ -61,7 +62,7 @@ type deploymentWrap struct {
 	*storage.Deployment
 	registryOverride string
 	original         interface{}
-	portConfigs      map[portRef]*storage.PortConfig
+	portConfigs      map[service.PortRef]*storage.PortConfig
 	pods             []*v1.Pod
 	// registryStore is the image registry store to use when determining if an image is cluster-local.
 	registryStore *registry.Store
@@ -391,12 +392,12 @@ func (w *deploymentWrap) populatePorts() {
 	w.mutex.Lock()
 	defer w.mutex.Unlock()
 
-	w.portConfigs = make(map[portRef]*storage.PortConfig)
+	w.portConfigs = make(map[service.PortRef]*storage.PortConfig)
 	for _, c := range w.GetContainers() {
 		for _, p := range c.GetPorts() {
-			w.portConfigs[portRef{Port: intstr.FromInt(int(p.ContainerPort)), Protocol: v1.Protocol(p.Protocol)}] = p
+			w.portConfigs[service.PortRef{Port: intstr.FromInt(int(p.ContainerPort)), Protocol: v1.Protocol(p.Protocol)}] = p
 			if p.Name != "" {
-				w.portConfigs[portRef{Port: intstr.FromString(p.Name), Protocol: v1.Protocol(p.Protocol)}] = p
+				w.portConfigs[service.PortRef{Port: intstr.FromString(p.Name), Protocol: v1.Protocol(p.Protocol)}] = p
 			}
 		}
 	}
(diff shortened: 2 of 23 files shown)


sensor/kubernetes/listener/resources/convert.go
```diff
@@ -451,17 +452,17 @@ func (w *deploymentWrap) resetPortExposureNoLock() {
...
-	defer w.mutex.Unlock()
-
-	w.resetPortExposureNoLock()
-
-	svcs := store.getMatchingServicesWithRoutes(w.Namespace, w.PodLabels)
-	for _, svc := range svcs {
-		w.updatePortExposureUncheckedNoLock(svc)
-	}
-}
+//func (w *deploymentWrap) updatePortExposureFromStore(store *serviceStore) {
```
fredrb, line 455: Remove

sensor/kubernetes/listener/resources/convert.go
```diff
@@ -451,17 +452,17 @@ func (w *deploymentWrap) resetPortExposureNoLock() {
...
+//
+//	w.resetPortExposureNoLock()
+//
+//	svcs := store.getMatchingServicesWithRoutes(w.Namespace, w.PodLabels)
+//	for _, svc := range svcs {
+//		w.updatePortExposureUncheckedNoLock(svc)
+//	}
+//}
 
 func (w *deploymentWrap) updatePortExposureFromServices(svcs ...serviceWithRoutes) {
```
fredrb, line 467: Maybe this method isn't needed anymore either. Any parent call with services can do the same thing with the exposure object.

sensor/common/store/types.go
```diff
@@ -35,3 +40,15 @@ type ServiceAccountStore interface {
 	Remove(sa *storage.ServiceAccount)
 	GetImagePullSecrets(namespace, name string) []string
 }
+
+// ServiceStore provides functionality to find port exposure infos from in-memory services
+//go:generate mockgen-wrapper
```
fredrb, line 45: Commit generated code

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -125,14 +126,32 @@ func (ds *DeploymentStore) GetAll() []*storage.Deployment {
...
+	return wrap
+}
+
+// Get returns deployment for supplied id.
+func (ds *DeploymentStore) Get(id string) *storage.Deployment {
+	wrap := ds.getWrap(id)
 	return wrap.GetDeployment()
 }
+
+// BuildDeploymentWithDependencies creates storage.Deployment object using external object dependencies
```
fredrb, line 146:
```suggestion
// BuildDeploymentWithDependencies creates storage.Deployment object using external object dependencies.
```

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -125,14 +126,32 @@ func (ds *DeploymentStore) GetAll() []*storage.Deployment {
...
-func (ds *DeploymentStore) Get(id string) *storage.Deployment {
+func (ds *DeploymentStore) getWrap(id string) *deploymentWrap {
 	ds.lock.RLock()
 	defer ds.lock.RUnlock()
 
 	wrap := ds.deployments[id]
 	if wrap == nil {
 		return nil
 	}
+	return wrap
```
vikin91, line 137: Do we need that `if`? If wrap is `nil` then the returned value will be also `nil`. Or do we care here about typed nils?

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -125,14 +126,32 @@ func (ds *DeploymentStore) GetAll() []*storage.Deployment {
...
+// Get returns deployment for supplied id.
+func (ds *DeploymentStore) Get(id string) *storage.Deployment {
+	wrap := ds.getWrap(id)
 	return wrap.GetDeployment()
 }
+
+// BuildDeploymentWithDependencies creates storage.Deployment object using external object dependencies.
+func (ds *DeploymentStore) BuildDeploymentWithDependencies(id string, dependencies store.Dependencies) (*storage.Deployment, error) {
+	wrap := ds.getWrap(id)
+	clonedWrap := wrap.Clone()
```
vikin91, line 149: `wrap` can be nil -> this can panic

sensor/kubernetes/listener/resources/store_provider.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/stackrox/rox/sensor/common/store"
+	"github.com/stackrox/rox/sensor/kubernetes/listener/resources/rbac"
+)
+
+// InMemoryStoreProvider holds all stores used in sensor and exposes a public interface for each that can be used outside of the listeners.
+type InMemoryStoreProvider struct {
+	serviceStore *serviceStore
+	rbacStore    rbac.Store
```
vikin91, line 11: Why not 

```suggestion
	rbacStore    *rbac.Store
```

to avoid copying when calling `RBAC()`?

sensor/kubernetes/listener/resources/dispatcher.go
```diff
@@ -65,16 +65,17 @@ func NewDispatcherRegistry(
 	namespaces *orchestratornamespaces.OrchestratorNamespaces,
 	credentialsManager awscredentials.RegistryCredentialsManager,
 	traceWriter io.Writer,
+	storeProvider *InMemoryStoreProvider,
 ) DispatcherRegistry {
-	serviceStore := newServiceStore()
+	serviceStore := storeProvider.serviceStore
+	rbacUpdater := storeProvider.rbacStore
```
vikin91, line 71: Why not using the `storeProvider.Services()` and `RBAC()` respectively?

sensor/kubernetes/sensor/sensor.go
```diff
@@ -101,9 +101,12 @@ func CreateSensor(cfg *CreateOptions) (*sensor.Sensor, error) {
 		return nil, errors.Wrap(err, "creating enforcer")
 	}
 
+	// TODO: Move other singleton stores into the store provider
```
vikin91, line 104: You got a ticket ref for this? Otherwise this todo would be easy to forget

sensor/kubernetes/listener/resources/dispatcher.go
```diff
@@ -65,16 +65,17 @@ func NewDispatcherRegistry(
 	namespaces *orchestratornamespaces.OrchestratorNamespaces,
 	credentialsManager awscredentials.RegistryCredentialsManager,
 	traceWriter io.Writer,
+	storeProvider *InMemoryStoreProvider,
```
vikin91, line 68: This should be maybe an interface? `InMemoryStoreProvider` is an implementation.
No need to do it now, but I think this is what we want to have long term.

vikin91 (changes requested):
Looks really nice!

There is one potential panic and one copy operation that could be avoided, hence requesting changes.

sensor/kubernetes/listener/resources/store_provider.go
fredrb: Because `rbac.Store` is an interface

sensor/kubernetes/listener/resources/dispatcher.go
fredrb: Because we have a public and a private version API for the stores. The listeners need access to the private methods (E.g.: upserting services). If we pass the public interface (as in defined `common/store/types.go`) then listeners won't be able to operate the stores properly. 

Other components would only use the public interface (i.e. methods that don't need to expose internal wrappers).

I wanted to introduce a unified interface that can be used both internally and externally. But this requires a _much_ bigger change, because we would have to extract all wrappers and parsing logic from the listeners component.

sensor/kubernetes/listener/resources/dispatcher.go
fredrb: The downside of making this an interface is that it prevents us for having the _private_ and _public_ split of stores. This only works because the provider exposes the public interface of stores using methods, but allows the listeners component to access the concrete store implementation, and hence their private methods.

sensor/kubernetes/listener/resources/deployment_store.go
fredrb: Good catch. I'll fix this

fredrb: Done

sensor/kubernetes/listener/resources/dispatcher.go
vikin91: oh, I see. But if that is the case then the benefits from having the storeProvider in the current shape are much smaller than anticipated. It means, that we cannot have another storeProvider (e.g. a mock) unless we implement exactly the same set methods for the particular stores (both public and private).

Having said this is not great, I would be for adding the interfaces in one of the next PRs.

vikin91 (approved):
Thanks for fixing the blockers!
While the current shape of the store provider is not ideal, this PR is a step in the right direction.

lvalerom (approved):
LGTM. Just a couple of questions/suggestions. 

* [nit] I think we can pass `store.ServiceStore` to the `deploymentHandler` since I think it doesn't need to know about the private implementation of the `serviceStore`

* We don't need to address it in this PR but I've noticed that the `nodeDispatcher` never uses the `serviceStore`. Do you think we can remove it?

sensor/kubernetes/sensor/sensor.go
lvalerom: +1

fredrb:
@lvalerom Thanks. Those were great observations! Fixed both.

PR #3852, opened 2022-11-18
ROX-12316: Remove deprecated ClusterCVE resource

rhybrillou:
## Description

The `ClusterCVE` resource (for permission sets) was deprecated in 3.72.
Following the branching of 3.73, the associated code can be removed for 3.74.

## Checklist

## Testing Performed

CI run

Manual testing:
- Deploy central.
- Go to the `Access Control` section.
- Open the `Permission sets` tab.
- Open one permission set (e.g. Admin), and check that `ClusterCVE` is not in the Resource list anymore.
- Back to the `Permission sets` tab, click the `Create permission set` button, check that the form does not expose `ClusterCVE` in the list of resources (only occurrences of the name should be in the warning banner, and as comment for the `Cluster` resource).

Migration testing:
- Deploy an older central version (e.g. 3.72)
- Go to the `Access Control` section
- Perform the steps listed in manual testing, but checking for the presence of the `ClusterCVE` resource rather than its absence.
- Create a permission set
- With the CLI, ensure that the `ClusterCVE` permission is present in the created permission set.
- Upgrade Central to a newer version (e.g. the one built with this pull request), check for migration steps.
- Repeat the UI manual test step, checking this time for `ClusterCVE` absence.
- Use the CLI to ensure that the `ClusterCVE` permission is not present in the existing permission sets.


janisz (approved):

pkg/migrations/internal/seq_num.go
```diff
@@ -6,5 +6,5 @@ var (
 	// It is a shared constant between central and the migrator binary.
 	CurrentDBVersionSeqNum = 112
 	// PostgresDBVersionPlus is the current DB version number with Postgres DB data migration.
-	PostgresDBVersionPlus = 56
+	PostgresDBVersionPlus = 58
```
rukletsov, line 9: Why not 57? Is it because it was not bumped last time?

pkg/migrations/internal/seq_num.go
```diff
@@ -6,5 +6,5 @@ var (
 	// It is a shared constant between central and the migrator binary.
 	CurrentDBVersionSeqNum = 112
```
rukletsov, line 7: Is my understanding correct that you don't bump this because you only write postgres migration?

migrator/migrations/n_57_to_n_58_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const (
+	batchSize = 500
+)
+
+var (
+	migration = types.Migration{
+		StartingSeqNum: pkgMigrations.CurrentDBVersionSeqNumWithoutPostgres() + 57,
+		VersionAfter:   &storage.Version{SeqNum: 58},
```
rukletsov, line 22: Shouldn't this be `pkgMigrations.CurrentDBVersionSeqNumWithoutPostgres() + 58`?

migrator/migrations/n_57_to_n_58_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+	}
+	clusterCVEResourceName = "ClusterCVE"
+)
+
+func cleanupPermissionSets(db *pgxpool.Pool) error {
+	ctx := context.Background()
+	store := postgres.New(db)
+	permissionSetsToInsert := make([]*storage.PermissionSet, 0, batchSize)
+	err := store.Walk(ctx, func(obj *storage.PermissionSet) error {
```
rukletsov, line 38: It took me some time to realize how this magically works without specifying the table. Can we rename the package with `postgres_plugin.go` into something more specific so that it's clear that `.New()` is a local function that produces a store specializing on permission sets? Maybe `postgresPlugin.New()`?

rukletsov (changes requested):
Looks like this is the first migration from postgres to postgres 🎊. People will be looking at it and its successors, copy-pasting the patterns and the code. A couple of thoughts:
* I guess the "postgres_plugin" approach is a new pattern everyone will be using. I guess you haven't written it yourself but copy-pasted from _somewhere_ — please leave a clear not about this.
* I see you're copying this pattern yourself but I also see alternative patterns, like `postgrespermissionsets` folder names. Let's agree on one pattern (folder structure plus package names) here and hope others will follow.

pkg/migrations/internal/seq_num.go
c-du: This is the 57th migrations in the n_ series, so it should be 57. #2630 was in before I hooked up Postgres migrations.

rukletsov: Then the files should also be named 56_to_57.

c-du: No, it should be named 57_to_58 and PostgresDBVersionPlus is 57. We have test case to cover if the number is wrong.

c-du: I understand why you think it should be 56_to_57. You may have noticed n_ series is 1-based not 0-based. I should have changed it.

migrator/README.md
```diff
@@ -1,23 +1,98 @@
...
-    1. Ensure that the `summary_of_migration` follows the naming convention of previous migrations, i.e., postfix `_policy` if it modifies policies
-3. Create at least two files: `migration.go` and `migration_test.go`. These files should belong to package `m{currentDBVersion}tom{currentDBVersion+1}`
-4. To better understand how to write these two files, look at existing examples: [#1](https://github.com/stackrox/rox/pull/8609) [#2](https://github.com/stackrox/rox/pull/7581) [#3](https://github.com/stackrox/rox/pull/7921) in `migrations` directory. Avoid depending on code that might change in the future as **migration should produce consistent results**.
-5. Add to `migrator/runner/all.go` line
+1. Determine which database group will be targeted. The options should be `BoltDB or RocksDB`,
+`Postgres and Gorm` or a migration from the former to the latter.
+
+    Note: the rule of thumb to determine the target group is based on the first release where the migration should be
+    applied.
+    1. Before 3.73, the only target group should be `BoltDB or RocksDB`.
```
c-du, line 25: We are already in 74. We should not recommend to add new migration on top of legacy databases.

migrator/migrations/n_57_to_n_58_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stackrox/rox/migrator/types"
+)
+
+const (
+	batchSize = 500
+)
+
+var (
+	migration = types.Migration{
+		StartingSeqNum: 168,
```
dashrews78, line 20: Should we use the constants like all the others?

migrator/README.md
```diff
@@ -1,23 +1,98 @@
 # StackRox Database migration
 
+## Purpose of database migrations
```
dashrews78, line 3: Great write up.  this should prove very useful going forward.

md2119: Couldn't agree more!

migrator/migrations/n_57_to_n_58_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+			return nil
+		},
+	}
+	clusterCVEResourceName = "ClusterCVE"
+)
+
+func cleanupPermissionSets(db *pgxpool.Pool) error {
+	ctx := context.Background()
+	permissionSetStore := permissionsetpostgresstore.New(db)
```
md2119, line 35: nit: `permissionSetPostgresStore.New(db)`

migrator/migrations/n_57_to_n_58_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,62 @@
...
+	permissionSetsToInsert := make([]*storage.PermissionSet, 0, batchSize)
+	err := permissionSetStore.Walk(ctx, func(obj *storage.PermissionSet) error {
+		if _, ok := obj.GetResourceToAccess()[clusterCVEResourceName]; ok {
+			delete(obj.ResourceToAccess, clusterCVEResourceName)
+			permissionSetsToInsert = append(permissionSetsToInsert, obj)
```
md2119, line 40: The changelog states `...the permission 'Cluster' replaces the deprecated permission 'ClusterCVE'`. However, you are only deleting the `ClusterCVE` resource. There is an assumption that the permission set contains resource `Cluster`.

I think either you need to changelog the changelog or make sure Cluster resource exists in the permission set.

migrator/migrations/n_01_to_n_02_postgres_clusters/migration.go
```diff
@@ -19,10 +19,14 @@ import (
 	"gorm.io/gorm"
 )
 
+const (
+	startingSeqNum = pkgMigrations.BasePostgresDBVersionSeqNum() + 1, // 112
+)
+
```
keyallis, line 25: This appears to be a change to the autogenerated migration code, is there a functional reason why this change is being done? Or is it to align the migrations with the form we would like to maintain going forward

migrator/runner/all.go
```diff
@@ -14,6 +14,7 @@ import (
 	_ "github.com/stackrox/rox/migrator/migrations/m_109_to_m_110_networkpolicy_guidance_2"
 	_ "github.com/stackrox/rox/migrator/migrations/m_110_to_m_111_replace_deprecated_resources"
 	_ "github.com/stackrox/rox/migrator/migrations/m_111_to_m_112_groups_invalid_values"
+	_ "github.com/stackrox/rox/migrator/migrations/m_168_to_m_169_postgres_remove_clustercve_permission"
```
keyallis, line 17: Do we want to pull this import into its own section of postgres -> postgres migrations in a way similar to the rocks -> postgres migrations are in their own section? We could start a new section at the top, which would also give us an opportunity to leave a descriptive message describing why there's a number skip and that the new particular section is postgres -> postgres only

md2119 (approved):
Fantastic job with the documentation!

migrator/migrations/m_168_to_m_169_postgres_remove_clustercve_permission/migration.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/migrator/migrations"
+	permissionSetPostgresStore "github.com/stackrox/rox/migrator/migrations/m_168_to_m_169_postgres_remove_clustercve_permission/permissionsetpostgresstore"
+	"github.com/stackrox/rox/migrator/types"
+)
+
+const (
```
dashrews78, line 14: nit:  can we combine the multiple const and var groups, respectively?

dashrews78 (approved):

migrator/migrations/m_168_to_m_169_postgres_remove_clustercve_permission/permissionsetpostgresstore/postgres_plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	pkgSchema "github.com/stackrox/rox/pkg/postgres/schema"
+	"github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/search/postgres"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+// This file is a partial copy of central/role/store/permissionset/postgres/store.go
+// in the state it had when the migration was written.
+// Only the relevant functions (Walk, DeleteMany and UpsertMany) are kept.
+// The kept functions are stripped from the scoped access control checks.
```
rukletsov, line 24: Thank you, @rhybrillou

migrator/migrations/m_168_to_m_169_postgres_remove_clustercve_permission/permissionsetpostgresstore/postgres_plugin.go
```diff
@@ -0,0 +1,283 @@
...
+	var err error
+
+	// This is a copy so first we must delete the rows and re-add them
+	// Which is essentially the desired behaviour of an upsert.
+	var deletes []string
+
+	copyCols := []string{
+
+		"id",
+
```
rukletsov, line 97: Not sure how these empty lines ended up in the code but they are present in `role/store/permissionset/postgres/store.go` as well, so NNTR.

migrator/README.md
```diff
@@ -1,23 +1,99 @@
...
+    ```go
+   import (
+       "github.com/stackrox/rox/migrator/migrations"
+       "github.com/stackrox/rox/migrator/types"
+       // If needed for the sequence number management.
+       pkgMigrations "github.com/stackrox/rox/pkg/migrations"
+   )
+   
+   var (
+       startSeqNum = {curentDBVersion}
```
rukletsov, line 65: I guess you call it `startingSeqNum` in this PR so let's rename here for clarity. I also suggest to add a comment here saying how we recommend to declare the value, specifically via `pkgMigrations.BasePostgresDBVersionSeqNum() + N` and say how to figure out `N`. Maybe also clarify this relationship above around L44.

rukletsov (approved):
Thanks, Yann, I only have minor comments. Feel free to either address&merge or merge&follow-up.

/lgtm
/woof

PR #3853, opened 2022-11-18
ROX-13584: Fix compatibility tests

fredrb:
## Description

Compatibility tests had a few issues:
1. Tests weren't really being run with older versions of sensor
2. The most recent helm charts were being used and only the image was (supposedly) being changed. This could lead to issues if the resource definition changed from one version to another
3. Sensor version was being swapped, but collector remained the same. 

This PR proposes a different way to deploy older sensors: by using open source helm charts. After central is deployed, we can fetch the init bundle and deploy a public version of the secured cluster. 

## Checklist
- [ ] Update script that automatically fetches sensor versions to get chart versions instead (fyi @jschnath)

## Testing Performed

- CI


tests/e2e/lib.sh
```diff
@@ -31,6 +61,7 @@ deploy_stackrox() {
     sensor_wait
 
     touch "${STATE_DEPLOYED}"
+
```
fredrb, line 64: Remove

tests/e2e/lib.sh
```diff
@@ -78,6 +109,25 @@ deploy_central() {
...
+    info "Deploying secured cluster (v$1) from Helm Charts (init bundle $2)"
+
+    helm repo update
+
+    helm install -n stackrox stackrox-secured-cluster-services \
+        stackrox-oss/stackrox-secured-cluster-services \
+        -f "$2" \
+        --set clusterName="remote" \
+        --version "$1"
```
fredrb, line 128: Meaningful variable names

scripts/ci/jobs/gke_version_compatibility_tests.py
```diff
@@ -14,10 +14,11 @@
 # don't use postgres
 os.environ["ROX_POSTGRES_DATASTORE"] = "false"
 
-versions=get_latest_release_versions(4)
+# versions=get_latest_release_versions(4)
+chart_versions=["72.1.0", "71.2.0", "70.0.0"]
```
fredrb, line 18: TODO: Probably need to update `get_latest_release_versions` to use `helm search` instead.

qa-tests-backend/scripts/run-compatibility.sh
```diff
@@ -31,8 +31,10 @@ compatibility_test() {
         remove_existing_stackrox_resources
         setup_default_TLS_certs
 
-        deploy_stackrox
+        # deploy_stackrox
```
fredrb, line 34: remove

scripts/ci/jobs/gke_version_compatibility_tests.py
```diff
@@ -14,10 +14,11 @@
...
 
 gkecluster=GKECluster("qa-e2e-test")
```
gavin-stackrox, line 20: while you are here can you also:
```suggestion
gkecluster=GKECluster("compat-test")
```
(I picked something short because long names can have issues)

scripts/ci/jobs/gke_version_compatibility_tests.py
gavin-stackrox: Are you going to do this in this PR? Otherwise this'll be a chore to periodically update.

tests/e2e/lib.sh
```diff
@@ -17,11 +17,41 @@ deploy_stackrox() {
...
+    # Bounce collectors to avoid restarts on initial module pull
+    kubectl -n stackrox delete pod -l app=collector --grace-period=0
+
+    sensor_wait
+
+    touch "${STATE_DEPLOYED}"
+}
+
+deploy_stackrox_with_custom_sensor() {
+    if [ -z $1 ]; then
```
gavin-stackrox, line 33: This'll blow up due to set -u and it being an unbound variable. This is used elsewhere in the file:
```suggestion
    if [[ "$#" -ne 1 ]]; then
```

qa-tests-backend/scripts/run-compatibility.sh
```diff
@@ -13,7 +13,7 @@ source "$ROOT/qa-tests-backend/scripts/lib.sh"
 set -euo pipefail
 
 compatibility_test() {
```
gavin-stackrox, line 15: whatabout doing this all in bash? in that way it'll reuse the same wrapping cluster. Which will save a bit of time.

scripts/ci/jobs/gke_version_compatibility_tests.py
gavin-stackrox: `helm search repo stackrox-oss -l -o json | jq -r '.[] | select( .name=="stackrox-oss/stackrox-secured-cluster-services" ) | .app_version'` got me:
```
3.72.1
3.72.0
3.71.2
3.71.0
3.70.0
```

qa-tests-backend/scripts/run-compatibility.sh
```diff
@@ -31,8 +31,9 @@ compatibility_test() {
         remove_existing_stackrox_resources
         setup_default_TLS_certs
 
-        deploy_stackrox
+        deploy_stackrox_with_custom_sensor "${SENSOR_CHART_VERSION}"
         echo "Stackrox deployed"
+        kubectl -n stackrox get deploy,ds -o wide
```
jschnath, line 36: Don't forget to remove this line before merging

scripts/ci/jobs/gke_version_compatibility_tests.py
jschnath: Are the release versions available on Helm a subset of all available release versions?

Anyways using helm search to get the latest versions is a oneliner now thanks to Gavins command, I think all it should take is plugging that command into the `make_image_tag` function and calling it instead of `subprocess.check_output(["make", "--quiet", "--no-print-directory", "tag"]).decode(encoding="utf-8")`

tests/e2e/lib.sh
```diff
@@ -78,6 +108,29 @@ deploy_central() {
     "$ROOT/${DEPLOY_DIR}/central.sh"
 }
 
+deploy_sensor_from_helm_charts() {
+    if [ -z $1 ]; then
+        die "deploy_sensor_from_helm_charts should receive a helm chart version\nusage: deploy_sensor_from_helm_charts <Chart version> <path to init bundle>"
+    fi
+	chart_version="$1"
+    if [ -z $2 ]; then
```
jschnath, line 116:
```suggestion
    if [[ "$#" -ne 1 ]]; then
        die "deploy_sensor_from_helm_charts should receive a helm chart version\nusage: deploy_sensor_from_helm_charts <Chart version> <path to init bundle>"
    fi
	chart_version="$1"
    if [[ "$#" -ne 2 ]]; then
```
Does the same as what Gavin mentioned apply here? Unsure if that's the case but it may be better to have a look.

jschnath (commented):
Thanks Fred, I left a few comments but I didn't find many problems aside from what Gavin mentioned.

scripts/ci/jobs/gke_version_compatibility_tests.py
fredrb: Yep. Just a tiny correction: it should be `.version` instead of `.app_version`.

fredrb: Done

tests/e2e/lib.sh
fredrb: Done

qa-tests-backend/scripts/run-compatibility.sh
fredrb: I think we should keep it. It doesn't harm to output the actual images in the cluster and it might help catch some problems in case we're deploying incorrect versions of collector or scanner.

tests/e2e/lib.sh
fredrb: Done

jschnath:
Running the version compatibility script using helm I get
```
72.1.0
71.2.0
70.0.0
```
is it correct that there is no version 69 on helm and no patches for version 70?

jschnath (approved):
LGTM, have you forwarded the failing DefaultPoliciesTest and NetworkFlowTest to the teams that own them?

fredrb:
> LGTM, have you forwarded the failing DefaultPoliciesTest and NetworkFlowTest to the teams that own them?

Not yet. Still on my todo list.

PR #3854, opened 2022-11-18
Start using the batch/v1 api for CronJobs

Molter73:
## Description

Closes stackrox/collector#907

The `v1beta1.CronJob` api is unavailable in k8s v1.25 and it causes sensor to not be able to synchronize k8s resources, which in turn causes all information gathered by collector to be discarded (since sensor is not able to link container IDs to k8s resources).

The `v1.CronJob` API has been available since k8s v1.21 (which is already EOL). Older versions are no longer supported by k8s or openshift, so we should be safe to make this change.

## Checklist

## Testing Performed
- [x] Tested on k3s with kubernetes v1.22.11
- [x] Tested on k3s with kubernetes v1.25.3


vikin91 (approved):
The code change looks good! 
Maybe we could check whether there are some unit tests and adapt them as well?

@lvalerom @fredrb would this change affect the safety-net tests for re-sync?

sensor/kubernetes/listener/resources/convert.go
```diff
@@ -194,6 +195,11 @@ func (w *deploymentWrap) populateNonStaticFields(obj interface{}, action *centra
 		podLabels = o.Spec.JobTemplate.Spec.Template.GetLabels()
 		podSpec = o.Spec.JobTemplate.Spec.Template.Spec
 		labelSelector = o.Spec.JobTemplate.Spec.Selector
+	case *batchv1.CronJob:
```
fredrb, line 198: I assume it's safe to remove the case block above (`*v1beta1.CronJob`). With the handler being registered using `V1()` informer there won't be any events of that type coming in.

fredrb:
> would this change affect the safety-net tests for re-sync?

We don't have any CronJob resources in the replay tests nor specific tests in the integration tests, so no.

sensor/kubernetes/listener/resources/convert.go
lvalerom: +1

md2119: +1

md2119 (approved):

Molter73:
@vikin91, I can try adding a cronjob test case [here](https://github.com/stackrox/stackrox/blob/9582bc260468a16f928f5d2c5c671ed0fbad1100/sensor/kubernetes/listener/resources/convert_test.go#L377), think that could work/be enough?

vikin91:
> @vikin91, I can try adding a cronjob test case [here](https://github.com/stackrox/stackrox/blob/9582bc260468a16f928f5d2c5c671ed0fbad1100/sensor/kubernetes/listener/resources/convert_test.go#L377), think that could work/be enough?

Assuming that `deploymentType` could accept the cron resource, that may be okay. I am not too familiar with this test.

sensor/kubernetes/listener/resources/convert.go
Molter73: Done

PR #3856, opened 2022-11-20
UI dev dependencies typescript 4.9.3

pedrottimark:
## Description

TypeScript releases a minor update every quarter.

https://github.com/microsoft/TypeScript/releases/tag/v4.9.3

1. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#the-satisfies-operator

    > TypeScript developers are often faced with a dilemma: we want to ensure that some expression *matches* some type, but also want to keep the *most specific* type of that expression for inference purposes.

    > The new `satisfies` operator lets us validate that the type of an expression matches some type, without changing the resulting type of that expression.

    > `satisfies` can be used to catch lots of possible errors.

2. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#unlisted-property-narrowing-with-the-in-operator

    > As developers, we often need to deal with values that aren’t fully known at runtime. In fact, we often don’t know if properties exist, whether we’re getting a response from a server or reading a configuration file. JavaScript’s `in` operator can check whether a property exists on an object.

    > TypeScript 4.9 makes the `in` operator a little bit more powerful when narrowing types that *don’t* list the property at all. Instead of leaving them as-is, the language will intersect their types with `Record<"property-key-being-checked", unknown>`.

3. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#checks-for-equality-on-nan

    > TypeScript now errors on direct comparisons against `NaN`, and will suggest using some variation of `Number.isNaN` instead.

4. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#file-watching-now-uses-file-system-events

    > In earlier versions, TypeScript leaned heavily on *polling* for watching individual files. Using a polling strategy meant checking the state of a file periodically for updates.

    > In TypeScript 4.9, file watching is powered by file system events by default, only falling back to polling if we fail to set up event-based watchers. For most developers, this should provide a much less resource-intensive experience when running in `--watch` mode, or running with a TypeScript-powered editor like Visual Studio or VS Code.

5. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#remove-unused-imports-and-sort-imports-commands-for-editors

    > In TypeScript 4.3, we introduced a command called "Sort Imports" which would only sort imports in the file, but not remove them.

    > TypeScript 4.9 adds the other half, and now provides "Remove Unused Imports". TypeScript will now remove unused import names and statements, but will otherwise leave the relative ordering alone.

    > Visual Studio Code (1.73 and later) will have support built in and will surface these commands via its Command Palette.

6. https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#go-to-definition-on-return-keywords

    > In the editor, when running a go-to-definition on the `return` keyword, TypeScript will now jump you to the top of the corresponding function. This can be helpful to get a quick sense of which function a return belongs to.

    > We expect TypeScript will expand this functionality to more keywords such as `await` and `yield` or `switch`, `case`, and `default`.

7. What’s Next

    > Our next release will be TypeScript 5.0 – but have no fear, we believe 5.0 should be pretty easy to adopt.

    https://github.com/microsoft/TypeScript/issues/51362

### Errors

Because `yarn build` reports only the first error, and not even the name of the file:

1. `yarn tsc` in ui/apps/platform

    * src/Containers/Clusters/ClustersSidePanel.tsx

        > error TS2345: Argument of type '"DEPLOYMENT"' is not assignable to parameter of type 'SetStateAction<"FORM">'.

        ```js
        setWizardStep(wizardSteps.DEPLOYMENT);
        ```

        > error TS2367: This comparison appears to be unintentional because the types '"FORM"' and '"DEPLOYMENT"' have no overlap.

        ```js
        {!isBlocked && wizardStep === wizardSteps.DEPLOYMENT && (
        ```

2. `yarn tsc` in ui/packages/ui-components

### Solution

1. Replace object in helpers file with string union type in component file.
2. Declare type of `useState` hook.

## Checklist

## Testing Performed

1. `yarn lint` in ui
2. `yarn build` in ui
3. `wc build/static/js/*.js` in ui
    * branch - master: -188 = 11083100 - 11083288 because replace object properties with literal strings

### DIFF
--- a/ui/apps/platform/package.json
+++ b/ui/apps/platform/package.json
@@ -167,7 +167,7 @@
         "redux-immutable-state-invariant": "^2.1.0",
         "redux-saga-test-plan": "^3.7.0",
         "tailwindcss": "^2.0.3",
-        "typescript": "^4.8.4"
+        "typescript": "^4.9.3"
     },
     "browserslist": [
         ">0.2%",
--- a/ui/apps/platform/src/Containers/Clusters/ClustersSidePanel.tsx
+++ b/ui/apps/platform/src/Containers/Clusters/ClustersSidePanel.tsx
@@ -28,7 +28,6 @@ import DownloadHelmValues from './DownloadHelmValues';
 import {
     clusterDetailPollingInterval,
     newClusterDefault,
-    wizardSteps,
     centralEnvDefault,
 } from './cluster.helpers';
 import { CentralEnv, ClusterManagerType } from './clusterTypes';
@@ -47,6 +46,8 @@ const validate = (values) => {
     return errors;
 };
 
+type WizardStep = 'FORM' | 'DEPLOYMENT';
+
 type MessageState = {
     type: 'warn' | 'error';
     message: JSX.Element | string;
@@ -62,7 +63,7 @@ function ClustersSidePanel({ selectedClusterId, setSelectedClusterId }) {
     const [clusterRetentionInfo, setClusterRetentionInfo] =
         useState<DecommissionedClusterRetentionInfo>(null);
     const [centralEnv, setCentralEnv] = useState<CentralEnv>(centralEnvDefault);
-    const [wizardStep, setWizardStep] = useState(wizardSteps.FORM);
+    const [wizardStep, setWizardStep] = useState<WizardStep>('FORM');
     const [loadingCounter, setLoadingCounter] = useState(0);
     const [messageState, setMessageState] = useState<MessageState | null>(null);
     const [isBlocked, setIsBlocked] = useState(false);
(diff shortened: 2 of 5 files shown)


dvail (approved):
Ah it looks like TypeScript inferred `useState(wizardSteps.FORM)` to be `useState<'FORM'>(wizardSteps.FORM)`, when on 4.8 it was `string`. Interesting I wonder if that has to do with an inference change from one of the first few release notes.

pedrottimark:
Yes, although difficult to say which change caused TypeScript to check more carefully, it is a helpful error report.

I was pleasantly surprised only one such error.

PR #3863, opened 2022-11-21
Dashrews/ROX-11248 add capability to prune log imbues

dashrews78:
## Description

Add in logic to prune logimbues.  Set up to prune once a day and to prune logimbues that are more than a week old.  The time is not stored in the database so currently pruning same way most everything else prunes.  We may want a future ticket to add the time to the table.

## Checklist

## Testing Performed

Added unit tests and updated the Makefile so that the pruning tests actually run for Postgres in CI.


central/pruning/pruning.go
```diff
@@ -144,13 +151,16 @@ func (g *garbageCollectorImpl) pruneBasedOnConfig() {
...
+
+		g.pruneLogImbues()
 	}
 
 	log.Info("[Pruning] Finished garbage collection cycle")
 }
 
 func (g *garbageCollectorImpl) runGC() {
 	lastClusterPruneTime = time.Now().Add(-24 * time.Hour)
+	lastLogImbuePruneTime = time.Now()
```
c-du, line 163: Look at line 162, I see the purpose of setting the last time back 24 hours is to make sure the prune could happen right at the start up time. It looks like log imbue prune also need it?

pkg/postgres/pgtest/postgres.go
```diff
@@ -22,6 +22,7 @@ import (
 type TestPostgres struct {
 	*pgxpool.Pool
 	database string
+	GormDB   *gorm.DB
```
c-du, line 25: It does not seem to be necessary to add this? We do not have GormDB in central production codes.

c-du (approved):

md2119 (approved):

PR #3865, opened 2022-11-21
ROX-13173 + ROX-13174: Add advanced filters and entity name search input in baselines tab

sachaudh:
## Description

This PR adds the entity name search input and advanced filters to the baselines tab (similar to the flows/traffic tab). I moved things around as a result of using the same piece of code in multiple places:
1. `AdvancedFlowsFilter` related components + utils were moved into their own directory in `common/`
2. Created a new `EntityNameSearchInput` component. I'm not really sure if it's all that necessary, but if we need to modify it, we can modify it in one spot and see it change in both places
3. The `Flow` and `FlowBase` types were moved out into a common location (ie. `types/`)
4. A new `getAllUniqPorts` function was made in the `utils/flowUtils` file. We can keep any common flow-related functions here

## Checklist
- ~[] Unit test and regression tests added~

### Screenshots

Screenshot 2022-11-21 at 10 48 49 AM


ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentBaselines.tsx
```diff
@@ -1,14 +1,96 @@
...
+    const [entityNameFilter, setEntityNameFilter] = React.useState<string>('');
+    const [advancedFilters, setAdvancedFilters] = React.useState<AdvancedFlowsFilterType>(
+        defaultAdvancedFlowsFilters
+    );
 
     const handleAlertingOnViolations = (checked: boolean) => {
         setIsAlertingOnViolations(checked);
     };
 
+    const allUniqPorts = getAllUniqPorts(baselines);
```
vjwilson, line 92: nit: Even though Lodash abbreviates `uniquie` as `uniq`, it's confusing for not a lot of saving. (Two characters. You could say it's going to be auto-completed in most editors most of the time, but that begs the question "why save the two characters then?")

vjwilson (approved):

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentBaselines.tsx
sachaudh: True. I'm ok with the complete name vs. the shortened name. We can do that

PR #3867, opened 2022-11-21
Dynamically determine directory for UI auth token script

dvail:
## Description

The utility script that gets a valid auth token (e.g. for Cypress runs) previously needed to be run with `pwd` equal to the location of the script. This change makes it so that the script can be run from anywhere and it will correctly find the files it needs.

## Checklist

## Testing Performed

Run `get-auth-token.sh` from a variety of directories. Each time a valid auth token should be echoed.

Run `cypress open` and ensure Cypress is able to authenticate and run tests.

### DIFF
--- a/ui/apps/platform/scripts/get-auth-token.sh
+++ b/ui/apps/platform/scripts/get-auth-token.sh
@@ -4,11 +4,13 @@
 # Env vars ROX_USERNAME and ROX_PASSWORD are expected to be set for basic auth,
 # otherwise k8s deployment assumed from where basic auth creds are retrieved.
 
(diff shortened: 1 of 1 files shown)


dvail:
* master
  * **PR #3867** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3867" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3867?utm_source=stack-comment).

sachaudh (approved):
LGTM 👍🏼

PR #3868, opened 2022-11-21
Fix duplicate fetch requests in collections view more button

dvail:
## Description

An oversight in the logic determining whether or not there are more collections to be fetched was leading to more network requests than were needed. This fixes the logic and causes only one request per view-more click to be fired, as well as firing only a single request on the load of the collection form.

## Checklist

## Testing Performed

Load the collection form page and view the network tab to ensure only a single request is sent to the "list collections" endpoint. Click the view more button multiple times, and ensure that each click only results in a single request.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/hooks/useEmbeddedCollections.ts
+++ b/ui/apps/platform/src/Containers/Collections/hooks/useEmbeddedCollections.ts
@@ -64,7 +64,8 @@ function fetchDetachedCollections(
         );
         const detached = aggregateResult.concat(newDetached);
         const lastResponseSize = collections.length;
-        const shouldFetchMore = lastResponseSize > 0 && detached.length < minimumUpdateSize;
+        const shouldFetchMore =
+            lastResponseSize === pageSize && detached.length < minimumUpdateSize;
         const nextPage = pageNumber + 1;
 
         if (shouldFetchMore) {


dvail:
* master
  * **PR #3868** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3868" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3868?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/hooks/useEmbeddedCollections.ts
```diff
@@ -64,7 +64,8 @@ function fetchDetachedCollections(
         );
         const detached = aggregateResult.concat(newDetached);
         const lastResponseSize = collections.length;
-        const shouldFetchMore = lastResponseSize > 0 && detached.length < minimumUpdateSize;
+        const shouldFetchMore =
```
dvail, line 67: With this change. we only automatically fetch more for the available collection list if the following conditions are met:
- The `lastResponseSize` of the previous fetch was equal to the size of the pages we are fetching (if it is less, then no more data exists on the server)
- The amount of collections we are adding to the available list is less than the `minimumUpdateSize`, which is half the page size. This handles situations where we fetch a page of data, but > half of those collections already exist in the attached list.

Previously, with `lastResponseSize > 0`, we would get one redundant fetch when the last page was returned and we were below the minimum update size. This happened frequently when there was very little data in the system and you were viewing a collection that has any embedded collections.

PR #3878, opened 2022-11-22
ROX-13449: Keep compatibility tests running after failure

jschnath:
Disabled the FAIL-FAST flag for gke-version-compatibility-tests to ensure test runs continue after a failure, added a try / catch block to the python loop calling those tests to keep running tests for the next sensor version after failure.

Added an always failing test for to ensure test failing behaviour can be checked, this will be deleted before merging.

### DIFF
--- a/qa-tests-backend/src/test/groovy/CertExpiryTest.groovy
+++ b/qa-tests-backend/src/test/groovy/CertExpiryTest.groovy
@@ -6,7 +6,7 @@ import util.Cert
 
 @Category([BAT, COMPATIBILITY])
 class CertExpiryTest extends BaseSpecification {
-    def: "Failing test"() {
+    def "Failing test - Delete before merging"() {
         when:
         "This test is supposed to fail"
 
--- a/qa-tests-backend/src/test/groovy/CertExpiryTest.groovy
+++ b/qa-tests-backend/src/test/groovy/CertExpiryTest.groovy
@@ -6,6 +6,7 @@ import util.Cert
 
 @Category([BAT, COMPATIBILITY])
 class CertExpiryTest extends BaseSpecification {
+//TODO(ROX-13449): Delete the following test case before merging
     def "Failing test - Delete before merging"() {
         when:
         "This test is supposed to fail"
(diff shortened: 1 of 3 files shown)


jschnath:
This is looking good:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3878/pull-ci-stackrox-stackrox-master-gke-version-compatibility-tests/1595153979824672768
My assert false triggered 4 times so catching the exception in the python loop and continuing worked. I'll add some logging too.

qa-tests-backend/src/test/groovy/CertExpiryTest.groovy
```diff
@@ -6,6 +6,15 @@ import util.Cert
...
 class CertExpiryTest extends BaseSpecification {
+//TODO(ROX-13449): Delete the following test case before merging
+    def "Failing test - Delete before merging"() {
+        when:
+        "This test is supposed to fail"
+
+        then:
+        "Check the behavior of the fail-fast flag in Prow"
+        assert False
+    }
```
vikin91, line 17: Have you been able to see it failing already? Is this piece of code now ready to be removed?

jschnath: Yes, seen it failing twice. I'll remove it now.

jschnath:
Since the failing test has been removed now the next run should be green:

/test images
/test push-images
/test gke-version-compatibility-tests

fredrb:
This is great, Jan. However, I think the whole build should fail if at least one sensor version test failed.

jschnath:
>I think the whole build should fail if at least one sensor version test failed.
You make a great point, I hadn't noticed that the build passed. I'll add that functionality

jschnath:
Failing as expected now
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/3878/pull-ci-stackrox-stackrox-master-gke-version-compatibility-tests/1595728641172443136

jschnath:
@fredrb I think it's ready to merge.

fredrb (approved):

PR #3879, opened 2022-11-22
Install minimal python in docs image

janisz:
## Description

To serve docs we only need http server. There is no need to install whole python that often contains CVEs.

Refs: https://quay.io/repository/rhacs-eng/docs/manifest/sha256:aaa6cc9f409b70f5c504944f47837aec0f7cc11fd22e615aa12577ab0a8f4a63?tab=vulnerabilities

## Testing Performed
```
docker build  .
docker run ... // and open web page in browser
```

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -4,14 +4,21 @@ WORKDIR /workspace
 COPY content content
 COPY tools tools
 COPY Makefile ./
-RUN wget -O busybox https://busybox.net/downloads/binaries/1.31.0-defconfig-multiarch-musl/busybox-`arch` && chmod +x busybox
 
 RUN make
 
-FROM registry.access.redhat.com/ubi8-micro:8.7
+FROM registry.access.redhat.com/ubi8-minimal:latest
+
+# Update RPMs to hopefully aid with vulns that aren't yet patched in the base image. Also remove package management
+# utils because our own policies don't recommend having them in the image. For that we need to be root.
+RUN microdnf upgrade && \
+    microdnf install --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
+    python39 && \
+    microdnf clean all && \
+    rpm --verbose -e --nodeps $(rpm -qa curl '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
+    rm -rf /var/cache/dnf /var/cache/yum
 
 WORKDIR /public
 COPY --from=builder /workspace/build/ ./
-COPY --from=builder /workspace/busybox /bin/busybox
 EXPOSE 8080
-CMD busybox httpd -f -p 8080
+CMD python -m http.server 8080
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -4,21 +4,14 @@ WORKDIR /workspace
 COPY content content
 COPY tools tools
 COPY Makefile ./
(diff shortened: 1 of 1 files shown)


docs/Dockerfile
```diff
@@ -4,21 +4,14 @@ WORKDIR /workspace
 COPY content content
 COPY tools tools
 COPY Makefile ./
+RUN wget -O busybox https://busybox.net/downloads/binaries/1.31.0-defconfig-multiarch-musl/busybox-`arch` && chmod +x busybox
```
msugakov, line 7: Maybe I'm getting older but I would resist this kind of thing - downloading an executable from the internet outside of well-known package repositories and not performing any kind of signature fingerprint validation (is `busybox` signed at all?)

An alternative approach is to use nginx to host docs. `nginx` is available as a package on `ubi8-minimal`.

Not quite the same, but we use nginx container for docs downstream https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-docs/Dockerfile.in

Apart from that, the initiative to remove product docs built in the product and as a stand-alone container is scheduled for 3.74.0 - https://issues.redhat.com/browse/ROX-12465
Maybe we don't need this change at all?

msugakov (changes requested):

docs/Dockerfile
janisz: I need this to release 3.73 due to known vulns. I agree that downloading something from internet is not the best option unfortunately I can't find busybox on ubi minimal :(
Ideal solution will be to use official busybox image.

docs/Dockerfile
```diff
@@ -12,8 +12,8 @@ FROM registry.access.redhat.com/ubi8-minimal:latest
 # Update RPMs to hopefully aid with vulns that aren't yet patched in the base image. Also remove package management
 # utils because our own policies don't recommend having them in the image. For that we need to be root.
 RUN microdnf upgrade && \
-    microdnf install \
-        python39 && \
+    microdnf install --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
```
janisz, line 15: This will keep python installation minimal.

```diff
Installing: libtirpc;1.1.4-8.el8;x86_64;ubi-8-baseos-rpms
Installing: libnsl2;1.2.0-2.20180605git4a062cf.el8;x86_64;ubi-8-baseos-rpms
Installing: python39-setuptools-wheel;50.3.2-4.module+el8.5.0+12204+54860423;noarch;ubi-8-appstream-rpms
Installing: python39-pip-wheel;20.2.4-7.module+el8.6.0+13003+6bb2c488;noarch;ubi-8-appstream-rpms
Installing: expat;2.2.5-10.el8;x86_64;ubi-8-baseos-rpms
Installing: gdbm-libs;1:1.18-2.el8;x86_64;ubi-8-baseos-rpms
Installing: python39;3.9.13-2.module+el8.7.0+17195+44752b34;x86_64;ubi-8-appstream-rpms
Installing: python39-libs;3.9.13-2.module+el8.7.0+17195+44752b34;x86_64;ubi-8-appstream-rpms
< Installing: python39-setuptools;50.3.2-4.module+el8.5.0+12204+54860423;noarch;ubi-8-appstream-rpms
< Installing: python39-pip;20.2.4-7.module+el8.6.0+13003+6bb2c488;noarch;ubi-8-appstream-rpms

```

docs/Dockerfile
janisz: Since it's going to be removed maybe removing problematic packages solves the problem.

msugakov (approved):

PR #3881, opened 2022-11-22
Add error screen for collections with configurations unsupported by the UI

dvail:
## Description

This adds a UI screen for when a collection response returned from the server contains valid configuration rules that cannot be displayed in the UI. Although these collections cannot be viewed or edited with the UI controls, their results can still be viewed and they can be attached to other collections.

This should occur rarely (or not at all) since the server checks most of these situations on create/edit, but this is a more user-friendly fallback instead of an error.

### Follow ups

- Work with UX and/or docs team on better wording for the specific error messages.
- If possible, remove the "Edit" and "Clone" options from the header dropdown.

## Checklist

## Testing Performed

Navigate to a collection page with rules that are unsupported by the UI. Note that this can only be accomplished currently by creating the collection via the API, and only by creating a collection with rules for two different fields for the same entity type. (e.g. A single collection with a rule for "Namespace" and "Namespace Label".) The collection page will display the following UI. Collection results will still be able to be displayed.
image

The  "More info" section can expand to give details information on why the collection cannot be displayed. (Better error info is TBD.)
image

This collection can still be attached to other collections without error:
image

Viewing the collection in a modal will still display the error screen:
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/UnsupportedCollectionState.tsx
+++ b/ui/apps/platform/src/Containers/Collections/UnsupportedCollectionState.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { ExpandableSection, List, ListItem } from '@patternfly/react-core';
+import { InfoCircleIcon } from '@patternfly/react-icons';
+
+import EmptyStateTemplate from 'Components/PatternFly/EmptyStateTemplate';
+import { CollectionParseError } from './converter';
+
+export type UnsupportedCollectionStateProps = {
+    errors: CollectionParseError['errors'];
+    className?: string;
+};
+
+function UnsupportedCollectionState({ errors, className = '' }: UnsupportedCollectionStateProps) {
+    return (
+        <div className={className}>
+            <EmptyStateTemplate
+                title="This collection cannot be displayed"
+                headingLevel="h2"
+                icon={InfoCircleIcon}
+            >
+                <p className="pf-u-pb-lg">
+                    This collection is valid but cannot be displayed nor edited through the ACS UI
+                </p>
+                <ExpandableSection toggleText="More info">
+                    <List className="pf-u-text-align-left">
+                        {errors.map((err) => (
+                            <ListItem key={err}>{err}</ListItem>
+                        ))}
+                    </List>
+                </ExpandableSection>
+            </EmptyStateTemplate>
+        </div>
+    );
+}
+
+export default UnsupportedCollectionState;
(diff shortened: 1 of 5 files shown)


dvail:
* master
  * **PR #3881** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3881" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3881?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/converter.ts
```diff
@@ -25,12 +25,29 @@ const fieldToEntityMap: Record<SelectorField, SelectorEntityType> = {
 
 const LABEL_SEPARATOR = '=';
 
+export type CollectionParseError = {
```
dvail, line 28: The `AggregateError` was a little clunky to work with so I replaced it with a simple custom type to track error messages.

(`AggregateError` is not generic and its `errors` property  was always `any[]` so this helps avoid runtime checks everywhere else the error is used.)

pedrottimark: Thank you for another TypeScript lesson!

The non-empty requirement in `CollectionParseError` type and predicate function makes me curious to ask whether TypeScript reports an error if `parseCollection` has an incorrect code path which returns an empty `errors` array?

ui/apps/platform/src/Containers/Collections/UnsupportedCollectionState.tsx
```diff
@@ -0,0 +1,36 @@
...
+function UnsupportedCollectionState({ errors, className = '' }: UnsupportedCollectionStateProps) {
+    return (
+        <div className={className}>
+            <EmptyStateTemplate
+                title="This collection cannot be displayed"
+                headingLevel="h2"
+                icon={InfoCircleIcon}
+            >
+                <p className="pf-u-pb-lg">
+                    This collection is valid but cannot be displayed nor edited through the ACS UI
```
pedrottimark, line 22: For product branding, can we replace **the ACS UI** with something like **the user interface**.

pedrottimark: Although in this case, collapsed by default seems reasonable, I will mention a counter-example where it seems unreasonable and caused me problems yet again yesterday: when an integration test fails because the page throws an exception, `ErrorBoundary` renders the error message as collapsed in the cypress snapshot.

pedrottimark (approved):
Interesting point about page and table actions.

PR #3068, opened 2022-09-13
ROX-8674: Replace vmWorkflowUtils helper functions to fix timing problems in integration tests

pedrottimark:
### Test failures

> Components list Page and its entity detail page, (related entities) sub list validations with VM updates ON should display all the columns expected in node components list page

screenshot-1567400461261082624

cypress/integration/vulnmanagement/componentsListPages.test.js

1. 1567400461261082624 from branch build for #2984 on 2022-09-07
    > Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `auth/status`. No request ever occurred.
2. 1567566650423644160 from branch build for #2996 on 2022-09-07
    > Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `credentialexpiry_SCANNER`. No request ever occurred.

The trend continues: during past 12 hours on 2022-09-13, 6 more failures for master builds compared to only 3 successes.

Recent test failures are similar to an earlier test failure on 2021-11-30

> Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `getImages`. No request ever occurred.

ROX-8674

### Analysis

The test files for entities have a big test, for example: `'Clusters list Page and its single entity detail page, and sub list validations'`
* Much opportunity for request timeouts. Not enough information to wait on requests.
* In hypothetical case that test fails because of a problem other than timeout, where is the problem: this, that, or the other?
* If test criteria are skipped conditionally because of data (that is, because first row of table does not have a link) then unreliable criteria can remain dormant, and then suddenly cause tests to fail at unconvenient times.

componentsListPages has 2 big tests back-to-back, and the second fails often: double trouble!

### Solution

Sadly, the solution is difficult to break down into smaller steps.

1. Separate big test into smaller tests, each of which has a short description.
2. Rewrite complicated helper functions to explicitly identify steps through a workflow.
3. Add constants for all entities so helper functions can wait on relevant requests.
4. Verify the first link in the column instead of verify only if the first row has a link in the column.

### Changed files

1. Edit cypress/constants/apiEndpoints.js
    * Rename objects and add missing properties (see Residue point 2).

2. Edit cypress/helpers/vulnmanagement/entities.js
    * Factor out `requestConfigForDashboard` for functions which visit dashboard.
    * Import opname objects and add `keyForEntity` object.
[...]

PR #3885, opened 2022-11-22
ROX-13564: Replace vmWorkflowUtils helper functions in non-postgress vulnmanagement integration tests

pedrottimark:
### Overview

1. Because a minority of vulnmanagement tests have `'ROX_POSTGRES_DATASTORE'` feature flag disabled, we have waited for evidence of timing problems.
2. Because previous true positives for non-postgress regressions were mixed with simultaneous false positives, apply the solution to replace confusion with clarity in the case of future regressions.
3. Prepare now for future removal of non-postgress tests and conditional skips in postgress tests. Move implicit head knowledge into explicit test organization, so it becomes a no-brainer.

### Test failure

1. 1592676501193494528 from master build for #3789 on 2022-11-15

> Vulnerability Management Deployments with VM updates OFF should display all the columns and links expected in deployments list page

> Timed out retrying after 4000ms: Expected to find element: `[data-testid="allCvesLink"]`, but never found it.

at validateAllCVELinks cypress/helpers/vmWorkflowUtils.js:126:7
at allCVECheck cypress/helpers/vmWorkflowUtils.js:233:4
at Context.eval cypress/integration/vulnmanagement/deploymentsListPages.test.js:60:32

deploymentsListPages

### Analysis

The test files for entities have a big test:
* Much opportunity for request timeouts. Not enough information to wait on requests.
* In hypothetical case that test fails because of a problem other than timeout, where is the problem: this, that, or the other?
* If test criteria are skipped conditionally because of data (that is, because first row of table does not have a link) then unreliable criteria can remain dormant, and then suddenly cause tests to fail at unconvenient times.

### Solution

Apply pattern for tests with `'ROX_POSTGRES_DATASTORE'` feature flag enabled in #3068

### Organization

1. Rename and separate entity tests so camelCase corresponds to hyphen-case path segments for entities.
    * Remove ListPages suffix for **postgress** test files to leave plural camelCase of entity.
        For example, rename clusterCvesListPages.test.js as clusterCves.test.js
        * Two-part entities like clusterCves are postgress only.
        * Objective of test file is to test entities, even if many of its tests start from list page.
    * Replace ListPages suffix with NonPostgress for **non-postgress** test files.
        For example, rename cvesListPages.test.js as cvesNonPostgress.test.js
    * Separate test files which have combined tests.
        * deployments.test.js and deploymentsNonPostgress.test.js
        * namespaces.test.js and namespacesNonPostgress.test.js
2. Replace test patterns which import from vmWorkflowUtils with entities helpers file.
3. Do not separate tests in files which do not skip tests, but instead have ternary expressions:
    * dashboard.test.js
    * entitypages.test.js

### Residue

Because `hasTableColumnHeadings` function explicitly asserts existence of checkbox, hidden, and invisible column headings:
* Increment `columnIndex` arguments of verify function calls to be consistent with arrays of column heading strings in the function call.
* Replace `columnIndex + 1` with `columnIndex` in `nth-child` pseudo-selector.

### Future removal of non-postgress tests

1. Delete whateverNonPostgress.test.js files

2. Delete skip from test files:

    ```js
    before(function beforeHook() {
        if (!hasFeatureFlag('ROX_POSTGRES_DATASTORE')) {
            this.skip();
        }
    });
    ```

3. Delete skip from tests:

    ```js
    if (!hasFeatureFlag('ROX_POSTGRES_DATASTORE')) {
        this.skip();
    }
    ```

4. Replace conditional ternary expressions with postgress-only value, for example:

    * replace `const entitiesKey = hasFeatureFlag('ROX_POSTGRES_DATASTORE') ? 'image-cves' : 'cves';`
    * with `const entitiesKey = 'image-cves';`

## Checklist

## Testing Performed


vjwilson (approved):
/lgtm

PR #3888, opened 2022-11-23
ROX-9115: Run e2e tests outside of CI

gavin-stackrox:
## Description

This PR creates a wrapper script to run e2e tests from a developers working directory. It runs the same image as CI to avoid local system version and environment issues. The required secrets are pulled from the OpenShift CI vault. This PR only supports part I of the groovy tests in `qa-tests-backend`. Future PRs will address:

- Other e2e
  - non-groovy e2e tests
  - upgrade test
  - scale test
  - UI e2e (if desired)
- Make setup() optional and reusable e.g. at present e2e is deployed with `--default-tls-*` but that data is lost after test and could not be used for redeploy/retest.
- Make deploy() optional (I think it is already reusable but we shall see.)
- Invoke Suite/Case not the entire test (but relying on prior data from setup()/deploy()).
- Make the vault use less of a pain. It currently needs a login per run. Or a VAULT_TOKEN setting but that is short lived.

## Requires
https://github.com/stackrox/rox-ci-image/pull/174

## Checklist

## Testing Performed

CI is sufficient.

Tested `./tests/e2e/run-e2e-tests.sh` with GKE and OpenShift. DeploymentTest.groovy fails regularly due to docker.io rate limiting and that will be addressed in a followup PR.


janisz (commented):
I think we need readme with info when and how it could be used

tests/e2e/run-e2e-tests.sh
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1091
+
+# Run a full suite of e2e tests using the code and against the cluster defined
+# in the calling environment via the rox-ci-image / stackrox-test container.
```
porridge, line 5:
```suggestion
# Run a full suite of e2e tests using the code via the rox-ci-image / stackrox-test container,
# against the cluster defined in the calling environment.
```

tests/e2e/run-e2e-tests.sh
```diff
@@ -0,0 +1,159 @@
...
+    docker run \
+      -v "$ROOT:$ROOT:z" \
+      -w "$ROOT" \
+      -e "KUBECONFIG=${kubeconfig}" \
+      -v "${kubeconfig}:${kubeconfig}:z" \
+      -e VAULT_TOKEN \
+      --platform linux/amd64 \
+      --rm -it \
+      --entrypoint="$0" \
+      quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.52-3-g49dc4af1cd "$@"
```
porridge, line 35: This will eventually use `/BUILD_IMAGE_VERSION`, right?

porridge (approved):

tests/e2e/run-e2e-tests.sh
gavin-stackrox: Sure. I was going to just make it a released version when the rox-ci-image PR lands but s/-build-/-test-/ on BUILD_VERSION makes sense.

gavin-stackrox:
> I think we need readme with info when and how it could be used

In subsequent PRs I've expanded the usage() to include examples. Will that be sufficient do you think?

tests/e2e/run-e2e-tests.sh
gavin-stackrox: Actually, I'll need to wait until the next BUILD_IMAGE_VERSION bump as the rox-ci-image required for this is later.

PR #3897, opened 2022-11-23
ROX-12626: Collections E2E Tests

keyallis:
## Description

Add E2E tests for collections, mainly testing the logical matching of deployments.
Fix the service endpoint requirements for creation and update of collections (*at least* one selector rule or embedded obj).

## Checklist

## Testing Performed

Haha

### DIFF
--- a/tests/resourcecollection_test.go
+++ b/tests/resourcecollection_test.go
@@ -131,7 +131,7 @@ func (s *CollectionE2ETestSuite) SetupSuite() {
 	}
 	for _, req := range createCollectionRequests {
 		resp, err := s.service.CreateCollection(s.ctx, req)
-		assert.NoError(s.T(), err)
+		s.NoError(err)
 		s.collectionIDs = append(s.collectionIDs, resp.GetCollection().GetId())
 	}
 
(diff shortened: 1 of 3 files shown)


tests/resourcecollection_test.go
```diff
@@ -0,0 +1,499 @@
...
+	suite.Suite
+
+	ctx        context.Context
+	nsClient   k8scorev1.NamespaceInterface
+	service    v1.CollectionServiceClient
+	depService v1.DeploymentServiceClient
+}
+
+func (s *CollectionE2ETestSuite) SetupSuite() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() || !features.ObjectCollections.Enabled() {
```
theencee, line 61: I think it's safe to make the assumption that if collections is enabled, postgres is enabled

tests/resourcecollection_test.go
```diff
@@ -0,0 +1,499 @@
...
+	} {
+		s.T().Run(tc.name, func(t *testing.T) {
+			resp, err := s.service.DryRunCollection(ctx, tc.request)
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, tc.expectedListDeployments, resp.GetDeployments())
+		})
+	}
+
+	// clean up test
+	for _, id := range collectionIDs {
```
theencee, line 450: I think this block may not run if test fails in between collection creation? What about putting it in `TearDownTest` (and the collection creation to `SetupTest`)?

Or put this in a defer block

tests/resourcecollection_test.go
```diff
@@ -0,0 +1,499 @@
...
+				MatchLabels: map[string]string{},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "nginx",
```
theencee, line 474: This will pull nginx:latest from docker.io which can cause us to get rate limited. How about `quay.io/rhacs-eng/qa:nginx-1-14-alpine` instead? Or any other image from https://quay.io/repository/rhacs-eng/qa?tab=tags would work as well.

tests/resourcecollection_test.go
keyallis: Changed to reflect that

tests/resourcecollection_test.go
keyallis: Moved the embedded collection creation to setup and the teardown now includes cleaning those up

md2119 (approved):
Nice! I recommend adding some regex rule collections to the test case.

PR #3904, opened 2022-11-23
ROX-13013: Add query string search parsing to PF Network Graph

vjwilson:
## Description

This is a good stopping place to check the URL flow.

Because the breadcrumb style is not in PatternFly yet, I have left the basic select input in, so that we can test the flow

## Checklist

## Testing Performed

Manual testing
Screen Shot 2022-11-23 at 6 41 12 PM

Screen Shot 2022-11-23 at 6 41 16 PM
Screen Shot 2022-11-23 at 6 41 34 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
@@ -1,146 +1,40 @@
-import React, { useCallback, ChangeEvent } from 'react';
-import {
-    Breadcrumb,
-    BreadcrumbItem,
-    Select,
-    SelectOption,
-    SelectVariant,
-} from '@patternfly/react-core';
+import React from 'react';
+import { Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
 
 import { Cluster } from 'types/cluster.proto';
 import useURLSearch from 'hooks/useURLSearch';
-import useFetchClusterNamespaces from 'hooks/useFetchClusterNamespaces';
-import useSelectToggle from 'hooks/patternfly/useSelectToggle';
-import { ClusterIcon, NamespaceIcon } from '../common/NetworkGraphIcons';
+import ClusterSelector from './ClusterSelector';
+import NamespaceSelector from './NamespaceSelector';
 import getScopeHierarchy from '../utils/getScopeHierarchy';
-
-function filterElementsWithValueProp(
-    filterValue: string,
-    elements: React.ReactElement[] | undefined
-): React.ReactElement[] | undefined {
-    if (filterValue === '' || elements === undefined) {
-        return elements;
-    }
-
-    return elements.filter((reactElement) =>
-        reactElement.props.value?.toLowerCase().includes(filterValue.toLowerCase())
-    );
-}
 
 type NetworkBreadcrumbsProps = {
     clusters: Cluster[];
 };
 
 function NetworkBreadcrumbs({ clusters = [] }: NetworkBreadcrumbsProps) {
-    const {
-        isOpen: isClusterOpen,
-        toggleSelect: toggleIsClusterOpen,
-        closeSelect: closeClusterSelect,
-    } = useSelectToggle();
-    const {
-        isOpen: isNamespaceOpen,
-        toggleSelect: toggleIsNamespaceOpen,
-        closeSelect: closeNamespaceSelect,
-    } = useSelectToggle();
-
     const { searchFilter, setSearchFilter } = useURLSearch();
 
-    const {
-        cluster: selectedClusterName,
-        namespaces: selectedNamespaces,
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        deployments: deploymentsFromUrl,
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        remainingQuery,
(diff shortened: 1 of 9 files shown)


ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+                        className="namespace-select"
+                        placeholderText="Namespaces"
+                        isDisabled={!selectedClusterId || loading || Boolean(error)}
+                        selections={selectedNamespaces}
+                        variant={SelectVariant.checkbox}
+                        maxHeight="275px"
+                        hasInlineFilter
+                    >
+                        {namespaceSelectOptions}
+                        {/* {availableNamespaceFilters.map((nsFilter) => (
```
sachaudh, line 139: Comment can be deleted

ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+import { Cluster } from 'types/cluster.proto';
+import useURLSearch from 'hooks/useURLSearch';
+import useFetchClusterNamespaces from 'hooks/useFetchClusterNamespaces';
+import { ClusterIcon, NamespaceIcon } from '../common/NetworkGraphIcons';
+import getScopeHierarchy from '../utils/getScopeHierarchy';
+
+function filterElementsWithValueProp(
+    filterValue: string,
+    elements: React.ReactElement[] | undefined
```
sachaudh, line 18: Should we specify that these will only be `SelectOption` components?

ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+            setSearchFilter(modifiedSearchObject);
+        }
+    };
+
+    const onFilterNamespaces = useCallback(
+        (e: ChangeEvent<HTMLInputElement> | null, filterValue: string) =>
+            filterElementsWithValueProp(
+                filterValue,
+                namespaces.map((filteredNS) => (
+                    <SelectOption key={filteredNS} value={filteredNS}>
```
sachaudh, line 71: A little farther down I noticed something similar where we take the namespaces and map them to get an array of SelectOptions. Can we re-use that logic here too?

sachaudh: Also nit: I think calling it `namespace` rather than `filteredNS` makes more sense. I was a little confused with the use of that term

ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    </SelectOption>
+                ))
+            ),
+        [namespaces]
+    );
+
+    const onNamespaceSelect = (_, selected) => {
+        setIsNamespaceOpen((prevIsOpen: boolean) => !prevIsOpen);
+
+        const newSelection = selectedNamespaces.find((nsFilter) => nsFilter === selected)
```
sachaudh, line 84: nit: Can we use `namespace` here as well? `nsFilter` seems like the wrong term for this

ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+            <span>
+                <NamespaceIcon /> {namespace}
+            </span>
+        </SelectOption>
+    ));
+
+    return (
+        <>
+            <Breadcrumb>
+                <BreadcrumbItem isDropdown>
```
sachaudh, line 111: Do you think it would be worthwhile to encapsulate the cluster-related breadcrumb logic and namespaces-related breadcrumb logic into their own separate components?

ui/apps/platform/src/Containers/NetworkGraph/components/NetworkBreadcrumbs.tsx
```diff
@@ -0,0 +1,149 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        deployments: deploymentsFromUrl,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        remainingQuery,
+    } = getScopeHierarchy(searchFilter);
+    const selectedClusterId =
+        clusters.find((cluster) => cluster.name === selectedClusterName)?.id || '';
+    const { loading, error, namespaces } = useFetchClusterNamespaces(selectedClusterId);
+
+    const onClusterToggle = (newClusterOpenValue) => setIsClusterOpen(newClusterOpenValue);
```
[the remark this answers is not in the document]
dvail: We do have one that I think fits this use case: https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/hooks/patternfly/useSelectToggle.ts

It is unfortunately named `useSelectToggle`, when in reality is should probably be `useToggle` since it can be used as a generic toggle for things other than `<Select>` components.

sachaudh (approved):
Approved to let you move things along

PR #3912, opened 2022-11-25
ROX-12936: added pprof profiling option to operator

johannes94:
## Description

In order to better debug OOM issues in the operator controller, I added the option to enable a heap dump once the memory consumption reaches a certain threshold. 

## Checklist

## Testing Performed

Did deployment to openshift local (crc) with dirty tag:
```
# Setup docker repositories with push rights
# I created with my RH user @ quay 3 repos
# stackrox-operator, stackrox-operator-bundle, stackrox-operator-index

# Push latest major release base version to stackrox-operator-index
docker pull quay.io/rhacs-eng/stackrox-operator-index:v3.72.0
docker tag quay.io/rhacs-eng/stackrox-operator-index:v3.72.0 quay.io/jmalsam/stackrox-operator-index:v3.72.0 
docker push quay.io/jmalsam/stackrox-operator-index:v3.72.0

# Make everything, this build and pushes operator, operator-bundle and operator-index images
# there might be an error if you run this on MacOS since the hack for opm render command tries to download 
# from a private repo of @porridge  https://github.com/porridge/operator-registry/releases
# you can build your own binary by checking out @porridge repo and run make build
# then comment the line downloading it and cp ./bin/opm to the target location
export IMAGE_REPO=quay.io/<your-name
make everything

# Create olm deployment in local cluster
make deploy-dirty-tag-via-olm

# Verify operator is running
kubectl get pods -n stackrox-operator

# Activate Profiling
kubectl patch subscriptions.operators.coreos.com -n stackrox-operator stackrox-operator-test-subscription --type "json" -p '[{"op":"add", "path":"/spec/config/env/-", "value": {"name": "ENABLE_PROFILING", "value":"true"}}]'

# Verify new pod with ENABLE_PROFILING is up and passes probes
kubectl get pods -n stackrox-operator

# Verify no dumps are written, there is no heapdump dir in /tmp if default threshold 0.80 is not hit
kubectl exec -n stackrox-operator <pod-name> -- ls /tmp

# Set PROFILING_THRESHOLD_FRACTION to a low value to force heap dumps
kubectl patch subscriptions.operators.coreos.com -n stackrox-operator stackrox-operator-test-subscription --type "json" -p '[{"op":"add", "path":"/spec/config/env/-", "value": {"name": "PROFILING_THRESHOLD_FRACTION", "value":"0.01"}}]'

# Verify that new pod is up and valid heap dump is written to /tmp/heapdump/<filename>
kubectl get pods -n stackrox-operator
# We can't use kubectl cp here because the operator container doesn't have "tar" installed, which is required by kubectl cp
kubectl exec -n stackrox-operator <pod-name> -- cat /tmp/heapdump/20221208T15-58-25.dump > dump.out
# This opens the pprof console type web to open a .svg file showing memory graph in your browser.
# Verify it looks like a valid memory profile
go tool pprof ./dump.out

# Test that 30 sec. default backoff between heapdumps is working properly by querying the heapdump folder and comparing timestamps 
 k exec -n stackrox-operator <pod-name> -- ls /tmp/heapdump/

# Test that file rotation after 10 files is working by waiting for the 11th file to be created and verifying that the 1st file was deleted
 k exec -n stackrox-operator <pod-name> -- ls /tmp/heapdump

# Test that you can dump to a specified path and volume mount in operator subscription
# add this to your subscription config
....
  config
    - name: HEAP_DUMP_PARENT_DIR
      value: "/dump"
    volumeMounts:
    - mountPath: /dump
      name: test-volume
    volumes:
    - emptyDir: {}
      name: test-volume
.... 
# Verify that dumps are written to expected folder
k exec -n stackrox-operator <pod-name> -- ls -l /dump/heapdump/
```

### DIFF
--- a/operator/main.go
+++ b/operator/main.go
@@ -17,9 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/pkg/errors"
 	platform "github.com/stackrox/rox/operator/apis/platform/v1alpha1"
@@ -29,6 +31,7 @@ import (
 	"github.com/stackrox/rox/pkg/buildinfo"
 	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/fileutils"
+	"github.com/stackrox/rox/pkg/profiling"
 	"github.com/stackrox/rox/pkg/version"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -48,7 +51,8 @@ var (
 	scheme          = runtime.NewScheme()
 	enableWebhooks  = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
 	enableProfiling = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
-
+	memLimit        = env.RegisterIntegerSetting("MEMORY_LIMIT_BYTES", 0)
+	heapDumpDir     = env.RegisterSetting("HEAP_DUMP_DIR", env.AllowEmpty())
 	// Default place where controller-runtime looks for TLS artifacts.
 	// see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/webhook/server.go#L96-L104
 	defaultCertDir  = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
@@ -106,6 +110,14 @@ func run() error {
 			return errors.Wrap(err, "unable to create Central webhook")
 		}
 	}
+
+	if enableProfiling.BooleanSetting() {
+		heapProfiler := profiling.NewHeapProfiler(0.80, uint64(memLimit.IntegerSetting()), heapDumpDir.EnvVar())
+		ctx, cancelProfiler := context.WithCancel(context.Background())
+		heapProfiler.DumpHeapOnThreshhold(ctx, time.Second)
+		defer cancelProfiler()
+	}
+
 	// The following comment marks the place where `operator-sdk` inserts new scaffolded code.
 	//+kubebuilder:scaffold:builder
 
--- a/operator/main.go
+++ b/operator/main.go
@@ -117,7 +117,7 @@ func run() error {
 		thresholdS := profilingThreshold.EnvVar()
 		thresholdF, err := strconv.ParseFloat(thresholdS, 32)
 		if err != nil {
-			return errors.Wrapf(err, "unable to parse PROFILING_THREASHOLD set to '%s' as a float", thresholdString)
+			return errors.Wrapf(err, "unable to parse PROFILING_THREASHOLD set to '%s' as a float", thresholdS)
 		}
 		heapProfiler := profiling.NewHeapProfiler(thresholdF, uint64(memLimit.IntegerSetting()), heapDumpDir.EnvVar())
 		ctx, cancelProfiler := context.WithCancel(context.Background())
--- a/operator/main.go
+++ b/operator/main.go
@@ -121,7 +121,7 @@ func run() error {
 		if err != nil {
 			return errors.Wrapf(err, "unable to parse PROFILING_THRESHOLD set to '%s' as a float", thresholdS)
 		}
-		heapProfiler := profiling.NewHeapProfiler(thresholdF, uint64(memLimit.IntegerSetting()), heapDumpDir.Setting())
+		heapProfiler := profiling.NewHeapProfiler(thresholdF, uint64(memLimit.IntegerSetting()), heapDumpDir.Setting(), profiling.DefaultHeapProfilerBackoff)
 		ctx, cancelProfiler := context.WithCancel(context.Background())
 		go heapProfiler.DumpHeapOnThreshhold(ctx, time.Second)
 		defer cancelProfiler()
--- a/operator/main.go
+++ b/operator/main.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -47,12 +48,13 @@ import (
 )
 
 var (
-	setupLog        = ctrl.Log.WithName("setup")
-	scheme          = runtime.NewScheme()
-	enableWebhooks  = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
-	enableProfiling = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
-	memLimit        = env.RegisterIntegerSetting("MEMORY_LIMIT_BYTES", 0)
-	heapDumpDir     = env.RegisterSetting("HEAP_DUMP_DIR", env.AllowEmpty())
+	setupLog           = ctrl.Log.WithName("setup")
+	scheme             = runtime.NewScheme()
+	enableWebhooks     = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
+	enableProfiling    = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
+	profilingThreshold = env.RegisterSetting("PROFILING_THRESHOLD", env.WithDefault("0.8"))
+	memLimit           = env.RegisterIntegerSetting("MEMORY_LIMIT_BYTES", 0)
+	heapDumpDir        = env.RegisterSetting("HEAP_DUMP_DIR", env.AllowEmpty())
 	// Default place where controller-runtime looks for TLS artifacts.
 	// see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/webhook/server.go#L96-L104
 	defaultCertDir  = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
@@ -112,7 +114,12 @@ func run() error {
 	}
 
 	if enableProfiling.BooleanSetting() {
-		heapProfiler := profiling.NewHeapProfiler(0.80, uint64(memLimit.IntegerSetting()), heapDumpDir.EnvVar())
+		thresholdS := profilingThreshold.EnvVar()
+		thresholdF, err := strconv.ParseFloat(thresholdS, 32)
+		if err != nil {
+			return errors.Wrapf(err, "unable to parse PROFILING_THREASHOLD set to '%s' as a float", thresholdString)
+		}
+		heapProfiler := profiling.NewHeapProfiler(thresholdF, uint64(memLimit.IntegerSetting()), heapDumpDir.EnvVar())
 		ctx, cancelProfiler := context.WithCancel(context.Background())
 		heapProfiler.DumpHeapOnThreshhold(ctx, time.Second)
 		defer cancelProfiler()
--- a/operator/main.go
+++ b/operator/main.go
@@ -54,7 +54,9 @@ var (
 	enableProfiling    = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
 	profilingThreshold = env.RegisterSetting("PROFILING_THRESHOLD", env.WithDefault("0.8"))
 	memLimit           = env.RegisterIntegerSetting("MEMORY_LIMIT_BYTES", 0)
-	heapDumpDir        = env.RegisterSetting("HEAP_DUMP_DIR", env.AllowEmpty())
+	// Default place to put the heap dump is the /tmp directory because the container process has rights
+	// to write to this directory without creating and mounting a PVC
+	heapDumpDir = env.RegisterSetting("HEAP_DUMP_DIR", env.WithDefault("/tmp"))
 	// Default place where controller-runtime looks for TLS artifacts.
 	// see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/webhook/server.go#L96-L104
 	defaultCertDir  = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
--- a/operator/main.go
+++ b/operator/main.go
@@ -44,9 +44,10 @@ import (
 )
 
 var (
-	setupLog       = ctrl.Log.WithName("setup")
-	scheme         = runtime.NewScheme()
-	enableWebhooks = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
+	setupLog        = ctrl.Log.WithName("setup")
+	scheme          = runtime.NewScheme()
+	enableWebhooks  = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
+	enableProfiling = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
 
 	// Default place where controller-runtime looks for TLS artifacts.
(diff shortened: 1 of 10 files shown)


porridge (commented):
To quickly check it works at all, just `make -C operator ENABLE_PROFILING=1 run` and launch `go profile` against it manually from another terminal window. That would be enough for this PR I think.

Then the next step would probably be to enable profiling [in the subscription](https://github.com/stackrox/stackrox/blob/master/operator/hack/operator.envsubst.yaml#L29) (it should not be a problem to run all our tests with profiling enabled) and then throw in a `50` step into the [upgrade test](https://github.com/stackrox/stackrox/tree/master/operator/tests/upgrade/upgrade) that fetches a profile and makes sure it's a valid one.
Getting this to work would also serve as a useful starting point for how to actually do this on a customer's cluster.

pkg/debug/debug.go
```diff
@@ -1,10 +1,75 @@
...
 )
 
 // FreeOSMemory runs a GC and then tries to relinquish as much memory back to the OS as possible
 func FreeOSMemory() {
 	debug.FreeOSMemory()
 }
+
+type HeapProfiler struct {
+	Threshold float64
+	Limit     uint64
```
porridge, line 20:
```suggestion
	ThresholdFraction float64
	LimitBytes     uint64
```
Or is it KiB?

pkg/debug/debug.go
```diff
@@ -1,10 +1,75 @@
...
+	}
+}
+
+func writeHeapProfile(t time.Time, dir string) error {
+	path := dumpFilePath(t, dir)
+	file, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
```
porridge, line 69: Please (debug?)log the fact that we are writing a profile.

pkg/debug/debug.go
```diff
@@ -1,10 +1,75 @@
...
+	return pprof.Lookup("heap").WriteTo(file, 0)
+}
+
+func dumpFilePath(t time.Time, dir string) string {
+	return fmt.Sprintf("%s/%s.dump", dir, t.Format("20060102T15-04-05"))
```
porridge, line 74: Not that it matters for us, but perhaps it's better to use a portable way to join dir and basename?

pkg/debug/debug_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestHeapDump(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	p := NewHeapProfiler(0.80, 2, tmpDir)
```
porridge, line 14:
```suggestion
	limitBytes := 2  // to be sure the test blows this limit
	p := NewHeapProfiler(0.80, limitBytes, tmpDir)
```

porridge (commented):
This looks promising overall, but I think we also need an upper bound on the size of the destination directory, either as bytes or number of files, and GC it on a FIFO basis, to make sure this keeps working if a process rides on the edge for a longer while.

To keep concerns  separate, I wonder if we could use a [filesystem abstraction](https://pkg.go.dev/io/fs) that would take care of this without tangling this issue with the rest of the code here.

pkg/debug/debug.go
johannes94: I added debug logging at the dumpHeapOnThreshhold method

pkg/debug/debug.go
johannes94: I changed this to use path.Join instead

pkg/debug/debug_test.go
johannes94: added the comment, thanks

johannes94:
> This looks promising overall, but I think we also need an upper bound on the size of the destination directory, either as bytes or number of files, and GC it on a FIFO basis, to make sure this keeps working if a process rides on the edge for a longer while.
> 
> To keep concerns separate, I wonder if we could use a [filesystem abstraction](https://pkg.go.dev/io/fs) that would take care of this without tangling this issue with the rest of the code here.

Thanks for bringing this up, I'll take a look if I can find something suitable to solve this.

pkg/profiling/profiling.go
```diff
@@ -0,0 +1,149 @@
...
+		for i, e := range entries {
+			oldestEntryInfo := entries[oldestEntryIndex]
+
+			if e.ModTime().Before(oldestEntryInfo.ModTime()) {
+				oldestEntryIndex = i
+			}
+		}
+
+		rmPath := path.Join(fd.dirPath, entries[oldestEntryIndex].Name())
+		os.Remove(rmPath)
```
porridge, line 57: If I'm reading this right, this only removes a single file. I think we should keep deleting the oldest one until we get below `fd.maxFileCount`.

pkg/profiling/profiling.go
```diff
@@ -0,0 +1,149 @@
...
+var (
+	log = logging.LoggerForModule()
+)
+
+const (
+	fifoDefaultMaxFileCount = 10
+	heapdumpSubfolderName   = "heapdump"
+)
+
+type fifoDir struct {
```
porridge, line 27: Looks like this object+method could go into a separate source file (probably too small for whole a separate package though).

pkg/profiling/profiling_test.go
```diff
@@ -0,0 +1,79 @@
...
+	// This should be bigger than maxFileCount to test FIFO deletion is done properly
+	numFilesToCreate := 10
+	for i := 0; i < numFilesToCreate; i++ {
+		fs.Create(fmt.Sprintf("%d.test.dump", i))
+	}
+
+	actualFiles, err := ioutil.ReadDir(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
```
porridge, line 30: I think we typically use an expectation library elsewhere to ensure no errors..

pkg/profiling/profiling_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if len(actualFiles) != maxFileCount {
+		t.Fatalf("expected count of files in test directory: %d, got %d", maxFileCount, len(actualFiles))
```
porridge, line 33: Ditto for comparisons.

pkg/profiling/profiling_test.go
```diff
@@ -0,0 +1,79 @@
...
+	expectedIndex := numFilesToCreate - maxFileCount
+	for _, f := range actualFiles {
+		parts := strings.Split(f.Name(), ".")
+		if len(parts) < 1 {
+			t.Fatal("expected filename to have index prefix seperated by '.'")
+		}
+
+		if fmt.Sprintf("%d", expectedIndex) != parts[0] {
+			t.Fatalf("expected index of file: %s to be: %d", f.Name(), expectedIndex)
+		}
```
porridge, line 49: This seems quite complex and does not actually check that the right files are deleted. Perhaps just pre-generate the file names at the beginning of the test, then create them, and then compare  sorted list to a sub-slice of the generated names?

pkg/profiling/profiling_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func Test_fifoDir(t *testing.T) {
+	dir := t.TempDir()
+	maxFileCount := 3
+	fs := fifoDir{dirPath: dir, maxFileCount: maxFileCount}
+
+	// This should be bigger than maxFileCount to test FIFO deletion is done properly
+	numFilesToCreate := 10
+	for i := 0; i < numFilesToCreate; i++ {
+		fs.Create(fmt.Sprintf("%d.test.dump", i))
```
porridge, line 24: Depending on the time resolution of underlying FS, you might have to inject a tiny delay here for the test to not flake.

porridge (changes requested):
This is excellent, but I have a bunch of nitpicks inline.

pkg/profiling/profiling_test.go
johannes94: Changed it to contains logic, which is in fact much simpler when using testify assert.

pkg/profiling/profiling_test.go
johannes94: Done.

pkg/profiling/profiling_test.go
johannes94: Done.

pkg/profiling/profiling.go
johannes94: You're right good catch. This only works correctly when starting with an empty directory. I changed it to sorting, which fortunately also made the code more simple :)

operator/main.go
```diff
@@ -44,10 +48,15 @@ import (
...
-
+	setupLog           = ctrl.Log.WithName("setup")
+	scheme             = runtime.NewScheme()
+	enableWebhooks     = env.RegisterBooleanSetting("ENABLE_WEBHOOKS", true)
+	enableProfiling    = env.RegisterBooleanSetting("ENABLE_PROFILING", false)
+	profilingThreshold = env.RegisterSetting("PROFILING_THRESHOLD", env.WithDefault("0.8"))
+	memLimit           = env.RegisterIntegerSetting("MEMORY_LIMIT_BYTES", 0)
+	// Default place to put the heap dump is the /tmp directory because the container process has rights
+	// to write to this directory without creating and mounting a PVC
+	heapDumpDir = env.RegisterSetting("HEAP_DUMP_DIR", env.WithDefault("/tmp"))
```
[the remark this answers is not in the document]
porridge: Did you perhaps forgot to push some changes? I still see:
```
heapDumpDir = env.RegisterSetting("HEAP_DUMP_DIR", env.WithDefault("/tmp"))
```

pkg/profiling/fifo_dir.go
```diff
@@ -0,0 +1,73 @@
...
+		return entryInfos[i].ModTime().Before(entryInfos[j].ModTime())
+	})
+
+	for len(entryInfos) >= fd.maxFileCount {
+		rmPath := path.Join(fd.dirPath, entryInfos[0].Name())
+		err := os.Remove(rmPath)
+		if err != nil {
+			return nil, errors.Wrapf(err, "removing file: %s", rmPath)
+		}
+		entryInfos = entryInfos[1:]
```
porridge, line 49: Wouldn't it be more straightforward to compute the "to delete" sub-slice in one go, and the just iterate over it, rather than do this re-slicing? :thinking:

pkg/profiling/fifo_dir_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	var filesToBeDeleted []string
+	var filesToBeKept []string
+	// This should be bigger than maxFileCount to test FIFO deletion is done properly
+	numFilesToCreate := 10
+	for i := 0; i < numFilesToCreate; i++ {
+		fileName := fmt.Sprintf("%d.test.dump", i)
+		_, err := fs.Create(fileName)
+		require.NoError(t, err, "creating file: %s", fileName)
+		if i < numFilesToCreate-3 {
```
porridge, line 26:
```suggestion
		if i < numFilesToCreate-maxFileCount {
```

But personally I would try to make this test easier to read by just statically initializing `filesToBeDeleted` and `filesToBeKept` using literals, rather than generate the names in a loop, since it's so easy to have an off-by-one in such loop. Also we probably don't need to test with 10 files if limit is 3. IMHO 5 would be fine too.

pkg/profiling/fifo_dir_test.go
```diff
@@ -0,0 +1,52 @@
...
+			filesToBeDeleted = append(filesToBeDeleted, fileName)
+		} else {
+			filesToBeKept = append(filesToBeKept, fileName)
+		}
+		time.Sleep(time.Millisecond * 100)
+	}
+
+	actualFilesEntries, err := os.ReadDir(dir)
+	require.NoErrorf(t, err, "reading directory %s", dir)
+	actualFiles, err := dirEntriesToFileInfo(actualFilesEntries)
```
porridge, line 36: This function seems to serve both as code under test and test code, which some people might call questionable :wink:

pkg/profiling/profiling.go
```diff
@@ -0,0 +1,94 @@
...
+	heapdumpSubfolderName = "heapdump"
+	// time to wait between heap dumps when hitting threshold in seconds
+	defaultHeapProfilerBackoff = 30
+)
+
+// HeapProfiler is used to start a ticker that periodically checks if heap memory consumption
+// exceed the thresholdFraction, if so the heap gets dumped to a File in Directory.
+type HeapProfiler struct {
+	// Backoff limits the maximum frequency of creating heap dumps
+	Backoff           time.Duration
```
porridge, line 26: I'm a little lost, why would we want to tweak the `Backoff` field on a live `HeapProfiler` object from a different package? :thinking:

pkg/profiling/profiling.go
```diff
@@ -0,0 +1,94 @@
...
+
+var (
+	log                   = logging.LoggerForModule()
+	heapdumpSubfolderName = "heapdump"
+	// time to wait between heap dumps when hitting threshold in seconds
+	defaultHeapProfilerBackoff = 30
+)
+
+// HeapProfiler is used to start a ticker that periodically checks if heap memory consumption
+// exceed the thresholdFraction, if so the heap gets dumped to a File in Directory.
```
porridge, line 23:
```suggestion
// exceed the thresholdFraction, if so the heap gets dumped to a file in directory, subject to Backoff.
```

pkg/profiling/profiling.go
johannes94: If I'd use a HeapProfiler from another package in my program, I would want this to be configurable from the caller site e.g. with env variables for the binary. Let's say we want to analyze a binary but it dies within 40 seconds because of OOM, I'd want to be able to set the backoff to 10 secs to get multiple dumps in this timeframe. IMO we have 3 ways to allow that:

1. Extend the NewHeapProfiler function with a backoff parameter
       - we can add it here, so that the whole config is done with the NewHeapProfiler and we don't mix different types of field access here
2. Make the field public 
       - We have a good default, so why should we extend the NewHeapProfiler function, this makes it more complicated, just to support a use case that is probably not used very often. That's why I took this option
3. Use Setter/Getter for this
       - I personally don't like getter/setter just for the purpose of making a field private

operator/main.go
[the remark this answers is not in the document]
johannes94: Yes, I did :)

pkg/profiling/fifo_dir.go
porridge: It does to me, but meh :shrug:

pkg/profiling/fifo_dir_test.go
```diff
@@ -0,0 +1,46 @@
...
+	fs := fifoDir{dirPath: dir, maxFileCount: maxFileCount}
+
+	// This should be more files than maxFileCount to test FIFO deletion is done properly
+	filesToCreate := []string{"1.test.dump", "2.test.dump", "3.test.dump", "4.test.dump", "5.test.dump"}
+	filesToBeDeleted := filesToCreate[:2]
+	filesToBeKept := filesToCreate[2:]
+
+	for _, fileName := range filesToCreate {
+		_, err := fs.Create(fileName)
+		time.Sleep(time.Millisecond * 100)
```
porridge, line 24:
```suggestion
		time.Sleep(time.Millisecond * 100) // to prevent flakes due to inconsistent ordering in case FS time resolution is low
```

pkg/profiling/profiling.go
porridge: I'd definitely go for (1), if only not to have to worry about thread-safety issues.

porridge (approved):

PR #3913, opened 2022-11-25
ROX-12948 Disable DeploymentExposure tests

lvalerom:
## Description

This will disable the `DeploymentExposure` tests until we refactor the `helper.go` to enable it to use fake k8s clients.

**Context**: These tests are flaky due to ports not being released on time for the next test to run. We decided to use a fake k8s client to solve this issue but, since mayor refactors are needed in `helper.go`, we will disable the test to avoid failures in CI in the meantime.

## Checklist

## Testing Performed

* CI
* Manual testing shows the tests are skipped

### DIFF
--- a/sensor/tests/resource/service/service_test.go
+++ b/sensor/tests/resource/service/service_test.go
@@ -117,6 +117,7 @@ type DeploymentExposureSuite struct {
 }
 
 func Test_DeploymentExposure(t *testing.T) {
+	t.Skip("Disabling these tests until we refactor the helper.go to enable fake k8s clients (ROX-13644)")
 	suite.Run(t, new(DeploymentExposureSuite))
 }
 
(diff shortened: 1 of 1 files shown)


sensor/tests/resource/service/service_test.go
```diff
@@ -117,6 +117,7 @@ type DeploymentExposureSuite struct {
 }
 
 func Test_DeploymentExposure(t *testing.T) {
+	t.Skip("Disabling these tests until we refactor the helper.go to enable fake k8s clients (ROX-13644)")
```
msugakov, line 120: Please also add a todo note in our standard format.

```suggestion
	// TODO(ROX-13644): reenable the test
	t.Skip("Disabling these tests until we refactor the helper.go to enable fake k8s clients (ROX-13644)")
```

msugakov (approved):

sensor/tests/resource/service/service_test.go
lvalerom: Done

PR #3924, opened 2022-11-28
ROX-13596: Determine user-agent in gRPC-gateway and HTTP

ivan-degtiarenko:
## Description

This is fix to telemetry interceptors to correctly determine `User-Agent`.

## Checklist

## Testing Performed

1. UI calls are tracked with browser `User-Agent`
2. `roxctl` is tracked with `roxctl/<version` user agent
3. `/v1/ping` is tracked with `kube-probe` user agent


central/telemetry/phonehome/interceptor.go
```diff
@@ -60,6 +62,19 @@ func getRequestDetails(ctx context.Context, err error, info *grpc.UnaryServerInf
...
+		// https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L106-L114
+		// User-Agent header is occupied with internal grpc-go value:
+		// https://github.com/grpc/grpc-go/blob/0238b6e1cec37b55820b461d3d30652c54efe2c4/clientconn.go#L211-L215
+		userAgentKey := fmt.Sprintf("%suser-agent", runtime.MetadataPrefix)
+		userAgent = strings.Join(ri.Metadata.Get(userAgentKey), ", ")
+	}
+	// Fallback to HTTP
+	if !throughGrpcGateway || userAgent == "" {
+		userAgent = strings.Join(ri.Metadata.Get("User-Agent"), ", ")
+	}
```
0x656b694d, line 77: May we just try to find the `grpcgatewayuser-agent` key and use it if found or fallback to `User-Agent` otherwise?

ivan-degtiarenko: I think the code here has clearer intent: if `grpc-gateway` is used, use `grpc-gateway` header; if it does not work **or** we are not on gRPC protocol, use HTTP header. It would be unexpected to see `grpcgateway-user-agent` in the HTTP interceptor, and I would think we should ignore such header in such context.

0x656b694d: I think there is too much code for this intent: you even had to add more parameters to two functions and compute the normal User-Agent twice.
Instead, you could move this grpc related check (i.e. try to read the `grpcgatewayuser-agent` header, and you don't need any bool parameter) to the beginning of the function, and put the existing User-Agent retrieval under `else`.

central/telemetry/phonehome/interceptor.go
```diff
@@ -60,6 +62,19 @@ func getRequestDetails(ctx context.Context, err error, info *grpc.UnaryServerInf
...
+		// By default, all permanent HTTP headers are added grpcgateway- prefix:
+		// https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L106-L114
+		// User-Agent header is occupied with internal grpc-go value:
+		// https://github.com/grpc/grpc-go/blob/0238b6e1cec37b55820b461d3d30652c54efe2c4/clientconn.go#L211-L215
+		userAgentKey := fmt.Sprintf("%suser-agent", runtime.MetadataPrefix)
+		userAgent = strings.Join(ri.Metadata.Get(userAgentKey), ", ")
+	}
+	// Fallback to HTTP
+	if !throughGrpcGateway || userAgent == "" {
+		userAgent = strings.Join(ri.Metadata.Get("User-Agent"), ", ")
```
rukletsov, line 76: FWIW, with such branching you don't need `userAgent` declaration at L43. If you keep the current structure, pull L43 right before branching and make it `var userAgent string`. However, I'm in favor of Michaël's suggestion below.

central/telemetry/phonehome/interceptor.go
rukletsov: ➕

central/telemetry/phonehome/interceptor.go
0x656b694d: :heavy_plus_sign:

central/telemetry/phonehome/interceptor.go
```diff
@@ -17,6 +18,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+const grpcGatewayUserAgentHeader = runtime.MetadataPrefix + "user-agent"
```
rukletsov, line 21:
```suggestion
const grpcGatewayUserAgentHeader = runtime.MetadataPrefix + "User-Agent"
```
I understand that `metadata.MD` converts keys to lowercase but this is not obvious here, is not common knowledge, and who knows when it changes. [Canonical form](https://github.com/grpc-ecosystem/grpc-gateway/blob/8952e38d5addd28308e29c272c696a578aa8ace8/runtime/context.go#L288) is "User-Agent".

central/telemetry/phonehome/interceptor.go
```diff
@@ -60,6 +61,16 @@ func getRequestDetails(ctx context.Context, err error, info *grpc.UnaryServerInf
 		method = "unknown"
 		code = -1
 	}
+
+	// By default, all permanent HTTP headers in grpc-gateway are added grpcgateway- prefix:
+	// https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L106-L114
```
rukletsov, line 66:
```suggestion
	// https://github.com/grpc-ecosystem/grpc-gateway/blob/8952e38d5addd28308e29c272c696a578aa8ace8/runtime/mux.go#L106-L114
```
Always press `y` in GitHub to get a link with pinned line references.

central/telemetry/phonehome/interceptor.go
```diff
@@ -60,6 +61,16 @@ func getRequestDetails(ctx context.Context, err error, info *grpc.UnaryServerInf
...
+	// User-Agent header is occupied with internal grpc-go value:
+	// https://github.com/grpc/grpc-go/blob/0238b6e1cec37b55820b461d3d30652c54efe2c4/clientconn.go#L211-L215
```
rukletsov, line 68: It's actually not clear to me what this code reference proves. `cc.dopts.copts.UserAgent += " " + grpcUA` hints that `User-Agent` value is not overwritten but appended to, so it will be something like `ACS roxctl user agent blah grpc-go/1.50.1`.  Which sort of contradicts what's claimed in ROX-13596.

Could you please split the "Testing Performed" section into before / after?

rukletsov (changes requested):

central/telemetry/phonehome/interceptor.go
ivan-degtiarenko: Regarding `cc.dopts.copts.UserAgent`: this shows where the  user agent's value `grpc-go/1.51.0` is initially taken from. 

Answering the question of why there's no other `user-agent` value: 
1. To transform from HTTP to GRPC request, we connect to our gRPC server as a client from within the central
https://github.com/stackrox/stackrox/blob/0529062eb7396c82871e5f59305108638c7b27c0/pkg/grpc/server.go#L352-L353
https://github.com/stackrox/stackrox/blob/0529062eb7396c82871e5f59305108638c7b27c0/pkg/grpc/server.go#L311-L317
2. Connection for this client is created once during central initialization, thus we cannot override it on every request.
3. Value for `user-agent` dial option is not specified in our code, thus we use the default
https://github.com/stackrox/stackrox/blob/0529062eb7396c82871e5f59305108638c7b27c0/pkg/grpc/server.go#L226-L233

rukletsov: Ah, I see, didn't know that we persist the connection — thanks. This makes sense.

It does *not* make sense to capture your reply as a code comment because it builds on non-local knowledge which might change any time and render the comment misleading, but it *does* make sense to explain that in the PR description.

One last thing: do you want to follow up and augment grpc server init code with User-Agent for HTTP muxer?

rukletsov (approved):
Thanks, Ivan. Please update the PR description per my recent comment and 🚢 it.

central/telemetry/phonehome/interceptor.go
ivan-degtiarenko: > One last thing: do you want to follow up and augment grpc server init code with User-Agent for HTTP muxer?

If I understand correctly what you suggest, that would add some additional `user-agent` value to all requests going through the muxer. Essentially, what @0x656b694d did here:
https://github.com/stackrox/stackrox/pull/3672/files#diff-af62cb3a3a7c777bd1b8feed18657c3ecae951d2dbdea55527f08c8c67967c71R233

I think it works as a fallback option. It's better than displaying `grpc-go/1.51.0`, so why not?

I had a small concern regarding introducing more variety to potential `user-agent` values in Segment, but Segment supports `contains` filter, so that shouldn't be a problem.

rukletsov: Yeah, essentially what's done in https://github.com/stackrox/stackrox/pull/3672, so my comment should be irrelevant. Why do we still see `"User-Agent": "grpc-go/1.51.0"` in the Before section then?

pkg/telemetry/phonehome/interceptor.go
```diff
@@ -39,24 +42,36 @@ func getGrpcRequestDetails(ctx context.Context, err error, info *grpc.UnaryServe
...
+	// By default, all permanent HTTP headers in grpc-gateway are added grpcgateway- prefix:
+	// https://github.com/grpc-ecosystem/grpc-gateway/blob/8952e38d5addd28308e29c272c696a578aa8ace8/runtime/mux.go#L106-L114
+	// User-Agent header is occupied with internal grpc-go value:
+	// https://github.com/grpc/grpc-go/blob/0238b6e1cec37b55820b461d3d30652c54efe2c4/clientconn.go#L211-L215
+	userAgentValues := ri.Metadata.Get(grpcGatewayUserAgentHeader)
+	// If endpoint is accessed not via grpc-gateway, extract from User-Agent header.
+	// If endpoint is accessed via grpc-gateway, append grpc-go value to the resultinguser agent.
+	userAgentValues = append(userAgentValues, ri.Metadata.Get("User-Agent")...)
+	userAgent := strings.Join(userAgentValues, " ")
+	return userAgent
```
0x656b694d, line 64:
```suggestion
	return strings.Join(userAgentValues, " ")
```

0x656b694d (approved):
LGTM

PR #3925, opened 2022-11-28
Add visitAndAssertBeforeResponses helper function for integration tests (part 1)

pedrottimark:
### Problem

Under pressure to solve timing failures in tests for various containers:

1. I replaced `cy.visit` method with generic `visit` function which intercepts requests and waits for responses according to `routeMatcherMapForAuthenticatedRoutes` object.

    * That worked well, with one exception: I incorrect includes credential expiry requests, which are not in sequence preceding, but instead in parallel with container-specific requests. Fixed in #3577

2. I wrote container-specific `visitWhatever` functions with `routeMatcherMapForWhatever` object and optional `staticResponseMap` argument.

    * However, I made a mistake in the pattern to call `visit` function with those arguments, **and then** assert page address or page header. Destination page address is an obvious prerequisite to requests and components usually render page header **before** instead of after responses to requests. Because the order of wait (API, API, and then DOM) contradicts the order of events, the pattern wastes the extra 4 seconds wait for DOM assertions.

    * Especially if containers or entity components take responsibility to make requests, but even if they render data from Redux store via sagas, it is possible to assert page address or page header **and then** intercept requests and wait for responses.

### Solution

This helper function adds another step to the pattern for interact and visit in #3917

To prevent timing problems if central slows down, invest cypress timeout in a more effective pattern which **alternates** API and DOM:

1. API: 10 seconds wait for requests and 30 seconds wait for responses in `routeMatcherForAuthenticatedRoutes` object.
2. DOM: 4 seconds wait for **container-specific** assertions in `assertionCallback` function.
3. API: 10 seconds wait for requests and 30 seconds wait for responses for **container-specific** `routeMatcherMapForWhatever` object.
4. DOM: 4 seconds wait for **test-specific** assertions after `visitWhatever` function call.

In part 1, apply the pattern to the following containers:
1. workflow
    * compliance
    * configmanagement
    * vulnmanagement
2. policies
3. violations

**Review recommendation**: before you review the test files, look at cypress/helpers/visit.js file to compare improved `visitAndAssertBeforeResponses` with baseline `visit` function (expand the following lines).

### Residue

1. cypress/helpers/policies.js
    * Replace inline intercept and wait with routeMatcherMap and helper functions.
2. cypress/integration/compliance/complianceList.test.js
    * Encapsulate reference to container path in tests.
3. cypress/integration/policies/policiesTable.test.js
    * Investigate how to encapsulate reference to policies path in 2 redirect tests.
4. cypress/integration/violations/tags.test.js
    * Encapsulate reference to container path in skipped tests.

## Checklist

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/helpers/compliance.js
+++ b/ui/apps/platform/cypress/helpers/compliance.js
@@ -1,7 +1,7 @@
 import { headingPlural, selectors, url } from '../constants/CompliancePage';
 
 import { getRouteMatcherMapForGraphQL, interactAndWaitForResponses } from './request';
-import { visit } from './visit';
+import { visitAndAssertBeforeResponses } from './visit';
 
 const routeMatcherMapForComplianceDashboard = getRouteMatcherMapForGraphQL([
     'clustersCount',
@@ -22,10 +22,16 @@ const routeMatcherMapForComplianceDashboard = getRouteMatcherMapForGraphQL([
     'complianceStandards_PCI_DSS_3_2',
 ]);
 
-export function visitComplianceDashboard() {
-    visit(url.dashboard, routeMatcherMapForComplianceDashboard);
+const dashboardTitle = 'Compliance';
 
-    cy.get('h1:contains("Compliance")');
+export function visitComplianceDashboard() {
+    visitAndAssertBeforeResponses(
+        url.dashboard,
+        () => {
+            cy.get(`h1:contains("${dashboardTitle}")`);
+        },
+        routeMatcherMapForComplianceDashboard
+    );
 }
 
 /*
@@ -75,9 +81,13 @@ export function visitComplianceEntities(entitiesKey) {
         opnameForEntities[entitiesKey],
     ]);
 
-    visit(url.entities[entitiesKey], routeMatcherMap);
-
-    cy.get(`h1:contains("${headingPlural[entitiesKey]}")`);
+    visitAndAssertBeforeResponses(
+        url.entities[entitiesKey],
+        () => {
+            cy.get(`h1:contains("${headingPlural[entitiesKey]}")`);
+        },
+        routeMatcherMap
+    );
 }
 
 /*
@@ -90,7 +100,11 @@ export function visitComplianceStandard(standardName) {
         'controls',
     ]);
 
-    visit(`${url.controls}?s[standard]=${standardName}`, routeMatcherMap);
-
-    cy.get(`h1:contains("${standardName}")`);
+    visitAndAssertBeforeResponses(
+        `${url.controls}?s[standard]=${standardName}`,
+        () => {
+            cy.get(`h1:contains("${standardName}")`);
+        },
+        routeMatcherMap
+    );
 }
--- a/ui/apps/platform/cypress/helpers/compliance.js
+++ b/ui/apps/platform/cypress/helpers/compliance.js
@@ -22,13 +22,13 @@ const routeMatcherMapForComplianceDashboard = getRouteMatcherMapForGraphQL([
     'complianceStandards_PCI_DSS_3_2',
 ]);
 
-const dashboardTitle = 'Compliance';
+const containerTitle = 'Compliance';
 
 export function visitComplianceDashboard() {
     visitAndAssertBeforeResponses(
         url.dashboard,
         () => {
-            cy.get(`h1:contains("${dashboardTitle}")`);
+            cy.get(`h1:contains("${containerTitle}")`);
         },
         routeMatcherMapForComplianceDashboard
     );
--- a/ui/apps/platform/cypress/helpers/configWorkflowUtils.js
+++ b/ui/apps/platform/cypress/helpers/configWorkflowUtils.js
@@ -144,13 +144,13 @@ const routeMatcherMapForConfigurationManagementDashboard = getRouteMatcherMapFor
     'secrets',
 ]);
 
-const dashboardTitle = 'Configuration Management';
+const containerTitle = 'Configuration Management';
 
 export function visitConfigurationManagementDashboard() {
     visitAndAssertBeforeResponses(
         basePath,
         () => {
-            cy.get(`h1:contains("h1:contains("${dashboardTitle}")")`);
+            cy.get(`h1:contains("h1:contains("${containerTitle}")")`);
         },
         routeMatcherMapForConfigurationManagementDashboard
     );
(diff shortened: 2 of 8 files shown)


pedrottimark:
Although changed files include Configuration Management, the integration test failure is unrelated:

> Timed out retrying after 4000ms: Expected to find element: `.rt-td:nth-child(4):contains("fail"):nth(0)`, but never found it.

cypress/integration/configmanagement/ciscontrols.test.js:83:11

dvail (commented):
For my understanding, what are the timing problems that this change is looking to solve? Does the Central slow-down result in race conditions where we are making assertions that will never be true, given the ordering of network requests and user actions? Or does the slow down result in the test executing beyond the Cypress timeouts and causing a flake?

pedrottimark:
Yes, that is the question. Here is a more complete analysis as I reviewed my thinking yesterday:
* **Pro** of original pattern: It was straightforward to factor out cypress intercept-visit-wait sandwich in tests as container-level helper functions which call `visit(path, routeMatcherMap, staticResponseMap)` **and then** make container-level assertions. That is, maximum return on investment to prevent timing failures throughout the tests.
* **Con** of original pattern: The container-level assertions are **after** responses to container-level requests, instead of the actual order before (or maybe more precisely, in parallel with) the requests. Therefore, it misleads readers and writers of tests and allows mixing assertions which do or do not depend on data from responses. It indicated my fuzzy thinking at the time and had my thinking been clearer, this pattern would have been only an additional increment of effort.

Using one or more DOM assertions to separate requests for authenticated routes from requests for container pages **might be a prerequisite** to future changes.

The most recent timeouts have been with the requests for **authenticated routes** instead of requests for container pages.
* One possible solution **does not** depend on this proposed change: increase `requestTimeout` for `routeMatcherMapForAuthenticatedRoutes` in visit.js file.
* Another possible solution **does** depend on this proposed change: if any more requests (in addition to credentialexpiry) in `routeMatcherMapForAuthenticatedRoutes` are not prerequisites to rendering pages, then remove them and let the container-level DOM assertions become the prerequisite to start waiting for container-level requests.

By the way, I was concerned how well this pattern will compose with other variations of visit helper functions:
* visit from **left nav**: It seems like a similar incremental update, because the corresponding tests are so limited.
* visit with static response for auth status or especially for **permissions**: Because we have not written many tests related to `'READ_ACCESS'` versus `'READ_WRITE_ACCESS'` I have not yet looked ahead to this as it deserves.

pedrottimark:
Aha, super example!

Container-specific helper functions encapsulate this issue so readers and writers **of tests** do not think about it.

Here are some factors for container-specific requests and assertions:
* It was important to wait for responses to prevent premature evaluation of jQuery-style method chains in assertions.
* Even for `:contains("whatever")` assertions which cypress can retry, there is a subtle difference: timeout for DOM assertion is 4 seconds but timeout for request is now 10 seconds.
* In rare cases where there is significant frontend computation (for example, Network Graph) the extra 4 seconds to render after the 10 seconds for response might be significant.

Test-writing idioms to solve the above problems include:
* Container-specific helper functions call `interactAndWaitForResponses` with relevant requests.
* Source container tests call `interactAndVisitWhatever` import from destination container which encapsulates requests.

The proposed change affects helper functions like `visitVulnerabilityManagementEntities` that usually make an assertion like the following:

```js
cy.get(`h1:contains("${headingPlural[entitiesKey]}")`);
```

That is, it is the correct page (independent of data).

Therefore, the audience is readers and writers of container-specific helper functions. Same audience, us of course, but an analogy is **at what point** in development process we care about **implementation** of hooks versus **use** of hooks.

The proposed change addresses what I see at this point as two remaining problems:
1. Conceptual problem for readers and writers of container-specific helper functions to understand the order of events: when does initial container-specific **rendering** occur compared to container-specific **requests**. Initial rendering is in parallel with requests, therefore more accurate to assert DOM elements from initial rendering **before requests** instead of **after responses**. This is a testing motivation for the discussion about pro and con to render **Collection** breadcrumb in parallel with spinner.
2. Timing problems which do not yet have known causes might have solutions which need to distinguish (via `assertionCallback` function) assertions which **do not** depend on container-specific responses from test-specific assertions which **do** depend on responses. If we build that distinction into the emerging pattern how the intermediate layer of container-specific helper function call the inner layer of generic helper functions, then the latter have somewhat more information to make changes which might benefit all tests.

pedrottimark:
> it isn't immediately apparent to me where a central slowdown in CI would cause a flake

Dave, here is something that I forgot to include in previous comments. The test visits the following pages:
1. clusters list: click fixable link to visit…
2. cluster **fixable** cves list in side panel: click left arrow to visit…
3. cluster in side panel: click related entities cves link to visit…
4. cluster **all** cves list in side panel: see `'related entities tile links should unset search params upon navigation'`

After the first step, robust DOM assertions might be enough, especially in local deployment.

The patterns in progress are intended to be robust for less than ideal characteristics of tests:
* Simple tests, especially if they assert on elements which do **not** depend on responses, sometimes succeed **before** responses to all requests. Without wait for responses, this has been a problem when the next test starts and attempts to **mock the response** to a request before the response from the corresponding request from the previous test.
* Every vulnmanagement test file for a specific entity called an original helper function that started from step 1 for each successive step: 1, 1-2, 1-2-3, 1-2-3-4, 1-2-3-4-5 (that is, quadratic instead of linear time complexity) and had only the last DOM assertion for each sequence. After recent work had put two such long-running tests in succession within a new test file for two entities, it suddenly started to fail with timeout in 2/3 of CI runs. Double the quadratic trouble! To stop the failures required an emergency rewrite of helper functions to visit pages 1-2-3-4-5 and also wait for the primary request of each page. Plus splitting the combined test file into two separate files.

And there are unknowns about CI environment in past, present, and future:
* Cypress and central seem to run more asynchronous in CI environment than on our computers.
* There are more CI runs in a day now than in a 2-week sprint 2 years ago. I think quantity is a reason why more failures on branch than merge and more on merge than on nightly. As engineers ironically referred to early electronic computers as huge vacuum tube test apparatus, we might think of CI as a fuzzer to find tests which have async weaknesses.
* Some intermittent failures are associated with longer-running tests and `Refreshing the GKE auth token` in build-log.txt file, which occurs every 15 minutes on CI.

pedrottimark:
Having dumped thoughts why the relationship between DOM asserts and request waits is **both-and** instead of **either-or**, the worthwhile discussion has brought me to the conclusion:
* the current visit pattern has proved its worth
* the proposed changes are hard to understand and the potential benefits might beyond point of diminishing return

dvail:
| Without wait for responses, this has been a problem when the next test starts and attempts to mock the response to a request before the response from the corresponding request from the previous test.

Oh no, this was something I was unaware of. 😬 

I suppose I thought Cypress automatically handled this type of thing but I can see how this could become a debugging nightmare, especially in an unpredictable environment. I've run into something similar in the ui-unit-tests, but with pending React state updates instead of network requests and it is incredibly time consuming to track down and resolve.

PR #3928, opened 2022-11-28
Update create-bundle.sh

0x656b694d:
## Description

Fix for: `Error: "/tmp/tmp.2GlLsngtaL/stackrox/static-data/" could not be found on the host: no such file or directory` happening on `make image`.

## Testing Performed

No error:
```sh
$ /usr/bin/env DEBUG_BUILD="no" /home/mipetrov/go/src/github.com/stackrox/stackrox/image/rhel/create-bundle.sh /home/mipetrov/go/src/github.com/stackrox/stackrox/image stackrox-data:3.73.x-120-g84989f7bb4-dirty quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.49 /home/mipetrov/go/src/github.com/stackrox/stackrox/image/rhel
bed13ada14d1876a544720ce2cd682125ea86bbb7f478cb433b969db088f2122
copier
df5f95f5b4e6271f66e800239e217c495efe4b6e4a1825b93fd3b761fa48674b
copier
```

### DIFF
--- a/image/rhel/create-bundle.sh
+++ b/image/rhel/create-bundle.sh
@@ -23,6 +23,7 @@ extract_from_image() {
   [[ -n "$image" && -n "$src" && -n "$dst" ]] \
       || die "extract_from_image: <image> <src> <dst>"
 
+  mkdir -p "${dst}"
   docker create --name copier "${image}"
   docker cp "copier:${src}" "${dst}"
   docker rm copier


image/rhel/create-bundle.sh
```diff
@@ -23,6 +23,7 @@ extract_from_image() {
   [[ -n "$image" && -n "$src" && -n "$dst" ]] \
       || die "extract_from_image: <image> <src> <dst>"
 
+  mkdir -p "${dst}"
```
vikin91, line 26: Do we have a guarantee that `dst` is a folder and not file?
- If yes, then this is fine.
- If no, then we must add code that would prevent creating a directory that would be named exactly as the file. Illustration:

We want to do `cat /tmp/path/file`

```shell
$ touch file
$ cp file /tmp/path/file
cp: /tmp/path/file: No such file or directory
# cannot copy, because /tmp/path is absent

# we fix that as in this PR
$  mkdir -p /tmp/path/file
# and copy again
$ cp file /tmp/path/file
# now it worked, but what is the result?
$ ls /tmp/path/file
file
$  cat /tmp/path/file
cat: /tmp/path/file: Is a director
# ups, that is not what we wanted
```

gavin-stackrox: `dst` can be a file e.g. the `ldb` case in that script.

gavin-stackrox (changes requested):

gavin-stackrox (approved):
LGTM

vikin91 (approved):

PR #3930, opened 2022-11-28
Use UUID in postgres unit tests

md2119:
## Description
We missed a few instances of IDs when migrating to UUID.

image

## Checklist

## Testing Performed
CI


dashrews78 (changes requested):
Looks like we (I) missed these because they are not executed as part of `go-postgres-unit-tests`.  We should probably upgrade that target in the `Makefile` to include these tests.

md2119:
> Looks like we (I) missed these because they are not executed as part of `go-postgres-unit-tests`. We should probably upgrade that target in the `Makefile` to include these tests.

Rather than upgrading the `Makefile`,  I renamed the test file per the original convention we put in place for postgres tests.

dashrews78 (approved):
LGTM

PR #3931, opened 2022-11-28
ROX-13661: Refactor `Stopper` and use it more

msugakov:
## Description

Refactor existing `Stopper` to be used in all places where we previously had `stopC` and `stoppedC`.
Induced by https://github.com/stackrox/stackrox/pull/3800

Note that I am not trying to fix things in this PR, just reduce the amount of different patterns. Some things look very questionable but I refrain from fixing them in the absence of unit tests.

It is best to review this PR by going through commits. This way the size isn't overwhelming.

## Checklist

No need in the following:

## Testing Performed

* Relying on CI.

### DIFF
--- a/central/pruning/pruning.go
+++ b/central/pruning/pruning.go
@@ -97,8 +97,7 @@ func newGarbageCollector(alerts alertDatastore.DataStore,
 		k8sRoles:        k8sRoles,
 		k8sRoleBindings: k8sRoleBindings,
 		logimbueStore:   logimbueStore,
-		stopSig:         concurrency.NewSignal(),
-		stoppedSig:      concurrency.NewSignal(),
+		stopper:         concurrency.NewStopper(),
 	}
 }
 
@@ -120,8 +119,7 @@ type garbageCollectorImpl struct {
 	k8sRoles        k8sRoleDataStore.DataStore
 	k8sRoleBindings roleBindingDataStore.DataStore
 	logimbueStore   logimbueDataStore.Store
-	stopSig         concurrency.Signal
-	stoppedSig      concurrency.Signal
+	stopper         concurrency.Stopper
 }
 
 func (g *garbageCollectorImpl) Start() {
@@ -159,6 +157,8 @@ func (g *garbageCollectorImpl) pruneBasedOnConfig() {
 }
 
 func (g *garbageCollectorImpl) runGC() {
+	defer g.stopper.Flow().ReportStopped()
+
 	lastClusterPruneTime = time.Now().Add(-24 * time.Hour)
 	lastLogImbuePruneTime = time.Now().Add(-24 * time.Hour)
 	g.pruneBasedOnConfig()
@@ -168,8 +168,7 @@ func (g *garbageCollectorImpl) runGC() {
 		select {
 		case <-t.C:
 			g.pruneBasedOnConfig()
-		case <-g.stopSig.Done():
-			g.stoppedSig.Signal()
+		case <-g.stopper.Flow().StopRequested():
 			return
 		}
 	}
@@ -876,6 +875,6 @@ func (g *garbageCollectorImpl) pruneLogImbues() {
 }
 
 func (g *garbageCollectorImpl) Stop() {
-	g.stopSig.Signal()
-	<-g.stoppedSig.Done()
+	g.stopper.Client().Stop()
+	_ = g.stopper.Client().Stopped().Wait()
 }
--- a/compliance/collection/auditlog/auditlog.go
+++ b/compliance/collection/auditlog/auditlog.go
@@ -18,15 +18,15 @@ type Reader interface {
 	// StartReader will start the audit log reader process which will continuously read and send events until stopped.
 	// Returns true if the reader can be started (log exists and can be read). Log file missing is not considered an error.
 	StartReader(ctx context.Context) (bool, error)
-	// StopReader will stop the reader if it's started. Will return false if it was already stopped.
-	StopReader() bool
+	// StopReader will stop the reader if it's started.
+	StopReader()
 }
 
 // NewReader returns a new instance of Reader
 func NewReader(client sensor.ComplianceService_CommunicateClient, nodeName string, clusterID string, startState *storage.AuditLogFileState) Reader {
 	return &auditLogReaderImpl{
 		logPath:    defaultLogPath,
-		stopC:      concurrency.NewSignal(),
+		stopper:    concurrency.NewStopper(),
 		sender:     newAuditLogSender(client, nodeName, clusterID),
 		startState: startState,
 	}
--- a/compliance/collection/auditlog/auditlog_impl.go
+++ b/compliance/collection/auditlog/auditlog_impl.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/gogo/protobuf/types"
 	"github.com/nxadm/tail"
-	"github.com/pkg/errors"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/pkg/logging"
@@ -34,16 +33,12 @@ var (
 
 type auditLogReaderImpl struct {
 	logPath    string
-	stopC      concurrency.Signal
+	stopper    concurrency.Stopper
 	sender     auditLogSender
 	startState *storage.AuditLogFileState
 }
 
 func (s *auditLogReaderImpl) StartReader(ctx context.Context) (bool, error) {
-	if s.stopC.IsDone() {
-		return false, errors.New("Cannot start reader because stopC is already done. Reader might have already been stopped")
-	}
-
 	t, err := tail.TailFile(s.logPath, tail.Config{
 		ReOpen:    true,
 		MustExist: true,
@@ -53,10 +48,13 @@ func (s *auditLogReaderImpl) StartReader(ctx context.Context) (bool, error) {
 	if err != nil {
 		if os.IsNotExist(err) {
 			log.Errorf("Audit log file %s doesn't exist on this compliance node", s.logPath)
-			return false, nil
+			err = nil
+		} else {
+			// handle other errors
+			log.Errorf("Failed to open file: %v", err)
 		}
-		// handle other errors
-		log.Errorf("Failed to open file: %v", err)
+		s.stopper.Flow().StopWithError(err)
+		s.stopper.Flow().ReportStopped()
 		return false, err
 	}
 
@@ -64,18 +62,20 @@ func (s *auditLogReaderImpl) StartReader(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
-func (s *auditLogReaderImpl) StopReader() bool {
-	return s.stopC.Signal()
+func (s *auditLogReaderImpl) StopReader() {
+	s.stopper.Client().Stop()
+	_ = s.stopper.Client().Stopped().Wait()
 }
 
 func (s *auditLogReaderImpl) readAndForwardAuditLogs(ctx context.Context, tailer *tail.Tail) {
+	defer s.stopper.Flow().ReportStopped()
 	defer s.cleanupTailOnStop(tailer)
 
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case <-s.stopC.Done():
+		case <-s.stopper.Flow().StopRequested():
 			return
 		case line := <-tailer.Lines:
 			if line == nil {
@@ -118,7 +118,7 @@ func (s *auditLogReaderImpl) readAndForwardAuditLogs(ctx context.Context, tailer
 			if err := s.sender.Send(ctx, &auditLine); err != nil {
 				// It's very likely that this failure is due to Sensor being unavailable
 				// In that case when Sensor next comes available it will ask to restart from the last
-				// message it got. Therefore this event will end up being sent at that point.
+				// message it got. Therefore, this event will end up being sent at that point.
 				// Therefore, we skip retrying at this point in time.
 				log.Errorf("Failed sending event to Sensor: %v", err)
 				continue
@@ -142,7 +142,7 @@ func (s *auditLogReaderImpl) shouldSendEvent(event *auditEvent, eventTS *types.T
 		s.startState = nil
 	}
 
-	// Only send when both the stage and resource type are in their corresponding allow list
+	// Only send when both the stage and resource type are in their corresponding allow-list
 	// and when verb is not disallowed
 	return stagesAllowList.Contains(event.Stage) &&
 		resourceTypesAllowList.Contains(event.ObjectRef.Resource) &&
@@ -150,7 +150,7 @@ func (s *auditLogReaderImpl) shouldSendEvent(event *auditEvent, eventTS *types.T
 }
 
 func (s *auditLogReaderImpl) cleanupTailOnStop(tailer *tail.Tail) {
-	// Only want to call tailer.Stop here as that stops the tailing. However do _not_ call tailer.Cleanup() because as the docs mention
+	// Only want to call tailer.Stop here as that stops the tailing. However, do _not_ call tailer.Cleanup() because as the docs mention
 	// "If you plan to re-read a file, don't call Cleanup in between."
 	// Cleanup is recommended for after the process exits, but that's not strictly necessary as if the process exits the container will be restarted anyway
 	// See https://pkg.go.dev/github.com/nxadm/tail#Tail.Cleanup for details on Cleanup()
--- a/pkg/concurrency/stopper.go
+++ b/pkg/concurrency/stopper.go
@@ -1,35 +1,116 @@
 package concurrency
 
-// NewStopper creates a new Stopper interface
+// NewStopper creates a new Stopper for arranging a graceful shutdown.
 func NewStopper() Stopper {
-	return Stopper{
-		stop:    NewSignal(),
-		stopped: NewSignal(),
+	return &stopperImpl{
+		stop:    NewErrorSignal(),
+		stopped: NewErrorSignal(),
 	}
 }
 
-// Stopper is an object that encapsulates both stop and stopped signals
-type Stopper struct {
-	stop    Signal
-	stopped Signal
+// Stopper encapsulates stop and stopped signals for arranging a graceful shutdown sequence for goroutines (and async
+// processing in general). These are referred as "stoppable" goroutines.
+type Stopper interface {
+	// Client returns an interface that should be used outside the goroutine to request a graceful shutdown and check
+	// the shutdown status.
+	Client() StopperClient
+
+	// Flow returns an interface that should be used in stoppable goroutine control flow to check and modify the state
+	// of Stopper. Flow should not be used outside the stoppable goroutine.
+	Flow() StopperFlow
+
+	// LowLevel allows to interact with the stopper at lower level. Callers should avoid using it unless there's no
+	// better alternative.
+	LowLevel() StopperLowLevel
 }
 
-// Stop signals the internal stop signal and effectively causes a push to StopDone
-func (s *Stopper) Stop() {
+// StopperClient represents an interface for an external client (external to a stoppable goroutine) to request the
+// stoppable goroutine to gracefully shutdown and to check the shutdown status.
+type StopperClient interface {
+	// Stop requests stoppable goroutine to shut down (or stop).
+	Stop()
+
+	// Stopped returns ReadOnlyErrorSignal that can be used for waiting until the stoppable goroutine reports that it
+	// finally stopped.
+	Stopped() ReadOnlyErrorSignal
+}
+
+// StopperFlow represents an interface that should be used by a stoppable goroutine to check and report its status of a
+// graceful shutdown.
+type StopperFlow interface {
+	// StopWithError initiates a shutdown due to an error that happened during async processing in goroutine. Should be
+	// called from the stoppable goroutine.
+	StopWithError(error)
+
+	// StopRequested provides a channel that should be checked in goroutine in select statements for goroutine to
+	// recognize that the shutdown was requested, and it needs to wrap up its work.
+	StopRequested() <-chan struct{}
+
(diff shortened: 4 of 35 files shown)


pkg/concurrency/stoppable.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Stopped returns ReadOnlyErrorSignal that can be used for waiting until the stoppable reports that it finally
+	// stopped.
+	Stopped() ReadOnlyErrorSignal
+}
+
+// StoppableFlow represents an interface for goroutine to check and report status of a graceful shutdown.
+type StoppableFlow interface {
+	// StopWithError initiates a shutdown due to an error that happened during async processing in goroutine.
+	StopWithError(error)
```
vikin91, line 37: Shouldn't `StoppableClient.Stop` and `StoppableFlow.StopWithError` be closer together in the same object?

msugakov: This was intentional.

`.Client().Stop()` is called externally (outside of goroutine), and I haven't found any cases when non-`nil` error was provided as parameter. That's why `.Client().Stop()` doesn't take `error` parameters.

OTOH, I found goroutines self-stop due to error (e.g. a channel closed). That's why I added `.Flow().StopWithError(error)` for internal interface of goroutines (that's what I call `StoppableFlow`).

sensor/common/compliance/command_handler_impl.go
```diff
@@ -40,12 +39,12 @@ func (c *commandHandlerImpl) Start() error {
 	return nil
 }
 
-func (c *commandHandlerImpl) Stop(err error) {
-	c.stopC.SignalWithError(err)
+func (c *commandHandlerImpl) Stop(_ error) {
+	c.stopper.Client().Stop()
```
vikin91, line 43: It looks like we are loosing information about the error passed here. As `c.stopper.Flow().StopWithError(err)` does not fit here, then maybe we need a replacement for sending a command to stop with error?

vikin91 (commented):
Looks good so far (as still WIP). I am curious to see it in the NodeScanHandler (or NodeInventoryHandler after landing https://github.com/stackrox/stackrox/pull/3755)

sensor/common/compliance/command_handler_impl.go
msugakov: No, we are not loosing information because the error is always `nil`.

It is exactly the same topic as above - https://github.com/stackrox/stackrox/pull/3931#discussion_r1034489073

I recommend you to check with IDE Find Usages what calls this `Stop()` method and whether there's any single place that gets non-`nil` error. I found zero.

sensor/common/sensor/sensor.go
```diff
@@ -217,11 +217,10 @@ func (s *Sensor) Start() {
 
 	select {
 	case <-errSig.Done():
-		s.stoppedSig.SignalWithErrorWrap(errSig.Err(), "getting connection from connection factory")
+		s.stopper.Flow().StopWithError(errors.Wrap(errSig.Err(), "getting connection from connection factory"))
+		s.stopper.Flow().ReportStopped()
 		return
 	case <-okSig.Done():
-	case <-s.stoppedSig.Done():
-		return
```
msugakov, line 224: This was simply redundant because nothing modified `s.stoppedSig` before this point, and `Sensor.Stop()` doesn't alter it either (maybe it should, but `Sensor.Stop()` looks like a can of worms at the moment).

sensor/common/compliance/command_handler_impl.go
vikin91: I see, so the error was a "plan for later" that we never used

msugakov: Probably. Something of this kind, or API was not separated for internal/external (from goroutine point of view) use cases and external functions were just a copy of internal needs. I guess we'd never know unless we go in archeology.

sensor/common/compliance/nodescan_handler_test.go
```diff
@@ -116,7 +115,7 @@ func (s *NodeScanHandlerTestSuite) TestHandlerRegularRoutine() {
 	s.NoError(h.Stopped().Wait())
 }
 
-func (s *NodeScanHandlerTestSuite) TestHandlerStoppedError() {
+func (s *NodeScanHandlerTestSuite) TestHandlerStopIgnoresError() {
```
msugakov, line 118: I think `TestHandlerStopIgnoresError` test became redundant and can be removed. Would it be ok with you @vikin91 ?

sensor/admission-control/manager/manager_impl.go
```diff
@@ -150,22 +149,22 @@ func (m *manager) Sync(ctx context.Context) error {
...
-	case <-m.stoppedSig.Done():
-		return m.stoppedSig.ErrorWithDefault(pkgErr.New("manager was stopped"))
+	case <-m.stopper.Client().Stopped().Done():
+		return m.stopper.Client().Stopped().ErrorWithDefault(pkgErr.New("manager was stopped"))
 	}
 }
 
 func (m *manager) Start() error {
-	if !m.stopSig.Reset() {
+	if m.stopper.LowLevel().ResetStopRequest() {
```
msugakov, line 167: There was a bug in this line before.
It was
```go
if !m.stopSig.Reset() {
```
but it should have been
```go
if m.stopSig.Reset() {
```

The reason it still worked was because `m.stopSig` wasn't initialized! Check that `NewManager` does not set it.

That's the reason it's better to use interfaces instead of structs. You can use uninitialized struct and have weird effects but you can't use `nil` interface.

compliance/collection/auditlog/auditlog_impl.go
```diff
@@ -34,16 +33,12 @@ var (
...
-	stopC      concurrency.Signal
+	stopper    concurrency.Stopper
 	sender     auditLogSender
 	startState *storage.AuditLogFileState
 }
 
 func (s *auditLogReaderImpl) StartReader(ctx context.Context) (bool, error) {
-	if s.stopC.IsDone() {
-		return false, errors.New("Cannot start reader because stopC is already done. Reader might have already been stopped")
-	}
```
msugakov, line 45: This was redundant because nothing in the current code tries to stop the reader before it's started.

compliance/collection/auditlog/auditlog_impl.go
```diff
@@ -53,29 +48,34 @@ func (s *auditLogReaderImpl) StartReader(ctx context.Context) (bool, error) {
...
 
 	go s.readAndForwardAuditLogs(ctx, t)
 	return true, nil
 }
 
-func (s *auditLogReaderImpl) StopReader() bool {
-	return s.stopC.Signal()
+func (s *auditLogReaderImpl) StopReader() {
+	s.stopper.Client().Stop()
+	_ = s.stopper.Client().Stopped().Wait()
```
msugakov, line 67: Absence of this line lead to the following error appear from time to time in logs when running tests.
```
collection/auditlog: 2022/11/30 12:19:56.751189 auditlog_impl.go:158: Error: Audit log tailer stopped with an error no such file or directory
```

Looks like I fixed a bug.

compliance/collection/auditlog/auditlog.go
```diff
@@ -18,15 +18,15 @@ type Reader interface {
 	// StartReader will start the audit log reader process which will continuously read and send events until stopped.
 	// Returns true if the reader can be started (log exists and can be read). Log file missing is not considered an error.
 	StartReader(ctx context.Context) (bool, error)
-	// StopReader will stop the reader if it's started. Will return false if it was already stopped.
-	StopReader() bool
```
msugakov, line 22: There was nothing that looks at the returned value.

pkg/concurrency/stopper.go
```diff
@@ -1,35 +1,115 @@
...
+
+	// Stopped returns ReadOnlyErrorSignal that can be used for waiting until the stoppable goroutine reports that it
+	// finally stopped.
+	Stopped() ReadOnlyErrorSignal
 }
 
-// Stop signals the internal stop signal and effectively causes a push to StopDone
-func (s *Stopper) Stop() {
+// StopperFlow represents an interface that should be used by a stoppable goroutine to check and report its status of a
+// graceful shutdown.
```
fredrb, line 39: If I understood your idea correctly, the `Flow` construct should be used by the sensor components if a parent (or any other component) requests a shutdown?

sensor/common/sensor/central_communication.go
```diff
@@ -22,7 +22,6 @@ func NewCentralCommunication(components ...common.SensorComponent) CentralCommun
 		sender:     NewCentralSender(components...),
 		components: components,
 
-		stopC:    concurrency.NewErrorSignal(),
-		stoppedC: concurrency.NewErrorSignal(),
+		stopper: concurrency.NewStopper(),
```
fredrb, line 25: Looking at the changes on this component and the others, my understanding is that each component has its own Stopper instance and external shutdown requests should come from `Stop` function. 

This is a bit conflicting with my understanding of the responsibility of `Flow` and `Client` (as per I commented above). I thought the `Client` was to be used by a parent caller (e.g. sensor's main goroutine), and the `Flow` by each component?

If `Stopper` contained in a single component's lifecycle, then I don't understand why we need these two interfaces. Why not have all the methods in the same one then? It seems to make the client code more difficult to reason about. Example:

```go
s.stopper.Flow().StopWithError(err)
// vs
s.stopper.StopWithError(err)

s.stopper.Client().Stop()
// vs
s.stopper.Stop()
```

pkg/concurrency/stopper_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	countdown := stdSync.WaitGroup{}
+	countdown.Add(6)
+
+	// This is how to implement a gracefully stoppable goroutine.
+	go func() {
+		// The goroutine must report that it shut down at the end.
+		defer stopper.Flow().ReportStopped()
+		// Here ticker simulates some async source of data that the goroutine handles.
+		ticker := time.NewTicker(50 * time.Millisecond)
```
fredrb, line 32: This might be a bit pedantic, but why not use 50 microseconds here?

Since this is a unit test with nothing else happening other than waiting, the test will needlessly wait `300ms`.

pkg/concurrency/stopper.go
msugakov: Flow is to be used by goroutine. Flow allows to check if shutdown was requested and allows to report that the shutdown was completed.
Client is to be used outside of goroutine. Client allows to request shutdown and allows to way for the shutdown to complete.
Makes sense?

sensor/common/sensor/central_communication.go
msugakov: It could have been one interface, but the problem is that I saw a number of misinterpretations and problems when the pattern was implemented with two (Error)Signal-s. Forcing separation of interfaces is intentional to help spot when wrong things are used in wrong places.

`StopWithError` is part of `Flow` because I found that reporting stop with error only makes sense from within the goroutine, but not outside of it.

pkg/concurrency/stopper_test.go
msugakov: Shorter timeout would mean more chance of flake if something in threading or in runtime takes a bit longer.
I don't mind to decrease but I don't know how to reason about which value is sufficiently big to keep the test stable.

I tried running 300 times with 1ms delay (`go test -count=300 -v -run ^TestStopper$ ./pkg/concurrency`) and the test flaked.
10 ms - no flake.
5 ms - no flake.
3 ms - no flake.
2 ms - no flake.
1 ms - got flakes.

Therefore, 50 microseconds is definitely too short.
I also don't feel very comfortable about 10ms.

I suggest to decrease the countdown from 6 to 3 and delay from 50 to 20ms.

msugakov: Well. `sync` package from standard Go library is banned by our linters therefore I had to change the implementation so that it doesn't rely on timing.

fredrb: > Well. sync package from standard Go library is banned by our linters therefore I had to change the implementation so that it doesn't rely on timing.

We have out own `pkg/sync` which wraps the standard sync functions. You should use that one instead

msugakov: Please check the recent code. There is no use of standard `sync`.

pkg/concurrency/stopper_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestStopper(t *testing.T) {
+	suite.Run(t, &StopperTestSuite{})
+}
+
+// TestCommonCase demonstrates how to use Stopper for implementing a gracefully stoppable goroutine.
+func (s *StopperTestSuite) TestCommonCase() {
+	stopper := NewStopper()
+
+	s1 := NewSignal()
+	s2 := NewSignal()
```
vikin91, line 23: I guess some readers of this code may ask themselves how to understand s1 and s2.
It s1 a signal to start the work in the goroutine and s2 a signal to mark the end of that work? Maybe we could rename s1 and s2 so that their roles are easier to understand?

sensor/common/component.go
```diff
@@ -10,7 +10,7 @@ import (
 // as well as sending messages back to central.
 type SensorComponent interface {
 	Start() error
-	Stop(err error)
+	Stop(err error) // TODO: get rid of err argument as it always seems to be effectively nil.
```
vikin91, line 13: I bet we forget about this TODO. Could you maybe create a ticket for it and link it? Or at least start a new draft PR for this?

vikin91 (approved):
It looks really good! The split of Client and Flow interfaces will help developers avoid mistakes that may be difficult to find later.

I left few nit-comments.

central/pruning/pruning.go
```diff
@@ -876,6 +875,6 @@ func (g *garbageCollectorImpl) pruneLogImbues() {
 }
 
 func (g *garbageCollectorImpl) Stop() {
-	g.stopSig.Signal()
-	<-g.stoppedSig.Done()
+	g.stopper.Client().Stop()
+	_ = g.stopper.Client().Stopped().Wait()
```
fredrb, line 879: You don't need the left hand `_` assignment. Any reasons to explicitly do that?

fredrb: Same comment for every other left hand empty assignments.

sensor/admission-control/manager/manager_impl.go
fredrb: Good catch on the issue! 

However, I'm struggling to understand why is this needed in the first place. It seems like this manager's `Start()` is only called once on admission control's `main` function: https://github.com/stackrox/stackrox/blob/master/sensor/admission-control/main.go#L79-L80

If the component is just starting, in which case would the signal be already triggered during startup? 

Additionally, the function call and the error message are even more confusing. If the reset was successful (i.e. `ResetStopRequest()` returns `true`) it returns and error saying the manager has already started. AFAIU this would mean that the manager was _stopped_ instead. The only reason I can see this would return true is if the component started, received a stop signal, and the component was requested to start again.

pkg/concurrency/stopper.go
fredrb: Ok, after reading the code again a few times I think it's making more sense now. Thanks.

sensor/common/component.go
fredrb: +1

sensor/common/detector/detector.go
```diff
@@ -362,9 +362,9 @@ func (d *detectorImpl) runAuditLogEventDetector() {
 				return alerts[i].GetPolicy().GetId() < alerts[j].GetPolicy().GetId()
 			})
 			select {
-			case <-d.auditStopper.StopDone():
+			case <-d.auditStopper.Flow().StopRequested():
```
fredrb, line 365: [nit / rant / no immediate suggestion]

I'm not so happy with the pattern of nested selects with the same stop signal repeated. Like (pseudo-code):

```
select {
   <- stoppedPath:
         return
   <- workPath:
         select {
              <-stoppedPath:
                     return
              <- workPath:
                     // work
         }
}
```

I don't have a better suggestion right now. But maybe it would make sense to use non-blocking operations on the inner block and only handle stop signals at the top level. Since it's an infinite loop it will go back to the top level select anyway. It will still be a problem if the buffer becomes full then writing into it becomes blocking. One could make the argument that if the buffered `d.output` chan is full we probably have other serious problems.

fredrb (commented):
Code looks good. Some general comments and thoughts. No blockers.

The distinction between `Client` and `Flow` is more clear now on the second pass.

However, I have one last optional suggestion: I wonder if there's a way to enforce `Client`/`Flow` distinction (at least as a code pattern) in a way that avoid misuse of the `Stopper` interface inside the goroutine? For other dev that wants to build a new sensor component the distinction might not be so obvious. 

We don't want the code inside the component's goroutine to call `stopper.Client()` functions, only `stopper.Flow()`, right? So maybe the goroutine should be anonymous and receive only what it should operate on? 

Something like:

```go
func (c *MyComponent) Start() {
    // do pre-work
    go func() {
           // Component main loop
    }(c.stopper.Flow(), someOtherDependency)
}
```

Maybe out of the scope for this PR. WDYT?

pkg/concurrency/stopper_test.go
fredrb: Looks good, thanks.

sensor/common/detector/detector.go
vikin91: Note that I think this is required here as there are two sources of potential blocking:
1. `auditEvents := <-d.auditEventsChan`
2. `d.output <- &central.MsgFromSensor{`

We want to be sure to stop the goroutine in case of any of this places block, thus we are checking the stopper twice.

fredrb: That's exactly what I pointed out. My idea is having a **non-blocking** operation in the inner select.

sensor/admission-control/manager/manager_impl.go
msugakov: You're right, the `Start()` is called right after creating a `New()` manager. It is impossible that something calls `Stop()` on `mgr` between those lines.

Either the code is over-designed or there _was_ some different usage in the past.

Regarding `ResetStopRequest()` return value, what you write here

>  The only reason I can see this would return true is if the component started, received a stop signal, and the component was requested to start again.

matches my understanding.

There is no way of identifying whether the manager got **started** from `stopper` (currently) or `stopSig` previously.

The only thing I can offer in this case is to delete the entire `if-error` block to make the code less confusing. However, I prefer to not take this too far as there are **many** simple and difficult bugs in the code I touched and I don't want to be fixing everything for everyone.

pkg/concurrency/stopper_test.go
```diff
@@ -0,0 +1,126 @@
...
+				s2.Signal()
+			}
+		}
+	}()
+
+	s1.Signal()
+	s2.Wait() // Wait until it is time to shut down the goroutine.
+
+	// This is how to tell goroutine to stop.
+	// Note that if this doesn't go through or goes slowly the `countdown.Done()` call will panic.
```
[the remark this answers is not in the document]
msugakov: Yes, it was. Thanks!

pkg/concurrency/stopper_test.go
msugakov: I thought about this but can't find good names. Also, the code is small and localized that just reading it explains what `s1` and `s2` do. I think, in this case it's like explaining a joke. Don't want to spoil the joy :-)

central/pruning/pruning.go
msugakov: `Stopped()` is `ReadOnlyErrorSignal`, therefore its `Wait()` returns an `error`. The `error` will be `nil` if this signal was signalled "normally" or will be an error value if the signal was provided that. For example, in case of abnormal shutdown.

By assigning the result to `_` I calm down GoLand which otherwise tells me that there's a potential `error` that wasn't respected.

With underscore:
image

Without underscore:
image

sensor/common/detector/detector.go
msugakov: 1. I agree that Go syntax in this scenario is clumsy. If it looked better, it might have annoyed less.
2. My _opinion_ is that this pattern is actually fine. I don't know of a more elegant way to express "read from here and write over there" async work in general case.
3. Introducing buffered channels may break things. I prefer to make mostly cosmetic changes than breaking things in this PR. Also don't want to fix too many things that I did not break.

Would you be ok to accept this unchanged?

fredrb: > Would you be ok to accept this unchanged?

Yes

sensor/admission-control/manager/manager_impl.go
fredrb: I'm fine leaving the code as is. Just wanted to point out. 

Maybe add a comment linking to this discussion? Just in case someone finds this code in the future and feels brave enough to try removing the if and testing if anything breaks?

msugakov: Sorry, I wasn't clear enough. I meant that I will fix this issue (the fix is simple and I just committed it) but don't want to get on a path of fixing all issues because some might be more involved.

msugakov:
@fredrb Sorry, I missed your suggestion previously.

> I wonder if there's a way to enforce Client/Flow distinction (at least as a code pattern)

That's the ideal state I wanted to get it to but unfortunately couldn't find a way. Mainly the reason is that we use `Stopper` in a struct as a field and so every struct's function has access to everything in the Stopper. Even in the snippet you provide
```go
func (c *MyComponent) Start() {
    // do pre-work
    go func(flow concurrency.Flow, dependency T) {
           // Component main loop
    }(c.stopper.Flow(), someOtherDependency)
}
```
there is `c.stopper` and goroutine can use that instead of `flow`. If some things _can_ be done differently, humans _will_ do them differently. The ideal state would be if there's only a single way and it hides things that should not be seen, but, as I mentioned, I did not find it.

> We don't want the code inside the component's goroutine to call `stopper.Client()` functions, only `stopper.Flow()`, right?

For me it is actually the opposite. We don't want anything outside of goroutine to use `stopper.Flow()`. Not having `stopper.Client()` in goroutine would be nice, but there are some cases that violate that, e.g. `sensor/common/sensor/central_receiver_impl.go`.

PR #3934, opened 2022-11-28
Add basic smoke tests for PatternFly version of Network Graph

vjwilson:
## Description

The journey of 1000 miles starts with one step.

## Checklist

## Testing Performed

Run locally
Screen Shot 2022-11-28 at 5 52 24 PM

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -43,7 +43,7 @@ deploy_stackrox_with_custom_sensor() {
 
     # generate init bundle
     password_file="$ROOT/deploy/$ORCHESTRATOR_FLAVOR/central-deploy/password"
-    if [ ! -f "$password_file" ]; then 
+    if [ ! -f "$password_file" ]; then
         die "password file $password_file not found after deploying central"
     fi
     kubectl -n stackrox exec deploy/central -- roxctl --insecure-skip-tls-verify \
@@ -82,6 +82,7 @@ export_test_environment() {
     ci_export ROX_QUAY_ROBOT_ACCOUNTS "${ROX_QUAY_ROBOT_ACCOUNTS:-true}"
     ci_export ROX_SEARCH_PAGE_UI "${ROX_SEARCH_PAGE_UI:-true}"
     ci_export ROX_SYSTEM_HEALTH_PF "${ROX_SYSTEM_HEALTH_PF:-true}"
+    ci_export ROX_NETWORK_GRAPH_PATTERNFLY "${ROX_NETWORK_GRAPH_PATTERNFLY:-true}"
 }
 
 deploy_central() {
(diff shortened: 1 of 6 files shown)


tests/e2e/lib.sh
```diff
@@ -82,6 +82,7 @@ export_test_environment() {
     ci_export ROX_QUAY_ROBOT_ACCOUNTS "${ROX_QUAY_ROBOT_ACCOUNTS:-true}"
     ci_export ROX_SEARCH_PAGE_UI "${ROX_SEARCH_PAGE_UI:-true}"
     ci_export ROX_SYSTEM_HEALTH_PF "${ROX_SYSTEM_HEALTH_PF:-true}"
+    ci_export ROX_NETWORK_GRAPH_PATTERNFLY "${ROX_NETWORK_GRAPH_PATTERNFLY:-true}"
```
pedrottimark, line 85: Aha, this is step 4 in the instructions:
https://github.com/stackrox/stackrox/tree/master/ui/apps/platform#add-a-feature-flag-to-frontend-code

ui/apps/platform/cypress/helpers/networkGraphPF.js
```diff
@@ -0,0 +1,111 @@
+import * as api from '../constants/apiEndpoints';
+import { selectors as networkGraphSelectors } from '../constants/NetworkPage';
```
pedrottimark, line 2: Recommendation not a blocker to merge:

If there is a helper function for a semantic interaction, prefer to encapsulate the relevant api endpoints and DOM selectors in the helpers file.
* Only Network Graph tests`api.network.whatever` properties. Most if not all deserve to be encapsulated by visit or interact helper functions.
* For selectors, the only counter-example that comes to mind for **both** helpers **and** tests to use selectors is disabled/enabled assertions about Save or Test buttons in integration form tests.

ui/apps/platform/src/routePaths.js
```diff
@@ -149,7 +149,7 @@ const vulnManagementPathToLabelMap = {
 export const basePathToLabelMap = {
     [dashboardPath]: 'Dashboard',
     [networkBasePath]: 'Network Graph',
-    [networkBasePathPF]: 'Network Graph-Patternfly',
+    [networkBasePathPF]: 'PatternFly Network Graph',
```
pedrottimark, line 152: Because the container renders the page title, which is the preferred pattern:
https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx#L65

Can you see if it works to delete this title property and instead provide a literal string in the `NavigationSidebar` component? Although the procedure to add a container must include `Body` and usually includes `NavigationSidebar` components, every extra “global registration” step is a barrier to inclusion and risks being the straw that breaks the camel’s brain, according to the principle of up-to-7 items in short-term memory.

vjwilson:
Superseded by https://github.com/stackrox/stackrox/pull/5264

PR #3937, opened 2022-11-29
ROX-13644 enable fake k8s client in sensor integration tests

lvalerom:
## Description

This PR enables the `TestContext` used in sensor's integration tests to use a fake k8s client.

The PR also reenables the `DeploymentExposure` tests that now use this fake client. By changing this, the flake [ROX-12948](https://issues.redhat.com/browse/ROX-12948) should be fixed.

## Checklist

## Testing Performed

* [ ] Five consecutive successful runs in CI
* [x] Manual testing with `make sensor-integration-test`

### DIFF
--- a/sensor/tests/resource/helper.go
+++ b/sensor/tests/resource/helper.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


lvalerom:
We decided to close this because it doesn't make sense to complicate the integration test utilities by adding the fake k8s client, which will also taint the nature of these tests.

PR #3943, opened 2022-11-29
Reuse NetworkEndpoint in "processes listening on ports" user-facing API

ovalenti:
This is an attempt to use the NetworkEndpoint object from network-flow to report processes listening on ports through the customer API. This is achieved by enriching the NetworkEntityInfo with originator process information (in the form of ProcessIndicator).
The user API would return a list of NetworkEndpoint objects.


ovalenti:
Closing this as it doesn't seem worth the effort; in particular, process-originator won't make much sense in network-graph, since endpoints are per deployment.

PR #3949, opened 2022-11-30
ROX-13685: Add query field to deployment matching on collections

keyallis:
Support for filter queries on collection endpoints returning deployment matches

## Checklist

## Testing Performed

ROX-13686 filed to track expanding E2E tests to include this logic


md2119 (approved):

proto/api/v1/resource_collection_service.proto
```diff
@@ -25,7 +25,7 @@ message GetCollectionRequest {
 
 message CollectionDeploymentMatchOptions {
     bool with_matches = 1;
-    Pagination matches_pagination = 2;
+    RawQuery filter_query = 3;
```
md2119, line 28: You do not need to update the tag number. The tag `2` is still not used by end-users.

theencee (approved):

proto/api/v1/resource_collection_service.proto
keyallis: Ah good catch, I forgot to do that

PR #3957, opened 2022-11-30
Start a pprof server for realtime metric observation

fredrb:
## Description

Other than outputting the pprof metric bundle to files, we could provide an http pprof server for realtime observations. This also allows to explore other pprof metrics that are not exported now, like `goroutines` and `mutext`. 

## Checklist

## Testing Performed

Example:
```
go tool pprof -http localhost:9090 http://localhost:6060/debug/pprof/goroutine
```

### DIFF
--- a/tools/local-sensor/main.go
+++ b/tools/local-sensor/main.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


lvalerom (approved):
Looks good to me.

PR #3961, opened 2022-11-30
Add Cypress tests for collection CRUD operations

dvail:
## Description

Add Cypress e2e tests for core collection CRUD workflows: create/edit/clone/delete.

Enhances the aria-label values for existing value selectors for better accessibility and more reliable test hooks.

Fixes a bug with the collection name in the form, where in some cases the `(COPY)` string was not appended in clone mode.

## Follow ups

- Tests for the embedded collection feature
- Tests that assert on values in the collection results sidebar at a basic level

## Checklist

## Testing Performed

Local Cypress test run:
image

Additionally, each test was manually interrupted and then the entire spec was run again from the start to ensure mid-spec failures correctly clean up data stored server side.

### DIFF
--- a/ui/apps/platform/cypress/constants/apiEndpoints.js
+++ b/ui/apps/platform/cypress/constants/apiEndpoints.js
@@ -182,10 +182,3 @@ export const extensions = {
 export const permissions = {
     mypermissions: '/v1/mypermissions',
 };
-
-export const collections = {
-    baseUrl: '/v1/collections',
-    count: '/v1/collectionscount',
-    dryRun: '/v1/collections/dryrun',
-    autocomplete: '/v1/collections/autocomplete',
-};
--- a/ui/apps/platform/cypress/constants/apiEndpoints.js
+++ b/ui/apps/platform/cypress/constants/apiEndpoints.js
@@ -182,3 +182,10 @@ export const extensions = {
 export const permissions = {
     mypermissions: '/v1/mypermissions',
 };
+
+export const collections = {
+    baseUrl: '/v1/collections',
+    count: '/v1/collectionscount',
+    dryRun: '/v1/collections/dryrun',
+    autocomplete: '/v1/collections/autocomplete',
+};
--- a/ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
+++ b/ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
@@ -1,14 +1,16 @@
 import withAuth from '../../helpers/basicAuth';
-import { collections as collectionsApi } from '../../constants/apiEndpoints';
 import { visitCollections } from '../../helpers/collections';
 import { hasFeatureFlag } from '../../helpers/features';
+
+const baseUrl = '/v1/collections';
+const autocompleteUrl = `${baseUrl}/autocomplete`;
 
 // Cleanup an existing collection via API call
 function tryDeleteCollection(collectionName) {
     const auth = { bearer: Cypress.env('ROX_AUTH_TOKEN') };
 
     cy.request({
-        url: `${collectionsApi.baseUrl}?query.query=Collection Name:"${collectionName}"`,
+        url: `${baseUrl}?query.query=Collection Name:"${collectionName}"`,
         auth,
     }).as('listCollections');
 
@@ -16,12 +18,16 @@ function tryDeleteCollection(collectionName) {
         const collection = res.body.collections.find(({ name }) => name === collectionName);
         if (collection) {
             const { id } = collection;
-            const url = `${collectionsApi.baseUrl}/${id}`;
+            const url = `${baseUrl}/${id}`;
             cy.request({ url, auth, method: 'DELETE' });
         }
     });
 }
 
+/* 
+    Each test in this spec builds upon the previous by executing another piece
+    of the collection CRUD workflow.
+*/
 describe('Create collection', () => {
     withAuth();
 
@@ -31,7 +37,7 @@ describe('Create collection', () => {
         }
         // Ignore autocomplete requests
         // TODO Remove this once the feature is in
-        cy.intercept(`${collectionsApi.autocomplete}`, {});
+        cy.intercept(autocompleteUrl, {});
     });
 
     const collectionName = 'Financial deployments';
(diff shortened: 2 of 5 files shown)


dvail:
* master
  * **PR #3961** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3961" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3961?utm_source=stack-comment).

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
```diff
@@ -0,0 +1,150 @@
...
+        cy.get(`button:contains('Add "payments"')`).click();
+
+        cy.get('button:contains("All clusters")').click();
+        cy.get('button:contains("Clusters with names matching")').click();
+        cy.get('input[aria-label="Select value 1 of 1 for the cluster name"]').type('production');
+        cy.get(`button:contains('Add "production"')`).click();
+
+        cy.get('button:contains("Save")').click();
+
+        cy.get('a:contains("Financial deployments")');
```
pedrottimark, line 77: What are pro and con for understandability to add table cell to selector?

```js
cy.get('td[data-label="Collection"] a:contains("Financial deployments")');
```

pedrottimark: A separate stylistic question which I ask but have no strong opinion, whether to use template literal:

```js
cy.get(`a:contains("${collectionName}")`);
```

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
```diff
@@ -0,0 +1,150 @@
+import withAuth from '../../helpers/basicAuth';
+import { collections as collectionsApi } from '../../constants/apiEndpoints';
+import { visitCollections } from '../../helpers/collections';
+import { hasFeatureFlag } from '../../helpers/features';
+
+// Cleanup an existing collection via API call
```
pedrottimark, line 6: You keep teaching me new idioms!

If first impression is correct that most or all tests depend on each other:
1. Can you add a comment to make the dependency explicit?
2. What are pro and con to call `tryDeleteCollection` at beginning of test versus to add an explicit test at end to delete created and cloned collections via user interface?

pedrottimark (approved):
Bravo to improve accessibility while you write tests! A few comments as usual.

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
dvail: Aha, I meant to change the second one before I pushed and looks like it slipped by.

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
dvail: 1. Sure. Ideally I'd have each test fully independent, but I think being able to test the full workflow here is worthwhile and I'm not sure how to keep the tests independent without a lot of repetition. Having a single spec file as independent is probably good enough in this case.
2. This was actually a recommendation from the Cypress docs so that each test is capable of cleaning up before assertions are made. This covers the cases where we run tests out of order (in development, `it.skip`, `it.only`, etc), or a test fails before its cleanup functions can be called. That said, why not both? It might be beneficial to add a catch all cleanup test at the end to keep data from leaking out of this spec, plus we get the benefit of more actual UI test coverage.

pedrottimark: Indeed. I admit that I did not analyze carefully to what extent you can `skip` or `only` in this test file.

ui/apps/platform/cypress/constants/apiEndpoints.js
```diff
@@ -182,3 +182,10 @@ export const extensions = {
 export const permissions = {
     mypermissions: '/v1/mypermissions',
 };
+
+export const collections = {
```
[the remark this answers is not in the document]
pedrottimark: Yes, please.

1. You might use string literal for transient uses of endpoint like autocomplete mock in test file.
2. As a follow up, let’s discuss helper functions to encapsulate `cy.get('button:contains("Save")').click();`
    * `saveCreatedCollection` for POST and GET requests
    * `saveUpdatedCollection` for PUT and GET requests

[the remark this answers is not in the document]
pedrottimark: Also for `deleteCollection(collectionName)` as discussed in another comment.

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
```diff
@@ -0,0 +1,184 @@
...
+        cy.get(`td[data-label="Collection"] a:contains("${clonedName}")`);
+    });
+
+    it('should delete the previously created collections', () => {
+        // From the main table
+        visitCollections();
+
+        // Delete the clone first, since the `:contains()` selector cannot do an exact match
+        // Delete one from the main collection table
+        cy.get(`tr:has(a:contains("${clonedName}")) button[aria-label="Actions"]`).click();
```
dvail, line 166: Just realized that the long awaited `:has` CSS selector that landed in Chrome v105 can be used in our e2e tests since the current version of Chrome that Cypress is using is 108.

IMO this can simplify some complicated selections that make sense from a user perspective but were difficult to implement in a single selector so that we can take advantage of Cypress retry mechanisms. This is especially helpful when working with tables.

From an end user view, I'd scan the table for the collection I want to delete by name, go to the menu button, then click it. This CSS selector effectively says "Find me a `tr` that has a descendant `a` containing the text `"${clonedName}"`, then get me a button within that `tr` that is labeled `"Actions"` ".

I'm not sure if there is another way to do this with a single selector without hard coding a `tr` index or making other assumptions about the DOM structure.

pedrottimark: Super! Please share this excellent example in next team meeting.

Cypress folder has 7 occurrences in 4 files, which might have been via jQuery:
* https://api.jquery.com/has-selector/
* https://developer.mozilla.org/en-US/docs/Web/CSS/:has

dvail: Aww, I thought I discovered something. Chances are this is still just using the jQuery selector under the hood...

Still good to know we can use it though :)

PR #3967, opened 2022-11-30
Replace label selector dropdowns with plain text inputs

dvail:
## Description

This replaces the typeahead dropdown inputs for _by label_ values with a plain text input. Since we are not doing autocomplete for these fields this should be a more appropriate UX.

Additionally, this removes the red "required" star from inputs when the form is in read only mode to reduce visual clutter.

## Checklist

## Testing Performed

Manual testing + verification of functionality covered by existing e2e tests.

Readonly variant of input fields:
image

Editable variant of input fields:
image

### DIFF
--- a/ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
+++ b/ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
@@ -57,16 +57,13 @@ describe('Create collection', () => {
         cy.get('button:contains("All deployments")').click();
         cy.get('button:contains("Deployments with labels matching")').click();
         cy.get('input[aria-label="Select label key for deployment rule 1 of 1"]').type('meta/name');
-        cy.get(`button:contains('Add "meta/name"')`).click();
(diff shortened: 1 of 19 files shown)


dvail:
* master
  * **PR #3961** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3961" target="_blank">Graphite</a> 
    * **PR #3967** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3967" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3967?utm_source=stack-comment).

dvail:
Note: this has implications with how we handle the debouncing for the results panel, since the form state will update on every keystroke instead of `onChange` of the entire field. I'll hold this PR until we decide whether or not it makes more sense to handle the debouncing at the results panel, or if these should be changed to uncontrolled inputs that emit debounced value updates.

pedrottimark (approved):
The `readOnlyVariant` prop became available just in time!

PR #3968, opened 2022-11-30
Dashrews/ROX-13662 make sure tests run pg

dashrews78:
## Description

Turns out that various tests that attempt to use Postgres as a datastore were not actually being executed via the 'go-postgres-unit-tests' Makefile target.  This leads to the potential for regressions and various other issues by not running these tests.  Due to the fact that many of these tests run at the datastore level and thus can use either Rocks or Postgres, I added them to the Makefile target vs renaming them to have postgres in the name so the target would pick them up as it was.

As part of this work, ROX-13701 and ROX-13703 were written to deal with some tests that were not simple fixes as the problems referenced in those tickets may in fact be bugs.

Note:  This PR does not include `pruning_test.go`.  That one was already in the process of being taken care of with the work for https://github.com/stackrox/stackrox/pull/3863

## Checklist

## Testing Performed

Ran the tests with Postgres enabled.


central/image/datastore/store/common/v2/parts_test.go
```diff
@@ -120,12 +120,12 @@ func TestSplitAndMergeImage(t *testing.T) {
 			},
 		},
 		ImageCVEEdges: map[string]*storage.ImageCVEEdge{
-			"cve1": {
+			cve.ID("cve1", ""): {
```
md2119, line 123: ✅

central/image/datastore/store/common/v2/parts_test.go
```diff
@@ -120,12 +120,12 @@ func TestSplitAndMergeImage(t *testing.T) {
...
 				Id:         postgres.IDFromPks([]string{"sha", cve.ID("cve1", "")}),
 				ImageId:    "sha",
 				ImageCveId: cve.ID("cve1", ""),
 			},
-			"cve2": {
+			cve.ID("cve2", ""): {
```
md2119, line 128: ✅

central/search/service/service_impl_test.go
```diff
@@ -229,6 +234,7 @@ func (s *SearchOperationsTestSuite) TestAutocomplete() {
 		},
 	} {
 		s.Run(fmt.Sprintf("Test case %q", testCase.query), func() {
+			log.Info(testCase.query)
```
md2119, line 237: rm

central/search/service/service_impl_test.go
```diff
@@ -244,14 +250,25 @@ func (s *SearchOperationsTestSuite) TestAutocompleteForEnums() {
...
-	policyIndexer := policyIndex.New(idx)
-	s.NoError(policyIndexer.AddPolicy(fixtures.GetPolicy()))
-	policySearcher := policySearcher.New(policyStore, policyIndexer)
-	ds := policyDatastore.New(policyStore, policyIndexer, policySearcher, nil, nil)
+	var policyIndexer policyIndex.Indexer
+	var ds policyDatastore.DataStore
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		policyStore := policyPostgres.New(s.pool)
+		policyIndexer = policyPostgres.NewIndexer(s.pool)
+		s.NoError(policyIndexer.AddPolicy(fixtures.GetPolicy()))
```
md2119, line 258: rm. This is only meant for bleve indexing.

central/search/service/service_impl_test.go
```diff
@@ -244,14 +250,25 @@ func (s *SearchOperationsTestSuite) TestAutocompleteForEnums() {
...
+		s.NoError(policyStore.Upsert(ctx, fixtures.GetPolicy()))
```
md2119, line 259: Not related to your PR but inserting directly into store is an upsert pattern. (My guess is it is done this way likely to skip some steps in ds)

pkg/nodes/enricher/enricher_impl_test.go
```diff
@@ -211,10 +221,18 @@ func TestCVESuppression(t *testing.T) {
...
 
-	node := &storage.Node{Id: "id"}
+	node := &storage.Node{Id: fixtureconsts.Node1}
 	err := enricherImpl.EnrichNode(node)
 	require.NoError(t, err)
-	assert.True(t, node.Scan.Components[0].Vulns[0].Suppressed)
+
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		for _, c := range node.GetScan().GetComponents() {
+			// `vulnerabilities` is the new field.
```
md2119, line 230: ✅

md2119 (approved):

PR #3969, opened 2022-11-30
ROX-13673 + ROX-13674 + ROX-13675: Create cidr block side panel layout

sachaudh:
## Description

This PR is an iterative step to creating the CIDR Block side panel. I decided to include three of the subtasks in this PR to minimize PRs. You can find all the subtasks for the CIDR Block side panel work here: https://issues.redhat.com/browse/ROX-13672

## Checklist

## Screenshots

Screenshot 2022-11-30 at 10 42 59 AM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -151,8 +151,12 @@ const TopologyComponent = ({ model }: TopologyComponentProps) => {
                             edges={model?.edges || []}
                         />
                     )}
-                    {selectedEntity && selectedEntity?.data?.type === 'CIDR_BLOCK' && (
-                        <CidrBlockSideBar />
+                    {selectedEntity && selectedEntity?.data?.type === 'EXTERNAL_SOURCE' && (
+                        <CidrBlockSideBar
+                            id={selectedEntity.id}
+                            nodes={model?.nodes || []}
+                            edges={model?.edges || []}
+                        />
                     )}
                     {selectedEntity && selectedEntity?.data?.type === 'EXTERNAL_ENTITIES' && (
                         <ExternalEntitiesSideBar />
--- a/ui/apps/platform/src/Containers/NetworkGraph/cidr/CidrBlockSideBar.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/cidr/CidrBlockSideBar.tsx
@@ -1,22 +1,167 @@
-import React from 'react';
-import { Badge, Flex, FlexItem, Text, TextContent, TextVariants } from '@patternfly/react-core';
+import React, { ReactElement } from 'react';
+import {
+    Divider,
+    Flex,
+    FlexItem,
+    Stack,
+    StackItem,
+    Text,
+    TextContent,
+    TextVariants,
+    Toolbar,
+    ToolbarContent,
+    ToolbarItem,
+} from '@patternfly/react-core';
+import { EdgeModel, NodeModel } from '@patternfly/react-topology';
 
-function CidrBlockSideBar() {
+import { getNodeById } from '../utils/networkGraphUtils';
+import { getAllUniquePorts, getNumFlows } from '../utils/flowUtils';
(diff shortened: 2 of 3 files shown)


sachaudh:
@alanonthegit @zhenpesky where can I find this icon?

Screenshot 2022-11-30 at 10 49 39 AM

ui/apps/platform/src/Containers/NetworkGraph/cidr/CidrBlockSideBar.tsx
```diff
@@ -1,22 +1,167 @@
...
+        defaultAdvancedFlowsFilters
+    );
+    const initialExpandedRows = flows
+        .filter((row) => row.children && !!row.children.length)
+        .map((row) => row.id); // Default to all expanded
+    const [expandedRows, setExpandedRows] = React.useState<string[]>(initialExpandedRows);
+    const [selectedRows, setSelectedRows] = React.useState<string[]>([]);
+
+    // derived data
+    const cidrBlockNode = getNodeById(nodes, id);
```
vjwilson, line 87: Are you going to need all the `nodes` at a later iteration of this component? (similar question about `edges` )

If not, it may be worth having the parent compute this and past it in.

ui/apps/platform/src/Containers/NetworkGraph/common/NetworkGraphIcons.tsx
```diff
@@ -8,3 +8,7 @@ export function DeploymentIcon() {
 export function NamespaceIcon() {
     return <Badge style={{ backgroundColor: 'rgb(32,79,23)' }}>NS</Badge>;
 }
+
+export function CidrBlockIcon() {
+    return <Badge style={{ backgroundColor: 'rgb(0,0,0)' }}>IP</Badge>;
```
vjwilson, line 13: Question that I forgot to ask earlier, about the colors of these "badges".

Is our UX team giving these colors, and will we try to get them added to PatternFly, so that all Red Hat apps can standardize on colors for k8s entities?

vjwilson (approved):
Approving to unblock, but would like to hear back on the few questions I had.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -151,8 +151,12 @@ const TopologyComponent = ({ model }: TopologyComponentProps) => {
                             edges={model?.edges || []}
                         />
                     )}
-                    {selectedEntity && selectedEntity?.data?.type === 'CIDR_BLOCK' && (
-                        <CidrBlockSideBar />
+                    {selectedEntity && selectedEntity?.data?.type === 'EXTERNAL_SOURCE' && (
```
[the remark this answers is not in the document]
sachaudh: Yea you have a good point. I made the change because internally we kept it as the backend values and only used the UI values in the rendered view so I made the change to allow myself to see things for now. I will change it to `CIDR_BLOCK` in a future PR. We will eventually type out the `data` field so there is a more consistent data structure

ui/apps/platform/src/Containers/NetworkGraph/cidr/CidrBlockSideBar.tsx
sachaudh: Good point. I think for now we'll keep it and decide if we need to refactor things in a future PR

ui/apps/platform/src/Containers/NetworkGraph/common/NetworkGraphIcons.tsx
sachaudh: I wasn't given color values, but I used what I saw from the mocks. I'm not entirely sure about standardizing these for other apps though. For the CIDR block icon, I just tried to make it similar to what we had already. I asked a question about where to get that IP icon. I'll wait on a response from the UX team about that. @alanonthegit @zhenpesky maybe you have thoughts on this

sachaudh: Will merge but make changes as necessary based on UX's reply

PR #3973, opened 2022-11-30
ROX-9115: Saner tag handling

gavin-stackrox:
## Description

Non groovy tests are particularly reliant on the output of `make tag`. This PR allows that to be overriden by the e2e test runner to more closely align with how CI e2e functions. It also drops `-dirty` if present to facilitate a test development workflow where tests code is changed in a working directory and tested against the most recent commit.

Testing found that the `config_only` option was working by accident due to poor choice of names :(. And when fixed I had to remove the separate output prefix for config & test because `something | other` is a subshell and export is lost from `something`.

## Checklist

## Testing Performed

- [x] missing image
- [x] -dirty tags
- [x] Full QA
- [x] Full NonGroovy E2e
- [x] Full QA - master tag
- [x] Full NonGroovy E2e - master tag
- [x] QA DeploymentTest only

### DIFF
--- a/make/env.mk
+++ b/make/env.mk
@@ -43,3 +43,8 @@ GOTAGS := $(RELEASE_GOTAGS)
 TAG := $(CIRCLE_TAG)
 endif
 endif
+
+# Support `make tag` override for non CI testing
+ifneq ($(TAG_OVERRIDE),)
+TAG := $(TAG_OVERRIDE)
+endif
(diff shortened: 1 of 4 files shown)


janisz (approved):
Do we need to update docs?

gavin-stackrox:
> Do we need to update docs?

I think the in-script `usage()` has as much info as it can hold. I'll add a README.md and link to it with more usage and background.

gavin-stackrox:
'Full NonGroovy E2e' test runs are pretty flaky. In a future PR I plan to make it so the separate chunks of 'non groovy' tests can be run independently.

PR #3984, opened 2022-12-01
ROX-13714: Implement a new rate limited logger, and use it to log failed auth messages.

clickboo:
## Description
1. Added a new rate limited logger
2. Added support to rate limit failed auth messages by hostname
3. Added hostname to failed auth log messages

## Checklist

## Testing Performed
Manual using curl on cluster.

### DIFF
--- a/pkg/grpc/authn/basic/extractor.go
+++ b/pkg/grpc/authn/basic/extractor.go
@@ -5,16 +5,26 @@ import (
 	"encoding/base64"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/pkg/auth/authproviders"
+	"github.com/stackrox/rox/pkg/errox"
 	"github.com/stackrox/rox/pkg/grpc/authn"
 	"github.com/stackrox/rox/pkg/grpc/requestinfo"
 	"github.com/stackrox/rox/pkg/logging"
 )
 
-var log = logging.LoggerForModule()
+const (
+	cacheSize          = 500
+	rateLimitFrequency = 5 * time.Minute
+	logBurstSize       = 5
+)
+
+var (
+	log = logging.NewRateLimitLogger(logging.LoggerForModule(), cacheSize, 1, rateLimitFrequency, logBurstSize)
+)
 
 // Extractor is the identity extractor for the basic auth identity.
 type Extractor struct {
@@ -52,11 +62,16 @@ func (e *Extractor) IdentityForRequest(ctx context.Context, ri requestinfo.Reque
 
 	username, password, err := parseBasicAuthToken(basicAuthToken)
 	if err != nil {
-		log.Warnf("failed to parse basic auth token: %s", err)
+		log.WarnL(ri.Hostname, "failed to parse basic auth token from %q: %v", ri.Hostname, err)
 		return nil, errors.New("failed to parse basic auth token")
 	}
 
-	return e.manager.IdentityForCreds(ctx, username, password, e.authProvider)
+	id, err := e.manager.IdentityForCreds(ctx, username, password, e.authProvider)
+	if errors.Is(err, errox.NotAuthorized) {
+		log.WarnL(ri.Hostname, "%q: %v", ri.Hostname, err)
+		return nil, err
+	}
+	return id, err
 }
 
 // NewExtractor returns a new identity extractor for basic auth.
--- a/pkg/logging/rate_limited_logger.go
+++ b/pkg/logging/rate_limited_logger.go
@@ -1,0 +1,68 @@
+package logging
+
+import (
+	"time"
+
+	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/time/rate"
+)
+
+// RateLimitedLogger wraps a zap.SugaredLogger that supports rate limiting.
+type RateLimitedLogger struct {
+	*Logger
+	frequency    float64
+	burst        int
+	rateLimiters *lru.Cache
+}
+
+// NewRateLimitLogger returns a rate limited logger
+func NewRateLimitLogger(l *Logger, size int, logLines int, interval time.Duration, burst int) *RateLimitedLogger {
+	cache, err := lru.New(size)
+	if err != nil {
+		l.Errorf("unable to create rate limiter cache for logger in module %q: %v", l.module.name, err)
+		return nil
+	}
+	return &RateLimitedLogger{
+		l,
+		float64(logLines) / interval.Seconds(),
+		burst,
+		cache,
+	}
+}
+
+// ErrorL logs a templated error message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) ErrorL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
+		rl.Errorf(template, args...)
+	}
+}
+
+// WarnL logs a templated warn message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) WarnL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
+		rl.Warnf(template, args...)
+	}
+}
+
+// InfoL logs a templated info message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) InfoL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
+		rl.Infof(template, args...)
+	}
+}
+
+// DebugL logs a templated debug message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) DebugL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
+		rl.Debugf(template, args...)
+	}
+}
+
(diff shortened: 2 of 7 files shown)


clickboo:
@rukletsov @mtodor @theencee Ping

pkg/logging/rate_limited_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return &RateLimitedLogger{
+		l,
+		cache,
+	}
+}
+
+// SetLimiter sets how many logLines are emitted in a given interval for a specific identifier.
+func (rl *RateLimitedLogger) SetLimiter(limiter string, logLines float64, interval time.Duration, burst int) {
+	if logLines == float64(burst) {
```
rukletsov, line 32: What does this do?

pkg/logging/rate_limited_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	*Logger
+	rateLimiters *lru.Cache
+	// needs a mutex to protect burst count increase? //@TODO: khushboo
+}
+
+// NewRateLimitLogger returns a rate limited logger
+func NewRateLimitLogger(l *Logger, size int) *RateLimitedLogger {
+	cache, err := lru.New(size)
+	if err != nil {
+		l.Errorf("unable to create rate limiter cache for logger in module %q: %q", l.module.name, err)
```
rukletsov, line 21:
```suggestion
		l.Errorf("unable to create rate limiter cache for logger in module %q: %v", l.module.name, err)
```

pkg/logging/rate_limited_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	// Check first, and only add the limiter if it doesn't exist already.
+	_, _ = rl.rateLimiters.ContainsOrAdd(limiter, rate.NewLimiter(rate.Limit(logLines/interval.Seconds()), burst))
+}
+
+// ErrorL logs a templated error message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) ErrorL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
+		rl.Errorf(template, args...)
+	}
```
rukletsov, line 43: I'd add an `else` path here and below printing a—also rate-limited!—note saying that log entries are being rate-limited.

pkg/logging/rate_limited_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	if logLines == float64(burst) {
+		burst++
+	}
+	// Check first, and only add the limiter if it doesn't exist already.
+	_, _ = rl.rateLimiters.ContainsOrAdd(limiter, rate.NewLimiter(rate.Limit(logLines/interval.Seconds()), burst))
+}
+
+// ErrorL logs a templated error message if allowed by the rate limiter corresponding to the identifier
+func (rl *RateLimitedLogger) ErrorL(limiter string, template string, args ...interface{}) {
+	if rl.allowLog(limiter) {
```
rukletsov, line 41: What do you think about removing `limiter` from the interface of `RateLimitedLogger` and calculating it on the fly via `fmt.Sprintf(template, args...)`? This will guarantee there is a rate limiter for every log entry, ensure different errors from the same hostname are not dropped, and simplify the interface.

pkg/grpc/authn/basic/extractor.go
```diff
@@ -52,11 +56,17 @@ func (e *Extractor) IdentityForRequest(ctx context.Context, ri requestinfo.Reque
 
 	username, password, err := parseBasicAuthToken(basicAuthToken)
 	if err != nil {
-		log.Warnf("failed to parse basic auth token: %s", err)
+		log.SetLimiter(ri.Hostname, 1, 5*60*time.Second, 5)
+		log.WarnL(ri.Hostname, "failed to parse basic auth token from %q: %q", ri.Hostname, err)
```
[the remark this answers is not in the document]
clickboo: Done

pkg/logging/rate_limited_logger.go
clickboo: Wouldn't that defeat the purpose of rate limiting the log messages in the first place?

pkg/logging/rate_limited_logger.go
clickboo: 1) We decided in our conversation to rate limit on hostname so hence the limiter is hostname
2) As for different errors from the same host name - only auth failures are being rate limited. Any other error is not. So they will continue to be logged as is.

pkg/logging/rate_limited_logger.go
theencee: So generally the rate limiter system would log when things are being rate limited. Otherwise it's hard to tell if it's due to rate limits vs low traffic vs bug. 

But the challenge here is that we are rate limiting logs, so logging that is counter-productive.

I can think of a few alternatives:

1. Log that we are rate limiting in _debug_ mode. Pros: No log spam in prod, we can still see logs when debugging. Con: If we run into an issue in prod, we can't tell without switching logs to debug mode. Debug logs are noisy.

2. Create a new (global) rate limiter, and use that to log something like "Rate limiting error logs including this log". 

3. Write the "Rate limiting error logs" message on a timer every 5 minutes.

I'm leaning towards #1 purely because it's easy to comprehend. Thoughts?

clickboo: +1, on 1. Other options are overkill in the light of an ingress rate limiter which has to be implemented for GA.

pkg/logging/rate_limited_logger.go
clickboo: Removed

mtodor (commented):
I have looked at the implementation, and I would suggest some bigger changes.

The main concern for me is `SetLimiter`.
- we need to call it before every log line. Otherwise, logs could be ignored
- the first call of `SetLimiter` defines the limit for all logs afterward

I would suggest defining rate limiting when an instance of `NewRateLimitLogger`. That allows us to create logs with different rates and use them. i.e.
```
logRateLimit5min = logging.NewRateLimitLogger(logging.LoggerForModule(), 100, 1, 5*time.Minute, 1)
logRateLimit10sec = logging.NewRateLimitLogger(logging.LoggerForModule(), 100, 1, 10*time.Second, 1)
```

And then we can call log.
```
logRateLimit5min.WarnL(ri.Hostname, "Token validation failed for hostname %v: %v", ri.Hostname, err)
```

This would require a change in `allowLog`. We should create a limiter if it does not exist. i.e.
```
func (rl *RateLimitedLogger) allowLog(limiter string) bool {
   if lim, ok := rl.rateLimiters.Get(limiter); ok {
      return lim.(*rate.Limiter).Allow()
   }

   rl.rateLimiters.Add(limiter, rate.NewLimiter(rl.baseLimit, rl.burst))

   return true
}
```
I assume that `rl.baseLimit` and `rl.burst` are created with a constructor.

@clickboo What do you think about these suggestions?

pkg/logging/rate_limited_logger.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"time"
+
+	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/time/rate"
+)
+
+// RateLimitedLogger wraps a zap.SugaredLogger that supports rate limiting.
+type RateLimitedLogger struct {
```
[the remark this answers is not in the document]
clickboo: What would the test do? Test that SetLimiter creates a cache entry?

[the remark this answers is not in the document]
mtodor: @clickboo Test would check that rate limiting works as expected. A simplified example:
```
// log has a rate limit of 2 seconds

log.InfoL("localhost", "Log entry 1")
log.InfoL("localhost", "Log entry 1")
log.InfoL("localhost", "Log entry 2")
log.InfoL("localhost", "Log entry 2")

time.Sleep(3*time.Second)

log.InfoL("localhost", "Log entry 1")
log.InfoL("localhost", "Log entry 1")
log.InfoL("localhost", "Log entry 2")
log.InfoL("localhost", "Log entry 2")
```

The output should be:
```
Log entry 1
Log entry 2
Log entry 1
Log entry 2
```

The only problem could be if we can easily create a logger that will output to buffer that we can compare afterward.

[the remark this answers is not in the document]
clickboo: @mtodor I believe this is overkill ^ and a manual test on this one should suffice, especially given this is a short term solution to ingress rate limiting.

mtodor (approved):
I have looked at the solution. It looks good. I like what you did here.

I would not completely rely on ingress. Ingres is something that we will use in ACSCS. And with that in place, we don't even need this functionality. This PR is really relevant for on-prem deployments too. For on-prem deployments, we don't know if the customer's infrastructure has such kind of protection or not. We know that without rate-limiting logs, there is a problem that central crashes because of accumulated logs. As I remember from our meeting, that's one reason we are adding this functionality in central.

There is one thing that I'm interested in, and I don't know if that is relevant for this implementation because it's not described in the ticket, so it's hard to know.

Hypothetical case. Let's say we have code like this:
```
log.WarnL(ri.Hostname, "Record 1")
// several lines of code
log.WarnL(ri.Hostname, "Record 2")
```

Our current implementation will only output the following:
```
Record 1
```

2nd log line will never be output, regardless of execution reaching it. Could that be a problem that we have partial context and we are losing important information?

Besides that, we can get this in. If we notice that additional functionality is required or that we can improve something, we can always add it.

PR #3986, opened 2022-12-01
ROX-13735: create consistent data structure for network graph node

sachaudh:
## Description

This PR makes a change in how we transform the data to be used for the topology node model. This also includes adding the grouping for external nodes. Made some updates to different files to make things work with Typescript as well

Screenshot 2022-12-01 at 8 50 08 PM

## Checklist


ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -1,6 +1,17 @@
...
+    CIDRBlockData,
+    CustomModel,
+    CustomNodeData,
+    CustomNodeModel,
+    DeploymentData,
+    ExternalEntitiesData,
+    ExternalNodeModel,
+    NamespaceData,
+    NamespaceNodeModel,
+} from '../types/topology.type';
```
vjwilson, line 14: Question about naming structure:

Because there are `NamespaceNodeModel` and `ExternalNodeModel`, that makes me wonder if `CustomNodeModel` only applies to deployments ?

If that's the case, would it be better to prefix all the "Custom..." things `Deployment` instead, like `DeploymentNodeModel` ?

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -16,19 +27,56 @@ function getNameByEntity(entity: NetworkEntityInfo): string {
...
+            ...entity,
+            type: 'EXTERNAL_ENTITIES',
+        };
+        return data;
+    }
+    const data: CIDRBlockData = {
+        ...entity,
+        type: 'CIDR_BLOCK',
+    };
+    return data;
```
vjwilson, line 49: nit: Would it be more clear to make `DEPLOYMENT` the default return type ?

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -0,0 +1,70 @@
+import { Model, NodeModel } from '@patternfly/react-topology';
+
+import { ListenPort } from 'types/networkFlow.proto';
+
+type Override<T1, T2> = Omit<T1, keyof T2> & T2;
```
vjwilson, line 5: TIL

vjwilson (commented):
I just had some naming comments. I'd fine with the tenor of these changes.

I would like @alwayshooin to validate this PR, as she is more familiar with the current data model than I.

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
sachaudh: Oh actually we have 5 different types of nodes:
1. `Namespace`
2. `Deployment`
3. `External`
4. `External Entities`
5. `CIDR Block`

I made `CustomNodeModel` as a type that can be one of these 5 

The hierarchy is as follows:
`Namespace` -- `groups` --> `Deployment`
`External` -- `groups` --> `External Entities`
`External` -- `groups` --> `CIDR Block`

So `Namespace` and `External` are the group nodes and the rest are just nodes (possible children)

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
sachaudh: Hmm, yea we can do that. I wanted to actually make the default be either throwing an error or returning another value, but Typescript was acting up because of that

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
sachaudh: I was thinking this `Override` type might be useful to have somewhere more global so it can be used in other instances. What do you think? (cc: @vjwilson @alwayshooin @pedrottimark @dvail @bradr5)

dvail: +1

I've been putting things like this in `src/utils/type.utils.ts` as they come up.

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
dvail: Speaking of `type.utils.ts`, you could use the `ensureExhaustive` function with a switch/case here to make it so that it is impossible for the app to compile if it would hit the default case:
```typescript
function getDataByEntityType(entity: NetworkEntityInfo, policyIds: string[]): CustomNodeData {
    switch (entity.type) {
        case 'DEPLOYMENT':
            return { ...entity, policyIds };
        case 'INTERNET':
            return { ...entity, type: 'EXTERNAL_ENTITIES' };
        case 'EXTERNAL_SOURCE':
            return { ...entity, type: 'CIDR_BLOCK' };
        default:
            return ensureExhaustive(entity);
    }
}
```

If you miss a case, or if the `NetworkEntityInfo["type"]` field gets an additional value down the line, the default case here will result in a type error.

sachaudh: Oh that's really useful! Let me try that out as well

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -51,7 +52,8 @@ function DeploymentSideBar({ deploymentId, nodes, edges }: DeploymentSideBarProp
     const numExternalFlows = getNumExternalFlows(nodes, edges, deploymentId);
     const numInternalFlows = getNumInternalFlows(nodes, edges, deploymentId);
     const listenPorts = getListenPorts(nodes, deploymentId);
-    const deploymentPolicyIds = deploymentNode?.data?.policyIds || [];
+    const deploymentPolicyIds =
+        deploymentNode?.data.type === 'DEPLOYMENT' ? deploymentNode?.data?.policyIds : [];
```
alwayshooin, line 56: oh just curious why this was necessary?

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ListenPort } from 'types/networkFlow.proto';
+import { Override } from 'utils/type.utils';
+
+export type CustomModel = Override<Model, { nodes?: CustomNodeModel[] }>;
+
+export type CustomNodeModel =
+    | NamespaceNodeModel
+    | DeploymentNodeModel
+    | ExternalNodeModel
+    | ExternalEntitiesNodeModel
```
alwayshooin, line 12: what's the difference between ExternalNodeModel and ExternalEntitiesNodeModel here? is ExternalEntitiesNodeModel the External to Cluster node?

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -0,0 +1,69 @@
...
+export type DeploymentData = {
+    type: 'DEPLOYMENT';
+    id: string;
+    deployment: {
+        cluster: string;
+        listenPorts: ListenPort[];
+        name: string;
+        namespace: string;
+    };
+    policyIds: string[];
```
alwayshooin, line 47: i think you can use the `DeploymentNetworkEntityInfo` type here

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -22,13 +43,28 @@ export const graphModel = {
...
+        children: [],
+        group: true,
+        label: 'External to cluster',
+        style: { padding: 15 },
+        data: {
+            type: 'EXTERNAL',
+            collapsible: true,
+            showContextMenu: false,
+        },
+    };
```
alwayshooin, line 66: this will render as an empty group even if there are no external nodes, btw

alwayshooin (approved):
lgtm! just a few comments/questions in the code.

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
sachaudh: I had to do that so typescript knew it had the fields

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
sachaudh: Screenshot 2022-12-05 at 4 41 25 PM

External refers to the grouping. This is based on the types defined in the `NetworkGraph` component.

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
sachaudh: Yea true, I can do that. We have those types defined in `networkFlow.proto.ts`. The thing is that I do want to keep the types different. For example, in that `networkFlow.proto.ts` we use the type `INTERNET` but I would want to use `EXTERNAL_ENTITIES`. Since the deployment is pretty much the same, I could use the already existing type, but then it would basically be the only one using the existing type. I decided not to use it because it seemed cleaner and simpler to just define it here instead.

PR #3987, opened 2022-12-01
Disable Edit button for uneditable integrations

vjwilson:
## Description

This little enhancement grew out of confusion during the HackFest for the new ACS Cloud Service. Kylie provided the UX pattern that will be used in new features, too, going forward.

This was a pain point for new users with the Cluster Init Bundle, but the same pattern applies to API Tokens.

(Note: We have to use the `isAriaDisabled` prop for the PatternFly component, instead of its `isDisabled` prop, in order for the Tooltip to be displayed. Reference: https://www.patternfly.org/v4/components/button#aria-disabled-examples

## Checklist

## Testing Performed

Cluster Init Bundle
Screen Shot 2022-12-01 at 3 30 26 PM

API Token
Screen Shot 2022-12-01 at 3 30 11 PM

No tooltip when edit button is clickable
Screen Shot 2022-12-01 at 3 37 06 PM

### DIFF
--- a/ui/apps/platform/src/Containers/Integrations/IntegrationPage.tsx
+++ b/ui/apps/platform/src/Containers/Integrations/IntegrationPage.tsx
@@ -9,10 +9,14 @@ import {
     Breadcrumb,
     BreadcrumbItem,
     Divider,
+    Tooltip,
 } from '@patternfly/react-core';
 
 import { integrationsPath } from 'routePaths';
-import { getIntegrationLabel } from 'Containers/Integrations/utils/integrationUtils';
+import {
+    getEditDisabledMessage,
+    getIntegrationLabel,
+} from 'Containers/Integrations/utils/integrationUtils';
 import PageTitle from 'Components/PageTitle';
 import LinkShim from 'Components/PatternFly/LinkShim';
 import BreadcrumbItemLink from 'Components/BreadcrumbItemLink';
@@ -35,6 +39,8 @@ function IntegrationPage({ title, children }: IntegrationPageProps): ReactElemen
     const integrationsListPath = `${integrationsPath}/${source}/${type}`;
     const integrationEditPath = `${integrationsPath}/${source}/${type}/edit/${id as string}`;
 
+    const editDisabledMessage = getEditDisabledMessage(type);
+
     return (
         <>
             <PageTitle title={title} />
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/Integrations/IntegrationPage.tsx
```diff
@@ -55,13 +61,26 @@ function IntegrationPage({ title, children }: IntegrationPageProps): ReactElemen
...
-                            >
-                                Edit
-                            </Button>
+                            {editDisabledMessage ? (
+                                <Tooltip content={editDisabledMessage}>
+                                    <Button
+                                        variant={ButtonVariant.secondary}
+                                        component={LinkShim}
+                                        href={integrationEditPath}
+                                        isAriaDisabled={!!editDisabledMessage}
```
pedrottimark, line 70: Any reason why not `isAriaDisabled` to emphasize the difference between the `Button` elements?

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
```diff
@@ -53,6 +53,16 @@ export function clearStoredCredentials<I extends IntegrationBase>(
     return integration;
 }
 
+export function getEditDisabledMessage(type) {
+    if (type === 'clusterInitBundle') {
+        return 'This Cluster Init Bundle can not be edited. Create a new Cluster Init Bundle or delete an existing one';
```
pedrottimark, line 58: * **This** sounds like a contrast with that other which can be edited.
* **Create** or **delete** is an unusual contrast. Is the intended meaning to delete an existing one if it has become obsolete after you create a new one? If yes, ask the docs team to suggest a clear and concise statement. Probably as separate sentences. If docs do not already have a starting point for the wording, then it is important to fill the gap there too.

pedrottimark (approved):
Good attention to detail. 2 comments for your consideration, but not a blocker to merge.

ui/apps/platform/src/Containers/Integrations/utils/integrationUtils.ts
vjwilson: That language is already in the response. I just moved it out to the tooltip. I'm not opposed to new language, but I'd rather that be a follow-on task, to prevent this Occam's Razor solution from expanding into a long discussion.

ui/apps/platform/src/Containers/Integrations/IntegrationPage.tsx
vjwilson: I thought about that redundancy, but then I considered that the PatternFly idiom might change in the future, and this guards the disabled state breaking.

PR #3988, opened 2022-12-01
CI: Wait longer for images

gavin-stackrox:
## Description

We've had a few (well at least two) recent flakes where the images were slow to build. Waiting 15 minutes more might get them and it is no harm.

## Checklist

## Testing Performed

CI is sufficient


keyallis (approved):
LGMT

PR #3992, opened 2022-12-02
ROX-12440: Use a fixed kuttl

porridge:
## Description

Looks like the fix might take a while to be merged upstream, as they are in the middle of a CI migration.
Use a build from my private repo for now.

## Checklist

## Testing Performed

See the result of CI for penultimate commit - logs were fetched as expected.
I also made sure this works OK from one's workstation.
I also made sure darwin binaries for intel and arm are present in [the release](https://github.com/porridge/kuttl/releases/tag/v0.14.0-kubeconfig).

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -182,8 +182,8 @@ operator-sdk: ## Download operator-sdk necessary for scaffolding and bundling.
 	$(GET_GITHUB_RELEASE_FN); \
 	get_github_release $(OPERATOR_SDK) $${OPERATOR_SDK_URL}
 
-KUTTL_VERSION = 0.13.0
-KUTTL_UPSTREAM = kudobuilder
+KUTTL_VERSION = 0.14.0-kubeconfig
+KUTTL_UPSTREAM = porridge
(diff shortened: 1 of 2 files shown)


operator/Makefile
```diff
@@ -182,8 +182,8 @@ operator-sdk: ## Download operator-sdk necessary for scaffolding and bundling.
 	$(GET_GITHUB_RELEASE_FN); \
 	get_github_release $(OPERATOR_SDK) $${OPERATOR_SDK_URL}
 
-KUTTL_VERSION = 0.13.0
-KUTTL_UPSTREAM = kudobuilder
+KUTTL_VERSION = 0.14.0-kubeconfig
+KUTTL_UPSTREAM = porridge
```
SimonBaeumer, line 186: Can we move kuttl upstream to `stackrox` org instead for future upstream fixes?

SimonBaeumer (approved):

operator/Makefile
porridge: I'll try to remember to do so :wink:

PR #3995, opened 2022-12-02
[ROX-13741] Fix migration for active components

mtodor:
## Description

We have discovered this bug during testing of the 3.73.0 release.

## Checklist

## How to reproduce

1. Create Openshift 4 cluster
2. Create Central via operator
3. Create Secured Cluster via operator that connects to previously created central
4. After everything is connected and RocksDB is populated - edit Central CR and change `isEnabled` to `Enabled` for `db` property.
5. This will deploy the database and change settings in Central to use it. Migration will start and the problem should appear

## Testing

1. Execute unit tests for relevant migration but with the setting `batchAfter` to `1` so that Copy path is executed
2. It fails without this fix - an example of a test run can be found here: #3996 
3. With this fix, it works properly, and with executing the test with coverage, we can confirm that this execution path was used

### DIFF
--- a/migrator/migrations/n_05_to_n_06_postgres_active_components/postgres/postgres_plugin.go
+++ b/migrator/migrations/n_05_to_n_06_postgres_active_components/postgres/postgres_plugin.go
@@ -148,7 +148,7 @@ func (s *storeImpl) copyFromActiveComponents(ctx context.Context, tx pgx.Tx, obj
 
 			obj.GetId(),
 
-			obj.GetDeploymentId(),
+			pgutils.NilOrUUID(obj.GetDeploymentId()),
 
 			obj.GetComponentId(),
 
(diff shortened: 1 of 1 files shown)


mtodor:
* master
  * **PR #3995** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/3995" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/3995?utm_source=stack-comment).

dashrews78 (approved):
I looked through all the other non-generated migrations.  It appears this is the only one we missed.

dashrews78 (changes requested):
line 80

dashrews78 (approved):

PR #3997, opened 2022-12-02
ROX-13469: Skip auth integration tests

pedrottimark:
## Description

Three strikes and auth tests are out.

1. Replaced main Dashboard with System Configuration to reduce need to mock responses in #3917
2. Added the following line to the failing test in #3985

    ```js
    localStorage.removeItem('access_token'); // replace possible valid token left over from previous test file
    ```

    Behavior **before** the change: failures in either postgress or non-postgress, but not usually both for any particular merge, for the **first** test in the file:
    `should redirect user to login page, authenticate and redirect to the requested page`

3. Behavior **after** the change: failures often in both postgress and non-postgress for the **second** test in the file:
     `'should allow authenticated user to enter'`

## Checklist
- [x] Skipped integration tests

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/integration/auth.test.js
+++ b/ui/apps/platform/cypress/integration/auth.test.js
@@ -33,7 +33,7 @@ describe.skip('Authentication', () => {
     };
 
     it('should redirect user to login page, authenticate and redirect to the requested page', () => {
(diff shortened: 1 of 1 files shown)


gavin-stackrox (approved):
Yep.

PR #4000, opened 2022-12-02
Factor out helpers for Access Control integration tests

pedrottimark:
## Description

Apply consistent patterns and prevent timing failures.

Follow up replacement for accessControlAuthProviders.test.js in #2805

This test folder has 11 `cy.visit` method calls, similar size to 10 in userinfo.test.js which were replaced in #3869

### Background information

Here are links to cypress data structures:
* `routeMatcher` https://docs.cypress.io/api/commands/intercept#Matching-with-RouteMatcher
* `staticResponse` https://docs.cypress.io/api/commands/intercept#StaticResponse-objects

### Naming convention

Distinguish the **scope** of constants and functions:
1. Test-specific: `it` block for test or module scope in test file for tests.
2. Container-specific: whatever.helpers.js or whatever.selectors.js file in cypress/integration/whatever subfolder. Only rare import from other containers, like `interactAndVisitWhatever` function (for example, from Risk to Network Graph).
3. Global
    * cypress/helpers/whatever.js for example, basicAuth.js or features.js
    * cypress/selectors/whatever.js for example, navigation.js or pagination.js

End goals:
1. Move most of cypress/helpers/whatever.js to container subfolders as whatever.helpers.js files.
2. Move most of cypress/constants/whatever.js to container subfolders as whatever.selectors.js files.
3. Encapsulate most of apiEndpoints.js in container-specific helpers files.

Here is an example of contrast between global helpers file and container-specific helpers and selectors files:

```js

import {
    accessScopesKey as entitiesKey,
    assertAccessControlEntityDoesNotExist,
    clickEntityNameInTable,
    visitAccessControlEntities,
    visitAccessControlEntitiesWithStaticResponseForPermissions,
    visitAccessControlEntity,
} from './accessControl.helpers';
import { selectors } from './accessControl.selectors';
```

### Changed files

1. Rename cypress/constants/AccessControlPage.js as cypress/integration/accessControl/accessControl.selectors.js
    * Because only Access Control tests import from the file, move it into the container subfolder and name as counterpart to helpers file.
    * Delete whateverUrl because helpers file encapsulates page addresses.
    * Delete selectors either used only in helpers file or obvious like `th` elements.

2. Edit cypress/helpers/request.js
    * Return array of objects from `cy.wait` method in case test needs to assert.

3. Add cypress/integration/accessControl/accessControl.helpers.js
    * Apply light version of entities pattern from Configuration Management or Vulnerability Management helpers.
    * Write minimum number of helpers for create, update, and delete, because:
        * Workflow containers rarely mutate data.
        * API has are some irregularities, like POST /v1/roles/roleName insteads of /v1/roles
        * Test coverage is limited: too bad, so sad.

4. Edit cypress/integration/accessControl/accessControlAccessScopes.test.js
    * Use literal strings for obvious selectors like `th` element.
    * Keep import selectors for form.

5. Edit cypress/integration/accessControl/accessControlAuthProviders.test.js
    * Ditto the previous item.
    * Click **Cancel** button at end of tests which do not save created auth provider.

6. Edit cypress/integration/accessControl/accessControlPermissionSets.test.js

7. Edit cypress/integration/accessControl/accessControlRoles.test.js

## Checklist

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/helpers/request.js
+++ b/ui/apps/platform/cypress/helpers/request.js
@@ -64,8 +64,10 @@ export function waitForResponses(routeMatcherMap) {
     if (routeMatcherMap) {
         const aliases = Object.keys(routeMatcherMap).map((alias) => `@${alias}`);
 
-        cy.wait(aliases);
+        return cy.wait(aliases);
     }
+
+    return [];
 }
 
 /**
@@ -84,5 +86,5 @@ export function interactAndWaitForResponses(
 
     interactionCallback();
 
-    waitForResponses(routeMatcherMap);
+    return waitForResponses(routeMatcherMap);
 }
--- a/ui/apps/platform/cypress/helpers/request.js
+++ b/ui/apps/platform/cypress/helpers/request.js
@@ -59,6 +59,7 @@ export function interceptRequests(routeMatcherMap, staticResponseMap) {
  * Wait for responses after initial page visit or subsequent interaction.
  *
  * @param {Record<string, { method: string, url: string }>} [routeMatcherMap]
+ * @returns {{ request: Record<string, unknown>, response: Record<string, unknown>}[]}
  */
 export function waitForResponses(routeMatcherMap) {
     if (routeMatcherMap) {
@@ -76,6 +77,7 @@ export function waitForResponses(routeMatcherMap) {
  * @param {() => void} interactionCallback
  * @param {Record<string, { method: string, url: string }>} [routeMatcherMap]
  * @param {Record<string, { body: unknown } | { fixture: string }>} [staticResponseMap]
+ * @returns {{ request: Record<string, unknown>, response: Record<string, unknown>}[]}
  */
 export function interactAndWaitForResponses(
     interactionCallback,
--- a/ui/apps/platform/cypress/integration/accessControl/accessControlAccessScopes.test.js
+++ b/ui/apps/platform/cypress/integration/accessControl/accessControlAccessScopes.test.js
@@ -1,13 +1,16 @@
 import { accessScopesUrl, selectors } from '../../constants/AccessControlPage';
-import {
-    accessScopes as accessScopesApi,
-    permissions as permissionsApi,
-} from '../../constants/apiEndpoints';
 
 import withAuth from '../../helpers/basicAuth';
 import { getRegExpForTitleWithBranding } from '../../helpers/title';
 
-const h1 = 'Access Control';
+import {
+    accessScopesKey as entitiesKey,
+    clickEntityNameInTable,
+    visitAccessControlEntities,
+    visitAccessControlEntitiesWithStaticResponseForPermissions,
+    visitAccessControlEntity,
+} from './accessControl.helpers';
+
 const h2 = 'Access scopes';
 
 const defaultNames = ['Unrestricted', 'Deny All'];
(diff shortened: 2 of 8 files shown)


ui/apps/platform/cypress/integration/accessControl/accessControlRoles.test.js
```diff
@@ -1,75 +1,56 @@
...
-const rolesApi = {
-    list: '/v1/roles',
-};
 
-const h1 = 'Access Control';
-const h2 = 'Roles';
+import {
+    assertAccessControlEntityDoesNotExist,
+    clickEntityNameInTable,
+    rolesKey as entitiesKey,
```
dvail, line 6: Nit: why the alias for these keys? Reading below I see
```typescript
        visitAccessControlEntities(entitiesKey);
```
which at first glance looks like `entitiesKey` is variable somewhere else in the file. Without the alias it would be more clear which entity we are visiting.
```typescript
        visitAccessControlEntities(rolesKey);
```
At least if the pattern to only visit one access control entity per test file is the plan for the future.

ui/apps/platform/cypress/integration/accessControl/accessControlAccessScopes.test.js
```diff
@@ -1,83 +1,65 @@
...
-        cy.get(selectors.navLinkCurrent).should('have.text', h2);
+        cy.get('button:contains("Create access scope")');
 
-        cy.contains(selectors.h2, /^\d+ results? found$/).should('exist');
-        cy.get(selectors.list.createButton).should('have.text', 'Create access scope');
-
-        cy.get(`${selectors.list.th}:contains("Name")`);
-        cy.get(`${selectors.list.th}:contains("Description")`);
-        cy.get(`${selectors.list.th}:contains("Roles")`);
+        cy.get('th:contains("Name")');
```
dvail, line 40: A much appreciated removal of indirection. 😃

dvail (approved):
LGTM, the colocation of endpoints, selectors, etc is a good trend for places that don't need wider access to those values.

ui/apps/platform/cypress/integration/accessControl/accessControlRoles.test.js
pedrottimark: Good question. First impression from reader viewpoint is valuable to know.

Although test files for these “entities” have more differences than for workflow entities, some tests are similar:
* `'displays alert if no permission'`
* `'list has table head cells'`
* `'displays message instead of form if entity id does not exist'`

The pro and con are worth consideration whether during copy-paste of tests to edit:
* only what is **different** for an entity
* also distinguish **which** entity: `accessScopesKey` versus `authProvidersKey` versus `permissionSetsKey` versus `rolesKey`

ui/apps/platform/cypress/helpers/request.js
```diff
@@ -59,13 +59,16 @@ export function interceptRequests(routeMatcherMap, staticResponseMap) {
...
  *
  * @param {Record<string, { method: string, url: string }>} [routeMatcherMap]
+ * @returns {{ request: Record<string, unknown>, response: Record<string, unknown>}[]}
  */
 export function waitForResponses(routeMatcherMap) {
     if (routeMatcherMap) {
         const aliases = Object.keys(routeMatcherMap).map((alias) => `@${alias}`);
 
-        cy.wait(aliases);
+        return cy.wait(aliases);
```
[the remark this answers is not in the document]
dvail: Aha, no need for improved comments, I just missed it. I was parsing for a `const <something> = saveUpdatedAuthProvider()` and didn't think to consider using the return value directly to take advantage of Cypress "yield".

PR #4001, opened 2022-12-02
Create collection results component

bradr5:
## Description

This add a live preview of currently selected deployments based off the rules selected

## Checklist

## Testing Performed

Before any rules have been created or a separate collection has been attached
Screenshot 2022-12-07 at 11 09 04 AM

Once a rule has been created, the first 10 deployments matching the rules should display with the option to view 10 more using the "view more" link
Screenshot 2022-12-07 at 11 09 20 AM

The ability to filter results will fetch deployments using the input field
Screenshot 2022-12-07 at 11 09 40 AM

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
@@ -1,7 +1,153 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState, useMemo } from 'react';
+import {
+    Button,
+    Divider,
+    EmptyState,
+    EmptyStateIcon,
+    EmptyStateVariant,
+    Flex,
+    FlexItem,
+    Select,
+    SearchInput,
+    SelectOption,
+    debounce,
+} from '@patternfly/react-core';
+import { ListIcon } from '@patternfly/react-icons';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import ResourceIcon from 'Components/PatternFly/ResourceIcon';
 
-function CollectionResults() {
-    return <></>;
+import { dryRunCollection } from 'services/CollectionsService';
+import { ListDeployment } from 'types/deployment.proto';
+
+function CollectionResults({ dryRunConfig }) {
+    const { isOpen, onToggle, closeSelect } = useSelectToggle();
+    const [selected, setSelected] = useState('Deployment');
+    const [filterText, setFilterText] = useState('');
+    const [isEndOfResults, setIsEndOfResults] = useState(false);
+    const [deployments, setDeployments] = useState<ListDeployment[]>([]);
+    const [hasRuleSelected, setHasRuleSelected] = useState(false);
+
+    function onRuleOptionSelect(_, value) {
+        setSelected(value);
+        setFilterText('');
+        closeSelect();
+    }
+
+    const fetchDryRun = useCallback(
+        (page: number) => {
+            const pageSize = 10;
+            const query = filterText ? `${selected}:${filterText}` : '';
+            const { request } = dryRunCollection(dryRunConfig, query, page, pageSize);
+            request
+                .then((results) => {
+                    setIsEndOfResults(results.length < 10);
+                    setDeployments((current) => (page === 0 ? results : [...current, ...results]));
+                    setHasRuleSelected(true);
+                })
+                .catch(() => {
+                    // TODO: indicate results not loading properly?
+                });
+        },
+        [dryRunConfig, filterText, selected]
+    );
+
+    useEffect(() => {
+        if (dryRunConfig.resourceSelectors?.[0]?.rules?.length > 0) {
+            fetchDryRun(0);
+        } else {
+            setHasRuleSelected(false);
+        }
+    }, [dryRunConfig, fetchDryRun]);
+
+    const onSearchInputChange = useMemo(
+        () =>
+            debounce((value: string) => {
+                setFilterText(value);
+            }, 800),
+        []
+    );
+
+    const currentPage = Math.floor((deployments.length - 1) / 10);
+
+    return (
+        <>
+            <Divider />
+            {!hasRuleSelected ? (
+                <EmptyState variant={EmptyStateVariant.xs}>
+                    <EmptyStateIcon icon={ListIcon} />
+                    <p>
+                        Add selector rules or attach existing collections to view resource matches
+                    </p>
+                </EmptyState>
+            ) : (
+                <Flex
+                    spaceItems={{ default: 'spaceItemsNone' }}
+                    alignItems={{ default: 'alignItemsCenter' }}
+                    className="pf-u-mt-lg"
+                >
+                    <FlexItem>
+                        <Select
+                            toggleAriaLabel="Select by name or label"
+                            isOpen={isOpen}
+                            onToggle={onToggle}
+                            selections={selected}
+                            onSelect={onRuleOptionSelect}
+                            isDisabled={false}
+                        >
+                            <SelectOption value="Deployment">Deployment</SelectOption>
+                            <SelectOption value="Namespace">Namespace</SelectOption>
+                            <SelectOption value="Cluster">Cluster</SelectOption>
+                        </Select>
+                    </FlexItem>
+                    <FlexItem grow={{ default: 'grow' }}>
+                        <SearchInput
+                            aria-label="Filter by name"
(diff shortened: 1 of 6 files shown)


ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -188,12 +188,28 @@ export type CollectionDryRunResponse = {
...
-    // TODO `ListDeployment` will make this impossible to paginate without loading the entire
-    // dataset client side. Ask [BE] if there is an efficient way to aggregate namespaces/clusters
-    // under a matching deployment name similar to the graphql query. An alternative might be to
-    // change the rendering of the list to not group deployments, but to sort alphabetically.
+    collectionRequest: CollectionRequest,
+    searchFilter: string,
+    page: number,
+    pageSize: number
 ): CancellableRequest<ListDeployment[]> {
+    const dryRunRequest: CollectionDryRunRequest = {
```
bradr5, line 196: What do you think of this change? I figured it would be nice to get the options object out of the component and into the service but I'm not sure if this call will eventually be used in other areas of the code and things like `sortOption` would need to be flexible

bradr5:
Lmk if we think some of the `CollectionResults` component should be broken down a bit. It doesn't appear too bloated to me but I could see an argument for splitting out something like `CollectionResultsFilter` or `CollectionResultsList`

ui/apps/platform/src/services/CollectionsService.ts
dvail: Yeah I like this, breaking apart the `CollectionDryRunRequest` and making it so the caller doesn't have to construct it itself is a good change, and I think more inline with other service functions in the app.

ui/apps/platform/src/services/CollectionsService.ts
```diff
@@ -188,12 +188,28 @@ export type CollectionDryRunResponse = {
...
+        ...collectionRequest,
+        options: {
+            filterQuery: {
+                query: searchFilter,
+                pagination: {
+                    offset: page * pageSize,
+                    limit: pageSize,
+                    sortOption: {
```
dvail, line 204: I think you can omit the `sortOption` in the API call which we should do instead if it works. I'd also make the `sortOption` an optional parameter to the function so that if we decide to reuse this elsewhere we have the ability to change it on demand.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,153 @@
...
+import { ListIcon } from '@patternfly/react-icons';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import ResourceIcon from 'Components/PatternFly/ResourceIcon';
 
-function CollectionResults() {
-    return <></>;
+import { dryRunCollection } from 'services/CollectionsService';
+import { ListDeployment } from 'types/deployment.proto';
+
+function CollectionResults({ dryRunConfig }) {
```
dvail, line 22: Can you add the type of this component's props and `export` it?

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,153 @@
...
+                })
+                .catch(() => {
+                    // TODO: indicate results not loading properly?
+                });
+        },
+        [dryRunConfig, filterText, selected]
+    );
+
+    useEffect(() => {
+        if (dryRunConfig.resourceSelectors?.[0]?.rules?.length > 0) {
```
dvail, line 55: We also need to account for the case where there are no rules, but the user has attached a collection. I would say that a collection is considered empty if:
- it has no `resourceSelectors`
or
- Every `resourceSelector` has zero rules _and_ the embedded collection array is empty

dvail: Also regarding the above, you should be able to calculate this right in the component's render call and remove the need to track `hasRuleSelected` in state.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,153 @@
...
+            const query = filterText ? `${selected}:${filterText}` : '';
+            const { request } = dryRunCollection(dryRunConfig, query, page, pageSize);
+            request
+                .then((results) => {
+                    setIsEndOfResults(results.length < 10);
+                    setDeployments((current) => (page === 0 ? results : [...current, ...results]));
+                    setHasRuleSelected(true);
+                })
+                .catch(() => {
+                    // TODO: indicate results not loading properly?
```
dvail, line 48: I think the error handling here is going to be combined with the current "onSave" error handling, since the possible errors that are returned from the dry run API should match what is returned when the user tries to save.

Lets create a new ticket for it though so that it doesn't hold up this PR.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,153 @@
...
+                </EmptyState>
+            ) : (
+                <Flex
+                    spaceItems={{ default: 'spaceItemsNone' }}
+                    alignItems={{ default: 'alignItemsCenter' }}
+                    className="pf-u-mt-lg"
+                >
+                    <FlexItem>
+                        <Select
+                            toggleAriaLabel="Select by name or label"
```
dvail, line 90:
```suggestion
                            toggleAriaLabel="Select an entity type to filter the results by"
```

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,153 @@
...
+    function onRuleOptionSelect(_, value) {
+        setSelected(value);
+        setFilterText('');
+        closeSelect();
+    }
+
+    const fetchDryRun = useCallback(
+        (page: number) => {
+            const pageSize = 10;
+            const query = filterText ? `${selected}:${filterText}` : '';
```
dvail, line 39: There is a `getRequestQueryStringForSearchFilter` util function that will create the search string for you that would come in handy if/when we apply multiple searches to this section.

dvail (changes requested):
This looks good! Just a handful of suggestions, with the main one being that we need to display collection results if the rules are empty but another collection has been attached.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
dvail: https://issues.redhat.com/browse/ROX-13780

dvail (approved):
LGTM. The final (hopefully) big rock for the collection feature on the UI side. 🎉

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
pedrottimark: Does error handling need to distinguish failure of the request itself versus backend validation error?

We might compare to policy dry run, although not assuming they have identical behavior.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -1,7 +1,157 @@
...
+    );
+
+    return (
+        <>
+            <Divider />
+            {!selectorRulesExist ? (
+                <EmptyState variant={EmptyStateVariant.xs}>
+                    <EmptyStateIcon icon={ListIcon} />
+                    <p>
+                        Add selector rules or attach existing collections to view resource matches
```
pedrottimark, line 83: For my understanding, no rules implies create/clone/edit and not view, correct?

ui/apps/platform/src/services/CollectionsService.ts
pedrottimark: Ditto.

Wow, you never know when API payloads and responses will split versus join:
* Instead of split `{ collection, options }`
* It joins/merges `options` property with collection properties for `CollectionDryRunRequest`

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
bradr5: No rules just implies not having made any changes to the "Collection Rules" section – I probably should've included that section in the screenshots

bradr5: I updated the screenshots to give a better representation of the logic

PR #4002, opened 2022-12-02
ROX-13708 Fix pagination of violating deployments in VM Policy table

dvail:
## Description

This adds pagination to the VM PolicyOverview table of deployments that are failing the policy.

## Approach

From what I can tell in staging and based on the headings of each section, the deployment table that is displayed when you click a row in the policy list should be identical to the table that is displayed when you click the "# failing deployments" text in the table cell.

VM -> Policies -> Click "Fixable severity at least important" row
image

VM -> Policies -> Click "29 failing deployments" link in "Fixable severity at least important" row
image

The general idea here is that we _remove_ the `deployment` subresolver from the entity query in VulnMgmtEntityPolicy and replace it with a separate query in the child component VulnMgmtPolicyOverview.

This has the downside of requiring an extra request when the details panel is opened.

The upsides are:
- We don't request the deployment list until it is needed
- We can actually do pagination of this secondary table
- When we paginate the second table, it doesn't require us to refetch the entire policy entity query each time the deployment query changes

## Checklist

## Testing Performed

Manual comparison of local branch and staging UI against staging data for both of the above pages via YARN_START_TARGET. For each policy with failing deployments, click the table row as well as the "# failing deployments" link and view the list of deployments in the table. Sorting by name, verify that the exact same deployments appear in each table. The UI should be identical across the two builds, with the exception of this PR having the data split into table pages.

Test against a scaled data cluster and ensure that very large lists of deployments do not cause a timeout and successfully paginate the data. (Note that the "related entities" side panel is not present in the screenshots below because the endpoint returns a `deploymentCount` of `0` due to the data being mocked - it is rendered correctly when running against staging data.)

Clicking on the table row:
image
image

Clicking on the failed deployments link:
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
@@ -64,7 +64,7 @@ const noop = () => {};
 const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
     const workflowState = useContext(workflowStateContext);
     const { isFeatureFlagEnabled } = useFeatureFlags();
-    const { page, onSetPage } = usePagination();
+    const [page, setPage] = useState(1);
     const [sort, setSort] = useState({
         field: entityPriorityField.DEPLOYMENT,
         reversed: false,
@@ -255,27 +255,25 @@ const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
                     if (loading && !policyData) {
                         return <Loader />;
                     }
-                    const { deployments, deploymentCount } = policyData;
+                    const failingDeployments = policyData.deployments;
+                    const totalCount = policyData.deploymentCount;
+                    const headerText = `${totalCount} ${pluralize(
+                        entityTypes.DEPLOYMENT,
+                        totalCount
+                    )} ${pluralizeHas(totalCount)} failed across this policy`;
                     return (
                         <div className="pdf-page pdf-stretch pdf-new flex shadow rounded relative bg-base-100 mb-4 mx-4">
                             <TableWidget
-                                header={`${deploymentCount} ${pluralize(
-                                    entityTypes.DEPLOYMENT,
-                                    deploymentCount
-                                )} ${pluralizeHas(deploymentCount)} failed across this policy`}
-                                rows={deployments}
+                                header={headerText}
+                                rows={failingDeployments}
                                 entityType={entityTypes.DEPLOYMENT}
                                 noDataText="No deployments have failed across this policy"
                                 className="bg-base-100"
                                 columns={getDeploymentTableColumns(workflowState)}
+                                /* TODO Remove `pageSize` */
                                 pageSize={2}
-                                parentPageState={{
-                                    page,
-                                    setPage: (newPage) => onSetPage(undefined, newPage),
-                                    totalCount: deploymentCount,
-                                }}
+                                parentPageState={{ page, setPage, totalCount }}
                                 sortHandler={setSort}
-                                defaultSorted={[sort]}
                             />
                         </div>
                     );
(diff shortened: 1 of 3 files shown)


dvail:
* master
  * **PR #4002** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4002" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4002?utm_source=stack-comment).

ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
```diff
@@ -122,20 +134,6 @@ const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
         </div>
     );
 
-    // @TODO: extract this out to make it re-usable and easier to test
```
dvail, line 125: I'm guessing that this logic was put in before the `"Policy Violated"` search option was implemented, and was just never removed when the "# failing deployments" link was implemented. As far as I can tell this client side filtering achieves the same thing the new (circa 2020 "new") option does.

ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
```diff
@@ -216,26 +214,70 @@ const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
...
+            }),
+            ...vulMgmtPolicyQuery,
+            pagination: queryService.getPagination(sort, page, 2 /* TODO Remove this param */),
+        };
+
         policyFindingsContent = (
-            <div className="pdf-page pdf-stretch pdf-new flex shadow rounded relativebg-base-100 mb-4 mx-4">
-                <TableWidget
-                    header={`${failingDeployments.length} ${pluralize(
+            <Query query={failingDeploymentQuery} variables={failingDeploymentQueryVariables}>
```
dvail, line 251: I'm using the old-style render props pattern for the GQL query here since the child component that is affected is being conditionally rendered. The alternatives seemed to be:

- Always pull the deployment data via a top level `useQuery` hook, even when we don't need it
- Extract the `<TableWidget>` here out into a new wrapped component specifically for this use case and use the hook there

This seemed like the lighter-touch approach.

ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
```diff
@@ -216,26 +214,70 @@ const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
...
+                        <div className="pdf-page pdf-stretch pdf-new flex shadow rounded relative bg-base-100 mb-4 mx-4">
+                            <TableWidget
+                                header={headerText}
+                                rows={failingDeployments}
+                                entityType={entityTypes.DEPLOYMENT}
+                                noDataText="No deployments have failed across this policy"
+                                className="bg-base-100"
+                                columns={getDeploymentTableColumns(workflowState)}
+                                /* TODO Remove `pageSize` */
+                                pageSize={2}
```
dvail, line 273: I've hard coded a page size of 2 for easier testing, but if this approach seems OK I'll remove this and test the component at actual scale.

ui/apps/platform/src/Containers/VulnMgmt/Entity/Policy/VulnMgmtPolicyOverview.js
```diff
@@ -55,6 +63,11 @@ const noop = () => {};
 const VulnMgmtPolicyOverview = ({ data, entityContext, setRefreshTrigger }) => {
     const workflowState = useContext(workflowStateContext);
     const { isFeatureFlagEnabled } = useFeatureFlags();
+    const [page, setPage] = useState(0);
```
dvail, line 66: I'm not sure if it is appropriate to put the pagination and sorting for the child table in `workflowState`, since that seemed to be used primarily for when the workflow sidebar is a full table.

LMK if this really should be handled in the `workflowState` though.

dvail:
@vjwilson I took a stab at this based on some assumptions about how this VM section is supposed to work. If the approach looks valid I'll go through and see what can be cleaned up and do a more thorough test. Let me know if I am misunderstanding the intent here though.

vjwilson (approved):
/lgtm

PR #4003, opened 2022-12-02
Dashrews/ROX-12930 postgres postgres upgrade

dashrews78:
## Description

Updates to the upgrade e2e tests to add in a postgres->postgres upgrade step.  Additionally when we first cut this we left the sensor upgrader tests running only in a rocksDB only test.  Those have been pulled over as well to run against a Postgres backed central.  Changes to the rocksdb run.sh were to simply move some scripts to lib.sh.

## Checklist

## Testing Performed

The whole thing is testing.


tests/upgrade/postgres_run.sh
```diff
@@ -7,6 +7,10 @@ set -euo pipefail
 
 TEST_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
 
+INITIAL_POSTGRES_TAG="3.73.x-75-gbab217c487"
```
c-du, line 10: Can you add a comment on why you would choose this image?

c-du (approved):

PR #4005, opened 2022-12-03
ROX-13676: Create external entities side panel layout

sachaudh:
## Description

This PR fleshes out the External Entities side bar a little bit more. I've added a flows table that is not editable and made a few little edits to make the data more in line with what you see in the current network graph

Screenshot 2022-12-02 at 4 23 53 PM

## Checklist


vjwilson (approved):
/lgtm

PR #4006, opened 2022-12-03
ROX-13371: Add GKE Log links for life of cluster

gavin-stackrox:
## Description

Adds links for GKE logs to prow results. [e.g.](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4006/pull-ci-stackrox-stackrox-master-gke-postgres-qa-e2e-tests/1599849433204264960)

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/scripts/ci/gke.sh
+++ b/scripts/ci/gke.sh
@@ -295,6 +295,7 @@ create_log_explorer_links() {
         <title><h4>GKE Logs Explorer</h4></title>
     </head>
     <body>
+    <p>(these links require a 'left-click -> open in tab')
     <ul style="padding-bottom: 28px; padding-left: 30px; font-family: Roboto,Helvetica,Arial,sans-serif;">
 HEAD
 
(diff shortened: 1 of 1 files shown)


scripts/ci/gke.sh
```diff
@@ -275,6 +277,53 @@ teardown_gke_cluster() {
...
+\<li\>\
+\<a href=\"https://console.cloud.google.com/logs/query\;query=\
+resource.type=%22k8s_container%22%0A\
+resource.labels.cluster_name%3D%22"$CLUSTER_NAME"%22%0A\
+resource.labels.namespace_name%3D%22stackrox%22%0A\
+\;timeRange="$start_ts"%2F"$end_ts"\
+\;cursorTimestamp="$start_ts"\
+?authuser="$authUser"\
+\&project="$project"\
+\&orgonly=true\&supportedpurview=organizationId\"\>authUser "$authUser"\</a\>\
```
porridge, line 319:
```suggestion
\&orgonly=true\&supportedpurview=organizationId\" target=\"_blank\" \>authUser "$authUser"\</a\>\
```

scripts/ci/gke.sh
```diff
@@ -275,6 +277,53 @@ teardown_gke_cluster() {
...
+    local start_ts
+    start_ts="$(cat /tmp/GKE_CLUSTER_CREATED_TIMESTAMP)"
+    local end_ts
+    end_ts="$(date -u +"%Y-%m-%dT%H:%M:%SZ")"
+    local project
+    project="$(gcloud config get project --quiet)"
+
+    for authUser in {0..2}; do
+        echo \
+\<li\>\
```
porridge, line 310: Why not use a here-doc in this case as well? To avoid having to put everything on a single line?

scripts/ci/gke.sh
```diff
@@ -275,6 +277,53 @@ teardown_gke_cluster() {
     gcloud container clusters delete "$CLUSTER_NAME" --async
 
     info "Cluster deleting asynchronously"
+
+    create_log_explorer_links
+}
+
+create_log_explorer_links() {
+    if [[ -z "${ARTIFACT_DIR:-}" ]]; then
+        info "No place for artifacts, skipping explorer links"
```
porridge, line 286:
```suggestion
        info "No place for artifacts, skipping generation of links to logs explorer"
```

porridge (approved):
Nice, but here is what I get when just clicking on the link:

Screenshot from 2022-12-06 09-35-34

looks like this does not work in a frame? How about using a target attribute?

scripts/ci/gke.sh
gavin-stackrox: Unfortunately `target=_blank` does not work in an iframe (at least in chrome). It opens a new tab with a 'content blocked' message. Maybe it needs a 'right click and open in new tab message/label' but that seems overly silly IMO.

PR #4014, opened 2022-12-05
Dashrews/ROX-13703 fix autocomplete test

dashrews78:
## Description

By default a deployment search appends a descending order by of the risk score.  So the primary order of a deployment is dictated by the risk score and not the field being searched.  This is how it works within the app.  With bleve it appears this sort was taking place after the fact whereas with Postgres it takes place with the query through the searcher.  So the test needs to utilize risk score to verify the results are returned in the order in which we expect them to be.

## Checklist

## Testing Performed

This is simply fixing a test.


md2119 (approved):
lgtm. 

The current changes are sufficient for now since the test is not enabled on postgres. Once you enable the test for postgres, you will have to update the Risk Score field on deployment as I pointed out on the JIRA.

dashrews78:
> lgtm.
> 
> The current changes are sufficient for now since the test is not enabled on postgres. Once you enable the test for postgres, you will have to update the Risk Score field on deployment as I pointed out on the JIRA.

@md2119 
I did set the Risk Score on the deployment BUT I had to do so through the deploymentRanker.  Since we are using the datastore and not the store itself the risk score on the deployment is calculated/pulled from the deploymentRanker.  See:
`	// Update deployment with latest risk score
	deployment.RiskScore = ds.deploymentRanker.GetScoreForID(deployment.GetId())`
in `upsertDeployment` of `central/deployment/datastore/datastore_impl.go`

It is also now on for Postgres with the Makefile change.  Does any of that change your opinion on this PR?

c-du (approved):

md2119:
🚢

PR #4019, opened 2022-12-05
Run builds in parallel on GHA

janisz:
## Description

Previous implementation of makefile for cli prevents multiple jobs.
Now we should get significant speedup by passing `-j` option.

PR #4018, opened 2022-12-05
Simplify make rules for CLI

janisz:
## Description

Currently in Makefile we have multiple copied lines to build CLI.
This PR fixes that by adding generic target `cli_%` that is able to build desired binaries by providing `cli_$OS-$ARCH` as target.

## Testing Performed

CI

PR #4020, opened 2022-12-05
ROX-12574: make grpc client to retry on errors

janisz:
## Description

Currently we are not retrying failed GRPC calls. This PR makes 3 retries on indepotent errors like
network issues (hard to unit test), `ResourceExhausted` and `Unavailable`.

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/99612e884e71d96a3c8dd30ad4d8eccacfa91798/retry/options.go#L15-L19

## Testing Performed

CI


janisz:
* master
  * **PR #4018** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4018" target="_blank">Graphite</a> 
    * **PR #4019** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4019" target="_blank">Graphite</a> 
      * **PR #4020** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4020" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4020?utm_source=stack-comment).

pkg/clientconn/client_test.go
```diff
@@ -138,3 +140,34 @@ func (t *ClientTestSuite) TestAuthenticatedHTTPTransport_WebSocket() {
 		})
 	}
 }
+
+func (t *ClientTestSuite) TestGRPCShouldRetry() {
+	testCases := map[string]int{
+		"retry on too large message": 3,
```
0x656b694d, line 146: nit: do we expect it to be not that large second time? Might be something like "try to turn it off and on again"?

0x656b694d (approved):
LGTM

pkg/clientconn/client.go
```diff
@@ -365,7 +366,17 @@ func AuthenticatedHTTPTransport(endpoint string, server mtls.Subject, baseTransp
...
-	allDialOpts := make([]grpc.DialOption, 0, len(dialOpts)+3)
+	allDialOpts := make([]grpc.DialOption, 0, len(dialOpts)+5)
+
+	retryOpts := []grpc_retry.CallOption{
+		grpc_retry.WithBackoff(grpc_retry.BackoffExponential(100 * time.Millisecond)),
+		grpc_retry.WithMax(3),
+	}
+
+	allDialOpts = append(allDialOpts,
+		grpc.WithStreamInterceptor(grpc_retry.StreamClientInterceptor(retryOpts...)),
```
mtodor, line 377: This change would affect the `scanner` and other components of our stack, and I'm a bit concerned if we would like a change like that in other services.

If we want retry logic for roxctl, then we should add it there. And I would suggest that we move these options to roxctl only. `roxctl/common/flags.go`

janisz: You are right. I hoped this change will not affect them that much but it's better to limit the scope.

janisz:
@mtodor PTAL

roxctl/common/tls.go
```diff
@@ -48,26 +48,26 @@ func (v *insecureVerifierWithWarning) VerifyPeerCertificate(leaf *x509.Certifica
...
 	if serverName == "" {
 		var err error
 		serverName, _, _, err = netutil.ParseEndpoint(endpoint)
 		if err != nil {
-			return "", "", errors.Wrap(err, "could not parse endpoint")
+			return "", "", false, errors.Wrap(err, "could not parse endpoint")
 		}
 	}
-	return endpoint, serverName, nil
+	return endpoint, serverName, usePlaintext, nil
```
0x656b694d, line 66: nit refactoring (due to two functions of the function, and therefore occasionally unused return values):
1. extract `endpoint, usePlaintext, err := flags.EndpointAndPlaintextSetting()`;
2. make `endpoint` the function parameter;
3. rename the function to `getServerName(endpoint string) (string, error)`.

janisz: I don't get it. You would like to extract function that will be only used in this single place?

roxctl/common/connection_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestGRPCShouldRetry(t *testing.T) {
+	ctx := context.Background()
+	called := 0
+	handlerFunc := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		called++
+		w.Header().Set("content-type", "application/grpc+proto")
+		if called == 1 {
+			fmt.Fprint(w, "too large message, please retry")
+		}
+	})
```
mtodor, line 25: For me, it's difficult to understand what produces errors in this test. And also, why the 2nd call is successful, but we still get an error from `conn.Invoke`. 😕 

 Can we change it to something easier to understand? i.e.
```suggestion
	handlerFunc := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("content-type", "application/grpc+proto")

		if called == 0 {
			w.WriteHeader(http.StatusInternalServerError)
		}
		called++
	})
```

And also, add a comment on why an error is returned from `conn.Invoke`. I assume it's because we use a GRPC client and talk with an HTTP server.

mtodor (approved):
Looks good. Added a few nitpicks, but nothing crucial.

roxctl/common/tls.go
0x656b694d: No, move it out of the function (to the caller) and pass the result as the argument.

roxctl/common/tls.go
```diff
@@ -48,26 +48,33 @@ func (v *insecureVerifierWithWarning) VerifyPeerCertificate(leaf *x509.Certifica
...
+}
+
+func getServerName(endpoint string) (string, error) {
+	if serverName := flags.ServerName(); serverName != "" {
+		return serverName, nil
+	}
+	serverName, _, _, err := netutil.ParseEndpoint(endpoint)
+	if err != nil {
+		return "", errors.Wrap(err, "could not parse endpoint")
```
0x656b694d, line 71: Well, this is not what I meant, I'm sorry for being unclear. Please revert the refactoring, as it doesn't make any sense now. I'll come up with a follow-up PR, if it will worth it.

PR #4022, opened 2022-12-05
ROX-13680: Add external to cluster side panel

sachaudh:
## Description

This PR adds the External to Cluster side panel

Screenshot 2022-12-05 at 12 46 24 PM

## Checklist


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -21,7 +21,8 @@ import defaultComponentFactory from './components/defaultComponentFactory';
 import DeploymentSideBar from './deployment/DeploymentSideBar';
 import NamespaceSideBar from './namespace/NamespaceSideBar';
 import CidrBlockSideBar from './cidr/CidrBlockSideBar';
-import ExternalEntitiesSideBar from './external/ExternalEntitiesSideBar';
+import ExternalEntitiesSideBar from './externalEntities/ExternalEntitiesSideBar';
+import ExternalToClusterSideBar from './external/ExternalToClusterSideBar';
```
vjwilson, line 25: These two names
```
ExternalEntitiesSideBar
ExternalToClusterSideBar
```

are a little too similar to parse easily.
What are some other names we could use?

ui/apps/platform/src/Containers/NetworkGraph/external/ExternalToClusterSideBar.tsx
```diff
@@ -0,0 +1,146 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function ExternalToClusterSideBar({ id, nodes, edges }: CidrBlockSideBarProps): ReactElement {
+    // component state
+    const [entityNameFilter, setEntityNameFilter] = React.useState<string>('');
+
+    // derived data
+    const externalToClusterNode = getNodeById(nodes, id) as ExternalNodeModel;
+    const externalNodes = nodes.filter(
+        (node) => node.data.type === 'CIDR_BLOCK' || node.data.type === 'EXTERNAL_ENTITIES'
+    ) as (ExternalEntitiesNodeModel | CIDRBlockNodeModel)[];
```
vjwilson, line 50: I wonder if there is a way we can get this list pre-filtered from the topology library, since it needs to already know this to render them in the correct group on the graph?

ui/apps/platform/src/Containers/NetworkGraph/external/ExternalToClusterSideBar.tsx
```diff
@@ -0,0 +1,146 @@
...
+                        <TextContent>
+                            <Text component={TextVariants.h1} className="pf-u-font-size-xl">
+                                {externalToClusterNode?.label}
+                            </Text>
+                        </TextContent>
+                        <TextContent>
+                            <Text
+                                component={TextVariants.h2}
+                                className="pf-u-font-size-sm pf-u-color-200"
+                            >
```
vjwilson, line 66: I think we already have an H1 on the page. Since this is a sub-section of the whole page, it would be more accessible to start this heading level with an H2, and so on. (The sizing seems custom regardless.)

```suggestion
                            <Text component={TextVariants.h2} className="pf-u-font-size-xl">
                                {externalToClusterNode?.label}
                            </Text>
                        </TextContent>
                        <TextContent>
                            <Text
                                component={TextVariants.h3}
                                className="pf-u-font-size-sm pf-u-color-200"
                            >
```

vjwilson (commented):
Some suggestions

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
sachaudh: Honestly, External Entities always seemed like an odd name for a single node because it's plural. If I had to think of another name, maybe we could change `ExternalToCluster` to `ExternalGroup`?

ui/apps/platform/src/Containers/NetworkGraph/external/ExternalToClusterSideBar.tsx
vjwilson: Sorry, I missed your reply.

I'm not sure what options are available from the controller; this was mainly me wishing out loud--for better documentation--I guess. Didn't mean for it to hold up this PR.

vjwilson (approved):

PR #4023, opened 2022-12-05
ROX-13730: Remove OpenShift 3 mode from Cluster form options

vjwilson:
## Description

A tester during the HackFest noticed that our Cluster form still has an OpenShift 3 option, but we no longer support that version of OpenShift.

## Checklist

## Testing Performed

Before this code change
Screen Shot 2022-12-05 at 10 47 37 AM

After this code change
Screen Shot 2022-12-05 at 11 14 08 AM

### DIFF
--- a/ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
+++ b/ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
@@ -35,7 +35,6 @@ export const runtimeOptions = [
 
 export const clusterTypes = {
     KUBERNETES: 'KUBERNETES_CLUSTER',
-    OPENSHIFT_3: 'OPENSHIFT_CLUSTER',
     OPENSHIFT_4: 'OPENSHIFT4_CLUSTER',
 };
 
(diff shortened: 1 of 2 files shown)


dvail (approved):
I don't have a lot of background here, but it looks like there are a handful of places in `ui/apps/platform/src/Containers/Clusters/StaticConfigurationSection.js` that can be deleted in a follow up since `OPENSHIFT_3` is no longer a possible option.

bradr5 (approved):

msugakov:
A potential problem here (thinking of recent KernelModules enum value removal) is that existing OpenShift 3 clusters (running under ACS 3.74) can **still** be connected to the ACS 4.y Central that does not any more have drop down value.

I think, this should work in a way to no longer allow selecting OpenShift 3 upon creation of the cluster, but should **still display** OpenShift 3 for the already existing cluster.

pedrottimark:
Van, it looks like this might be the last use of src/Components/Select.js so if we need to support disabled options, how about replace with PatternFly Select element and delete the classic component?

PR #4025, opened 2022-12-05
Rox 13506 add UI for extra fields in syslog notifier

vjwilson:
## Description

This PR adds the FE changes for the Extra Fields in the Syslog notifier integration.

It basically repeats the pattern for the same type of option in the Generic Webhook notifier integration. (If we do it a third time for yet another integration, we could consider if it's worth factoring out to a helper component, but that seems premature right now--especially with the extra handling required for an optional list of values.)

Note: this PR is pointed at the backend PR that adds the feature. If that one merges first, this one can be merged into the master branch.

## Checklist

## Testing Performed

New fields in the create/edit form
Screen Shot 2022-12-05 at 4 26 28 PM

Read-only view for the new fields
Screen Shot 2022-12-05 at 4 29 04 PM


sachaudh:
@vjwilson Looks like you might have to do a rebase

sachaudh (approved):
LGTM

PR #4028, opened 2022-12-06
ROX-13433: Adding Extraneous Flows graph

alwayshooin:
This will need to be rebased off of Saif's data structure PR, but it adds the extraneous flows graph logic, refactors some of the node/grouped node generating utils, and adds a new `fakeGroup` type to accommodate for the allowed in/egress flows node that appears for the extraneous flows graph when a node is selected. 

The `ingress flows` and `egress flows` will be added in a separate PR to incrementally add changes

image

active traffic w central selected: 
Screenshot 2022-12-05 at 4 10 51 PM

extraneous flows w central selected:
Screenshot 2022-12-05 at 4 11 30 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -17,7 +17,7 @@ import useDeepCompareEffect from 'use-deep-compare-effect';
 
 import useFetchClusters from 'hooks/useFetchClusters';
 import useURLSearch from 'hooks/useURLSearch';
-import { fetchNetworkFlowGraph } from 'services/NetworkService';
+import { fetchNetworkFlowGraph, fetchNetworkPolicyGraph } from 'services/NetworkService';
 import { getQueryString } from 'utils/queryStringUtils';
 import timeWindowToDate from 'utils/timeWindows';
 
@@ -42,6 +42,8 @@ const includePorts = true;
 
 function NetworkGraphPage() {
     const [edgeState, setEdgeState] = useState<EdgeState>('active');
+    const [activeModel, setActiveModel] = useState<CustomModel>(emptyModel);
+    const [extraneousFlowsModel, setExtraneousFlowsModel] = useState<CustomModel>(emptyModel);
     const [model, setModel] = useState<CustomModel>(emptyModel);
     const [isLoading, setIsLoading] = useState(false);
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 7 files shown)


sachaudh:
@alwayshooin you can rebase now

sachaudh (commented):
LGTM to me so far. I'll wait for approval until the rebase occurs and I take another look

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -85,3 +145,109 @@ export function transformData(nodes: Node[]): Model {
...
+            showContextMenu: false,
+        },
+    };
+    const extraneousIngressNode: NodeModel = {
+        id: 'extraneous-ingress',
+        type: 'fakeGroup',
+        width: 75,
+        height: 75,
+        label: 'Ingress flows',
+        // TODO: figure out how to fake group node
```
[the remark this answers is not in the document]
alwayshooin: yeah, i need to figure out how to represent it visually in the graph and add the edges. the ones here are mostly just placeholders for now

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -10,7 +10,9 @@ export type CustomNodeModel =
     | DeploymentNodeModel
     | ExternalNodeModel
     | ExternalEntitiesNodeModel
-    | CIDRBlockNodeModel;
+    | CIDRBlockNodeModel
+    | PolicyNodeModel
```
sachaudh, line 14: What is the `PolicyNodeModel` and how is it different from a `DeploymentNodeModel`? I was wondering more why have a separate type. If a `PolicyNodeModel` has `nonIsolatedIngress` and `nonIsolatedEgress` as false, then would it be the same as a `DeploymentNodeModel`?

sachaudh (approved):
LGTM. Just a comment

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -42,6 +48,8 @@ const includePorts = true;
 
 function NetworkGraphPage() {
     const [edgeState, setEdgeState] = useState<EdgeState>('active');
+    const [activeModel, setActiveModel] = useState<CustomModel>(emptyModel);
+    const [extraneousFlowsModel, setExtraneousFlowsModel] = useState<CustomModel>(emptyModel);
```
vjwilson, line 52: Ok for this PR, but I think we need a comment and a Jira task to update this state to be URL-driven, based on the plan in our design doc.

> For controls that cause the front-end to display the data differently, without having to make another API call, we will have to create appropriate param names. Because our Search options are always Title-Case strings with spaces between distinct words, to avoid collisions we should use appropriate camelCase query param names for display-only features, like showConnections=allowed and namespaceFlows=hide
https://docs.google.com/document/d/1QMI0PQ_duKJ_6aqD3SU8gNjBTUDPyNBMOktEG5opABc/edit#heading=h.pan7jow300u7

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -67,16 +75,68 @@ function NetworkGraphPage() {
...
+                        includePorts
+                    ),
+                ])
+                    .then((values) => {
+                        const activeNodeMap: Record<string, CustomNodeModel> = {};
+                        const activeEdgeMap: Record<string, EdgeModel> = {};
+                        const policyNodeMap: Record<string, PolicyNodeModel> = {};
+
+                        // get policy nodes from api response
+                        const { nodes: policyNodes } = values[1].response;
```
vjwilson, line 100: Nit: replace hard-coded position "magic number" with a constant.

```
const POLICY_RESPONSE_INDEX = 1;
...
    const { nodes: policyNodes } = values[POLICY_RESPONSE_INDEX].response;
```

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -67,16 +75,68 @@ function NetworkGraphPage() {
...
+                        // to enhance active node data
+                        policyDataModel.nodes?.forEach((node) => {
+                            // no grouped nodes in policy graph data model
+                            if (!policyNodeMap[node.id]) {
+                                policyNodeMap[node.id] = node as PolicyNodeModel;
+                            }
+                        });
+
+                        // get active nodes from api response
+                        const { nodes: activeNodes } = values[0].response;
```
vjwilson, line 113: Nit: magic number

ui/apps/platform/src/Containers/NetworkGraph/components/stylesComponentFactory.tsx
```diff
@@ -33,6 +33,12 @@ const stylesComponentFactory: ComponentFactory = (kind: ModelKind, type: string)
                 // @ts-ignore
                 withSelection()(StyleGroup)
             );
+        case 'fakeGroup':
+            return withDragNode(nodeDragSourceSpec('node', true, true))(
```
vjwilson, line 37: It looks like we could group this case with `case 'node':` above.

vjwilson (commented):
I know this is already merged, but I left a few comments for future considertation.

I noticed a couple of odd behaviors in the `EdgeStateSelect` component when smoke-testing this on a cluster.

1. The dropdown doesn't close on selecting an option.
2. Clicking either option toggles between the two options, which is counter-intuitive. I would expect that clicking an option would be "idempotent".

PR #4031, opened 2022-12-06
ROX-10895: Detect role bindings to cluster roles within namespace RBAC evaluator

dhaus67:
## Description

The RBAC evaluator gives the ability to provide an overview of permissions within either a specific cluster or namespace for e.g. service accounts, users, groups.

This is done by inspecting the role bindings within the specific namespace / cluster, resolving the bound role and subject, and the given permissions for the specific subjects.

As a high-level overview of RBAC within Kubernetes, [refer to this documentation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding).

Now, role bindings, which are scoped to a specific namespace, have the ability to bind to cluster roles. The role binding won't grant permission on the whole cluster, but apply the permissions defined within the cluster role to subjects within the specific namespace.

Previously, for namespaced RBAC evaluation (i.e. when a user doesn't have an access scope with full cluster access), we skipped role bindings which bind to cluster role bindings.

This has some unwanted drawbacks:
Take the following role binding as an example:
```yaml
kind: RoleBinding
metadata:
  name: test-rolebinding
  namespace: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: dh-test
  namespace: default 
```
The role binding will grant the `dh-test` service account within the namespace cluster-admin permission, as defined within the cluster role:
```yaml
kind: ClusterRole
metadata:
  name: cluster-admin
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '*'
  verbs:
  - '*'
```

However, when viewing that specific service account within e.g. the UI / API **with a scope that doesn't have full access to the cluster, e.g. only the default namespace**, he will get the impression that the service account has _no_ permissions available:
service_account_rolebinding_clusterrole

This gives a wrong depiction of the actual RBAC permission state within the specific namespace.

Instead of confusing people with limited access scopes for this particular issue, we will allow them to view permissions of cluster roles, in case they have role bindings which are binding them, irrespective of whether they have the full cluster in their access scope.

This PR will do so, on a conditional basis, and only elevate the context for retrieving _cluster roles_. The overall behavior will stay the same for non-cluster roles.

## Checklist

## Testing Performed

```
1. Create an access scope that is scoped to a single namespace, e.g. `rox-10895`.

2. Create a serviceaccount + rolebinding within the namespace the access scope is scoped to:
apiVersion: v1
kind: ServiceAccount
metadata:
  creationTimestamp: null
  name: dh-test
  namespace: rox-10895
---
kind: RoleBinding
metadata:
  name: rolebinding
  namespace: rox-10895
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: dh-test
  namespace: rox-10895

3. As admin user / user with full cluster scope, check the service account within the UI and observe that the service account has permission to change _everything_ within the namespace (Config Management -> RBAC -> Service Accounts).

4. Switch to a user that has a role which has the previously created access scope.
5. You should observe the same permissions on the service account albeit the specific user doesn't have full access to the cluster (Config Management -> RBAC -> Service Accounts).
```


dhaus67:
* master
  * **PR #4031** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4031" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4031?utm_source=stack-comment).

janisz (commented):
Does it make sense to split this into two?

dhaus67:
@janisz split this one out!

central/risk/multipliers/deployment/service_account_permissions_test.go
```diff
@@ -214,7 +215,7 @@ func TestPermissionScore(t *testing.T) {
 
 	for _, c := range clusterCases {
 		t.Run(c.name, func(t *testing.T) {
-			ctx := context.Background()
+			ctx := sac.WithGlobalAccessScopeChecker(context.Background(), sac.AllowAllAccessScopeChecker())
```
rhybrillou, line 218: nit: `sac.WithAllAccess(context.Background())` does the same, and might be easier to understand for an external reader (also applies to other occurrences in the PR).

rhybrillou (approved):

central/risk/multipliers/deployment/service_account_permissions_test.go
rukletsov: ➕

PR #4036, opened 2022-12-06
ROX-13023: Use database count rather than result count in postgres scoped access control context

rhybrillou:
## Description

So far, queries to the datastore to count items when the user has a limited access scope do perform a scoped search, and the response is the number of items returned by the search.
This has the drawback that scoped count queries generate unnecessary traffic between the datastore and the service layer.
The goal of this change is in postgres mode to let the database return only the number of items matching the requester scope rather than all items matching the requester scope and letting the application code count the returned items.

## Checklist

## Testing Performed

Scoped access control tests do cover the Count function in most datastores. A manual run of datastore tests for sample central subdirectories was done, and triggered the new code path.

CI should be sufficient.

### DIFF
--- a/pkg/sac/search_helper.go
+++ b/pkg/sac/search_helper.go
@@ -273,8 +273,17 @@ func (h *pgSearchHelper) executeCount(ctx context.Context, q *v1.Query, searcher
 		return searcher.Count(ctx, q)
 	}
 
-	results, err := h.executeSearch(ctx, q, searcher)
-	return len(results), err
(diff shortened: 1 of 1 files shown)


janisz (approved):

rukletsov (approved):
@rhybrillou could you please confirm in the testing done section that this PR fixes the problem described in https://issues.redhat.com/browse/ROX-13786? And if so, also add ROX-13786 to the title of the PR.

PR #4037, opened 2022-12-06
Jv rox 13774 add cluster id to processes listening on ports api

JoukoVirtanen:
## Description

Adds cluster_id to processes listening on ports API. This is needed to help customers find the process and port reported by the API.

## Checklist
- [x] Updated e2e tests

## Testing Performed

CI is sufficient.


proto/storage/process_listening_on_port.proto
```diff
@@ -16,6 +16,7 @@ message ProcessListeningOnPort {
     L4Protocol protocol = 2;
     ProcessIndicatorUniqueKey process = 3;
     google.protobuf.Timestamp close_timestamp = 4;
+    string cluster_id = 6;
```
erthalion, line 19: What do you think about doing the same as with `ProcessIndicatorUniqueKey`, i.e. create a `ProcessInfo` structure with the `cluster_id` in it? This will make things more structured, especially since we have prospects of enriching this with more information.

JoukoVirtanen: We are trying to match the process API more closely. That API does not have cluster_id as part of the process signal. Therefore, cluster_id should be separate from the process information. This is the structure of the process API response 

`{
  "processes": [
    {
      "id": "string",
      "deploymentId": "string",
      "containerName": "string",
      "podId": "string",
      "podUid": "string",
      "signal": {
        "id": "string",
        "containerId": "string",
        "time": "2019-08-24T14:15:22Z",
        "name": "string",
        "args": "string",
        "execFilePath": "string",
        "pid": 0,
        "uid": 0,
        "gid": 0,
        "lineage": [
          "string"
        ],
        "scraped": true,
        "lineageInfo": [
          {
            "parentUid": 0,
            "parentExecFilePath": "string"
          }
        ]
      },
      "clusterId": "string",
      "namespace": "string",
      "containerStartTime": "2019-08-24T14:15:22Z",
      "imageId": "string"
    }
  ]
}`

erthalion: > That API does not have cluster_id as part of the process signal. Therefore, cluster_id should be separate from the process information.

This doesn't sound logical to me at all, neither the idea of mixing listening and process information together.

JoukoVirtanen: cc @robbycochran @md2119 @theencee Can you help resolve this disagreement? I will also post an example request to the API and response from the API on slack. I will also create a document with my plan for the API here https://docs.google.com/document/d/1kKDgZXQfTI1hUCQpatC1RgkoNx8viIfChI8bZsBTAN8/edit

PR #4030, opened 2022-12-06
ROX-13593: Add telemetry options to central Helm chart

stehessel:
## Description

The current telemetry implementation has been deprecated, but not removed from the code base as a whole. It will be replaced by a new telemetry implementation (see #3819). We are planning to use an installation toggle to give opt-in/opt-out of the new telemetry collection. If enabled, the installation sets up a telemetry storage key, which enables the collection within Central.

The current `disableTelemetry` flag is removed. The current telemetry data store and api are deprecated and will be removed later. The installation toggle will be the only way to enable/disable telemetry - even if that then requires rolling the central pod.

This PR sets up the following configuration for the new telemetry stack:
- A user facing boolean setting to enable/disable telemetry (`telemetry.enabled`).
- Two options to configure the telemetry storage backend (`telemetry.storage`). Initially, at least the storage key has to be provided manually to enable telemetry. Eventually, the storage will be hardcoded with a default value in the chart.

An addition to the central CRD for the operator installation method will follow.

Note that we are not planning to roll out telemetry to on-premise users at this time. Telemetry will initially be a feature only enabled for the ACS cloud service. However, we want to lay the config ground work for future a on-premise rollout.

We opted to not document the flags publicly to avoid confusion of on-prem users, for whom the telemetry options are currently ineffective. The current `disableTelemetry` flag is still part of upstream documentation (https://docs.openshift.com/acs/3.72/installing/installing_helm/install-helm-customization.html#central-services-public-configuration-file-central_acs-install-helm-customization) - this entry will be removed from the documentation.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- [x] Evaluated and added CHANGELOG entry if required
- [ ] ~Determined and documented upgrade steps~
- [x] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

PR #4041, opened 2022-12-06
ROX 13593: Add telemetry to central CRD

stehessel:
## Description

The primary use case of this change is setting up central telemetry via fleetshard-sync for the ACS cloud service. Down the line, users will be able to opt-in/opt-out of telemetry during installation via the `enabled` toggle.

We opted to not document the flags publicly to avoid confusion of on-prem users, for whom the telemetry options are currently ineffective.

Note that the diff is larger than expected because `make generate` changed some formatting options.

## Checklist

## Testing

Tested Central install with custom ACS operator.


stehessel:
* master
  * **PR #4030** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4030" target="_blank">Graphite</a> 
    * **PR #4041** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4041" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4041?utm_source=stack-comment).

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -69,7 +69,7 @@ type Egress struct {
 }
 
 // ConnectivityPolicy is a type for values of spec.egress.connectivityPolicy.
-//+kubebuilder:validation:Enum=Online;Offline
+// +kubebuilder:validation:Enum=Online;Offline
```
dhaus67, line 72: more of a super nit: Sometimes its `// +kubebuilder...` or `//+kubebuilder`.  Would be great to be consistent now that we change it here / or just _not_ change this.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -120,6 +120,11 @@ type CentralComponentSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=6,displayName="Central DB Settings (Technology Preview)"
 	DB *CentralDBSpec `json:"db,omitempty"`
 
+	// Configures telemetry settings for Central. If enabled, Central transmits telemetry and diagnostic
+	// data to a remote storage backend.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=7,displayName="Central telemetry settings"
```
dhaus67, line 125: So, as far as I understand this:
This [API marker](https://sdk.operatorframework.io/docs/building-operators/golang/references/markers/) will render the specific option within the UI for customers to discover during installation of the CR within e.g. the OpenShift console.
Since we opted out to _officially_ document it, personally I'd say it would be best to just not show it within the UI during operator installation for now.

Might be a good question for someone from Draco: In case we have knobs in the CR we do not _yet_ want to expose, is it possible to skip this API marker?

Alternatively, we could do something vaguely similar to the central DB stuff and declare it as "Experimental" or something like that (albeit, no customer will be actually able to use it yet, so I still prefer not showing it at all).

operator/pkg/central/values/translation/translation.go
```diff
@@ -231,6 +233,24 @@ func getCentralDBComponentValues(c *platform.CentralDBSpec) *translation.ValuesB
...
 }
 
+func getTelemetryValues(t *platform.Telemetry) *translation.ValuesBuilder {
+	cv := translation.NewValuesBuilder()
+	if t == nil {
+		return &cv
+	}
+
+	if t.Enabled != nil {
+		cv.SetBoolValue("enabled", *t.Enabled)
```
dhaus67, line 243:
```suggestion
		cv.SetBool("enabled", t.Enabled)
```

operator/pkg/central/values/translation/translation.go
```diff
@@ -231,6 +233,24 @@ func getCentralDBComponentValues(c *platform.CentralDBSpec) *translation.ValuesB
 	return &cv
 }
 
+func getTelemetryValues(t *platform.Telemetry) *translation.ValuesBuilder {
+	cv := translation.NewValuesBuilder()
```
dhaus67, line 237: Looking at the file, this seems more in line with the naming (`telemetry values -> tv`):
```suggestion
	tv := translation.NewValuesBuilder()
```

operator/tests/central/basic/80-assert.yaml
```diff
@@ -1,10 +1,16 @@
-apiVersion: v1
```
dhaus67, line 1: Any reason for these changes / renames? Why not use the `90` prefix for the telemetry tests?

dhaus67 (changes requested):
You still need to run `make -C operator manifests bundle` to apply your changes to the `config/ & bundle/` generated files.

Not directly related:
The changes you mentioned related to the `make generate` command: what exactly has changed? Running this locally for me doesn't result in the same changes you had, so curious what changed (and what I'm probably missing).

operator/apis/platform/v1alpha1/central_types.go
stehessel: I didn't change these by hand. I don't why know, but the formatter is behaving very strangely...

operator/tests/central/basic/80-assert.yaml
stehessel: The previous `80` now `90` prefix is a deletion step and seemed to me like it was last intentionally.

stehessel:
Ok so I now executed `make generate` again and it removed a lot of the formatting diff again. Not sure how this happened in the first place 🤔

operator/apis/platform/v1alpha1/central_types.go
stehessel: Agree that we should not show this during installation if possible. Asked the Draco team about this topic.

stehessel: I added a `hidden` marker to the Telemetry options. This hides the option in the UI.

operator/apis/platform/v1alpha1/central_types.go
stehessel: So I finally figured it out. I get a different formatting result depending on whether I use go1.18 (`//+kubebuilder`) vs go1.19 (`// +kubebuilder`).

stehessel:
I dropped the default value as well to allow us a path to opt-out at some point if we want (otherwise all CRDs would have `telemetry.enabled=false` opting out by default).

dhaus67 (approved):
🚢

PR #4042, opened 2022-12-07
ROX-9115: E2e test runner enhancements

gavin-stackrox:
## Description

- Makes configuration reusable by saving items in the working directory which allows the test to run independently.
- Adds long options for this and future extension.
- Adds a --spin-cycle option to repeatedly run tests to weed out flakes.

Note: I've focused on making this work for `qa-tests-backend/`. Similar changes could be applied to the UI/NonGroovy/Upgrade tests but that remains TBD until someone is driven to undertake it.

## Checklist

## Testing Performed

CI is sufficient


qa-tests-backend/scripts/lib.sh
```diff
@@ -2,22 +2,15 @@
 
 # Common functions for deploying a cluster for QA tests
 
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
+
 set -euo pipefail
```
porridge, line 7: How about swapping these lines?

porridge (approved):

qa-tests-backend/scripts/lib.sh
gavin-stackrox: > How about swapping these lines?

There is a weird error when the set is before the ROOT determination that I was not able to resolve. This is why this is ordered this way in all the CI YAML -> bash scripts.

porridge: > There is a weird error when the set is before the ROOT determination that I was not able to resolve

WAT?

```
$ cat a.sh 
#!/bin/bash
set -euo pipefail
ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
echo a
$ ./a.sh 
a
$ echo $?
0
$ 
```

You got me thinking. Can you show an example?

gavin-stackrox: Try changing the order in `scripts/ci/lib.sh` and you should see it. I'll see if I can repro/dig up my notes to justify the order.

PR #4043, opened 2022-12-07
ROX-13782: Fix Compliance crash with operator standards

vjwilson:
## Description

The Compliance Operator standards,
* ocp4-cis
* opc4-cis-node
* opc4-high

were causing a crash when their GraphQL queries were parsed on the Compliance section dashboard.

This change sanitizes those query names, and replaces all non-word characters with underscores, `_`.

## Checklist

## Testing Performed

Manual testing

Here is the non-crashing dashboard, which includes those operator standards.
Screen Shot 2022-12-06 at 8 53 57 PM

### DIFF
--- a/ui/apps/platform/src/queries/standard.js
+++ b/ui/apps/platform/src/queries/standard.js
@@ -93,7 +93,10 @@ export const LIST_STANDARD_NO_NODES = gql`
 `;
 
 export const COMPLIANCE_STANDARDS = (standardId) => gql`
-    query complianceStandards_${standardId}($groupBy: [ComplianceAggregation_Scope!], $where: String) {
+    query complianceStandards_${standardId.replace(
+        /\W/g,
+        '_'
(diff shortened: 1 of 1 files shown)


ui/apps/platform/src/queries/standard.js
```diff
@@ -93,7 +93,10 @@ export const LIST_STANDARD_NO_NODES = gql`
 `;
 
 export const COMPLIANCE_STANDARDS = (standardId) => gql`
-    query complianceStandards_${standardId}($groupBy: [ComplianceAggregation_Scope!], $where: String) {
+    query complianceStandards_${standardId.replace(
+        /\W/g,
```
pedrottimark, line 97: This is okay for a patch.

The permanent change we will make on master, which removes the need for any additional testing:
* Go back in product code to `query complianceStandards(` without suffix.
* Go back in compliance test code to unique method to wait for multiple synonymous GraphQL requests.

PR #4046, opened 2022-12-07
Start post-Postgres migrations

c-du:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dashrews78 (approved):
LGTM, but I'm not sure it will make it clearer for anyone else.  Though I do like the idea of the "Last" Rocks migration.

keyallis (commented):
It might be good to add a little README or something just to give a more direct guidance to anyone stumbling across this of where new migrations should be implemented and any other changes they would need to make ie bumping the version number

keyallis (changes requested):
This will need to be rebased for changes made in #4068

dashrews78 (commented):
We may need to think about this one some more.  We have a situation where #4068 used a change in the patch release which hoses up migrations.  I have some ideas on how to work around it, but I think we need to know the Postgres version explicitly to do that.  Without storing it in the DB, I think we may be better served keeping PostgresDBVersionPlus, at least for now.  But I'm open to ideas.

migrator/migrations/registry.go
```diff
@@ -14,7 +14,7 @@ var (
 
 // MustRegisterMigration registers a Migration, panic-ing if there's an error.
 func MustRegisterMigration(m types.Migration) {
-	if !env.PostgresDatastoreEnabled.BooleanSetting() && m.StartingSeqNum > migrations.CurrentDBVersionSeqNumWithoutPostgres() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() && m.StartingSeqNum > migrations.BasePostgresDBVersionSeqNum() {
```
rhybrillou, line 17: Does this mean non-postgres migrations beyond `LastRocksDBVersionSeqNum` will be silently ignored ?
If having extra non-postgres migrations is an error, wouldn't it be better to panic here, as suggested in the function documentation comment ?

pkg/migrations/internal/seq_num.go
```diff
@@ -4,7 +4,8 @@ var (
 	// CurrentDBVersionSeqNum is the current DB version number.
 	// This must be incremented every time we write a migration.
 	// It is a shared constant between central and the migrator binary.
-	CurrentDBVersionSeqNum = 112
-	// PostgresDBVersionPlus is the current DB version number with Postgres DB data migration.
-	PostgresDBVersionPlus = 56
+	CurrentDBVersionSeqNum = 168
```
rhybrillou, line 7: Thank you for this semantic change. I think it makes things easier not to have to compute what the `PostgresDBVersionPlus` should be.

pkg/migrations/seq_num.go
```diff
@@ -10,14 +10,14 @@ import (
...
 }
 
-// CurrentDBVersionSeqNumWithoutPostgres is the base of current DB version number
-// without Postgres migrations. This function should only be used in Postgres
+// BasePostgresDBVersionSeqNum is the base of DB version number
+// for Postgres migrations. This function should only be used in Postgres
 // migrations.
-func CurrentDBVersionSeqNumWithoutPostgres() int {
-	return internal.CurrentDBVersionSeqNum - 1
+func BasePostgresDBVersionSeqNum() int {
```
rhybrillou, line 21: What is the lifetime of this function ? Should it be restrained to the initial Postgres DB load, or should it be used in newer migrations too, or could it be dropped so `CurrentDBVersionSeqNum` can be used directly in all n_ migrations ?

rhybrillou (commented):
Thank you, I think this goes in the direction of making migration-writing easier.

pkg/migrations/seq_num.go
c-du: This should not be used in new Postgres to Postgres migrations. In your PR, I can make change accordingly after your PR merged.

keyallis (approved):
Unblocking

PR #4054, opened 2022-12-07
Refactor front end types for collection responses

dvail:
## Description

This cleans up some of the type naming throughout Collections to be more understandable. The core changes are:

`Collection` => `ClientCollection`
This is the limited representation that we only use client side for Collection UI logic, so that is called out more explicitly.

`ResolvedCollectionResponse` => `CollectionResponse`
The `Resolved` piece here really didn't have much meaning, and was being used to distinguish it from the inner collection configuration returned by the server.

`CollectionResponse` => `Collection`
The former `CollectionResponse` is really the canonical representation of a Collection as understood by the server, so it is renamed to a plain `Collection`.

## Checklist

## Testing Performed

Automated tests.

The only non-type-level change is that the response for the create and update service calls is now automatically unwrapped. `response.data` => `response.data.collection`. This return value is not currently used anywhere in the app.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionAttacher.tsx
@@ -4,16 +4,16 @@ import { Alert, Button, debounce, Flex, SearchInput } from '@patternfly/react-co
 import BacklogListSelector, {
     BacklogListSelectorProps,
 } from 'Components/PatternFly/BacklogListSelector';
-import { CollectionResponse } from 'services/CollectionsService';
+import { Collection } from 'services/CollectionsService';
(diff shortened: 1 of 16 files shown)


dvail:
* master
  * **PR #4054** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4054" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4054?utm_source=stack-comment).

pedrottimark (approved):
This helps especially to contrast:
* `ClientCollection` for form values which can change during create/clone/edit action.
* `Collection` for attached/available collections which do not change.

PR #4055, opened 2022-12-07
Debounce dry run request function at the point of request

dvail:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
@@ -16,9 +16,21 @@ import { ListIcon } from '@patternfly/react-icons';
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import ResourceIcon from 'Components/PatternFly/ResourceIcon';
 
-import { dryRunCollection, CollectionRequest } from 'services/CollectionsService';
+import { CollectionRequest, dryRunCollection } from 'services/CollectionsService';
 import { ListDeployment } from 'types/deployment.proto';
 import { SelectorEntityType } from './types';
+
+function fetchMatchingDeployments(
+    dryRunConfig: CollectionRequest,
+    page: number,
+    filterText: string,
+    entity: SelectorEntityType
+) {
+    const pageSize = 10;
+    const query = { [entity]: filterText };
+    const { request } = dryRunCollection(dryRunConfig, query, page, pageSize);
+    return request;
+}
 
 export type CollectionResultsProps = {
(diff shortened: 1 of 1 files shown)


dvail:
* master
  * **PR #4054** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4054" target="_blank">Graphite</a> 
    * **PR #4055** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4055" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4055?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -138,7 +138,9 @@ function CollectionResults({ dryRunConfig }: CollectionResultsProps) {
                                 variant="link"
                                 isInline
                                 className="pf-u-text-align-center"
-                                onClick={() => fetchDryRun(currentPage + 1)}
+                                onClick={() =>
+                                    fetchDryRun(dryRunConfig, currentPage + 1, filterText, selected)
```
pedrottimark, line 142: Off the topic of this pull request, seeing `currentPage + 1` reminds me of the following:

https://reactjs.org/docs/faq-state.html#what-is-the-difference-between-passing-an-object-or-a-function-in-setstate

* object `this.setState({ count: this.state.count + 1})`
* function
  ```js
  this.setState((state) => {
    // Important: read `state` instead of `this.state` when updating.
    return {count: state.count + 1}
  });
  ```

And also discussion how to render reloading state. Independent of other details about visual presentation, what is pro and con of a `disabled={isFetchingDryRun}` prop to make explicit that `currentPage` is a stable value?

Your viewpoint of the scenario is welcome if I am misreading or misunderstanding!

pedrottimark (approved):
So good to see that previous decisions enable instead of impede adjustments like this.

One question to improve my understanding.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
dvail: I think that is definitely a good idea. In addition to the point about being a stable value, I think we would get into weird behavior if the user clicks the "view more" link multiple times before the request returns. I'll make a note to handle this as a follow up in the reloading state change, as well as hit the other places we are doing "view more" in the collections workflow.

bradr5 (approved):

PR #4062, opened 2022-12-08
ROX-13814: s/k8s.gcr.io/registry.k8s.io/g

porridge:
## Description

Let us see how much breaks.

## Checklist

## Testing Performed

Relying on CI.


janisz (approved):

pkg/fixtures/image.go
```diff
@@ -3328,19 +3328,19 @@ var (
...
-		{"quay.io/openshift/community-e2e-images:e2e-36-k8s-gcr-io-sd-dummy-exporter-v0-2-0-pPp4DDTpVamU2YNiNfTgVQ", "sha256:6057499be1ceca52655553ff2d9361d88580da2edd5fb1c5d4cff1156ae0fdaa"},
-		{"quay.io/openshift/community-e2e-images:e2e-30-k8s-gcr-io-prometheus-dummy-exporter-v0-1-0-fNZIUrKnAwcvKTvFkmEBKw", "sha256:dfd3c8dbb1caec41ca21fe6c77138cbf730b2481987793aaefccbf8b2e80f803"},
+		{"quay.io/openshift/community-e2e-images:e2e-31-registry.k8s.io-prometheus-to-sd-v0-5-0-6JI59Yih4oaj3oQOjRfhyQ", "sha256:14666989f40bb7c896c3e775a93c6873e2b791d65bc65579f58a078b7f9a764e"},
+		{"quay.io/openshift/community-e2e-images:e2e-36-registry.k8s.io-sd-dummy-exporter-v0-2-0-pPp4DDTpVamU2YNiNfTgVQ", "sha256:6057499be1ceca52655553ff2d9361d88580da2edd5fb1c5d4cff1156ae0fdaa"},
+		{"quay.io/openshift/community-e2e-images:e2e-30-registry.k8s.io-prometheus-dummy-exporter-v0-1-0-fNZIUrKnAwcvKTvFkmEBKw", "sha256:dfd3c8dbb1caec41ca21fe6c77138cbf730b2481987793aaefccbf8b2e80f803"},
 		{"quay.io/openshift/community-e2e-images:e2e-33-gcr-io-kubernetes-e2e-test-images-regression-issue-74839-amd64-1-0-zKI5eDEBSTF6x1_LICVZ9w", "sha256:3b36bd80b97c532a774e7f6246797b8575d97037982f353476c703ba6686c75c"},
```
janisz, line 3334: This should not be changed

porridge: Oops, should have used `[.]` rather than `.` in the regex.
But WAT? Why didn't this cause any failures? :thinking:

central/imageintegration/store/defaults.go
```diff
@@ -28,7 +28,7 @@ var DefaultImageIntegrations = []*storage.ImageIntegration{
 		Categories: []storage.ImageIntegrationCategory{storage.ImageIntegrationCategory_REGISTRY},
 		IntegrationConfig: &storage.ImageIntegration_Docker{
 			Docker: &storage.DockerConfig{
-				Endpoint: "k8s.gcr.io",
```
porridge, line 31: Is this safe to change in situ? Or should we introduce a new entry, and leave the current one alone? :thinking:

porridge:
cc @stackrox/ui for the UI bits

vjwilson (approved):
UI files are okay.

qa-tests-backend/src/test/groovy/ImageScanningTest.groovy
```diff
@@ -654,7 +654,7 @@ class ImageScanningTest extends BaseSpecification {
 
         where:
         image                                              | registry
-        "k8s.gcr.io/ip-masq-agent-amd64:v2.4.1"            | "gcr registry"
+        "registry.k8s.io/ip-masq-agent-amd64:v2.4.1"            | "gcr registry"
```
gavin-stackrox, line 657: nit: ```suggestion
        "registry.k8s.io/ip-masq-agent-amd64:v2.4.1"       | "gcr registry"
```

pkg/fixtures/image.go
RTann: This is used for scale testing, so it's possible those images were not used or the errors were ignored. Let's switch them back

central/imageintegration/store/defaults.go
RTann: I think this is ok, though I don't know if this would require a migration as well. Maybe @c-du or @md2119 knows?

theencee: What do you want to do about all customers that already have an integration with k8s.gcr.io? Keep or migrate?

RTann: It's unclear when `k8s.gcr.io` will be obsolete at the moment (yes, it's deprecated, but are images still pushed there for now? Also, it's still the [actual blob storage backend for `registry.k8s.io`](https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/archeio/docs/request-handling.md#request-handling)). Perhaps no need to migrate.

However, I also started thinking if we need to consider web proxies. It's possible customers may need to update proxies to support traffic to `registry.k8s.io`, so perhaps we'd need to warn people before doing this? Perhaps @porridge brings up a good idea and we should make a separate image integration for `registry.k8s.io` and then eventually delete the `k8s.gcr.io` integration. No matter what, I'd also suggest we update our docs or the CHANGELOG so people know this is an endpoint to which they may want to consider allowing traffic from Central/Scanner

porridge: >Also, it's still the [actual blob storage backend](https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/archeio/docs/request-handling.md#request-handling)
It is for now, **unless** the client is running on AWS.

> Perhaps no need to migrate.

There can be various grades of "need", but I think the following are good reasons to migrate:
- better performance for AWS workloads
- responsible citizenship: the k8s project makes it quite clear that they are requesting everyone to migrate ASAP to reduce costs

But :100: to the other points you're making!

pkg/fixtures/image.go
porridge: Done.

porridge:
I changed the approach to add a new integration struct, and just mark the old one as deprecated.
Also added a couple of changelog entries, please let me know if they make sense.
Addressed other comments too. PTAL

porridge:
Also, does anyone think we need upgrade steps or docs updates? :thinking:

porridge:
Friendly ping to PTAL since the approach changed somewhat, @md2119 @gavin-stackrox @RTann @theencee @janisz 
I'd also appreciate comments w.r.t. the changelog/docs approach.

janisz (approved):

gavin-stackrox (approved):

porridge:
Rebasing to pick up new required CI job.

porridge:
No comments so far regarding docs, I assume none are needed. Can always be addressed later. Merging.

PR #4066, opened 2022-12-08
ROX-13351: Use an image tagged as latest in quay.io/rhacs-eng for DefaultPoliciesTest

theencee:
## Description

Grabbed a nginx image (nginx:1.22-alpine) and tagged it as qa:latest. This prevents rate limiting or slow down from docker.io from impacting these tests

## Checklist

## Testing Performed

this is a test

### DIFF
--- a/qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
+++ b/qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
@@ -85,7 +85,7 @@ class DefaultPoliciesTest extends BaseSpecification {
     static final private List<Deployment> DEPLOYMENTS = [
         new Deployment()
             .setName (NGINX_LATEST)
-            .setImage ("quay.io/rhacs-eng/qa:latest") // this is docker.io/redis:7.0.8-alpine but tagged as latest
+            .setImage ("quay.io/rhacs-eng/qa:latest") // this is docker.io/nginx:1.22-alpine but tagged as latest
             .addPort (22)
             .addLabel ("app", "test")
             .setEnv([SECRET: 'true']),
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -85,7 +85,7 @@ class DefaultPoliciesTest extends BaseSpecification {
     static final private List<Deployment> DEPLOYMENTS = [
         new Deployment()
             .setName (NGINX_LATEST)
-            .setImage ("nginx")
+            .setImage ("quay.io/rhacs-eng/qa:latest") // this is docker.io/redis:7.0.8-alpine but tagged as latest
```
gavin-stackrox, line 88: why mirror redis and not nginx? there are a bunch more .groovy tests that use nginx:latest and could benefit from us defining `quay.io/rhacs-eng/qa:latest` as nginx.

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
```diff
@@ -85,7 +85,7 @@ class DefaultPoliciesTest extends BaseSpecification {
     static final private List<Deployment> DEPLOYMENTS = [
         new Deployment()
             .setName (NGINX_LATEST)
```
gavin-stackrox, line 87:
```suggestion
            .setName (QA_LATEST)
```

gavin-stackrox (changes requested):

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
theencee: I ran into flaky tests by mirroring an _existing_ nginx version as latest. Ttbomk, it's because some tests are querying for an nginx image with the same sha as the one I tagged, and then throwing up because it got the wrong tagged one. Or it picked up the wrong deployment. Effectively the tests rely on there being only one ever. 

But let me try using a brand new nginx version and tagging that as latest only.

qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy
theencee: Not needed anymore

gavin-stackrox (approved):

PR #4067, opened 2022-12-08
Integrate collection dry run error handling with save error handling

dvail:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dvail:
* master
  * **PR #4067** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4067" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4067?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -182,19 +182,8 @@ function CollectionForm({
...
-        // before validation and the declarative state has `isValid === true` and `isValidation == false`
-        // at the same time that the actual value object is invalid.
-        validationSchema
-            .validate(values)
-            .then(() => onFormChange(values))
-            .catch(() => {
-                /* Validation failed, do not propagate change */
-            });
-    }, [values, saveError, onFormChange]);
+        onFormChange(values);
```
dvail, line 185: We need to send dry run requests even when the form is invalid to get correct error handling (e.g. the user goes to create a collection, skips entering a name, and configures a collection rule), so we can discard this logic and just propagate all form changes. Since we debounce at the request level now, this won't cause a flood of requests to hit the server.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -54,104 +63,146 @@ function CollectionResults({ dryRunConfig }: CollectionResultsProps) {
...
+                });
+        },
+        [setSaveError]
+    );
 
     const fetchDryRunDebounced = useMemo(() => debounce(fetchDryRun, 800), [fetchDryRun]);
 
     useEffect(() => {
+        if (saveError) {
+            setDeployments([]);
```
dvail, line 86: Flush old data on error to prevent the display of incorrect deployments the next time the rules correctly validate.

ui/apps/platform/src/Containers/Collections/hooks/useDryRunConfiguration.ts
```diff
@@ -28,11 +28,12 @@ export default function useDryRunConfiguration(
                 dryRunConfig.resourceSelectors,
                 nextConfig.resourceSelectors
             );
-            if (isEmbeddedCollectionsChanged || isResourceSelectorChanged) {
+            const isNameChange = dryRunConfig.name !== nextConfig.name;
+            if (isEmbeddedCollectionsChanged || isResourceSelectorChanged || isNameChange) {
```
dvail, line 32: It might seem like overkill, but we need to send dry run requests on name change, since name is a part of the server side validation. This will force users to enter _some_ correct name value before previewing deployments via rules.

pedrottimark:
Off the subject of the pull request, poor eyesight reacts to error message in **Collection results** panel:
* Insult: small gray text is hard to read for the most important information.
* Injury: incomplete information for **unable to process** error.

Even with auto-scrolling behavior, the panel gives a backup which is always in view. Recommendation:
* Same font size and weight as in alert message (bold for first line, regular if there is a second line) but text color can be regular black instead of red.
* Align error lines at left even though icon and title have centered horizontal alignment.

pedrottimark (approved):
One comment and two questions about `saveError` which you wisely plan to rename:
* Is it limited to request errors, or now includes frontend errors?
* Can you show an example of simulated error because the request itself failed?

dvail:
@pedrottimark Good catch about the missing additional information in the sidebar error. How does this representation look?
image

And here is a visual of what would happen if the server returns an error (HTTP 400 here), with the wording slightly modified from the code in this PR.
image

pedrottimark:
Yes, even my poor eyes understand at a glance the updated pictures of errors in sidebar!

PR #4070, opened 2022-12-08
ROX-13582: Central telemetry configuration. Metadata response update.

0x656b694d:
## Description

This and following PRs deprecate #3819 to simplify review.

* Central telemetry configuration with no actual use (see following PRs);
* `/v1/metadata` response with telemetry configuration.
* Telemetry environment variables reused:
  * `ROX_TELEMETRY_FREQUENCY` — the data flush interval (i.e. API call to Segment);
  * `ROX_TELEMETRY_ENDPOINT` — the telemetry storage endpoint, implies default segment endpoint if left empty;
* New environment variable:
  * `ROX_TELEMETRY_STORAGE_KEY_V1` — segment write key, disables telemetry if left empty.
* `rhacs.redhat.com/tenant` central deployment label is read for the Tenant ID, which will serve as the group ID for clients, such as central, central API users, UI users etc;
* Central installation information provides the central ID.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- [ ] ~Evaluated and added CHANGELOG entry if required~
- [ ] ~Determined and documented upgrade steps~
- [ ] ~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed

Unit tests.

`/v1/metadata` response:
```
{
  "version": "",
  "buildFlavor": "development",
  "releaseBuild": false,
  "licenseStatus": "VALID",
  "userId": "3wf49KZA31rMxvgqCKVi8o3ZvINhjDqx04G5r16+Y3g=",
  "telemetryEndpoint": "",
  "storageKeyV1": "..."
}
```

PR #4072, opened 2022-12-08
ROX-13582: Intercepting API calls

0x656b694d:
## Description

Another piece of #3819.
API call interceptor, that issues various events from configured filters.

`ROX_TELEMETRY_API_WHITELIST` environment variable may be set on the central deployment with a comma separated list of API paths (or "*" for any path except a hardcoded list of ignored ones), which will trigger generic "API Call" events, with three properties:
* Path — the API Path;
* Code — the service response status code;
* User-Agent — the client's User Agent;
* HTTP Method, or something alike computed from gRPC method.

Other implemented events are:
* "Cluster Registered", for when a cluster is posted to the DB;
* "Cluster Initialized", for when a cluster changes its state from `UNINITIALIZED` to something;
* "roxctl", for when an API is called by `roxctl`.

## Checklist

## Testing Performed

Unit tests.

Example of an "API Call" event received by Segment:
```json
{
  "context": {
    "Client ID": "2a3e1829-8f84-40c1-a761-006f07a59666",
    "Client Name": "Central",
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    }
  },
  "event": "API Call",
  "integrations": {},
  "messageId": "59a4f17d-616f-4588-bfc0-c88890330441",
  "originalTimestamp": "2022-12-13T17:51:42.079249978Z",
  "properties": {
    "Code": 500,
    "Method": "GET",
    "Path": "/api/extensions/scannerdefinitions",
    "User-Agent": "Go-http-client/1.1"
  },
  "receivedAt": "2022-12-13T17:52:18.204Z",
  "sentAt": "2022-12-13T17:52:18.113Z",
  "timestamp": "2022-12-13T17:51:42.170Z",
  "type": "track",
  "userId": "fxVLkxH4kOj9UdfZOP3Rcz8Yw/64sAyZgoMXJ6cGynE=",
  "writeKey": "..."
}
```

### DIFF
--- a/central/main.go
+++ b/central/main.go
@@ -165,6 +165,7 @@ import (
 	"github.com/stackrox/rox/pkg/grpc/authz/or"
 	"github.com/stackrox/rox/pkg/grpc/authz/perrpc"
 	"github.com/stackrox/rox/pkg/grpc/authz/user"
+	"github.com/stackrox/rox/pkg/grpc/client/authn/basic"
 	"github.com/stackrox/rox/pkg/grpc/errors"
 	"github.com/stackrox/rox/pkg/grpc/routes"
 	"github.com/stackrox/rox/pkg/httputil"
@@ -538,6 +539,10 @@ func startGRPCServer() {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
 		config.HTTPInterceptors = append(config.HTTPInterceptors, cfg.GetHTTPInterceptor())
 		config.UnaryInterceptors = append(config.UnaryInterceptors, cfg.GetGRPCInterceptor())
+		// Central adds itself to the tenant group, with no group properties:
+		cfg.Telemeter().Group(cfg.GroupID, cfg.ClientID, nil)
+		// Add the local admin user as well, with no extra group properties:
+		cfg.Telemeter().Group(cfg.GroupID, cfg.HashUserID(basic.DefaultUsername, basicAuthProvider.ID()), nil)
 	}
 
 	// Before authorization is checked, we want to inject the sac client into the context.
--- a/central/main.go
+++ b/central/main.go
@@ -535,6 +535,11 @@ func startGRPCServer() {
 	)
 	config.HTTPInterceptors = append(config.HTTPInterceptors, observe.AuthzTraceHTTPInterceptor(authzTraceSink))
 
+	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
+		config.HTTPInterceptors = append(config.HTTPInterceptors, cfg.GetHTTPInterceptor())
+		config.UnaryInterceptors = append(config.UnaryInterceptors, cfg.GetGRPCInterceptor())
+	}
+
 	// Before authorization is checked, we want to inject the sac client into the context.
 	config.PreAuthContextEnrichers = append(config.PreAuthContextEnrichers,
 		centralSAC.GetEnricher().GetPreAuthContextEnricher(authzTraceSink),
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -76,8 +76,7 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 			StorageKey:   key,
 			Endpoint:     env.TelemetryEndpoint.Setting(),
 			PushInterval: env.TelemetryFrequency.DurationSetting(),
-		},
-		map[string]any{
+		}, map[string]any{
 			"Central version":    version.GetMainVersion(),
 			"Chart version":      version.GetChartVersion(),
 			"Orchestrator":       orchestrator,
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -19,6 +19,10 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	apiPathsAnnotation = "rhacs.redhat.com/telemetry-apipaths"
+)
+
 var (
 	config *phonehome.Config
 	once   sync.Once
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -2,6 +2,7 @@ package centralclient
 
 import (
 	"context"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/installation/store"
@@ -9,6 +10,7 @@ import (
 	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/rox/pkg/sync"
 	"github.com/stackrox/rox/pkg/telemetry/phonehome"
 	"github.com/stackrox/rox/pkg/version"
@@ -47,6 +49,9 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 		return nil, nil, errors.Wrap(err, "cannot get central deployment")
 	}
 
+	paths := central.GetAnnotations()[apiPathsAnnotation]
+	trackedPaths = set.NewFrozenSet(strings.Split(paths, ",")...)
+
 	orchestrator := storage.ClusterType_KUBERNETES_CLUSTER.String()
 	if env.OpenshiftAPI.BooleanSetting() {
 		orchestrator = storage.ClusterType_OPENSHIFT_CLUSTER.String()
@@ -71,7 +76,8 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 			StorageKey:   key,
 			Endpoint:     env.TelemetryEndpoint.Setting(),
 			PushInterval: env.TelemetryFrequency.DurationSetting(),
-		}, map[string]any{
+		},
+		map[string]any{
 			"Central version":    version.GetMainVersion(),
 			"Chart version":      version.GetChartVersion(),
 			"Orchestrator":       orchestrator,
@@ -98,6 +104,18 @@ func InstanceConfig() *phonehome.Config {
 		}
 		log.Info("Central ID: ", config.ClientID)
 		log.Info("Tenant ID: ", config.GroupID)
+		log.Info("API path telemetry enabled for: ", trackedPaths)
+
+		for event, funcs := range interceptors {
+			for _, f := range funcs {
+				config.AddInterceptorFunc(event, f)
+			}
+		}
+		// Central adds itself to the tenant group, adding its properties to the
+		// group properties:
+		config.Telemeter().Group(config.GroupID, config.ClientID, props)
+		// Add the local admin user as well, with no extra group properties:
+		config.Telemeter().Group(config.GroupID, config.HashUserID("admin", ""), nil)
 
 		config.Gatherer().AddGatherer(func(ctx context.Context) (map[string]any, error) {
 			return props, nil
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -114,11 +114,6 @@ func InstanceConfig() *phonehome.Config {
 				config.AddInterceptorFunc(event, f)
 			}
 		}
-		// Central adds itself to the tenant group, adding its properties to the
-		// group properties:
-		config.Telemeter().Group(config.GroupID, config.ClientID, props)
-		// Add the local admin user as well, with no extra group properties:
-		config.Telemeter().Group(config.GroupID, config.HashUserID("admin", ""), nil)
 
 		config.Gatherer().AddGatherer(func(ctx context.Context) (map[string]any, error) {
 			return props, nil
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -19,11 +19,9 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-const (
-	apiPathsAnnotation = "rhacs.redhat.com/telemetry-apipaths"
-)
-
 var (
+	apiWhiteList = env.RegisterSetting("ROX_TELEMETRY_API_WHITELIST", env.AllowEmpty())
+
 	config *phonehome.Config
 	once   sync.Once
 	log    = logging.LoggerForModule()
@@ -53,8 +51,7 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 		return nil, nil, errors.Wrap(err, "cannot get central deployment")
 	}
 
-	paths := central.GetAnnotations()[apiPathsAnnotation]
-	trackedPaths = set.NewFrozenSet(strings.Split(paths, ",")...)
+	trackedPaths = set.NewFrozenSet(strings.Split(apiWhiteList.Setting(), ",")...)
 
 	orchestrator := storage.ClusterType_KUBERNETES_CLUSTER.String()
 	if env.OpenshiftAPI.BooleanSetting() {
@@ -89,7 +86,7 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 }
 
 // InstanceConfig collects the central instance telemetry configuration from
-// central Deployment labels and annotations, installation store and
+// central Deployment labels and environment variables, installation store and
 // orchestrator properties. The collected data is used for configuring the
 // telemetry client.
 func InstanceConfig() *phonehome.Config {
--- a/central/telemetry/centralclient/interceptors.go
+++ b/central/telemetry/centralclient/interceptors.go
@@ -1,0 +1,104 @@
+package centralclient
+
+import (
+	"strings"
+	"sync"
+
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/set"
+	"github.com/stackrox/rox/pkg/telemetry/phonehome"
+)
+
+var (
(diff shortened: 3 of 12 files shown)


0x656b694d:
* master
  * **PR #4070** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4070" target="_blank">Graphite</a> 
    * **PR #4071** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4071" target="_blank">Graphite</a> 
      * **PR #4072** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4072" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4072?utm_source=stack-comment).

central/telemetry/centralclient/instance_config.go
```diff
@@ -18,7 +20,8 @@ import (
 )
 
 const (
-	tenantIDLabel = "rhacs.redhat.com/tenant"
+	apiPathsAnnotation = "rhacs.redhat.com/telemetry-apipaths"
```
dhaus67, line 23: I'd have the same question as in the previous PR within the series: whether we should move this to a more generic, importable location to re-use it within the fleetshard-synchronizer code.

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,103 @@
...
+		props["User-Agent"] = rp.UserAgent
+		props["Method"] = rp.GetMethod()
+		props["Protocol"] = rp.GetProtocol()
+		return true
+	}
+	return false
+}
+
+// Adds Post Cluster call specific properties to the Post Cluster event.
+func postCluster(rp *phonehome.RequestParams, props map[string]any) bool {
```
dhaus67, line 49: (same for `putCluster`): I feel like both of those would be good to be covered by unit tests.

pkg/telemetry/phonehome/request_params.go
```diff
@@ -0,0 +1,59 @@
...
+	UserAgent string
+	UserID    authn.Identity
+	Path      string
+	Code      int
+	GRPCReq   any
+	HTTPReq   *http.Request
+}
+
+// GetProtocol returns HTTP for requests with HTTP request and gRPC otherwise.
+func (rp *RequestParams) GetProtocol() string {
```
dhaus67, line 21: I think this is more of an edge case and not something we necessarily need to address now:

We provide the option to also use gRPC when the server is exposed behind load balancers which do not necessarily support it (there's https://github.com/stackrox/go-grpc-http1 which covers this).

This means, internally while the API call _may_ be a gRPC one, the protocol it actually uses _may_ be something different (e.g. wss).

Internally, we use this for all sorts of clients (e.g. roxctl, sensor <-> central, etc.). I don't know the full history about this, but probably we ran into some issues within some environments, which lead to us coming up with this solution.

I'm not sure whether there's the desire to also collect the _specific_ protocol used, to further determine the sort of environments customers are using, but it might be worth to consider in the future (essentially, what I'm trying to say is that protocol `gRPC` a.k.a. `http2` _may not be correct_ for all gRPC calls).

central/telemetry/centralclient/instance_config.go
0x656b694d: This one is really central specific. We've discussed this annotation with @rukletsov, and looks like we want to move it to an env var, as _we_ don't use annotations for configuration today.

central/telemetry/centralclient/instance_config.go
```diff
@@ -47,6 +53,9 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 		return nil, nil, errors.Wrap(err, "cannot get central deployment")
 	}
 
+	paths := central.GetAnnotations()[apiPathsAnnotation]
```
stehessel, line 56: As discussed this should be converted to an env variable.

pkg/telemetry/phonehome/request_params.go
0x656b694d: I agree that the existing protocol detection is misleading. And, as was not explicitly requested, I'd rather remove this field from the report.

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,107 @@
...
+		"API Call":            {apiCall},
+		"Cluster Registered":  {clusterRegistered},
+		"Cluster Initialized": {clusterInitialized},
+		"roxctl":              {roxctl},
+	}
+)
+
+// Adds Path, Code and User-Agent properties to API Call events for the API
+// paths which start from the prefixes specified in the
+// rhacs.redhat.com/telemetry-apipaths central deployment annotation
```
stehessel, line 29: This should be based on the API whitelist env variable now, right?

central/telemetry/centralclient/interceptors_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"testing"
+
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/set"
+	"github.com/stackrox/rox/pkg/telemetry/phonehome"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_clusterRegistered(t *testing.T) {
```
stehessel, line 12: I find these tests a little bit hard to follow with the number of asserts. Could you maybe add some comments to make the structure and purpose of the test more obvious?

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,107 @@
...
+
+var (
+	trackedPaths set.FrozenSet[string]
+	ignoredPaths = []string{"/v1/ping", "/v1.PingService/Ping", "/v1/metadata", "/static/"}
+
+	uninitializedClusters = set.NewSet[string]()
+	mux                   = &sync.Mutex{}
+
+	interceptors = map[string][]phonehome.Interceptor{
+		"API Call":            {apiCall},
```
stehessel, line 20: I assume the plan is to start with a generic API capture and then replace them more customised events down the road?

stehessel:
LGTM overall, just the minor comments.

central/telemetry/centralclient/interceptors.go
0x656b694d: We actually start with 3 other events (sensor cluster registration/initialization and roxctl), while the "API Call" is there as a generic configurable "backdoor" for tracking specific calls without recompiling.

stehessel (approved):

pkg/telemetry/phonehome/client_config.go
```diff
@@ -19,6 +24,12 @@ var (
 // services tenant ID. The value of the label becomes the group ID if not empty.
 const TenantIDLabel = "rhacs.redhat.com/tenant"
 
+// Interceptor is a function which will be called on every API call if none of
+// the previous interceptors in the chain returned false.
```
rukletsov, line 28: Here there is no chain and no workflow logic, so I'd move this comment to:
* `track()` describing the workflow logic
* `AddInterceptorFunc()` saying that the order in which interceptors are added might matter

Here I'd describe when an `Interceptor` instance is supposed to return `true` and when `false`.

pkg/telemetry/phonehome/client_config.go
```diff
@@ -82,3 +97,43 @@ func (cfg *Config) Telemeter() Telemeter {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			statusTrackingWriter := httputil.NewStatusTrackingWriter(w)
+			handler.ServeHTTP(statusTrackingWriter, r)
+			rp := getHTTPRequestDetails(r.Context(), r, statusCodeToError(statusTrackingWriter.GetStatusCode()))
+			go cfg.track(rp)
+		})
+	}
+}
+
+func statusCodeToError(code *int) error {
```
rukletsov, line 134: We need this for the second time — maybe it's time to extract this into `pkg/httputil/`?

pkg/telemetry/phonehome/interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+	erroxGRPC "github.com/stackrox/rox/pkg/errox/grpc"
+	"github.com/stackrox/rox/pkg/grpc/authn"
+	grpcError "github.com/stackrox/rox/pkg/grpc/errors"
+	"github.com/stackrox/rox/pkg/grpc/requestinfo"
+	"github.com/stackrox/rox/pkg/sync"
+	"google.golang.org/grpc"
+)
+
+var (
+	mux = &sync.Mutex{}
```
rukletsov, line 17: I find it hard to track vars defined in one file and used in other files. IIUC, this is only used to guard write access to `Config.interceptors`. If so, then I think:
* it shall at least be declared in `client_config.go`
* but actually it should be part of `Config`: `Config.interceptorsLock` because otherwise a single mutex is used for all `Config`s (I know we have a singleton but this is not clear here and hence breaks the abstraction)
* `track()` should likely copy `cfg.interceptors` under the same mutex
* probably should be a `sync.RWMutex`

pkg/telemetry/phonehome/client_config.go
```diff
@@ -19,6 +24,12 @@ var (
 // services tenant ID. The value of the label becomes the group ID if not empty.
 const TenantIDLabel = "rhacs.redhat.com/tenant"
 
+// Interceptor is a function which will be called on every API call if none of
+// the previous interceptors in the chain returned false.
+// An Interceptor function may add custom properties to the props map so that
+// they appear in the event.
+type Interceptor func(rp *RequestParams, props map[string]any) bool
```
rukletsov, line 31: Looks like `props` is only expected to return values,, any reason it's passed as a function param?

pkg/telemetry/phonehome/interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func (cfg *Config) track(rp *RequestParams) {
+	id := cfg.HashUserAuthID(rp.UserID)
+	for event, funcs := range cfg.interceptors {
+		props := map[string]any{}
```
rukletsov, line 23: Coming back to the `props` discussion: if you only want a single interceptor to do meaningful work, then `props` should be a retvalue. If you envision interceptor to leave a trace in the same global to their scope `props` object _regardless whether they return true or false_, then it's not the interceptor who should decide on how collisions should be handled.

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,107 @@
...
+	"strings"
+
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/set"
+	"github.com/stackrox/rox/pkg/sync"
+	"github.com/stackrox/rox/pkg/telemetry/phonehome"
+)
+
+var (
+	trackedPaths set.FrozenSet[string]
```
rukletsov, line 13: I find it hard to reason about the lifetime, ownership, and initialization of `trackedPaths`. It is initialized in `getInstanceConfig()` which is called once as part of `phonehome.Config` singleton initialization. So there is an implicit assumption that `getInstanceConfig()` can (or should) only be called once. 

I'd suggest to make it part of `Config` and pass into interceptors from the `Config` singleton.

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,107 @@
...
+	ignoredPaths = []string{"/v1/ping", "/v1.PingService/Ping", "/v1/metadata", "/static/"}
+
+	uninitializedClusters = set.NewSet[string]()
+	mux                   = &sync.Mutex{}
```
rukletsov, line 17: If this is only to guard access to `uninitializedClusters`, let's call it so.

central/telemetry/centralclient/interceptors.go
```diff
@@ -0,0 +1,107 @@
...
+	interceptors = map[string][]phonehome.Interceptor{
+		"API Call":            {apiCall},
+		"Cluster Registered":  {clusterRegistered},
+		"Cluster Initialized": {clusterInitialized},
+		"roxctl":              {roxctl},
+	}
+)
+
+// Adds Path, Code, User-Agent and Method properties to API Call events for the
+// API paths specified in the ROX_TELEMETRY_API_WHITELIST central deployment
```
rukletsov, line 28: As someone not familiar with the codebase, I don't understand how `ROX_TELEMETRY_API_WHITELIST` is related to this function. One should bring context from different place to understand that `ROX_TELEMETRY_API_WHITELIST` impacts `apiWhiteList` which in turn impacts `trackedPaths` which is used here (but initialized in a different file).

rukletsov (changes requested):

central/telemetry/centralclient/instance_config.go
```diff
@@ -18,6 +20,8 @@ import (
 )
 
 var (
+	apiWhiteList = env.RegisterSetting("ROX_TELEMETRY_API_WHITELIST", env.AllowEmpty())
```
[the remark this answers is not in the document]
0x656b694d: This is a central specific setting, not to be shared with other telemetry clients.

central/main.go
```diff
@@ -535,6 +536,15 @@ func startGRPCServer() {
 	)
 	config.HTTPInterceptors = append(config.HTTPInterceptors, observe.AuthzTraceHTTPInterceptor(authzTraceSink))
 
+	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
+		config.HTTPInterceptors = append(config.HTTPInterceptors, cfg.GetHTTPInterceptor())
+		config.UnaryInterceptors = append(config.UnaryInterceptors, cfg.GetGRPCInterceptor())
+		// Central adds itself to the tenant group, with no group properties:
+		cfg.Telemeter().Group(cfg.GroupID, cfg.ClientID, nil)
+		// Add the local admin user as well, with no extra group properties:
+		cfg.Telemeter().Group(cfg.GroupID, cfg.HashUserID(basic.DefaultUsername, basicAuthProvider.ID()), nil)
```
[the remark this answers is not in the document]
0x656b694d: It could, but here we've got `basicAuthProvider` available.

pkg/telemetry/phonehome/request_params.go
```diff
@@ -0,0 +1,29 @@
...
+func (rp *RequestParams) GetMethod() string {
+	switch {
+	case rp.HTTPReq == nil:
+		return rp.Path // i.e. in the form of /service/method for gRPC requests.
+	case rp.HTTPReq.Method != "":
+		return rp.HTTPReq.Method
+	default:
+		return http.MethodGet
+	}
+}
```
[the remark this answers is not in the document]
0x656b694d: `grpc.Method()` function have this comment:
```
// Method returns the method string for the server context.  The returned
// string is in the format of "/service/method".
```
I considered consistency with this method.

Go `http.Request` mentions this:
```
// Method specifies the HTTP method (GET, POST, PUT, etc.).
// For client requests, an empty string means GET.
```
That's why `GET` is returned if it's empty. But normally it shouldn't happen.

pkg/telemetry/phonehome/client_config.go
```diff
@@ -82,3 +97,43 @@ func (cfg *Config) Telemeter() Telemeter {
...
+func (cfg *Config) GetHTTPInterceptor() httputil.HTTPInterceptor {
+	return func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			statusTrackingWriter := httputil.NewStatusTrackingWriter(w)
+			handler.ServeHTTP(statusTrackingWriter, r)
+			rp := getHTTPRequestDetails(r.Context(), r, statusCodeToError(statusTrackingWriter.GetStatusCode()))
+			go cfg.track(rp)
+		})
+	}
+}
```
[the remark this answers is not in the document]
0x656b694d: These are methods of the Config struct, so they're here close to the definition.

pkg/telemetry/phonehome/client_config.go
```diff
@@ -82,3 +97,43 @@ func (cfg *Config) Telemeter() Telemeter {
...
 }
+
+// AddInterceptorFunc appends the custom list of telemetry interceptors with the
+// provided function.
+func (cfg *Config) AddInterceptorFunc(event string, f Interceptor) {
+	mux.Lock()
+	defer mux.Unlock()
+	if cfg.interceptors == nil {
+		cfg.interceptors = make(map[string][]Interceptor, 1)
+	}
```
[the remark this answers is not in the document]
0x656b694d: There's no c-tor for Config because there would be too many arguments to pass to it. Instead, a client just fills the required fields of the struct. `interceptors` map is not required to be initialized unless it's going to be used.

pkg/telemetry/phonehome/client_config.go
0x656b694d: The idea was that a chain of interceptors (per event type) could augment this map one after another. For example, there could be a filtering interceptor that would filter the requests and possibly add some generic properties, and the second one would add more specific properties.
Originally, I was thinking about adding the current `roxctl` event as the second interceptor for `API Call` event, but then made it a separate event.
Today, there's no need for this complexity, as all the events have a unique interceptor.

pkg/telemetry/phonehome/interceptor.go
0x656b694d: Filtering interceptors in the chain will decide if the event will fire or not.

central/telemetry/centralclient/interceptors.go
0x656b694d: `trackedPath` is a central specific feature, so it shouldn't be part of `phonehome.Config`. 
`getInstanceConfig()` is a central telemetry configuration initialization function, which initializes this central specific package level variable.

PR #4078, opened 2022-12-09
ROX-10895: Ensure namespaced bindings specify cluster role correctly

dhaus67:
## Description

This is a pre-factor for the follow-up PR.
For more information on the issue at hand, refer to the child PR #4031.

Within [Kubernetes RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding), there's the concept of Roles, ClusterRoles, RoleBindings, and ClusterRoleBindings.

A RoleBinding may bind either a role or a cluster role to a specific namespace.
A ClusterRoleBinding may bind a cluster role cluster-wide.

Previously, when we detected role bindings, we automatically assumed `ClusterRole = false`.
However, this is not entirely correct. A role binding may also bind a cluster role, in which case we shall set `ClusterRole = true`.

This PR fixes this and will ensure that we now correctly detect `ClusterRole = true` also for role bindings.
Since the data will be sent from sensor to central, and updated periodically, there's no need for a separate migration of the datastore.

## Checklist

## Testing Performed

- see CI.


dhaus67:
* master
  * **PR #4078** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4078" target="_blank">Graphite</a>  👈
    * **PR #4031** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4031" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4078?utm_source=stack-comment).

janisz (approved):

clickboo (approved):
lgtm

central/graphql/resolvers/generated.go
```diff
@@ -1320,6 +1320,7 @@ func registerGeneratedTypes(builder generator.SchemaBuilder) {
 		"skipTLSVerify: Boolean!",
 	}))
 	utils.Must(builder.AddType("Syslog", []string{
+		"extraFields: [KeyValuePair]!",
```
rukletsov, line 1323: How changes to this file are related to the rest of the PR?

dhaus67: So, not entirely sure _why_ the change is happening, but its done by the code generation (i.e. `make go-generated-srcs`).

I'll try removing those and check whether CI is passing, otherwise the change has to be made (not _entirely_ sure why the CI isn't failing in the first place though).

rukletsov (approved):

PR #4083, opened 2022-12-09
Dashrews/migration prob

dashrews78:
## Description

We need to be able to not panic when a legacy migration is added as part of a patch release or for some other reason.  We can take advantage of the fact that if we determine that Postgres is the source database, we will pass nils to the migrations for both Bolt and Rocks.  If we determine we need to migrate from legacy we are ensured that Bolt and Rocks are not nil because we fail before calling the runner if they are.  So here we are just taking the Postgres migrations and performing a nil check on the legacy DB and skipping that migration if it is nil.

This solution is a quick and dirty one based on the state that we have saved currently for migrations.  A larger refactor and storing additional state is also an option, but that would take longer to release and would prove unnecessary in a couple of releases.

## Checklist

## Testing Performed

Manual tests for now.  Spin up an environment from earlier in 3.73.  Upgrade to these images and note in the logs that migration 169 was skipped.  #4003 will have added this case to CI runs upon its completion.

### DIFF
--- a/migrator/types/migration.go
+++ b/migrator/types/migration.go
@@ -37,4 +37,10 @@ type Migration struct {
 	// All other (optional) metadata can be whatever the user desires, and has no bearing on the
 	// functioning of the migrator.
 	VersionAfter *storage.Version
+	// LegacyToPostgres tells us if the migration is a legacy database to postgres migration.
+	// This allows us the ability to ensure that all databases exist before executing the migration.
+	// Particularly helpful in the event a patch release caused a new legacy migration to be added.
+	// This will allow us to skip migrations when the legacy databases are not passed to the runner
+	// because we have already migrated to Postgres.
+	LegacyToPostgres bool
 }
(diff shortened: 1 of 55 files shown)


migrator/types/migration.go
```diff
@@ -37,4 +37,10 @@ type Migration struct {
 	// All other (optional) metadata can be whatever the user desires, and has no bearing on the
 	// functioning of the migrator.
 	VersionAfter *storage.Version
+	// LegacyToPostgres tells us if the migration is a legacy database to postgres migration.
+	// This allows us the ability to ensure that all databases exist before executing the migration.
+	// Particularly helpful in the event a patch release caused a new legacy migration to be added.
+	// This will allow us to skip migrations when the legacy databases are not passed to the runner
+	// because we have already migrated to Postgres.
+	LegacyToPostgres bool
```
keyallis, line 45: I could be wrong but I think this would only be needed for the #4084 flow

dashrews78: That is correct.  I started down this path before I started to go 2 separate paths.  Will clean up depending upon which PR lives and which one doesn't.

PR #4100, opened 2022-12-12
[ROX-13890] Update Vuln reporting to use collections to match deployments when collections' feature flag is enabled

charmik-redhat:
## Description

- When collections feature flag is enabled, scope_id field in reportConfiguration will be treated as collection id.
- Vuln reporting workflow will use the collections id to fetch the collections and then create a report based on the deployments that match that collection.
- The final results will be primarily grouped by cluster name and secondarily by namespace.

## Checklist

## Testing Performed

Manual testing by creating a collection and using it in a vuln report

517FF47A-E277-4624-8B39-FE821C7CE345

1B68E5D4-9168-4FD7-A7FF-14B6DF967ACE_1_105_c

CF36F461-6352-44F4-9C33-3EB9D114A2DE_1_105_c

Created separate task to add unit tests, and performing manual testing for this one.

### DIFF
--- a/central/reportconfigurations/service/service_impl.go
+++ b/central/reportconfigurations/service/service_impl.go
@@ -9,12 +9,14 @@ import (
 	notifierDataStore "github.com/stackrox/rox/central/notifier/datastore"
 	"github.com/stackrox/rox/central/reportconfigurations/datastore"
 	"github.com/stackrox/rox/central/reports/manager"
+	collectionDS "github.com/stackrox/rox/central/resourcecollection/datastore"
 	accessScopeStore "github.com/stackrox/rox/central/role/datastore"
 	"github.com/stackrox/rox/central/role/resources"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/permissions"
 	"github.com/stackrox/rox/pkg/errox"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/grpc/authz"
 	"github.com/stackrox/rox/pkg/grpc/authz/perrpc"
 	"github.com/stackrox/rox/pkg/grpc/authz/user"
@@ -48,10 +50,12 @@ var (
 type serviceImpl struct {
 	v1.UnimplementedReportConfigurationServiceServer
 
-	manager           manager.Manager
-	reportConfigStore datastore.DataStore
-	notifierStore     notifierDataStore.DataStore
-	accessScopeStore  accessScopeStore.DataStore
+	manager                 manager.Manager
+	reportConfigStore       datastore.DataStore
+	notifierStore           notifierDataStore.DataStore
+	accessScopeStore        accessScopeStore.DataStore
+	collectionDataStore     collectionDS.DataStore
+	collectionQueryResolver collectionDS.QueryResolver
 }
 
 func (s *serviceImpl) GetReportConfigurations(ctx context.Context, query *v1.RawQuery) (*v1.GetReportConfigurationsResponse, error) {
@@ -203,12 +207,19 @@ func (s *serviceImpl) validateReportConfiguration(ctx context.Context, config *s
 		}
 	}
 
-	_, found, err := s.accessScopeStore.GetAccessScope(ctx, config.GetScopeId())
-	if !found || err != nil {
-		return errors.Wrapf(errox.NotFound, "Access scope %s not found. Error: %s", config.GetScopeId(), err)
+	if features.ObjectCollections.Enabled() {
+		_, found, err := s.collectionDataStore.Get(ctx, config.GetId())
+		if !found || err != nil {
+			return errors.Wrapf(errox.NotFound, "Collection %s not found. Error: %s", config.GetScopeId(), err)
+		}
+	} else {
+		_, found, err := s.accessScopeStore.GetAccessScope(ctx, config.GetScopeId())
+		if !found || err != nil {
+			return errors.Wrapf(errox.NotFound, "Access scope %s not found. Error: %s", config.GetScopeId(), err)
+		}
 	}
 
-	_, found, err = s.notifierStore.GetNotifier(ctx, config.GetEmailConfig().GetNotifierId())
+	_, found, err := s.notifierStore.GetNotifier(ctx, config.GetEmailConfig().GetNotifierId())
 	if err != nil {
 		return errors.Wrapf(errox.NotFound, "Failed to fetch notifier %s with error %s", config.GetEmailConfig().GetNotifierId(), err)
 	}
--- a/central/reportconfigurations/service/service_impl.go
+++ b/central/reportconfigurations/service/service_impl.go
@@ -207,7 +207,7 @@ func (s *serviceImpl) validateReportConfiguration(ctx context.Context, config *s
 	}
 
 	if features.ObjectCollections.Enabled() {
-		_, found, err := s.collectionDatastore.Get(ctx, config.GetId())
+		_, found, err := s.collectionDatastore.Get(ctx, config.GetScopeId())
 		if !found || err != nil {
 			return errors.Wrapf(errox.NotFound, "Collection %s not found. Error: %s", config.GetScopeId(), err)
 		}
--- a/central/reportconfigurations/service/service_impl.go
+++ b/central/reportconfigurations/service/service_impl.go
@@ -9,7 +9,7 @@ import (
 	notifierDataStore "github.com/stackrox/rox/central/notifier/datastore"
 	"github.com/stackrox/rox/central/reportconfigurations/datastore"
 	"github.com/stackrox/rox/central/reports/manager"
-	collectionDS "github.com/stackrox/rox/central/resourcecollection/datastore"
+	collectionDataStore "github.com/stackrox/rox/central/resourcecollection/datastore"
 	accessScopeStore "github.com/stackrox/rox/central/role/datastore"
 	"github.com/stackrox/rox/central/role/resources"
 	v1 "github.com/stackrox/rox/generated/api/v1"
@@ -50,12 +50,11 @@ var (
 type serviceImpl struct {
 	v1.UnimplementedReportConfigurationServiceServer
 
-	manager                 manager.Manager
-	reportConfigStore       datastore.DataStore
-	notifierStore           notifierDataStore.DataStore
-	accessScopeStore        accessScopeStore.DataStore
-	collectionDataStore     collectionDS.DataStore
-	collectionQueryResolver collectionDS.QueryResolver
+	manager             manager.Manager
+	reportConfigStore   datastore.DataStore
+	notifierStore       notifierDataStore.DataStore
+	accessScopeStore    accessScopeStore.DataStore
+	collectionDatastore collectionDataStore.DataStore
 }
 
 func (s *serviceImpl) GetReportConfigurations(ctx context.Context, query *v1.RawQuery) (*v1.GetReportConfigurationsResponse, error) {
@@ -208,7 +207,7 @@ func (s *serviceImpl) validateReportConfiguration(ctx context.Context, config *s
 	}
 
 	if features.ObjectCollections.Enabled() {
-		_, found, err := s.collectionDataStore.Get(ctx, config.GetId())
+		_, found, err := s.collectionDatastore.Get(ctx, config.GetId())
 		if !found || err != nil {
 			return errors.Wrapf(errox.NotFound, "Collection %s not found. Error: %s", config.GetScopeId(), err)
 		}
--- a/central/reportconfigurations/service/singleton.go
+++ b/central/reportconfigurations/service/singleton.go
@@ -4,6 +4,7 @@ import (
 	notifierDataStore "github.com/stackrox/rox/central/notifier/datastore"
 	reportConfigDS "github.com/stackrox/rox/central/reportconfigurations/datastore"
 	"github.com/stackrox/rox/central/reports/manager"
+	collectionDS "github.com/stackrox/rox/central/resourcecollection/datastore"
 	accessScopeStore "github.com/stackrox/rox/central/role/datastore"
 	"github.com/stackrox/rox/pkg/sync"
 )
@@ -14,7 +15,8 @@ var (
 )
 
 func initialize() {
-	svc = New(reportConfigDS.Singleton(), notifierDataStore.Singleton(), accessScopeStore.Singleton(), manager.Singleton())
+	collectionDataStore, collectionQueryResolver := collectionDS.Singleton()
+	svc = New(reportConfigDS.Singleton(), notifierDataStore.Singleton(), accessScopeStore.Singleton(), collectionDataStore, collectionQueryResolver, manager.Singleton())
 }
 
 // Singleton provides the instance of the service to register.
--- a/central/reportconfigurations/service/singleton.go
+++ b/central/reportconfigurations/service/singleton.go
@@ -4,7 +4,7 @@ import (
 	notifierDataStore "github.com/stackrox/rox/central/notifier/datastore"
 	reportConfigDS "github.com/stackrox/rox/central/reportconfigurations/datastore"
 	"github.com/stackrox/rox/central/reports/manager"
-	collectionDS "github.com/stackrox/rox/central/resourcecollection/datastore"
+	collectionDataStore "github.com/stackrox/rox/central/resourcecollection/datastore"
 	accessScopeStore "github.com/stackrox/rox/central/role/datastore"
 	"github.com/stackrox/rox/pkg/sync"
 )
@@ -15,8 +15,8 @@ var (
 )
 
 func initialize() {
-	collectionDataStore, collectionQueryResolver := collectionDS.Singleton()
-	svc = New(reportConfigDS.Singleton(), notifierDataStore.Singleton(), accessScopeStore.Singleton(), collectionDataStore, collectionQueryResolver, manager.Singleton())
+	collectionDS, _ := collectionDataStore.Singleton()
+	svc = New(reportConfigDS.Singleton(), notifierDataStore.Singleton(), accessScopeStore.Singleton(), collectionDS, manager.Singleton())
 }
 
 // Singleton provides the instance of the service to register.
--- a/central/reports/common/query_builder.go
+++ b/central/reports/common/query_builder.go
@@ -1,12 +1,15 @@
 package common
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
 
+	collectionDataStore "github.com/stackrox/rox/central/resourcecollection/datastore"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/sac/effectiveaccessscope"
 	"github.com/stackrox/rox/pkg/search"
 )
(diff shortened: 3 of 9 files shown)


central/reportconfigurations/service/service_impl.go
```diff
@@ -203,12 +206,19 @@ func (s *serviceImpl) validateReportConfiguration(ctx context.Context, config *s
 		}
 	}
 
-	_, found, err := s.accessScopeStore.GetAccessScope(ctx, config.GetScopeId())
-	if !found || err != nil {
-		return errors.Wrapf(errox.NotFound, "Access scope %s not found. Error: %s", config.GetScopeId(), err)
+	if features.ObjectCollections.Enabled() {
+		_, found, err := s.collectionDatastore.Get(ctx, config.GetScopeId())
+		if !found || err != nil {
+			return errors.Wrapf(errox.NotFound, "Collection %s not found. Error: %s", config.GetScopeId(), err)
```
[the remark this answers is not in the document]
clickboo: If the latter, I urge the collections teams to look into it with PM since if its just collection ... it could be a collection of ... rocks! Or marbles! Or artwork.  :-)

central/reportconfigurations/service/singleton.go
```diff
@@ -14,7 +15,8 @@ var (
 )
 
 func initialize() {
-	svc = New(reportConfigDS.Singleton(), notifierDataStore.Singleton(), accessScopeStore.Singleton(), manager.Singleton())
+	collectionDS, _ := collectionDataStore.Singleton()
```
clickboo, line 18: Odd that the singleton func should return an error ... and if it does, we shouldn't ignore it. panic(err).

central/reports/common/query_builder.go
```diff
@@ -1,47 +1,64 @@
...
 	"github.com/stackrox/rox/pkg/search"
 )
 
 // ReportQuery encapsulates the cve specific fields query, and the resource scope
 // queries to be used in a report generation run
 type ReportQuery struct {
-	CveFieldsQuery string
-	ScopeQueries   []string
+	CveFieldsQuery   string
+	ScopeQueries     []string
```
clickboo, line 21: Comments to explain why both scope queries and deployment query exists would be nice

central/reports/scheduler/schedule.go
```diff
@@ -364,24 +385,42 @@ func formatMessage(rc *storage.ReportConfiguration, emailTemplate string, date t
...
+		gqlPaginationOffset := offset
 		if env.PostgresDatastoreEnabled.BooleanSetting() {
 			gqlQuery = reportQueryPostgres
+			if features.ObjectCollections.Enabled() {
+				deploymentIds, err := s.getDeploymentIDs(ctx, deploymentsQuery, int32(offset))
+				if err != nil || len(deploymentIds) == 0 {
+					return common.Result{}, err
+				}
+				scopeQuery = fmt.Sprintf("%s:%s", search.DeploymentID.String(), strings.Join(deploymentIds, ","))
+				log.Infof("ROX-12629 : deployments scopeQuery : %s", scopeQuery)
```
clickboo, line 421: remove? debug?

central/reportconfigurations/service/singleton.go
charmik-redhat: Ah, the singleton func returns both a datastore and a queryResolver instance. queryResolver has functions to generate a v1.Query from collection. Ignoring it here since config service just needs access to collection datastore.

central/reports/scheduler/schedule.go
charmik-redhat: Yeah, had added it for debugging to see if query isn't too long, will remove it.

central/reports/common/report_formatter.go
```diff
@@ -36,7 +36,8 @@ var (
 	}
 )
 
-type imageVulnerability struct {
+// ImageVulnerability data used for generating vuln reports
+type ImageVulnerability struct {
```
clickboo, line 40: Why were the changes in this file required?

charmik-redhat: Had to export `Deployment` struct in this file, so it can be used by `groupByClusterAndNamespace` func, but I didn't need to export everything else. I will change to just export Deployment.

central/reports/common/query_builder.go
```diff
@@ -1,47 +1,64 @@
...
-	scopeQueries, err := q.buildScopeQueries()
+func (q *queryBuilder) BuildQuery(ctx context.Context) (*ReportQuery, error) {
+	var err error
+	var deploymentsQuery *v1.Query
+	var scopeQueries []string
+	if !features.ObjectCollections.Enabled() {
+		scopeQueries, err = q.buildScopeQueries()
+	} else {
+		deploymentsQuery, err = q.collectionQueryResolver.ResolveCollectionQuery(ctx, q.collection)
+	}
```
md2119, line 60: Flip it to make it easier to parse.
```suggestion
	if features.ObjectCollections.Enabled() {
		deploymentsQuery, err = q.collectionQueryResolver.ResolveCollectionQuery(ctx, q.collection)
	} else {
		scopeQueries, err = q.buildScopeQueries()
	}
```

central/reports/scheduler/schedule.go
```diff
@@ -285,17 +298,25 @@ func (s *scheduler) sendReportResults(req *ReportRequest) error {
...
 
-	scope, found, err := s.roleDatastore.GetAccessScope(req.Ctx, rc.GetScopeId())
+	var found bool
+	var scope *storage.SimpleAccessScope
+	var collection *storage.ResourceCollection
+	if !features.ObjectCollections.Enabled() {
+		scope, found, err = s.roleDatastore.GetAccessScope(req.Ctx, rc.GetScopeId())
+	} else {
+		collection, found, err = s.collectionDatastore.Get(req.Ctx, rc.GetScopeId())
+	}
```
md2119, line 308: flip it

central/reports/scheduler/schedule.go
```diff
@@ -364,24 +385,42 @@ func formatMessage(rc *storage.ReportConfiguration, emailTemplate string, date t
...
-		r = append(r, resultData)
+		return r, nil
 	}
-	return r, nil
+	result, err := s.runPaginatedQuery(ctx, "", rQuery.CveFieldsQuery, rQuery.DeploymentsQuery)
+	if err != nil {
+		return nil, err
+	}
+	result = groupByClusterAndNamespace(result)
+	return []common.Result{result}, nil
```
md2119, line 404: flip it

central/reports/scheduler/schedule.go
```diff
@@ -364,24 +385,42 @@ func formatMessage(rc *storage.ReportConfiguration, emailTemplate string, date t
...
 }
 
-func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery string) (common.Result, error) {
+func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery string, deploymentsQuery *v1.Query) (common.Result, error) {
 	offset := 0
```
md2119, line 408: not your code, but can you declare a const `startOffset = 0` and assign offset to it.

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +450,43 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
...
+	}
+	results, err := s.deploymentDatastore.SearchDeployments(ctx, deploymentsQuery)
+	if err != nil {
+		return nil, err
+	}
+	ids := make([]string, 0, len(results))
+	for _, res := range results {
+		ids = append(ids, res.GetId())
+	}
+	return ids, nil
```
md2119, line 466: `return search.ResultsToIDs(results), nil`

central/reports/scheduler/schedule.go
```diff
@@ -364,24 +385,42 @@ func formatMessage(rc *storage.ReportConfiguration, emailTemplate string, date t
...
 		var gqlQuery string
+		gqlPaginationOffset := offset
 		if env.PostgresDatastoreEnabled.BooleanSetting() {
 			gqlQuery = reportQueryPostgres
+			if features.ObjectCollections.Enabled() {
+				deploymentIds, err := s.getDeploymentIDs(ctx, deploymentsQuery, int32(offset))
+				if err != nil || len(deploymentIds) == 0 {
+					return common.Result{}, err
+				}
+				scopeQuery = fmt.Sprintf("%s:%s", search.DeploymentID.String(), strings.Join(deploymentIds, ","))
```
md2119, line 420: Why do you need to fetch deployment IDs? You can pass deployment query to deployment resolver.

md2119 (changes requested):

central/reports/common/query_builder.go
keyallis: +1

central/reports/common/query_builder_test.go
```diff
@@ -66,8 +68,10 @@ var vulnFilters = &storage.VulnerabilityReportFilters{
 }
 
 func TestBuildQuery(t *testing.T) {
-	qb := NewVulnReportQueryBuilder(clusters, namespaces, accessScope, vulnFilters, time.Now())
-	rq, err := qb.BuildQuery()
+	// TODO : Skip is collections is enabled
```
keyallis, line 71: Is this task to be completed in this PR? If yes it appears that still needs to be done. If not do we have a JIRA filed for this, and can we include that reference in the comment here?

keyallis (commented):
Some comments, looks like there are some other comments from other folks so I'll leave it at that until it has another pass from you

central/reports/scheduler/schedule.go
charmik-redhat: Do you mean a global const? If local const, why would that be better than the current way?

central/reports/scheduler/schedule.go
charmik-redhat: deploymentQuery is a `*v1.Query` but graphQL needs a string query. The problem with building an equivalent string version of  deploymentQuery occurs with nested collections. Nested collections will  form queries like  
`[(Cluster: c1 AND Namespace: n1 AND Deployment: d1) OR (Cluster: c2 AND Namespace: n2 AND Deployment: d2)]` 

However, we cannot make a similar string query of the form 
`scopeQuery: (Cluster:c1+Namespace:n1+Deployment:d1) OR (Cluster:c2+Namespace:n2+Deployment:d2)` because the string query language doesn't support disjunction of conjunctions. It only supports conjunction of disjunctions where each fieldLabel can only occur once. This is why I need to fetch deploymentIDs first and then create a scopeQuery from those to pass to graphQL.

central/reports/common/query_builder_test.go
charmik-redhat: I think that should be done when collections is enabled by default. Created [ROX-13958](https://issues.redhat.com/browse/ROX-13958) and referenced it in the comment.

central/reportconfigurations/service/service_impl.go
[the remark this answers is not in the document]
charmik-redhat: That's right, keeping this open for now to get answers on whether resource collections is good enough or we need a different name for would be better.

central/reportconfigurations/service/singleton.go
charmik-redhat: @keyallis Correct, the collectionDatastore's `Singleton` does error out on failure to build the collection graph. But it crashes on error and does not return any error.

central/reportconfigurations/service/service_impl.go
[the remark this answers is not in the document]
md2119: Follow the name used in the REST API.

central/reports/scheduler/schedule.go
md2119: Add the explanation as code comment

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +451,39 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
 	return resultData, nil
 }
 
+func (s *scheduler) getDeploymentIDs(ctx context.Context, deploymentsQuery *v1.Query, offset int32) ([]string, error) {
+	deploymentsQuery.Pagination = &v1.QueryPagination{
```
md2119, line 455: Fetch deployments by one deterministic order and keep the responsibility here, with the caller. Do not rely on default ordering in the searcher.

central/reports/scheduler/schedule.go
```diff
@@ -375,13 +406,22 @@ func (s *scheduler) getReportData(ctx context.Context, rQuery *common.ReportQuer
...
+func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery string, deploymentsQuery *v1.Query) (common.Result, error) {
+	offset := paginatedQueryStartOffset
 	var resultData common.Result
 	for {
 		var gqlQuery string
+		gqlPaginationOffset := offset
 		if env.PostgresDatastoreEnabled.BooleanSetting() {
 			gqlQuery = reportQueryPostgres
+			if features.ObjectCollections.Enabled() {
+				deploymentIds, err := s.getDeploymentIDs(ctx, deploymentsQuery, int32(offset))
```
md2119, line 418: Since this is not happening in one transaction, it is likely that the ID fetched on the previous page appears in the next page because of the changing DB state. You have to make sure seen ids are not fetched again.

I do not see a point of paginating the IDs if the report itself is not paginated. graphQL is paginated because it pulls out full objects.

central/reportconfigurations/service/service_impl.go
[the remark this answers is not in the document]
charmik-redhat: The REST API uses the name "Collections".

central/reports/scheduler/schedule.go
charmik-redhat: Ah great catch, didn't think of that case. Changed it. Fetching all deploymentIDs in a single transaction now and iterating over the ids in a paginated manner to exec graphQL query.

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +439,91 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
...
+		// string equivalent of deploymentsQuery for graphQL. Because of this, we first fetch deploymentIDs from
+		// deploymentDatastore using deploymentsQuery and then build string query for graphQL using those deploymentIDs.
+		scopeQuery := fmt.Sprintf("%s:%s", search.DeploymentID.String(),
+			strings.Join(deploymentIds[offset:mathutil.MinInt(offset+numDeploymentsLimit, len(deploymentIds))], ","))
+		r, err := s.execReportDataQuery(ctx, reportQueryPostgres, scopeQuery, cveQuery, paginatedQueryStartOffset)
+		if err != nil {
+			return r, err
+		}
+		resultData.Deployments = append(resultData.Deployments, r.Deployments...)
+		offset += numDeploymentsLimit
```
keyallis, line 463: Is this increment a pagination value? This name might be changed for clarity

keyallis (approved):
LGTM, but 1 nit comment. Also I think it would be best if tests were included with this rather than having them in a separate PR

keyallis:
While these changes are being done, can you add the collection id search tag to the `scope_id` in the proto of report configurations?

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +439,91 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
...
+		deploymentsByNamespace := make(map[string][]*common.Deployment)
+		for _, deployment := range deployments {
+			deploymentsByNamespace[deployment.Namespace] = append(deploymentsByNamespace[deployment.Namespace], deployment)
+		}
+		for _, deps := range deploymentsByNamespace {
+			groupedDeployments = append(groupedDeployments, deps...)
+		}
+	}
+
+	return common.Result{Deployments: groupedDeployments}
```
md2119, line 524: Deployments IDs are fetch in the order of risk priority, but that order is lost while building the groups. Maps do not guarantee order.
Furthermore, I don't think you need maps. You can simply sort the `[]*common.Deployment` array.
```
	sort.SliceStable(deployments, func(i, j int) bool {
		if deployments[i].Cluster.GetName() == deployments[j].Cluster.GetName() {
			return deployments[i].Namespace < deployments[j].Namespace
		} 
		return deployments[i].Cluster.GetName() < deployments[j].Cluster.GetName()
	})
```

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +439,91 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
...
+		return nil, err
+	}
+	return search.ResultsToIDs(results), nil
+}
+
+func groupByClusterAndNamespace(result common.Result) common.Result {
+	groupedDeployments := make([]*common.Deployment, 0, len(result.Deployments))
+	deploymentsByCluster := make(map[string][]*common.Deployment)
+	for _, deployment := range result.Deployments {
+		clusterName := deployment.Cluster.GetName()
```
md2119, line 510: Why are you storing the entire cluster object when we only need its name?

central/reports/scheduler/schedule.go
```diff
@@ -411,6 +439,91 @@ func (s *scheduler) runPaginatedQuery(ctx context.Context, scopeQuery, cveQuery
...
+		},
+	}
+	results, err := s.deploymentDatastore.Search(ctx, deploymentsQuery)
+	if err != nil {
+		return nil, err
+	}
+	return search.ResultsToIDs(results), nil
+}
+
+func groupByClusterAndNamespace(result common.Result) common.Result {
```
md2119, line 506: Could you pass the pointer to the constructed result set instead of creating copy?

md2119 (changes requested):

central/reports/scheduler/schedule.go
charmik-redhat: Yeah, the global sort order will go away, but since we are iterating the sorted deployments and adding them to slices in  the map's value, the deployments will still be sorted within each group. This replicates the ordering that is in current vuln reports. I used maps for grouping to avoid sorting's nlogn time complexity but I agree that it affects the readability and the time for creating/deleting maps might probably be making it worse when there are large number of namespaces. So I am fine changing that to use sorting here.

That being said, I am in favor of removing the grouping altogether and just return deployments globally ordered by RiskPriority. We are sending the  results in csv format and the end user can group them if they want. I also don't see any documentation promising any ordering.

central/reports/scheduler/schedule.go
charmik-redhat: That is right, the graphQL query used was `deployments($query) { cluster { name } } ...` which was fetching the the entire cluster just to get the name, I changed that to `deployments($query) { clusterName ...` instead.

charmik-redhat:
> While these changes are being done, can you add the collection id search tag to the `scope_id` in the proto of report configurations?

Created [ROX-14220](https://issues.redhat.com/browse/ROX-14220) to make report configs searchable by scope_id. Adding  it in a separate PR so that this can be merged to unblock the unit tests PR.

central/reports/scheduler/schedule.go
md2119: Sorry, I think you misunderstood me. My comment was multi-fold:
- Is retrieving deployments by risk priority necessary? It is now ordering through this PR.
- The deployment ordering according to risk priority needs is not maintained during grouping, which made the effort to obtain deployments ordered by risk priority moot.
**- You can group the rows by sorting. The use of two maps seems unnecessary.**

Overall, I firmly believe we should not order csv rows by risk priority and keep the csv rows ordered by cluster+namespace combination because:
- Risk priority column is not included in the CSV.
- The report is a CSV report and not a deployment report. The data insights in csv are strictly focused on vulnerability and not the overall risk of the deployments.
- The readability of the csv drops dramatically without grouping. We do not care how customers manipulate the csv later, but the ACS output should be readable without sheet tools.

md2119 (changes requested):
One last comment about using sort to group instead of mem alloc

central/reports/scheduler/schedule.go
charmik-redhat: Yes that makes sense and it simplifies the implementation, thanks for clarification. I would skip the ordering by riskPriority altogether and just sort by cluster+namespace since riskPriority is not a part of csv anyways.

md2119 (approved):

PR #4106, opened 2022-12-12
Move helpers into same subfolder as Vulnerability Management Reporting integration tests

pedrottimark:
## Description

Reduce import dependencies before update when collections replace access scopes.

### Naming convention

Distinguish the **scope** of constants and functions:
1. Test-specific: `it` block for test or module scope in test file for tests.
2. Container-specific: whatever.helpers.js or whatever.selectors.js file in cypress/integration/whatever subfolder. Only rare import from other containers, like `interactAndVisitWhatever` function (for example, from Risk to Network Graph).
3. Global
    * cypress/helpers/whatever.js for example, basicAuth.js or features.js
    * cypress/selectors/whatever.js for example, navigation.js or pagination.js

End goals:
1. Move most of cypress/helpers/whatever.js to container subfolders as whatever.helpers.js files.
2. Move most of cypress/constants/whatever.js to container subfolders as whatever.selectors.js files.
3. Encapsulate most of apiEndpoints.js in container-specific helpers files.

### Changed files

1. Edit cypress/constants/VulnManagementPage.js
    * Delete `reportSection` object.

2. Edit and move cypress/helpers/reporting.js to reporting.**helpers**.js
    * Replace import from apiEndpoints with literal strings.
    * Factor out title.
    * Encapsulate both action and assertion about left navigation.

3. Edit and rename cypress/integrations/vulnmanagement/reporting/**displayReports**.test.js as **reportingTable**.test.js
    * Replace import from selectors for HTML table elements.
    * Replace `describe` title with titles from left nav and delete second `describe` level.
    * Delete `should('be.visible')` method calls.
    * Replace `should('contain', 'whatever')` method calls with `:contains("whatever")` pseudo-selectors. Although the former are retryable assertions, give our eyes and fingers practice with the latter pattern, because it is more robust than longer method chains.

4. Edit and rename cypress/integrations/vulnmanagement/reporting/**manageReports**.test.js as **reportingForm**.test.js
    * Replace `describe` title with titles from left nav and delete second `describe` level.
    * Call `interactAndVisitVulnerabilityReporting` helper function.

## Checklist

## Testing Performed


dvail (approved):
🚢 LGTM

PR #4108, opened 2022-12-12
ROX-12620: Deprecate Policy and VulnerabilityReports resources

dhaus67:
## Description

With the introduction of the `WorkflowAdministration` resource, the long-term goal is to replace both `Policy` and `VulnerabilityReports` resources in favor of it.

This PR will mark the `Policy` and `VulnerabilityReports` resources as `deprecated`.
Marking them as deprecated will have the following effect:
- There will be a "replacing" resources assigned to it (in this case `WorkflowAdministration`). SAC will give access when the context has access to _either_ resources.
- There will be a banner within the access control page, which contains a note about the deprecation as well as the version in which it will be removed.

## Checklist
- [ ] Investigated and inspected CI test results
- [X] Unit test and regression tests added
- [X] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

- see CI.

PR #4111, opened 2022-12-12
ROX-13757: Workflow Administration Migration

keyallis:
## Description

Providing a new postgres -> postgres migration to allow permissions sets that have read/write permissions on report configuration to have the same level of access on the new Workflow Administration permission.

## Checklist

## Testing Performed

Manually verified that upgrading a cluster from a previous version to this current version provides the new sac resource.

### DIFF
--- a/central/role/datastore/singleton.go
+++ b/central/role/datastore/singleton.go
@@ -155,12 +155,13 @@ var vulnReportingDefaultRoles = map[string]roleAttributes{
 		postgresID:  vulnReporterPermissionSetID,
 		description: "For users: use it to create and manage vulnerability reporting configurations for scheduled vulnerability reports",
 		resourceWithAccess: []permissions.ResourceWithAccess{
-			// TODO: ROX-13888 Replace VulnerabilityReports with WorkflowAdministration.
-			permissions.View(resources.VulnerabilityReports),   // required for vuln report configurations
-			permissions.Modify(resources.VulnerabilityReports), // required for vuln report configurations
-			permissions.View(resources.Role),                   // required for scopes
-			permissions.View(resources.Image),                  // required to gather CVE data for the report
-			permissions.View(resources.Integration),            // required for vuln report configurations
+			permissions.View(resources.WorkflowAdministration),   // required for vuln report configurations
+			permissions.Modify(resources.WorkflowAdministration), // required for vuln report configurations
+			permissions.View(resources.Role),                     // required for scopes
+			permissions.View(resources.Image),                    // required to gather CVE data for the report
+			permissions.View(resources.Integration),              // required for vuln report configurations
+			permissions.View(resources.VulnerabilityReports),     // required for vuln report configurations prior to collections
+			permissions.Modify(resources.VulnerabilityReports),   // required for vuln report configurations prior to collections
 		},
 	},
 }
--- a/central/role/datastore/singleton.go
+++ b/central/role/datastore/singleton.go
@@ -50,20 +50,25 @@ func Singleton() DataStore {
 		// Which role format is used is determined solely by the feature flag.
 		ds = New(roleStorage, permissionSetStorage, accessScopeStorage)
 
-		// TODO ROX-13888 this filtering can be removed once we are fully migrated
-		if features.ObjectCollections.Enabled() {
-			roleAttribute := vulnReportingDefaultRoles[rolePkg.VulnReporter]
-			collectionsResources := make([]permissions.ResourceWithAccess, 0, len(roleAttribute.resourceWithAccess))
-			for _, permission := range roleAttribute.resourceWithAccess {
-				if permission.Resource == resources.Role || permission.Resource == resources.VulnerabilityReports {
-					continue
-				}
-				collectionsResources = append(collectionsResources, permission)
-			}
-			roleAttribute.resourceWithAccess = collectionsResources
-		}
-
 		for r, a := range vulnReportingDefaultRoles {
+			// TODO ROX-13888 this filtering can be removed once we are fully migrated
+			if features.ObjectCollections.Enabled() && r == rolePkg.VulnReporter {
+				collectionResourceWithAccess := make([]permissions.ResourceWithAccess, 0, len(a.resourceWithAccess))
+				for _, permission := range a.resourceWithAccess {
+					if permission.Resource == resources.Role || permission.Resource == resources.VulnerabilityReports {
+						continue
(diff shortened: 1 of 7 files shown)


central/role/datastore/singleton.go
```diff
@@ -155,12 +155,13 @@ var vulnReportingDefaultRoles = map[string]roleAttributes{
...
 		resourceWithAccess: []permissions.ResourceWithAccess{
-			// TODO: ROX-13888 Replace VulnerabilityReports with WorkflowAdministration.
-			permissions.View(resources.VulnerabilityReports),   // required for vuln report configurations
-			permissions.Modify(resources.VulnerabilityReports), // required for vuln report configurations
-			permissions.View(resources.Role),                   // required for scopes
-			permissions.View(resources.Image),                  // required to gather CVE data for the report
-			permissions.View(resources.Integration),            // required for vuln report configurations
+			permissions.View(resources.WorkflowAdministration),   // required for vuln report configurations
+			permissions.Modify(resources.WorkflowAdministration), // required for vuln report configurations
+			permissions.View(resources.Role),                     // required for scopes
```
md2119, line 160: We do not need `Role` permissions are required anymore.

migrator/migrations/m_169_to_m_170_collections_sac_resource_migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/stackrox/rox/pkg/sac"
+)
+
+const (
+	batchSize = 500
+
+	workflowAdminResource       = "WorkflowAdministration"
+	reportConfigurationResource = "ReportConfiguration"
+
+	startSeqNum = 169
```
md2119, line 21:
```suggestion
	startSeqNum = migrations.CurrentDBVersionSeqNum()
```

md2119:
We need to include deprecation notice for `VulnerabilityReports` sac resource in this release. @keyallis will you be doing it or someone from auth team is looking into it?

central/role/datastore/singleton.go
keyallis: As of right now, `Role` is still required in the service layer implementation for report configurations. Is the migration from `ReportConfiguration` to `WorkflowAdministration` something we intend to do with MVP or is that a post-MVP task?

keyallis:
@md2119 The deprecation notice was included in #4108 and there is a follow up jira tracker for the replacement currently on the auth team

migrator/migrations/m_169_to_m_170_collections_sac_resource_migration/migration.go
keyallis: That function does not exist, and that's not the form of the previous (168 to 169) migration

central/role/datastore/singleton.go
md2119: With collections we do not require `Role` because `Role` was required by `AccessScopes` which is being replaced.

migrator/migrations/m_169_to_m_170_collections_sac_resource_migration/migration.go
md2119: It exists in the `seq_num`, but the real reason we cannot follow the above suggestion is because sequence number changes with each migration.

md2119:
You will need to modify the service layer of allow access if user has `permissions.View(resources.VulnerabilityReports) OR permissions.View(resources.WorkflowAdministration)`

TAL the following for an example:
https://github.com/stackrox/stackrox/blob/d8199792f1f777293b7914c91c3725b1ccae23cc/central/vulnerabilityrequest/service/service_impl.go

central/role/datastore/singleton.go
keyallis: If we cut a release without collections being enabled `Role` will still be required for the flow to work. It seems to me like it would be a good idea to leave it in there until it's fully deprecated, unless there's some significant reason why it should be removed

dashrews78 (approved):
Migration looks good.

central/role/datastore/singleton.go
md2119: Those perms need to be added conditionally based on the feature flag.

keyallis: Couldn't accomplish this in place because go doesn't like ternary. I added some filtering code before the object is used and will verify with the image that the values are excluded before merging

central/role/datastore/singleton.go
```diff
@@ -149,18 +168,20 @@ var defaultRoles = map[string]roleAttributes{
 	},
 }
 
+// TODO ROX-13888 when we migrate to WorkflowAdministration we can remove VulnerabilityReports and Role resources
 var vulnReportingDefaultRoles = map[string]roleAttributes{
```
md2119, line 172: You can drop this with above code suggestion

md2119 (commented):
Make sure the changelog captures the change in required permissions for configuring vulnerability reports. (Maybe a separate PR).

central/role/datastore/singleton.go
```diff
@@ -149,19 +150,27 @@ var defaultRoles = map[string]roleAttributes{
...
```
md2119, line 154: not your code, but I don't understand the point of var. 
Take it or ignore it: add this k-v directly to `defaultRoles`.

md2119 (approved):

md2119 (changes requested):
I believe this should be a `n` type migration.

md2119:
@c-du can you confirm if all the postgres migrations should be `n` type or `m` type?

keyallis:
@md2119 From the migrations dir readme:
```
After 3.73, the key-value stores are deprecated and Postgres becomes the only data store.

The migration returns to the old scheme, restarting from the database version after all data moves to Postgres.

All migrations again have the form m_{currentDBVersion}_to_m_{currentDBVersion+1}_{summary_of_migrations} .
```

md2119:
> @md2119 From the migrations dir readme:
> 
> ```
> After 3.73, the key-value stores are deprecated and Postgres becomes the only data store.
> 
> The migration returns to the old scheme, restarting from the database version after all data moves to Postgres.
> 
> All migrations again have the form m_{currentDBVersion}_to_m_{currentDBVersion+1}_{summary_of_migrations} .
> ```

I am aware of it. However, I want @c-du  to confirm it.

md2119 (approved):
@dashrews78 has confirmed that the `m` migration is the way to go.

PR #4115, opened 2022-12-13
Add refresh button to collection results panel

dvail:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -248,8 +254,20 @@ function CollectionResults({
 
     return (
         <>
+            <div className="pf-u-p-lg pf-u-display-flex pf-u-align-items-center">
```
dvail, line 257: Do we have any team preferences for when to use the PF `<Flex>` component versus `<div className="pf-u-display-flex">`? I know the former is required if we need to change the flex properties based on screen size, but it can be fairly verbose for simple flex cases like this.

ui/apps/platform/src/Containers/Collections/CollectionResults.tsx
```diff
@@ -248,8 +254,20 @@ function CollectionResults({
...
+                <div className="pf-u-display-flex pf-u-flex-direction-column pf-u-flex-grow-1">
+                    <Title headingLevel="h2">Collection results</Title>
+                    <Text>See a preview of current matches.</Text>
+                </div>
+                <Button variant="plain" onClick={refreshResults} title="Refresh results">
```
pedrottimark, line 262: What do you think about `isDisabled={isRefreshingResults}` prop?

pedrottimark (approved):

PR #4126, opened 2022-12-13
Run style action on PRs, tags and master

janisz:
## Description

Currently we are running style action only on PRs. This is suboptimal as cache is not populated by master branch. This change enables `style` on all PRs and pushes to master

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -1,11 +1,9 @@
 name: Style
 on:
+  push:
+    branches:
+      - master
   pull_request:
-    types:
-    - labeled
-    - opened
-    - reopened
-    - synchronize
 
 env:
   ROX_PRODUCT_BRANDING: RHACS_BRANDING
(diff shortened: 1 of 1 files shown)


janisz:
I created a PR to show how it will look like https://github.com/stackrox/stackrox/pull/4131
I did not focused on UX of this change but if tools supports sarif then we could use it (e.g. golangci-lint provides an action that comment directly on PR https://github.com/golangci/golangci-lint-action#golangci-lint-action).
So far I tried to keep it simple and use `make` instead of actions but maybe we switch to actions over our scripts ;) 

This PR is mostly about populating cache after every merge so `make style` will take ~20min instead of ~60.

Refs:
https://www.npmjs.com/package/@microsoft/eslint-formatter-sarif

gavin-stackrox (approved):
Got it.

janisz:
Here is the example failure https://github.com/stackrox/stackrox/pull/4131
For now it will stop when first `check` fails. I created couple of issues but only golangci was reported.

PR #4127, opened 2022-12-13
ROX-13912: Migrate UI unit tests

janisz:
## Description

Migrate UI unit tests to GHA.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,36 @@
+name: Tests
+on:
+  push:
+    tags:
+    - '*'
+    branches:
+    - master
+    - release-*
+  pull_request:
+    types:
+    - opened
+    - reopened
+    - synchronize
+jobs:
+  ui:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.49
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Create combined mono repo lock file
+      uses: ./.github/actions/create-concatenated-ui-monorepo-lock
+
+    - name: Cache UI dependencies
+      uses: ./.github/actions/cache-ui-dependencies
+
+    - name: UI Unit Tests
+      run: make ui-test
+
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v3
+      if: always()
+      with:
+        report_paths: 'ui/test-results/reports/*.xml'


janisz:
* master
  * **PR #4127** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4127?utm_source=stack-comment).

fredrb (approved):

central/pruning/pruning_test.go
```diff
@@ -614,7 +615,8 @@ func (s *PruningTestSuite) TestClusterPruning() {
 		s.T().Skip("Skipping because ROX_DECOMMISSIONED_CLUSTER_RETENTION feature flag isn't set.")
 	}
 
-	s.T().Setenv("ROX_IMAGE_FLAVOR", "rhacs")
+	s.T().Setenv(defaults.ImageFlavorNameStackRoxIORelease, defaults.ImageFlavorNameRHACSRelease)
```
msugakov, line 618: This has no effect and should be removed
```suggestion
```

central/pruning/pruning_test.go
```diff
@@ -847,7 +849,7 @@ func (s *PruningTestSuite) TestClusterPruningCentralCheck() {
...
```
msugakov, line 852: Do you mean this?
```suggestion
	s.T().Setenv(defaults.ImageFlavorEnvName, defaults.ImageFlavorNameRHACSRelease)
```

central/scanner/handler_test.go
```diff
@@ -34,6 +35,8 @@ func (s *handlerTestSuite) SetupTest() {
 }
 
 func (s *handlerTestSuite) TestGenerateScannerHTTPHandler() {
+	s.T().Setenv(defaults.ImageFlavorNameStackRoxIORelease, defaults.ImageFlavorNameDevelopmentBuild)
```
msugakov, line 38: Please remove
```suggestion
```

msugakov (changes requested):

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,88 @@
...
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
```
msugakov, line 57: I'm not sure such `ref` will work for non-pull_request events. Maybe just omit it?
[Documentation](https://github.com/actions/checkout) says:

```yaml
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+    - name: Cache Go dependencies
+      uses: ./.github/actions/cache-go-dependencies
+
+    - name: Go Unit Tests
+      run: ${{ matrix.gotags }} make go-unit-tests
```
msugakov, line 30: I understand what you do is better as you're making tests autonomous and runnable from IDE without having to set `ROX_IMAGE_FLAVOR` environment variable.

However, if this becomes too much of a burden, you can cut that corner and set
```
ROX_IMAGE_FLAVOR=development_build
```
whenever you run `make go-unit-tests`.

The former logic
https://github.com/stackrox/stackrox/blob/6a423ab0693b0cbe4f4208e8d265de3983d832df/scripts/ci/jobs/go-unit-tests.sh#L13-L17
is not relevant because `ROX_IMAGE_FLAVOR=stackrox.io` has no effect any more.

.github/workflows/unit-tests.yaml
janisz: https://github.com/stackrox/stackrox/pull/4213

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,88 @@
...
+      matrix:
+        gotags: [ 'GOTAGS=""', 'GOTAGS=release' ]
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.50
+    services:
+      # Label used to access the service container
+      postgres:
+        # Docker Hub image
+        image: postgres
```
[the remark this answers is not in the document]
janisz: Good point! How about using image from ghcr https://github.com/EnterpriseDB/docker-postgresql/pkgs/container/postgresql

.github/workflows/unit-tests.yaml
janisz: It looks like there is a valid usecase for it.
Fetch 0 is used to download whole history so `make tag` will work and `ref` is used to build against commit instead of a merge https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

msugakov: Found answer to the Q: will `ref: ${{ github.event.pull_request.head.sha }}` work on non-pull requests?
A: from [here](https://docs.github.com/en/actions/learn-github-actions/contexts#about-contexts)

> If you attempt to dereference a non-existent property, it will evaluate to an empty string.

.github/workflows/unit-tests.yaml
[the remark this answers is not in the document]
msugakov: That makes for another kind of dependency with another kind of problems. I think the safest choice is to stay with Red Hat registries whose failure scenarios we know better.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  ui:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.49
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - uses: ./.github/actions/create-concatenated-ui-monorepo-lock
```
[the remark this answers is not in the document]
janisz: Technically name is not required and in other places we also do not give it a name although it will looks better. I'll fix it

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+      uses: actions/checkout@v3
+
+    - name: Create combined mono repo lock file
+      uses: ./.github/actions/create-concatenated-ui-monorepo-lock
+
+    - name: Cache UI dependencies
+      uses: ./.github/actions/cache-ui-dependencies
+
+    - name: Fetch UI deps
+      run: make -C ui deps
```
msugakov, line 29: You could drop this because `make -C ui test` depends on `deps`. Optional and up to you.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: UI Unit Tests
+      run: make ui-test
+
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v3
```
msugakov, line 35: Stupid question. I see that this action runs and reports how many tests are successful. But where do I see the test results? E.g. test names, their status (passed/failed) and any stdout+stderr from them (that's captured in junit output).

msugakov (changes requested):

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    tags:
+    - '*'
+    branches:
+    - master
+    - release-*
+  pull_request:
+    types:
+    - opened
+    - synchronize
```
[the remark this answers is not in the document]
janisz: Originally I left this empty but then it was not starting on stacked PRs so I looked how we do it in other places and copied it https://github.com/stackrox/stackrox/blob/2995fffcebbfd186d1284fedcdd1895e93eee1ed/.github/workflows/build.yaml#L3-L6

.github/workflows/unit-tests.yaml
msugakov: This looks extremely bare-bones and pretty much on the border of being completely unusable.

Would you be so kind to try find some GitHub Action or a different tool that displays results in more actionable way, as a backlog/follow-up item, or at least create a follow-up JIRA?

msugakov (approved):
Green from my side

PR #4132, opened 2022-12-13
Fix vuln test

janisz:
## Description

Looks like there is no property `tewakesortedVulnsdArray`.
This PR fixes this and also add types and use Set instead of ordered list.

### DIFF
--- a/qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
+++ b/qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
@@ -157,9 +157,9 @@ class VulnMgmtSACTest extends BaseSpecification {
     }
 
     // GraphQL does not provide ordering guarantees, so to compare the results
-    // of two GraphQL queries we extract just the CVE names and sort them.
-    def extractCVEsAndSort(queryCallResult) {
-        return queryCallResult.results*.cve.sort()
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/VulnMgmtSACTest.groovy
```diff
@@ -240,9 +240,9 @@ class VulnMgmtSACTest extends BaseSpecification {
     }
 
     // GraphQL does not provide ordering guarantees, so to compare the results
-    // of two GraphQL queries we extract just the CVE names and sort them.
-    def extractCVEsAndSort(queryCallResult) {
-        return queryCallResult.results*.cve.sort()
+    // of two GraphQL queries we extract just the CVE names as a Set.
+    Set<String> extractCVEs(queryCallResult) {
+        return (queryCallResult.results*.cve.toSet() + UUID.randomUUID().toString()) as Set<String>
```
rhybrillou, line 245: I suppose the extra UUID.randomUUID().toString() is there in order to have a failing test and validate the error reporting code.

janisz: Correct, I'll revert it once I see it's working correctly

janisz:
@rhybrillou Output is now cleaner and missing property is fixed.

rhybrillou (approved):

PR #4134, opened 2022-12-13
Skip TLS verification during e2e tests

BradLugo:
## Description

During end-to-end tests, `roxctl` throws an error about the validity of the certification when connecting to the newly made cluster:
```
ERROR:	Certificate validation error: x509: certificate signed by unknown authority
```

OpenShift CI sees this in the build log and will highlight it in the UI, which causes confusion when someone's triaging a broken build. This PR removes TLS verification, which is safe since we created the cluster, and will remove the above error from the build log.

## Testing Performed

In progress


BradLugo:
OBE https://github.com/stackrox/stackrox/pull/4130

PR #4136, opened 2022-12-13
ROX-13729: Add Central DB values to the public and private examples

c-du:
## Description

Postgres will be GA in 74 and hence we need to add Central DB specific values to the public and private examples
so that the customer can start using it.

In addition, there are a few value configs are not rendered. I filed two jiras

ROX-13920 | Central db node selector config is not rendered
ROX-13921 | Central DB config value db.extraMounts is not rendered.

It looks like ROX-13920 should be needed for hostpath pvc so I put it as 74 blocker. It looks to me that there is no need for ROX-13921 as of now, correct me if I am wrong.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


msugakov:
Not sure I'm the best reviewer here (also in terms of response time currently). Added @stackrox/draco team so that they have a look.

vladbologa (approved):

PR #4178, opened 2022-12-16
Fixed reinit issue w topology controller

alwayshooin:
as title states -- the new network graph should no longer reinitialize when selecting/unselecting a node, nor when switching between active and extraneous states. everything else (edge rendering, extraneous nodes, etc) should remain the same.


sachaudh (approved):
Nice, looks good to me!

PR #4181, opened 2022-12-16
ROX-13887: Revert rules to previously applied YAML

sachaudh:
## Description

1. Updating the `NetworkPolicySimulatorSidePanel` to show the previously applied YAML when we are in the `UNDO` state of the network policy simulator.

Screenshot 2022-12-16 at 7 21 11 AM
Screenshot 2022-12-16 at 7 21 18 AM

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/simulation/NetworkPoliciesYAML.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/simulation/NetworkPoliciesYAML.tsx
@@ -1,21 +1,31 @@
 import React from 'react';
 import { CodeEditor, CodeEditorControl, Language } from '@patternfly/react-code-editor';
-import { DownloadIcon, MoonIcon, ProcessAutomationIcon, SunIcon } from '@patternfly/react-icons';
-import { Bullseye, EmptyState, EmptyStateVariant, Title } from '@patternfly/react-core';
+import {
+    DownloadIcon,
+    MoonIcon,
+    ProcessAutomationIcon,
+    SunIcon,
+    UndoIcon,
+} from '@patternfly/react-icons';
 
 import { useTheme } from 'Containers/ThemeProvider';
 import download from 'utils/download';
 
 type NetworkPoliciesYAMLProp = {
     yaml: string;
-    generateNetworkPolicies?: () => void;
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/simulation/NetworkPoliciesYAML.tsx
```diff
@@ -53,17 +63,15 @@ function NetworkPoliciesYAML({ yaml, generateNetworkPolicies }: NetworkPoliciesY
...
-                    </Title>
-                </EmptyState>
-            </Bullseye>
-        );
-    }
+    const revertRecentYAML = undoNetworkPolicies ? (
+        <CodeEditorControl
+            icon={<UndoIcon />}
+            aria-label="Revert most recently applied YAML"
+            toolTipText="Revert most recently applied YAML"
```
vjwilson, line 70: nit: If these are the same, perhaps we should define a string constant and reuse it in both props.

ui/apps/platform/src/Containers/NetworkGraph/simulation/NetworkPoliciesYAML.tsx
```diff
@@ -53,17 +63,15 @@ function NetworkPoliciesYAML({ yaml, generateNetworkPolicies }: NetworkPoliciesY
...
-            <Bullseye>
-                <EmptyState variant={EmptyStateVariant.xs}>
-                    <Title headingLevel="h4" size="md">
-                        No network policies
-                    </Title>
-                </EmptyState>
-            </Bullseye>
-        );
-    }
+    const revertRecentYAML = undoNetworkPolicies ? (
```
vjwilson, line 66: Why is this optional?

vjwilson (approved):
Just a couple of minor questions.

ui/apps/platform/src/Containers/NetworkGraph/simulation/NetworkPoliciesYAML.tsx
sachaudh: Hmm, actually that's a good question. I think the controls should be visible regardless and not hidden when viewing active YAMLs. I'll change that

PR #4182, opened 2022-12-16
ROX-13904: Excent plop lifecycle with closed status [Do not merge, used for Pipeline testing]

erthalion:
## Description

Derive "closed" status for ProcessListeningOnPort and modify the lifecycle accordingly. Closed PLOP objects are going to be ignored from the API.

## Checklist

## Testing Performed

Local unit tests run.


central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -105,6 +107,40 @@ func (ds *datastoreImpl) AddProcessListeningOnPort(
...
+
+	if len(indicatorIds) > 0 {
+		// If no corresponding processes found, we can't verify if the PLOP
+		// object is closing an existing one. Collect existingPLOPMap only if
+		// there are some matching indicators.
+
+		// XXX: This has to be done in a single join query fetching both
+		// ProcessIndicator and needed bits from PLOP.
+		existingPLOP, err := ds.storage.GetByQuery(ctx, search.NewQueryBuilder().
+			AddStrings(search.ProcessID, indicatorIds...).
```
JoukoVirtanen, line 122: Above we are using DisjunctionQuery. Here we are using an array of indicatorIds. We could use DisjuctionQuery here or we could use an array of indicatorIds above were we use DisjunctionQuery. What are the advantages and disadvantages of the methods? Why use these two methods? I don't know that one is better than the other.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -127,15 +163,40 @@ func (ds *datastoreImpl) AddProcessListeningOnPort(
...
+
+				activePLOP.Closed = true
+				plopObjects[i] = activePLOP
+			} else {
+				log.Warnf("Found no matching PLOP to close for %s", key)
+				plopObjects[i] = newPLOP
+			}
+		} else {
+			plopObjects[i] = newPLOP
+		}
```
JoukoVirtanen, line 199: This is a very long function. Maybe break it into smaller pieces.

proto/storage/process_listening_on_port.proto
```diff
@@ -50,4 +50,7 @@ message ProcessListeningOnPortStorage {
 
     string process_indicator_id = 5
         [(gogoproto.moretags) = 'search:"Process ID,store" sql:"fk(ProcessIndicator:id),no-fk-constraint,index=btree,type(uuid)"'];
+
+    // XXX: Make it a partial index on only active, not closed, PLOP
+    bool closed = 6 [(gogoproto.moretags) = 'search:"Closed,store"  sql:"index=btree"'];
```
JoukoVirtanen, line 55: I don't think you need this additional field. The close_timestamp already keeps track of if the port is open or closed. If close_timestamp is nil the port is open. If close_timestamp is not nil the port is closed.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -127,15 +163,40 @@ func (ds *datastoreImpl) AddProcessListeningOnPort(
...
 	}
 
 	// Now save actual PLOP objects
```
JoukoVirtanen, line 202: I think this is overly complicated. I don't think you needed to make so many changes to the existing code that you had. I don't think this Closed field is needed. I commented on this below in the protobuf definition file, but for convenience I will include my comment there, here. 

"I don't think you need this additional field [closed]. The close_timestamp already keeps track of if the port is open or closed. If close_timestamp is nil the port is open. If close_timestamp is not nil the port is closed."

It seems like all of this code is here just to set the Closed field and that field is not needed.I think probably all that needs to be done is to modify the query in store.go so that it only returns entries where close_timestamp is nil. We are using upsert so we should be able to insert new plops without worrying if they already exist.

central/processlisteningonport/datastore/datastore_impl_test.go
```diff
@@ -254,11 +467,24 @@ func (suite *PLOPDataStoreTestSuite) TestPLOPAddMultipleIndicators() {
...
+		Endpoint: &storage.ProcessListeningOnPort_Endpoint{
+			Port:     1234,
+			Protocol: storage.L4Protocol_L4_PROTOCOL_TCP,
+		},
+		Signal: &storage.ProcessSignal{
+			Name:         "test_process1",
+			Args:         "test_arguments1",
+			ExecFilePath: "test_path1",
+		},
+	})
```
JoukoVirtanen, line 489: Unit tests look good. Could you try writing datastore_impl.go without the closed field and without getting existing plops and see if the unit tests pass?

central/processlisteningonport/store/postgres/store.go
```diff
@@ -57,8 +57,7 @@ const (
 		"FROM process_listening_on_ports plop " +
 		"JOIN process_indicators proc " +
 		"ON plop.processindicatorid = proc.id " +
-		"WHERE proc.deploymentid = $1"
-
+		"WHERE proc.deploymentid = $1 AND plop.closed = false"
```
JoukoVirtanen, line 60: Try checking if close_timestamp is null instead.

proto/storage/process_listening_on_port.proto
erthalion: `close_timestamp` is part of serialized field and is not extracted into a separate column to operate via sql. I was thinking about this option, but decided to use a boolean value instead because we have to create an index on it, and in this way it would be smaller than an index over a timestamp field (especially when partial index support will be introduced into the generator).

central/processlisteningonport/datastore/datastore_impl.go
erthalion: > It seems like all of this code is here just to set the Closed field and that field is not needed.I think probably all that needs to be done is to modify the query in store.go so that it only returns entries where close_timestamp is nil. We are using upsert so we should be able to insert new plops without worrying if they already exist.

Not quite, most of the code is busy updating closed_timestamp instead of creating another duplicating record of the PLOP object with different closed_timestamp. To achieve that there should be extra lookup (or ideally an extension of process indicator with the join, but it's more cumbersome with the current store generator) to fetch already existing potentially fitting PLOP objects (as neither Collector nor Sensor know about PLOP id) -- this is what constitutes most of the code changes.

JoukoVirtanen: Hmm. I might not be clear on how upsert works in this case. If we do an upsert with different close_timestamps, why do we get duplicates, if we don't get duplicates when upserting different closed values? I might need to play around with this. I have read about upsert in postgres, but I don't know exactly what the UpsertMany function does.

erthalion: > If we do an upsert with different close_timestamps, why do we get duplicates

The `UPSERT` in Postgres relies on a unique constraint to verify, if the constraint is failing -- the update will be performed, otherwise the insert. The storage generator uses an id unique constraint for this purpose, which means:

* If we call `upsert` on an object with `id` field set and there is already existing record with this value -- an actual upsert is happening
* If we call `upsert` on an object without `id` set, there will be no clash possible and it will be always inserted.

The data Central received from Sensor cannot contain a PLOP `id`, because only Central knows it. Thus  there should be fetching logic. Does it make things clearer to you?

JoukoVirtanen: Yes. I was thinking that the key used for the upsert would be process_indicator_id, but I see that it is not the primary key and id is the primary key. Could we use process_indicator_id as the primary key?

erthalion: > Could we use process_indicator_id as the primary key?

Not really, this will break cases when one process listens to multiple ports.

JoukoVirtanen: You can also specify a column to be unique. E.g. in storage/cluster.proto there is the following 

message Cluster {
    string                    id                               = 1 [(gogoproto.moretags) = 'search:"Cluster ID,hidden,store" sql:"pk,type(uuid)"'];
    string                    name                             = 2 [(gogoproto.moretags) = 'search:"Cluster,store" sql:"unique"'];

Perhaps process_indicator_id could be set to be unique. Would that simplify things?

central/processindicator/datastore/datastore_impl.go
```diff
@@ -137,12 +142,28 @@ func (ds *datastoreImpl) removeMatchingIndicators(ctx context.Context, results [
...
 func (ds *datastoreImpl) removeIndicators(ctx context.Context, ids []string) error {
+	log.Infof("Process Pruning")
 	if len(ids) == 0 {
 		return nil
 	}
 	if err := ds.storage.DeleteMany(ctx, ids); err != nil {
 		return err
 	}
+
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
```
dashrews78, line 153: Keep an eye on this in scale and performance tests.  I tried some similar things with process indicators and network flows but in those scale environments I would get statement timeouts on these big deletes.  That is why DeleteMany batches.  It may be OK here, just something to keep in mind when testing.

central/processlisteningonport/datastore/datastore_impl.go
erthalion: > Perhaps process_indicator_id could be set to be unique. Would that simplify things?

As I've already said, no, this will break cases when one process listens to multiple ports. You have to keep in mind that the relation here is "one to many", i.e. one process to many ports.

central/processindicator/datastore/datastore_impl.go
erthalion: Good point, thanks. Generally speaking I believe it would be much better to rework pruning to do the cleanup by dropping old partitions, but it's quite a lot of work.

I've planned one round of perf evaluation for sql queries, so will check this out.

central/processlisteningonport/datastore/datastore_impl.go
dashrews78: I see both sides of this conversation.  This is a lot of logic to insert a record at the same time, based on how we set up the table it is necessary.  I do wonder, though, would it be feasible to change to a composite key and take advantage of that?  Say we eliminate the plop ID itself and got with a composite key on indicator_id, port, and protocol.  Something like:

`message ProcessListeningOnPortStorage {
    // Ideally it has to be GENERATED ALWAYS AS IDENTITY, which will make it a
    // bigint with a sequence. Unfortunately at the moment some bits of store
    // generator assume an id has to be a string.
    string id = 1 [(gogoproto.moretags) = 'sql:"pk,type(uuid)"'];

    uint32 port = 2  [(gogoproto.moretags) = 'sql:"pk"`;
    L4Protocol protocol = 3 [(gogoproto.moretags) = 'sql:"pk"`;

    string process_indicator_id = 5
        [(gogoproto.moretags) = 'search:"Process ID,store" sql:"pk,fk(ProcessIndicator:id),no-fk-constraint,index=btree,type(uuid)"'];

    // XXX: Make it a partial index on only active, not closed, PLOP
    bool closed = 6 [(gogoproto.moretags) = 'search:"Closed,store"  sql:"index=btree"'];
}`

I think the generators would support that and it may help simplify some of this logic.  That said, my background on PLOP is high level so there may be reasons that what I suggest is not feasible.

erthalion: It's an interesting idea, I haven't though about this one -- in this way it doesn't make sense to keep anything serialized anymore so we can unwrap it. There are of course limitations as well:

* In this case we loose possibility to store those PLOP objects what we were not able to correlate with any process information. There could be cases like this, so the idea was to still keep them to not loose the data and for introspection about why did it happen like this.
* The primary index is going to be rather large with certain amount of duplication as endpoint information would be often repeated. The proposal is essentially to trade one read query for an index over three columns. It would be interesting to see if PG14 posting lists could optimize this case.

The amount of changes required to implement this is comparable to the original approach, pruning and custom store parts have to be changed.

erthalion: > This is a lot of logic to insert a record

Btw, for both to keep in mind: the whole acrobatics around existing PLOP map is needed only because it's hard to fetch the indicator + plop info in one join. I've started doing this, but already tired of fighting with it's limitations and the custom version would be annoying as well because of batching. But as soon as joins will be normally supported by query builder this whole patch will become:

* fetch indicator and correlated plop if exists
* based on the presence of the latter fill up the structure to upsert with either existing id or with the new one.

erthalion: > all these options have drawbacks. IMO if the generators and query builder are holding you back, it might be time to tackle those updates that you need even though I know you would rather not do that as part of MVP.

It would be nice, but we've already spent much more than expected time on this, so not sure.

 > There could be cases like this, so the idea was to still keep them to not loose the data and for introspection about why did it happen like this.

That reminds me, I think we've forgotten to actually keep needed process information in this case in the serialized column :( I guess there should be a process info field, that will contain the signal details we've got from Sensor which were not found in the db, and null otherwise.

PR #4188, opened 2022-12-16
ROX-13929: Add log to debug VM GraphQL test error

theencee:
## Description

Unable to repro this locally and it has only happened twice, so print out the actual error from the response for next time.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/qa-tests-backend/src/test/groovy/VulnMgmtTest.groovy
+++ b/qa-tests-backend/src/test/groovy/VulnMgmtTest.groovy
@@ -256,7 +256,14 @@ fragment cveFields on ImageVulnerability {
 
         def embeddedImageRes = gqlService.Call(getEmbeddedImageQuery(),
                 [id: imageDigest, query: "CVE:CVE-2017-10684"])
-        assert embeddedImageRes.hasNoErrors()
+
+        // Expanded instead of using hasErrors() for easier debugging if there are errors
+        // as the test framework will actually print out the errors now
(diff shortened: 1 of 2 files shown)


keyallis (approved):

qa-tests-backend/src/test/groovy/VulnMgmtTest.groovy
```diff
@@ -256,7 +256,14 @@ fragment cveFields on ImageVulnerability {
 
         def embeddedImageRes = gqlService.Call(getEmbeddedImageQuery(),
                 [id: imageDigest, query: "CVE:CVE-2017-10684"])
-        assert embeddedImageRes.hasNoErrors()
+
+        // Expanded instead of using hasErrors() for easier debugging if there are errors
+        // as the test framework will actually print out the errors now
+        assert embeddedImageRes.code == 200
```
md2119, line 262: Is the concern only with the above graphQL call?

md2119 (approved):

qa-tests-backend/src/test/groovy/VulnMgmtTest.groovy
theencee: This is the one that has failed. Others might, but I'm hoping given this one has for sure we can get logs out of it.

theencee:
The EKS failure seems to be unrelated

PR #4195, opened 2022-12-19
Dashrews/ROX-13964 central db accessible flag in metadata

dashrews78:
## Description

Now that the database is separate from central we need to alert users if the database becomes unreachable.  For the purposes of the UI, the simplest mechanism for the backend to inform the UI is through the /metadata endpoint.  This simply adds a flag that will alert the caller of /metadata whether or not the database is reachable.  What to do with that information is up to the caller and will be addressed separately.

## Checklist

## Testing Performed

Added a unit tests to get the metadata.


vjwilson (approved):
Tested the response locally in UI, and that works.

c-du (approved):

PR #4197, opened 2022-12-19
Add usePaginatedQuery hook to abstract over infinite list queries

dvail:
## Description

This adds a hook that allows us to fetch data that will be presented in an "infinite list" type UI.

This is something that would likely be better of handled by a library like react-query or SWR, but integration with those libraries is likely a ways off in the future. On the other hand, since this pattern occurs 3 times in Collections already I thought it would be valuable to extract this logic instead of copy-pasting similar code in each component. I think we should have an end goal of deleting this hook and replacing with something standard down the line.

## Follow ups

- Convert existing CollectionResults logic to use this hook ([Draft] https://github.com/stackrox/stackrox/pull/4198)
- Use this hook for the Collection dropdown in the updated Vuln Mgmt form
- Use this hook for the Collection Attacher component (if possible)

## Checklist

## Testing Performed

Core functionality is tested via unit tests.

Manual functionality is tested in the implementation of this for the first two follow up items above. (PRs coming soon.)

### DIFF
--- a/ui/apps/platform/src/hooks/usePaginatedQuery.ts
+++ b/ui/apps/platform/src/hooks/usePaginatedQuery.ts
@@ -4,18 +4,18 @@ import { debounce } from 'lodash';
 function filterNextPage<Item, ItemKey>(
     results: Item[],
     itemKeys: Set<ItemKey>,
-    keyFn: ((item: Item) => ItemKey) | undefined
+    dedupKeyFn: ((item: Item) => ItemKey) | undefined
 ) {
     // If a deduplication function has been provided, use it to ensure only unique items
     // are added to the data results. Otherwise, return the results directly without the need
     // to update the key Set.
-    if (!keyFn) {
+    if (!dedupKeyFn) {
         return { nextPageItems: results, nextKeys: itemKeys };
     }
     const nextPageItems: Item[] = [];
     const nextKeys = new Set(itemKeys);
     results.forEach((item) => {
-        const itemKey = keyFn(item);
+        const itemKey = dedupKeyFn(item);
         if (!nextKeys.has(itemKey)) {
             nextKeys.add(itemKey);
             nextPageItems.push(item);
@@ -80,6 +80,9 @@ export function usePaginatedQuery<Item, ItemKey>(
         manualFetch?: boolean;
     } = {}
 ): UsePaginatedQueryReturn<Item> {
+    // These initialOptions are stored in state instead of being destructured directly since
+    // the `fetchPageHandler` function below relies on the reference of `dedupKeyFn`. By storing these
+    // in state, it removes the burden of wrapping in a useCallback or useMemo from the caller.
     const [{ debounceRate, dedupKeyFn, onError, manualFetch = false }] = useState(initialOptions);
     const [isRefreshingResults, setIsRefreshingResults] = useState(false);
     const [isFetchingNextPage, setIsFetchingNextPage] = useState(false);
--- a/ui/apps/platform/src/hooks/usePaginatedQuery.ts
+++ b/ui/apps/platform/src/hooks/usePaginatedQuery.ts
@@ -1,0 +1,187 @@
+import { useState, useCallback, useEffect, useMemo } from 'react';
+import { debounce } from 'lodash';
+
+function filterNextPage<Item, ItemKey>(
+    results: Item[],
+    itemKeys: Set<ItemKey>,
+    keyFn: ((item: Item) => ItemKey) | undefined
+) {
+    // If a deduplication function has been provided, use it to ensure only unique items
+    // are added to the data results. Otherwise, return the results directly without the need
+    // to update the key Set.
+    if (!keyFn) {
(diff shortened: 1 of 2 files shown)


dvail:
* master
  * **PR #4197** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4197" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4197?utm_source=stack-comment).

ui/apps/platform/src/hooks/usePaginatedQuery.ts
```diff
@@ -0,0 +1,187 @@
...
+        /** Callback function to fire if a request results in an error */
+        onError?: (err: Error) => void;
+        /**
+         *  Whether or not the initial fetch call for the first page should be automatic or require
+         * a manual call to `fetchNextPage`
+         */
+        manualFetch?: boolean;
+    } = {}
+): UsePaginatedQueryReturn<Item> {
+    const [{ debounceRate, dedupKeyFn, onError, manualFetch = false }] = useState(initialOptions);
```
pedrottimark, line 83: Because there is no set function from `useState` hook, is it possible to destructure directly?

```js
const { debounceRate, dedupKeyFn, onError, manualFetch = false } = initialOptions;
```

ui/apps/platform/src/hooks/usePaginatedQuery.ts
```diff
@@ -0,0 +1,187 @@
+import { useState, useCallback, useEffect, useMemo } from 'react';
+import { debounce } from 'lodash';
+
+function filterNextPage<Item, ItemKey>(
+    results: Item[],
+    itemKeys: Set<ItemKey>,
+    keyFn: ((item: Item) => ItemKey) | undefined
```
pedrottimark, line 7: What do you think about `dedupKeyFn`
* Be consistent with function call.
* Let the name provide some of the meaning and let comment reinforce it.

ui/apps/platform/src/hooks/usePaginatedQuery.ts
```diff
@@ -0,0 +1,187 @@
...
+                return pageFetcher.immediate(queryFn, page, itemKeys);
+            }
+            return pageFetcher.debounced(queryFn, page, itemKeys);
+        },
+        [pageFetcher, queryFn, page, itemKeys]
+    );
+
+    const clearPages = useCallback(() => {
+        setItemPages([]);
+        setItemKeys(new Set());
```
pedrottimark, line 156: For my understanding, does any state update depend on new set which changes referential identity?

pedrottimark (approved):
Thank you for writing tests. 2 comments and 1 question.

pedrottimark:
Dave, a question that I forgot to ask in comments is whether `usePaginatedQuery` needs to be more specific to indicate its use (pardon pun) for infinite lists? Or is it as useful for ordinary paginated tables?

ui/apps/platform/src/hooks/usePaginatedQuery.ts
dvail: Not with the hook as-is. The `fetchPageHandler` is wrapped in a `useCallback` that has `dedupKeyFn` in the dependency array. I initially did try to destructure directly but it led to a loop since the `dedupKeyFn` was a new instance each time it was called from the parent.

I figured instead of making calling components responsible for the memoization of the `dedupKeyFn` I would have the options be non-reactive and store them in local state (hence the _initial_ part of the parameter name). not as intuitive as I would like, but I thought it was better keeping the complexity inside and not forcing callers to `useCallback` themselves.

I can add a comment here explaining why.

ui/apps/platform/src/hooks/usePaginatedQuery.ts
dvail: +1

I think this function was initially inside the hook and I used this naming to avoid shadowing.

ui/apps/platform/src/hooks/usePaginatedQuery.ts
dvail: Yes, only the `fetchNextPage` function that is wrapped in a `useCallback`. I captures the current value of `itemKeys` in a closure and needs to be replaced each time that value reference changes.

dvail:
@pedrottimark This _could_ be used for paginated tables, but I'm not sure that it would be any more useful than a non-infinite alternative. (Unless we needed to dedup the pages for some reason).

I can change the name to `useInfiniteQuery` to be even closer to the react-query version if you think that is a good idea.

ui/apps/platform/src/hooks/usePaginatedQuery.ts
pedrottimark: Yes, a comment sounds good.

pedrottimark:
Unless the hook API is the same (or very similar) then same name has some risk of misunderstanding.

Better to leave well enough alone. The comment explains the intended use.

PR #4199, opened 2022-12-19
ROX-13729: Unhide central db generate and add CLI unit tests

c-du:
## Description
Expose central db generate CLI for version 74.
In addition to the functional test with bats, we want to add light weight unit tests to cover the cli command lines and the error messages.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dashrews78 (approved):
LGTM

roxctl/central/db/generate/generate_test.go
```diff
@@ -0,0 +1,406 @@
...
+				words := strings.Fields(lines[i])
+				s.Assert().Greater(len(words), 1)
+				flags[words[0]] = words[1:]
+			}
+		}
+	}
+	return
+}
+
+func (s *centralDBGenerateCliTestSuite) verifyConfig(c testCaseType, config *renderer.Config) {
```
janisz, line 318: This function has a lot of logic. I think it should be clean assertion `Equal(actual, expected *renderer.Config)` and `expected` config should be part of `testCaseType`

roxctl/central/db/generate/generate_test.go
```diff
@@ -0,0 +1,406 @@
...
+	for _, cmd := range cmds {
+		if cmd.Use == target {
+			return cmd
+		}
+	}
+	s.Require().Fail("Cannot find command %s", target)
+	return nil
+}
+
+func (s *centralDBGenerateCliTestSuite) parseUsage(output string) (usage string, commands map[string][]string, flags map[string][]string) {
```
janisz, line 286: Why do we need to parse usage? Why can't we just compare output with expected output? This will simplify the test and make it clear what if failing

roxctl/central/db/generate/generate_test.go
```diff
@@ -0,0 +1,406 @@
...
+	s.setRunE(testCase, genCmd, "k8s", "none")
+	s.setRunE(testCase, genCmd, "openshift", "pvc")
+	s.setRunE(testCase, genCmd, "openshift", "hostpath")
+	s.setRunE(testCase, genCmd, "openshift", "none")
+
+	return genCmd
+}
+
+func (s *centralDBGenerateCliTestSuite) setRunE(testCase testCaseType, root *cobra.Command, subCmds ...string) {
+	command := root
```
janisz, line 260: Why do we need this? If `command` is better name then rename argument

PR #4198, opened 2022-12-19
Update collection results to use `usePaginatedQuery` hook

dvail:
## Description

Small refactor to use hook to handle infinite loading queries from https://github.com/stackrox/stackrox/pull/4197.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Repeat manual testing steps from:
https://github.com/stackrox/stackrox/pull/4001
https://github.com/stackrox/stackrox/pull/4110
https://github.com/stackrox/stackrox/pull/4115

PR #4200, opened 2022-12-19
Update Vuln Reporting UI to use collections instead of resource scopes

dvail:
## Description

Switches over the VM reporting section to use Collections instead of Access Scopes when the `ROX_OBJECT_COLLECTIONS` feature flag is enabled.

**Note that this change comes with the following caveats**

1. The back end being used is a build from the (currently un-merged) branch that implements the BE functionality: https://github.com/stackrox/stackrox/pull/4100 
2. A database migration _has not_ occurred to handle existing VM report configurations that we saved using access scopes as the scoping mechanism. Existing VM reports will not work against this branch, we can only test the creation, edit, and view behavior of VM reports created against the branch in PR 4100 above.

From a reviewing perspective, there are three main sections that are changed:

1. [The Cypress tests are updated](https://github.com/stackrox/stackrox/pull/4200/files#diff-9039ca8a276bc8943a3a87f7637e11ad94b8545871e8f291a7be720a8b1fcec3) (everything changed under `/cypress/`) to wait on the collection request instead of the access scope request. I don't believe the tests care about the data that is returned at this point, just that the request succeeds. 
2. [The Collection modal is updated](https://github.com/stackrox/stackrox/pull/4200/files#diff-1bb9bd1a8a08241dc7d2dab94ee53244c04b2f36c537b063ad9aed419de857e1) (everything changed under `/src/Containers/Collections/`) to allow creation of new collections. Previously the modal was restricted to the read-only use case.
3. [The VM reporting section](https://github.com/stackrox/stackrox/pull/4200/files#diff-9e00882975b1b1629a0e9161a272ec236578f12b20109709f988ecaaa230bb92) (everything under `/src/Containers/VulnMgmt/`) is updated to use collections and the collection modal.

## Follow ups

- The layout of the collection dropdown combined with the floating bottom bar is a little hard to use. We should evaluate if there is a good way to avoid the back and forth scrolling that needs to occur when browsing the list.

## Checklist

## Testing Performed

Visit the VM reporting page and click the creation button. The prior access scope dropdown has been replaced with a collections dropdown.
image

Opening the collections dropdown will reveal the first 10 collection results, the rest accessible via a "View more" lazy loading component.
image

Clicking the view more button will load more collections into the list.
image

Selecting an item in the list makes it the selected report scope for the VM report.
image

Opening the dropdown and entering text into the typeahead box will filter the dropdown to only include collections that have a name starting with the entered text. If there are more than 10 matching collections, the view more button will allow more to be loaded.
image

Clicking the "Create collection" button from the form will open a collection modal and allow creation of a new collection. Results previews, filtering, and error handling will have the full functionality as if the collection is being created directly via the collection page.
image
image
image

Saving the collection will create it in the database, populate the collection dropdown, and set the scope ID for the report form.
image

Opening the dropdown and clicking 'view more' until the server results are exhausted will not cause a duplicate entry in the dropdown. (The newly creation collection has not yet been stored in the pagination results, so it is possible that it will be returned by the server when viewing more collections.)
image

Filling out the rest of the form and submitting it correctly saves the VM report config as well as the referenced collection.
image
image
Clicking on the report scope link in the details view correctly navigates the user to the collection's page.
image

### DIFF
--- a/ui/apps/platform/cypress/fixtures/collections/collections.json
+++ b/ui/apps/platform/cypress/fixtures/collections/collections.json
@@ -1,0 +1,20 @@
+{
+    "collections": [
+        {
+            "id": "34e57a20-c630-4767-baa9-0cb26c1274b9",
+            "name": "Security Cluster Collection (CYPRESS)",
+            "description": "Test collection for Cypress e2e tests",
+            "resourceSelectors": [
+                {
+                    "rules": [
+                        {
+                            "operator": "OR",
+                            "fieldName": "Cluster",
+                            "values": [{ "value": "security" }]
+                        }
+                    ]
+                }
+            ]
+        }
+    ]
+}
--- a/ui/apps/platform/cypress/integration/vulnmanagement/reporting/reporting.helpers.js
+++ b/ui/apps/platform/cypress/integration/vulnmanagement/reporting/reporting.helpers.js
@@ -97,8 +97,10 @@ export function visitVulnerabilityReportingWithFixture(fixturePath) {
 // action create
 
 export const accessScopesAlias = 'simpleaccessscopes';
+export const collectionsAlias = 'collections';
 export const notifiersAlias = 'notifiers';
 
+// TODO This object can be deleted once the ROX_OBJECT_COLLECTIONS feature flag is removed
 const routeMatcherMapToCreate = {
     [accessScopesAlias]: {
         method: 'GET',
@@ -110,10 +112,31 @@ const routeMatcherMapToCreate = {
     },
 };
 
-export function visitVulnerabilityReportingToCreate(staticResponseMap) {
-    visit(`${basePath}?action=create`, routeMatcherMapToCreate, staticResponseMap);
+const routeMatcherMapToCreateWithCollections = {
+    [collectionsAlias]: {
+        method: 'GET',
+        url: '/v1/collections*',
+    },
+    [notifiersAlias]: {
+        method: 'GET',
+        url: '/v1/notifiers',
+    },
+};
+
+export function visitVulnerabilityReportingToCreate(staticResponseMap, isCollectionsEnabled) {
+    const routeMatcherMap = isCollectionsEnabled
+        ? routeMatcherMapToCreateWithCollections
+        : routeMatcherMapToCreate;
+    visit(`${basePath}?action=create`, routeMatcherMap, staticResponseMap);
 }
 
-export function interactAndWaitToCreateReport(interactionCallback, staticResponseMap) {
-    interactAndWaitForResponses(interactionCallback, routeMatcherMapToCreate, staticResponseMap);
+export function interactAndWaitToCreateReport(
+    interactionCallback,
+    staticResponseMap,
+    isCollectionsEnabled
+) {
+    const routeMatcherMap = isCollectionsEnabled
+        ? routeMatcherMapToCreateWithCollections
+        : routeMatcherMapToCreate;
+    interactAndWaitForResponses(interactionCallback, routeMatcherMap, staticResponseMap);
 }
--- a/ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import {
+    Alert,
     Bullseye,
     Button,
     Divider,
@@ -19,11 +20,18 @@ import { Collection } from 'services/CollectionsService';
 import useCollection from './hooks/useCollection';
 import CollectionFormDrawer, { CollectionFormDrawerProps } from './CollectionFormDrawer';
 import CollectionLoadError from './CollectionLoadError';
+import { CollectionPageAction } from './collections.utils';
 
 export type CollectionsFormModalProps = {
     hasWriteAccessForCollections: boolean;
-    collectionId: string;
+    modalAction: Extract<
+        { type: 'create' } | { type: 'view'; collectionId: string },
+        CollectionPageAction
+    >;
     onClose: () => void;
+    onSubmit?: CollectionFormDrawerProps['onSubmit'];
+    configError?: CollectionFormDrawerProps['configError'];
+    setConfigError?: CollectionFormDrawerProps['setConfigError'];
 };
 
 function getCollectionTableCells(): ReturnType<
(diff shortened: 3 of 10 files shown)


dvail:
* master
  * **PR #4197** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4197" target="_blank">Graphite</a> 
    * **PR #4200** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4200" target="_blank">Graphite</a>  👈
    * **PR #4198** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4198" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4200?utm_source=stack-comment).

ui/apps/platform/src/Containers/VulnMgmt/Reports/Form/CollectionSelection.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+import FormLabelGroup from 'Components/PatternFly/FormLabelGroup';
+import { Collection, listCollections } from 'services/CollectionsService';
+import CollectionsFormModal from 'Containers/Collections/CollectionFormModal';
+import { useCollectionFormSubmission } from 'Containers/Collections/hooks/useCollectionFormSubmission';
+import { useFormik } from 'formik';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+import { usePaginatedQuery } from 'hooks/usePaginatedQuery';
+
+const COLLECTION_PAGE_SIZE = 10;
```
dvail, line 23: Thoughts on a smaller page size? I feel like a page size of ~5 might make users more likely to attempt to search by name before scrolling down the infinite list.

ui/apps/platform/src/Containers/VulnMgmt/Reports/Form/CollectionSelection.tsx
```diff
@@ -0,0 +1,167 @@
...
+        },
+        [search]
+    );
+
+    const { data, isEndOfResults, isFetchingNextPage, fetchNextPage } = usePaginatedQuery(
+        requestFn,
+        COLLECTION_PAGE_SIZE
+    );
+
+    // Combines the server-side fetched pages of collections data with the local cache
```
dvail, line 61: This section is required due to the following:
1. The collection dropdown is lazy loaded, and we likely won't have all collections client side when a new collection is created.
2. When a new collection is created, we want to have it inserted into the dropdown so that it is visible to the user.
3. After creating a collection, if the user loads more items via the "view more" button, it is likely that the duplicate collection will be returned. Since the pagination filtering happens inside the hook, we need to reconcile the difference here again.

ui/apps/platform/cypress/integration/vulnmanagement/reporting/reporting.helpers.js
```diff
@@ -110,10 +112,31 @@ const routeMatcherMapToCreate = {
...
+        method: 'GET',
+        url: '/v1/collections*',
+    },
+    [notifiersAlias]: {
+        method: 'GET',
+        url: '/v1/notifiers',
+    },
+};
+
+export function visitVulnerabilityReportingToCreate(staticResponseMap, isCollectionsEnabled) {
```
dvail, line 126: I went with an additional parameter based on the feature flag here since I think that will make these easier to track down and modify once the flag is deleted, but I could create separate helper functions as well if we would prefer.

pedrottimark: This is good.

ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
```diff
@@ -90,35 +103,32 @@ function CollectionsFormModal({
     } else if (data) {
         content = (
             <CollectionFormDrawer
-                // We do not want to present the user with options to change the collection when in this modal
-                hasWriteAccessForCollections={false}
-                action={{
-                    type: 'view',
-                    collectionId,
-                }}
+                hasWriteAccessForCollections={hasWriteAccessForCollections}
```
pedrottimark, line 106: Questions for my understanding. If role has **write** access:
1. For view action, you can click **Edit** button to open collection in a new page.
    * Is this also possible in reporting?
2. For create action, it is implied.
    * That is, create and read-only access is a contradiction?
    * You can create **only** in reporting, **not** in collections?

ui/apps/platform/cypress/fixtures/collections/collections.json
```diff
@@ -0,0 +1,20 @@
...
+            "id": "34e57a20-c630-4767-baa9-0cb26c1274b9",
+            "name": "Security Cluster Collection (CYPRESS)",
+            "description": "Test collection for Cypress e2e tests",
+            "resourceSelectors": [
+                {
+                    "rules": [
+                        {
+                            "operator": "OR",
+                            "fieldName": "Cluster",
+                            "values": [{ "value": "security" }]
```
vjwilson, line 13: "Values-value"
Interesting data structure from the API.

ui/apps/platform/src/Containers/VulnMgmt/Reports/Components/ScopeName.tsx
```diff
@@ -3,14 +3,17 @@ import { Button, ButtonVariant, Spinner } from '@patternfly/react-core';
...
+import useCollection from 'Containers/Collections/hooks/useCollection';
 import { getAxiosErrorMessage } from 'utils/responseErrorUtils';
+import { collectionsBasePath } from 'routePaths';
 
 type ScopeNameProps = {
     scopeId: string;
 };
 
-function ScopeName({ scopeId }: ScopeNameProps): ReactElement {
+function ResourceScopeName({ scopeId }: ScopeNameProps): ReactElement {
```
vjwilson, line 16: "Resource" was only a distinction used when we were abusing AccessScope by using it in VM reports.
It probably doesn't matter since this is going away after Collections goes live, but if we were going to rename ScopeName, `AccessScopeName` would be the parallel construction to `CollectionScopeName` below.

vjwilson (approved):

ui/apps/platform/src/Containers/Collections/CollectionFormModal.tsx
dvail: 1. This is not possible in reporting. The only path to the collection form from the reporting page place the form in "create" mode where the edit/clone/delete menu buttons are not applicable. Other collections that are linked below in the form are marked to open a new tab on the individual collection's page:
image
2. 
- An action of "create" when the user only has read access is a contradiction that is currently only enforced in FE logic. This is unfortunately currently not enforced in the type system. 
- Correct, the collection modal is not used for creation via the main collection pages, only in VM reporting.

PR #3902, opened 2022-11-23
Enable caching for eslint

dvail:
## Description

Enabled the [eslint cache](https://eslint.org/docs/latest/user-guide/command-line-interface#caching) to only lint files that have been changed since the last lint run. In CI this should always run the full lint job since `.eslintcache` is .gitignored.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

#### From top level ui/

Before change:
<pre>
yarn lint
yarn run v1.22.19
$ yarn build:packages && lerna run lint --parallel --stream
$ lerna run build --stream --ignore @stackrox/*-app
lerna notice cli v3.22.1
lerna info versioning independent
lerna notice filter excluding "@stackrox/*-app"
lerna info filter [ '!@stackrox/*-app' ]
lerna info Executing command in 2 packages: "yarn run build"
@stackrox/tailwind-config: $ postcss ./index.tw.css -o ./tailwind.css
@stackrox/ui-components: $ npm-run-all build:*
@stackrox/ui-components: $ postcss ./src/index.tw.css -o ./lib/ui-components.css
@stackrox/ui-components: $ tsc --build
lerna success run Ran npm script 'build' in 2 packages in 4.6s:
lerna success - @stackrox/tailwind-config
lerna success - @stackrox/ui-components
lerna notice cli v3.22.1
lerna info versioning independent
lerna info Executing command in 3 packages: "yarn run lint"
@stackrox/tailwind-config: $ npm-run-all lint:*
@stackrox/platform-app: $ NODE_PATH=src eslint --cache --quiet --ext .js,.ts,.tsx src cypress tailwind.config.js .prettierrc.js .eslintrc.js
@stackrox/ui-components: $ npm-run-all lint:*
@stackrox/tailwind-config: $ prettier --check '**/*.{md,css,json}'
@stackrox/ui-components: $ prettier --check '**/*.{md,css,json}'
@stackrox/tailwind-config: Checking formatting...
@stackrox/ui-components: Checking formatting...
@stackrox/tailwind-config: All matched files use Prettier code style!
[...]

PR #4201, opened 2022-12-19
Revert eslint caching for apps/platform

dvail:
## Description

Reverts https://github.com/stackrox/stackrox/pull/3902/files

I'm not sure that having a second yarn command (that we would have to run separately) in each pacakge of the repo is worth it for the relatively minor speed boost we get from having this enabled. I was hoping there was an env var way to configure this but alas I could not find one.

I might revisit this when time in the future but it isn't high priority enough to figure out a better solution for now.


dvail:
* master
  * **PR #4201** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4201" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4201?utm_source=stack-comment).

pedrottimark (approved):

PR #4219, opened 2022-12-20
Dashrews/ROX-14026 add metric to state whether central is connected to Postgres or not

dashrews78:
## Description

Add a couple metrics to prometheus for the Postgres database `postgres_active_db_size` provides the size of the active database and `postgres_connected` is a simple flag indicating if central is connected to the database or not.  `0` being NOT connected and `1` being connected.  The idea with these is to hit some immediate needs as we work the greater issue of what metrics to have.

From prometheus metrics
```
# TYPE rox_central_postgres_connected gauge
rox_central_postgres_connected 1
# TYPE rox_central_postgres_db_size_bytes gauge
rox_central_postgres_db_size_bytes{database="central_active"} 1.9427887e+07
rox_central_postgres_db_size_bytes{database="postgres"} 8.090159e+06
rox_central_postgres_db_size_bytes{database="template0"} 7.938563e+06 
rox_central_postgres_db_size_bytes{database="template1"} 7.938563e+06
```

from telemetry-data.json
```
          "databaseDetails": [
            { 
              "databaseName": "postgres",
              "databaseSizeBytes": 8090159
            },
            { 
              "databaseName": "central_active",
              "databaseSizeBytes": 19427887
            },
            { 
              "databaseName": "template1",
              "databaseSizeBytes": 7938563
            },
            { 
              "databaseName": "template0",
              "databaseSizeBytes": 7938563
            } 
          ] 
```

## Checklist

## Testing Performed

Local testing by running diagnostic bundles to see the data gets dumped into `metrics-1` and `telemetry-data.json`.  Additionally added a unit test.

### DIFF
--- a/central/globaldb/metrics/metrics.go
+++ b/central/globaldb/metrics/metrics.go
@@ -177,6 +177,20 @@ var (
 		Name:      "postgres_table_toast_bytes",
 		Help:      "bytes being used by toast for a table",
 	}, []string{"Table"})
+
+	PostgresDBSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_db_size",
+		Help:      "bytes being used by Postgres Database",
+	})
+
+	PostgresConnected = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_connected",
+		Help:      "flag indicating if central is connected to the Postgres Database",
+	})
 )
 
 // SetGaugeInt sets a value for a gauge from an int
--- a/central/globaldb/metrics/metrics.go
+++ b/central/globaldb/metrics/metrics.go
@@ -151,47 +151,47 @@ var (
 		Subsystem: metrics.CentralSubsystem.String(),
 		Name:      "postgres_table_size",
 		Help:      "estimated number of rows in the table",
-	}, []string{"Table"})
+	}, []string{"table"})
 
 	PostgresIndexSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metrics.PrometheusNamespace,
 		Subsystem: metrics.CentralSubsystem.String(),
 		Name:      "postgres_table_index_bytes",
 		Help:      "bytes being used by indexes for a table",
-	}, []string{"Table"})
+	}, []string{"table"})
 
 	PostgresTableTotalSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metrics.PrometheusNamespace,
 		Subsystem: metrics.CentralSubsystem.String(),
 		Name:      "postgres_table_total_bytes",
 		Help:      "bytes being used by the table overall",
-	}, []string{"Table"})
+	}, []string{"table"})
 
 	PostgresTableDataSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metrics.PrometheusNamespace,
 		Subsystem: metrics.CentralSubsystem.String(),
 		Name:      "postgres_table_data_bytes",
 		Help:      "bytes being used by the data for a table",
-	}, []string{"Table"})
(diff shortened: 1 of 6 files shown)


central/globaldb/postgres.go
```diff
@@ -157,13 +172,26 @@ func CollectPostgresStats(ctx context.Context, db *pgxpool.Pool) []*stats.TableS
...
 	}
 
-	return statsSlice
+	dbStats.Tables = statsSlice
+	return dbStats
+}
+
+// CollectPostgresDatabaseStats -- collect database level stats for Postgres
+func CollectPostgresDatabaseStats(postgresConfig *pgxpool.Config) {
+	cloneSize, err := pgadmin.GetDatabaseSize(postgresConfig, migrations.CurrentDatabase)
```
c-du, line 181: This could be misleading. If we only count the current database, the customer would see more spaces used. We can add another stat for the currently used and keep UsedBytes for the totally used.

central/globaldb/metrics/metrics.go
```diff
@@ -177,6 +179,20 @@ var (
...
 	}, []string{"Table"})
+
+	PostgresDBSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_active_db_size",
+		Help:      "bytes being used by Postgres Database",
+	})
+
+	PostgresConnected = prometheus.NewGauge(prometheus.GaugeOpts{
```
c-du, line 190: It would be more helpful to have the accumulated number of flapping or the frequency of the disconnect.

dashrews78: @c-du for alerting purposes, I think the 0/1 is best for that purpose.  What I would propose then is an additional metric such as `Disconnects` that is a running count.

dashrews78: @c-du I took a look at this.  I think for what we need now so the ACSCS team can alert on the condition that the flag best serves that purpose.  I'll gladly looking at some info on tracking frequency as we build out the rest of ROX-14171

central/globaldb/metrics/metrics.go
```diff
@@ -177,6 +180,27 @@ var (
 		Name:      "postgres_table_toast_bytes",
 		Help:      "bytes being used by toast for a table",
 	}, []string{"Table"})
+
+	PostgresDBSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_db_size",
```
stehessel, line 187: Can you add `_bytes` as a suffix to the metric name here? That's a common Prometheus convention.

central/globaldb/metrics/metrics.go
```diff
@@ -177,6 +180,27 @@ var (
...
+		Help:      "bytes being used by a Postgres Database",
+	}, []string{"Database"})
+
+	PostgresTotalSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_total_size",
```
stehessel, line 194: Same here

central/globaldb/postgres.go
```diff
@@ -173,13 +187,58 @@ func CollectPostgresStats(ctx context.Context, db *pgxpool.Pool) []*stats.TableS
...
+	detailsSlice := make([]*stats.DatabaseDetailsStats, 0)
+
+	for _, database := range databases {
+		dbSize, err := pgadmin.GetDatabaseSize(postgresConfig, database)
+		if err != nil {
+			log.Errorf("error fetching clone size: %v", err)
+			return detailsSlice
+		}
+
+		databaseLabel := prometheus.Labels{"Database": database}
```
stehessel, line 207: Prometheus convention usually prefers snake case. So I would propose lower case labels.

stehessel:
What about a metric for the number of active database connections?

dashrews78:
> What about a metric for the number of active database connections?

I have another ticket written for that.  https://issues.redhat.com/browse/ROX-14271

stehessel (approved):

c-du (approved):

PR #4226, opened 2022-12-20
ROX-13885: applying network policies in network simulator

sachaudh:
## Description

1. Adding another function returned by the `useNetworkSimulator` hook called `applyNetworkPolicyModification`. This function applies the network policy modification value currently stored, and changes the state accordingly. If it applies successfully, it will then switch to the active YAMLs with the new network policy added in the select dropdown. If it was not successful, it will remain in the same state and show an error

Note: I could possibly add another field to the `NetworkPolicySimulator` type called `success` or `message` that shows any successful messages. That way when we do switch to the active YAMLs we can see a message that the action was successful. If you look at the first attached video, try to determine if it's intuitive that the action was successful

### Successful apply
https://user-images.githubusercontent.com/4805485/208966204-01265613-3249-4ed1-acf1-67faf94e1027.mov

### Error when applying
https://user-images.githubusercontent.com/4805485/208966528-370bfc05-f889-402f-9547-6fd11c7cf1d4.mov

## Checklist


vjwilson (approved):
/lgtm

PR #4232, opened 2022-12-21
ROX-11513, ROX-14077: Do not run QA part 2 if part 1 fails.

porridge:
## Description

See the ticket for the rationale.
TL;DR: a failure in part 1 sometimes leaves environment in a state unpalatable to part 2

## Checklist

## Testing Performed

A [followup PR](https://github.com/stackrox/stackrox/pull/4254) with a change that breaks part1 and verify that part2 is skipped.


gavin-stackrox (approved):
Yeah, I think this makes sense (assuming it works as advertised). Similar to the rationale for enabling --fail-fast for nightlies.

PR #4234, opened 2022-12-21
ROX-14039: Pad test timeout to fix flake

porridge:
## Description

The [service can take up to 20s to return](https://github.com/stackrox/stackrox/blob/751e9fbcb93715110f542fe155744d76bd26d90e/central/development/service/service_impl.go#L107), so the timeout in the test needs to be a little higher.

As a side note, this test is inherently flaky, because among other things it depends on an external service that is [*not intended* to be used in this way](https://github.com/chromium/badssl.com#disclaimer). Tracking that in https://issues.redhat.com/browse/ROX-14078

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/tests/ca_setup_test.go
+++ b/tests/ca_setup_test.go
@@ -75,7 +75,9 @@ func TestCASetup(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(c.url, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


kurlov (approved):
LGTM

PR #4240, opened 2022-12-21
ROX-13899: Clear selected deployments when namespace deselected in new Network Graph

vjwilson:
## Description

By passing two additional parameters--the currently selected deployments and an array of all namespaces with their child deployments--to the Namespace select component of the Network Graph breadcrumbs, this additional logic filters the selected deployments whenever a namespace select change event occurs.

1. Whenever the Namespace select component re-renders, it build a map with each NS as a key, and a simple array of string values of its deployments as the key's value.
2. When a NS select event occurs, the change handler takes the new selection and uses it to filter that map of NS-deployments (object -> array of entries -> array filter -> back to filtered object)
3. It then flattens all those deployments into a single array of the allowed deployments.
4. It then filters the lists of currently selected deployments against the new list of allowed deployments.
5. In addition to updating the URL with selected namespaces, it updates the URL with filtered list of selected deployments.

Notes:
* I did not refactor this logic into a separate function, because it is currently only used in this one place. Refactoring for unit testing would be a good follow-up task.
* I also added a new optional parameter to the Network Service's  `fetchNetworkFlowGraph` and `fetchNetworkPolicyGraph` functions, that allows the new Network Graph to force Orchestrator components to always be shown.

## Checklist

## Testing Performed

Two namespaces and deployments from each selected
Screen Shot 2022-12-21 at 5 25 35 PM

After de-selecting one of those namespaces, the deployments from the deselected NS no longer appear (gone from the URL, and hence graph and the select)
Screen Shot 2022-12-21 at 5 25 55 PM

### DIFF
--- a/ui/apps/platform/src/services/NetworkService.js
+++ b/ui/apps/platform/src/services/NetworkService.js
@@ -159,7 +159,8 @@ export function fetchNetworkPolicyGraph(
     deployments,
     query,
     modification,
-    includePorts
+    includePorts,
+    includeOrchestratorComponents = false
 ) {
     const urlParams = query ? { query } : {};
     const namespaceQuery = namespaces.length > 0 ? `Namespace:${namespaces.join(',')}` : '';
@@ -172,7 +173,10 @@ export function fetchNetworkPolicyGraph(
     }
 
     // for openshift filtering toggle
-    if (localStorage.getItem(ORCHESTRATOR_COMPONENTS_KEY) !== 'true') {
+    if (
+        !includeOrchestratorComponents &&
+        localStorage.getItem(ORCHESTRATOR_COMPONENTS_KEY) !== 'true'
+    ) {
         urlParams.scope = {
             query: 'Orchestrator Component:false',
         };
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/services/NetworkService.js
```diff
@@ -236,7 +242,10 @@ export function fetchNetworkFlowGraph(
         urlParams.includePorts = true;
     }
     // for openshift filtering toggle
-    if (localStorage.getItem(ORCHESTRATOR_COMPONENTS_KEY) !== 'true') {
+    if (
+        !includeOrchestratorComponents &&
+        localStorage.getItem(ORCHESTRATOR_COMPONENTS_KEY) !== 'true'
```
sachaudh, line 247: Ohh I wasn't aware we were keeping that value in local storage. Whats the story about that?

sachaudh (approved):
/lgtm

ui/apps/platform/src/services/NetworkService.js
vjwilson: Localstorage for the "show orchestrator components" setting was part of the change we implemented when we joined Red Hat, so that people wouldn't see all the OpenShift namespaces in the Network Graph and listed on the Risk page. The setting is supposed to be the same on both pages, and since we were pulling back from Redux, localstorage was a good alternative.

PR #4241, opened 2022-12-22
ROX-13555: adding externally connected decorator to deployments

alwayshooin:
also refactored a bit of the logic in `modelUtils` to be:
* more type specific for external entities
* allow for cross referencing nodes before adding them to the data model (needed to check if the node is externally connected)
* renamed `EXTERNAL` node type to `EXTERNAL_GROUP` to be more specific

if a deployment is connected (ingress or egress) to any nodes inside the `External to cluster` group on the graph, it should have the `IP` decorator on the top right corner of the node.
image
image

everything else should work as before/expected on both active and extraneous graphs.

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
@@ -255,6 +255,9 @@ function getNetworkPolicyState(
     return networkPolicyState;
 }
 
+// external connections can only be active, so this is hard coded to false
+const POLICY_NODE_EXTERNALLY_CONNECTED_VALUE = false;
+
 export function transformPolicyData(nodes: Node[], flows: number): CustomModel {
     const dataModel = {
         graph: graphModel,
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -196,7 +263,7 @@ export function transformPolicyData(nodes: Node[], flows: number): CustomModel {
     };
     nodes.forEach(({ entity, policyIds, outEdges, nonIsolatedEgress, nonIsolatedIngress }) => {
         const networkPolicyState = getNetworkPolicyState(nonIsolatedEgress, nonIsolatedIngress);
-        const node = getNodeModel(entity, policyIds, networkPolicyState);
+        const node = getNodeModel(entity, policyIds, networkPolicyState, false, outEdges);
```
vjwilson, line 266: nit: Bare values, like `false` in this line, are hard to understand later. If we always pass the same constant, best to name it.
```
const THIS_VALUE_IS_FALSE_BECAUSE_OF_THIS_REASON = false;
```
```suggestion
        const node = getNodeModel(entity, policyIds, networkPolicyState, THIS_VALUE_IS_FALSE_BECAUSE_OF_THIS_REASON, outEdges);
```

vjwilson (approved):
Looks sane to me. Would like Saif to sanity-check, too.

(one minor nit about naming a param)

sachaudh (approved):
LGTM

PR #4244, opened 2022-12-22
Update Violations by severity widget to use search

dvail:
## Description

This PR updates the "Violations by severity" widget on the dashboard to use a graphql query with four resolvers, one for each severity, instead of the alert aggregation API in order to reduce memory pressure in environments with a large number of violations.

---

**Edit: Everything below this point, up to "Testing performed", refers to an older version of this PR that was kept here for reference.**

---

This contains two potential mitigations for https://issues.redhat.com/browse/ROX-13948 to prevent Central crashes and OOMs when visiting the main dashboard. (This is a WIP and needs some code cleanup.)

The first is a fix for the "Policy violations by severity widget" that changes the existing alert summary query to use a single graphql query that resolves the four severity violations.

```gql
query q {
  low: violationCount(query: "Severity:LOW")
  med: violationCount(query: "Severity:MEDIUM")
  high: violationCount(query: "Severity:HIGH")
  crit: violationCount(query: "Severity:CRITICAL")
}
```

The second fix is similar for the "Policy violations by category" widget, but requires a dynamic graphql query to be built that contains four resolvers for _every policy category_ that exists in the system. This looks similar to:

```gql
query violationCountsByPolicyCategory {
  Anomalous_Activity_LOW_SEVERITY: violationCount(
    query: "Category:Anomalous Activity+Severity:LOW_SEVERITY"
  )
  Anomalous_Activity_MEDIUM_SEVERITY: violationCount(
    query: "Category:Anomalous Activity+Severity:MEDIUM_SEVERITY"
  )
  Anomalous_Activity_HIGH_SEVERITY: violationCount(
    query: "Category:Anomalous Activity+Severity:HIGH_SEVERITY"
  )
  Anomalous_Activity_CRITICAL_SEVERITY: violationCount(
    query: "Category:Anomalous Activity+Severity:CRITICAL_SEVERITY"
  )
  Cryptocurrency_Mining_LOW_SEVERITY: violationCount(
    query: "Category:Cryptocurrency Mining+Severity:LOW_SEVERITY"
  )
  ...and so on
}
```

## Performance testing

These queries were tested against a cluster with 300,000 Violations to gauge memory usage and query response time. Full results are available in this [spreadsheet](https://docs.google.com/spreadsheets/d/1EMHF4C9GGPJO-Oy8ktEoFEdeylNzUhmGIfFHC2St4Ng/edit#gid=0).

### Methodology

The system was tested both with the original queries "**Old queries**" and with the queries in this PR, "**New Queries**". Query response times as well as central CPU/memory usage were recorded for a handful of scenarios. The following cycle was repeated 4 times:
1. Load the dashboard with no filters
2. Load the dashboard with 1 cluster and 3 namespace filters
3. Load the dashboard with 1 cluster and 10 namespace filters
4. Test each query individually via API request

### Results

In general, when loading the dashboard, the newer queries performed _worse_ in regards to response time, but _better_ when hit directly via the API. This was especially apparent as cluster/namespace filters were added to the queries. The old query response times became much faster when additional query parameters were added, while the new queries became slightly slower.

In contrast, the memory and CPU usage of the new queries is much less than the original queries, with pod memory usage almost static during testing for the new queries. This was the case even during the unfiltered queries, which caused either a timeout or central crash due to OOM 100% of the time that the dashboard was loaded.

Additional testing was done with only the first widget's query changed, and using the old query for the second widget. This scenario resulted in the overall most consistent response times, but memory usage was still erratic. Central would periodically crash during testing even when requests succeeded.

### Downsides

Although the new queries appear to fix the memory issues with the main dashboard, the second query suffers from two problems:
1. It scales linearly with the number of policy categories in the system. Each additional policy category will add four additional resolver calls to the query. There are query response times recorded for queries containing 1x, 3x, and 9x the number of policy categories that come with ACS by default.
2. The code it self dynamically builds the query and relies on string concatenation and parsing to work. This is a little more error prone than regular query usage.

## Checklist

## Testing Performed

Run a local copy with `YARN_START_TARGET` pointed at staging with another browser loading the staging site directly to compare counts for accuracy.

With no filter:
image

With "Cluster:production"
image

With "Cluster:production+Namespace:backend"
image

With "Cluster:production,security+Namespace:backend,frontend,stackrox"
image


dvail:
* master
  * **PR #4244** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4244" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4244?utm_source=stack-comment).

md2119 (commented):
The graphQL API looks good. I will defer thecode review to UI team.

ui/apps/platform/src/Containers/Dashboard/Widgets/ViolationsByPolicySeverity.tsx
```diff
@@ -79,8 +108,8 @@ function ViolationsByPolicySeverity() {
         CRITICAL_SEVERITY: 0,
     };
     let totalCount = 0;
-    severityCounts.forEach(({ severity, count }) => {
-        counts[severity] = parseInt(count, 10);
+    Object.entries(alertCountData ?? {}).forEach(([severity, count]) => {
+        counts[severity] = count;
         totalCount += counts[severity];
```
vjwilson, line 113: I think this is a case where a reduce would be more functional, but also make an already confusing database optimization even harder to follow.

vjwilson (approved):
/lgtm

ui/apps/platform/src/Containers/Dashboard/Widgets/ViolationsByPolicySeverity.tsx
pedrottimark: Maybe if only computation of total, but not with assignment to `counts` is a side-effect.

@dvail For my understanding, is the assignment still needed now that the GraphQL query seems to return `AlertCounts` type which is same as `counts` object?

pedrottimark (approved):

ui/apps/platform/src/Containers/Dashboard/Widgets/ViolationsByPolicySeverity.tsx
dvail: Ah sorry I missed this comment since I had automerge on and was focused on the CI failures. Yeah I think the assignment is unnecessary after the change and the sum could be changed to a one line reducer like Van mentioned `Object.values(alertCountData).reduce((a, b) => a + b)`

PR #3302, opened 2022-10-04
ROX-12826 improve re-sync service tests

lvalerom:
## Description

Re-sync service tests no longer wait for the re-sync to happen.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Unit test and regression tests added
- ~[ ] Evaluated and added CHANGELOG entry if required~
- ~[ ] Determined and documented upgrade steps~
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

If any of these don't apply, please comment below.

## Testing Performed

Manually and CI:

```
go test -race -count=1 github.com/stackrox/rox/sensor/tests/resource/service
```

PR #4246, opened 2022-12-22
ROX-13629 Add delete pod/deployment integration tests

lvalerom:
## Description

Adds integration tests for the deletion of pods and deployments.

## Checklist

## Testing Performed

* CI
* Manual testing with and without the `ROX_RESYNC_DISABLED` feature flag enabled.

### DIFF
--- a/sensor/tests/resource/helper.go
+++ b/sensor/tests/resource/helper.go
@@ -295,6 +295,32 @@ func (c *TestContext) LastResourceStateWithTimeout(matchResourceFn MatchResource
 	}
 }
 
+// WaitForDeploymentEvent waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEvent(name string) {
+	c.WaitForDeploymentEventWithTimeout(name, 3*time.Second)
+}
+
+// WaitForDeploymentEventWithTimeout waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEventWithTimeout(name string, timeout time.Duration) {
+	timer := time.NewTimer(timeout)
+	ticker := time.NewTicker(500 * time.Millisecond)
+	lastErr := errors.Errorf("the deployment %s was not processed", name)
+	for {
+		select {
+		case <-timer.C:
+			c.t.Fatalf("timeout reached waiting for deployment: %s", lastErr)
+		case <-ticker.C:
+			messages := c.GetFakeCentral().GetAllMessages()
+			lastDeploymentUpdate := GetLastMessageWithDeploymentName(messages, DefaultNamespace, name)
+			deployment := lastDeploymentUpdate.GetEvent().GetDeployment()
+			if deployment != nil {
+				return
+			}
+		}
+	}
+
+}
+
 // LastDeploymentState checks the deployment state similarly to `LastDeploymentStateWithTimeout` with a default 3 seconds timeout.
 func (c *TestContext) LastDeploymentState(name string, assertion AssertFunc, message string) {
 	c.LastDeploymentStateWithTimeout(name, assertion, message, 3*time.Second)
@@ -378,17 +404,17 @@ func (c *TestContext) LastViolationStateByID(id string, assertion AlertAssertFun
 func (c *TestContext) LastViolationStateByIDWithTimeout(id string, assertion AlertAssertFunc, message string, checkEmptyAlertResults bool, timeout time.Duration) {
 	timer := time.NewTimer(timeout)
 	ticker := time.NewTicker(10 * time.Millisecond)
-	var lastErr error
+	lastErr := errors.New("no alerts asserted")
 	for {
 		select {
 		case <-timer.C:
 			c.t.Fatalf("timeout reached waiting for violation state (%s): %s", message, lastErr)
 		case <-ticker.C:
 			messages := c.GetFakeCentral().GetAllMessages()
-			alerts := GetAllAlertsForDeploymentID(messages, id, checkEmptyAlertResults)
-			var lastViolationState *central.AlertResults
-			if len(alerts) > 0 {
-				lastViolationState = alerts[len(alerts)-1].GetEvent().GetAlertResults()
+			lastAlert := GetLastAlertsWithDeploymentID(messages, id, checkEmptyAlertResults)
+			lastViolationState := lastAlert.GetEvent().GetAlertResults()
+			if lastViolationState == nil {
+				continue
 			}
 			if lastErr = assertion(lastViolationState); lastErr == nil {
 				// Assertion matched the case. We can return here without failing the test case
@@ -405,24 +431,6 @@ func GetAllAlertsForDeploymentName(messages []*central.MsgFromSensor, name strin
 	for _, m := range messages {
 		for _, alert := range m.GetEvent().GetAlertResults().GetAlerts() {
 			if alert.GetDeployment().GetName() == name {
-				selected = append(selected, m)
-				break
-			}
-		}
-	}
-	return selected
-}
-
-// GetAllAlertsForDeploymentID filters sensor messages and gets all alerts for a deployment with `id`. If checkEmptyAlerts is set to true it will also check for AlertResults with no Alerts
-func GetAllAlertsForDeploymentID(messages []*central.MsgFromSensor, id string, checkEmptyAlertResults bool) []*central.MsgFromSensor {
-	var selected []*central.MsgFromSensor
-	for _, m := range messages {
-		if checkEmptyAlertResults && m.GetEvent().GetAlertResults().GetDeploymentId() == id && len(m.GetEvent().GetAlertResults().GetAlerts()) == 0 {
-			selected = append(selected, m)
-			break
-		}
-		for _, alert := range m.GetEvent().GetAlertResults().GetAlerts() {
-			if alert.GetDeployment().GetId() == id {
 				selected = append(selected, m)
 				break
 			}
@@ -606,10 +614,14 @@ func GetLastMessageWithDeploymentName(messages []*central.MsgFromSensor, ns, nam
 	return lastMessage
 }
 
-// GetLastAlertsWithDeploymentID find most recent alert message by deployment ID
-func GetLastAlertsWithDeploymentID(messages []*central.MsgFromSensor, id string) *central.MsgFromSensor {
+// GetLastAlertsWithDeploymentID find most recent alert message by deployment ID, If checkEmptyAlerts is set to true it will also check for AlertResults with no Alerts
+func GetLastAlertsWithDeploymentID(messages []*central.MsgFromSensor, id string, checkEmptyAlertResults bool) *central.MsgFromSensor {
 	var lastMessage *central.MsgFromSensor
 	for i := len(messages) - 1; i >= 0; i-- {
+		if checkEmptyAlertResults && messages[i].GetEvent().GetAlertResults().GetDeploymentId() == id && len(messages[i].GetEvent().GetAlertResults().GetAlerts()) == 0 {
+			lastMessage = messages[i]
+			break
+		}
 		if messages[i].GetEvent().GetAlertResults().GetDeploymentId() == id {
 			lastMessage = messages[i]
 			break
--- a/sensor/tests/resource/helper.go
+++ b/sensor/tests/resource/helper.go
@@ -260,7 +260,7 @@ func runPermutation(files []YamlTestFile, i int, cb func([]YamlTestFile)) {
 }
 
 // AssertFunc is the deployment state assertion function signature.
-type AssertFunc func(deployment *storage.Deployment) error
+type AssertFunc func(deployment *storage.Deployment, action central.ResourceAction) error
 
 // MatchResource is a function to match sensor messages to be filtered.
 type MatchResource func(resource *central.MsgFromSensor) bool
@@ -314,8 +314,9 @@ func (c *TestContext) LastDeploymentStateWithTimeout(name string, assertion Asse
 			messages := c.GetFakeCentral().GetAllMessages()
 			lastDeploymentUpdate := GetLastMessageWithDeploymentName(messages, "sensor-integration", name)
 			deployment := lastDeploymentUpdate.GetEvent().GetDeployment()
+			action := lastDeploymentUpdate.GetEvent().GetAction()
 			if deployment != nil {
-				if lastErr = assertion(deployment); lastErr == nil {
+				if lastErr = assertion(deployment, action); lastErr == nil {
 					return
 				}
 			}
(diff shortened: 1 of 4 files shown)


sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -66,6 +66,8 @@ func MergeResourceEvents(dest, src *ResourceEvent) *ResourceEvent {
 		dest.CompatibilityReprocessDeployments = append(dest.CompatibilityReprocessDeployments, src.CompatibilityReprocessDeployments...)
 		dest.ForwardMessages = append(dest.ForwardMessages, src.ForwardMessages...)
 		dest.CompatibilityDetectionDeployment = append(dest.CompatibilityDetectionDeployment, src.CompatibilityDetectionDeployment...)
+		dest.DeploymentReference = src.DeploymentReference
+		dest.ParentResourceAction = src.ParentResourceAction
```
fredrb, line 70: Is this still needed?

sensor/tests/resource/helper.go
```diff
@@ -338,6 +370,24 @@ func GetAllAlertsForDeploymentName(messages []*central.MsgFromSensor, name strin
 	return selected
 }
 
+// GetAllAlertsForDeploymentID filters sensor messages and gets all alerts for a deployment with `id`. If checkEmptyAlerts is set to true it will also check for AlertResults with no Alerts
+func GetAllAlertsForDeploymentID(messages []*central.MsgFromSensor, id string, checkEmptyAlertResults bool) []*central.MsgFromSensor {
```
fredrb, line 374: I think this logic can be simplified. If you're only interested in the last alert, you don't need to have a function that returns _all_ alerts for a deployment and then pick the last element on the list. The pre-filter is unnecessary.

My suggestion is to simply do the following:
```go
func GetLastViolationStateForDeploymentID(messages []*central.MsgFromSensor, id string, func GetAllAlertsForDeploymentID(messages []*central.MsgFromSensor, id string, checkEmptyAlertResults bool) []*central.MsgFromSensor {) *central.MsgFromSensor {
        for i := len(messages) - 1; i >= 0; i-- {
                // check empty and check ID, if all good return here.
        }
}
```

The improvement here is that helper doesn't need to filter all alerts (potentially a large number) just to always pick up the last element in the assertion.

sensor/tests/resource/helper.go
```diff
@@ -324,6 +325,37 @@ func (c *TestContext) LastViolationStateWithTimeout(name string, assertion Alert
...
+func (c *TestContext) LastViolationStateByID(id string, assertion AlertAssertFunc, message string, checkEmptyAlertResults bool) {
+	c.LastViolationStateByIDWithTimeout(id, assertion, message, checkEmptyAlertResults, 3*time.Second)
+}
+
+// LastViolationStateByIDWithTimeout checks that a violation state for a deployment must match `assertion`. If violation state does not match
+// until `timeout` the test fails.
+func (c *TestContext) LastViolationStateByIDWithTimeout(id string, assertion AlertAssertFunc, message string, checkEmptyAlertResults bool, timeout time.Duration) {
+	timer := time.NewTimer(timeout)
+	ticker := time.NewTicker(10 * time.Millisecond)
+	var lastErr error
```
fredrb, line 338: Ideally you should start with a message here, just in case the test fails without any assertion failed. 
```suggestion
	lastErr := errors.New("no alerts asserted")
```

sensor/tests/resource/helper.go
```diff
@@ -324,6 +325,37 @@ func (c *TestContext) LastViolationStateWithTimeout(name string, assertion Alert
...
+	for {
+		select {
+		case <-timer.C:
+			c.t.Fatalf("timeout reached waiting for violation state (%s): %s", message, lastErr)
+		case <-ticker.C:
+			messages := c.GetFakeCentral().GetAllMessages()
+			alerts := GetAllAlertsForDeploymentID(messages, id, checkEmptyAlertResults)
+			var lastViolationState *central.AlertResults
+			if len(alerts) > 0 {
+				lastViolationState = alerts[len(alerts)-1].GetEvent().GetAlertResults()
```
fredrb, line 348: This can panic on `nil` if you run this before any alerts for this deployment ID are sent. 
I suggest to check on `nil` and `continue` if there are no alerts to assert against.

sensor/tests/resource/pod/pod_test.go
```diff
@@ -123,3 +126,70 @@ func (s *PodHierarchySuite) Test_ParentlessPodsAreTreatedAsDeployments() {
...
+		// Check the deployment is processed
+		testC.LastDeploymentState("nginx-deployment", func(deployment *storage.Deployment, _ central.ResourceAction) error {
+			id = deployment.GetId()
+			return nil
+		}, "deployment should be processed")
+		testC.GetFakeCentral().ClearReceivedBuffer()
+
+		// Delete the deployment
+		err = deleteDep()
+		require.NoError(t, err)
```
fredrb, line 145:
```suggestion
		require.NoError(t, deleteDep())
```

sensor/tests/resource/pod/pod_test.go
```diff
@@ -123,3 +126,70 @@ func (s *PodHierarchySuite) Test_ParentlessPodsAreTreatedAsDeployments() {
...
 }
+
+func (s *PodHierarchySuite) Test_DeleteDeployment() {
+	s.testContext.RunBare("Removing a deployment should send empty violation", func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
+		deleteDep, err := testC.ApplyFileNoObject(context.Background(), resource.DefaultNamespace, NginxDeployment)
+		require.NoError(t, err)
+
+		var id string
+		// Check the deployment is processed
+		testC.LastDeploymentState("nginx-deployment", func(deployment *storage.Deployment, _ central.ResourceAction) error {
```
fredrb, line 137: This makes think that we should have an assertion for "the deployment event with name X was received", so you don't have to do it like that.

sensor/tests/resource/pod/pod_test.go
```diff
@@ -123,3 +126,70 @@ func (s *PodHierarchySuite) Test_ParentlessPodsAreTreatedAsDeployments() {
 			"Should have received four different pod events (3 from nginx-deployment and 1 from nginx-rouge")
 	})
 }
+
+func (s *PodHierarchySuite) Test_DeleteDeployment() {
+	s.testContext.RunBare("Removing a deployment should send empty violation", func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
+		deleteDep, err := testC.ApplyFileNoObject(context.Background(), resource.DefaultNamespace, NginxDeployment)
```
fredrb, line 132: Suggestion: If you need the deployment ID, I _think_ you can do that by using `ApplyFile` instead, and pass a `*v1.Deployment` reference. You'd have to test, because AFAIU the object passed will only contain the parsed properties from the yaml. But, the object is also passed to the resources API to be be created in k8s. Maybe the ID will be updated on the `*v1.Deployment` and you don't need to use the callback to figure out what is the ID?

sensor/tests/resource/pod/pod_test.go
```diff
@@ -123,3 +126,70 @@ func (s *PodHierarchySuite) Test_ParentlessPodsAreTreatedAsDeployments() {
...
+		// Check the pod is processed
+		testC.LastDeploymentState("nginx-rogue", func(deployment *storage.Deployment, _ central.ResourceAction) error {
+			id = deployment.GetId()
+			return nil
+		}, "rogue pod should be processed")
+		testC.GetFakeCentral().ClearReceivedBuffer()
+
+		// Delete the pod
+		err = deletePod()
+		require.NoError(t, err)
```
fredrb, line 178:
```suggestion
		require.NoError(t, deletePod())
```

fredrb (commented):
Some suggestions to improve the tests.

sensor/tests/resource/helper.go
lvalerom: 🤦 There was an unused function `GetLast AlertsWithDeploymentID` that does that. I've switch the code to use that one.

sensor/tests/resource/helper.go
lvalerom: Done

sensor/tests/resource/helper.go
lvalerom: I've stopped using `GetAllAlertsForDeploymentID` so now the code is different here.

sensor/tests/resource/pod/pod_test.go
lvalerom: Done

sensor/tests/resource/pod/pod_test.go
lvalerom: I've added a new function `WaitForDeploymentEvent` that basically does that. Let me know what you think.

sensor/tests/resource/pod/pod_test.go
lvalerom: Good idea. I've changed it accordingly.

sensor/tests/resource/pod/pod_test.go
lvalerom: Done

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: It should be gone after the rebase.

sensor/tests/resource/helper.go
```diff
@@ -295,6 +295,32 @@ func (c *TestContext) LastResourceStateWithTimeout(matchResourceFn MatchResource
...
+// WaitForDeploymentEvent waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEvent(name string) {
+	c.WaitForDeploymentEventWithTimeout(name, 3*time.Second)
+}
+
+// WaitForDeploymentEventWithTimeout waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEventWithTimeout(name string, timeout time.Duration) {
+	timer := time.NewTimer(timeout)
+	ticker := time.NewTicker(500 * time.Millisecond)
+	lastErr := errors.Errorf("the deployment %s was not processed", name)
```
fredrb, line 307: Technically it could've been processed, it was just never sent by Sensor.
```suggestion
	lastErr := errors.Errorf("the deployment %s was not sent" name)
```

sensor/tests/resource/helper.go
```diff
@@ -378,17 +404,17 @@ func (c *TestContext) LastViolationStateByID(id string, assertion AlertAssertFun
 func (c *TestContext) LastViolationStateByIDWithTimeout(id string, assertion AlertAssertFunc, message string, checkEmptyAlertResults bool, timeout time.Duration) {
 	timer := time.NewTimer(timeout)
 	ticker := time.NewTicker(10 * time.Millisecond)
-	var lastErr error
+	lastErr := errors.New("no alerts asserted")
```
fredrb, line 407: IIUC error won't be set if there are no violations with _this deployment ID_. Therefore, I think the message can be a bit clearer: 
```suggestion
	lastErr := errors.Errorf("no alerts sent for deployment ID %s", id)
```

sensor/tests/resource/helper.go
```diff
@@ -606,10 +614,14 @@ func GetLastMessageWithDeploymentName(messages []*central.MsgFromSensor, ns, nam
 	return lastMessage
 }
 
-// GetLastAlertsWithDeploymentID find most recent alert message by deployment ID
-func GetLastAlertsWithDeploymentID(messages []*central.MsgFromSensor, id string) *central.MsgFromSensor {
+// GetLastAlertsWithDeploymentID find most recent alert message by deployment ID, If checkEmptyAlerts is set to true it will also check for AlertResults with no Alerts
```
fredrb, line 617:
```suggestion
// GetLastAlertsWithDeploymentID find most recent alert message by deployment ID. If checkEmptyAlerts is set to true it will also check for AlertResults with no Alerts
```

sensor/tests/resource/pod/pod_test.go
```diff
@@ -129,20 +130,17 @@ func (s *PodHierarchySuite) Test_ParentlessPodsAreTreatedAsDeployments() {
...
 func (s *PodHierarchySuite) Test_DeleteDeployment() {
 	s.testContext.RunBare("Removing a deployment should send empty violation", func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
-		deleteDep, err := testC.ApplyFileNoObject(context.Background(), resource.DefaultNamespace, NginxDeployment)
-		require.NoError(t, err)
-
 		var id string
+		k8sDeployment := &appsv1.Deployment{}
+		deleteDep, err := testC.ApplyFile(context.Background(), resource.DefaultNamespace, NginxDeployment, k8sDeployment)
+		require.NoError(t, err)
+		id = string(k8sDeployment.GetUID())
```
fredrb, line 137: Cool that this works! I wasn't sure that the ID would be set in the object after a applying it.

fredrb (approved):
Looks good. Some improvement suggestions.

sensor/tests/resource/helper.go
```diff
@@ -295,6 +295,32 @@ func (c *TestContext) LastResourceStateWithTimeout(matchResourceFn MatchResource
...
 
+// WaitForDeploymentEvent waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEvent(name string) {
+	c.WaitForDeploymentEventWithTimeout(name, 3*time.Second)
+}
+
+// WaitForDeploymentEventWithTimeout waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEventWithTimeout(name string, timeout time.Duration) {
+	timer := time.NewTimer(timeout)
+	ticker := time.NewTicker(500 * time.Millisecond)
```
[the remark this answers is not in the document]
lvalerom: 10ms was making the tests burn too much CPU ultimately making them run slower to the point that in some cases they were reaching the 10min timeout. I think the 10ms ticker was left out from #3302 by mistake.

I've created the constant `defaultTicker` so that can be reused in these functions as you suggested.

sensor/tests/resource/helper.go
```diff
@@ -295,6 +295,32 @@ func (c *TestContext) LastResourceStateWithTimeout(matchResourceFn MatchResource
 	}
 }
 
+// WaitForDeploymentEvent waits until sensor process a given deployment
+func (c *TestContext) WaitForDeploymentEvent(name string) {
+	c.WaitForDeploymentEventWithTimeout(name, 3*time.Second)
```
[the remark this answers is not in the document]
lvalerom: Done

sensor/tests/resource/helper.go
lvalerom: Done

sensor/tests/resource/helper.go
lvalerom: Done

sensor/tests/resource/helper.go
lvalerom: 👍

sensor/tests/resource/pod/pod_test.go
lvalerom: Definitely makes things easier 🙂

PR #4247, opened 2022-12-22
ROX-14108: Adding simulation frame

sachaudh:
## Description

1. Adding a `SimulationFrame` which shows a border around the network graph when simulating

Screenshot 2022-12-22 at 1 00 04 PM

## Checklist
- ~[ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge~ into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/simulation/SimulationFrame.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/simulation/SimulationFrame.tsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { ScreenIcon } from '@patternfly/react-icons';
+
+import { Flex, FlexItem } from '@patternfly/react-core';
+import { NetworkPolicySimulator } from '../hooks/useNetworkPolicySimulator';
+
+type SimulationFrameProps = {
+    simulator: NetworkPolicySimulator;
+    children: React.ReactNode;
+};
+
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/simulation/SimulationFrame.tsx
```diff
@@ -0,0 +1,51 @@
...
+        style = {};
+    }
+    return (
+        <div className="pf-ri__topology-demo" style={style}>
+            {children}
+            {isSimulating && (
+                <Flex
+                    className="pf-u-p-sm"
+                    style={{
+                        backgroundColor: 'rgb(224,233,242)',
```
vjwilson, line 31: @alanonthegit We used to be able to inspect Invision for things like color values, or we'd use standard ones. Let's make a new year's resolution to get our color management back in shape. 🆕 📆

vjwilson (approved):
/lgtm

ui/apps/platform/src/Containers/NetworkGraph/simulation/SimulationFrame.tsx
sachaudh: I'll sync with Alan on some of these color values

PR #4251, opened 2022-12-23
Add e2e tests for collection deployment matching

dvail:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dvail:
* master
  * **PR #4251** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4251" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4251?utm_source=stack-comment).

bradr5 (approved):
You made it easy to understand 👍

ui/apps/platform/cypress/integration/collections/Collections.helpers.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+// visit
+
+export function visitCollections(staticResponseMap) {
+    visit(basePath, routeMatcherMapForCollections, staticResponseMap);
+
+    cy.get('h1:contains("Collections")');
+    cy.get(`${navSelectors.navExpandable}:contains("Platform Configuration")`);
+    cy.get(`${navSelectors.nestedNavLinks}:contains("Collections")`).should(
```
pedrottimark, line 30: As a follow up, we (probably me) might write an assertion helper function:
* Encapsulate navigation selectors so simple `selectors` can imply collection selectors.
* Increase parallel form with `visitFromLeftNavExpandable` below.

What do you think about:

```js
assertLeftNavExpandable('Platform Configuration', 'Collections');
```

ui/apps/platform/cypress/integration/collections/collectionsCrudWorkflow.test.js
```diff
@@ -1,29 +1,10 @@
 import withAuth from '../../helpers/basicAuth';
-import { visitCollections } from '../../helpers/collections';
 import { hasFeatureFlag } from '../../helpers/features';
+import { tryDeleteCollection, visitCollections } from './Collections.helpers';
 
 const baseUrl = '/v1/collections';
 const autocompleteUrl = `${baseUrl}/autocomplete`;
```
pedrottimark, line 6: Because the only other uses of `baseUrl` were in `tryDeleteCollection` function and only use of `autocompleteUrl` is temporary intercept, what do you think about `'/v1/collections/autocomplete'` string where it is used?

ui/apps/platform/cypress/integration/collections/deploymentMatching.test.js
```diff
@@ -0,0 +1,183 @@
...
+    assertDeploymentResultCountEquals,
+    assertDeploymentsAreMatched,
+    assertDeploymentsAreNotMatched,
+    tryDeleteCollection,
+    visitCollections,
+} from './Collections.helpers';
+import { collectionSelectors as selectors } from './Collections.selectors';
+
+const baseUrl = '/v1/collections';
+const autocompleteUrl = `${baseUrl}/autocomplete`;
```
pedrottimark, line 13: Ditto comment above about `'/v1/collections/autocomplete'`

ui/apps/platform/cypress/integration/collections/deploymentMatching.test.js
```diff
@@ -0,0 +1,183 @@
...
+        cy.get('input[name="name"]').type(sampleCollectionName);
+        cy.get('input[name="description"]').type('Matches some stackrox deployments');
+
+        cy.get('button:contains("All namespaces")').click();
+        cy.get('button:contains("Namespaces with names matching")').click();
+        cy.get('input[aria-label="Select value 1 of 1 for the namespace name"]').type('stackrox');
+        cy.get(`button:contains('stackrox')`).click();
+
+        // Test that Stackrox deployments are matched
+        assertDeploymentsAreMatched('central', 'central-db', 'collector', 'scanner', 'sensor');
```
pedrottimark, line 55: For my understanding, what is the relationship between helper functions:
* `assertDeploymentsAreMatched`
* `assertDeploymentResultCount`

Here only one is called, below both are called.
* Is that because here there might be other results than the StackRox deployments?
* Below, where the count is the number of deployments, what do you think about `deployments` array argument instead of arguments, so `deployments.length` is the expected count?

pedrottimark (approved):
A few comments. Tests make good use of accessibility attributes. Bravo!

ui/apps/platform/cypress/integration/collections/Collections.selectors.js
```diff
@@ -0,0 +1,43 @@
+const resultsPanel = `.pf-c-drawer__panel:has(h2:contains("Collection results")) `;
```
pedrottimark, line 1: Is there a trailing space in the template literal?

ui/apps/platform/cypress/integration/collections/Collections.helpers.js
dvail: Sure, or maybe `assertLeftNavCurrentPage()`?

ui/apps/platform/cypress/integration/collections/deploymentMatching.test.js
dvail: Yep, this is to ensure that 'collector' and 'sensor' are matched and that there are no other matches. Switching to an array arg and using the length is a good idea 👍

pedrottimark: Oh, here is a delayed thought. How about 2 functions whose names distinguish:
* match deployments, but allow others (current function behavior)
* match deployments, but do not allow any others (current function plus length function)

That is, second function calls the first and also encapsulates the length assertion.

pedrottimark: The assertion about the button precedes the others, so its failure indicates a false assumption in test itself?

dvail: Yeah the assertion that the "view more" doesn't exist could be a false assumption in the test itself if the developer tried to pass in 10+ deployments for matching. We could add a length check and an explicit `throw` to make it more clear as a follow up.

PR #4266, opened 2023-01-03
Telemetry: get request body from request parameters

0x656b694d:
## Description

When intercepting an API call, a telemetry interceptor function needs a way to access the request body.
For gRPC requests, it means casting the `any` object to the according type. For pure HTTP (not translated to gRPC), the body is not captured, as it would require creating a buffer to store a potentially positive body. This is probably ok, as HTTP requests are those like `/static`, `/docs` and such, which we won't track today.

## Checklist

## Testing Performed

Unit tests, real events in the following PRs.

### DIFF
--- a/pkg/telemetry/phonehome/request_params.go
+++ b/pkg/telemetry/phonehome/request_params.go
@@ -5,11 +5,8 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/pkg/errors"
 	"github.com/stackrox/rox/pkg/grpc/authn"
 )
-
-var errBadType = errors.New("unexpected body type")
 
 // RequestParams holds intercepted call parameters.
 type RequestParams struct {
--- a/pkg/telemetry/phonehome/request_params.go
+++ b/pkg/telemetry/phonehome/request_params.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stackrox/rox/pkg/grpc/authn"
 )
 
-var errNoBody = errors.New("empty body")
+var ErrNoBody = errors.New("empty body")
 var errBadType = errors.New("unexpected body type")
 
 // RequestParams holds intercepted call parameters.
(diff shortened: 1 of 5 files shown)


0x656b694d:
* master
  * **PR #4265** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4265" target="_blank">Graphite</a> 
    * **PR #4266** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4266" target="_blank">Graphite</a>  👈
      * **PR #4267** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4267" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4266?utm_source=stack-comment).

pkg/telemetry/phonehome/request_params.go
```diff
@@ -48,3 +49,12 @@ func (rp *RequestParams) HasPathIn(patterns []string) bool {
 func (rp *RequestParams) Is(s *ServiceMethod) bool {
 	return rp.Method == s.GRPCMethod || (rp.Method == s.HTTPMethod && rp.PathMatches(s.HTTPPath))
 }
+
+// GetGRPCRequestBody sets the output body argument. Returns io.EOF on empty body.
+func GetGRPCRequestBody[T any](rp *RequestParams, body **T) error {
```
stehessel, line 54: Why do you use a double pointer construction here instead of returning a pointer to the body via `(T*, error)` return type?

0x656b694d: Go cannot infer the type by return value, unfortunately, and I don't have a way to get the right type for explicit type specification.

pkg/telemetry/phonehome/request_params.go
```diff
@@ -48,3 +50,17 @@ func (rp *RequestParams) HasPathIn(patterns []string) bool {
 func (rp *RequestParams) Is(s *ServiceMethod) bool {
 	return rp.Method == s.GRPCMethod || (rp.Method == s.HTTPMethod && rp.PathMatches(s.HTTPPath))
 }
+
+// GetGRPCRequestBody returns the request body with the type inferred from the
+// API handler provided as the first argument. Returns io.EOF on nil body.
```
stehessel, line 55: Since we now return `body` directly, do we still need the error return? Can the caller not check directly if the returned body is nil?

0x656b694d: True.

stehessel (approved):

PR #4268, opened 2023-01-03
CI: Port image scan to GHA

gavin-stackrox:
## Description

Images are not pushed in OSCI so this check is moved to GHA. Also removes the docs image from the scan per ROX-12465.

## Checklist

## Testing Performed

- [x] Only runs for pushes to master

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -446,6 +446,7 @@ jobs:
           quay.io/rhacs-eng/grpc-server:${{ steps.tag.outputs.TAG }}
 
   scan-images-with-roxctl:
+    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     needs:
       - build-and-push-main
@@ -482,5 +483,4 @@ jobs:
 
       - name: Scan images
         run: |
-          roxctl version
           ./release/scripts/scan-images-with-roxctl.sh
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -467,9 +467,20 @@ jobs:
 
       - uses: ./.github/actions/handle-tagged-build
 
+      - uses: actions/download-artifact@v3
+        with:
+          name: cli-build
+
       - name: Unpack cli build
         run: |
           tar xvzf cli-build.tgz
-          source ./scripts/ci/lib.sh
-          install_built_roxctl_in_gopath
+
+      - name: Install roxctl
+        run: |
+          ./scripts/ci/lib.sh install_built_roxctl_in_gopath
           roxctl version
+
+      - name: Scan images
+        run: |
+          roxctl version
+          ./release/scripts/scan-images-with-roxctl.sh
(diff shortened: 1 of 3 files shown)


scripts/ci/jobs/release-mgmt.sh
```diff
@@ -30,10 +30,6 @@ release_mgmt() {
 
     fi
 
-    "$ROOT/release/scripts/scan-images-with-roxctl.sh" || {
```
tommartensen, line 33: @gavin-stackrox Does the `release-mgmt.sh` script run for X.Y.Z tags and release candidates? 
If so, because the GHA workflow does not build these tags yet, we would not scan the images any more IIUC.

tommartensen:
> Only runs for pushes to master

You _could_ add another branch in your workflow file under the `on.push.branches` section 🙂

.github/workflows/build.yaml
```diff
@@ -444,3 +444,43 @@ jobs:
...
+          tar xvzf cli-build.tgz
+
+      - name: Install roxctl
+        run: |
+          ./scripts/ci/lib.sh install_built_roxctl_in_gopath
+          roxctl version
+
+      - name: Scan images
+        run: |
+          ./release/scripts/scan-images-with-roxctl.sh
```
janisz, line 486: Since roxctl does not support sarif I'm afraid about output of this test (who will check it?). How about sending slack notification when there's a violation?

gavin-stackrox: With https://github.com/stackrox/stackrox/pull/4272 this will slack on scan failures

scripts/ci/jobs/release-mgmt.sh
gavin-stackrox: @tommartensen _currently_ that depends on the config in the openshift/releases repo and the versions in the release branches. So existing release patches and RCs will still be scanned this way. The next release should use GHA but how that will function is TBD.

tommartensen: Ok, just to make it absolutely clear to me: we remove the image scan from this script here, and this script is used by openshift/releases, and we do not migrate to GHA for the release building in time, do we scan the images or not?
Thanks :)

gavin-stackrox: We *have* to migrate image builds to GHA for the next release so we will scan images via GHA on the next release branches. (I say we *have* to because it is not really feasible or beneficial to support two varieties of build systems)

tommartensen: Understood.

.github/workflows/build.yaml
```diff
@@ -444,3 +444,43 @@ jobs:
...
+
+      - uses: ./.github/actions/handle-tagged-build
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: cli-build
+
+      - name: Unpack cli build
+        run: |
+          tar xvzf cli-build.tgz
```
tommartensen, line 477: nit: we could merge this step with `Install roxctl`.

tommartensen (approved):
LGTM

janisz (approved):

PR #4269, opened 2023-01-03
ROX-14172: Fix Configmanagement UI test.

dhaus67:
## Description

With the change of #4031 , the cluster permission evaluator was changed.

Previously, it was sufficient enough to only specify `Cluster role == true` for cluster role bindings, however this has changed.

A cluster role binding has `Cluster role == true && Namespace == ""`.

The query that is currently being used in the cluster permission evaluator (using `AddNullField`) does not match. The reason for this is that there are two namespace fields within the role binding proto message: one for the binding, and one for the subject. Since the subject's namespace is never empty, this query didn't work properly.

This lead to a failing UI test, since the cluster admin roles were not shown correctly.

By fixing the query as well as deleting unused code + making sure that other places respect the condition of `cluster role == true && namespace == ""`, this should be fixed.

## Checklist

## Testing Performed

- see CI.

### DIFF
--- a/pkg/k8srbac/evaluator.go
+++ b/pkg/k8srbac/evaluator.go
@@ -44,9 +44,7 @@ func MakeClusterEvaluator(roles []*storage.K8SRole, bindings []*storage.K8SRoleB
 	// Collect cluster role bindings.
 	var clusterBindings []*storage.K8SRoleBinding
 	for _, binding := range bindings {
-		// A binding is only binding cluster-wide if cluster role is true _and_ namespace is empty. Namespaced
-		// bindings may bind to cluster roles as well.
-		if binding.GetClusterRole() && binding.GetNamespace() == "" {
+		if IsClusterRoleBinding(binding) {
 			clusterBindings = append(clusterBindings, binding)
 		}
 	}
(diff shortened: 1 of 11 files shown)


dhaus67:
* master
  * **PR #4269** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4269" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4269?utm_source=stack-comment).

pkg/k8srbac/evaluator.go
```diff
@@ -45,39 +44,11 @@ func MakeClusterEvaluator(roles []*storage.K8SRole, bindings []*storage.K8SRoleB
 	// Collect cluster role bindings.
 	var clusterBindings []*storage.K8SRoleBinding
 	for _, binding := range bindings {
-		if binding.GetClusterRole() {
+		// A binding is only binding cluster-wide if cluster role is true _and_ namespace is empty. Namespaced
+		// bindings may bind to cluster roles as well.
+		if binding.GetClusterRole() && binding.GetNamespace() == "" {
```
janisz, line 49: This condition appears 3 times how about extracting it to well named function in `pkg`?

janisz (approved):

PR #4250, opened 2022-12-22
ROX-14056: Disable failing gke_version_compatibility_tests

gavin-stackrox:
## Description

This was broken by the API change in https://github.com/stackrox/stackrox/pull/4195 (I assume). Disabling until fixed.

ref: https://srox.slack.com/archives/C0321S70YK1/p1671658169495739

## Checklist
- [x] Investigated and inspected CI test results

## Testing Performed

CI is sufficient.

PR #4218, opened 2022-12-20
ROX-13963: Show banner when db unavailable

vjwilson:
## Description

A simple banner to indicate that the database is unavailable. (Closing a gap in UX, now that the Postgres database can run in a separate pod or cloud service, and be unavailable even if Central is up and serving the UI.)

Depending on the changes by @dashrews78 currently in this PR:
https://github.com/stackrox/stackrox/pull/4195

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

Deleted the database deployment and saw the message.
Screen Shot 2022-12-20 at 12 32 51 PM

PR #4270, opened 2023-01-03
Dashrews/revert metadata service change that breaks old sensors

dashrews78:
## Description

This reverts #4195, #4218, and #4250 that were part of the updates to update the data returned with the `/metadata` endpoint to inform the caller if central could communicate with the database.  This caused compatibility issues as old sensors could not process the new field.  Which is why #4250 was needed.  This change reverts all the associate changes so we can ensure master is in a good state with compatability while we look for a less impactful solution to ROX-13963 and ROX-13964

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dvail (approved):
LGTM on the UI side. We'll still want the banner for whatever the eventual solution is but we can pull that code from the initial commit when needed.

msugakov (approved):
Assuming reverts were done exclusively via `git revert` and there were no manual changes on top.

dashrews78:
> Assuming reverts were done exclusively via `git revert` and there were no manual changes on top.

@msugakov only exception was I had to generate `central/graphql/resolvers/generated.go` as the flag got added to that in some other PR after my original one.

fredrb (approved):

PR #4276, opened 2023-01-04
ROX-14176: retry protoc download

porridge:
## Description

Rather than invoke a silent `wget` once, reuse a macro that calls `curl` with retries, that already works for the operator  build.

While at it, I also fixed a broken clean target and improved portability a bit.

## Checklist

## Testing Performed

I ran the relevant targets manually twice after cleaning the files. Otherwise relying on CI.

### DIFF
--- a/make/github.mk
+++ b/make/github.mk
@@ -12,6 +12,8 @@ GET_GITHUB_RELEASE_FN = get_github_release() { \
 		for i in $$(seq $$attempts); do \
 			curl --silent --show-error --fail --location --output $${1} $${2} && break ;\
 			[ $$i -eq $$attempts ] && exit 1; \
+			echo "Retrying after $$(i*i) seconds..."; \
+			sleep $$(i*i); \
 		done ;\
 		[[ "$$(uname -s)" != "Darwin" ]] || xattr -c $${1} ;\
 		chmod +x $${1} ;\
--- a/make/github.mk
+++ b/make/github.mk
@@ -1,0 +1,18 @@
+# github.mk
+# Helpers for fetching released binaries from github projects.
+
+# For usage instructions, see uses of this macro elsewhere, since there is no
+# single standard for architecture naming.
+GET_GITHUB_RELEASE_FN = get_github_release() { \
+	[ -x $${1} ] || { \
+		set -euo pipefail ;\
+		mkdir -p bin ;\
+		attempts=5 ;\
+		for i in $$(seq $$attempts); do \
+			curl --silent --show-error --fail --location --output $${1} $${2} && break ;\
+			[ $$i -eq $$attempts ] && exit 1; \
+		done ;\
+		[[ "$$(uname -s)" != "Darwin" ]] || xattr -c $${1} ;\
+		chmod +x $${1} ;\
+	} \
+}
--- a/make/github.mk
+++ b/make/github.mk
@@ -6,6 +6,7 @@
 GET_GITHUB_RELEASE_FN = get_github_release() { \
 	[ -x $${1} ] || { \
 		set -euo pipefail ;\
+		echo "+ $${1}" ;\
 		mkdir -p bin ;\
 		attempts=5 ;\
 		for i in $$(seq $$attempts); do \
--- a/make/protogen.mk
+++ b/make/protogen.mk
@@ -37,15 +37,16 @@ endif
 PROTOC_VERSION := 21.4
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
-PROTOC_ARCH = linux
+PROTOC_OS = linux
 endif
 ifeq ($(UNAME_S),Darwin)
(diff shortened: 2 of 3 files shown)


make/protogen.mk
```diff
@@ -37,15 +37,16 @@ endif
...
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
-PROTOC_ARCH = linux
+PROTOC_OS = linux
 endif
 ifeq ($(UNAME_S),Darwin)
-PROTOC_ARCH = osx
+PROTOC_OS = osx
 endif
+PROTOC_ARCH=$(shell case $$(uname -m) in (aarch64) echo -n armv6 ;; (*) echo -n $$(uname -m) ;; esac)
```
msugakov, line 45: Would just `uname -m` work for the default case?

```suggestion
PROTOC_ARCH=$(shell case $$(uname -m) in (aarch64) echo -n armv6 ;; (*) uname -m ;; esac)
```

make/protogen.mk
```diff
@@ -37,15 +37,16 @@ endif
...
 
 PROTO_PRIVATE_DIR := $(BASE_PATH)/.proto
 
-PROTOC_DIR := $(PROTO_PRIVATE_DIR)/protoc-$(PROTOC_ARCH)-$(PROTOC_VERSION)
+PROTOC_DIR := $(PROTO_PRIVATE_DIR)/protoc-$(PROTOC_OS)-$(PROTOC_VERSION)
```
msugakov, line 49: I suggest including both OS and Arch to the directory name:

```suggestion
PROTOC_DIR := $(PROTO_PRIVATE_DIR)/protoc-$(PROTOC_OS)-$(PROTOC_ARCH)-$(PROTOC_VERSION)
```

make/github.mk
```diff
@@ -0,0 +1,19 @@
...
+# single standard for architecture naming.
+GET_GITHUB_RELEASE_FN = get_github_release() { \
+	[ -x $${1} ] || { \
+		set -euo pipefail ;\
+		echo "+ $${1}" ;\
+		mkdir -p bin ;\
+		attempts=5 ;\
+		for i in $$(seq $$attempts); do \
+			curl --silent --show-error --fail --location --output $${1} $${2} && break ;\
+			[ $$i -eq $$attempts ] && exit 1; \
```
msugakov, line 14: Thinking aloud: let's maybe add 1-5 seconds sleep between attempts?
```suggestion
			curl --silent --show-error --fail --location --output $${1} $${2} && break ;\
			sleep 2 ;\
			[ $$i -eq $$attempts ] && exit 1; \
```

msugakov (approved):

make/protogen.mk
porridge: Yeah, I don't think the `-n` for `echo` is providing any value. Also changed in the operator Makefile.

make/github.mk
porridge: Good point. I added exponential backoff :wink:

make/protogen.mk
```diff
@@ -37,15 +37,16 @@ endif
...
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
-PROTOC_ARCH = linux
+PROTOC_OS = linux
 endif
 ifeq ($(UNAME_S),Darwin)
-PROTOC_ARCH = osx
+PROTOC_OS = osx
 endif
+PROTOC_ARCH=$(shell case $$(uname -m) in (aarch64) echo armv6 ;; (*) uname -m ;; esac)
```
connorgorman, line 45: Not sure I'm following this logic? This should move arm64 to aarch64 right?

connorgorman: This broke building on m1 macs

connorgorman: at least for downloading protoc

porridge: Sorry for the inconvenience @connorgorman .

Interesting, because previously line 64 was hardcoded to `x86_64` regardless of runtime CPU architecture.
Are you saying that `x86_64` tarball actually works on M1?? Or did you just happen to have an M1-compatible binary stashed away manually at `$(PROTO_PRIVATE_DIR)/protoc-$(PROTOC_ARCH)-$(PROTOC_VERSION)` and now this makefile changed to `$(PROTO_PRIVATE_DIR)/protoc-$(PROTOC_OS)-$(PROTOC_ARCH)-$(PROTOC_VERSION)` and your local copy is no longer used?

Can you please provide more information about how it's actually broken now?

The reason I chose this specific recipe for converting `uname -m` to download file infix is by comparing the download directories for `protoc` to the one for `kuttl`, which I assume works because nobody ever complained :wink:

porridge: Anyway, this should be solved by #4370

connorgorman: Ah! it probably did work on m1 with amd64 because I would assume that rosetta did its magic :)

PR #4277, opened 2023-01-04
ROX-14177: Retry `go mod download`.

porridge:
## Description

Apparently it can fail sometimes due to go proxy unavailability, see parent of the ticket.

Ideally we could just reuse the cached modules inside the container, but that will be a more involved change, if it's at all possible.

## Checklist

## Testing Performed

Relying on CI.

### DIFF
--- a/operator/Dockerfile
+++ b/operator/Dockerfile
@@ -25,7 +25,8 @@ COPY go.sum go.sum
 # However it does _not_ seem to resolve packages _incorrectly_ and so should be safe especially given that downloaded
 # packages are only used during build and later this docker layer is discarded (only resulting binary goes in the final
 # image).
-RUN go mod download
+# Retry as the proxy can be unavailable at times.
+RUN go mod download || go mod download || go mod download
 
 # Copy operator source
 COPY operator/ operator/


porridge:
It seems the failures are unrelated, enabling automerge on approval.

operator/Dockerfile
```diff
@@ -25,7 +25,8 @@ COPY go.sum go.sum
 # However it does _not_ seem to resolve packages _incorrectly_ and so should be safe especially given that downloaded
 # packages are only used during build and later this docker layer is discarded (only resulting binary goes in the final
 # image).
-RUN go mod download
+# Retry as the proxy can be unavailable at times.
+RUN go mod download || go mod download || go mod download
```
janisz, line 29: How about using multiple GOPROXY
https://github.com/stackrox/stackrox/blob/538a30411ba7b7ed5fad629a7b28880459ea9895/make/env.mk#L32
See: https://github.com/stackrox/stackrox/pull/164

janisz, line 29:
```suggestion
ENV GOPROXY=https://proxy.golang.org|https://goproxy.io|direct 
RUN go mod download
```

porridge:
Rebased to hopefully pick up a fix for expired quay token.
PTAL @janisz

porridge:
Operator build and tests passed, re-enabling automerge on approval.

porridge:
Looks like the discussion in the other PR might take some time. Can we merge this in the meantime @janisz ?

janisz:
@porridge sure let's merge it. But why you don't want to use proxy fallback?

janisz (approved):

porridge:
Ah, I must have reverted that during the rebase by mistake. Let me prepare a followup PR.

porridge:
Ah, you already have https://github.com/stackrox/stackrox/pull/4343, thanks!

PR #4278, opened 2023-01-04
Use RWMutex for the list of API interceptors

0x656b694d:
## Checklist

## Testing Performed

Unit tests.

### DIFF
--- a/pkg/telemetry/phonehome/client_config.go
+++ b/pkg/telemetry/phonehome/client_config.go
@@ -52,7 +52,7 @@ type Config struct {
 	// Map of event name to the list of interceptors, that gather properties for
 	// the event.
 	interceptors     map[string][]Interceptor
-	interceptorsLock sync.Mutex
+	interceptorsLock sync.RWMutex
 }
 
 // Enabled tells whether telemetry data collection is enabled.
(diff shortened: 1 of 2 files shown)


0x656b694d:
* master
  * **PR #4278** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4278" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4278?utm_source=stack-comment).

pkg/telemetry/phonehome/client_config.go
```diff
@@ -52,7 +52,7 @@ type Config struct {
 	// Map of event name to the list of interceptors, that gather properties for
 	// the event.
 	interceptors     map[string][]Interceptor
-	interceptorsLock sync.Mutex
+	interceptorsLock sync.RWMutex
 }
 
 // Enabled tells whether telemetry data collection is enabled.
```
dhaus67, line 58: One unrelated question that isn't related to the PR:

We now have the package scoped variable (`onceGatherer/onceTelemeter`) and the struct variables intertwined.
I wonder if it wouldn't be more beneficial to make sure we split the package scoped variables and struct variables by doing the following:

- Splitting the creation of the singleton into a static function (e.g. `GathererSingleton()/TelemeterSingleton()`)
- Within the `cfg.Gatherer/Telemeter` methods, setting the value on the struct.

WDYT?

0x656b694d: Gatherer and Telemeter not need to be singletons, they're instances referenced by a configuration singleton instead.
What's suboptimal, indeed, is that their initialization controlled by Once package variables. Therefore, I'll rather get rid of the package level variables ensuring synchronization on telemeter and gatherer initialization for a single configuration instance.
Thanks for noticing that.

PR #4279, opened 2023-01-04
Remove hardcoded Postgres minor version

vladbologa:
## Description

The Postgres minor version that is hardcoded in `image/rhel/create-bundle.sh` is not guaranteed to exist. For example, for `arm64` architectures, the file does not exist anymore.

The minor version is actually determined programmatically in `image/postgres/create-bundle.sh`, so I used the same approach here.

## Checklist
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs]~~(https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

I used `make image` to check that build works on `arm64`. CI should test that it works on `amd64`.

### DIFF
--- a/image/rhel/create-bundle.sh
+++ b/image/rhel/create-bundle.sh
@@ -34,6 +34,8 @@ INPUT_ROOT="$1"
 DATA_IMAGE="$2"
 BUILDER_IMAGE="$3"
 OUTPUT_DIR="$4"
+# Install the PG repo natively if true (versus using a container)
+NATIVE_PG_INSTALL="${5:-false}"
 
 [[ -n "$INPUT_ROOT" && -n "$DATA_IMAGE" && -n "$BUILDER_IMAGE" && -n "$OUTPUT_DIR" ]] \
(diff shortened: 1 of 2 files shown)


image/rhel/create-bundle.sh
```diff
@@ -34,6 +34,8 @@ INPUT_ROOT="$1"
 DATA_IMAGE="$2"
 BUILDER_IMAGE="$3"
 OUTPUT_DIR="$4"
+# Install the PG repo natively if true (versus using a container)
+NATIVE_PG_INSTALL="${5:-false}"
```
janisz, line 38: when it will be set to `false`? Do we use this script in other places?

janisz (approved):

image/rhel/create-bundle.sh
vladbologa: It's `false` e.g. when making local builds. It's only set to `true` on OpenShift CI, where we already have RHEL (and there's no docker server)

PR #4281, opened 2023-01-04
Add e2e tests for Collection permissions

dvail:
## Description

Adds tests to check the behavior of a user having `NO_ACCESS`, `READ_ACCESS`, or `READ_WRITE_ACCESS` on the `WorkflowAdministration` permission as it pertains to rendering the Collections UI.

## Checklist

## Testing Performed

Local Cypress test run.
image

### DIFF
--- a/ui/apps/platform/cypress/integration/collections/permissions.test.js
+++ b/ui/apps/platform/cypress/integration/collections/permissions.test.js
@@ -69,7 +69,7 @@ describe('Collection permission checks', () => {
         cy.get(`button:contains("Save")`).should('not.exist');
     });
 
-    it('should not provide the full UI to users with read-write access', () => {
+    it('should provide the full UI to users with read-write access', () => {
         // Mock a 'READ_WRITE_ACCESS' permission response
         visitWithStaticResponseForPermissions('/main', {
             body: { resourceToAccess: { WorkflowAdministration: 'READ_WRITE_ACCESS' } },
--- a/ui/apps/platform/cypress/integration/collections/permissions.test.js
+++ b/ui/apps/platform/cypress/integration/collections/permissions.test.js
@@ -1,0 +1,116 @@
+import withAuth from '../../helpers/basicAuth';
+import { hasFeatureFlag } from '../../helpers/features';
+import { visit, visitWithStaticResponseForPermissions } from '../../helpers/visit';
+import navSelectors from '../../selectors/navigation';
+import { tryDeleteCollection, visitCollections } from './Collections.helpers';
+import { collectionSelectors } from './Collections.selectors';
+
+// Quick utility to create a basic collection for permission testing
+function createCollection(name) {
+    // Ignore autocomplete requests
+    // TODO Remove this once the feature is in
+    cy.intercept('/v1/collections/autocomplete', {});
+    localStorage.setItem('access_token', Cypress.env('ROX_AUTH_TOKEN'));
+    tryDeleteCollection(name);
+    visitCollections();
+    cy.get('a:contains("Create collection")').click();
+    cy.get('input[name="name"]').type(name);
+    cy.get('input[name="description"]').type('A collection for permission testing purposes');
+    cy.get('button:contains("All namespaces")').click();
+    cy.get('button:contains("Namespaces with names matching")').click();
+    cy.get('input[aria-label="Select value 1 of 1 for the namespace name"]').type('stackrox');
+    cy.get(`button:contains('stackrox')`).click();
+    cy.get('button:contains("Save")').click();
+}
+
+describe('Collection permission checks', () => {
+    withAuth();
+
+    const collectionName = 'Permission test collection';
+
+    beforeEach(function beforeHook() {
+        if (!hasFeatureFlag('ROX_OBJECT_COLLECTIONS')) {
+            this.skip();
+        }
+    });
+
+    // Ensure a collection exists in the system for permission tests
+    if (hasFeatureFlag('ROX_OBJECT_COLLECTIONS')) {
+        before(() => createCollection(collectionName));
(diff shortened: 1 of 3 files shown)


dvail:
* master
  * **PR #4251** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4251" target="_blank">Graphite</a> 
    * **PR #4281** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4281" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4281?utm_source=stack-comment).

ui/apps/platform/cypress/integration/collections/permissions.test.js
```diff
@@ -0,0 +1,116 @@
...
+    tryDeleteCollection(name);
+    visitCollections();
+    cy.get('a:contains("Create collection")').click();
+    cy.get('input[name="name"]').type(name);
+    cy.get('input[name="description"]').type('A collection for permission testing purposes');
+    cy.get('button:contains("All namespaces")').click();
+    cy.get('button:contains("Namespaces with names matching")').click();
+    cy.get('input[aria-label="Select value 1 of 1 for the namespace name"]').type('stackrox');
+    cy.get(`button:contains('stackrox')`).click();
+    cy.get('button:contains("Save")').click();
```
pedrottimark, line 23: Can be a follow up to write a helper function like `saveCollection` to encapsulate:
* click button
* wait on POST request

ui/apps/platform/cypress/integration/collections/permissions.test.js
```diff
@@ -0,0 +1,116 @@
...
+        if (!hasFeatureFlag('ROX_OBJECT_COLLECTIONS')) {
+            this.skip();
+        }
+    });
+
+    // Ensure a collection exists in the system for permission tests
+    if (hasFeatureFlag('ROX_OBJECT_COLLECTIONS')) {
+        before(() => createCollection(collectionName));
+        after(() => tryDeleteCollection(collectionName));
+    }
```
pedrottimark, line 41: This `before` and `after` idiom makes sense because there are no default collections.

The other side of the coin is (as you have done) for test files to have clear purpose. In the past, imperative idioms like `beforeEach` outside of tests caused confusion in large test files which had no clear focus.

ui/apps/platform/cypress/integration/collections/permissions.test.js
pedrottimark: In case you need or want to reuse a few collections for various tests, then if the helper function creates only if the collection does not already exist:
* It achieves reuse without coupling between tests or test files.
* It seems as if conditional create replaces need to try delete for view-only collections.

ui/apps/platform/cypress/integration/collections/permissions.test.js
pedrottimark: Discussion above might remove need for this idiom.

ui/apps/platform/cypress/integration/collections/permissions.test.js
dvail: Agreed, I'll switch this to an API based request.

I like the idea of deleting at the end of the test just for the sake of cleanup. I suppose in this case since the helper to create the function will always check existence first it isn't as much of a priority though.

ui/apps/platform/cypress/integration/collections/permissions.test.js
dvail: Hmm it could, if the first test in the suite runs the creation function and the last test in the suite runs deletion. (Although if we opt to not delete, that isn't necessary.)

I kind of like the `before`/`after` idiom though, when it is fairly minimal like this. It gives a clear location for setup/teardown type tasks.

pedrottimark: Yeah, good point. Although I prefer explicit function call in each test which requires a (possibly conditionally created) view-only collection, teardown to delete at the end of the test file closes the door on a subtle intermittent failure if cypress occasionally runs test files out or order so (usually) following test file depends on the collection without having called the function.

pedrottimark: Although so far I have seen only subfolders out of order, cypress makes no guarantee about order of test files.

ui/apps/platform/cypress/integration/collections/permissions.test.js
dvail: I moved this out to a helper function and verified that Cypress waits for the request to complete before proceeding, ince we can't `cy.wait()` on a request initiated via `cy.request()`.

ui/apps/platform/cypress/integration/collections/permissions.test.js
```diff
@@ -0,0 +1,104 @@
...
+        // Ensure table rows do not have an action menu
+        cy.get(`tr:has(${linkSelector}) button[aria-label="Actions"]`).should('not.exist');
+        // Visit page for individual collection and verify action button is not present
+        cy.get(linkSelector).click();
+        cy.get(`h1:contains("${collectionName}")`);
+        cy.get(`button:contains("Actions")`).should('not.exist');
+        cy.get(`button:contains("Save")`).should('not.exist');
+    });
+
+    it('should not provide the full UI to users with read-write access', () => {
```
pedrottimark, line 72: Is the **not** in test title from copy-paste?

pedrottimark (approved):
This will become a valuable template for similar permissions tests in other containers.

ui/apps/platform/cypress/integration/collections/permissions.test.js
dvail: Guilty...

PR #4286, opened 2023-01-04
ROX-13936: Introduce ID for RHELComponents

Maddosaurus:
## Description

This PR introduces an ID for RHELComponent conversions in Node Inventories.
As described in ROX-13936, a unique ID is needed for all components to be correctly displayed.

`ID`s only need to be unique per `Node` (and therefore, per `NodeInventory`).
Packages are generally identified by the unique tuple `(Name, Version, Arch, Module)` in Scanner, whereas the ID is currently only used in [Scanners rhelv2_vulnerability.go](https://github.com/stackrox/scanner/blob/4aec14bf42567fc3a8972826b9537a444aae5e0d/database/pgsql/rhelv2_vulnerability.go#L211) to deduplicate a list of incoming packages.
Hence, I have introduced a local structure to check for duplicates locally by checking the component tuple before converting said component.

## Checklist

### DIFF
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -92,8 +92,5 @@ func convertAndDedupRHELComponents(rc *database.RHELv2Components) []*scannerV1.R
 }
 
 func makeComponentKey(component *scannerV1.RHELComponent) string {
-	if component == nil {
-		return ""
-	}
 	return component.Name + ":" + component.Version + ":" + component.Arch + ":" + component.Module
 }
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -65,13 +65,13 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
 	}
 	v1rhelc := make([]*scannerV1.RHELComponent, 0, len(rc.Packages))
 	for _, rhelc := range rc.Packages {
-		rhelcId, err := hashstructure.Hash(rhelc, hashstructure.FormatV2, nil)
+		rhelcID, err := hashstructure.Hash(rhelc, hashstructure.FormatV2, nil)
 		if err != nil {
-			log.Warnf("Could not create id for RHELComponent %d", rhelc.Name)
-			rhelcId = 0
+			log.Warnf("Could not create id for RHELComponent %q", rhelc.Name)
+			rhelcID = 0
 		}
 		v1rhelc = append(v1rhelc, &scannerV1.RHELComponent{
-			Id:          int64(rhelcId),
+			Id:          int64(rhelcID),
 			Name:        rhelc.Name,
 			Namespace:   rc.Dist,
 			Version:     rhelc.Version,
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -95,5 +95,5 @@ func makeComponentKey(component *scannerV1.RHELComponent) string {
 	if component == nil {
 		return ""
 	}
-	return component.Name +":"+  component.Version +":"+ component.Arch +":"+ component.Module
+	return component.Name + ":" + component.Version + ":" + component.Arch + ":" + component.Module
 }
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -6,6 +6,7 @@ import (
 	"github.com/stackrox/scanner/database"
 	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
 	"github.com/stackrox/scanner/pkg/analyzer/nodes"
+	"golang.org/x/exp/maps"
 )
 
 // NodeInventorizer is the interface that defines the interface a scanner must implement
@@ -46,7 +47,7 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
 	}
 
 	// For now, we only care about RHEL components, but this must be extended once we support non-RHCOS
-	rhelComponents := convertRHELComponents(c.CertifiedRHELComponents)
+	rhelComponents := convertAndDedupRHELComponents(c.CertifiedRHELComponents)
 
 	protoComponents := &scannerV1.Components{
 		Namespace:          c.OSNamespace.Name,
@@ -57,29 +58,16 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
 	return protoComponents
 }
 
-func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELComponent {
+func convertAndDedupRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELComponent {
 	if rc == nil || rc.Packages == nil {
 		log.Warn("No RHEL packages found in scan result")
 		return nil
 	}
 
-	seenEntries := make(map[int]*database.RHELv2Package)
-	v1rhelc := make([]*scannerV1.RHELComponent, 0, len(rc.Packages))
-
-	// referencing this label by the inner `continue` lets us skip the rest of the outer loop as well,
-	// resulting in the colliding component not being added to v1rhelc.
-COMPONENTS:
+	convertedComponents := make(map[string]*scannerV1.RHELComponent, 0)
 	for i, rhelc := range rc.Packages {
-		for _, candidate := range seenEntries {
-			if equalRHELv2Packages(candidate, rhelc) {
-				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %v for id %v", candidate, i)
-				continue COMPONENTS
-			}
-		}
-		// If the entry didn't produce a collision, add it to seenEntries and result
-		seenEntries[i] = rhelc
-		log.Debugf("Adding component %v to v1rhelc", rhelc.Name)
-		v1rhelc = append(v1rhelc, &scannerV1.RHELComponent{
+		comp := &scannerV1.RHELComponent{
+			// The loop index is used as ID, as this field only needs to be unique for each NodeInventory result slice
 			Id:          int64(i),
 			Name:        rhelc.Name,
 			Namespace:   rc.Dist,
@@ -88,17 +76,24 @@ COMPONENTS:
 			Module:      rhelc.Module,
 			Cpes:        rc.CPEs,
 			Executables: rhelc.Executables,
-		})
+		}
+		compKey := makeComponentKey(comp)
+		if compKey != "" {
+			if _, contains := convertedComponents[compKey]; !contains {
+				log.Debugf("Adding component %v to convertedComponents", comp.Name)
+				convertedComponents[compKey] = comp
+			} else {
+				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %v at index %v", comp, i)
+			}
+		}
+
 	}
-	return v1rhelc
+	return maps.Values(convertedComponents)
 }
 
-func equalRHELv2Packages(a *database.RHELv2Package, b *database.RHELv2Package) bool {
-	if a == nil || b == nil {
-		return false
+func makeComponentKey(component *scannerV1.RHELComponent) string {
+	if component == nil {
+		return ""
 	}
-	if a.Name == b.Name && a.Version == b.Version && a.Arch == b.Arch && a.Module == b.Module {
-		return true
-	}
-	return false
+	return component.Name + component.Version + component.Arch + component.Module
 }
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -2,6 +2,7 @@ package nodeinventorizer
 
 import (
 	timestamp "github.com/gogo/protobuf/types"
+	"github.com/mitchellh/hashstructure/v2"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/scanner/database"
 	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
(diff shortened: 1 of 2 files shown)


compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -78,3 +92,13 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
...
 }
+
+func equalRHELv2Packages(a *database.RHELv2Package, b *database.RHELv2Package) bool {
+	if a == nil || b == nil {
+		return false
+	}
+	if a.Name == b.Name && a.Version == b.Version && a.Arch == b.Arch && a.Module == b.Module {
+		return true
+	}
+	return false
```
vikin91, line 103:
```suggestion
	return a.Name == b.Name && a.Version == b.Version && a.Arch == b.Arch && a.Module == b.Module
```

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -63,10 +62,25 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
 		log.Warn("No RHEL packages found in scan result")
 		return nil
 	}
+
+	seenEntries := make(map[int]*database.RHELv2Package)
```
vikin91, line 66: How about indexing by name+version+module+arch while the value is the ID?

```suggestion
	seenEntries := make(map[string]int)
```

This would allow us to search for duplicates in O(1).

We could go further and not care about the ID in the map as we use the index anyway - this could lead to `make(map[string]struct{})`. As this makes it practically a set, I think there was already a set implementation available somewhere in our code.

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -63,10 +62,25 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
...
 	v1rhelc := make([]*scannerV1.RHELComponent, 0, len(rc.Packages))
-	for _, rhelc := range rc.Packages {
+
+	// referencing this label by the inner `continue` lets us skip the rest of the outer loop as well,
+	// resulting in the colliding component not being added to v1rhelc.
+COMPONENTS:
```
vikin91, line 71: Following the suggestion about map key should help to get rid of this label

compliance/collection/nodeinventorizer/nodeinventory_test.go
```diff
@@ -0,0 +1,211 @@
...
+	for caseName, testCase := range testCases {
+		s.Run(caseName, func() {
+			mockComponents := &database.RHELv2Components{
+				Dist:     "MockDist",
+				CPEs:     nil,
+				Packages: testCase.inComponents,
+			}
+			convertedComponents := convertRHELComponents(mockComponents)
+			if testCase.inComponents != nil {
+				convertedIDs := make([]int64, 0, len(convertedComponents))
```
vikin91, line 86: Why do we need `convertedIDs`? I cannot see it being used anywhere after it is filled

compliance/collection/nodeinventorizer/nodeinventory_test.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+type NodeInventorizerTestSuite struct {
+	suite.Suite
+}
+
+func (s *NodeInventorizerTestSuite) TestConvertRHELComponentIDs() {
+	testCases := map[string]struct {
+		inComponents  []*database.RHELv2Package
+		outComponents []*scannerV1.RHELComponent
```
vikin91, line 22: Looks like this is leftover from a previous state of this test, where you compared in against out. I think it can be removed now.

vikin91 (changes requested):
Looks good, but I think there are some bits that should be corrected before merging.

compliance/collection/nodeinventorizer/nodeinventory.go
jvdm: Nit: If we keep this (not necessary if we use a map): is it OK to break in different lines? Easier to tell what we are comparing to someone reading it.

compliance/collection/nodeinventorizer/nodeinventory.go
jvdm: I agree with @vikin91 suggestion. Going even further, is there a reason that prevents us from using only one map, map[string]*scannerV1.RHELComponent? In the return, we would return all the values of the map.

Although we plan to use the index as the ID, we don't care if the ID reflects the order in the final return value, do we?

I would also rename this method to `convertAndDedupRHELComponents`.

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -63,10 +62,25 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
...
+	for i, rhelc := range rc.Packages {
+		for _, candidate := range seenEntries {
+			if equalRHELv2Packages(candidate, rhelc) {
+				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %v for id %v", candidate, i)
```
jvdm, line 75: s/for id/at index/

jvdm (changes requested):

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -63,10 +62,25 @@ func convertRHELComponents(rc *database.RHELv2Components) []*scannerV1.RHELCompo
...
+				continue COMPONENTS
+			}
+		}
+		// If the entry didn't produce a collision, add it to seenEntries and result
+		seenEntries[i] = rhelc
+		log.Debugf("Adding component %v to v1rhelc", rhelc.Name)
 		v1rhelc = append(v1rhelc, &scannerV1.RHELComponent{
-			// TODO(ROX-13936): Find out if ID field is needed here
+			Id:          int64(i),
```
jvdm, line 83: A quick comment on why the index is being used as the ID would be helpful.

compliance/collection/nodeinventorizer/nodeinventory_test.go
Maddosaurus: Thank you for spotting this. Both, this and the `convertedIds` were leftovers of earlier iterations

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: Done!

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: Done and done!

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: Both are great ideas, I like them a lot.

It is true that we are not interested in the order of items, only that the IDs are unique.
The uniqueness check here is more of a sanity safeguard than an actual dedup, as we don't expect to encounter any duplicates in an inventory run.

That said, I've switched to a `map[string]*RHELComponent` and return the values of it as a slice.

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -58,23 +58,42 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
...
+	}
+	return maps.Values(convertedComponents)
+}
+
+func makeComponentKey(component *scannerV1.RHELComponent) string {
+	if component == nil {
+		return ""
 	}
-	return v1rhelc
+	return component.Name + component.Version + component.Arch + component.Module
```
vikin91, line 98: I would suggest to use a separator here. Imagine the following name-version pairs:
- name: ab, version: cd => abcd
- name: abc, version d => abcd

We could do instead:
```suggestion
	return component.Name +":"+  component.Version +":"+ component.Arch +":"+ component.Module
```

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -58,23 +58,42 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
...
 			Executables: rhelc.Executables,
-		})
+		}
+		compKey := makeComponentKey(comp)
+		if compKey != "" {
+			if _, contains := convertedComponents[compKey]; !contains {
+				log.Debugf("Adding component %v to convertedComponents", comp.Name)
+				convertedComponents[compKey] = comp
+			} else {
+				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %v at index %v", comp, i)
```
vikin91, line 86: Why not being more specific than `%v`?
```suggestion
				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %s at index %d", compKey, i)
```

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -58,23 +58,42 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
...
+			} else {
+				log.Warnf("Detected package collision in Node Inventory scan. Skipping package %s at index %d", compKey, i)
+			}
+		}
+
+	}
+	return maps.Values(convertedComponents)
+}
+
+func makeComponentKey(component *scannerV1.RHELComponent) string {
```
RTann, line 94: nit: Maybe make this return a struct instead of a string

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -58,23 +58,42 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
...
 		return nil
 	}
-	v1rhelc := make([]*scannerV1.RHELComponent, 0, len(rc.Packages))
-	for _, rhelc := range rc.Packages {
-		v1rhelc = append(v1rhelc, &scannerV1.RHELComponent{
-			// TODO(ROX-13936): Find out if ID field is needed here
+
+	convertedComponents := make(map[string]*scannerV1.RHELComponent, 0)
+	for i, rhelc := range rc.Packages {
+		comp := &scannerV1.RHELComponent{
```
RTann, line 69: nit: maybe make this after we ensure we need it so we can reduce allocations/de-allocations when a duplicate is found

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -58,23 +58,42 @@ func protoComponentsFromScanComponents(c *nodes.Components) *scannerV1.Component
...
+			}
+		}
+
+	}
+	return maps.Values(convertedComponents)
+}
+
+func makeComponentKey(component *scannerV1.RHELComponent) string {
+	if component == nil {
+		return ""
```
jvdm, line 96: I don't get this validation.

I think you are trying to make the code resilient to bad input, which is good. But looking at the caller, we don't anticipate the component to be `nil`. There is no path in the logic that depends on input that might yield a `nil` component either.

I think people reading calls to this function will always think, "Which components will generate empty keys?" to understand "Why are we skipping some components?", and this question is completely irrelevant at the moment.

Even if we could have `nil` components when calling `makeComponentKey()` -- let's say we reuse it in another place where that will be the case -- I would still say it is more readable to leave to the caller the responsibility of checking the condition and dealing with it instead of delegating the check to the key generation function.

jvdm (changes requested):

compliance/collection/nodeinventorizer/nodeinventory.go
jvdm: @Maddosaurus Since this is not a blocker, and everything else LGTM, I decided to approve, and you can either follow up on this here or another time.

jvdm (approved):
@Maddosaurus Since my last comment is not a blocker, and everything else LGTM, I decided to approve, and you can either follow up on the comment here or another time.

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: I could make it return a struct, but then I'd either need to convert it to a string in the caller, or switch the callers' `convertedComponents` map to take a struct as key.
The latter would bring us back to the idea discussed [here](https://github.com/stackrox/stackrox/pull/4286#discussion_r1065254337), which is a viable alternative.

But doing that introduces a possible problem:
We're considering the tuple (name,version,arch,module) for uniqueness checks. `RHELComponents` does have more fields than these four, which could lead to us producing duplicates, e.g. when attached `Executables`differ. To the dedup code here it would be two separate entries, whereas the code that uses the tuple would see it as a collision.

Would you be okay with redesigning this when we move the conversion functions into Scanner?

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: I did the conversion first with two thoughts in mind:
1. My understanding is that we only expect very rare cases of duplicates, if any at all
2. I wanted to derive the key from the converted object to improve code readability

As with the other nit: Would you be okay with revisiting this when we move the conversion into Scanner?

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: You're right that it's highly unlikely that we ever receive a `nil` here, as we would most likely have crashed before already.
Also, a very valid point what people reading this function might think. I'll remove the check here to improve readability.

That said, it's an interesting situation, as in an earlier PR review Misha suggested to move `nil` checks further down in the call stack (e.g. let the receiving function handle them).

jvdm: I'll make sure to make my point and opinion clear.

If the function's success depends on factors beyond the programmer's control, it should handle the bad scenarios and signal the caller, usually with status, to let them decide what to do. Otherwise, the function should panic as a sign of a bug in the calling code, a precondition of the function that was not met, as it should never happen in well-written programs. In other words, a bug.

In our case, there is nothing outside our control, and as such, `makeComponentKey()` should simply assume the caller will meet its preconditions and not pass `nil` as input.

If the Component struct would depend on conditions beyond the programmer's control (maybe it comes from some I/O) potentially leading to `nil` during normal program execution, then we should handle the `nil`. I agree that we should handle it down in the call stack, but down to the point where there is context to decide what to do with it. Otherwise, you are masking the `nil` with something else -- in our case, returning an empty key.

Note that sometimes the function you call can deal with `nil`. But most of the time, that should be part of the function's contract, and `nil` should be considered valid input for that function.

PR #4289, opened 2023-01-04
Dashrews/ROX-13964 central db accessible compatible solution

dashrews78:
## Description

Take 2.  Updating the object returned in `/metadata` caused compatibility issues with sensors.  Thus it seemed the best course of action was to create a new endpoint for database status and generic database information.  This can be used by the UI to populate a banner and/or add some data to the system health page if desired.  UI changes will need to be made via ROX-13963 to utilize this endpoint.

New endpoint is `database-status`.  Initially it will return this object:
```
    // indicates whether or not central can communicate with the database
    bool database_available = 1;
    // type of database serving central
    string database_type = 2;
    // version of the database
    string database_version = 3;
```
If unauthenticated, only the database_available flag will be set.  This mimics the functionality of the `/metadata` endpoint in that regard.

## Checklist

## Testing Performed

manual testing with curl on the endpoint and added UT.


proto/api/v1/metadata_service.proto
```diff
@@ -60,4 +69,10 @@ service MetadataService {
             get: "/v1/tls-challenge"
         };
     }
+
+    rpc GetDatabaseStatus(Empty) returns (DatabaseStatus) {
+        option (google.api.http) = {
+            get: "/v1/database-status"
```
c-du, line 75: How about "/v1/database/status"? It could be extended in the future.

central/metadata/service/service_impl.go
```diff
@@ -133,3 +138,31 @@ func (s *serviceImpl) TLSChallenge(ctx context.Context, req *v1.TLSChallengeRequ
...
 	return resp, nil
 }
+
+// GetDatabaseStatus returns the database status for Rox.
+func (s *serviceImpl) GetDatabaseStatus(ctx context.Context, _ *v1.Empty) (*v1.DatabaseStatus, error) {
+	dbStatus := &v1.DatabaseStatus{
+		DatabaseAvailable: true,
+	}
+
+	dbType := "RocksDB"
```
c-du, line 148: It would be better to declare an enum to keep the type stable.

central/metadata/service/service_impl.go
```diff
@@ -133,3 +138,31 @@ func (s *serviceImpl) TLSChallenge(ctx context.Context, req *v1.TLSChallengeRequ
...
+			dbStatus.DatabaseAvailable = false
+			log.Warn("central is unable to communicate with the database.")
+			return dbStatus, nil
+		}
+
+		dbVersion = globaldb.GetPostgresVersion(ctx, s.db)
+	}
+
+	// Only return the database type and version to logged in users, not anonymous users.
+	if authn.IdentityFromContextOrNil(ctx) != nil {
```
c-du, line 162: Why do we want to provide the database availability for anonymous users?
And why not expose the auth failure so the customer can fix it?

dashrews78: @c-du Because if central cannot talk to the database then UI cannot provide the auth providers on the login page.  So allowing for that flag to be returned when the user is not authenticated can let them know earlier that the database is the issue.  this function is similar to how the /metadata endpoint works.

c-du: Yes, that is right.

c-du (approved):

PR #4291, opened 2023-01-04
ROX-14181: Add traits message to proto message for declarative resources

dhaus67:
## Description

This is a follow-up from #4290 to add the traits message to all relevant proto messages for the declarative configuration feature.

In scope of the declarative configuration feature are:
- Auth Provider
- Groups
- Role
- Permission Set
- Simple Access Scope

Note that both Auth Provider and Groups had the traits message beforehand due to its usage in the ACSCS use-case.

As next steps, the traits message will be later on used to restrict certain functionality (e.g. changes to resources created declaratively) within [ROX-14182](https://issues.redhat.com/browse/ROX-14182).

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

- see CI passing (no new functionality added besides proto definition).

PR #4294, opened 2023-01-04
ROX-14181: Add default origin to default resources

dhaus67:
## Description

This is a follow-up from #4291, where the traits message has been added to all relevant resources for declarative configuration.

The origin enum within the traits message can be used to identify default resources, and all default resources created (role, permission sets, access scopes) will specify it correctly now.

Note that a migration is not required in this case, as the default resources will be upserted on each restart of central irrespective of whether they have existed previously or not.

## Checklist

## Testing Performed

- see CI.

### DIFF
--- a/central/role/datastore/datastore_impl.go
+++ b/central/role/datastore/datastore_impl.go
@@ -127,7 +127,6 @@ func (ds *dataStoreImpl) RemoveRole(ctx context.Context, name string) error {
 		return err
 	}
 
-	// Verify storage constraints
 	if err := ds.verifyRoleForDeletion(ctx, name); err != nil {
 		return err
 	}
@@ -456,7 +455,7 @@ func (ds *dataStoreImpl) verifyRoleReferencesExist(ctx context.Context, role *st
 
 // Returns errox.InvalidArgs if the given role is a default one.
 func verifyNotDefaultRole(role *storage.Role) error {
-	if rolePkg.IsDefaultRoleName(role) {
+	if rolePkg.IsDefaultRole(role) {
 		return errors.Wrapf(errox.InvalidArgs, "default role %q cannot be modified or deleted", role.GetName())
 	}
 	return nil
@@ -489,7 +488,7 @@ func (ds *dataStoreImpl) verifyPermissionSetIDDoesNotExist(ctx context.Context, 
 }
 
 // Returns errox.InvalidArgs if the given permission set is a default
-// one. Note that IsDefaultRoleName() is reused due to the name sameness.
+// one. Note that IsDefaultRole() is reused due to the name sameness.
 func verifyNotDefaultPermissionSet(permissionSet *storage.PermissionSet) error {
 	if rolePkg.IsDefaultPermissionSet(permissionSet) {
 		return errors.Wrapf(errox.InvalidArgs, "default permission set %q cannot be modified or deleted",
--- a/central/role/datastore/datastore_impl.go
+++ b/central/role/datastore/datastore_impl.go
@@ -77,7 +77,7 @@ func (ds *dataStoreImpl) AddRole(ctx context.Context, role *storage.Role) error 
 	if err := rolePkg.ValidateRole(role); err != nil {
 		return errors.Wrap(errox.InvalidArgs, err.Error())
 	}
-	if err := verifyNotDefaultRole(role.GetName()); err != nil {
+	if err := verifyNotDefaultRole(role); err != nil {
 		return err
 	}
 
@@ -103,7 +103,7 @@ func (ds *dataStoreImpl) UpdateRole(ctx context.Context, role *storage.Role) err
 	if err := rolePkg.ValidateRole(role); err != nil {
 		return errors.Wrap(errox.InvalidArgs, err.Error())
 	}
-	if err := verifyNotDefaultRole(role.GetName()); err != nil {
+	if err := verifyNotDefaultRole(role); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


dhaus67:
* master
  * **PR #4290** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4290" target="_blank">Graphite</a> 
    * **PR #4291** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4291" target="_blank">Graphite</a> 
      * **PR #4294** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4294" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4294?utm_source=stack-comment).

janisz (approved):

janisz:
Where it will be used? Can you add an assertion in e2e for this?

ivan-degtiarenko (commented):
I'm curious what is the plan for functions used to distinguish default objects like 
https://github.com/stackrox/stackrox/blob/728c7e3520e336ec1b89445611407928c8770252/central/role/default.go#L81-L90

Will they be rewritten to use `Traits` object? Perhaps we should do that in the same PR?

central/role/datastore/datastore_impl.go
```diff
@@ -103,7 +103,7 @@ func (ds *dataStoreImpl) UpdateRole(ctx context.Context, role *storage.Role) err
 	if err := rolePkg.ValidateRole(role); err != nil {
 		return errors.Wrap(errox.InvalidArgs, err.Error())
 	}
-	if err := verifyNotDefaultRole(role.GetName()); err != nil {
+	if err := verifyNotDefaultRole(role); err != nil {
```
ivan-degtiarenko, line 106: That does not work in case the user decides to modify default role and sends it without `traits`

central/role/datastore/datastore_impl.go
```diff
@@ -77,7 +77,7 @@ func (ds *dataStoreImpl) AddRole(ctx context.Context, role *storage.Role) error
...
```
ivan-degtiarenko, line 80: I wonder if there is a benefit of having this check in `AddRole` method considering adding role with default name should be impossible just because the name is occupied.

central/role/datastore/datastore_impl.go
```diff
@@ -126,14 +126,20 @@ func (ds *dataStoreImpl) RemoveRole(ctx context.Context, name string) error {
...
 	// Verify storage constraints.
 	if err := ds.verifyRoleNameExists(ctx, name); err != nil {
 		return err
 	}
 
+	role, _, err := ds.roleStorage.Get(ctx, name)
+	if err != nil {
+		return err
+	}
+	if err := verifyNotDefaultRole(role); err != nil {
```
ivan-degtiarenko, line 139: I think this check can be unified with `verifyRoleNameExists` - then we won't have to call `ds.roleStorage.Get` twice

central/role/datastore/datastore_impl.go
dhaus67: I think to be on the safe side here, I've re-added the conditions for the `IsDefault..` functions to conditionally check the name as well, covering the case the traits field isn't set.

central/role/datastore/datastore_impl.go
dhaus67: I think it's better to be explicit here, as `verifyNotDefaultRole` returns `InvalidArgs` with a specific error message that default roles cannot be added / modified / deleted, compared to already exists.

I'm also quite sure this is a breaking change in our API, since we would return 400 instead of 409, which we return for "AlreadyExists".

central/role/datastore/datastore_impl.go
ivan-degtiarenko: Good thinking! That way we keep the existing restriction + it's impossible to create a role with "default" origin

central/role/datastore/datastore_impl.go
```diff
@@ -126,11 +126,9 @@ func (ds *dataStoreImpl) RemoveRole(ctx context.Context, name string) error {
 	if err := sac.VerifyAuthzOK(roleSAC.WriteAllowed(ctx)); err != nil {
 		return err
 	}
-	if err := verifyNotDefaultRole(name); err != nil {
-		return err
-	}
-	// Verify storage constraints.
-	if err := ds.verifyRoleNameExists(ctx, name); err != nil {
+
+	// Verify storage constraints
```
ivan-degtiarenko, line 130:
```suggestion
	// Verify storage constraints and that role is not default.
```

ivan-degtiarenko, line 130: nit: not sure default role check counts as "storage constraint". Considering you have a more detailed comment next to function implementation, maybe we could delete this comment altogether.

central/role/default.go
```diff
@@ -78,15 +84,21 @@ func getAccessScopeIncludeAllID() string {
 	return EnsureValidAccessScopeID("unrestricted")
 }
 
-// IsDefaultRoleName checks if a given role name corresponds to a default role.
-func IsDefaultRoleName(name string) bool {
-	return DefaultRoleNames.Contains(name)
+// IsDefaultRoleName checks if a given role corresponds to a default role.
+func IsDefaultRoleName(role *storage.Role) bool {
+	return role.GetTraits().GetOrigin() == storage.Traits_DEFAULT || DefaultRoleNames.Contains(role.GetName())
```
ivan-degtiarenko, line 89: we should probably update `role/datastore/datastore_test` with tests around traits(or at least have a TODO linking to the ticket where it will be done)

ivan-degtiarenko (approved):
LGTM modulo two minor comments

central/role/default.go
dhaus67: I've added tests under `default_test.go` which should be enough.

PR #4295, opened 2023-01-04
ROX-14195: Resync Prefactor quick fixes for follow-up PRs

fredrb:
## Description

There are three things in this PR which are needed for on-going work on RBACs and Services:

1. Disable re-sync entirely if feature is enabled (tests *should* fail). This allows us to run integration tests against the new code without needing to change the test setup.  
2. Fix issue where not all properties were copied when merging pipeline messages
3. Fix issue where updated deployment wasn't being added to the deployment store after processed

This will reduce the conflicts and repeated work on both PRs.

## Checklist

## Testing Performed

- CI should be green. Everything is behing a feature flag
- RBAC integration tests fail with FF enabled and without any change to the tests:
```
ROX_RESYNC_DISABLED="true" LOGLEVEL="info" ./scripts/go-test.sh -v -p 1 ./sensor/tests/resource/role
```


lvalerom (approved):
LGTM

PR #4296, opened 2023-01-04
ROX-14197: Resync Prefactor RBAC integration tests

fredrb:
## Description

Adds new RBAC integration tests that test:
- Role Binding has RoleID after update
- Group based subject (e.g. `system:serviceaccounts:sensor-integration`) isn't supported by the current solution

## Checklist

## Testing Performed

- This PR is just new tests, integration tests should continue to pass without the feature flag enabled


sensor/tests/resource/role/role_test.go
```diff
@@ -69,6 +87,66 @@ func (s *RoleDependencySuite) Test_PermutationTest() {
...
+
+		testC.LastResourceState(matchBinding(binding.GetNamespace(), string(binding.GetUID())), assertBindingHasRoleId(string(role.GetUID())), "Has RoleID")
+
+		testC.GetFakeCentral().ClearReceivedBuffer()
+
+		utils.IgnoreError(deleteRole)
+		utils.IgnoreError(deleteRoleBinding)
+		utils.IgnoreError(deleteDep)
+
+		testC.GetFakeCentral().ClearReceivedBuffer()
```
lvalerom, line 125: I think this block is redundant. Unless you want to make sure that the resources are deleted before calling `ClearReceivedBuffer`, in which case all the `defer utils.IgnoreError(deleteResource)` aren't needed.

sensor/tests/resource/role/role_test.go
```diff
@@ -69,6 +87,66 @@ func (s *RoleDependencySuite) Test_PermutationTest() {
...
+		if resource.GetEvent() == nil || resource.GetEvent().GetBinding() == nil {
+			return false
+		}
+		return resource.GetEvent().GetBinding().GetId() == id && resource.GetEvent().GetBinding().GetNamespace() == namespace
+	}
+}
+
+func (s *RoleDependencySuite) Test_BindingHasNoRoleId() {
+	s.testContext.RunBare("Binding should get an update if role gets created after", func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
+		deleteDep, err := testC.ApplyFileNoObject(context.Background(), "sensor-integration", NginxDeployment)
```
lvalerom, line 101: [nit] You could use the constant `DefaultNamespace` from the helper to set the namespace.

lvalerom (approved):
Looks good, just a couple of nitpicks.

PR #4297, opened 2023-01-04
[ROX-13701] Handle deployment label parsing correctly in GetLabels API

md2119:
## Description

This PR fixes the issue with `DeploymentService` endpoint `GetLabels` that returned nil in Postgres.

In postgres, map(string, string) matches are returned as`key=value` on the field path as compared to separate match paths `{fieldpath}.keypair.key` and `{fieldpath}.keypair.value`. 

Deployment API and global search API are the only spots where map key-value pair matches are processed. The global search API handles the kv pairs correctly since it also considers the map field path.
 
## Checklist

## Testing Performed
Unit


dashrews78 (approved):
LGTM

PR #4301, opened 2023-01-05
Do not build operator in docker

janisz:
## Description

Operator docerized build is different then one in main. Instead of running commands in docker it uses `docker build`. This prevent us from using cache (modules and build). To fix this and speedup builds let's build operator locally and copy binary to final image. This should reduce build time and decrease network usage used to download dependencies.
This should be safe because in fact we were using dockerized build but tests were run without docker. 

## Testing Performed

CI


operator/Dockerfile.osci
```diff
@@ -0,0 +1,72 @@
+# Deprecated: This is used only by OSCI.
```
msugakov, line 1: Something that often gets overlooked: `Dockerfile` is not a file name, it is actually an extension, just without leading dot :-)
IDEs enable syntax highlighting (at least GoLand does so) by extension and so you get highlighting for `Dockerfile` and `osci.Dockerfile` (yes, automatically) but not for `Dockerfile.osci`.

operator/Makefile
```diff
@@ -323,17 +323,18 @@ smuggled-status-sh:
 	)
 
 # Force re-building the file to make sure the current environment is correctly reflected.
+# Deprecated: This is used only by OSCI (aka prow) and should be remove soon.
```
porridge, line 326:
```suggestion
# Deprecated: This is used only by OSCI (aka prow) and should be removed soon.
```
Is there a ticket to track the migration that we can mention here?

operator/Makefile
```diff
@@ -323,17 +323,18 @@ smuggled-status-sh:
...
 .PHONY: build/Dockerfile.gen
-build/Dockerfile.gen: Dockerfile
+build/Dockerfile.gen: Dockerfile.osci
 	mkdir -p build/
 	sed -e 's,$${ROX_IMAGE_FLAVOR},$(ROX_IMAGE_FLAVOR),g; s,$${BUILD_IMAGE_VERSION},$(shell sed 's/\s*\#.*//' ../BUILD_IMAGE_VERSION),' < $< > $@
 
 .PHONY: docker-build
-docker-build: build/Dockerfile.gen test smuggled-status-sh ## Build docker image with the operator.
-	DOCKER_BUILDKIT=1 BUILDKIT_PROGRESS=plain docker build \
+docker-build: build test smuggled-status-sh ## Build docker image with the operator.
```
porridge, line 333: Would this work?
```suggestion
docker-build: build test ## Build docker image with the operator.
```

porridge (approved):
Wow, this seems to work!
I like this a lot, since it's a lot more straightforward than all the magic we had so far.

Additionally, the build environment is now controlled consistently for all builds in a [single place](https://github.com/stackrox/stackrox/blob/master/.github/workflows/build.yaml), right?

operator/Dockerfile.osci
porridge: Good point, although not so important if this file will disappear this week I guess...

porridge:
> The advantage of doing that is that steps for which inputs did not change are cached by the `docker` itself which should lead to quick local rebuilds.

Perhaps I'm missing something but the Go toolchain already builds only the things it needs to, so local builds should already be fast without the need to introduce docker, no?

msugakov:
> so local builds should already be fast without the need to introduce docker, no?

There's way more than just `go build` in the repo-root's `Makefile` (and my original message was about that). This "stuff" makes local rebuilds take **time**. Here's a simple benchmark: I run `time make image` followed by `time make image`. I expect the second command to complete really fast because I haven't changed _anything_. Instead, it produces a lot of output, and eventually:
```
$ time make image
...
make image  46,95s user 10,32s system 38% cpu 2:29,15 total
```

Two minutes isn't too bad but I've heard folks have slower hardware and it takes ~10 minutes for them.

janisz:
> The advantage of doing that is that steps for which inputs did not change are cached by the docker itself which should lead to quick local rebuilds.

Yes and we have solved this problem by introducing `go mod download` in dockerfile. This works because `go.mod` and `go.sum` are not changed often and this layer could be cached. This approach will not work for build as we will change source code that will be builded so `go build ./...` layer will not be cached. The same is with tests, linting, code generation and any other step. 

> Running docker build in CI isn't that exotic so I don't think that's unsolved problem?

Looks like this problem is solved by not in docker but in podman. Now do we really want to introduce podman just to build a binary? In my opinion there are better build tools than docker e.g. nix (cc @stehessel :)) or bazel. 

I think this is much bigger discussion not about tools but about approach to what we expect from build. Do we need reproducible builds? Now every build have some metadata including timestamp so it's not reproducible. Every image updates packages so it will also produce different output for the same source code.

msugakov:
> > Running docker build in CI isn't that exotic so I don't think that's unsolved problem?

> Looks like this problem is solved by not in docker but in podman. Now do we really want to introduce podman just to build a binary?

Sorry, this is not what I meant. I meant that having Go cache and `docker build` in GHA shouldn't be that exotic. I assume lots of people use GHA, and this problem should be known and approaches to that should exist.
One option I see could be this.

In GitHub Workflow (pseudo-GHA-yaml):
```yaml
steps:
# Try to download Go modules cache from previous runs.
- restore-go-mod-cache-action

# Download Go modules while outside of Dockerfile context.
# This should download anything that wasn't restored from the cache.
- go mod download

# Save cache for faster rebuilds in the future.
- save-go-mod-cache-action

# Create an image just with gomod cache.
- docker build -t ci-gomod-cache-image /var/local/<path-to-gomod-cache> -f <<EOF
FROM scratch
COPY . /my-fancy-gomod-cache
EOF

# Build 
- docker build --build-arg CACHE_IMAGE=ci-gomod-cache-image -f main.Dockerfile .
```

In `stackrox/stackrox` `Dockerfile` (pseudo-code):

```Dockerfile
# Dummy stage that just gives empty gomod cache location so that local builds without arguments don't fail.
FROM ubi-minimal AS shim-gomod-cache
RUN mkdir /my-fancy-gomod-cache

# Then, in real build stage:
FROM ubi AS builder

# An argument where to take gomod cache from. If not overridden, uses dummy location from the stage above.
ARG CACHE_IMAGE=shim-gomod-cache

COPY --from=$CACHE_IMAGE /my-fancy-gomod-cache /var/local/<whereever-gomod-cache-should-be>

# everything else
```

msugakov:
Answering few more:

> Now do we really want to introduce podman just to build a binary?

I wouldn't say that `podman` is too scary. At least not for Linux users who don't have to start yet another VM for this ;-)
In the end `podman`/`buildah` must produce images fully compatible with `docker run` and understand majority of the `Dockerfile` and `docker` commands, therefore I wouldn't dramatize the complexity of the migration (yet, I know it's not a small task).

> In my opinion there are better build tools than docker e.g. nix

I believe, `nix` isn't provided and supported by the Red Hat at the moment so it's not an option.
Not sure how `bazel` applies here.

> Every image updates packages so it will also produce different output for the same source code.

Sounds like go modules cache should be stored per branch with the branch name in the cache key. For non-master branches, we should restore `master` cache first, and branch cache after (if exists).

janisz:
> Additionally, the build environment is now controlled consistently for all builds in a [single place](https://github.com/stackrox/stackrox/blob/master/.github/workflows/build.yaml), right?

Correct

janisz:
@msugakov OK, this works for `go mod` cache which we can solve with [caching layer in GHA cache](https://docs.docker.com/build/cache/backends/gha/). How to use this approach with go build cache?

janisz:
> therefore I wouldn't dramatize the complexity of the migration (yet, I know it's not a small task at all).

This is a part I'd like to avoid now. We are migrating from OSCI to GHA. If we add another migration there's a chance we will never migrate as next step after migrating to podman will be migrating to buildah and ubi-micro or even something else. 
My goal is to have GHA up and running and then we can easily iterate on it. If we start migrating to podman and it will keep us tight to OSCI for longer than it's necessary causing need to keep 2 build systems at one time.

I think it's a valid concern and I think I understand the benefits of having "Dockerized" builds but I think we should postpone it after GHA migration is finished and then prepare action plan for all our images and repositories.

msugakov:
@janisz

1. Regarding

> OK, this works for go mod cache which we can solve with caching layer in GHA cache.

_This_ for me was different than GHA cache. Have you read what I proposed?

2. GHA cache thing that you referred is available in beta.
Worse, that it requires authentication whereas storing files locally and using `actions/cache@v2` does not require any credentials.
Worse, GHA cache size limit is set to ridiculous 10GB per repo. I visited https://github.com/stackrox/stackrox/actions/caches yesterday and saw that we were using 146GB over 10GB (yes, more than a limit). Today it shows that we use 7.91 of 10. Maybe there will be again a different picture by the end of the day.
Either we need to be super cautious about what, how much and when we cache, or we need to cache things externally (yes, I mean something like GCloud, with our own code that reads and writes to it).

3. Regarding

> How to use this approach with go build cache?

First, there **is** a way. After building the container, we rebuild only the `builder` stage which completes quickly. Then create it as a container (just `docker create`, no need for `docker run`) and copy contents interesting for us from the container to the host, then save that in cache. Injecting that back into build can be done like I showed above.

Second, I tried wiping the `~/.cache/go-build`. Here's what I observed

```sh
# First: rebuild with cache
$ time make image
...
make image  50,33s user 13,08s system 39% cpu 2:41,38 total

$ rm -rf ~/.cache/go-build

$ time make image
...
make image  1436,37s user 128,81s system 388% cpu 6:42,47 total

$ du -hs ~/.cache/go-build
5,1G	/home/misha/.cache/go-build
```

The most concerning thing is 5.1GB size of the build cache. This will drain our 10GB limit quite quickly causing cache thrashing.

4. Regarding **branches** and GHA caching: I gave https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows a slow read, and the behavior described there is very sane. We don't need to do anything extra on our end.

5. Regarding `podman` vs `docker`, dockerizing or un-dockerizing builds, and the bigger picture.
IIUC, the operator build in this change [takes ~7 minutes](https://github.com/stackrox/stackrox/actions/runs/3873014892/jobs/6602536340) whereas the operator build on master [takes ~9 minutes](https://github.com/stackrox/stackrox/actions/runs/3878473796/jobs/6614678451).
I don't feel too uncomfortable about 9 minutes to give up dockerization of builds. Equally, it's not a winning argument to force us migrate to `podman`.
Please let know what you think.

6. I researched and tried both [Docker cache backends](https://docs.docker.com/build/cache/backends/) and [Dockerfile RUN cache mounts](https://docs.docker.com/engine/reference/builder/#run---mounttypecache). These are two independent things.
* Local cache backend produces cache for the resulting image layers. For the operator build I saw the cache size of 33MB. Therefore: there is only final state of layers in that cache. It may be useful in case no rebuild is needed, but certainly won't speedup Go rebuilds. That's another reason to not onboard GHA Docker layers caching.
* It is not possible to provide host filesystem location for RUN cache mounts (as well as RUN bind mounts). Therefore, it's not possible to persist the cache through this mechanism.

janisz:
Since #4277 was merged I expect operator builds to be more stable than before. Let's close this PR

PR #4305, opened 2023-01-05
Rework collection rule selectors to contain an explicit match type

dvail:
## Description

This reworks the rule selection piece of the collection form to allow (and enforce) users to choose between "exact matching" and "regex matching" for name based rules, as well as combine the label based rule inputs into a single input field.

The reasons for this are:
1. There was no way for the user to specify if they wanted an exact match versus a regex match, which was unintuitive when a user selected a single option from one of the dropdowns. e.g. If a user selected "namespace=stackrox", they would not expect the rule to match deployments in namespace "stackrox2", which is what would happen.
2. Forcing the user to enter `^` and `$` characters, or automatically adding them, was a clunky solution that applied to what most likely would be the _default_ use case of selecting an exact match.
3. Splitting the label field made regex matching difficult as per https://issues.redhat.com/browse/ROX-14063. Additionally, it was decided that regex matching for this section was not needed for MVP so the fields were combined into a single field for simplicity and the ability to support autocomplete.
4. Combining the label key and value fields allows us to support autocomplete for those fields.

## TODO

- [x] There are a handful of places that are commented out that rely on the BE support of the "match_type" field that need to be fixed once completed.
- As a follow up, update tests to account for regex/exact field semantics and verify deployment matching.

## Review notes

- Many of the changes, especially in the test files, are due to the underlying rule data structure change from `{ values: string[] }` to `{ values: { value: string, matchType: string }[] }`
- Unfortunately there were enough changes to the ByName and ByLabel selector components that the diffs in those files are fairly large, mostly due to JSX changes. IMO these files are easier to review in GitHub's "split" view.
- The ByName and ByLabel selector components are very similar, but not quite similar enough that abstracting the functionality wouldn't lead to more confusing code. Instead I abstracted what I could from the styles into CSS to minimize the duplication of inline `classNames` and PF `<Flex>` props.

## Checklist

## Testing Performed

From the create collection page, click the "All deployments" dropdown and select "deployments with names matching".
image

Enter a value for "An exact value of" and verify that the deployments match:
image
Note that this matches only "central", _exactly_. Click the "OR" button to add an addtional rule to match exactly "central-db" as well.
image

Delete the second rule and change the dropdown next to the input box to "A regex value of". Since the regex `/central/` matches both central and central-db, both of those deployments are matched:
image

A more complex regex value with wildcards and boundaries should match the correct deployments:
image

Regex and exact value matches should be able to be combined in a single rule:
image

Delete the "by name" rules, and select "Deployments with labels matching exactly" from the dropdown. This should allow you to enter a value in to get an exact label match. Note that regex matching is not supported, so `app=central` should _only_ match the `central` deployment and not `central-db`. **This currently isn't working, as the testing so far has been done against a draft version of https://github.com/stackrox/stackrox/pull/4351. These tests steps need to be re-verified once the BE work is complete and merged.**
image

### DIFF
--- a/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
+++ b/ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
@@ -23,11 +23,16 @@ import * as yup from 'yup';
 
 import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 import { Collection } from 'services/CollectionsService';
-import { getIsValidLabelKey } from 'utils/labels';
+import { getIsValidLabelKey, getIsValidLabelValue } from 'utils/labels';
+import { ensureExhaustive } from 'utils/type.utils';
 import { CollectionPageAction } from './collections.utils';
 import RuleSelector from './RuleSelector';
 import CollectionAttacher, { CollectionAttacherProps } from './CollectionAttacher';
 import {
+    byLabelMatchTypes,
+    ByLabelResourceSelector,
+    byNameMatchType,
+    ByNameResourceSelector,
     ClientCollection,
     ScopedResourceSelector,
     SelectorEntityType,
@@ -90,31 +95,80 @@ export type CollectionFormProps = {
     headerContent?: ReactElement;
 };
 
-function yupResourceSelectorObject() {
-    return yup.lazy((ruleObject) => {
-        if (ruleObject.type === 'All') {
-            return yup.object().shape({});
-        }
+function yupLabelRuleObject({ field }: ByLabelResourceSelector) {
+    return yup.object().shape({
+        field: yup.string().required().matches(new RegExp(field)),
+        rules: yup.array().of(
+            yup.object().shape({
+                operator: yup.string().required().matches(/OR/),
+                values: yup
+                    .array()
+                    .of(
+                        yup.object().shape({
+                            value: yup
+                                .string()
+                                .trim()
+                                .required()
+                                .test(
+                                    'label-value-k8s-format',
+                                    'Label values must be valid k8s labels',
+                                    (val) => {
+                                        const parts = val.split('=');
+                                        if (parts.length !== 2) {
+                                            return false;
+                                        }
+                                        const validKey = getIsValidLabelKey(parts[0]);
+                                        const validLabel = getIsValidLabelValue(parts[1]);
+                                        return validKey && validLabel;
+                                    }
+                                ),
+                            matchType: yup
+                                .string()
+                                // TODO - requires BE
+                                // .required()
+                                .matches(new RegExp(byLabelMatchTypes.join('|'))),
+                        })
+                    )
+                    .required(),
+            })
+        ),
+    });
+}
 
-        const { field } = ruleObject;
-        return typeof field === 'string' && field.endsWith('Label')
-            ? yup.object().shape({
-                  field: yup.string().required().matches(new RegExp(field)),
-                  rules: yup.array().of(
-                      yup.object().shape({
-                          operator: yup.string().required().matches(/OR/),
-                          key: yup.string().trim().required().test(getIsValidLabelKey),
-                          values: yup.array().of(yup.string().trim().required()).required(),
-                      })
-                  ),
-              })
-            : yup.object().shape({
-                  field: yup.string().required().matches(new RegExp(field)),
-                  rule: yup.object().shape({
-                      operator: yup.string().required().matches(/OR/),
-                      values: yup.array().of(yup.string().trim().required()).required(),
-                  }),
-              });
+function yupNameRuleObject({ field }: ByNameResourceSelector) {
+    return yup.object().shape({
+        field: yup.string().required().matches(new RegExp(field)),
+        rule: yup.object().shape({
+            operator: yup.string().required().matches(/OR/),
+            values: yup
+                .array()
+                .of(
+                    yup.object().shape({
+                        value: yup.string().trim().required(),
+                        matchType: yup
+                            .string()
+                            // TODO - requires BE
+                            // .required()
+                            .matches(new RegExp(byNameMatchType.join('|'))),
+                    })
+                )
+                .required(),
+        }),
+    });
+}
+
+function yupResourceSelectorObject() {
+    return yup.lazy((ruleObject: ScopedResourceSelector) => {
+        switch (ruleObject.type) {
+            case 'All':
+                return yup.object().shape({});
+            case 'ByName':
+                return yupNameRuleObject(ruleObject);
+            case 'ByLabel':
+                return yupLabelRuleObject(ruleObject);
+            default:
+                return ensureExhaustive(ruleObject);
+        }
     });
 }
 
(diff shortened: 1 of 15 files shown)


dvail:
* master
  * **PR #4251** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4251" target="_blank">Graphite</a> 
    * **PR #4305** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4305" target="_blank">Graphite</a>  👈
    * **PR #4281** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4281" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4305?utm_source=stack-comment).

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -90,31 +95,80 @@ export type CollectionFormProps = {
     headerContent?: ReactElement;
 };
 
+function yupLabelRuleObject({ field }: ByLabelResourceSelector) {
```
dvail, line 98: I'd be curious to see what a manual validation function with equivalent functionality looks like when plugged into Formik. My initial feeling here is that this is a ton of boilerplate that duplicates what we already get from TS, with a couple of small validation functions on the values.

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
```diff
@@ -27,25 +40,51 @@ function getOptions(
...
     const { isOpen, onToggle, closeSelect } = useSelectToggle();
     const [typeahead, setTypeahead] = useState(selectedOption);
+
     // When isTyping is true, autocomplete results will not be displayed. This prevents
     // a clunky UX where the dropdown results and the user text get out of sync.
     const [isTyping, setIsTyping] = useState(false);
 
+    // We need to wrap this custom SelectOption component in a forward ref
+    // because PatternFly will pass a `ref` to it
+    const OptionComponent = forwardRef(
```
dvail, line 63: Initially I wanted to keep the custom option rendering and autocomplete calls outside of this component, to make it more reusable, but the extra complexity for the (small?) chance of us needing this someplace else didn't feel worth it.

Without the rendering and the autocomplete behavior, there wasn't a ton of behavior built on top of the PatternFly defaults. We can revisit this in the future if we think we will need this elsewhere.

ui/apps/platform/src/Containers/Collections/RuleSelector/MatchTypeSelect.tsx
```diff
@@ -0,0 +1,47 @@
...
+            </Select>
+        </>
+    );
+}
+
+export function NameMatchTypeSelect(props: MatchTypeSelectProps<ByNameMatchType>) {
+    return <MatchTypeSelect {...props} />;
+}
+
+export function LabelMatchTypeSelect(props: MatchTypeSelectProps<ByLabelMatchType>) {
```
dvail, line 45: This was added when we intended to have a match type dropdown for label fields as well, and I kept it for completeness. (The label match component only supports options of `["EXACT"]`, to mirror what is expected in the API.)

ui/apps/platform/src/Containers/Collections/converter.test.ts
```diff
@@ -157,58 +180,6 @@ describe('Collection parser', () => {
 
         expect(isCollectionParseError(parseCollection(collectionResponse))).toBeTruthy();
     });
-
-    it('should correctly handle label key/value splitting on `=` delimiter', () => {
```
dvail, line 161: This is no longer relevant since we are not splitting label keys/values in the converter anymore.

dvail:
Note: this needs to be rebased once the BE is in in order for the e2e tests to pass.

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
pedrottimark: Indeed. Next best thing is `yup.ObjectSchema<Whatever>` used only in Policies so far:
* https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Policies/Wizard/policyValidationSchemas.ts
* https://github.com/jquense/yup#ensuring-a-schema-matches-an-existing-type

ui/apps/platform/src/Containers/Collections/CollectionForm.tsx
```diff
@@ -319,12 +371,7 @@ function CollectionForm({
                             </Title>
                             <Badge isRead>{ruleCount}</Badge>
                         </Flex>
-                        {!isReadOnly && (
-                            <p>
-                                Select deployments via rules. You can use regular expressions (RE2
-                                syntax).
-                            </p>
-                        )}
+                        {!isReadOnly && <p>Select deployments using names or labels</p>}
```
pedrottimark, line 374: Bravo to replace **via rules** with **using names of labels**

ui/apps/platform/src/Containers/Collections/RuleSelector/AutoCompleteSelect.tsx
pedrottimark: Good call.

ui/apps/platform/src/Containers/Collections/RuleSelector/ByLabelSelector.tsx
```diff
@@ -1,59 +1,63 @@
...
-    TextInput,
-} from '@patternfly/react-core';
+import { Label, Button, Divider, ValidatedOptions } from '@patternfly/react-core';
 import { TrashIcon } from '@patternfly/react-icons';
 import { FormikErrors } from 'formik';
 import cloneDeep from 'lodash/cloneDeep';
 
 import useIndexKey from 'hooks/useIndexKey';
-import { SelectorEntityType, ScopedResourceSelector, ByLabelResourceSelector } from '../types';
+import {
```
pedrottimark, line 8: Small request to sort these import names, because so many of the others are sorted.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
```diff
@@ -15,10 +14,25 @@ import {
...
 
+import './RuleSelector.css';
+
 function isRuleSelectorOption(value: string): value is RuleSelectorOption {
     return selectorOptions.includes(value as RuleSelectorOption);
 }
 
+const placeholders = {
+    ByName: {
+        Deployment: 'nginx-deployment',
```
pedrottimark, line 25: Was there any discussion about placeholders less ambiguous with an actual names for people who might not be at their best attention?

Like `'deployment-name'` and so on.

pedrottimark (approved):
Although hard work to adjust integration and unit tests, especially worthwhile now.

One question about placeholders.

ui/apps/platform/src/Containers/Collections/RuleSelector/RuleSelector.tsx
dvail: We went with the placeholders that were available in the official k8s documentation when available, but we can bring it up at the sync and see if other have better suggestions.

PR #4307, opened 2023-01-05
ROX-14217: Disable resync for RoleBindings and ClusterRoleBindings

fredrb:
## Description

Before fully disabling the re-sync for deployment-like resources we first need to disable it for RBAC Bindings.

We don't need to re-process bindings on _every_ Role/ClusterRole update, only if the Role wasn't in the system when the binding event was received. The approach here is to hold bindings in a `pending` map, where they are stored until the related Role is received. This way Sensor no longer needs to wait for a re-sync event to send the updated Binding event with `RoleID`.

This logic is hidden behind a feature flag.

## Checklist

## Testing Performed
- Additional unit tests added to cover this case
- Tests added in #4296 cover Bindings getting updated on Role events
- E2E tests passed with version of the code without feature flag (re-sync disabled). Results here: https://prow.ci.openshift.org/pr-history/?org=stackrox&repo=stackrox&pr=4307

### DIFF
--- a/sensor/kubernetes/listener/resources/rbac/binding_fetcher_test.go
+++ b/sensor/kubernetes/listener/resources/rbac/binding_fetcher_test.go
@@ -1,0 +1,48 @@
+package rbac
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_FetchBindingRemovesLastAppliedConfig(t *testing.T) {
+	fakeClient := fake.NewSimpleClientset()
+	_, err := fakeClient.RbacV1().ClusterRoleBindings().Create(context.Background(), &v1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-binding",
+			UID:  "test-uid",
+			Annotations: map[string]string{
+				"SomeKey": "SomeValue",
+				"kubectl.kubernetes.io/last-applied-configuration": "{\"some_prop\": \"value\"}",
+			},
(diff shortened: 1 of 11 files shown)


sensor/kubernetes/listener/resources/rbac/binding_fetcher_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_FetchBindingRemovesLastAppliedConfig(t *testing.T) {
```
lvalerom, line 14: I would maybe add another case with a `Role` just to cover both paths in `generateDependentEvent`.

lvalerom (approved):
Looks good to me, just a minor comment/suggestion.

lvalerom (changes requested):
I just realize that the test for the rbac store will fail if the feature flag is not enable. I think you need to have two versions of the tests so they don't fail without the feature flag.

fredrb:
> I just realize that the test for the rbac store will fail if the feature flag is not enable. I think you need to have two versions of the tests so they don't fail without the feature flag.

Yea I noticed this. I fixed it locally and forgot to push. Check the newest commit. I'd prefer to keep only the FF enabled version of the tests, as it's pretty much a superset of features. Hopefully we won't have to add anything new that's exclusive to the old path, so not having these tests for a short period of time seems fine. 

If you're not okay with that, I'll copy the old tests into this branch and have them skip if FF is set.

lvalerom (approved):
> I'd prefer to keep only the FF enabled version of the tests, as it's pretty much a superset of features. Hopefully we won't have to add anything new that's exclusive to the old path, so not having these tests for a short period of time seems fine.

Makes sense

> If you're not okay with that

I'm ok with that 🙂

PR #4309, opened 2023-01-05
Consolidate node datastores: part 1

md2119:
## Description

This is first of several PRs to refactor and simplify the node datastore. Nodes has multiple exported datastore interfaces:
- `central/node/globaldatastore/datastore.go`
- `central/node/datastore/datastore.go`
- `central/node/datastore/datastore.go`
 
It isn't very clear (and not necessary), and could be simplified with consolidation. For example, one can retrieve nodes for a cluster with clusterID query.

Note that the actual code change is limited to 2 files only and can be found in the first commit. The second commit is the result of running `make style`

## Checklist
- ~Unit test and regression tests added~

## Testing Performed
CI

### DIFF
--- a/central/node/service/service_impl.go
+++ b/central/node/service/service_impl.go
@@ -5,16 +5,16 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
-	"github.com/stackrox/rox/central/node/globaldatastore"
+	"github.com/stackrox/rox/central/node/datastore/dackbox/datastore"
 	"github.com/stackrox/rox/central/role/resources"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/permissions"
-	"github.com/stackrox/rox/pkg/errox"
 	pkgGRPC "github.com/stackrox/rox/pkg/grpc"
 	"github.com/stackrox/rox/pkg/grpc/authz"
 	"github.com/stackrox/rox/pkg/grpc/authz/perrpc"
 	"github.com/stackrox/rox/pkg/grpc/authz/user"
+	"github.com/stackrox/rox/pkg/search"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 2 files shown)


central/node/service/service_impl.go
```diff
@@ -69,19 +65,11 @@ func (s *nodeServiceImpl) ListNodes(ctx context.Context, req *v1.ListNodesReques
...
-		return nil, errors.Errorf("could not access per-cluster node store for cluster %q: %v", req.GetClusterId(), err)
-	}
-
-	node, err := clusterLocalStore.GetNode(req.GetNodeId())
-	if err != nil {
-		return nil, errors.Errorf("could not locate node %q in per-cluster node store for cluster %s: %v", req.GetNodeId(), req.GetClusterId(), err)
+	// Nodes have non-composite unique node ID, hence, cluster ID is not required to retrieve a node.
+	// Note that whether we provide clusterID or not, the searcher will always perform SAC check in the graph.
+	node, found, err := s.nodeDatastore.GetNode(ctx, req.GetNodeId())
+	if err != nil || !found {
```
RTann, line 71: This message may look strange when `err == nil`, but `found == false`. May be best to split the error message for this scenario.

central/node/service/service_impl.go
```diff
@@ -69,19 +65,11 @@ func (s *nodeServiceImpl) ListNodes(ctx context.Context, req *v1.ListNodesReques
...
-	clusterLocalStore, err := s.nodeStore.GetClusterNodeStore(ctx, req.GetClusterId(), false)
-	if err != nil {
-		return nil, errors.Errorf("could not access per-cluster node store for cluster %q: %v", req.GetClusterId(), err)
-	}
-
-	node, err := clusterLocalStore.GetNode(req.GetNodeId())
-	if err != nil {
-		return nil, errors.Errorf("could not locate node %q in per-cluster node store for cluster %s: %v", req.GetNodeId(), req.GetClusterId(), err)
+	// Nodes have non-composite unique node ID, hence, cluster ID is not required to retrieve a node.
+	// Note that whether we provide clusterID or not, the searcher will always perform SAC check in the graph.
```
RTann, line 69: Where would `clusterID` factor in this? Looks like this is a `Get` instead of a `Search`

RTann (approved):
Been meaning to do this since we added nodes to DackBox. Thanks for doing this!

Approving, but I have two comments to address

central/node/service/service_impl.go
theencee: The way I read this comment is that the request makes it look like you need both clusterID _and_ nodeID but apparently you don't since nodeID is unique across all clusters? Is that correct?

If so, perhaps reword this to be explicit that nodeID is unique across all clusters and that clusterID in the request is not required.

md2119: > Looks like this is a Get instead of a Search

@RTann, Searcher in `GetNode` is the implementation detail of GetNode. Reworded the comment.

> If so, perhaps reword this to be explicit that nodeID is unique across all clusters and that clusterID in the request is not required.

@theencee, to avoid any user-facing changes, I will not change the API. Although, it is clear to me that internally we do not require it. Added context to the comment.

PR #4311, opened 2023-01-06
ROX-13646: Add logs for port-forwarding in E2E tests

kovayur:
## Description

This PR adds logging of port-forwarding processes and storing them on the gcp storage among other artifacts and logs.
This is necessary to continue CI flake investigation when the port forwarding process is aborted for an unknown reason.

## Checklist

## Testing Performed
Tested on CI that logs are uploaded correctly for the jobs where they are used. 
Also checked that QA jobs are not affected by the change.

Screenshot 2023-01-11 at 13 05 14
 
image

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -11,6 +11,7 @@ source "$TEST_ROOT/scripts/ci/lib.sh"
 source "$TEST_ROOT/scripts/ci/test_state.sh"
 
 export QA_TEST_DEBUG_LOGS="/tmp/qa-tests-backend-logs"
+export PORT_FORWARD_LOGS="/tmp/port-forward-logs"
 
 # shellcheck disable=SC2120
 deploy_stackrox() {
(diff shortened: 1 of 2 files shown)


tests/e2e/lib.sh
```diff
@@ -239,8 +240,12 @@ start_port_forwards_for_test() {
     ulimit -n 65535 || true
 
     central_pod="$(kubectl -n stackrox get po -lapp=central -oname | head -n 1)"
+
+    mkdir -p "$PORT_FORWARD_LOGS"
+
     for target_port in 8080 8081 8082 8443 8444 8445 8446 8447 8448; do
-        nohup kubectl -n stackrox port-forward "${central_pod}" "$((target_port + 10000)):${target_port}" </dev/null &>/dev/null &
+        log_file="$PORT_FORWARD_LOGS/central-${target_port}.log"
+        nohup kubectl -n stackrox port-forward "${central_pod}" "$((target_port + 10000)):${target_port}" > "${log_file}" 2>&1 &
```
porridge, line 248:
```suggestion
        nohup kubectl -n stackrox port-forward "${central_pod}" "$((target_port + 10000)):${target_port}" < /dev/null > "${log_file}" 2>&1 &
```
Or was there a reason to stop redirecting stdin from /dev/null?

porridge (approved):
Nice, just one nitpick.

gavin-stackrox (approved):

PR #4316, opened 2023-01-06
Dashrews/ROX-14206 jsonpb configure unmarshaler to allow unknown fields in pkg/jsonutil

dashrews78:
## Description

ROX-14206 is part of the greater ROX-14188 to ensure that uses of jsonpb unmarshalers are configured to allow for unknown fields to be present for compatibility.  This issue was exposed recently when a field was added to the `metadata` service that subsequently broke older sensors connected to the update central.  Updated `JSONToProto` to AllowUnknownFields.  Additionally added `JSONUnmarshaler` that will return an unmarshaler configured to AllowUnknownFields that should be used as the rest of ROX-14206 progresses.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/jsonutil/conversion_test.go
+++ b/pkg/jsonutil/conversion_test.go
@@ -45,11 +45,17 @@ func TestConversion(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			verifyJSONToProtoToJSON(t, jsonify(tc.str, tc.addUnknonwField), !tc.escaped, []ConversionOption{})
+			// Unknown fields will be excluded in the result so we need to build a target JSON
+			// that will exclude those to verify the proto conversion was good.
+			inputJSON := jsonify(tc.str, tc.addUnknonwField)
+			targetJSON := jsonify(tc.str, false)
+			verifyJSONToProtoToJSON(t, inputJSON, targetJSON, !tc.escaped, []ConversionOption{})
 		})
 
 		t.Run(tc.desc+" in compact JSON", func(t *testing.T) {
-			verifyJSONToProtoToJSON(t, jsonifyCompact(tc.str, tc.addUnknonwField), !tc.escaped, []ConversionOption{OptCompact})
+			inputJSON := jsonifyCompact(tc.str, tc.addUnknonwField)
+			targetJSON := jsonifyCompact(tc.str, false)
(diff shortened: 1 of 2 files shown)


pkg/jsonutil/conversion_test.go
```diff
@@ -99,5 +131,5 @@ func verifyJSONToProtoToJSON(t *testing.T, inputJSON string, shouldPreserve bool
 	// Conversion to only JSON escapes characters if specified in options.
 	convertedJSON, err := ProtoToJSON(&proto, options...)
 	assert.NoError(t, err)
-	assert.Equalf(t, shouldPreserve, inputJSON == convertedJSON, "original JSON:\n%s\nconvertedJSON:\n%s", inputJSON, convertedJSON)
+	assert.Equalf(t, shouldPreserve, targetJSON == convertedJSON, "original JSON:\n%s\ntargetJSON:\n%s\nconvertedJSON:\n%s", inputJSON, targetJSON, convertedJSON)
```
msugakov, line 134: The whole `addUnknownField` thing adds great deal of uncertainty here: is `targetJSON != convertedJSON` because of escaping special characters or did it have something to do with extra `unknownfield` attribute?

In fact, the effect of `unknownfield` attribute with the change is just that `JSONToProto()` call doesn't return error.

One can argue that the tests are still understandable (if you spend enough time reading), but I would object that it becomes harder to maintain these tests as they now try checking two things at once.

I suggest undoing all current changes to this `pkg/jsonutil/conversion_test.go` file and introducing a new simple and boring test (I have not tried compiling this):

```go
func TestNoErrorOnUnknownAttribute(t *testing.T) {
	const json = `{ "id": "6500", "unknownField": "junk" }`
	var proto v1.ResourceByID

	err := JSONToProto(json, &proto)

	assert.NoError(t, err)
	assert.Equal(t, "6500", proto.GetID())
}
```

msugakov (changes requested):
Thanks for taking initiative in fixing this!
I'm a bit concerned by the current changes to unit tests.

pkg/jsonutil/conversion_test.go
dashrews78: This approach was actually my first thought but I clearly decided to over complicate things and try to fit it in the existing tests.  I will update accordingly.

fredrb (approved):
Thanks @dashrews78!

PR #3768, opened 2022-11-10
[DO NOT MERGE] ROX-13246, ROX-13247, ROX-13248: add Clair v4 support

RTann:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #4318, opened 2023-01-06
add retries to Clair v4 image scanning HTTP requests.

RTann:
## Checklist

## Testing Performed

Tested with a live Clair v4 instance.


pkg/scanners/clairv4/clairv4.go
```diff
@@ -199,37 +205,38 @@ func (c *clairv4) index(manifest *claircore.Manifest) error {
...
-		return err
-	}
-
-	resp, err := c.client.Do(req)
-	if err != nil {
-		return err
-	}
-	defer utils.IgnoreError(resp.Body.Close)
+	return retry.WithRetry(func() error {
+		// Make a new request per retry to ensure the body is fully populated for each retry.
```
dcaravel, line 209: Dug into this a bit, nice catch - the need to re-create a request seems as if it would be a common cause for bugs. Interestingly I see that client.Do will ['reset'](https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/net/http/client.go;l=662-669;drc=1c65b69bd1dbc930c6246877f6c21c81f2a60d55;bpv=1;bpt=1) the body reader for certain types, but it only does that when following redirects, not errors.

dcaravel (commented):
Nice addition to `pkg/retry/retry.go`, I see other areas of stackrox using https://github.com/cenkalti/backoff - what led to the decision to add the backoff logic to the `retry`?

pkg/scanners/clairv4/clairv4.go
RTann: I only caught this because I ran into this error once when testing and looked it up 😂

RTann:
> Nice addition to `pkg/retry/retry.go`, I see other areas of stackrox using https://github.com/cenkalti/backoff - what led to the decision to add the backoff logic to the `retry`?

Honestly did this months ago https://github.com/stackrox/stackrox/pull/3768 and have been splitting it into reviewable and testable PRs since then, so I can't recall the exact reason. I think the API interface was simpler with this. Plus I saw other parts of the code which used `retry` and added its own implementation of backoff as an "in between" task, so I figured it'd be nice to add this as an option for this library.

dcaravel (approved):

pkg/scanners/clairv4/clairv4.go
```diff
@@ -199,37 +205,38 @@ func (c *clairv4) index(manifest *claircore.Manifest) error {
...
+		if err := json.NewDecoder(resp.Body).Decode(&ir); err != nil {
+			return err
+		}
+		if !ir.Success && ir.Err != "" {
+			return errors.Errorf("indexing error: %s", ir.Err)
+		}
 
-	return nil
+		return nil
+	}, retry.Tries(3), retry.WithExponentialBackoff(), retry.OnlyRetryableErrors())
```
md2119, line 239: pull retry count as global const

md2119 (approved):

PR #4321, opened 2023-01-09
ROX-14143,ROX-14144: Add declarative configuration settings to Helm chart.

dhaus67:
## Description

This PR adds the required settings to the Helm chart to support adding abriratry amounts of config maps / secrets that hold declarative configurations.

New settings have been introduced that will allow specifying the config maps / secrets:
```yaml
declarativeConfiguration:
  mounts:
    configMaps: []
    secrets: []
```

Each mount will have:
- the "optional" flag set, thus central pod will be able to start without them being present ([k8s doc for reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps))
- the "read only" flag set, thus central won't be able to change the contents of the config map ([k8s doc for reference](https://kubernetes.io/docs/concepts/storage/volumes/#csi))

The reason for supporting both config maps and secrets as means of specifying declarative configuration is due to the fact that some configurations include secret data (i.e. the auth provider), hence we do not want to force users to specify their sensitive data in config maps.

## Checklist

## Testing Performed

- see CI, added unit tests that cover the Helm chart.


dhaus67:
* master
  * **PR #4321** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4321" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4321?utm_source=stack-comment).

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -233,4 +239,10 @@ spec:
       - name: {{ $extraMount.name }}
         {{- $extraMount.source | toYaml | nindent 8 }}
       {{- end }}
+      {{- range $mount := (default list ._rox.central.declarativeConfiguration.mounts) }}
+      - name: {{ $mount }}
+        configMap:
+          name: {{ $mount }}
+          optional: true
```
ivan-degtiarenko, line 246: Why do we make this mount optional - do we expect it to be deleted? 

What will happen if the resource contained in the mount is upserted within Postgres, but after the restart, the mount is not existing anymore? Would the resource be deleted from Postgres? 

Also, what if a declarative resource from one mount references a declarative resource from another mount?

pkg/helm/charts/tests/centralservices/testdata/helmtest/declarative-configuration.test.yaml
```diff
@@ -0,0 +1,49 @@
...
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-1")
+     | assertThat(. != null)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-1")
+     | assertThat(.configMap.optional == true)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-2")
+     | assertThat(. != null)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-2")
+     | assertThat(.configMap.optional == true)
+
+- name: "Single value set for declarative configuration mound should lead to single additional volume mount"
```
ivan-degtiarenko, line 33: How does this test differ from multiple values test?

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
dhaus67: >Why do we make this mount optional - do we expect it to be deleted?

The reason to make this mount optional is that, in theory the config map could be deleted / misconfigured etc., but most importantly, we do not want central to _not_ come up. If we wouldn't set `optional==true`, the central pod wouldn't come up without the config map being there. Since declarative config isn't _required_ for a central to become "ready", I've decided to make it optional (we use this pattern across all mounts that are not required for central startup as well, I do like to be quite conservative here about what is required for a central boot).

 >What will happen if the resource contained in the mount is upserted within Postgres, but after the restart, the mount is not existing anymore? Would the resource be deleted from Postgres?

I think these questions are more or less jumping ahead in terms of the context of this PR and are rather part of [ROX-14147](https://issues.redhat.com/browse/ROX-14147). Without looking at the actual implementation, the expectation would be exactly what you stated: in case the declarative config has been removed, the resources should be removed as well (simply put, we want a reconciliation on the contents).

>Also, what if a declarative resource from one mount references a declarative resource from another mount?

I think that's an interesting question, and definitely a possibility. I suppose, again, this should be part of [ROX-14147](https://issues.redhat.com/browse/ROX-14147). The reconciliation should first read all contents of the mounts, and then "transform" the configuration to proto resources. There we should take care of resolving these references.

Note that the references part is still somewhat missing in JIRA, I'll be adding this shortly. Those are definitely cases we have to cover.

pkg/helm/charts/tests/centralservices/testdata/helmtest/declarative-configuration.test.yaml
dhaus67: I suppose not that much besides ensuring we can handle single and multiple values. If you think it's too much, I can remove, however I don't think there's much overhead of maintaining this test.

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -172,7 +172,13 @@ spec:
         - name: {{ $extraMount.name }}
           {{- $extraMount.mount | toYaml | nindent 10 }}
         {{- end }}
-        {{- range $mount := (default list ._rox.central.declarativeConfiguration.mounts) }}
+        {{- range $mount := (default list ._rox.central.declarativeConfiguration.mounts.configMaps) }}
+        - name: {{ $mount }}
+          mountPath: /run/secrets/stackrox.io/declarative-configuration/
```
ivan-degtiarenko: I'm just now realising that using `/run/secrets/...` path is not entirely correct considering that most of declarative configuration has nothing to do with secrets

pkg/helm/charts/tests/centralservices/testdata/helmtest/declarative-configuration.test.yaml
```diff
@@ -0,0 +1,140 @@
...
+     | assertThat(.secret.optional == true)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-3")
+     | assertThat(. != null)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-3")
+     | assertThat(.configMap != null)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-3")
+     | assertThat(.configMap.optional == true)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-4")
+     | assertThat(. != null)
+    .deployments.central.spec.template.spec.volumes[] | select(.name == "some-declarative-config-4")
```
ivan-degtiarenko, line 137: We could also check `secret.secretName` and `configMap.name` here

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
dhaus67: Yeah, its unfortunate that neither `config` nor `secrets` really matches and we probably have to use the non-default path for mounts in central.

I could switch it to `/run/stackrox.io/declartive-config` instead. WDYT?

ivan-degtiarenko (approved):
LGTM modulo switching mount path to `/run/stackrox.io/declartive-config`

PR #4324, opened 2023-01-09
ROX-13634 resolve deployments by labels

lvalerom:
## Description

Resolving deployment IDs based on labels. This enables us to resolve deployment dependencies on services/network policies updates.

This was part of #4015 

Depends on #4323

## Checklist

## Testing Performed

* Test the resources package locally
* Run the newly added benchmark `go test -bench=. -count=2 -run=^# github.com/stackrox/rox/sensor/kubernetes/listener/resources -test.bench FindDeploymentIDsByLabels`

### DIFF
--- a/sensor/kubernetes/listener/resources/deployment_store.go
+++ b/sensor/kubernetes/listener/resources/deployment_store.go
@@ -160,7 +160,7 @@ func (ds *DeploymentStore) FindDeploymentIDsByLabels(namespace string, sel selec
 			continue
 		}
 
-		if sel.Matches(selector.CreateLabelsWithLen(wrap.PodLabels)) {
+		if sel.Matches(selector.CreateLabelsWithLen(wrap.GetPodLabels())) {
 			resIDs = append(resIDs, id)
 		}
 	}
--- a/sensor/kubernetes/listener/resources/deployment_store.go
+++ b/sensor/kubernetes/listener/resources/deployment_store.go
@@ -145,6 +145,28 @@ func (ds *DeploymentStore) FindDeploymentIDsWithServiceAccount(namespace, sa str
 	return match
 }
 
+// FindDeploymentIDsByLabels returns a slice of deployments based on matching namespace and labels
+func (ds *DeploymentStore) FindDeploymentIDsByLabels(namespace string, sel selector.Selector) (resIDs []string) {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
+	ids, found := ds.deploymentIDs[namespace]
+	if !found || ids == nil {
+		return
+	}
+
+	for id := range ids {
+		wrap, found := ds.deployments[id]
+		if !found || wrap == nil {
+			continue
+		}
+
+		if sel.Matches(selector.CreateLabelsWithLen(wrap.PodLabels)) {
+			resIDs = append(resIDs, id)
+		}
+	}
+	return
+}
+
 func (ds *DeploymentStore) getWrap(id string) *deploymentWrap {
 	ds.lock.RLock()
 	defer ds.lock.RUnlock()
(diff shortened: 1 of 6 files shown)


sensor/kubernetes/listener/resources/deployment_store_test.go
```diff
@@ -145,6 +146,90 @@ func (s *deploymentStoreSuite) Test_BuildDeploymentWithDependencies_NoDeployment
...
+			expectedIDs: nil,
+		},
+		"Match": {
+			namespace: "test-ns",
+			labels: map[string]string{
+				"app": "nginx",
+			},
+			expectedIDs: []string{"uuid-2"},
+		},
+		"Labels do not match": {
```
fredrb, line 191: This test looks the same as the one above. Also the name says `"Labels do not match"` but the test expects one match?

sensor/kubernetes/listener/resources/deployment_store_test.go
```diff
@@ -145,6 +146,90 @@ func (s *deploymentStoreSuite) Test_BuildDeploymentWithDependencies_NoDeployment
...
+			namespace: "test-ns",
+			labels: map[string]string{
+				"app": "nginx",
+			},
+			expectedIDs: []string{"uuid-2"},
+		},
+		"Namespace do not match": {
```
fredrb, line 198: Same as above. Content of the test looks the same.

sensor/kubernetes/listener/resources/deployment_store_test.go
```diff
@@ -145,6 +146,90 @@ func (s *deploymentStoreSuite) Test_BuildDeploymentWithDependencies_NoDeployment
...
+			expectedIDs: nil,
+		},
+	}
+	for testName, c := range cases {
+		s.Run(testName, func() {
+			ids := s.deploymentStore.FindDeploymentIDsByLabels(c.namespace, selector.CreateSelector(c.labels))
+			s.Equal(len(c.expectedIDs), len(ids))
+			sort.Strings(ids)
+			sort.Strings(c.expectedIDs)
+			s.Equal(c.expectedIDs, ids)
```
fredrb, line 228: `s.Equal` on slices will assert their order as well. You should use `s.ElementsMatch` instead.

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -145,6 +145,28 @@ func (ds *DeploymentStore) FindDeploymentIDsWithServiceAccount(namespace, sa str
...
+		return
+	}
+
+	for id := range ids {
+		wrap, found := ds.deployments[id]
+		if !found || wrap == nil {
+			continue
+		}
+
+		if sel.Matches(selector.CreateLabelsWithLen(wrap.PodLabels)) {
```
fredrb, line 163:
```suggestion
		if sel.Matches(selector.CreateLabelsWithLen(wrap.GetPodLabels())) {
```
As a general rule you should always use protobuf getter methods rather than the property directly.

fredrb (changes requested):
Could you check if there is something wrong with the test cases?

sensor/kubernetes/listener/resources/deployment_store_test.go
lvalerom: I think I messed up with the copy-paste 😓

sensor/kubernetes/listener/resources/deployment_store_test.go
lvalerom: Same as above.

sensor/kubernetes/listener/resources/deployment_store_test.go
lvalerom: Done

sensor/kubernetes/listener/resources/deployment_store.go
lvalerom: Done

fredrb (approved):

PR #4326, opened 2023-01-09
ROX-13912: Migrate GO PG unit tests

janisz:


msugakov:
It looks like this is still draft. I took the liberty to convert it to Draft.

janisz:
@msugakov I accidentally removed `make test` when removing junit reports https://github.com/stackrox/stackrox/compare/d8942535ab5e5f848d00162599f38fb308b29fd6..a5ae36e316a1475c4d3ee69e2006577293cdecec#diff-6e608e02c595d53ab6b70822a2bf19abcfc6ddcc976c2f536ad5bfca20f0443fL50

rhybrillou (approved):
Provided the tests pass

.github/workflows/unit-tests.yaml
```diff
@@ -12,6 +12,39 @@ on:
...
+  go-postgres:
+    strategy:
+      matrix:
+        gotags: [ 'GOTAGS=""', 'GOTAGS=release' ]
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.49
+    services:
+      postgres:
+        image: ghcr.io/enterprisedb/postgresql:15
```
msugakov, line 24: Is there any chance to find quay.io or registry.access.redhat.com alternative?

msugakov: Quick search revealed these:
- https://quay.io/repository/fedora/postgresql-13?tab=info
- https://quay.io/repository/openshift-community-operators/postgresql?tab=info - would require installing the operator on OpenShift in order to figure out where the actual postgres container comes from.

msugakov: The most reliable would be to create RH image pull credentials and use this https://catalog.redhat.com/software/containers/rhel8/postgresql-13/5ffdbdef73a65398111b8362?container-tabs=gti

.github/workflows/unit-tests.yaml
```diff
@@ -28,9 +61,3 @@ jobs:
 
     - name: UI Unit Tests
       run: make ui-test
-
-    - name: Publish Test Report
-      uses: mikepenz/action-junit-report@v3
-      if: always()
-      with:
-        report_paths: 'ui/test-results/reports/*.xml'
```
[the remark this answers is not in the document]
janisz: https://issues.redhat.com/browse/ROX-14330

.github/workflows/unit-tests.yaml
janisz: > passing unit tests may not catch problems with real image

So we should use `central-db` image. I'd try to make it work as it has some expectations about environment (e.g. mounted configuration file)

janisz: I give up, let's copy approach from osci and maybe revisit usage of services later.

janisz:
@msugakov PTAL

.github/workflows/unit-tests.yaml
```diff
@@ -12,6 +12,45 @@ on:
...
+#        image: quay.io/stackrox-io/central-db:3.73.x-369-gee573ecd8d #TODO: Change to latest or current tag
+#        env:
+#          PGDATA: /tmp/data
+#          POSTGRES_PASSWORD: postgres
+#        options: >-
+#          --entrypoint docker-entrypoint.sh
+#          --health-cmd pg_isready
+#          --health-interval 10s
+#          --health-timeout 5s
+#          --health-retries 5
```
msugakov, line 33: I'd suggest to remove this dead code. If you'd like to resurrect the `service` approach some day, maybe you can add a TODO comment instead.
```suggestion
```

msugakov (approved):
Doesn't look bad at all. Please ship after removing commented-out code.

PR #4210, opened 2022-12-20
ROX-10003: Adjust ...Id naming to ...ID for style compliance

rhybrillou:
## Description

Pieces of the postgres generated store code do not comply with style requirements.
The goal of this change is to ensure generated postgres store code match style requirements.

## Checklist
- [ ] Investigated and inspected CI test results
~~- [ ] Unit test and regression tests added~~
~~- [ ] Evaluated and added CHANGELOG entry if required~~
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

If any of these don't apply, please comment below.

## Testing Performed

CI is sufficient.

PR #3451, opened 2022-10-17
Use generics in RunGetManyQueryForSchema

janisz:
## Description

After switching to Go 1.18 we can use generics. This will allow us to not parse output from database to `[][]byte` and then unmarshal it to `[]storage.*` but immediately parse to `[]storage.*`. So no intermediate slice will be created and we can save up some memory and allocations. 

Below tests shows comparison with previous commit not master.

```
name                   old time/op    new time/op    delta
GetMany/1_alerts-8        161µs ±11%     157µs ± 4%     ~     (p=0.065 n=19+19)
GetMany/2_alerts-8        196µs ± 8%     196µs ± 4%     ~     (p=0.708 n=19+20)
GetMany/4_alerts-8        238µs ± 7%     237µs ± 5%     ~     (p=0.687 n=19+20)
GetMany/8_alerts-8        338µs ± 7%     310µs ± 8%   -8.16%  (p=0.000 n=19+17)
GetMany/16_alerts-8       559µs ± 4%     470µs ± 6%  -15.79%  (p=0.000 n=18+20)
GetMany/32_alerts-8      1.00ms ±14%    0.79ms ± 4%  -21.21%  (p=0.000 n=19+20)
GetMany/64_alerts-8      2.10ms ±14%    1.71ms ± 5%  -18.51%  (p=0.000 n=17+19)
GetMany/128_alerts-8     4.57ms ±62%    2.93ms ± 7%  -35.88%  (p=0.000 n=20+20)
GetMany/256_alerts-8     7.66ms ± 8%    5.28ms ± 5%  -30.99%  (p=0.000 n=19+20)
GetMany/512_alerts-8     14.2ms ± 9%    10.1ms ± 4%  -28.77%  (p=0.000 n=20+20)
GetMany/1024_alerts-8    27.2ms ±15%    19.1ms ± 4%  -29.99%  (p=0.000 n=19+18)
GetMany/2048_alerts-8    51.9ms ± 5%    37.4ms ± 4%  -27.95%  (p=0.000 n=18+19)

name                   old alloc/op   new alloc/op   delta
GetMany/1_alerts-8       10.4kB ± 0%    10.4kB ± 0%   -0.16%  (p=0.000 n=20+20)
GetMany/2_alerts-8       18.3kB ± 0%    18.3kB ± 0%   -0.39%  (p=0.000 n=20+20)
GetMany/4_alerts-8       34.1kB ± 0%    33.9kB ± 0%   -0.54%  (p=0.000 n=18+20)
[...]

PR #4328, opened 2023-01-09
Process Listening On Port: store

erthalion:
Implement datastore for Process Listening On Port objects.

## Checklist

There are no breaking changes, no update steps (the PR is about handling a type
of data) and no user facing changes yet (they come in the follow up PRs).

## Testing Performed

Local testing, unit-tests, verifying previous test CI pipelines (e.g. #4182).
As nothing is using the datastore yet, no E2E testing is present yet.


central/processindicator/datastore/singleton.go
```diff
@@ -33,20 +34,24 @@ var (
...
+	var plopStorage plopStore.Store
 	var indexer index.Indexer
 	if env.PostgresDatastoreEnabled.BooleanSetting() {
 		storage = postgres.New(globaldb.GetPostgres())
+		plopStorage = plopStore.New(globaldb.GetPostgres())
 		indexer = postgres.NewIndexer(globaldb.GetPostgres())
 	} else {
 		storage = rocksdb.New(globaldb.GetRocksDB())
+		// PLOP storage is only supported for PostgreSQL
+		// plopStorage = rocksdb.New(globaldb.GetRocksDB())
```
dashrews78, line 46: remove this line (or both)

central/processlisteningonport/datastore/datastore.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stackrox/rox/generated/storage"
+)
+
+// WalkFn is a convenient type alias to use for the Walk function
+type WalkFn = func(plop *storage.ProcessListeningOnPortStorage) error
+
+// DataStore interface for ProcessListeningOnPort object interaction with the database
+//
+//go:generate mockgen-wrapper
+type DataStore interface {
```
dashrews78, line 17: nit:  We should probably be consistent at least for the exported interface routines.
`AddProcessListeningOnPort`
`GetProcessListeningOnPort`
`RemovePLOP`

IMO we should either shorten them all or spell them all out.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+		indicatorID := ""
+		var processInfo *storage.ProcessIndicatorUniqueKey
+
+		key := getProcesUniqueKey(val)
+
+		if indicator, ok := indicatorsMap[key]; ok {
+			indicatorID = indicator.GetId()
+			log.Debugf("Got indicator %s: %+v", indicatorID, indicator)
+		} else {
+			// XXX: Create a metric for this
```
dashrews78, line 89: nit:  should reference the Jira for adding such metric

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+			log.Warnf("Found no matching indicators for %s", key)
+			processInfo = val.Process
+		}
+
+		plopKey := fmt.Sprintf("%d %d %s",
```
dashrews78, line 94: nit:  This likely doesn't matter at all, but if we were to ever look at these keys visually it might be better to have a delimiter other than a space.  a `-` or `_` would be more obvious that it is a multiple items mashed together into 1 than a space does.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+			val.GetProtocol(),
+			val.GetPort(),
+			indicatorID,
+		)
+
+		existingPLOP, prevExists := existingPLOPMap[plopKey]
+
+		// There are three options:
+		// * We found an existing PLOP object with different close timestamp.
```
dashrews78, line 103: what if the new close timestamp is before the existing close timestamp?  I don't think we guarantee these things arrive in order.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+		if !prevExists {
+			if val.CloseTimestamp != nil {
+				// We try to close a not existing Endpoint, something is wrong
+				log.Warnf("Found no matching PLOP to close for %s", key)
+			}
+
+			newPLOP := &storage.ProcessListeningOnPortStorage{
+				// XXX, ResignatingFacepalm: Use regular GENERATE ALWAYS AS
+				// IDENTITY, which would require changes in store generator
+				Id:                 uuid.NewV4().String(),
```
dashrews78, line 127: we've probably talked around the edges of this before, but what do you want the `Id` to be?  I understand not wanting to fix the generator now and I get not wanting to make it a custom store now.  But we need to consider what a future migration of the primary key would be if we were to change its type.  It is possible to likely that migration may be just as cumbersome to deal with as taking on the task of making the PK what you want it to be now.  Something to consider.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+
+	return nil
+}
+
+// fetchExistingPLOPs: Query already existing PLOP objects belonging to the
+// specified process indicators.
+//
+// XXX: This function queries all PLOP, no matter if they are matching port +
+// protocol we've got or not. This means potentially dangerous corner cases
+// when one process listenst to a huge number of ports. To address it we could
```
dashrews78, line 208: nit:  typo `listenst`

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+	var existingPLOPMap = map[string]*storage.ProcessListeningOnPortStorage{}
+
+	if len(indicatorIds) == 0 {
+		return nil, nil
+	}
+
+	// If no corresponding processes found, we can't verify if the PLOP
+	// object is opening/closing an existing one. Collect existingPLOPMap
+	// only if there are some matching indicators.
+	existingPLOP, err := ds.storage.GetByQuery(ctx, search.NewQueryBuilder().
```
dashrews78, line 227: recommend `existingPLOPs`

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+		AddStrings(search.ProcessID, indicatorIds...).ProtoQuery())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, val := range existingPLOP {
+		key := fmt.Sprintf("%d %d %s",
```
dashrews78, line 234: same comment as earlier on the spaces.  Additionally we should probably create a helper function to create the key.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+			val.GetProtocol(),
+			val.GetPort(),
+			val.GetProcessIndicatorId(),
+		)
+
+		// A bit of paranoia is always good
+		if old, ok := existingPLOPMap[key]; ok {
```
dashrews78, line 241: Are they returned in order, so we know we want to keep the most recent one in the list?

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+
+	indicatorsQuery := search.DisjunctionQuery(indicatorLookups...)
+	log.Debugf("Sending query: %s", indicatorsQuery.String())
+	indicators, err := ds.indicatorDataStore.SearchRawProcessIndicators(ctx, indicatorsQuery)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, val := range indicators {
+		key := fmt.Sprintf("%s %s %s %s %s",
```
dashrews78, line 290: same key comment on spaces

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+	closed []*storage.ProcessListeningOnPortFromSensor
+}
+
+// normalizePLOPs
+//
+// In the batch of PLOP events there could be many open & close events for the
+// same combination of port, protocol, process. Find and squash them into a
+// single event.
+//
+// Open/close state will be calculated from the totall number of open/close
```
dashrews78, line 323: typo `totall`

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+// seems to be more preferrable.
+func normalizePLOPs(
+	plops []*storage.ProcessListeningOnPortFromSensor,
+) []*storage.ProcessListeningOnPortFromSensor {
+
+	normalizedMap := map[string]OpenClosedPLOPs{}
+	normalizedResult := []*storage.ProcessListeningOnPortFromSensor{}
+
+	for _, val := range plops {
+		key := fmt.Sprintf("%d %d %s",
```
dashrews78, line 341: same key comment on spaces

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+	}
+
+	for _, value := range normalizedMap {
+		nOpen := len(value.open)
+		nClosed := len(value.closed)
+
+		// Take the last open PLOP if there are more open in total, otherwise
+		// the last closed PLOP
+		if nOpen > nClosed {
```
dashrews78, line 380: I might be missing something.  How does this handle the case of PLOPs that are out of order?  This seems to return the last one you processed which would assume that `plops` was passed in ordered.  May not matter so much for open, but could pose issues with closed and the closedTimestamp.

central/processlisteningonport/store/postgres/index.go
```diff
@@ -0,0 +1,70 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	storage "github.com/stackrox/rox/generated/storage"
+	ops "github.com/stackrox/rox/pkg/metrics"
+	search "github.com/stackrox/rox/pkg/search"
+	"github.com/stackrox/rox/pkg/search/blevesearch"
+	"github.com/stackrox/rox/pkg/search/postgres"
+	"github.com/stackrox/rox/pkg/search/postgres/mapping"
+)
+
+func init() {
```
dashrews78, line 19: #4350 moves this to `pkg/postgres/schema`.  So watch for when that one is merged and adjust accordingly.

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,748 @@
+// Code generated by pg-bindings generator.
```
dashrews78, line 1: nit:  `// Code originally generated by pg-bindings generator.`

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,748 @@
...
+		"WHERE proc.deploymentid = $1 AND plop.closed = false"
+)
+
+var (
+	log            = logging.LoggerForModule()
+	schema         = pkgSchema.ProcessListeningOnPortsSchema
+	targetResource = resources.DeploymentExtension
+)
+
+type Store interface {
```
dashrews78, line 69: nit:  #4210 reordered the routines into a logical order.  Up to you if you want to mimic that.

central/pruning/pruning_test.go
```diff
@@ -567,7 +567,7 @@ func (s *PruningTestSuite) TestImagePruning() {
 			alerts, config, images, deployments, pods, indexQ := s.generateImageDataStructures(ctx)
```
dashrews78, line 567: Need to add a test for PLOP pruning.

pkg/postgres/schema/process_indicators.go
```diff
@@ -1,4 +1,7 @@
-// Code generated by pg-bindings generator. DO NOT EDIT.
+// Code generated by pg-bindings generator.
```
dashrews78, line 1: nit:  `// Code originally generated by pg-bindings generator.`

pkg/postgres/schema/process_indicators.go
```diff
@@ -45,16 +48,20 @@ const (
 )
 
 // ProcessIndicators holds the Gorm model for Postgres table `process_indicators`.
+// XXX: unique_indicator index is specified manually, as it has to cover
```
dashrews78, line 51: If you have the Jira ticket to do this later, reference it.

central/processlisteningonport/datastore/datastore_impl.go
erthalion: My proposal was described in ROX-13793, to summarize the idea is to let the database to generate the id value. I know the migrations could be cumbersome in the future, but this one ideally should be easy, as it will be not a data migration (there still will be uuid), but only schema migration (with an extra logic how to generate uuid). Does it answer your concerns?

central/processlisteningonport/store/postgres/index.go
erthalion: Already done, it was preventing the pipeline runs.

central/pruning/pruning_test.go
```diff
@@ -1307,6 +1308,100 @@ func (s *PruningTestSuite) TestRemoveOrphanedProcesses() {
...
+						ProcessName:         "test_process1",
+						ProcessArgs:         "test_arguments1",
+						ProcessExecFilePath: "test_path1",
+					},
+				},
+			},
+			expectedDeletions: []string{},
+		},
+		{
+			name: "Plop is closed but not expried so it is not removed",
```
rhybrillou, line 1341: nit: typo in `expired` ?

pkg/postgres/schema/process_indicators.go
```diff
@@ -47,16 +50,20 @@ const (
...
+// XXX: ROX-13790 unique_indicator index is specified manually, as it has to
+// cover multiple columns, which is not supported by the store generator.
+//
+// Note, that it will introduce some overhead even with PLOP feature disabled.
 type ProcessIndicators struct {
 	Id                 string `gorm:"column:id;type:uuid;primaryKey"`
 	DeploymentId       string `gorm:"column:deploymentid;type:uuid;index:processindicators_deploymentid,type:hash"`
-	ContainerName      string `gorm:"column:containername;type:varchar"`
-	PodId              string `gorm:"column:podid;type:varchar"`
+	ContainerName      string `gorm:"column:containername;type:varchar;index:unique_indicator"`
```
rhybrillou, line 60: Could the index name be prefixed similarly to the `processindicators_poduid` one ?

central/processindicator/datastore/singleton.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore.go
JoukoVirtanen: RemovePLOP was renamed to RemoveProcessListeningOnPort.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Changed to "TODO ROX-14377:"

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: We now have various functions to get the various keys and all keys are obtained by calling those functions. All of them use "_" as the delimiter.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: We now sort the existing plops by CloseTimestamp. The following function has been added

```
func sortByCloseTimestamp(values []*storage.ProcessListeningOnPortFromSensor) {
	sort.Slice(values, func(i, j int) bool {
		return values[i].CloseTimestamp.Compare(values[j].CloseTimestamp) == -1
	})
}
```

We also have a test for plops being out of order in central/processlisteningonport/datastore/singleton.go. That test is TestPLOPAddCloseBatchOutOfOrder.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Good catch.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: See above response to similar comment.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: See above response to similar comment.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: See above response to similar comment.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: See above response to similar comment.

central/pruning/pruning_test.go
JoukoVirtanen: We have added three tests for PLOP pruning.

Test that open PLOP is not pruned.
Test that closed PLOP which is not expired is not pruned.
Test that closed PLOP which is expired is pruned.

pkg/postgres/schema/process_indicators.go
JoukoVirtanen: Making this change causes style-check to run on this file, and that reports style errors for automatically generated code.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+			// XXX: Create a metric for this
+			log.Warnf("Found no matching indicators for %s", key)
+			processInfo = val.Process
+		}
+
+		plopKey := fmt.Sprintf("%d %d %s",
+			val.GetProtocol(),
+			val.GetPort(),
+			indicatorID,
+		)
```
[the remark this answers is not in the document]
JoukoVirtanen: We now have various functions to get the various keys and all keys are obtained by calling those functions.

pkg/postgres/schema/process_indicators.go
```diff
@@ -47,16 +50,20 @@ const (
...
+	PodId              string `gorm:"column:podid;type:varchar;index:unique_indicator"`
```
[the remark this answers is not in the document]
erthalion: Yep, we've already created ROX-13790 for that.

central/processindicator/store/postgres/gen.go
```diff
@@ -1,3 +0,0 @@
-package postgres
-
-//go:generate pg-table-bindings-wrapper --type=storage.ProcessIndicator --search-category PROCESS_INDICATORS --references storage.Deployment
```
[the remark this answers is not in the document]
erthalion: I'm a bit cautious about doing this, after we've learned that removing the header will make those files eligible for style checking, which they not always pass. In case of process indicator schema it looks fine for now, but maybe it will not be like this in the future. What do you think, how important is it to update the header, having in mind the clear statement about generation status later on and those practical considerations I've mentioned above?

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+	return nil
+}
+
+func (ds *datastoreImpl) GetProcessListeningOnPort(
+	ctx context.Context,
+	deployment string,
+) (
+	processesListeningOnPorts []*storage.ProcessListeningOnPort, err error,
+) {
+	if ok, err := plopSAC.ReadAllowed(ctx); err != nil {
```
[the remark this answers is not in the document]
erthalion: To be honest I'm not familiar enough with the scope access mechanism, could you show an example how to modify `ReadAllowed` call to check ClusterID and Namespace as well?

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,503 @@
...
+	for _, val := range portProcesses {
+		indicatorID := ""
+		var processInfo *storage.ProcessIndicatorUniqueKey
+
+		key := getPlopProcessUniqueKey(val)
+
+		if indicator, ok := indicatorsMap[key]; ok {
+			indicatorID = indicator.GetId()
+			log.Debugf("Got indicator %s: %+v", indicatorID, indicator)
+		} else {
```
dashrews78, line 89: do we ever clean up and/or reconcile the case where there is no indicatorID?  Theoretically that could result in a fair number of collisions with the map and orphaned data in the database.

erthalion: They're getting cleaned up during pruning. There are ideas about reconciliation, but first we wanted to get understanding how frequent this happening in practice.

central/processlisteningonport/datastore/datastore_impl.go
dashrews78: looks like this one is still built with spaces.

central/processindicator/store/postgres/gen.go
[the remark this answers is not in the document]
dashrews78: I think my preference would be to update this stuff on the process indicators now as painful as it may be.  My reasoning is that we have no idea what else may need to change in process indicators outside of your PLOP work over the next few sprints and we would want it to be as obvious as possible to other contributors as to what is occurring.  That is just one person's opinion of course.

central/processlisteningonport/datastore/datastore_impl_test.go
```diff
@@ -0,0 +1,1149 @@
...
+
+	hasNoneCtx  context.Context
+	hasReadCtx  context.Context
+	hasWriteCtx context.Context
+
+	mockCtrl *gomock.Controller
+}
+
+func (suite *PLOPDataStoreTestSuite) SetupTest() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
```
dashrews78, line 44: if we use a `SetupSuite` we should be able to do this check one time and then call a `suite.T().Skip()` or something like that and skip the entire suite so we don't have to put this check in each test.  Something like in `central/node/datastore/dackbox/datastore/datastore_impl_test.go` only the reverse case.

central/processlisteningonport/datastore/datastore_impl_test.go
```diff
@@ -0,0 +1,1149 @@
...
+	// Reopen PLOP objects
+	suite.NoError(suite.datastore.AddProcessListeningOnPort(
+		suite.hasWriteCtx, plopObjectsActive...))
+
+	// Fetch inserted PLOP back
+	newPlops, err := suite.datastore.GetProcessListeningOnPort(
+		suite.hasWriteCtx, fixtureconsts.Deployment1)
+	suite.NoError(err)
+
+	// It's being closed and excluded from the API response
```
dashrews78, line 366: Comment seems wrong since it has been re-opened.

central/processlisteningonport/store/postgres/store.go
dashrews78: My preference would be to do it now and get it over with so it doesn't linger.  It also clears it up that the store is in fact no longer generated.  But again that is my opinion.

pkg/postgres/schema/process_listening_on_ports.go
```diff
@@ -0,0 +1,57 @@
+// Code generated by pg-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: Is this one still being generated?  the line in `gen.go` is commented out so I suspect it is not.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,503 @@
...
+		return nil
+	}
+
+	if ok, err := plopSAC.WriteAllowed(ctx); err != nil {
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
+
+	portProcesses, completedInBatch := normalizePLOPs(portProcesses)
```
dashrews78, line 64: Do we really want to reuse the name of the incoming parameter?  I would think we would want to name it something else to differentiate since we pass that into `normalizePLOPs` and get 2 things back.

central/processlisteningonport/datastore/datastore_impl.go
dashrews78: to clarify:  In the case of this function is it that we only ever have the most recent in the database as was calculated through the combination of normalize and Add functions?  Meaning this particular if check should not occur?

central/processlisteningonport/store/postgres/store.go
erthalion: I have to disagree here. Having this done will introduce unnecessary divergence between generated version and generated bits of the custom version, making it much more painful to reconcile them in the future. I mean, how silly would that be when we will get conflicts just because of `podId` will be different from `podID`. And the fast that the store is not longer generated is stated clearly everywhere else.

If you have a strong opinion on this topic, I would be fine following your suggestion, but will recruit you in the future for reconciliation work :)

pkg/postgres/schema/process_indicators.go
erthalion: Sure, makes sense.

central/processlisteningonport/store/postgres/store.go
dashrews78: One thing to note, such divergence is already unavoidable as #4210 resolved a lot of the style issues for these generated stores.  that specifically targeted things like changing podId to be podID.  of course an option could be to save off the custom bits, regenerate with #4210 and then add the custom stuff back.

central/pruning/pruning_test.go
JoukoVirtanen: Fixed

central/pruning/pruning.go
```diff
@@ -417,6 +426,41 @@ func (g *garbageCollectorImpl) removeOrphanedProcessBaselines(deployments set.Fr
...
+// on per-deployment basis, since the deployment and cluster info is taken from
+// the process indicator, so clean without such filtering in batches.
+func (g *garbageCollectorImpl) removeOrphanedPLOP() {
+	var plopToPrune []string
+	now := types.TimestampNow()
+
+	// This walk seems unnecessary, but ClosedTimestamp is a part of serialized
+	// column, not visible to Postgres. So if we would like to keep using
+	// orphanWindow, unfortunately it has to be this way, at least for now.
+	walkFn := func() error {
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processindicator/datastore/datastore.go
```diff
@@ -43,9 +44,10 @@ type DataStore interface {
 }
 
 // New returns a new instance of DataStore using the input store, indexer, and searcher.
-func New(store store.Store, indexer index.Indexer, searcher search.Searcher, prunerFactory pruner.Factory) (DataStore, error) {
+func New(store store.Store, plopStore plopStore.Store, indexer index.Indexer, searcher search.Searcher, prunerFactory pruner.Factory) (DataStore, error) {
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processindicator/datastore/datastore_impl.go
```diff
@@ -137,12 +142,28 @@ func (ds *datastoreImpl) removeMatchingIndicators(ctx context.Context, results [
 }
 
 func (ds *datastoreImpl) removeIndicators(ctx context.Context, ids []string) error {
+	log.Infof("Process Pruning")
```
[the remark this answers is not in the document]
JoukoVirtanen: Removed

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
+
+	return ds.removePLOP(ctx, ids)
+}
+
+func (ds *datastoreImpl) removePLOP(ctx context.Context, ids []string) error {
+	log.Infof("Deleting PLOP records")
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+
+	if len(ids) == 0 {
+		return nil
+	}
+	if err := ds.storage.DeleteMany(ctx, ids); err != nil {
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+// XXX: This function queries all PLOP, no matter if they are matching port +
+// protocol we've got or not. This means potentially dangerous corner cases
+// when one process listenst to a huge number of ports. To address it we could
+// introduce filtering by port and protocol to the query, and even without
+// extra indices PostgreSQL will be able to do it relatively efficiently using
+// bitmap scan.
+func (ds *datastoreImpl) fetchExistingPLOPs(
+	ctx context.Context,
+	indicatorIds []string,
+	portProcesses ...*storage.ProcessListeningOnPortFromSensor,
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	for _, val := range indicators {
+		key := fmt.Sprintf("%s %s %s %s %s",
+			val.GetContainerName(),
+			val.GetPodId(),
+			val.GetSignal().GetName(),
+			val.GetSignal().GetArgs(),
+			val.GetSignal().GetExecFilePath(),
+		)
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,485 @@
...
+func sortByCloseTimestamp(values []*storage.ProcessListeningOnPortFromSensor) {
+	sort.Slice(values, func(i, j int) bool {
+		return values[i].CloseTimestamp.Compare(values[j].CloseTimestamp) == -1
+	})
+}
+
+func addNewPLOP(plopObjects []*storage.ProcessListeningOnPortStorage,
+	indicatorID string,
+	processInfo *storage.ProcessIndicatorUniqueKey,
+	value *storage.ProcessListeningOnPortFromSensor) []*storage.ProcessListeningOnPortStorage {
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Sorry for the confusion my previous response in this thread caused. I deleted it. We should not have entries with with identical keys here where the key consists of the protocol, port, and indicatorID. That is because of the normalize and Add functions. We should never see the log statement "A PLOP %s is already present, overwrite with %s"

central/processlisteningonport/store/postgres/store.go
erthalion: We have generated those bits again to apply the new ordering coming from this change.

pkg/postgres/schema/process_listening_on_ports.go
erthalion: `gen.go` for PLOP is still in place, only one for ProcessIndicator was removed. PLOP doesn't generate the store itself due to `--schema-only`, but the db schema is still automated and in sync.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Oops. Fixed now.

pkg/postgres/schema/process_indicators.go
JoukoVirtanen: Done

central/processindicator/store/postgres/store.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by pg-bindings generator. DO NOT EDIT.
+// Code originally generated by pg-bindings generator.
```
dashrews78, line 1: nit:  Can we add a TODO with the reference to the ticket(s) we plan to use to get this back to being generated?

central/processlisteningonport/store/postgres/index.go
```diff
@@ -0,0 +1,65 @@
+// Code generated by pg-bindings generator. DO NOT EDIT.
```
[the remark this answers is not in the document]
JoukoVirtanen: These comments have been changed to "Code originally generated by pg-bindings generator"

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,763 @@
...
+			&id, &serialized,
+			&podID, &containerName,
+			&signalContainerID,
+			&signalName, &signalArgs, &signalExecFilePath,
+			&clusterID, &procSerialized); err != nil {
+			return nil, pgutils.ErrNilIfNoRows(err)
+		}
+
+		var msg storage.ProcessListeningOnPortStorage
+		if err := proto.Unmarshal(serialized, &msg); err != nil {
```
rhybrillou, line 720: If I remember correctly, this pattern was replaced in `pkg/search/postgres/common.go` to leverage the `Unmarshal` function in the generated code. The change was driven by performance considerations (https://github.com/stackrox/stackrox/pull/3451).
https://github.com/stackrox/stackrox/blob/master/pkg/search/postgres/common.go#L883-L893
I think the pattern here would be 
```
    msg := &storage.ProcessListeningOnPortStorage{}
    if err := msg.Unmarshal(serialized); err != nil {
        return nil, err
    }
```
This change is not critical in my eyes, applies to the ProcessIndicator decoding as well, and can be the target of a follow-up PR.

central/processlisteningonport/datastore/datastore_impl_test.go
```diff
@@ -0,0 +1,1100 @@
...
+
+func (suite *PLOPDataStoreTestSuite) SetupSuite() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		suite.T().Skip("Skip PLOP tests if postgres is disabled")
+		suite.T().SkipNow()
+	}
+}
+
+func (suite *PLOPDataStoreTestSuite) SetupTest() {
+	suite.hasNoneCtx = sac.WithGlobalAccessScopeChecker(context.Background(), sac.DenyAllAccessScopeChecker())
```
dashrews78, line 51: nit:  If you don't need this stuff to be reset for each test, you can move it to `SetupSuite` if you want.

central/processlisteningonport/datastore/datastore_impl_test.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl_test.go
JoukoVirtanen: It has been corrected

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: No. portProcesses has been changed to normalizedPLOPs

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,763 @@
...
+	// XXX: Verify the query plan to make sure needed indexes are in use.
+	getByDeploymentStmt = "SELECT plop.id, plop.serialized, " +
+		"proc.podid, proc.containername, " +
+		"proc.signal_containerid, " +
+		"proc.signal_name, proc.signal_args, proc.signal_execfilepath, " +
+		"proc.clusterid, proc.serialized as procSerialized " +
+		"FROM process_listening_on_ports plop " +
+		"JOIN process_indicators proc " +
+		"ON plop.processindicatorid = proc.id " +
+		"WHERE proc.deploymentid = $1 AND plop.closed = false"
```
rhybrillou, line 59: AFAIK, In the current state of things, the pattern applied in our app is to duplicate the data from the serialized object in the table columns that can be used for searching.
As a consequence, I think the query here could simply select `plop.id, plop.serialized, proc.serialized as procSerialzed`. This would slightly reduce the amount of data sent over the wire, and decoded by the process.

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,763 @@
...
+		}
+
+		var procMsg storage.ProcessIndicator
+		if err := proto.Unmarshal(procSerialized, &procMsg); err != nil {
+			return nil, err
+		}
+
+		plop := &storage.ProcessListeningOnPort{
+			Endpoint: &storage.ProcessListeningOnPort_Endpoint{
+				Port:     msg.Port,
```
rhybrillou, line 731: General consideration applicable for all fields used to build the plop: the getters from the generated protobuf code should be nil-safe.
I'd prefer here `msg.GetPort()` to `msg.Port`, or later `procMsg.GetSignal().GetId()` to `procMsg.Signal.Id`.

central/processlisteningonport/store/postgres/store.go
```diff
@@ -0,0 +1,763 @@
...
+				Protocol: msg.Protocol,
+			},
+			DeploymentId:  procMsg.DeploymentId,
+			PodId:         podID,
```
rhybrillou, line 735: If the suggested simplified query statement were used, `podID` would not have to be decoded, and `procMsg.GetPodId()` could fill the same purpose.
Variation of this could be applied for `procMsg.GetContainerName()`, `procMsg.GetSignal().GetContainerId()` and the other select fields that could be deduced from the serialized payload.

rhybrillou (commented):
Minor comments on the store part.

central/processlisteningonport/store/postgres/store.go
erthalion: It's probably fine to use this approach for this particular PR, if everything else works this way. But in long term perspective such duplication is a huge problem (more data to store, bulletproof mechanism is required to make sure columns are really the same as in serialized field) and has to be addressed.

central/processlisteningonport/datastore/datastore_impl.go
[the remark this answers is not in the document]
erthalion: What happened here is that there is a possibility to do SAC verification more fine grained, and allow users who have only access to certain namespaces to see PLOPs. As in example [[1]]  we would need to fetch cluster id and namespace for that. The problems would be to do this efficiently via deployment argument that we have in this API, as we only have the deployment name, which is not an indexed column in the db. All in all we agree that this could be done as a follow up, and needs to be investigated first in ROX-14417.

[1]: https://github.com/stackrox/stackrox/blob/master/central/deployment/datastore/datastore_impl.go#L129-L140

central/processlisteningonport/store/postgres/store.go
JoukoVirtanen: Done

central/processlisteningonport/store/postgres/store.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,503 @@
...
+				// We try to close a not existing Endpoint, something is wrong
+				log.Warnf("Found no matching PLOP to close for %s", key)
+			}
+
+			plopObjects = addNewPLOP(plopObjects, indicatorID, processInfo, val)
+		}
+	}
+
+	// Verify what to do about pairs of open/close events that close the
+	// lifecycle withih the batch. There are only few options:
```
rhybrillou, line 126: typo: within

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,503 @@
...
+				// This events should always be closing by definition
+				log.Warnf("Found active PLOP completed in the batch %+v", val)
+			}
+
+			plopObjects = addNewPLOP(plopObjects, indicatorID, val.Process, val)
+		}
+	}
+
+	// Now save actual PLOP objects
+	err = ds.storage.UpsertMany(ctx, plopObjects)
```
rhybrillou, line 173: nit: `return ds.storage.UpsertMany(ctx, plopObjects)`

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,503 @@
...
+// bitmap scan.
+func (ds *datastoreImpl) fetchExistingPLOPs(
+	ctx context.Context,
+	indicatorIds []string,
+) (map[string]*storage.ProcessListeningOnPortStorage, error) {
+
+	var existingPLOPMap = map[string]*storage.ProcessListeningOnPortStorage{}
+
+	if len(indicatorIds) == 0 {
+		return nil, nil
```
rhybrillou, line 248: I'd rather return the empty map than nil here.

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
JoukoVirtanen: Done

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -0,0 +1,498 @@
...
+	)
+}
+
+func getPlopKey(plop *storage.ProcessListeningOnPortStorage) string {
+	return getPlopKeyFromParts(plop.GetProtocol(), plop.GetPort(), plop.GetProcessIndicatorId())
+}
+
+func sortByCloseTimestamp(values []*storage.ProcessListeningOnPortFromSensor) {
+	sort.Slice(values, func(i, j int) bool {
+		return values[i].CloseTimestamp.Compare(values[j].CloseTimestamp) == -1
```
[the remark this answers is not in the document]
rhybrillou: (ideally for both compared entries)

central/processlisteningonport/datastore/datastore_impl_test.go
```diff
@@ -0,0 +1,1099 @@
...
+	plopsFromDB := []*storage.ProcessListeningOnPortStorage{}
+	err = suite.datastore.WalkAll(suite.hasWriteCtx,
+		func(plop *storage.ProcessListeningOnPortStorage) error {
+			plopsFromDB = append(plopsFromDB, plop)
+			return nil
+		})
+	suite.NoError(err)
+	suite.Len(plopsFromDB, 1)
+	suite.Equal(plopsFromDB[0].CloseTimestamp, closedPlopObject3.CloseTimestamp)
+}
```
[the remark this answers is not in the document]
erthalion: Yeah, why not. Added a new test for that as well.

rhybrillou (approved):

dashrews78 (approved):

PR #4332, opened 2023-01-09
Fix nil pointer panic in telemetry config

dhaus67:
## Description

In restricted environments (e.g. ACSCS), Central may not be able to reach out to the Kubernetes API server to retrieve metadata (e.g. due to missing RBAC, proxy configuration etc.).

While we added logic to ignore potential errors during initialization (e.g. reading the service account token / CA from file, setting up the clients), we ignore any error returned from `discoveryclient.ServerVersion()`.

This lead to `v` being set to `nil` when `err != nil` ([see reference](https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L567-L578)) and subsequently to a nil pointer panic when accessing `v.GitVersion`.

Now, we do not ignore the error and only set `v` in case of `err == nil`.

## Checklist

### DIFF
--- a/central/telemetry/centralclient/instance_config.go
+++ b/central/telemetry/centralclient/instance_config.go
@@ -38,7 +38,9 @@ func getInstanceConfig() (*phonehome.Config, map[string]any, error) {
 	v := &k8sVersion.Info{GitVersion: "unknown"}
 	if rc, err := rest.InClusterConfig(); err == nil {
 		if clientset, err := kubernetes.NewForConfig(rc); err == nil {
-			v, _ = clientset.ServerVersion()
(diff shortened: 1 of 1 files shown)


dhaus67:
* master
  * **PR #4332** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4332" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4332?utm_source=stack-comment).

rukletsov (approved):

stehessel:
Should we add a regression test?

0x656b694d (approved):
Thank you.

PR #4315, opened 2023-01-06
Consolidate node datastores: part 2

md2119:
## Description

This PR replaces global node datastore in search. 

Part 1: #4309 

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

PR #4334, opened 2023-01-09
Consolidate node datastores: part 3

md2119:
## Description

Removes global datastore from message pipeline.

Part 1: #4309 
Part 2: #4315 

## Checklist

## Testing Performed
CI


RTann (approved):
nit: can we replace `clusterID` with the wildcard `_` in `Run`?

RTann:
> nit: can we replace `clusterID` with the wildcard `_` in `Run`?

never mind, I see it's still needed to get the cluster's name

PR #4337, opened 2023-01-09
ROX-14239: Fix VM report search when no results returned

vjwilson:
## Description

The way we nested the table and empty state message cause the user to get stuck with filters that cannot be cleared if there are no results for a set of filters.

This PR leaves the existing "empty" message when there are no reports even without filters, but now shows the complex table component when there are results OR a filter applied, so that the user can modify or clear filters to "back out" to results.

I based the new "No search results" message on the example in the PatternFly docs.
https://www.patternfly.org/v4/components/table#composable-empty-state

Notes:
1. For better UX, I disable the sorting on the table if it is shown but there are no results.
2. I add an override to the CSS, because when the table is shown with no results, the checkbox was getting cut off on the right. (I considered hiding the checkbox or the whole header row when there were no results, but that made the layout a little too "jumpy".)

## Checklist

## Testing Performed

Manual testing

With no reports or filters (this existing behavior remains the same)
Screen Shot 2023-01-09 at 5 47 12 PM

With report but no filters applied (this existing behavior remains the same also)
Screen Shot 2023-01-09 at 5 47 40 PM

With report, but filter causes returned result to be empty (new behavior, filter bar now available to allow filters to be modified or deleted)
Screen Shot 2023-01-09 at 5 47 54 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportTablePage.tsx
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportTablePage.tsx
@@ -110,7 +110,6 @@ function ReportTablePage({ query }: ReportTablePageProps): ReactElement {
             triggerRefresh();
         });
     }
-
     return (
         <>
             <PageSection variant={PageSectionVariants.light}>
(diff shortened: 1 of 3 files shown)


dvail (approved):
LGTM :shipit:

PR #4338, opened 2023-01-09
Show filters in Risk Acceptance even when no results

vjwilson:
## Description

Another area of the app that requires a slightly different empty state when filtering causes the results to be empty.

The way we nested the table and empty state message cause the user to get stuck with filters that cannot be cleared if there are no results for a set of filters.

This PR leaves the existing "empty" message when there are no pending, approved deferrals, or approved false positives, even without filters, but now shows the complex table component when there are results OR a filter applied, so that the user can modify or clear filters to "back out" to results.

I based the new "No search results" message on the example in the PatternFly docs.
https://www.patternfly.org/v4/components/table#composable-empty-state

This is the same approach as done in #4337 .
This may be worth abstracting later.

## Checklist

## Testing Performed

With nothing in the database or filters (this existing behavior remains the same)
Screen Shot 2023-01-09 at 6 42 53 PM
Screen Shot 2023-01-09 at 6 42 55 PM
Screen Shot 2023-01-09 at 6 42 58 PM

With results but no filters applied (this existing behavior remains the same also)
Screen Shot 2023-01-09 at 6 42 07 PM
Screen Shot 2023-01-09 at 6 42 10 PM
Screen Shot 2023-01-09 at 6 42 12 PM

With results in database, but filter causes returned result to be empty (new behavior, filter bar now available to allow filters to be modified or deleted)
Screen Shot 2023-01-09 at 6 27 03 PM
Screen Shot 2023-01-09 at 6 34 30 PM
Screen Shot 2023-01-09 at 6 41 49 PM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/RiskAcceptance/ApprovedDeferrals/ApprovedDeferrals.tsx
+++ b/ui/apps/platform/src/Containers/VulnMgmt/RiskAcceptance/ApprovedDeferrals/ApprovedDeferrals.tsx
@@ -40,7 +40,7 @@ function ApprovedDeferrals(): ReactElement {
     const rows = data?.vulnerabilityRequests || [];
     const itemCount = data?.vulnerabilityRequestsCount || 0;
 
-    if (!isLoading && rows && rows.length === 0) {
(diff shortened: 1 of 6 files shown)


dvail (approved):
Thanks again, LGTM

PR #4339, opened 2023-01-10
ROX-12902: adding legend to network graph

alwayshooin:
mock: 
image

actual: 
image

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import { useHistory, useParams } from 'react-router-dom';
 import {
     SELECTION_EVENT,
@@ -97,6 +97,7 @@ const TopologyComponent = ({
     const { detailId } = useParams();
     const selectedEntity = detailId && getNodeById(model?.nodes, detailId);
     const controller = useVisualizationController();
+    const [isLegendOpen, setIsLegendOpen] = useState(false);
 
     function rerenderGraph() {
         setNodes();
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -324,14 +326,17 @@ const TopologyComponent = ({
...
                     })}
                 />
             }
         >
             <VisualizationSurface state={{ selectedIds }} />
+            <Popover
+                aria-label="Network graph legend"
+                bodyContent={<LegendContent />}
+                hasAutoWidth
+                reference={() => document.getElementById('legend') as HTMLButtonElement}
```
alwayshooin, line 338: this is following an example on the popover patternfly docs (https://www.patternfly.org/v4/components/popover#popover-selector-ref)

ui/apps/platform/src/Containers/NetworkGraph/components/LegendContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+                <DescriptionListCompact isHorizontal termWidth="20px" className="pf-u-pl-md">
+                    <DescriptionListItem term={<BuilderImageIcon />} desc="Deployment" />
+                    <DescriptionListItem
+                        term={<PficonNetworkRangeIcon />}
+                        desc="External CIDR block"
+                    />
+                </DescriptionListCompact>
+            </FlexItem>
+            <FlexItem>
+                <Title headingLevel="h4" className="pf-u-pb-sm">
```
sachaudh, line 29: I would just double-check to see if the heading level is correct (not sure if it is from the code diff)

sachaudh (approved):
LGTM

ui/apps/platform/src/Containers/NetworkGraph/components/LegendContent.tsx
vjwilson: This may be a use-case for a hidden heading in the pop-up that reads "Legend", and has a heading level one higher than these.

```
   <Title headingLevel="h3" className="pf-u-screen-reader">Legend</Title>
```

ui/apps/platform/src/Containers/NetworkGraph/components/LegendContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+                    Deployment badges
+                </Title>
+                <DescriptionListCompact isHorizontal termWidth="20px" className="pf-u-pl-md">
```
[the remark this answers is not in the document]
alwayshooin: oh we're not doing that one bc of performance reasons so i left it out of the legend

PR #4340, opened 2023-01-10
ROX-12908: adding external entities/cidr block svg to network graph

alwayshooin:
should look like below: 
image

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/components/StyleNode.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/components/StyleNode.tsx
@@ -23,7 +23,6 @@ import {
     WithSelectionProps,
 } from '@patternfly/react-topology';
 import DefaultIcon from '@patternfly/react-icons/dist/esm/icons/builder-image-icon';
-import AlternateIcon from '@patternfly/react-icons/dist/esm/icons/regions-icon';
 import { PficonNetworkRangeIcon } from '@patternfly/react-icons';
 import useDetailsLevel from '@patternfly/react-topology/dist/esm/hooks/useDetailsLevel';
 import { SVGIconProps } from '@patternfly/react-icons/dist/esm/createIcon';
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -40,6 +40,13 @@ export type NamespaceData = {
 
 export type NetworkPolicyState = 'none' | 'both' | 'ingress' | 'egress';
 
+export type NodeDataType =
+    | 'DEPLOYMENT'
+    | 'EXTERNAL_GROUP'
+    | 'EXTERNAL_ENTITIES'
+    | 'CIDR_BLOCK'
+    | 'EXTRANEOUS';
+
```
alwayshooin, line 49: not sure what the best way to have this exist along side the actual `type` strings in the specific data types of each kind of node below. open to suggestions!

ui/apps/platform/src/Containers/NetworkGraph/components/StyleNode.tsx
```diff
@@ -56,22 +51,24 @@ type StyleNodeProps = {
...
 
-const getTypeIcon = (dataType?: DataTypes): any => {
-    switch (dataType) {
-        case DataTypes.Alternate:
-            return AlternateIcon;
+const getTypeIcon = (type?: NodeDataType): any => {
+    switch (type) {
+        case 'EXTERNAL_ENTITIES':
+        case 'CIDR_BLOCK':
+            return PficonNetworkRangeIcon;
```
sachaudh, line 58: Such a strange name

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
sachaudh: This is fine. I don't know if there is a fancy Typescript way of going through the `CustomNodeData` type and extracting the value for the `type` field similarly. Seems like overkill anyways

sachaudh (approved):

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
vjwilson: Could you use the TS `as const` pattern where the actual constants are exported to get both "for the price of one"?
```suggestion
export const type = {
    DEPLOYMENT: 'DEPLOYMENT',
    EXTERNAL_GROUP: 'EXTERNAL_GROUP',
    EXTERNAL_ENTITIES: 'EXTERNAL_ENTITIES',
    CIDR_BLOCK: 'CIDR_BLOCK',
    EXTRANEOUS: 'EXTRANEOUS',
} as const;
```

alwayshooin: thanks, will explore this in a later PR

PR #4341, opened 2023-01-10
Remove remnants of orchestrator vulnerability data from Central

RTann:
## Description

Central no longer needs to store orchestator scanning-related vulnerability data, as Scanner is now the source-of-truth. This PR removes the dependency on this data. It also updates Central to use the same codepath for orchestrator scanning in both online and offline mode, as having two separate paths is no longer required.

## Checklist

## Testing Performed

CI


central/cve/fetcher/manager_impl.go
```diff
@@ -16,66 +11,44 @@ import (
...
+// Start begins the process to periodically scan orchestrator-level components.
 func (m *orchestratorIstioCVEManagerImpl) Start() {
-	if m.mgrMode != online {
-		log.Error("can't fetch in non-online mode")
-		return
-	}
-
-	ticker := time.NewTicker(fetchDelay)
+	ticker := time.NewTicker(env.OrchestratorVulnScanInterval.DurationSetting())
 	defer ticker.Stop()
```
md2119, line 39: You can move this inside the anonymous func

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Technical Changes
 - ROX-12967: Re-introduce `rpm` to the main image in order to be able parse installed packages on RHCOS nodes (from Compliance container)
 - ROX-14280: ACS operator default channel changes from `latest` to `stable`. Users of older versions must follow the upgrade procedure in order to preserve ACS data in case of issues with the upgrade.
+- The k8s-istio.zip file inside of scanner-vuln-updates.zip (the file downloaded from https://install.stackrox.io/scanner/scanner-vuln-updates.zip for updating Scanner vulnerabilities in offline-mode)
+  is no longer needed. We will continue to populate it to support older versions of the product, but it will be ignored.
```
md2119, line 41: `We will continue to populate it to support older versions of the product, but it will be ignored.` not following this. Do you want to say that `https://install.stackrox.io/scanner/scanner-vuln-updates.zip for updating Scanner vulnerabilities in offline-mode` will continue to be populated? or Do you mean we will continue to populated the file `k8s-istio.zip` bundled in the image?

central/cve/fetcher/manager_impl.go
RTann: not "can". it's "must"! great catch. It is likely the ticket would have never gone off because it was stopped once `Start()` completed

CHANGELOG.md
RTann: we will continue to populate the `k8s-istio.zip` file found inside of `scanner-vuln-updates.zip` which is downloaded as part of the offline-dumps, but it is no longer needed (`scanner-vuln-updates.zip` is a ZIP of ZIPs)

central/cve/fetcher/manager_impl.go
RTann: this proves I did not test this well enough to catch this. Will test this more to make sure it runs as expected

connorgorman (approved):

PR #4347, opened 2023-01-10
Dashrews/ROX-14211 jsonpb allow unknown pkg utilities

dashrews78:
## Description

Trying to trickle these through as they are fairly quick after #4316.  Simply use the JSONUnmarshaler instead of jsonpb directly to ensure that AllowUnknownFields is set for unmarshaling to avoid compatibility issues.

## Checklist

## Testing Performed

Simply using a method that is already tested in `pkg/jsonutil/conversion_test.go`

### DIFF
--- a/pkg/defaults/categories/categories.go
+++ b/pkg/defaults/categories/categories.go
@@ -1,7 +1,6 @@
 package categories
 
 import (
-	"bytes"
 	"embed"
 	"path/filepath"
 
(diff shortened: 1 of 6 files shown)


md2119 (approved):

pkg/defaults/categories/categories.go
```diff
@@ -55,7 +55,7 @@ func readCategoryFile(path string) (*storage.PolicyCategory, error) {
 	utils.CrashOnError(err)
 
 	var category storage.PolicyCategory
-	err = jsonpb.Unmarshal(bytes.NewReader(contents), &category)
+	err = jsonutil.JSONUnmarshaler().Unmarshal(bytes.NewReader(contents), &category)
```
msugakov, line 58: Overall looks ok, but feels that `jsonutil.JSONUnmarshaler().Unmarshal(` could be made shorter, e.g. `jsonutil.JSONBytesToProto(` - a new wrapper that covers cases in this PR.
Feel free to treat it as optional.

msugakov (approved):

pkg/defaults/categories/categories.go
dashrews78: I like that suggestion.  That will make things smoother as we work through the rest of the tickets.

msugakov (approved):

PR #4354, opened 2023-01-11
ROX-13705: Use real data for CIDR block side panel

sachaudh:
## Description

This is an iterative approach for using real data in the CIDR Block side panel. Things I still need to do in the future:
- [x] Create network flows using the edges
- [x] Fix typing for out edges
- [x] Change usage of EdgeModel to a new CustomEdgeModel
- [x] Hook up filtering for network flows 
- [ ] Create nested network flows (expandable)
- [ ] Create tests for util functions (is a bit of work since we use a controller in the function)

https://user-images.githubusercontent.com/4805485/211851602-8bcf5e75-8da8-4274-8ccb-fdde11d019c5.mov

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
+++ b/ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
@@ -1,9 +1,11 @@
-import { Model, NodeModel } from '@patternfly/react-topology';
+import { EdgeModel, Model, NodeModel } from '@patternfly/react-topology';
 
-import { ListenPort, Edge } from 'types/networkFlow.proto';
+import { EdgeProperties, ListenPort, OutEdges } from 'types/networkFlow.proto';
 import { Override } from 'utils/type.utils';
 
-export type CustomModel = Override<Model, { nodes?: CustomNodeModel[] }>;
+export type CustomModel = Override<Model, { nodes: CustomNodeModel[]; edges: CustomEdgeModel[] }>;
+
+// Node types
 
 export type CustomNodeModel =
     | NamespaceNodeModel
@@ -72,7 +74,7 @@ export type ExternalGroupData = {
 export type ExternalEntitiesData = {
     type: 'EXTERNAL_ENTITIES';
     id: string;
-    outEdges: Edge[];
+    outEdges: OutEdges;
 };
 
 export type CIDRBlockData = {
@@ -83,7 +85,7 @@ export type CIDRBlockData = {
         default: boolean;
         name: string;
     };
-    outEdges: Edge[];
+    outEdges: OutEdges;
 };
 
 export type ExtraneousData = {
@@ -91,4 +93,12 @@ export type ExtraneousData = {
     collapsible: boolean;
     showContextMenu: boolean;
     numFlows: number;
+};
+
+// Edge types
+
+export type CustomEdgeModel = Override<EdgeModel, { data: EdgeData }>;
+
+export type EdgeData = {
+    properties: EdgeProperties[];
 };
--- a/ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
+++ b/ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
@@ -1,5 +1,18 @@
+import { Controller } from '@patternfly/react-topology';
 import { uniq } from 'lodash';
+import { AdvancedFlowsFilterType } from '../common/AdvancedFlowsFilter/types';
 import { Flow } from '../types/flow.type';
+import { CustomEdgeModel, CustomNodeData } from '../types/topology.type';
+
+const protocolLabel = {
+    L4_PROTOCOL_UNKNOWN: 'UNKNOWN',
+    L4_PROTOCOL_TCP: 'TCP',
+    L4_PROTOCOL_UDP: 'UDP',
+    L4_PROTOCOL_ICMP: 'ICMP',
+    L4_PROTOCOL_RAW: 'RAW',
+    L4_PROTOCOL_SCTP: 'SCTP',
+    L4_PROTOCOL_ANY: 'ANY',
+};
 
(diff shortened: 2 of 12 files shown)


ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
```diff
@@ -19,3 +32,120 @@ export function getNumFlows(flows: Flow[]): number {
...
+        if (flow.entity.includes(entityNameFilter)) {
+            matchedEntityName = true;
+        }
+
+        // check filtering by flow type
+        if (advancedFilters.flows.length) {
+            const isAnomalousFiltered =
+                advancedFilters.flows.includes('anomalous') && flow.isAnomalous;
+            const isBaselineFiltered =
+                advancedFilters.flows.includes('baseline') && !flow.isAnomalous;
```
alwayshooin, line 115: isn't it either anomalous or baseline filtered? i feel like it makes more sense to do this conditionally if so

ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
```diff
@@ -19,3 +32,120 @@ export function getNumFlows(flows: Flow[]): number {
...
+            matchedFlowType = isAnomalousFiltered || isBaselineFiltered;
+        }
+
+        // check filtering by directionality
+        if (advancedFilters.directionality.length) {
+            const isIngressFiltered =
+                advancedFilters.directionality.includes('ingress') && flow.direction === 'Ingress';
+            const isEgressFiltered =
+                advancedFilters.directionality.includes('egress') && flow.direction === 'Egress';
+            matchedDirectionality = isIngressFiltered || isEgressFiltered;
```
alwayshooin, line 125: same here

ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
```diff
@@ -19,3 +32,120 @@ export function getNumFlows(flows: Flow[]): number {
...
+        }
+
+        // check filtering by protocols
+        if (advancedFilters.protocols.length) {
+            const isTCPFiltered =
+                advancedFilters.protocols.includes('TCP') && flow.protocol === 'TCP';
+            const isUDPFiltered =
+                advancedFilters.protocols.includes('UDP') && flow.protocol === 'UDP';
```
alwayshooin, line 133: and here

ui/apps/platform/src/types/networkFlow.proto.ts
```diff
@@ -100,16 +100,17 @@ export type Node = {
...
-export type Edge = {
-    [key: string]: {
+export type OutEdges = Record<
+    string,
+    {
         properties: EdgeProperties[];
-    };
-};
+    }
+>;
```
alwayshooin, line 111: oh just curious why you decided to change this?

ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
sachaudh: I guess I'm not understanding what you mean. It's true a network flow is either anomalous or baseline, but with the filters, we can select both `Anomalous` and `Baseline`. So we show anomalous network flows when (1) No filter is selected (2) The anomalous filter is selected (3) When the anomalous filter is selected with the baseline filter

sachaudh: The filtering works this way. If there's a different way you think would be more ideal, and gets the same result, we can try that. I was just confused about what you meant by "it makes more sense to do this conditionally"

sachaudh: I'll merge for now and we can discuss

ui/apps/platform/src/types/networkFlow.proto.ts
sachaudh: Looks like I made a mistake here

sachaudh: It was actually incorrect. `outEdges` is an object where the key is the node index and the value is an object with the field `properties`. I just fixed it here since the autosuggestion wasn't working due to wrong typing

alwayshooin (approved):

vjwilson (approved):
Questions about naming can be addressed later.

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -72,7 +74,7 @@ export type ExternalGroupData = {
 export type ExternalEntitiesData = {
     type: 'EXTERNAL_ENTITIES';
     id: string;
-    outEdges: Edge[];
+    outEdges: OutEdges;
```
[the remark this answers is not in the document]
sachaudh: The type was actually incorrect before but wasn't caught because we weren't actually accessing the value in our code. It actually isn't an array but an object with the key as a number index and a value of an object with the `properties` key. Let me know if that changes your opinion about the renaming

ui/apps/platform/src/Containers/NetworkGraph/types/topology.type.ts
```diff
@@ -92,3 +94,11 @@ export type ExtraneousData = {
     showContextMenu: boolean;
     numFlows: number;
 };
+
+// Edge types
+
+export type CustomEdgeModel = Override<EdgeModel, { data: EdgeData }>;
+
+export type EdgeData = {
+    properties: EdgeProperties[];
```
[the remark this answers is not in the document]
sachaudh: Technically, I think it is. `EdgeProperities` is an object like this:
```
{ lastUpdated: '...', port: '443', protocol: 'L4_PROTOCOL_TCP' }
```
I guess you could say they are properties and not a property?

PR #4355, opened 2023-01-11
ROX-14507: Fix validation check getting skipped on missing daysOfWeek in reportConfig

charmik-redhat:
## Description

The validation check for daysOfWeek in reportConfig schedule was getting skipped when the days was non-nil empty array. This PR fixes the validation by adding an extra check on the length of days slice.

## Checklist

## Testing Performed

Added unit tests


md2119 (approved):
The schedule proto is weird which makes it hard to understand but your validation seems to be sufficient.

PR #4359, opened 2023-01-11
Telemetry: "Updated Identity" event to wake up the central analytics

0x656b694d:
## Description

Amplitude makes use of user traits when there are events from the user. For a user composition chart based on some user properties, there should be an event happened after identity update.

In this PR:
* Record the gathered client identity only if it is different from the one previously collected;
* Send "Updated Identity" event after client identity is recorded;
* Record the initial identity on start up, not waiting for the first timer event.

## Checklist

## Testing Performed

ongoing

### DIFF
--- a/central/main.go
+++ b/central/main.go
@@ -292,11 +292,6 @@ func main() {
 
 	go startGRPCServer()
 
-	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
-		// Issue an event so that the central become visible for analytics:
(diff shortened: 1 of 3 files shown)


0x656b694d:
* master
  * **PR #4359** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4359" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4359?utm_source=stack-comment).

stehessel (approved):

PR #4362, opened 2023-01-11
ROX-13535: Remove references to the embedded documentation

0x656b694d:
## Description

Make central redirect requests to "/docs/product/" to the online documentation.
See stackrox/test-gh-actions#83 for the according release automation update.

## Checklist

## Testing Performed

```sh
$ curl --head -k https://localhost:8000/docs/product/   
HTTP/2 301 
content-type: text/html; charset=utf-8
location: https://docs.openshift.com/acs/3.73/welcome/index.html
vary: Accept-Encoding
date: Wed, 11 Jan 2023 15:43:41 GMT
```

Opened help page from local UI.


0x656b694d:
* master
  * **PR #4362** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4362" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4362?utm_source=stack-comment).

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Added Features
 
 ### Removed Features
-- ROX-12316: As announced in 3.72, the permission `Cluster` replaces the deprecated permission `ClusterCVE`. 
+- ROX-12316: As announced in 3.72, the permission `Cluster` replaces the deprecated permission `ClusterCVE`.
+- ROX-13535: Built-in documentation link redirects now to the online version.
```
msugakov, line 14: Would you be so kind to mention that we also: don't ship the stand-alone docs container and don't embed docs in the product?

central/ui/mux.go
```diff
@@ -1,14 +1,20 @@
...
+	"github.com/stackrox/rox/pkg/version"
 )
 
 // Mux returns a HTTP Handler that knows how to serve the UI assets,
 // including Javascript, HTML, and other items.
 func Mux() http.Handler {
 	mux := http.NewServeMux()
-	mux.Handle("/docs/product/", http.StripPrefix("/docs/product/", http.FileServer(http.Dir("/stackrox/static-data/product-docs"))))
+	mux.Handle("/docs/product/", http.StripPrefix("/docs/product/",
+		http.RedirectHandler(fmt.Sprintf("https://docs.openshift.com/acs/%s/welcome/index.html",
```
msugakov, line 15: 1. Why redirect from the back-end and not change the link destination in UI?
2. Would be good to double-confirm with the Documentation team that `https://docs.openshift.com/acs/${x.yy}/welcome/index.html` is the canonical URL we should take users to.
3. [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) for this redirect may be questionable:
3.1. Does the browser remember the target forever and next time the user upgrades from 3.74 to 3.75 and further, the browser still goes to <https://docs.openshift.com/acs/3.74/welcome/index.html>?
3.2. What happens when we try redirect with `301` to URL that doesn't exist yet, e.g. <https://docs.openshift.com/acs/3.74/welcome/index.html>, or we mess up something (e.g. introduce different ACSCS versioning) (but we never make mistakes, right? right?...) and the target becomes <https://docs.openshift.com/acs/bogus/welcome/index.html>?

central/ui/mux.go
```diff
@@ -1,14 +1,20 @@
...
+			version.GetMajorMinor(version.GetMainVersion())),
```
pedrottimark, line 16: Is it worthwhile to write a short comment about the possible future scenario which you brought up, if I remember correctly: if **dev build** is looking forward to next release instead of looking backward to version number of previous release? In that hypothetical numbering convention, dev build 3.74.x-whatever would need to link to 3.73 last published version of docs instead of not-yet-existing 3.74 version, correct?

central/ui/mux.go
pedrottimark: Maybe out of scope to ask: if central knows that it is an **offline** installation (at the moment, UI does not, although it probably should for System Health) is there any difference in behavior that makes sense?

central/ui/mux.go
0x656b694d: Correct. That is a good point.

central/ui/mux.go
0x656b694d: 1. Some code should calculate the version for the link, why not backend. It could be changed after review from the UI guys;
2. Would be good. I wanted actually to put just "https://docs.openshift.com/acs/%s/", but copy-pasted too much;
3. We'll find a better code, thanks for noticing that.

msugakov: I see two approaches to get the version on the front-end. First, UI calls `/v1/metadata` which should report the product version. Second, we can embed another `REACT_APP_` quasi-environment variable with the version at the build time and use that. In the end, there's always an option to fall back to unversioned https://docs.openshift.com/acs/welcome/index.html

msugakov: And, in order to make things even more fun, I discovered this
https://github.com/stackrox/rox/pull/7416/files#diff-c5e13185526736c199edcc9f02470d99f481013c9d47d3effce0286315811456R58-R61

Which means both existing built-in documentation and your PR leak ACS version information.

pedrottimark: Brad and I will move the logic to compute the `href` attribute of the **Help Center** link into UI.

We observed on k8s demo, that version number is unknown for a few seconds until response from first /v1/metadata request. I have asked technical documentation team about pro and con to render a generic link without version number until it is available, or wait to render **Help Center** in the menu. There is a question whether current redirect from unversioned to versioned address on docs site incorrectly uses 301.

The initial plan is to adapt from the logic from this PR, and then amend this PR to remove backend redirect after frontend change has been merged.

@msugakov Concerning the leak of version information in docs and this PR, can you explain more specifically if there is any problem for us to avoid in the frontend PR?

msugakov: For some reason that wasn't mentioned in https://github.com/stackrox/rox/pull/7416 , back-end will not provide main version in `/v1/metadata` response if the user is unauthenticated.

I _suppose_ that could be to not reveal the product version for untrusted strangers, because they may find what vulnerabilities are present in that version to try exploit them.

What this means for UI is that you need to be prepared to handle the case when `/v1/metadata` returns empty/null main version. I suggest to redirect to unversioned `https://docs.openshift.com/acs/welcome/index.html` for such cases. If/when the version becomes available, take the user to `https://docs.openshift.com/acs/{X.YY}/welcome/index.html`.
OTOH, I don't know if there's any chance unauthenticated user will see the menu link at all. (What concerns me about the backend change here is that unauthenticated user can access `/docs/product/`.)

msugakov: Good point on `301` redirect in https://docs.openshift.com/acs/welcome/index.html itself. Any time I find something in Google about OpenShift in docs.openshift.com, I'm taken to docs of random 4.x version. I would prefer to get the latest 4.11 but that doesn't happen. Not sure if 301 is the cause.
image
image

release/scripts/allowed_vulns.json
```diff
@@ -4,11 +4,5 @@
         "image": ".*",
         "tag": ".*",
         "reason": "An example"
-    },
-    {
-        "vuln": "*",
-        "image": "^docs.*",
```
msugakov, line 10: Thanks for taking care of this.

scripts/ci/lib.sh
```diff
@@ -550,40 +550,6 @@ check_rhacs_eng_image_exists() {
     [[ "$(jq -r '.tags | first | .name' <<<"$check")" == "$tag" ]]
 }
 
-check_docs() {
```
msugakov, line 553: Thanks for taking care of this too.

scripts/offline-bundle/image-bundle/import.sh
```diff
@@ -4,10 +4,7 @@ IFS=$'\n\t'
```
msugakov, line 4: It seems to me, this is all part of the `offline-bundle` that existed and was needed during `stackrox.io` times. Now, as we sunset stackrox.io image registries, and we don't create offline bundle in CI, we can remove offline-bundle stuff completely, I think.
See related https://srox.slack.com/archives/C3W47R5LP/p1651081299957489?thread_ts=1651079204.647909&cid=C3W47R5LP

That said, it shouldn't be a part of this PR and doesn't have to be done by you. The change you have here looks sane, therefore feel free to treat this comment as no-op.

gavin-stackrox (approved):
LGTM
@tommartensen FYI ICYMI

0x656b694d:
To be merged after #4451.

msugakov (approved):
CI looks green -> approving.

tommartensen (approved):
👍 
Is there anything people need to do to their checked out repository or branches with respect to the removed submodules?

0x656b694d:
> +1 Is there anything people need to do to their checked out repository or branches with respect to the removed submodules?

Hopefully nothing to do.

PR #4365, opened 2023-01-11
ROX-14286: Prevent VM report form submission when no days selected

vjwilson:
## Description

I thought it was just awkward to validate PatternFly Select, but after a while of looking for workarounds, I remembered why we didn't add validate to the Days option originally: PF Select with checkboxes fire events on the individual checkbox, instead of the parent Select component. Trying to build a workaround turned into a widening spiral of convoluted code.

So, this is the punt.

Just manually checking that the appropriate `days` array has at least one item selected, and using the result as an additional "validation" variable for disabling the form submission button.

(The check is based on different parts of the form values object, because that is how different day arrays are represented in our protobuf, depending on whether the interval is WEEKLY or MONTHLY.)

(Note: I also fixed two warnings that showed up in the browser console:
1. a key error in the list of weekdays
2. an accessibility warning about the PatternFly PageSection component)

## Checklist

~Unit test and regression tests added~ 
See comment in `ui/apps/platform/cypress/integration/vulnmanagement/reporting/reportingForm.test.js` about how the same issue with PatternFly that prevents easy validation of the checkbox select, also prevents easy automated browser testing of it, too.

## Testing Performed

WEEKLY, no days selected, the submission button is disabled
Screen Shot 2023-01-11 at 2 15 07 PM

WEEKLY, with a day selected, the submission button is enabled
Screen Shot 2023-01-11 at 2 15 12 PM

MONTHLY, no days selected, the submission button is disabled
Screen Shot 2023-01-11 at 2 14 51 PM

MONTHLY, with a day selected, the submission button is enabled
Screen Shot 2023-01-11 at 2 14 58 PM

### DIFF
--- a/ui/apps/platform/src/Components/PatternFly/DayPickerDropdown.tsx
+++ b/ui/apps/platform/src/Components/PatternFly/DayPickerDropdown.tsx
@@ -38,7 +38,7 @@ function DayPickerDropdown({
     const selectOptions =
         intervalType === 'WEEKLY'
             ? [
-                  <SelectOption key="sunday" value="1">
+                  <SelectOption key="monday" value="1">
                       Monday
                   </SelectOption>,
                   <SelectOption key="tuesday" value="2">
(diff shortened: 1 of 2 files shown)


pedrottimark (approved):
A local variable makes sense.

dvail (approved):
No more three day weekends with two Sundays. 😔

PR #4369, opened 2023-01-11
ROX-13963: Use new database-status endpoint to show DB warning banner

vjwilson:
WIP

## Description

Take 2 on showing a warning banner in the UI when the Postgres-compatible DB is unavailable.

This version uses the new `/database-status` endpoint to get its info.

## Checklist

## Testing Performed

Blocking the endpoint to test the banner:
Screen Shot 2023-01-11 at 5 56 53 PM

And then unblocking the endpoint, and waiting no more than 60 seconds for the API call to poll:
Screen Shot 2023-01-11 at 5 57 54 PM


dashrews78:
@vjwilson slight adjustment to the endpoint based on feedback for my PR.

endpoint is now `/v1/database/status`

ui/apps/platform/src/types/databaseService.proto.ts
```diff
@@ -0,0 +1,5 @@
+export type DatabaseStatus = {
+    databaseAvailable: boolean;
+    databaseType?: string | null;
+    databaseVersion?: string | null;
```
pedrottimark, line 4: @dashrews78 Can you help clarify a few points on alignment of frontend and backend service?
@vjwilson Some recommended changes pending the clarification.

1. Concerning `?` and `| null`
    * Is the following comment from backend code `// Only return the database type and version to logged in users, not anonymous users.` the explanation why? If yes, Van please copy it to the type definition.
    * Is `null` value possible? I have seen `null` value for some properties which have `string whatever` in proto file. I did not read the Go code and unit tests deeply enough to answer this question.
2. Is database status conceptually a metadata service, or is it coincidence left over from initial attempt?
    * If it is, then from my viewpoint as reviewer or future maintainer, better to move both frontend type and function into MetadataService.ts file to align with corresponding backend file.
    * If it is not, is it possible to move the backend service to a separate file?

dashrews78: @pedrottimark The initial attempt was to simply use the metadata endpoint but that hosed up sensors so we had to pivot.  To me this is a metadata service of the likes of metadata.  The idea is databaseType and databaseVersion are useful information, but like the version in metadata endpoint we probably don't want to provide those details if unauthenticated.  I don't think you will receive a null unless your (java|type)script interprets an empty string as such.  Here is what a curl will provide you:

```
{"databaseAvailable":false,"databaseType":"Hidden","databaseVersion":""}
```

notice the databaseVersion is an empty string.

In regard to your 2nd point, in my mind it is conceptually a metadata service which is why I put it where I did on the backend.  Though I would disagree that the backend and UI need to house the related.  Things such as that could move at any time and keeping where the backend and UI house code seems like an unnecessary task.  IMO

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
```diff
@@ -98,6 +120,7 @@ function MainPage(): ReactElement {
                     hasServiceIdentityWritePermission={hasServiceIdentityWritePermission}
                 />
                 {metadata?.stale && <VersionOutOfDate />}
+                {!databaseAvailable && <DatabaseBanner />}
```
pedrottimark, line 123: Suggestion out of an abundance of caution, because polled request in `MainPage` element causes React to re-render the entire page (even though, in theory, reconciliation algorithm will decide no changes).

I am concerned if this new cause for re-rendering gives any descendant element that might have an unknown weakness not to re-render itself purely (that is, without side effects) an opening to cause subtle intermittent integration test failure because element detached from the DOM (which would be unrelated to test).

Therefore, what do you think if `DatabaseBanner` encapsulates such a frequent polled request. For what it is worth, that is the pattern in widgets on main dashboard and vulnerability management dashboard.

Sadly the precedents are equivocal:
* `UnreachableWarning` depends on `serverState` from Redux store via 'MainPage` but encapsulates `if return null`
* `Notifications` encapsulates its data access from Redux store
* `CredentialExpiryBanner` encapsulates data request
* `VersionOutOfDate` has conditional rendering which depends on `metadata?.state` from Redux store via `MainPage`

ui/apps/platform/src/types/databaseService.proto.ts
pedrottimark: The majority of my contributions are incremental fixes and updates in files that are new to me.

Reasons that caused me to align the files when I rewrote JavaScript as TypeScript:
* I needed to compare the services and types. Where were they? Wasted time to search and find.
* When there is an incremental change to add a field to an existing response, Europe engineers are willing and able to accelerate progress by doing both. Where are they? Wasted time to search and find.

vjwilson:
Superseded by #4403

PR #4370, opened 2023-01-11
Fix protoc download arch on arm64

c-du:
## Description

The arch packages is in the format of aarch_64 instead of arm64. This PR fix the protoc download for Apple M1 local building.

## Checklist
- ~[ ] Documented user facing changes (create PR based on ([openshift/openshift-docs](https://github.com/openshift/openshift-docs))) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed
Local build on Apple M1 Monterey

Screen Shot 2023-01-11 at 3 45 30 PM

### DIFF
--- a/make/protogen.mk
+++ b/make/protogen.mk
@@ -42,7 +42,7 @@ endif
 ifeq ($(UNAME_S),Darwin)
 PROTOC_OS = osx
 endif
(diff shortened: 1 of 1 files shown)


porridge:
I wonder if the same or similar should have be done for `kuttl` in #4276  :thinking:

PR #4372, opened 2023-01-12
ROX-13557: Make Operator central.db.isEnabled Enabled be default and drop the flag

kurlov:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


porridge:
Just a quick note based on the PR description before I dive in:

I don't think I understand this sentence:
> The central.db.isEnabled feature flag remains to ease migration to 3.74 if the flag is already presented in the CRD.

And regarding this:

> The flag will drop after 3.74 release.

I'd have to refresh my memory but I do not think it is possible to remove a field in a given `apiVersion`. So maybe say something along the lines of "this field has no effect and will be removed in a future API version"?

porridge:
Also, the title "Make Operator central.db.isEnabled Enabled by default" sounds much weaker than what actually seems to be happening here?

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -153,20 +153,20 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 }
 
 // CentralDBSpec defines settings for the "central db" component.
+// TODO(ROX-14395): drop `IsEnabled` field.
 type CentralDBSpec struct {
-	// Specify whether central-db is enabled, Default configures central to use rocksdb.
-	// If IsEnabled the operator will provision the Central DB or use an external DB if a connection string is provided.
+	// Deprecated field. It is no longer necessary to specify it.
+	// This field will be removed in a future release.
+	// Central configures to use PostgreSQL by default.
```
porridge, line 169:
```suggestion
	// Central is configured to use PostgreSQL by default.
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -153,20 +153,20 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
...
 	//+kubebuilder:validation:Default=Default
 	//+kubebuilder:default=Default
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=1
```
porridge, line 172: We probably want to hide it in UI now that it has no effect.

```suggestion
	//+operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:hidden"}
```
This will also require renumbering the order of fields below.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -194,15 +194,12 @@ type CentralDBSpec struct {
 }
 
 // CentralDBEnabled is a type for values of spec.central.db.enabled.
-//+kubebuilder:validation:Enum=Default;Enabled
+//+kubebuilder:validation:Enum=Default
```
porridge, line 197: Shrinking the allowed set of values is not a compatible change.
TBH neither is growing the set, but removing a previously allowed value is more likely to cause problems.

```suggestion
//+kubebuilder:validation:Enum=Default
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -194,15 +194,12 @@ type CentralDBSpec struct {
...
 type CentralDBEnabled string
 
 const (
-	// CentralDBEnabledDefault configures the central to use rocksdb
-	// TODO(ROX-13557): in 3.74.0 this should say that Default enables PostgreSQL.
+	// CentralDBEnabledDefault configures the central to use PostgreSQL database (Technology Preview)
```
porridge, line 201: It's no longer preview now that it's the only available option, right?

operator/tests/central/basic/74-errors.yaml
```diff
@@ -1,9 +0,0 @@
-apiVersion: v1
-kind: PersistentVolumeClaim
-metadata:
-  name: central-db
```
porridge, line 4: Shouldn't we add these to the assert file now to make sure these are created?

operator/tests/central/basic/74-assert.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: stackrox-db
```
porridge, line 4: I think I heard from someone that the `stackrox-db` PVC is still used for some things. If so, please also keep it here.

operator/tests/central/basic/76-switch-central-db-to-internal.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
 spec:
   central:
     db:
-      isEnabled: "Enabled"
```
porridge, line 8: This step makes no sense now, does it? There is nothing to switch, right?

porridge (changes requested):

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -194,15 +194,12 @@ type CentralDBSpec struct {
...
 	CentralDBEnabledDefault CentralDBEnabled = "Default"
-	// CentralDBEnabledTrue configures the central to use a PostgreSQL database (Technology Preview)
-	CentralDBEnabledTrue CentralDBEnabled = "Enabled"
```
vladbologa, line 205: I also don't think it's a good idea to remove the "Enabled" value. All our existing ACS instances in ACSCS use "Enabled" atm.

I think we should keep both values for now, the only difference being that "Default" now also means "Enabled". After the 3.74 release, we can omit setting this field in `fleetshard-sync`. Then in the 3.75 release we can remove "isEnabled" altogether.

vladbologa, line 205: I also don't think it's a good idea to remove the "Enabled" value. All our existing ACS instances in ACSCS use "Enabled" atm.

I think we should keep both values for now, the only difference being that "Default" now also means "Enabled". After the 3.74 release, we can omit setting this field in fleetshard-sync. Then in the 3.75 release we can remove "isEnabled" altogether.

vladbologa (changes requested):

operator/apis/platform/v1alpha1/central_types.go
msugakov: (Disclaimer: I scrolled through comments and randomly decided to respond.)

> Then in the 3.75 release we can remove "isEnabled" altogether.

I had extensive DM chat with @kurlov about this. Dropping `isEnabled` can be done only if we can prove that CRs created with this attribute will work without any errors after upgrading to the version that doesn't support this flag. If this can be proved reliably, we can drop `isEnabled` already and hardcode the logic that selects Postgres by default.

kurlov: I've tested with dropping the field but I am not 100% sure that I can cover all possible scenario. I think having hidden redundant field is an acceptable trade for have smooth and safe migration. So I lean towards having this field and dropping it in the future

operator/apis/platform/v1alpha1/central_types.go
kurlov: You are right, it's much safer

operator/tests/central/basic/74-errors.yaml
kurlov: It's there but you are right, it makes sense check for stackrox-db PVC as well

operator/tests/central/basic/76-switch-central-db-to-internal.yaml
kurlov: I had an impression that it's about switch from external postgres DB from `connectionString` to internal postgres central-db. if so, it makes sense because we assert that central PVC, deployment, secret is created.

operator/tests/central/basic/74-assert.yaml
kurlov: You are right, it makes sense to check for both central-db and stackrox-db PVCs

operator/apis/platform/v1alpha1/central_types.go
msugakov: Sorry, I don't quite follow the logic. Can you guys help clarify, please?

1. If we **drop** the `isEnabled` field, we need to be **absolutely** sure we can do it **safely**. Correct?
2. If we need to be absolutely sure we can safely drop `isEnabled`, it does not change if we decide to drop it in the future. Correct?
3. If the above is valid, and we simply remove `isEnabled` in the future without making any validations, there's a chance we break some users that have upgraded from 3.73.

By saying

> So I lean towards having this field and dropping it in the future

you mean that you **also** defer the **work of verifying** whether we can **safely** drop `isEnabled` to the future, correct?

porridge:
Unfortunately I won't have time to review today, but perhaps you can fix the failing operator e2e tests in the meantime @kurlov ?

operator/apis/platform/v1alpha1/central_types.go
kurlov: >2. If we need to be absolutely sure we can safely drop `isEnabled`, it does not change if we decide to drop it in the future. Correct?

I am not sure it's correct. I think there is an important difference between dropping right away and doing it in the next release. Basically, instead of `3.73 using isEnabled -> 3.74 dropping isEnabled` there will be `3.73 using isEnabled -> 3.74 ignoring isEnabled -> 3.75 dropping isEnabled`. This additional state helps to make sure that nothing should break because of missing `isEnabled`. If such approach sounds redundant I can drop it in this PR

msugakov: The thing is: I'm not sure how `dropping isEnabled` would look like. Will the operator be able to modify Central Custom Resource that has `isEnabled` attribute from 3.73 version? If not, and the flag will remain present in the Custom Resource, will the operator freak out when it encounters this attribute?

kurlov: The operator works normally if there is unknown field in the central CRD yaml. If `isEnabled` is introduced manually or inherited in 3.75 then this field will be considered as unknown and should not violate operator flow

msugakov: > central CRD yaml

Do you mean Central CR or Central CRD? The difference is significant.

I believe the operator, when deployed, brings CRDs with it. CRD defines validation schema for CRs.
If we remove `isEnabled` in, say, 3.75, its CRD will not have a definition for `isEnabled`. However, Central CRs that were created at the time of 3.73 version will contain `isEnabled` attribute. Those CRs will now be validated against the new CRD without the attribute.
I think the CR/CRD validation is not done by the operator, rather by Kubernetes. Yet, the operator does its own pass over the CR when loading it. Would both things be happy about CR having an unrecognized attribute?

kurlov: Sorry, I've meant CR. I totally agree that there are uncertainties.
@SimonBaeumer could you please your input?

SimonBaeumer: Removing a field from a CRD is considered a breaking API change. If the removal does not have __any__ side-effects we could do it but proving it is hard.
Therefore the only way to remove a field safely from a CRD is by promoting our CRD API version from `v1alpha1` to `v1alpha2` or `v1beta`.

imho the solution, defaulting to enabled and printing a message if a misconfiguration was applied is perfectly fine.

Going forward I would like to reconsider using annotations for temporary API fields (similar to feature flags) instead of adding them to the public API directly which would allow safe removals.

msugakov: > Going forward I would like to reconsider using annotations for temporary API fields (similar to feature flags) instead of adding them to the public API directly which would allow safe removals.

Indeed. We could have ab-used environment variables mechanism on Central CR to enable CentralDB.

kurlov: thanks for clarification, let's keep `isEnabled` and drop it if we will bump the CRD API version.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -115,9 +115,8 @@ type CentralComponentSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=5
 	Persistence *Persistence `json:"persistence,omitempty"`
```
porridge, line 116: It would be great to enumerate what actually is stored on this directly attached PVC (I think someone mentioned it's still used "for some things" but unclear for what). As a customer I would be confused by both `persistence` and `db` mentioning persistent data.

operator/tests/central/basic/76-switch-central-db-to-internal.yaml
porridge: Hm, I think we need to take a look at the whole directory holistically and perhaps rename or remove some steps.

If my understanding is correct, the `10-central-cr.yaml` file contains no `spec.central.db` field therefore it should provision a `central-db` since this is the new default and only supported option in the coming release, right? In that case we should probably add the associated resources to that step's assert file to make this explicit.

In that light, the `74-central-with-db-defaults.yaml` is mostly a no-op, and I think step 74 could be dropped?

Regardless of that, I think step 75 will delete the central-db deployment and thus (in our case) result in a non-functional central because the connection string is a dummy one, correct? I'm not sure this use of switching from an already provisioned central-db to an external DB is a case worth testing - it only makes sense if the user has cloned the database into an external postgres instance out-of-band. Since we're not doing the clone here, we don't learn much from this step... I guess we could leave it around, I have no strong opinions. If we leave it around then it perhaps makes sense to add the central-db PVC to the assert file to make it explicit that we're not deleting it.

Then step 76 is fine, to make sure we can go back to using the (now orphaned after step 75) PVC. It might make sense to set the whole `db` key value to `null` though to make the test a bit more robust (in case step 74 ever sets more properties under it that might not be compatible such as size or storage class).

operator/pkg/central/values/translation/translation.go
```diff
@@ -193,10 +193,9 @@ func getCentralComponentValues(c *platform.CentralComponentSpec) *translation.Va
 		cv.AddChild("exposure", &exposure)
 	}
 
-	if c.CentralDBEnabled() {
+	if c.DB != nil {
```
porridge, line 196: I take from this that **the helm chart** is already changed to default to provisioning a central-db if the `db` field is unset?

porridge (changes requested):
I think we're getting there, see comments inline.

operator/apis/platform/v1alpha1/central_types.go
vladbologa: BTW this was the idea from the beginning, to have `isEnabled` with the values `Default` and `Enabled`, so that we can keep the API the same, and just change the meaning of `Default` from `Disabled` to `Enabled`.

msugakov: @kurlov I want to make a blind suggestion without looking at the code about this:

> thanks for clarification, let's keep isEnabled and drop it if we will bump the CRD API version.

Could you please write a code comment for future-us on `isEnabled` that tells exactly what we learned here? Something in the spirit of

```go
// isEnabled is effectively no-op starting from the version 3.74.0. It should be removed when we
// bump API version of ACS custom resources. Removing it before is unsafe and may break compatibility.
// TODO: Remove isEnabled when bumping API version.
isEnabled attribute declared here
```

operator/apis/platform/v1alpha1/central_types.go
kurlov: You are right. Even though it's a small effort I would prefer to do it in the separate PR. I need to gather info about its usage

operator/pkg/central/values/translation/translation.go
kurlov: I've changed it to be a bit more robust if I understand its purpose correctly

operator/pkg/central/extensions/reconcile_central_db_password_test.go
```diff
@@ -212,11 +203,24 @@ func TestReconcileDBPassword(t *testing.T) {
 			Existing:      []*v1.Secret{canonicalPWSecretWithNoPassword},
 			ExpectedError: "secret must contain a non-empty",
 		},
+		"When using an external DB with specified password secret, that secret should be left untouched": {
+			Spec: v1alpha1.CentralSpec{
+				Central: &v1alpha1.CentralComponentSpec{
+					DB: &v1alpha1.CentralDBSpec{
+						IsEnabled:                v1alpha1.CentralDBEnabledPtr(v1alpha1.CentralDBEnabledDefault),
```
porridge, line 210:
```suggestion
```

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -213,11 +214,8 @@ type CentralDBSpec struct {
...
 const (
-	// CentralDBEnabledDefault configures the central to use rocksdb
-	// TODO(ROX-13557): in 3.74.0 this should say that Default enables PostgreSQL.
+	// CentralDBEnabledDefault configures the central to use PostgreSQL database.
 	CentralDBEnabledDefault CentralDBEnabled = "Default"
-	// CentralDBEnabledTrue configures the central to use a PostgreSQL database (Technology Preview)
-	CentralDBEnabledTrue CentralDBEnabled = "Enabled"
 )
 
 // CentralDBEnabledPtr return a pointer for the given CentralDBEnabled value
```
porridge, line 221: I believe this function will no longer needed once the other suggestions are applied, and can be removed?

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -157,14 +156,12 @@ func (c *CentralComponentSpec) GetAdminPasswordGenerationDisabled() bool {
 	return pointer.BoolPtrDerefOr(c.AdminPasswordGenerationDisabled, false)
 }
 
-// CentralDBEnabled returns true if central db should be created
-// TODO(ROX-13557): modify the logic to enable Postgres DB by default.
-func (c *CentralComponentSpec) CentralDBEnabled() bool {
-	if c == nil || c.DB == nil || c.DB.IsEnabled == nil {
+// IsExternalDB returns true if central DB is not managed by the Operator
+func (c *CentralComponentSpec) IsExternalDB() bool {
+	if c == nil || c.DB == nil {
```
porridge, line 161: FWIW, `IsExternal` also checks for nil, so this whole function can probably be simplified as `return c != nil && c.DB.IsExternal()`?

porridge (approved):
LGTM overall, just a few cleanups left!

porridge:
@kurlov in the light of discussion yesterday, please check with the postgres migration team whether this flip is still ok to merge now.

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -213,11 +214,8 @@ type CentralDBSpec struct {
 type CentralDBEnabled string
```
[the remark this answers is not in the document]
kurlov: Oh, it's quite popular feature request https://github.com/community/community/discussions/4452

vladbologa (approved):

PR #4383, opened 2023-01-12
ROX-12898: Adding Display Options dropdown select for Network Graph

alwayshooin:
the display options should work as expected (when one is toggled off, the graph should rerender w the specified badge turned off and vise versa)
image
image
image
image

I tried to not use the css file but couldn't find a way for the select option to directly get `display: flex` semantically 

I also moved the `EdgeStateSelect` to the components directory for semantics


ui/apps/platform/src/Containers/NetworkGraph/components/DisplayOptionsSelect.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react';
+import { Select, SelectVariant, SelectGroup, SelectOption } from '@patternfly/react-core';
+import { PficonNetworkRangeIcon } from '@patternfly/react-icons';
+
+import { ReactComponent as NoPolicyRules } from 'images/network-graph/no-policy-rules.svg';
```
vjwilson, line 5: funky syntax

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -171,6 +181,32 @@ function NetworkGraphPage() {
         }
     }, [edgeState, setModel, activeModel, extraneousFlowsModel]);
 
+    useEffect(() => {
+        // this is to update the display options visually for deployment nodes on the graph
+        if (model.nodes?.length) {
```
vjwilson, line 186: I hadn't thought of doing it this way, but seems valid.

vjwilson (approved):
/lgtm

PR #4386, opened 2023-01-13
ROX-14146: Add declarative config to roxctl

dhaus67:
## Description

This PR is a follow-up of #4321 and adds the declarative config setting to the central bundle generation as well as the helm output command for deriving local values.

The declarative config setting currently consists of two lists of mounts to add (config maps and secrets) to the Helm chart.

Additionally, the `helm derive-local-values` command has been adjusted to include deriving the values from an existing central.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

- see CI.

Additional manual tests:
```bash
roxctl k8s generate interactive
# Answer all the "questions", and add a list of config maps / secrets that exist / not exist.
Enter list of config maps to add as declarative configuration mounts in central (default: "[]"): cm-1, cm-2
Enter list of secrets to add as declarative configuration mounts in central (default: "[]"): sec-1, sec-2

# Install the bundle files on a cluster.
kubectl create -R -f central-bundle/central

# After installation, and the deployment successfully became ready, derive the helm configuration.
roxctl helm derive-local-values --output /path/to/some/dir central-services

# Inspect the public values, specifically the declarative configuration values, and observe that they match the initially provided values.
```
```yaml
...
declarativeConfiguration:
  mounts:
    configMaps:
    - "cm-1"
    - "cm-2
    secrets:
    - "sec-1"
    - "sec-2"
...
```

PR #4387, opened 2023-01-13
ROX-14145: Add declarative config to operator

dhaus67:
## Description

This PR is a follow-up of #4321 and adds the previously introduced settings for declarative config to the operator, specifically the central CR.

The fields on the central CR have been marked as "hidden" within the UI for now, similar to the telemetry settings as they are not yet officially released.

Besides that, the PR in itself is straightforward by adding the list of mount references to the CR and subsequently transforming these values to helm values.

## Checklist

## Testing Performed

- see CI.


dhaus67:
* master
  * **PR #4321** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4321" target="_blank">Graphite</a> 
    * **PR #4387** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4387" target="_blank">Graphite</a>  👈
    * **PR #4386** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4386" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4387?utm_source=stack-comment).

operator/pkg/central/values/translation/translation.go
```diff
@@ -249,6 +251,30 @@ func getTelemetryValues(t *platform.Telemetry) *translation.ValuesBuilder {
 	return &tv
 }
 
+func getDeclarativeConfigurationValues(c *platform.DeclarativeConfiguration) *translation.ValuesBuilder {
+	declarativeConfig := translation.NewValuesBuilder()
+	if c == nil {
+		return &declarativeConfig
```
ivan-degtiarenko, line 257: shouldn't we put empty 
```
mounts:
      configMaps: []
      secrets: []
```
even in that case?

operator/config/manifests/bases/rhacs-operator.clusterserviceversion.yaml
```diff
@@ -295,6 +295,11 @@ spec:
         path: misc
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
+      - description: Configures resources within Central in a declarative manner.
+        displayName: Declarative Configuration
+        path: central.declarativeConfiguration
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:hidden
```
ivan-degtiarenko, line 302: what does adding these descriptors do?

operator/tests/central/basic/100-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
dhaus67: So, the number is equivalent to the applied update version on the CR.
Since we create a new version here, I chose to bump the number, and keep the deletion step as the last step applied, hence bumping this one to `100-...`.

operator/pkg/central/values/translation/translation.go
dhaus67: Yeah I thought about this one as well, but looking at the other places we do something similar, it seems that's how we do things (e.g. telemetry values, secured cluster values). So instead of specifying any default values, we simply set it to nil and assume that the helm chart can handle if these values aren't set (which, it certainly can _and_ should).

operator/config/manifests/bases/rhacs-operator.clusterserviceversion.yaml
dhaus67: They are called [API Markers](https://sdk.operatorframework.io/docs/building-operators/golang/references/markers/) and impact the UI that is rendered when you want to e.g. create a CR.
This marker specifically allows us to hide the field, since its currently not released and shouldn't be used by users (effectively, it won't do anything if used, but we still shouldn't show it _yet_ as prominently). We've done similar things for the telemetry values.

Additionaly,[ there are other markes like the config map / secret one](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#3-k8sresourceprefix), which automatically provide users of the UI with a selection of config maps / secrets, so its quite handy.

ivan-degtiarenko: Cool, today I learned something new :)

operator/pkg/central/values/translation/translation.go
ivan-degtiarenko: Sounds good. Do you think it makes sense to test this case in `translation_test`? If that's not too much work

ivan-degtiarenko (approved):
LGTM modulo test comment

operator/pkg/central/values/translation/translation.go
dhaus67: Ain't this what we test with[ the empty spec test](https://github.com/stackrox/stackrox/blob/65820256abbcb6cc2809f7a6ece8ca7729dd904a/operator/pkg/central/values/translation/translation_test.go#L52-L68) in there?
The test that covers that the helm chart can handle these "empty" values is within [the Helm chart PR](https://github.com/stackrox/stackrox/pull/4321/files#diff-453f50cb4de0a2077bd7029131c35907ed20743c69b421e26ed095396477e830R1-R6).

PR #4391, opened 2023-01-13
Dashrews/ROX-14209 update sensor to ensure jsonpb unmarshal has AllowUnknownFields set to true

dashrews78:
## Description

Trying to trickle these through as they are fairly quick after https://github.com/stackrox/stackrox/pull/4316. Simply use the JSONUnmarshaler instead of jsonpb directly to ensure that AllowUnknownFields is set for unmarshaling to avoid compatibility issues.

## Checklist

## Testing Performed

Simply using a method that is already tested in pkg/jsonutil/conversion_test.go.  Added test to the additional wrapper method

### DIFF
--- a/pkg/jsonutil/conversion.go
+++ b/pkg/jsonutil/conversion.go
@@ -2,6 +2,7 @@ package jsonutil
 
 import (
 	"bytes"
+	"io"
 	"regexp"
 	"strings"
 
@@ -40,6 +41,11 @@ func JSONToProto(json string, m proto.Message) error {
 // JSONBytesToProto converts bytes containing JSON into a proto message.
 func JSONBytesToProto(contents []byte, m proto.Message) error {
 	return JSONUnmarshaler().Unmarshal(bytes.NewReader(contents), m)
+}
+
+// JSONReaderToProto converts bytes from a reader containing JSON into a proto message.
+func JSONReaderToProto(reader io.Reader, m proto.Message) error {
+	return JSONUnmarshaler().Unmarshal(reader, m)
 }
 
 // ProtoToJSON converts a proto message into a string containing JSON.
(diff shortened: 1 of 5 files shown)


pkg/jsonutil/conversion_test.go
```diff
@@ -111,7 +112,13 @@ func TestNoErrorOnUnknownAttribute(t *testing.T) {
...
+	jsonBytes := []byte(json)
+	err = JSONBytesToProto(jsonBytes, &proto)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "6500", proto.GetId())
+
+	err = JSONReaderToProto(bytes.NewReader(jsonBytes), &proto)
 
 	assert.NoError(t, err)
 	assert.Equal(t, "6500", proto.GetId())
```
msugakov, line 124: I suggest to unbreak lines to highlight different cases better
```suggestion
	jsonBytes := []byte(json)

	err = JSONBytesToProto(jsonBytes, &proto)
	assert.NoError(t, err)
	assert.Equal(t, "6500", proto.GetId())

	err = JSONReaderToProto(bytes.NewReader(jsonBytes), &proto)
	assert.NoError(t, err)
	assert.Equal(t, "6500", proto.GetId())
```

pkg/jsonutil/conversion.go
```diff
@@ -42,6 +43,11 @@ func JSONBytesToProto(contents []byte, m proto.Message) error {
 	return JSONUnmarshaler().Unmarshal(bytes.NewReader(contents), m)
```
msugakov, line 43: You can potentially change `JSONBytesToProto` and `JSONToProto` to function via newly created `JSONReaderToProto`:

```suggestion
	return JSONReaderToProto(bytes.NewReader(contents), m)
```

msugakov (approved):

PR #4414, opened 2023-01-16
ROX-14400: Deprecate BuildDate

janisz:
## Description

BuildTimestampt and BuildDate will be removed in future release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,6 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-14400: `BuildDate` is deprecated and will be removed in future releases. It won't be returned by
 `/debug/versions.json` endpoint and not available to use in `roxctl` helm templates.
 
-
-
 ### Required Actions
 - The permission `WorkflowAdministration` will replace `Policy, VulnerabilityReports` in permission sets starting with the 3.76 release.
   You should preemptively start replacing the `Policy` and `VulnerabilityReports` resources within your permission sets in favor of `WorkflowAdministration`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-12620: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
   - The permission `WorkflowAdministration` will deprecate the permissions `Policy, VulnerabilityReports`.
 
+- ROX-14400: `BuildDate` is deprecated and will be removed in future releases. It won't be returned by
+`/debug/versions.json` endpoint and not available to use in `roxctl` helm templates.
+
+
+
 ### Required Actions
 - The permission `WorkflowAdministration` will replace `Policy, VulnerabilityReports` in permission sets starting with the 3.76 release.
   You should preemptively start replacing the `Policy` and `VulnerabilityReports` resources within your permission sets in favor of `WorkflowAdministration`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,8 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-12620: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
   - The permission `WorkflowAdministration` will deprecate the permissions `Policy, VulnerabilityReports`.
 
-- ROX-14400: `BuildDate` is deprecated and will be removed in future releases. It won't be returned by
-`/debug/versions.json` endpoint and not available to use in `roxctl` helm templates.
+- ROX-14400: product `BuildDate` attribute is deprecated and will be removed in `3.75` release. It won't be returned by
+`/debug/versions.json` endpoint and `roxctl version --json` command.
 
 ### Required Actions
 - The permission `WorkflowAdministration` will replace `Policy, VulnerabilityReports` in permission sets starting with the 3.76 release.
--- a/pkg/buildinfo/buildinfo.go
+++ b/pkg/buildinfo/buildinfo.go
@@ -23,6 +23,7 @@ const (
 // CAVEAT: This function panics if no build timestamp information is available.
 //
 // Deprecated: It will be removed in 3.75. Please do not use it.
+// TODO(ROX-14336): delete it
 func BuildTimestamp() time.Time {
 	if timestamp.BuildTimestampParsingErr != nil {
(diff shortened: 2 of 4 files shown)


janisz:
* master
  * **PR #4414** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4414" target="_blank">Graphite</a>  👈
    * **PR #4345** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4345" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4414?utm_source=stack-comment).

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-12620: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
   - The permission `WorkflowAdministration` will deprecate the permissions `Policy, VulnerabilityReports`.
 
+- ROX-14400: `BuildDate` is deprecated and will be removed in future releases. It won't be returned by
```
msugakov, line 22: 1. Normally, we should announce in which specific release the thing will be removed.
2. Which ticket will be used for the actual removal?

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
...
+`/debug/versions.json` endpoint and not available to use in `roxctl` helm templates.
```
msugakov, line 23: Which helm templates?
I thought it's only exposed from `roxctl version --json`.

janisz: If user use
`roxctl helm output central-services --output-dir /tmp/helm --debug-path <path to custom templates> --debug`
and in template uses `[< .Versions.BuildDate >]` then it will not work

CHANGELOG.md
janisz: 1. 3.75 should be OK
2. ROX-14336

pkg/buildinfo/buildinfo.go
```diff
@@ -21,6 +21,8 @@ const (
 
 // BuildTimestamp returns the time when this build was created.
 // CAVEAT: This function panics if no build timestamp information is available.
+//
+// Deprecated: It will be removed in next release. Please do not use it.
```
dhaus67, line 25:
```suggestion
// Deprecated: It will be removed in 3.75. Please do not use it.
```

pkg/buildinfo/internal/timestamp/timestamp.go
```diff
@@ -9,8 +9,10 @@ var (
 	buildTimestampUnixSecs string //XDef:BUILD_TIMESTAMP
 
 	// BuildTimestamp is the time when this binary was built.
+	// Deprecated: It will be removed in next release. Please do not use it.
```
dhaus67, line 12:
```suggestion
	// Deprecated: It will be removed in 3.75. Please do not use it.
```

pkg/buildinfo/internal/timestamp/timestamp.go
```diff
@@ -9,8 +9,10 @@ var (
...
 	BuildTimestamp time.Time
 	// BuildTimestampParsingErr is the error encountered when parsing the build timestamp (if any).
+	// Deprecated: It will be removed in next release. Please do not use it.
```
dhaus67, line 15:
```suggestion
	// Deprecated: It will be removed in 3.75. Please do not use it.
```

pkg/version/version.go
```diff
@@ -29,6 +29,7 @@ func getCollectorVersion() string {
 
 // Versions represents a collection of various pieces of version information.
 type Versions struct {
+	// Deprecated: BuildDate will be removed in next release. Please do not use it.
```
dhaus67, line 32:
```suggestion
	// Deprecated: BuildDate will be removed in 3.75. Please do not use it.
```

CHANGELOG.md
dhaus67: I'm not sure how used it is by users, but we typically have a two release deprecation cycle. Looking at how minor this is though (it's somewhat tied to `debug`), I'm fine with 3.75.
We've added the `Required Actions` section to be more clear about remediations that may require user interactions, for this I don't really think it makes too much sense to add something in that section, WDYT?

dhaus67 (approved):

CHANGELOG.md
janisz: > I don't really think it makes too much sense to add something in that section

I agree this could be an overkill. Anyway if anybody is using this I'm curious why.

msugakov: > I'm not sure how used it is by users, but we typically have a two release deprecation cycle.

@dhaus67 Do you know if this is documented somewhere? If not, we should document it. My first thought is to put it in the beginning of https://github.com/stackrox/stackrox/blob/master/CHANGELOG.md

CHANGELOG.md
msugakov: > in template uses

I believe we control templates and don't let users modify templates (although they can provide different variables), especially at "`.htpl`" level (`[<` `>]` tags). Therefore, `.Versions.BuildDate` becomes unavailable for us, ACS developers.
Given that we don't seem to use `.Versions.BuildDate`, it does not deserve to be mentioned.

Moreover, the changelog is the communication tool with ACS/StackRox **users**, not developers. We put things that are significant for users to be aware of. Then, the corresponding part of the changelog is converted to product Release Notes by the documentation team. That's how users get to know it.

pkg/version/version.go
```diff
@@ -29,6 +29,7 @@ func getCollectorVersion() string {
 
 // Versions represents a collection of various pieces of version information.
 type Versions struct {
+	// Deprecated: BuildDate will be removed in 3.75. Please do not use it.
 	BuildDate time.Time `json:"BuildDate"`
```
msugakov, line 33:
```suggestion
	// Deprecated: BuildDate will be removed in 3.75. Please do not use it.
	// TODO(ROX-14336): delete it
	BuildDate time.Time `json:"BuildDate"`
```

msugakov (changes requested):
Suggesting alterations to the CHANGELOG given what we discussed.
Also it's good to add `// TODO(ROX-14336): ...` comments that will prevent you from merging any PR that mentions `ROX-14336` but does not delete these comments.

janisz:
@msugakov Thanks for comments. I applied them.

msugakov (approved):

PR #4415, opened 2023-01-16
ROX-13799: Expand Prometheus metrics for RHCOS

Maddosaurus:
## Description

Introducing Node Scanning also enables us to generate additional metrics that can be collected. This PR adds them.
Most notably, `Compliance` now exposes Metrics, which it didn't so far. As it runs together with `Collector` in the same Pod, a non-standard port `9091` was chosen to have both metrics collections available in parallel.

This PR covers both scenarios, deploying with `MONITORING_SUPPORT=true ./deploy/.../deploy.sh` as well as the Helm route when setting `exposeMonitoring: true` for `Collector` in the values.
The former is done by patching the deployed resources in `monitoring.sh`, while the latter is done by updating the Helm templates with new conditional includes.

As part of the review, I would be interested in discussing additional metrics reviewers can think of.

## Checklist

## Testing Performed

Testing was done locally on Colima as well as Infra Clusters running OpenShift 4.11.
On both  platforms, the code was deployed with either the env var set or the Helm var set, and I verified that in all cases I was able to see the exposed metrics, especially of the new Compliance metrics endpoint.


vikin91 (changes requested):
Looks good but there are some bits that need to be corrected - see comments.

Regarding the additional metrics here are my ideas:

Labels:
- error yes/no - allows us to see how much a regular and erroneous analysis take
- namespace (see the OS and version)
- cluster name

Metrics:
- number of content sets
- node inventory size in bytes (important for sensor-central network volume estimation)
- number of language components (for the future?)
- time duration between node scans (if possible)
- number of scans since the last start of container `scans_total`
- number of retries (only after [ROX-13164](https://issues.redhat.com/browse/ROX-13164))

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -49,6 +56,7 @@ func (n *NodeInventoryCollector) Scan(nodeName string) (*storage.NodeInventory,
 		Notes:      []storage.NodeInventory_Note{storage.NodeInventory_LANGUAGE_CVES_UNAVAILABLE},
 	}
 
+	metrics.ObserveNodeInventoryScan(m, endTime)
```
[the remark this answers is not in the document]
Maddosaurus: That is true, but as I rely on the nodeName, I have to do it rather late. 
I could also not include the name as label, but that information seems rather crucial.

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+var (
+	numberOfRHELPackages = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.ComplianceSubsystem.String(),
+		Name:      "num_packages_in_inventory",
+		Help:      "Number of packages discovered by the last Node Inventory (per Node)",
+	},
+		[]string{
+			// The Node this scan belongs to
+			"node_name",
```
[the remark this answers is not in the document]
Maddosaurus: AFAIK a pod doesn't know the cluster it runs on, so I don't think we have this available in collector.
I also checked the [Kubernetes Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/), which doesn't include it either.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 
 - ROX-13814: A new "Public Kubernetes Registry" image integration is now available as a replacement
   for the (now deprecated) "Public Kubernetes GCR" image integration.
+- ROX-13799: Introduce Prometheus metrics for Node Scanning.
```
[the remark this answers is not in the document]
Maddosaurus: That is a good idea. I've removed the CL entry for now

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+		})
+
+	scanTime = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.ComplianceSubsystem.String(),
+		Name:      "inventory_scan_time",
```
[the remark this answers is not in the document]
Maddosaurus: Thanks for the suggestion. I've switched the name, but w.r.t. milliseconds I have a question:
The [Metric and label naming](https://prometheus.io/docs/practices/naming/#base-units) page suggests to use base units, which would be seconds for time, which is why I chose seconds here. 
I concur that milliseconds might be more appropriate, though. WDYT?

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/metrics"
+)
+
+var (
+	numberOfRHELPackages = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
Maddosaurus: I chose a Gauge, as it was described as a value that can raise and lower, as opposed to a Counter, which is monotonically increasing. The number of packages is something that might fall. 

On the other hand, we don't really expect the package count on the same node to ever fall. It might increase though, especially with layering in mind.

compliance/collection/metrics/metrics.go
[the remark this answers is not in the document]
vikin91: Seconds are fine as long as we make sure that they are not rounded to integers. Seeing `0.035` is fine, while `0` is not :)

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+		Help:      "Number of content sets discovered by the last Node Inventory (per Node)",
+	},
+		[]string{
+			// The Node this scan belongs to
+			"node_name",
+			// The OS name and version of the Node
+			"os_namespace",
+		})
+
+	scanDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
vikin91, line 40: In the cluster I see that 100% of scans take between 2.5 and 5 seconds (query:`rox_compliance_inventory_scan_duration_seconds_bucket{node_name="mm-01-27-scientific-f-cvdq4-worker-b-ndhlf"})`. This is long. Could we double check that the values are correct (especially seconds vs milliseconds error should be avoided)?

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	},
+		[]string{
+			// The Node this scan belongs to
+			"node_name",
+		})
+
+	protobufMessageSize = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.ComplianceSubsystem.String(),
+		Name:      "protobuf_inventory_message_size",
```
vikin91, line 67: Shall we add unit name as postfix to the name?

vikin91: We need to configure histogram buckets here as 100% of observations fall into the `le="+Inf"` bucket.
I would propose the following buckets: 500B, 1kB, 10kB, 50kB, 100kB, 500kB, 1000kB,...

compliance/collection/metrics/metrics.go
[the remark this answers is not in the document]
vikin91: Okay, let's keep it as gauge.

sensor/common/detector/metrics/metrics.go
```diff
@@ -40,6 +41,16 @@ var (
...
+	receivedNodeInventory = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.SensorSubsystem.String(),
+		Name:      "node_inventories_received_total",
+		Help:      "Total number of Node Inventories received by this sensor",
+	},
+		[]string{
+			// Name of the node sending an inventory
+			"node_name",
+		})
```
vikin91, line 53: Cool that you though about sensor!

sensor/common/detector/metrics/metrics.go
```diff
@@ -61,6 +72,13 @@ func ObserveNetworkPolicyStoreEvent(event, namespace string, numSelectors int) {
 	}).Inc()
 }
 
+// ObserveReceivedNodeInventory observes the metric.
+func ObserveReceivedNodeInventory(inventory *storage.NodeInventory) {
+	receivedNodeInventory.With(prometheus.Labels{
+		"node_name": inventory.NodeName,
```
vikin91, line 78: Please getters and nil protection here as well.

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -21,10 +24,18 @@ type NodeInventoryCollector struct {
...
+	metrics.ObserveScansTotal(nodeName)
+	startTime := time.Now()
+
 	// uncertifiedRHEL is set to false, as scans are only supported on RHCOS for now,
 	// which only exists in certified versions
 	componentsHost, err := nodes.Analyze(nodeName, "/host/", nodes.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: true})
+
+	scanDuration := time.Since(startTime)
+	metrics.ObserveScanDuration(scanDuration, nodeName, err)
+	log.Debug("Finished collecting node inventory")
```
vikin91, line 37: Not sure we need this at the debug level. It looks more like println-debugging. I would vote to remove this.

deploy/common/monitoring.sh
```diff
@@ -22,4 +24,22 @@ spec:
     - {key: app, operator: In, values: [central, sensor]}
   policyTypes:
   - Ingress
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
+    app.kubernetes.io/name: stackrox
+  name: allow-cmonitoring
```
vikin91, line 33: Lets use an bit more informative name here: `allow-compliance-monitoring`?

deploy/charts/monitoring/templates/grafana.yaml
```diff
@@ -70,3 +70,6 @@ data:
 {{ .Files.Get "dashboards/cluster-health.json" | indent 4 }}
   dataplane.json: |-
 {{ .Files.Get "dashboards/dataplane.json" | indent 4 }}
+  node-scanning.json: |-
+{{ .Files.Get "dashboards/node-scanning.json" | indent 4 }}
+
```
vikin91, line 75: I thought we are not committing the Grafana dashboard 🤔

vikin91: Anyway, I looked at the dashboards and they are fine to be added. Of course they could be more fancy but maybe in a separate PR.

vikin91 (changes requested):
Looks much better, but I still want to request some changes.

The blocker for me are not using the getters for proto types and insufficient nil protection.
The custom bucket config and check why the scan takes so long is a strong **should-have**.

compliance/collection/metrics/metrics.go
Maddosaurus: Thanks for the insight w.r.t. bucket sizes. This is one of the things I wanted to test experimentally.
Now that we know, I have added your proposed buckets. 
I also configured buckets for the scan time hist to be `[]float64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20}`, as I expect the scan time to be a bit higher on production nodes that are running workloads.

compliance/collection/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+		[]string{
+			// The Node this scan belongs to
+			"node_name",
+		})
+)
+
+// ObserveNodeInventoryScan observes the metric.
+func ObserveNodeInventoryScan(inventory *storage.NodeInventory) {
+	rhelPackageCount := 0
+	if inventory.Components.RhelComponents != nil {
```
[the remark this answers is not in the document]
Maddosaurus: Thanks for the reminder! I have switched to getters and added `nil` checks

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: This was added to get a rough estimate of scan times. Now that we have the metrics, we might not need this anymore.
But if we remove this, I would also vote to remove the debug message noting the beginning of a Node Scan as well.

compliance/collection/metrics/metrics.go
Maddosaurus: I've checked the log messages in a collector, and it looks correct. We definitely need to  keep an eye on how this develops. Multiple seconds for what we're doing today seems rather long-ish.

```
$ ks logs collector-ss48q compliance | grep collecting
collection/nodeinventorizer: 2023/01/30 10:05:13.789081 nodeinventory.go:27: Debug: Start collecting node inventory
collection/nodeinventorizer: 2023/01/30 10:05:18.750015 nodeinventory.go:37: Debug: Finished collecting node inventory
```

compliance/collection/metrics/metrics.go
vikin91: That is fine, but note when we have scan times above 20, we will be blind how large they are as they fall into the bucket of `20 to +Inf`. I would suggest to always include some exponential buckets at the end, for example: `[]float64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 50, 100, 500, 1000}`

compliance/collection/nodeinventorizer/nodeinventory.go
vikin91: I see. Then I would suggest to rephrase this to sth like: `log.Debugf("Collecting node inventory took %ss", scanDuation.Seconds().String())`

deploy/charts/monitoring/templates/grafana.yaml
vikin91: The 30 minutes I understand, I was thinking that it will not be committed.

I think you should still consider not committing it, as when the FF is disabled, we will see an empty dashboard that may look broken to people who would be looking at it. We may keep the code and add the dashboard shortly before enabling the feature flag.

compliance/collection/nodeinventorizer/nodeinventory.go
Maddosaurus: That's a great idea, let me rework this

compliance/collection/metrics/metrics.go
Maddosaurus: That's a great practice to stick to, let me update

deploy/charts/monitoring/templates/grafana.yaml
Maddosaurus: That's a very good point. I've removed them from the PR and set up a new ticket (ROX-14626) to add them back before we pull the FF

vikin91 (approved):
I think it looks good now! Thanks for improving!

Before you merge, I would kindly ask you to deploy it on the test cluster with FF enabled **and disabled** to check whether the nil-checks are working correctly (especially when inventory==nil). We are close to the code freeze and want to avoid any last-minute fixes for something that could be avoided.

compliance/collection/metrics/metrics.go
Maddosaurus: After updating to a better debug log message, I can confirm the observation on an OpenShift Node:
`Debug: Collecting Node Inventory took 2.882744 seconds`

deploy/charts/monitoring/templates/grafana.yaml
Maddosaurus: > Before you merge, I would kindly ask you to deploy it on the test cluster with FF enabled **and disabled** to check whether the nil-checks are working correctly (especially when inventory==nil).

I have tested the latest version by switching the feature flag to off in the test cluster, followed by a clean teardown and redeploy with disabled feature flag.
All looks stable, I did not observe any crashes,  neither of sensor nor of compliance.

Maddosaurus:
> Before you merge, I would kindly ask you to deploy it on the test cluster with FF enabled **and disabled** to check whether the nil-checks are working correctly (especially when inventory==nil).

I have tested the latest version by switching the feature flag to off in the test cluster, followed by a clean teardown and redeploy with disabled feature flag.
All looks stable, I did not observe any crashes,  neither of sensor nor of compliance.

deploy/common/monitoring.sh
```diff
@@ -2,6 +2,8 @@
 
 kubectl -n stackrox patch svc/sensor -p '{"spec":{"ports":[{"name":"monitoring","port":9090,"protocol":"TCP","targetPort":9090}]}}'
 kubectl -n stackrox patch svc/central -p '{"spec":{"ports":[{"name":"monitoring","port":9090,"protocol":"TCP","targetPort":9090}]}}'
+kubectl -n stackrox patch daemonset/collector --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/1/ports", "value":[{"containerPort":9091,"name":"cmonitor","protocol":"TCP"}]}]'
+kubectl -n stackrox set env daemonsets/collector --containers="compliance" ROX_METRICS_PORT=":9091"
```
vikin91, line 6: I think this line is note executed in CI and the `gke-nongroovy-e2e-tests` try to pull up the compliance with monitoring on port 9090 which results in a panic:

```
No certificates found in /usr/local/share/ca-certificates
No certificates found in /etc/pki/injected-ca-trust
main: 2023/01/30 13:22:25.896767 main.go:253: Info: Running StackRox Version: 3.73.x-588-g8bbc45ad6a
main: 2023/01/30 13:22:25.897803 main.go:265: Info: Initialized Sensor gRPC stream connection
pkg/metrics: 2023/01/30 13:22:25.900635 http_server.go:64: Panic: Unexpected termination of metrics HTTP server: listen tcp :9090: bind: address already in use
panic: Unexpected termination of metrics HTTP server: listen tcp :9090: bind: address already in use

goroutine 13 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0008bc2c0, {0x0, 0x0, 0x0})
	go.uber.org/zap@v1.23.0/zapcore/entry.go:230 +0x486
go.uber.org/zap.(*SugaredLogger).log(0xc000850008, 0x4, {0x2169e65?, 0x0?}, {0xc00008bfa8?, 0x0?, 0x0?}, {0x0, 0x0, 0x0})
	go.uber.org/zap@v1.23.0/sugar.go:234 +0x13b
go.uber.org/zap.(*SugaredLogger).Panicf(...)
	go.uber.org/zap@v1.23.0/sugar.go:159
github.com/stackrox/rox/pkg/metrics.runForever(0x0?)
	github.com/stackrox/rox/pkg/metrics/http_server.go:64 +0x71
created by github.com/stackrox/rox/pkg/metrics.(*HTTPServer).RunForever
	github.com/stackrox/rox/pkg/metrics/http_server.go:58 +0x1e5
```

Maddosaurus: Ah, thank you for the investigation!
I have moved the port definition into the DaemonSet itself by default, as we always want to have a different port for the compliance metrics.

PR #4425, opened 2023-01-16
Feature/process listening on port store jv address yann review

JoukoVirtanen:
Made some changes requested by Yann.


rhybrillou:
Comparing with the current state of PR 4328, I do not think this follow-up PR is needed anymore.
The only change I'm not sure if I should pick from this PR or 4328 is the content of a comment in TestPLOPReopen (I did not check in details the difference in the PLOP postgres store code, though).

PR #4426, opened 2023-01-17
ROX-12423: Fix Artifactory Integration EOF Error

dcaravel:
## Description
Setting up a new `JFrog Artifactory` integration in Stackrox when Artifactory is configured to use the `Repository Path` Docker Access Method will trigger an error: `registry integration: EOF: invalid arguments: invalid arguments`, this error does not occur when using the `Generic Docker` integration.

The Artifactory and Docker integrations differ only in the method which performs the initial test; this is rumored to be due to Artifactory historically not implementing the endpoint called by [docker-registry-client's `Ping()`](https://github.com/stackrox/docker-registry-client/blob/3c566c2fa4277f181f06167309eb6f4b02de9b77/registry/registry.go#L119-L120) method. 

This PR changes the Artifactory integration's `Test()` method to mimic that of the Docker integration

## Checklist

No additional unit or regression tests were added - that would require setting up multiple private instances of Artifactory during test execution.  Additionally, not all settings necessary to re-create the appropriate test conditions can be  automated easily during Artifactory setup.

CHANGELOG was not updated as it appears bug fixes are not applicable based on reviewing recent past.

There are no changes to upgrade steps or user documentation necessary as part of this PR

## Testing Performed

Different versions of [JFrog Container Registry](https://www.jfrog.com/confluence/display/JFROG/JFrog+Container+Registry) (Artifactory) were installed and settings modified to duplicate the error as well as learn conditions under which the error would not occur. Based on testing it was determined the Artifactory [Docker Access Method](https://www.jfrog.com/confluence/display/RTF6X/Getting+Started+with+Artifactory+as+a+Docker+Registry#GettingStartedwithArtifactoryasaDockerRegistry-ConfiguringArtifactory.1) has a direct impact on the occurrence of this error. No other settings tested had an impact

Repository Path Access Method | Sub Domain Access Method
--- | --- 
image | image

The oldest (`v7.23.3`) and newest (`v7.49.5`) supported versions of JFrog Container Registry (JCR), per [End of Life](https://www.jfrog.com/confluence/display/JFROG/Artifactory+End+of+Life) doc, were installed and tested with the different access methods.

Installation of JCR was performed via the official [JFrog Container Registry Helm Chart](https://artifacthub.io/packages/helm/jfrog/artifactory-jcr) using the following values:

```yaml
artifactory:
  ingress:
    enabled: false
    hosts:
    - "arti.local"
    - "docker.arti.local"
  postgresql:
    enabled: false
  artifactory:
    persistence:
      enabled: false
  nginx:
    tlsSecretName: "nginx-tls"
```
`hosts` files were modified to enable resolution of `arti.local` and `docker.arti.local` as well as custom certs created/setup to allow for TLS validation (via `openssl`)

Integration setup was re-tested post change:

Repository Path Access Method | Sub Domain Access Method
--- | --- 
image | image

Additionally `roxctl image scan` was executed for various images for each of the Docker Access Methods - which worked as expected.

### DIFF
--- a/pkg/registries/artifactory/artifactory.go
+++ b/pkg/registries/artifactory/artifactory.go
@@ -34,7 +34,7 @@ func newRegistry(integration *storage.ImageIntegration) (types.Registry, error) 
 
 // Test implements a valid Test function for Artifactory
 func (r *registry) Test() error {
-	_, err := r.Client.Repositories()
+	err := r.Client.Ping()
 	if err != nil {
 		logger.Errorf("error testing Artifactory integration: %v", err)
 	}
--- a/pkg/registries/artifactory/artifactory.go
+++ b/pkg/registries/artifactory/artifactory.go
@@ -9,7 +9,6 @@ import (
 // Creator provides the type and registries.Creator to add to the registry of image registries.
 func Creator() (string, func(integration *storage.ImageIntegration) (types.Registry, error)) {
 	return "artifactory", func(integration *storage.ImageIntegration) (types.Registry, error) {
-		reg, err := docker.NewDockerRegistry(integration)
-		return reg, err
+		return docker.NewDockerRegistry(integration)
 	}
 }
--- a/pkg/registries/artifactory/artifactory.go
+++ b/pkg/registries/artifactory/artifactory.go
@@ -2,41 +2,14 @@ package artifactory
 
 import (
 	"github.com/stackrox/rox/generated/storage"
-	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/registries/docker"
 	"github.com/stackrox/rox/pkg/registries/types"
 )
 
-var (
-	logger = logging.LoggerForModule()
-)
-
 // Creator provides the type and registries.Creator to add to the registry of image registries.
 func Creator() (string, func(integration *storage.ImageIntegration) (types.Registry, error)) {
-	return "artifactory", newRegistry
-}
-
-var _ types.Registry = (*registry)(nil)
-
-type registry struct {
-	*docker.Registry
-}
-
-func newRegistry(integration *storage.ImageIntegration) (types.Registry, error) {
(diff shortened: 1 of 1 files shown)


pkg/registries/artifactory/artifactory.go
```diff
@@ -34,7 +34,7 @@ func newRegistry(integration *storage.ImageIntegration) (types.Registry, error)
 
 // Test implements a valid Test function for Artifactory
 func (r *registry) Test() error {
-	_, err := r.Client.Repositories()
+	err := r.Client.Ping()
```
RTann, line 37: I think we can simplify this even further by just removing this function completely. That way, we just use the `Test()` function as implemented for generic docker registries: https://github.com/stackrox/stackrox/blob/3.73.1/pkg/registries/docker/docker.go#L208

Note this works because the following indicates that the artifactory registry **is** a docker registry:

```
type registry struct {
	*docker.Registry
}
```

RTann: Actually, we probably no longer need that struct and we can just do what azure does: https://github.com/stackrox/stackrox/blob/3.73.1/pkg/registries/azure/azure.go

dcaravel: I considered that and my thought was the error logging specifically mentioning 'artifactory' would be helpful during troubleshooting, but looking at the other integrations like the one you shared - it appears most do as you suggest.  I'll make that change.

RTann (approved):
I think you can just `return` without the variables, but up to you

dcaravel:
> I think you can just `return` without the variables, but up to you

Certainly a debated topic :)

In some cases debugging is easier having the variables, in this case can insert a breakpoint in the `registries.factory` so not much benefit.

Also see it done without the vars in [registry.go](https://github.com/stackrox/stackrox/blob/master/pkg/registries/artifactregistry/registry.go#L12)

I went ahead and made the change, thanks Ross!

PR #4435, opened 2023-01-17
ROX-13646: Fix endpoints test flake

kovayur:
## Description
Replace port forwarding with load balancer for the problematic endpoint.

## Checklist

## Testing Performed
Tested locally on colima with the following settings:
```
colima start -c 8 -m 16 --kubernetes --kubernetes-disable="traefik" --network-address 
```


tests/e2e/lib.sh
```diff
@@ -219,15 +219,18 @@ patch_resources_for_test() {
...
     kubectl -n stackrox patch svc central-loadbalancer --patch "$(cat "$TEST_ROOT"/tests/e2e/yaml/endpoints-test-lb-patch.yaml)"
     kubectl -n stackrox apply -f "$TEST_ROOT/tests/e2e/yaml/endpoints-test-netpol.yaml"
-    # shellcheck disable=SC2034
-    for i in $(seq 1 20); do
-        if curl -sk --fail "https://${API_HOSTNAME}:8446/v1/metadata" &>/dev/null; then
-            return
-        fi
-        sleep 1
+
+   for target_port in 8444 8446; do
```
porridge, line 223: Maybe mention here what these two are for and it's sufficient for just one to be available?

porridge (commented):
I notice the tests are failing, is this related?

tests/e2e/lib.sh
kovayur: It was done by mistake. Thanks for highlighting!

kovayur:
Port forwarding has been entirely replaced by LB.

tests/endpoints_test.go
```diff
@@ -49,24 +49,20 @@ var (
 	}
 )
 
-func endpointForTargetPort(targetPort uint16, accessViaLoadBalancer bool) string {
```
kovayur, line 52: Remove `accessViaLoadBalancer` flag since it is always `true` now

janisz (approved):

tests/e2e/lib.sh
```diff
@@ -219,37 +218,25 @@ patch_resources_for_test() {
...
+        check_endpoint_availability "$target_port"
+    done
+}
+
+check_endpoint_availability() {
+    local target_port="$1"
     # shellcheck disable=SC2034
     for i in $(seq 1 20); do
-        if curl -sk --fail "https://${API_HOSTNAME}:8446/v1/metadata" &>/dev/null; then
+        if echo "Endpoint check" 2>/dev/null > /dev/tcp/"${API_HOSTNAME}"/"${target_port}"; then
```
porridge, line 231: I vaguely remember some distros disable this `/dev/tcp` functionality in bash, so this might become a portability problem, but we can always just revert, right? :wink:

porridge (approved):
:heart_eyes:

PR #4436, opened 2023-01-17
ROX-14345: Added ability to toggle alerting on baseline violations

sachaudh:
## Description

This PR adds the ability to toggle alerting on baseline violations within the deployment baselines tab

https://user-images.githubusercontent.com/4805485/212948012-2a0105db-f731-48d1-95b0-eb984c0b6fc4.mov

Note: Flickering happens because we're loading and we show a spinner, but it's pretty quick. I feel like maybe it would be better if the table becomes disabled and we show an overlay with a spinner, but not sure that is consistent with our other tables. Although, I feel like it would feel smoother

## Checklist
- ~[ ] Determined and documented upgrade steps`


vjwilson (approved):
/lgtm

PR #4447, opened 2023-01-18
Prepare Role resource grouping with Access

rhybrillou:
## Description

As part of the effort to simplify the permission management in the product, the `Role` resource will be grouped with the `Access` one.
This change announces the grouping and flags the code portions to alter in order to make this change happen.

## Checklist

## Testing Performed

Mostly comment addition in the code.
CI should be enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Deprecated Features
 - ROX-12620: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
   - The permission `WorkflowAdministration` will deprecate the permissions `Policy, VulnerabilityReports`.
+- ROX-14398: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
+  - The permission `Access` will deprecate the permissions `Role`.
+  - The default role `Scope Manager` will be removed.
 
 - ROX-14400: product `BuildDate` attribute is deprecated and will be removed in `3.75` release. It won't be returned by
 `/debug/versions.json` endpoint and `roxctl version --json` command.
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 ### Deprecated Features
 - ROX-12620: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
   - The permission `WorkflowAdministration` will deprecate the permissions `Policy, VulnerabilityReports`.
+- ROX-14398: We continue to simplify access control management by grouping some permissions in permission sets. As a result:
+  - The permission `Access` will deprecate the permissions `Role`.
+  - The default role `Scope Manager` will be removed.
```
dhaus67, line 23: We've added the `required actions` section below to address some feedback regarding customers not necessarily knowing what they have to do to remediate the deprecation. We should cover the role deprecation within there as well.

Especially with the removal of the `Scope Manager` role, this _does_ require an action if its used as a default minimum access role / group within auth providers.

dhaus67 (commented):
Just one comment regarding the changelog and what it entails.

Could you also include the role removal / migrations to [ROX-14398](https://issues.redhat.com/browse/ROX-14398)? Think its worth to explicitly call it out besides the role resource migration.

CHANGELOG.md
md2119, line 23: when will it be removed?

md2119 (approved):

CHANGELOG.md
rhybrillou: Target is 3.76 if this PR is merged before 3.74 release start, 3.77 otherwise.

dhaus67 (approved):

PR #4449, opened 2023-01-18
Encapsulate version string in HelpMenu component

pedrottimark:
## Description

Prerequisite to replace backend redirect for /docs/product with frontend link to OpenShift docs page.

### Changed files

1. Edit src/Containers/MainPage/Header/HelpMenu.tsx
    * Conditionally render elements that depend on version.
    * Adapt template literal for version string from hook.

2. Edit src/hooks/useMetadata.ts
    * Delete impure assignment to `versionString` property.

3. Edit src/types/metadataService.proto.ts
    * Delete `versionString` property which is not in /v1/metadata response.
        https://github.com/stackrox/stackrox/blob/master/proto/api/v1/metadata_service.proto#L11-L26

## Checklist

## Testing Performed

### DIFF
--- a/ui/apps/platform/src/Containers/MainPage/Header/HelpMenu.tsx
+++ b/ui/apps/platform/src/Containers/MainPage/Header/HelpMenu.tsx
@@ -12,7 +12,7 @@ import useMetadata from 'hooks/useMetadata';
 import { apidocsPath, productDocsPath } from 'routePaths';
 
 function HelpMenu(): ReactElement {
-    const metadata = useMetadata();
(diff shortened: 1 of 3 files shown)


bradr5:
Does it make sense to even have the metadata hook if we pull out the version string? Seems like we'd just be able to pull in metadata from redux directly in our components now

pedrottimark:
Brad, you ask a good question about `useMetadata` hook, so I will pause and think about that.

bradr5 (approved):

PR #4453, opened 2023-01-18
ROX-12900: Add filter autocomplete to PR Network Graph

vjwilson:
## Description

This adds the existing Network Graph filter autocomplete key-value input, already given a veneer of PatternFly style last year by Dave Vail, to the new PatternFly version of Network Graph.

As we were already doing in classic Network Graph, we remove the Cluster and Namespace categories, but here we also remove the Deployment category as well.

Two ancillary changes to call out:
1. I had to override some z-index styles to get the autocomplete dropdown to show up above the live area of the graph. This includes adding a high z-index on the toolbar parent, to create a new stacking context. (see https://web.dev/learn/css/z-index/#stacking-context for why we need a separate stacking context)
2. I had to change the query-object-to-query-string helper function we were using from the newer `{ getQueryString } from 'utils/queryStringUtils`, to the older `queryService from 'utils/queryService'`, because the newer function was causing incorrect handling of the nested query properties. For instance,`'Deployment Label': 'app=central'` was being parsed into `Deployment%20Label=app=central` instead of what the API expected, `Deployment%20Label:app=central`

## Checklist
- [x] Investigated and inspected CI test results (unrelated flakes because of test infrastructure network connection drop out)

## Testing Performed

Without a query filter
Screen Shot 2023-01-18 at 1 40 47 PM

With a query filter applied
Screen Shot 2023-01-18 at 1 39 19 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
@@ -14,11 +14,11 @@
 /* Create a stacking context that is higher than the context for the react-topology component
    that is rendered below the toolbar, so that the search filter dropdown will render above the graph when open */
 [data-testid="network-graph-toolbar"] {
-    z-index: 150;
+    z-index: 250;
 }
 [data-testid="network-graph-toolbar"] .pf-search-shim,
 [data-testid="network-graph-toolbar"] .react-select__menu {
-    z-index: 150;
+    z-index: 250;
 }
 
 /* Remove gray background from plain PF Select trigger button when disabled */
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
@@ -6,11 +6,20 @@
 }
 
 .pf-topology-resizable-side-bar {
-    /* we use the page header min height thrice to account for: nav header + page header + secondary header */ 
+    /* we use the page header min height thrice to account for: nav header + page header + secondary header */
     /* @TODO: Consider a more future-proof way of doing this. See https://github.com/stackrox/stackrox/pull/3703#discussion_r1014292340 */
     height: calc(100vh - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight));
 }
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/components/NetworkSearch.css
```diff
@@ -0,0 +1,52 @@
+/*
```
dvail, line 1: It doesn't really impact this PR, but one thing I noticed with this is that we need to move the CSS to a more general location and/or import it in the other places it is used.

I believe the Violations and Policies pages are using this shim via classname as well since they are in PatternFly, but neither of those components import the CSS file. Since the CSS is lazy-loaded, if you directly visit one of those old pages you will get the old Tailwind styles, but then if you visit the Network Graph and return to the original page you will get the updated PT styles.

This has been on my personal todo list but I wanted to call it out since this CSS is moving to a new location.

dvail (approved):
One comment on something I need to remember to take care of, but otherwise LGTM as an outsider.

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
```diff
@@ -6,11 +6,20 @@
...
 
+/* Create a stacking context that is higher than the context for the react-topology component
+   that is rendered below the toolbar, so that the search filter dropdown will render above the graph when open */
+[data-testid="network-graph-toolbar"] {
+    z-index: 150;
+}
+[data-testid="network-graph-toolbar"] .pf-search-shim,
+[data-testid="network-graph-toolbar"] .react-select__menu {
+    z-index: 150;
+}
```
sachaudh, line 22: I think this does work for the topology component, but seems like the side panel still shows above it

Screenshot 2023-01-18 at 1 54 42 PM

sachaudh (changes requested):
LGTM but one fix is needed

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
vjwilson: @sachaudh 
Pushed a change to fix this
Screen Shot 2023-01-19 at 12 08 35 PM

sachaudh (approved):
LGTM

PR #4457, opened 2023-01-19
ROX-12916: Adding ports/protocols edge labels to Network Graph

alwayshooin:
& bidirectional edge refactoring & some performance enhancements. 

should look like below:
the edge label logic is:
* if there is only one direction on the edge and there is only one port/protocol, show that port/protocol
* if there is only one direction on the edge and there are more than one port/protocols, show the sum
* if there is both directions on an edge and they have the same port/protocol, show that port/protocol
* if there is both directions on an edge and they don't have the same port/protocol, show the sum of both directions
image
image
image
image


sachaudh (approved):
LGTM. I think some functions are starting to get long. We might consider breaking them down into smaller subfunctions as needed

vjwilson (approved):
/lgtm

PR #4459, opened 2023-01-19
ROX-14352: Prepare for new `git` version

porridge:
## Description

A newer git provided by impeding test container bump does not like the ownership on the checked out repository both on GHA and OSCI:

```
fatal: detected dubious ownership in repository at '/go/src/github.com/stackrox/stackrox'
To add an exception for this directory, call:
	git config --global --add safe.directory /go/src/github.com/stackrox/stackrox
```

So we need to whitelist the directory.

Moreover, some of the jobs run in environment where `$HOME` is not set, so we need to do the setup before this `git config` invocation, otherwise it fails with:
```
error: could not lock config file //.gitconfig: Permission denied
```

I extracted these changes from the bump PR to keep things simple.

## Checklist

## Testing Performed

Ran tests in https://github.com/openshift/release/pull/35471 and https://github.com/stackrox/stackrox/pull/4446 where this change was included.

## Refs:

- https://github.com/actions/checkout/issues/760

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,6 +26,11 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
 
+      - name: Ignore dubious repository ownership
+        run: |
+          # Prevent fatal error "detected dubious ownership in repository" from recent git.
+          git config --global --add safe.directory "$(pwd)"
+
       - uses: ./.github/actions/create-concatenated-ui-monorepo-lock
 
       - uses: ./.github/actions/cache-ui-dependencies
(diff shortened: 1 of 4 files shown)


janisz:
I'm not sure this is the right solution. What files has wrong permissions/owners? Can we fix with chmod/chown and commit changes?

porridge:
> I'm not sure this is the right solution. What files has wrong permissions/owners?

I haven't investigated, but I suspect it's the directory resulting from `git clone`. I doubt we can chown since (I hope) we're not running as root.

> Can we fix with chmod/chown and commit changes?

I don't think we want to add more commits on top of the checked out one, since we make assumptions about the currently checked out git sha.

porridge:
We could file a followup JIRA to investigate this closer, but I'd really rather not dive deep into it as the yak shaving journey I've gone so far just to collect k8s API server audit logs is long enough already :dizzy_face:

janisz (approved):

PR #4467, opened 2023-01-19
ROX-14357: Correct the No Clusters message when empty results in cluster search

vjwilson:
## Description

Mandar noticed the same confusion in a clusters list with no search results, but some number of clusters when unfiltered, that had already been addressed in VM Reports and VM Risk Acceptance lists.

In order to make all these fixes more readable, I extracted the logic to a semantically named function, and updated all the places it is used.

## Checklist

## Testing Performed

Manually tested this new fix (screenshot below) and all the existing fixes.

Filtered Clusters list
Screen Shot 2023-01-19 at 1 05 34 PM

Unfiltered Clusters list
Screen Shot 2023-01-19 at 1 05 39 PM

### DIFF
--- a/ui/apps/platform/src/Containers/Clusters/ClustersTablePanel.js
+++ b/ui/apps/platform/src/Containers/Clusters/ClustersTablePanel.js
@@ -23,7 +23,7 @@ import {
     upgradeCluster,
 } from 'services/ClustersService';
 import { toggleRow, toggleSelectAll } from 'utils/checkboxUtils';
-import { filterAllowedSearch, convertToRestSearch } from 'utils/searchUtils';
+import { filterAllowedSearch, convertToRestSearch, getHasSearchApplied } from 'utils/searchUtils';
 
 import AutoUpgradeToggle from './Components/AutoUpgradeToggle';
 import { clusterTablePollingInterval, getUpgradeableClusters } from './cluster.helpers';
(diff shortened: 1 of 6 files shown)


ui/apps/platform/src/utils/searchUtils.ts
```diff
@@ -150,6 +150,17 @@ export function getUrlQueryStringForSearchFilter(
...
 
+/**
+ * Helper function to determine if any search has been applied.
+ *
+ * @param searchFilter The `SearchFilter` value to check.
+ *
+ * @returns boolean, true if there are any search params
+ */
+export function getHasSearchApplied(searchFilter: SearchFilter): boolean {
+    return Boolean(Object.keys(searchFilter).length);
```
pedrottimark, line 161: 1. Thank you for factoring out the semantic helper function!
2. An observation, not a blocker to merge: I have to pause for a few seconds to translate `Boolean` cast to `Object.keys(searchFilter).length !== 0` which is more intuitive to my mind.

ui/apps/platform/src/Containers/Clusters/ClustersTablePanel.js
```diff
@@ -271,28 +273,29 @@ function ClustersTablePanel({ selectedClusterId, setSelectedClusterId, searchOpt
...
-                                selection={checkedClusterIds}
-                                selectedRowId={selectedClusterId}
-                                noDataText="No clusters to show."
-                                minRows={20}
-                                pageSize={pageSize}
-                            />
-                        </div>
-                    )}
+                    {(!fetchingClusters || pollingCount > 0) &&
+                        currentClusters.length <= 0 &&
```
pedrottimark, line 277: An observation, not a blocker to merge: because a JavaScript array cannot have negative length, I have to pause a few seconds to translate to `currentClusters.length === 0` in the middle of an increasingly complicated condition.

pedrottimark (approved):
Approved with 2 comments for your consideration.

PR #4470, opened 2023-01-19
Network Graph PF edge performance improvement

alwayshooin:
adding a custom comparator for React.memo and a simple comparator value in the data model (`updateCount`) to use instead of deep comparing models

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
```diff
@@ -228,7 +228,7 @@ const TopologyComponent = ({
 
     function resetGraphToDefault() {
         removeExtraneousEdges();
-        controller.fromModel(model);
+        controller.fromModel(model, true);
```
sachaudh, line 231: What does the `true` do?

sachaudh (approved):
LGTM

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraph.tsx
alwayshooin: it makes it so that the new model is merged w the previously existing model. i think it's `true` by default but i just wanted to explicitly put it there

PR #4474, opened 2023-01-20
Clair v4 integration: only show vuln ID

RTann:
## Description

Clair names vulnerabilities with potentially more data than just the ID (see the unit tests for examples). This PR extracts the ID portion out (or keeps the original if an ID cannot be extracted)

## Checklist

## Testing Performed

CI


hdonnay (approved):
LGTM.

I don't think there's any other namespace in use within Clair -- perhaps [`OSV`](https://ossf.github.io/osv-schema/) in the near future.

pkg/scanners/clairv4/convert.go
```diff
@@ -149,6 +152,15 @@ func vulnerabilities(vulnerabilities map[string]*claircore.Vulnerability, ids []
 	return vulns
 }
 
+func name(original string) string {
```
jvdm, line 155: Can we have a more descriptive function name, maybe:

```suggestion
func vulnName(original string) string {
```

jvdm (approved):
LGTM.

PR #4475, opened 2023-01-20
Dashrews/ROX-14424 properly set migration yaml on pg upgrade

dashrews78:
## Description

After the migrations are executed we need to set the version of the rocksdb instance so it remains consistent in the event we need to rollback to it.  Previously we were updating the sequence number in a way that if we were in postgres mode we would set it to the max sequence number instead of the max rocksdb sequence number.  This resolves that AND should allow us to in some instances to switch from Postgres back to Rocksdb by changing the env var.

## Checklist

## Testing Performed

Manual testing, but will test more thoroughly once #4499 is merged.

### DIFF
--- a/pkg/migrations/migration_version.go
+++ b/pkg/migrations/migration_version.go
@@ -61,11 +61,11 @@ func Read(dbPath string) (*MigrationVersion, error) {
 
 // SetCurrent update the database migration version of a database directory.
 func SetCurrent(dbPath string) {
-	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != CurrentDBVersionSeqNum() {
+	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != LastRocksDBVersionSeqNum() {
 		newVersion := &MigrationVersion{
 			dbPath:      dbPath,
 			MainVersion: version.GetMainVersion(),
-			SeqNum:      CurrentDBVersionSeqNum(),
+			SeqNum:      LastRocksDBVersionSeqNum(), // Most recent possible RocksDB version
 		}
 		err := newVersion.atomicWrite()
 		if err != nil {


pkg/migrations/migration_version.go
```diff
@@ -61,11 +61,11 @@ func Read(dbPath string) (*MigrationVersion, error) {
 
 // SetCurrent update the database migration version of a database directory.
 func SetCurrent(dbPath string) {
-	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != CurrentDBVersionSeqNum() {
+	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != LastRocksDBVersionSeqNum() {
 		newVersion := &MigrationVersion{
 			dbPath:      dbPath,
 			MainVersion: version.GetMainVersion(),
-			SeqNum:      CurrentDBVersionSeqNum(),
+			SeqNum:      LastRocksDBVersionSeqNum(), // Most recent possible RocksDB version
```
c-du, line 68: Yes, that is right. We need to update LastPersistTime here.

c-du (approved):

pkg/migrations/migration_version.go
```diff
@@ -61,11 +61,11 @@ func Read(dbPath string) (*MigrationVersion, error) {
 
 // SetCurrent update the database migration version of a database directory.
 func SetCurrent(dbPath string) {
-	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != CurrentDBVersionSeqNum() {
+	if curr, err := Read(dbPath); err != nil || curr.MainVersion != version.GetMainVersion() || curr.SeqNum != LastRocksDBVersionSeqNum() {
```
connorgorman, line 64: I think this is just generally safer if we say > LastRocksDBVersion?

dashrews78: @connorgorman well maybe but this gets called on other clones and we have to account for the LastPersistedTime so we can properly go from rocks -> Postgres -> Rocks -> Postgres.  Take a look at #4511 where I had to make some additional changes to satisfy the case of  rocks -> Postgres -> Rocks -> Postgres all on the same software version as in that instance we need to ensure the migrations run again

PR #4476, opened 2023-01-20
ROX-14229: Deterministic UUIDs for declarative config resources

dhaus67:
## Description

This PR is a pre-factor for the declarative configuration project.

For declarative configuration, we allow users to specify resources in a declarative manner by providing YAML configuration via config map / secret mounts within central.

The YAML-based configuration will not be using any UUIDs, but instead will be uniquely identified by their name.
To support this, we need to make sure we map those names internally to UUIDs, since datastores typically work with UUIDs as unique identifiers, as well as make sure that we will be able to create the UUIDs deterministic, since the same _named declarative configuration_ should only create _a single set of resources_.

For this, we re-use the `uuid.NewV5FromNonUUID`, which under hood are `v5 UUIDs`, which are meant for generating UUIDs from names (for more info, you can read up [the RFC for UUIDs](https://www.rfc-editor.org/rfc/rfc4122#section-4.3)).

The newly created package contains helper functions to create UUIDs for all of the resources we envision for the first MVP of declarative configuration, including their namespaces.

Note that the `Role` resource has been omitted, since it is one of the few exceptions where no generated IDs are used for unique identification, but rather the name. Hence, we won't need any deterministic UUIDs for those resources.

## Checklist

## Testing Performed

- see CI (added unit tests).

### DIFF
--- a/pkg/declarativeconfig/uuid.go
+++ b/pkg/declarativeconfig/uuid.go
@@ -1,0 +1,34 @@
+package declarativeconfig
+
+import "github.com/stackrox/rox/pkg/uuid"
+
+const (
+	authProviderNS  string = "auth-provider"
+	groupNS         string = "group"
+	permissionSetNS string = "permission-set"
+	accessScopeNS   string = "access-scope"
+)
+
+// NewDeclarativeAuthProviderUUID creates a UUID from the name of a declarative auth provider configuration.
+// The returned UUID will be deterministic.
+func NewDeclarativeAuthProviderUUID(name string) uuid.UUID {
+	return uuid.NewV5FromNonUUIDs(authProviderNS, name)
+}
+
+// NewDeclarativeGroupUUID creates a UUID from the name of a declarative group configuration.
+// The returned UUID will be deterministic.
+func NewDeclarativeGroupUUID(name string) uuid.UUID {
+	return uuid.NewV5FromNonUUIDs(groupNS, name)
+}
+
+// NewDeclarativePermissionSetUUID creates a UUID from the name of a declarative permission set configuration.
+// The returned UUID will be deterministic.
+func NewDeclarativePermissionSetUUID(name string) uuid.UUID {
+	return uuid.NewV5FromNonUUIDs(permissionSetNS, name)
+}
+
+// NewDeclarativeAccessScopeUUID creates a UUID from the name of a declarative access scope configuration.
+// The returned UUID will be deterministic.
+func NewDeclarativeAccessScopeUUID(name string) uuid.UUID {
+	return uuid.NewV5FromNonUUIDs(accessScopeNS, name)
+}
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #4476** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4476" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4476?utm_source=stack-comment).

pkg/declarativeconfig/uuid_test.go
```diff
@@ -0,0 +1,37 @@
...
+	dummyName := "dummy-resource"
+
+	// 1. Test that using the same name will lead to the same UUID being created.
+	for _, f := range declarativeUUIDFuncs {
+		firstID := f(dummyName)
+		secondID := f(dummyName)
+		assert.Equal(t, firstID, secondID)
+	}
+
+	// 2. Test that using the same name won't lead to clashes within the different namespaces.
```
ivan-degtiarenko, line 26: I suggest adding the third test: testing that different names in the same namespace result in different UUIDs. 
Otherwise, currently this test can be passed by `func constant(a string) { return "constantValue" }`

pkg/declarativeconfig/uuid.go
```diff
@@ -0,0 +1,34 @@
+package declarativeconfig
+
+import "github.com/stackrox/rox/pkg/uuid"
+
+const (
+	authProviderNS  string = "auth-provider"
```
ivan-degtiarenko, line 6: nit: should we call them namespaces? Then they can be confused with k8s namespaces. Perhaps prefix is a better word?

dhaus67: Hm, I'm a bit on the fence with this one. While it may be true, in the context of UUID generation, namespace _is_ the correct name. Prefix would be something we currently do with our internal IDs[ such as the groups ID generation](https://github.com/stackrox/stackrox/blob/65820256abbcb6cc2809f7a6ece8ca7729dd904a/central/group/datastore/validate.go#L53-L56), so I'd rather stick with namespace for now as it's more explicit in the current context of the file.

rhybrillou: One option to reduce the confusion would be to name these namespaces UUID Namespaces.
An alternative I could see would be to put a comment stating that the namespaces used here are UUID namespaces and not k8s namespaces.

rhybrillou (approved):

pkg/declarativeconfig/uuid.go
ivan-degtiarenko: UUID Namespace sounds good. Anyway, I'll leave it up to @dhaus67 to decide

ivan-degtiarenko (approved):
🚢

PR #4480, opened 2023-01-20
Process Listening On Port: Internal API support

erthalion:
## Description

Introduce a Sensor pipeline to handle PLOP events from Collector. This actually
plugs the Store mechanism from #4328 into the workflow. The PR depends on
#4328 and has being separated only to make review more convenient.

## Checklist
~- [ ] Documented user facing changes~

There are no breaking changes, no update steps (the PR is about handling new
event type from Sensor) and no user facing changes yet (they come in the
follow-up PRs).

## Testing Performed

Local testing, unit-tests, verifying previous test CI pipelines (e.g. #4182).
To test manually one have to enable this feature in both Sensor and Collector
via `ROX_PROCESSES_LISTENING_ON_PORT`. After this any listening port will be
reported all the way up to the Central.


central/sensor/service/pipeline/processlisteningonport/pipeline.go
```diff
@@ -0,0 +1,70 @@
...
+	ctx context.Context,
+	clusterID string,
+	msg *central.MsgFromSensor,
+	injector common.MessageInjector,
+) error {
+	defer countMetrics.IncrementResourceProcessedCounter(
+		pipeline.ActionToOperation(msg.GetEvent().GetAction()), metrics.ProcessListeningOnPort)
+
+	update := msg.GetProcessListeningOnPortUpdate()
+	if s.dataStore != nil && update != nil {
```
dashrews78, line 56: do we want to log an error here?

erthalion: Do you mean if both are nil? Yeah, probably won't hurt.

erthalion: I ended up adding a warning instead.

JoukoVirtanen: I agree with adding a warning instead of an error. The dataStore will be a nil if postgres is not used. We could change this to an error when rocksdb is removed.

proto/internalapi/central/sensor_events.proto
```diff
@@ -80,6 +81,8 @@ message SensorEvent {
         storage.ImageIntegration        image_integration       = 13;
         ReprocessDeploymentRisk         reprocess_deployment    = 17;
         AlertResults                    alert_results           = 18;
+        // XXX: Is this really needed?
+        storage.ProcessListeningOnPort  process_listening_on_port = 26;
```
lvalerom, line 85: AFAICT you're not using `SensorEvent` messages in the pipeline. Unless you're planning on sending and handling `SensorEvents` in the future I think this line can be removed.

lvalerom (approved):
LGTM

dashrews78 (approved):

PR #4482, opened 2023-01-20
Process Listening On Port: Public API

erthalion:
## Description

Introduce the Public API to expose PLOP information about actively listening
ports, and E2E test using this API as well. The PR depends on #4480 and has
being separated only to make review more convenient.

## Checklist

There are no no update steps (the PR is about exposing new type of
information).

## Testing Performed

Local testing, unit-tests, verifying previous test CI pipelines. To test
manually one have to enable this feature in both Sensor and Collector via
`ROX_PROCESSES_LISTENING_ON_PORT`. After this any listening port will be
reported all the way up to the Central and could be requested from the API.


central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (s *serviceImpl) GetProcessesListeningOnPorts(
+	ctx context.Context,
+	req *v1.GetProcessesListeningOnPortsRequest,
+) (*v1.GetProcessesListeningOnPortsResponse, error) {
+
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		// PLOP is a Postgres-only feature, do nothing.
+		log.Warnf("Tried to request PLOP not on Postgres, ignore: %s", req.GetDeploymentId())
+		return emptyProcessesListeningOnPortsResponse()
```
dhaus67, line 61: Usually, IIRC, we return an error in case APIs are being tried to use that are behind a feature flag, so this should ideally become:
```go
return nil, errors.Errorf("Postgres env var is not enabled, PLOP APIs are disabled")
```

central/main.go
```diff
@@ -373,6 +374,7 @@ func servicesToRegister(registry authproviders.Registry, authzTraceSink observe.
 		probeUploadService.Singleton(),
 		processIndicatorService.Singleton(),
 		processBaselineService.Singleton(),
+		processListeningOnPorts.Singleton(),
```
dhaus67, line 377: You should move adding the service singleton to the condition at l.395, i.e. only add the service if postgres is enabled.

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func emptyProcessesListeningOnPortsResponse() (*v1.GetProcessesListeningOnPortsResponse, error) {
+	result := &v1.GetProcessesListeningOnPortsResponse{
+		ListeningEndpoints: make([]*storage.ProcessListeningOnPort, 0),
+	}
+	return result, nil
+}
+
+func (s *serviceImpl) GetProcessesListeningOnPorts(
```
dhaus67, line 53: super-nit: would be nice to have a comment here describing the method.

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+) (*v1.GetProcessesListeningOnPortsResponse, error) {
+
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		// PLOP is a Postgres-only feature, do nothing.
+		log.Warnf("Tried to request PLOP not on Postgres, ignore: %s", req.GetDeploymentId())
+		return emptyProcessesListeningOnPortsResponse()
+	}
+
+	deployment := req.GetDeploymentId()
+	processesListeningOnPorts, err := s.dataStore.GetProcessListeningOnPort(ctx, deployment)
```
dhaus67, line 65: I haven't looked too deep into the datastore, but in case the deployment ID cannot be found, will we emit a `errox.NotFound` error?

Reason I'm asking is that the `errox` types will be used to map to GRPC/ HTTP status codes (see [this reference here](https://github.com/stackrox/stackrox/blob/22225b4ec1f2dc1fa7b54678610e1dc72c2ed303/pkg/grpc/errors/interceptor.go#L88-L108)), so just want to ensure that we know this exists.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    @Tag("BAT")
+    def "Verify networking endpoints with processes appear in API at the deployment level"() {
+        given:
+        "Two deployments that listen on ports are started up"
+
+        if (! Env.CI_JOBNAME.contains("postgres")) {
+                return
+        }
```
dhaus67, line 116: You could also do this on the class level with an annotation as such:
```
@IgnoreIf({ Env.CI_JOBNAME.contains("postgres") })
```

central/processlisteningonport/service/service.go
```diff
@@ -0,0 +1,29 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/grpc"
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
+
+// Service provides the interface to the microservice that serves alert data.
```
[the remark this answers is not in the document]
JoukoVirtanen: Changed "alert" to "process listening on ports"

central/processlisteningonport/service/service.go
```diff
@@ -0,0 +1,29 @@
...
+
+	datastore "github.com/stackrox/rox/central/processlisteningonport/datastore"
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/grpc"
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
```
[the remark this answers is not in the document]
JoukoVirtanen: I removed this. I also moved all logging statements in service_impl.go. So there is no reason to have it in either file.

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/grpc/authz"
+	"github.com/stackrox/rox/pkg/grpc/authz/perrpc"
+	"github.com/stackrox/rox/pkg/grpc/authz/user"
+	"google.golang.org/grpc"
+)
+
+var (
+	authorizer = perrpc.FromMap(map[authz.Authorizer][]string{
+		user.With(permissions.View(resources.DeploymentExtension)): {
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if err != nil {
+		log.Warnf("In processlisteningonport service query return err: %+v", err)
+		return emptyProcessesListeningOnPortsResponse()
+	}
+
+	if processesListeningOnPorts == nil {
+		log.Debug("In processlisteningonport service query return nil")
+		return emptyProcessesListeningOnPortsResponse()
+	}
```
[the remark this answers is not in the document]
JoukoVirtanen: The logging statement for when GetProcessListeningOnPort returns an error and when it returns nil where moved to the datastore.

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+		log.Warnf("Tried to request PLOP not on Postgres, ignore: %s", req.GetDeploymentId())
+		return emptyProcessesListeningOnPortsResponse()
+	}
+
+	deployment := req.GetDeploymentId()
+	processesListeningOnPorts, err := s.dataStore.GetProcessListeningOnPort(ctx, deployment)
+
+	if err != nil {
+		log.Warnf("In processlisteningonport service query return err: %+v", err)
+		return emptyProcessesListeningOnPortsResponse()
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/service/service_impl.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if processesListeningOnPorts == nil {
+		log.Debug("In processlisteningonport service query return nil")
+		return emptyProcessesListeningOnPortsResponse()
+	}
+
+	return &v1.GetProcessesListeningOnPortsResponse{
+		ListeningEndpoints: processesListeningOnPorts,
+	}, err
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

central/processlisteningonport/service/service_impl.go
JoukoVirtanen: Done

central/main.go
JoukoVirtanen: Done

central/processlisteningonport/service/service_impl.go
JoukoVirtanen: Done

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
JoukoVirtanen: Using `@IgnoreIf({ !Env.CI_JOBNAME.contains("postgres") })`

dhaus67 (approved):
Changes look good to me, modulo the open comment regarding adding the v1.UnimplementedProcessesListeningOnPortsServiceServer to the struct implementing the grpc service.

c-du (approved):

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -0,0 +1,252 @@
...
+        given:
+        "Two deployments that listen on ports are started up"
+
+        rebuildForRetries()
+        def clusterId = ClusterService.getClusterId()
+
+        String deploymentId1 = targetDeployments[0].getDeploymentUid()
+        String deploymentId2 = targetDeployments[1].getDeploymentUid()
+
+        def gotCorrectNumElements = waitForResponseToHaveNumElements(2, deploymentId1, 240)
```
dhaus67, line 122: More of an observation, and not a comment blocking this PR:
Looking through the test logs, the first attempt fails after a couple of minutes, and the second attempt immediately succeeds after few seconds.. any idea what's going on there?

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -0,0 +1,252 @@
...
+
+    def cleanupSpec() {
+        destroyDeployments()
+    }
+
+    def rebuildForRetries() {
+        if (Helpers.getAttemptCount() > 1) {
+            log.info ">>>> Recreating test deployments prior to retest <<<<<"
+            destroyDeployments()
+            sleep(5000)
```
dhaus67, line 104: do we need these 5 second sleeps here? Ideally, `destroyDeployments/createDeployments` will only ever return after ensuring the deployments are deleted / created.

dhaus67 (approved):
Some minor comments within the tests, but they do not block this PR.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
JoukoVirtanen: Removed

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -0,0 +1,252 @@
...
+                        imagePullSecrets: ["quay"]
+                )
+        )
+
+        createDeployments()
+    }
+
+    def destroyDeployments() {
+        for (Deployment deployment : targetDeployments) {
+            orchestrator.deleteDeployment(deployment)
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
JoukoVirtanen: If there is a failure, the retry should be independent. It does seem like the first try succeeds less often than the second try. We will investigate. We do have planned improvements that should make this more reliable.

PR #4430, opened 2023-01-17
[ROX-12943,-12975,-13722] Send NodeInventory as a separate message between Sensor and Central

vikin91:
## Description

This PR is an alternative implementation to #3757

This draft contains also #3940 and #3892

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #3892, opened 2022-11-23
ROX-12943: Sensor maps the NodeInventory to Node

vikin91:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #4483, opened 2023-01-20
ROX-13722: Readd nodeID to NodeInventory proto

vikin91:
## Description

This PR was created as a split from #4430 in order to make the change easier to review.
This PR is meant to replace #3757 - as we will not attach NodeInventory to Node anymore, we need to change the NodeInventory so that it contains the information we will need in Central.

Here, we readd the `nodeID` field to `NodeInventory` as this field will be needed to uniquely identify the Node that the NodeInventory belongs to. The `NodeInventory.NodeID` will be filled by Sensor in another PR. 

Note that by abandoning #3757 and following #4430 we DO NOT:
- attach NodeInventory to Node,
- expose NodeInventory over public APIs,
- save NodeInventory in the Central-DB

anymore.

I also renamed central pipeline for handling Node Inventories from `nodescanv2` to `nodeinventory`.

## Checklist
- [X] Investigated and inspected CI test results

### Not apply
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed

- CI
- Deploy locally on Colima, enable FF and check log for errors

PR #4484, opened 2023-01-20
ROX-12943: Sensor sets NodeID in NodeInventory

vikin91:
## Description

This PR was created as a split from #4430 in order to make the change easier to review.
This PR is meant to replace #3892.

Here, we change Sensor to store the Nodes in the in-memory store, so that Sensor is able to find a Node object based on the node name that arrived from Compliance.
 
When NodeInventory arrives to Sensor from Compliance it contains only the node name. Next, Sensor searches the in-memory store to find the Node ID for a given name. If the ID is found, then `NodeInventory.NodeID` field is set and NodeInventory is sent to Central. When Node cannot be found, the NodeInventory message is dropped. 

## Checklist

### Not apply

## Testing Performed

- CI
- Deploy locally on Colima, enable FF and check logs

Selected logs
```
# sensor
sensor-66b879fbd7-hlhvz common/compliance: 2023/01/20 10:47:25.681686 node_inventory_handler_impl.go:107: Info: Mapping NodeInventory name 'colima' to Node ID '252afb4d-0694-45c3-9f86-e1bd7d7a26de'

# central
central-c5dcb8cc8-cnvq5 sensor/service/pipeline/nodeinventory: 2023/01/20 10:47:55.691034 pipeline.go:52: Info: Central received NodeInventory for Node name='colima' ID='252afb4d-0694-45c3-9f86-e1bd7d7a26de'
```


vikin91:
* master
  * **PR #4483** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4483" target="_blank">Graphite</a> 
    * **PR #4484** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4484" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4484?utm_source=stack-comment).

sensor/kubernetes/sensor/sensor.go
```diff
@@ -106,7 +106,7 @@ func CreateSensor(cfg *CreateOptions) (*sensor.Sensor, error) {
 
 	imageCache := expiringcache.NewExpiringCache(env.ReprocessInterval.DurationSetting())
 	policyDetector := detector.New(enforcer, admCtrlSettingsMgr, resources.DeploymentStoreSingleton(), resources.ServiceAccountStoreSingleton(), imageCache, auditLogEventsInput, auditLogCollectionManager, resources.NetworkPolicySingleton())
-	pipeline := eventpipeline.New(cfg.k8sClient, configHandler, policyDetector, k8sNodeName.Setting(), cfg.resyncPeriod, cfg.traceWriter, storeProvider)
+	pipeline := eventpipeline.New(storeProvider.Nodes(), cfg.k8sClient, configHandler, policyDetector, k8sNodeName.Setting(), cfg.resyncPeriod, cfg.traceWriter, storeProvider)
```
lvalerom, line 109: I don't think you need to pass the `NodeStore` here. We are already passing the `storeProvider` that contains the `NodeStore`. You can just use `storeProvider.Nodes()` when you are creating the `nodeDispatcher` in the `NewDispatcherRegistry` function.

lvalerom (commented):
Just commenting because I haven't reviewed the entire PR yet.

sensor/kubernetes/sensor/sensor.go
vikin91: You are absolutely right. I got that corrected in [660a42c](https://github.com/stackrox/stackrox/pull/4484/commits/660a42c645c6912ae3bb6596d3645c764684dba7).

sensor/common/compliance/node_matcher.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+
+	"github.com/stackrox/rox/sensor/common/store"
+)
+
+// nodeStore provides functionality to get nodes
+type nodeStore interface {
+	AddOrUpdateNode(node *store.NodeWrap) bool
```
fredrb, line 11: This doesn't seem to be used by production code, only for tests. Why does it need to be part of the interface?

sensor/common/store/nodes.go
```diff
@@ -0,0 +1,12 @@
+package store
+
+import (
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/net"
+)
+
+// NodeWrap adds address information to nodes to forward to central
+type NodeWrap struct {
```
fredrb, line 9: I don't understand the design decision of exposing this _internal_ struct of the listeners rather than using a separate interface returning only `*storage.Node`, which is what is really used outside the listener. 

I.e.: I don't see `Addresses` being used anywhere other than in the listener. So why do they need to be exposed in `sensor/common`?

Most stores have the similar requirements. The way this is handled for other stores is to have a sensor-wide API for the store (defined in `sensor/common`), which exposes `storage.*` type structs. And have its implementation be internal to `sensor/kubernetes`, where you can use any additional information needed.

Take deployments, for example:

```go
// In sensor/common/store:
type DeploymentStore interface {
	Get(id string) *storage.Deployment
}

// In sensor/kubernetes/listener/resources
type DeploymentStore struct { ... }

func (ds *DeploymentStore) addOrUpdateDeployment(wrap *deploymentWrap) { ... }

// Get returns deployment for supplied id.
func (ds *DeploymentStore) Get(id string) *storage.Deployment {
	wrap := ds.getWrap(id)
        // parse to a storage.Deployment type to comply with public common interface
	return wrap.GetDeployment()
}
```

fredrb (commented):
Some comments that we can discuss synchronously.

sensor/common/compliance/mocks/node_inventory.go
```diff
@@ -0,0 +1,51 @@
+package mocks
```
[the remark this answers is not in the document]
vikin91: If simple impl is enough then implement it in sensor/common/compliance/node_matcher_test.go
If it is more complex, then use mockgen.

In both cases remove `AddOrUpdateNode` from the interface.

sensor/kubernetes/listener/resources/node_store.go
```diff
@@ -3,84 +3,94 @@ package resources
...
-type nodeWrap struct {
-	*v1.Node
-	addresses []net.IPAddress
+// nodeStore represents a collection of NodeWraps
+type nodeStore interface {
+	AddOrUpdateNode(node *store.NodeWrap) bool
+	RemoveNode(node *storage.Node)
+	GetNode(nodeName string) *store.NodeWrap
+	GetNodes() []*store.NodeWrap
 }
```
vikin91, line 20: Should be in the common file with all other interfaces.

vikin91: Follow the example of Roles an RBACS

sensor/common/compliance/node_matcher.go
```diff
@@ -0,0 +1,38 @@
+package compliance
+
+import (
+	"fmt"
+
+	"github.com/stackrox/rox/sensor/common/store"
+)
+
+// nodeStore provides functionality to get nodes
+type nodeStore interface {
```
vikin91, line 10: We should return the storage types here - storage.Node.

sensor/kubernetes/listener/resources/endpoints.go
```diff
@@ -8,31 +8,32 @@ import (
...
+	OnNodeCreate(node *store.NodeWrap)
 	OnNodeUpdateOrRemove()
 }
 
 type endpointManagerImpl struct {
 	serviceStore    *serviceStore
 	deploymentStore *DeploymentStore
 	podStore        *PodStore
-	nodeStore       *nodeStore
+	nodeStore       nodeStore
```
vikin91, line 31: Use unexported nodeStoreImpl here or the second unexported interface.

vikin91:
I addressed most of the comments and stumbled upon an issue with code generation. I want the CI to give this a shot. Will ping you when the PR is ready to be looked at again.

vikin91:
The CI failure for `gke-postgres-nongroovy-e2e-tests` affects also `master`, so not a blocker here.

vikin91:
@fredrb @lvalerom @Maddosaurus @jschnath this is ready to be reviewed again.

CI went crazy for unknown reasons - will retry later, but code should be ready to give it another look.

vikin91:
I rebased again to fix the CI issues (see [msg](https://srox.slack.com/archives/C3W3U0RGW/p1674672979365629))

sensor/kubernetes/listener/resources/store_provider.go
```diff
@@ -56,3 +56,8 @@ func (p *InMemoryStoreProvider) RBAC() store.RBACStore {
 func (p *InMemoryStoreProvider) EndpointManager() store.EndpointManager {
 	return p.endpointManager
 }
+
+// Nodes returns the Nodes public interface
+func (p *InMemoryStoreProvider) Nodes() *nodeStoreImpl {
```
lvalerom, line 61: I know that this goes against the go convention of "Accept interfaces, return structs", but I would either change this to `store.NodeStore` or I would change the implementations above. Otherwise it looks like they are different when in reality they're meant to do the same.

central/sensor/service/pipeline/nodeinventory/pipeline.go
```diff
@@ -48,8 +48,8 @@ func (p *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
 		return errors.Errorf("unexpected resource type %T for node inventory", event.GetResource())
 	}
 
-	// TODO(ROX-12240, ROX-13053): Do something meaningful with the nodeInventory
-	log.Debugf("Central received NodeInventory: %+v", nodeInventory)
+	// TODO(ROX-12975): Handle nodeInventory
+	log.Infof("Central received NodeInventory for Node name='%s' ID='%s'", nodeInventory.GetNodeName(), nodeInventory.GetNodeId())
```
lvalerom, line 52: Just making sure that you didn't leave it as `Info` by mistake.

sensor/common/compliance/node_inventory_handler_impl.go
```diff
@@ -77,15 +77,24 @@ func (c *nodeInventoryHandlerImpl) run() <-chan *central.MsgFromSensor {
...
 					return
 				}
-				// TODO(ROX-12943): Do something with the inventory, e.g., attach NodeID
-				c.sendInventory(toC, inventory)
+				if inventory == nil {
+					log.Warnf("Received nil NodeInventory - not sending node inventory to Central")
+					break
+				}
+				if nodeID, err := c.nodeMatcher.GetNodeID(inventory.GetNodeName()); err != nil {
+					log.Warnf("Node '%s' unknown to sensor - not sending node inventory to Central", inventory.GetNodeName())
```
lvalerom, line 85: I think we should add a test for this path. What would be expected if we received a `nil` inventory? What would happen if we don't find the node in the store?

lvalerom (approved):
Looks good to me. I have some comments but I don't want to block the PR since I think most of them can be skipped or done in follow-up PRs.

central/sensor/service/pipeline/nodeinventory/pipeline.go
vikin91: No, this is on purpose - it allows us to see in the logs that this feature works. It happens only when the FF is enabled. 
We will revisit logging before release https://issues.redhat.com/browse/ROX-14358

sensor/common/compliance/node_inventory_handler_impl.go
vikin91: Not finding the Node in the store should by design result in ignoring the `nodeInventory` - spec in the ticket https://issues.redhat.com/browse/ROX-12943.

vikin91: Now if `nodeInvetory` is nil - this should not happen.

NodeInventories are written into the channel by [compliance service](https://github.com/stackrox/stackrox/blob/master/sensor/common/compliance/service_impl.go#L134-L151). The way the `switch` is built should prevent sending nils to the channel.

I understand that you ask for a testcase to see what happens if we get a nil here anyway. Let me see what I can implement in `sensor/common/compliance/node_inventory_handler_test.go`

vikin91: Got the test cases in [48c2883](https://github.com/stackrox/stackrox/pull/4484/commits/48c2883fe25debdeb89ebbc480230f8c7da300df) and [91c1ac1](https://github.com/stackrox/stackrox/pull/4484/commits/91c1ac188e3dc78adfd787e18a2b2c6519db5363) :)

sensor/kubernetes/listener/resources/store_provider.go
fredrb: +1 to this

fredrb (approved):
I'm fine with this once you address Luis' comment. Thanks for making the changes!

sensor/kubernetes/listener/resources/store_provider.go
vikin91: I see your point. Let me share mine.

There are ~three~ four options here:

1. `func (p *InMemoryStoreProvider) Nodes() store.NodeStore`

This is a problem cause we rely on the `nodeStore` interface in some places.

```
sensor/kubernetes/listener/resources/dispatcher.go:91:65: cannot use storeProvider.Nodes() (value of type store.NodeStore) as nodeStore value in argument to newNodeDispatcher: store.NodeStore does not implement nodeStore (missing method addOrUpdateNode) (typecheck)
```

2. `func (p *InMemoryStoreProvider) Nodes() nodeStore`

This is fine for now, but returning an unexported interface may limit the usability of this method in the future.

3.  `func (p *InMemoryStoreProvider) Nodes() *nodeStoreImpl`

The current state. It allows to add checks for the `impl` to make sure that it implements both interfaces:

```go
var _ nodeStore = (*nodeStoreImpl)(nil)
var _ store.NodeStore = (*nodeStoreImpl)(nil)
```

4. `func (p *InMemoryStoreProvider) Nodes() BigNodeStore`

Create new interface `BigNodeStore` (name subject to change) that would do:

```go
type BigNodeStore interface {
    nodeStore
    store.NodeStore
}
```

Based on that, I think that option 3 looks the best. What do you think?

vikin91: I think there is also a 5th option where we make `store.NodeStore` a copy of `nodeStore` but the methods from `nodeStore` (`addOrUpdateNode` etc.) will be unexported in the new bigger `store.NodeStore`.

vikin91: Following the other alternative (returning impls everywhere) would look like this:

```go
type InMemoryStoreProvider struct {
	deploymentStore *DeploymentStore  // Interface named the same as implementation - add Impl to type name?
	podStore        *PodStore
	serviceStore    *serviceStore
	rbacStore       rbac.storeImpl       // Unexported - would need to export `storeImpl`
	endpointManager *endpointManagerImpl 
	nodeStore       *nodeStoreImpl
	entityStore     *clusterentities.Store  // this is actually Impl - add Impl to type name?
}
```

As it contains a lot of renaming I would like to do this in a followup PR. Would that be okay?

sensor/common/compliance/node_inventory_handler_test.go
```diff
@@ -221,3 +221,62 @@ func (s *NodeInventoryHandlerTestSuite) TestInputChannelClosed() {
...
+
+// GetNodeID always finds a hardcoded ID "abc"
+func (c *mockAlwaysHitNodeIDMatcher) GetNodeID(nodename string) (string, error) {
+	return "abc", nil
+}
+
+// mockNeverHitNodeIDMatcher simulates inability to find a node when GetNodeResource is called
+type mockNeverHitNodeIDMatcher struct{}
+
+// GetNodeID neever finds a node and returns error hardcoded ID "abc"
```
Maddosaurus, line 279:
```suggestion
// GetNodeID never finds a node and returns error
```
Aside from the typo, a logic question: We return an error, no hardcoded ID at all, right?

Maddosaurus (approved):
Thanks for addressing the feedback from our sync review. Apart from the other ongoing discussions, I found a comment to fix, but that's non-blocking

sensor/common/compliance/node_inventory_handler_test.go
vikin91: Yes, this simulates situation when the node is not found, so we return empty ID.

sensor/kubernetes/listener/resources/store_provider.go
fredrb: I have a simple suggestion related to (1). The only place where it seems like you should be needing the private struct instead of the public interface is when instantiating the handlers. In `sensor/kubernetes/listener/resources/dispatcher.go` you could simply do:
```go
nodeDispatcher:            newNodeDispatcher(deploymentStore, storeProvider.nodeStore, endpointManager),
```
And the dispatcher will be able to call the private store functions. 

If I'm not mistaken, when you make this change, you should be able to switch the `.Nodes()` function to the public interface without any compiler errors.

This is what we do for RBACs and Services stores. Maybe it's not the ideal way of doing things and maybe you're on to something better. But I'd opt for the easiest thing here which is just passing the internal struct to the handler. We can have a sync discussion after this PR on how we could improve this if you'd like. 

Could you give it a try?

sensor/common/compliance/node_inventory_handler_impl.go
lvalerom: Thank you for adding the tests 🙂 

Sorry, I think my wording wasn't the best. My comment was simply pointing out that since this is the critical change in this PR, I think it should be tested to make sure it works the way it's meant to work.

sensor/kubernetes/listener/resources/store_provider.go
lvalerom: Add me to that sync discussion if possible 🙂 I have similar feelings as Fred but I feel they might not align with "the Go way".

sensor/common/compliance/node_matcher.go
```diff
@@ -0,0 +1,36 @@
+package compliance
+
+import (
+	"fmt"
+
+	"github.com/stackrox/rox/generated/storage"
+)
+
+type nodeStore interface {
```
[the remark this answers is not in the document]
vikin91: That is correct, I missed that one. Should be fixed now.

jschnath (approved):
I have nothing to add that wasn't mentioned yet, LGTM

vikin91:
4 approvals and 100% green CI :) Looks like it is time to merge this 🚢

sensor/kubernetes/listener/resources/store_provider.go
vikin91: Looks like it worked great! I will be merging this in a moment

PR #4485, opened 2023-01-20
ROX-14352: Bump CI images to latest version.

porridge:
## Description

The goal is to pick up a new `oc`. It also happens to pick up newer `git` - a fix for its new checks was merged in #4459 

The custom-built ARM image mentioned in the `Makefile` is not bumped. See https://issues.redhat.com/browse/ROX-12064

## Checklist

## Testing Performed

CI should suffice.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
   pre-build-ui:
     runs-on: ubuntu-latest
     container:
-      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.49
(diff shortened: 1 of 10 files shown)


janisz (approved):
We probably need a ci job to check if image is updated in all places

gavin-stackrox:
The tests on the rehearsal PR are all failing. There is a permission issue copying artifacts to GCS which might be that the rehearsal jobs have an old cred somehow. But there is also a meltdown in the Groovy tests that I cannot figure out.

porridge:
@gavin-stackrox do you think it's safe to merge this one while we still ponder https://github.com/openshift/release/pull/35572 ?

PR #4487, opened 2023-01-20
add more information to the assertions of the replay tests

lvalerom:
## Description

Adds more information to the assertions of the replay tests

## Checklist

## Testing Performed

* CI


sensor/tests/replay/replay_test.go
```diff
@@ -212,15 +212,15 @@ func (suite *ReplayEventsSuite) Test_ReplayEvents() {
 			}
 			expectedAlerts := getAlerts(expectedEvents)
 			receivedAlerts := getAlerts(allEvents)
-			assert.Equal(t, len(expectedAlerts), len(receivedAlerts))
+			assert.Equal(t, len(expectedAlerts), len(receivedAlerts), "Not the same number of alerts.\nExpected: %+v\nActual: %+v", expectedAlerts, receivedAlerts)
```
fredrb, line 215: nit: use `assert.Len` to assert length of slices. I think the message will be more clear that the number of events don't match. 

Check if using `.Len` would print out the elements of the slice. If so the extra message wouldn't be needed.

sensor/tests/replay/replay_test.go
```diff
@@ -229,10 +229,10 @@ func (suite *ReplayEventsSuite) Test_ReplayEvents() {
 			}
 			expectedDeployments := getLastStateFromDeployments(expectedEvents)
 			receivedDeployments := getLastStateFromDeployments(allEvents)
-			assert.Equal(t, len(expectedDeployments), len(receivedDeployments))
+			assert.Equal(t, len(expectedDeployments), len(receivedDeployments), "Not the same number of deployments.\nExpected: %+v\nActual: %+v", expectedDeployments, receivedDeployments)
```
fredrb, line 232: Same here for `.Len`

sensor/tests/replay/replay_test.go
lvalerom: `.Len` would print only one of the slices. We could do something like: 

```go
assert.Len(t, receivedAlerts, len(expectedAlerts), "Not the same number of alerts. Expected: %+v", expectedAlerts)
```

What do you think?

sensor/tests/replay/replay_test.go
lvalerom: Same as above

sensor/tests/replay/replay_test.go
fredrb: I think that it depends on what you want to assert. After looking at the code again it seems like this seem redundant to the code below? If the expected event isn't found in the list, the loop would print the specific event that is missing. If my understanding is correct, why would we benefit from printing the entire lists if the lengths are different?

lvalerom: The only use for the `len` assertions is in the case when we received more alerts/deployments than we expected. I'm no longer sure if that's a case we should test at all. If you agree, then I'll delete the `len` assertions.

fredrb: Sounds good

lvalerom: Done

fredrb (approved):

PR #4490, opened 2023-01-20
ROX-14101: Detect central installed on Openshift

0x656b694d:
## Description

Set the `ROX_OPENSHIFT_API` environment variable to the central deployment for Openshift installations, so that the telemetry code could identify Openshift as the platform on which the central is running:

https://github.com/stackrox/stackrox/blob/8f9fe014022f45953861e70ef55e4bcbc6c741dd/pkg/env/openshift.go#L5-L6

https://github.com/stackrox/stackrox/blob/8f9fe014022f45953861e70ef55e4bcbc6c741dd/central/telemetry/centralclient/instance_config.go#L49-L51

## Checklist

## Testing Performed

```
$ kubectl describe deployment central
...
    Environment:                                                                                                  
      ROX_NAMESPACE:            (v1:metadata.namespace)
      ROX_OFFLINE_MODE:        false                                                                              
      ROX_OPENSHIFT_API:       true                                                                               
      ROX_POSTGRES_DATASTORE:  true                                                                               
      ROX_DEVELOPMENT_BUILD:   true                                                                               
      ROX_HOTRELOAD:           false
      ROX_MANAGED_CENTRAL:     false
      ROX_NETWORK_ACCESS_LOG:  false
      ROX_POSTGRES_DATASTORE:  true
```

### DIFF
--- a/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
@@ -109,7 +109,7 @@ spec:
           value: "true"
         {{- end }}
         {{- if ._rox.env.openshift }}
-        - name: ROX_OPENSHIFT_API
+        - name: ROX_OPENSHIFT
           value: "true"
         {{- end }}
         {{- if ._rox.env.managedServices }}
--- a/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
@@ -108,6 +108,10 @@ spec:
         - name: ROX_ENABLE_OPENSHIFT_AUTH
           value: "true"
         {{- end }}
+        {{- if ._rox.env.openshift }}
+        - name: ROX_OPENSHIFT_API
+          value: "true"
+        {{- end }}
         {{- if ._rox.env.managedServices }}
         - name: ROX_MANAGED_CENTRAL
           value: "true"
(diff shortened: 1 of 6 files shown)


0x656b694d:
* master
  * **PR #4490** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4490" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4490?utm_source=stack-comment).

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -108,6 +108,10 @@ spec:
         - name: ROX_ENABLE_OPENSHIFT_AUTH
           value: "true"
         {{- end }}
+        {{- if ._rox.env.openshift }}
+        - name: ROX_OPENSHIFT_API
+          value: "true"
+        {{- end }}
```
dhaus67, line 114: I'm a bit confused by this change. If you look at the usages of ´ROX_OPENSHIFT_API´cas well as the description of it, you see that its only used within sensor deployments and used for changing the setup of informers there for resources on OpenShift.

I would say we should rather create a separate environment variable that is set, instead of reusing this one as it could be difficult to separate the contexts and meanings when its set (i.e. when set in sensor it means something regarding informers vs if its set in central its used to specify central is running on openshift for telemetry)

dhaus67 (changes requested):

dhaus67 (approved):

PR #4418, opened 2023-01-16
Merge API Reference and Help Center integration tests

pedrottimark:
## Description

Prerequisite to replace backend redirect for /docs/product with frontend link to OpenShift docs page.

### Changed files

1. Delete cypress/constants/Docs.js

2. Rename cypress/integration/apiReference.test.js as main/HelpMenu.test.js [too many changes to compare side-by-side]
    * Add `describe` block for Help Center.
    * Add `describe` block for version number.

3. Delete cypress/integration/docs.test.js

4. Edit src/Containers/MainPage/Header/HelpMenu.tsx
    * Merge items into one `ApplicationLauncherGroup` element to prepare for conditional rendering of Help Center, separator, and version number.

### Residue

1. Move files from credentialExpiry folder into main folder.
2. Add tests for other banners.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Edited integration tests

## Testing Performed

Ran tests individually with local deployment before and after editing.

Help menu has same appearance after merging two `ul` into one.
Help_menu

PR #4491, opened 2023-01-20
Add key prop in ApplicationLauncherGroup of HelpMenu

pedrottimark:
## Description

Dave, thank you for finding the error in the browser console:

key

My bad to omit `key` prop when I merged two groups in #4418

## Checklist

## Testing Performed

Manual testing with browser console.


dvail (approved):
Intuitively I probably would have done the same thing, given that there is just a single item being passed now. PatternFly must still `map` over the array internally though.

PR #4498, opened 2023-01-20
Track database backups and store database backup metadata

md2119:
## Description

Add infrastructure to track backups and persist backup metadata.

**Note:**
- The check to abort migration is not in this PR
- This is only forward-looking i.e. it is only enabled when running on postgres.
- If we decide to ship 3.74 with rocksDB default, I can go ahead and add rocksDB store and its rockdb-postgres migration.

## Checklist

## Testing Performed
CI

image


dashrews78 (changes requested):
I learned the hard way that we can't update the metadata service and its message.  It breaks sensors because they use jsonp to unmarshal in default mode which does not allow for unknown fields.  Most of that has been resolved BUT won't be delivered until 74 and won't make it to the sensors until the users chooses to upgrade them.  So sadly I think we need to put this info some other place.

Here is the ticket which references a slack thread for some additional info for you.
https://issues.redhat.com/browse/ROX-14188

md2119:
> I learned the hard way that we can't update the metadata service and its message. It breaks sensors because they use jsonp to unmarshal in default mode which does not allow for unknown fields. Most of that has been resolved BUT won't be delivered until 74 and won't make it to the sensors until the users chooses to upgrade them. So sadly I think we need to put this info some other place.
> 
> Here is the ticket which references a slack thread for some additional info for you. https://issues.redhat.com/browse/ROX-14188

Thanks for the heads-up, @dashrews78 . However, I am not changing the existing metadata service but adding a new service to the metadata server. Would that still break?

dashrews78:
> > I learned the hard way that we can't update the metadata service and its message. It breaks sensors because they use jsonp to unmarshal in default mode which does not allow for unknown fields. Most of that has been resolved BUT won't be delivered until 74 and won't make it to the sensors until the users chooses to upgrade them. So sadly I think we need to put this info some other place.
> > Here is the ticket which references a slack thread for some additional info for you. https://issues.redhat.com/browse/ROX-14188
> 
> Thanks for the heads-up, @dashrews78 . However, I am not changing the existing metadata service but adding a new service to the metadata server. Would that still break?

Yes that is fine.  That is what I had to do with the database status stuff in there.  Now that `google.protobuf.Timestamp backup_last_run_at = 5;` is out of the metadata message this is no longer a concern

central/globaldb/export/backup.go
```diff
@@ -20,57 +21,74 @@ import (
 )
 
 // Backup backs up the given databases (optionally removing secrets) and writes a ZIP archive to the given writer.
-func Backup(ctx context.Context, boltDB *bolt.DB, rocksDB *rocksdb.RocksDB, includeCerts bool, out io.Writer) error {
+func Backup(ctx context.Context, boltDB *bolt.DB, rocksDB *rocksdb.RocksDB, backupListener listener.BackupListener, includeCerts bool, out io.Writer) error {
 	zipWriter := zip.NewWriter(out)
 	defer utils.IgnoreError(zipWriter.Close)
 
+	listen := func(err error) error {
```
dashrews78, line 28: nit:  Not really using `listen` for this routine so we could remove it or not.

central/metadata/service/service_impl.go
```diff
@@ -166,3 +168,25 @@ func (s *serviceImpl) GetDatabaseStatus(ctx context.Context, _ *v1.Empty) (*v1.D
...
+		return nil, err
+	}
+	if !found {
+		return nil, errors.New("database backup status not found")
+	}
+	// Do not share requestor information for anonymous users.
+	if authn.IdentityFromContextOrNil(ctx) == nil {
+		sysInfo.BackupInfo.Requestor = nil
+	}
+	return &v1.DatabaseBackupStatus{
```
dashrews78, line 189: I wonder if all this info should be for authenticated users only.  I can't think what insight we would risk providing a backup time or that it succeeded to an user that is anonymous but it feels like too much.

md2119: I am not sure if the info is exploitable. I was wondering if and when operator communicates with Central for this info, would it be anonymous? 

That said, open to redact the message if you strongly feel about it.

dashrews78: I do not feel strongly one way or the other.  Figured I should pose the question just in case it triggered a thought for someone.  And that is a good point about operator but of course I do not know the answer to that.

dashrews78 (approved):

central/metadata/service/service_impl.go
md2119: defaulted to using admin sac

PR #4499, opened 2023-01-20
Dashrews/ROX-14469 multiple open rocks db in migrator

dashrews78:
## Description

Migrator opened up RocksDB in its main.  But later during a handful of migrations a call was made to `GetGlobalDackBox`.  The result of this call was there was another opening of RocksDB.  In cases where we were going from version X on Rocks to version X on Postgres, these 2 would work on the same backing directory.  That was the source of our corruption as you can only have one RocksDB open at a time.  So for migrator I mimicked what we did for Rocks and created a dackbox helper.  Part of this is because the migrator may not always want to point to current to do its work depending on the situation.  I also adjusted those to run once so we don't get in a situation of doing the same thing multiple times.

## Checklist

## Testing Performed

unit tests and gke-postgres-upgrade-tests cover most of this.  Spending a great deal of time manually though to make sure we can avoid the situation of getting in that corrupt state.


connorgorman (approved):
Looks good

c-du (approved):

migrator/dackboxhelper/singleton.go
```diff
@@ -0,0 +1,69 @@
...
+	toIndex  queue.WaitableQueue
+	registry indexer.WrapperRegistry
+
+	globalKeyLock concurrency.KeyFence
+
+	dackBox *dackbox.DackBox
+
+	initialized sync.Once
+
+	lazyStarted sync.Once
```
md2119, line 32: unused?

md2119 (approved):
lgtm

PR #4502, opened 2023-01-20
ROX-14372: Add PF version of time window dropdown

vjwilson:
## Description

This includes two related items:
1. Adding and wiring up a PF version of the Time Window dropdown to the new Network Graph toolbar
2. Adding the Last Updated time to the right side of the toolbar, as per the new mock, https://docs.google.com/presentation/d/1DdKZ0PoLk6yDyV9vFM4Q-HpB0N_lXwJ53G-CYAc56Sc/edit#slide=id.g1669bc43c7d_13_62 

## Checklist
- [x] Investigated and inspected CI test results (Go unit test failure on last run is test runner flake)

## Testing Performed

Default time window
Screen Shot 2023-01-20 at 2 24 46 PM

Changing to a new value
Screen Shot 2023-01-20 at 2 25 08 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/components/TimeWindowSelector.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/components/TimeWindowSelector.tsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+import { timeWindows } from 'constants/timeWindows';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+
+type TimeWindowSelectorProps = {
+    setActiveTimeWindow: (timeWindow) => void;
+    activeTimeWindow: string;
+    isDisabled?: boolean;
+};
+
+function TimeWindowSelector({
+    setActiveTimeWindow,
+    activeTimeWindow,
+    isDisabled = false,
+}: TimeWindowSelectorProps) {
+    const { closeSelect, isOpen, onToggle } = useSelectToggle();
+    function selectTimeWindow(_event, selection) {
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/components/TimeWindowSelector.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+import { timeWindows } from 'constants/timeWindows';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
+
+type TimeWindowSelectorProps = {
+    setActiveTimeWindow: (timeWindow) => void;
+    activeTimeWindow: string;
```
dvail, line 9: Would typing these values with the same `typeof timeWindows[number]` used in the state of NetworkGraphPage cause problems with PF's Select component props?

ui/apps/platform/src/Containers/NetworkGraph/components/TimeWindowSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+        <Select
+            isOpen={isOpen}
+            onToggle={onToggle}
+            onSelect={selectTimeWindow}
+            selections={activeTimeWindow}
+            isDisabled={isDisabled}
+        >
+            {timeWindows.map((window) => (
```
dvail, line 32: I know that this is shadowed correctly, but naming a local var the same thing as the global `window` object makes me uneasy.

dvail (approved):
Two nits but otherwise LGTM

PR #4508, opened 2023-01-23
ROX-14335: PVC extension should not touch PVCs that are not its target

porridge:
## Description

Each instance of the `ReconcilePVCExtension` should only care about the PVC resources which match its target (where matching is in broad sense, for backwards compatibility reasons). This was mostly true, except for the `handleDelete()` case, which lead to unintended removal of owner references from **all owned** PVCs.

See [more detailed description in the ticket](https://issues.redhat.com/browse/ROX-14335?focusedCommentId=21590623&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21590623).

This changes:
- `handleDelete()` to ignore PVCs that are not for the current target,
- adds a unit test and a kuttl test for this case
- improves logging a bit
- fixes some typos

## Checklist

This definitely requires some user-facing communication, since some repair actions would be necessary for almost all 3.73.x operator users if they want to re-gain full functionality. Still need to investigate how to best approach this.

## Testing Performed

Ran the new unit tests which failed on old codebase. Depending on CI otherwise.

### DIFF
--- a/operator/pkg/central/extensions/reconcile_pvc.go
+++ b/operator/pkg/central/extensions/reconcile_pvc.go
@@ -113,7 +113,7 @@ func reconcilePVC(ctx context.Context, central *platform.Central, persistence *p
 		persistence:      persistence,
 		target:           target,
 		defaultClaimName: defaultClaimName,
-		log:              log,
+		log:              log.WithValues("pvcReconciler", target),
 	}
 
 	return ext.Execute()
--- a/operator/pkg/central/extensions/reconcile_pvc.go
+++ b/operator/pkg/central/extensions/reconcile_pvc.go
@@ -186,7 +186,7 @@ func (r *reconcilePVCExtensionRun) Execute() error {
 }
 
 func (r *reconcilePVCExtensionRun) handleDelete() error {
-	ownedPVCs, err := r.getOwnedPVC()
+	ownedPVCs, err := r.getOwnedPVCsForCurrentTarget()
 	if err != nil {
 		return errors.Wrap(err, "fetching operator owned PVCs")
 	}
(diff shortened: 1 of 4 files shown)


operator/pkg/central/extensions/reconcile_pvc.go
```diff
@@ -275,52 +275,53 @@ func parseResourceQuantityOr(qStrPtr *string, d resource.Quantity) (resource.Qua
...
+
+func (r *reconcilePVCExtensionRun) getTargetLabelValue(pvc *corev1.PersistentVolumeClaim) string {
+	// If the PCV is not owned by the Central we're reconciling, do not make any assumptions about the target.
+	if !metav1.IsControlledBy(pvc, r.centralObj) {
+		return ""
+	}
+	if val, ok := pvc.Labels[pvcTargetLabelKey]; ok {
+		return val
+	}
+	// If the target annotation is not set, assume this (owned) PVC is a Central PVC for backwards compatibility.
```
SimonBaeumer, line 304: Is the Central PVC = rocksdb?

SimonBaeumer (approved):
@porridge LGTM for the fix.

Suggestion to fix existing customer environments:
To automatically fix Central CRs without CLI access,  could an annotation on the CRD force the operator to reclaim the PVC?
Something like: `stackrox.io/reclaim-pvc`?

porridge:
> Suggestion to fix existing customer environments: To automatically fix Central CRs without CLI access, could an annotation on the CRD force the operator to reclaim the PVC? Something like: `stackrox.io/reclaim-pvc`?

If a customer is able to change `medatadata.annotations` the CRD, then surely they can also use a similar method to change `metadata.ownerReferences` on the CR, no?

PR #4510, opened 2023-01-23
Dashrews/ROX-14271 metric active db connections

dashrews78:
## Description

Adds a metric for active database connections by database.

## Checklist

## Testing Performed

Ran and inspected the metrics files from the diagnostic and debug bundles.

From metrics-1:
```
# HELP rox_central_postgres_total_connections number of total connections to Postgres by state
# TYPE rox_central_postgres_total_connections gauge
rox_central_postgres_total_connections{state="active"} 1
rox_central_postgres_total_connections{state="idle"} 81
...
# HELP rox_central_postgres_maximum_db_connections number of total connections allowed to the Postgres database server
# TYPE rox_central_postgres_maximum_db_connections gauge
rox_central_postgres_maximum_db_connections 200
```

### DIFF
--- a/central/globaldb/metrics/metrics.go
+++ b/central/globaldb/metrics/metrics.go
@@ -56,7 +56,6 @@ func init() {
 		PostgresDBSize,
 		PostgresTotalSize,
 		PostgresConnected,
-		PostgresActiveConnections,
 		PostgresTotalConnections,
 		PostgresMaximumConnections,
 	)
@@ -208,13 +207,6 @@ var (
 		Name:      "postgres_connected",
 		Help:      "flag indicating if central is connected to the Postgres Database. 0 NOT connected, 1 connected",
 	})
-
-	PostgresActiveConnections = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Namespace: metrics.PrometheusNamespace,
-		Subsystem: metrics.CentralSubsystem.String(),
-		Name:      "postgres_active_db_connections",
-		Help:      "number of active connections to Postgres by database name",
-	}, []string{"database"})
 
 	PostgresTotalConnections = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metrics.PrometheusNamespace,
--- a/central/globaldb/metrics/metrics.go
+++ b/central/globaldb/metrics/metrics.go
@@ -56,6 +56,9 @@ func init() {
 		PostgresDBSize,
 		PostgresTotalSize,
 		PostgresConnected,
+		PostgresActiveConnections,
+		PostgresTotalConnections,
+		PostgresMaximumConnections,
 	)
 }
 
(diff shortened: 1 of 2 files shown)


central/globaldb/postgres.go
```diff
@@ -57,6 +57,8 @@ SELECT TABLE_NAME
 ) a;`
 
 	versionQuery = `SHOW server_version;`
+
+	connectionQuery = `SELECT datname, COUNT(datid) FROM pg_stat_activity WHERE state <> 'idle' AND datname IS NOT NULL GROUP BY datname;`
```
erthalion, line 61: Probably makes sense to exclude current connection where this query is executed?

```
where pid != pg_backend_pid()
```

Also would be quite useful to know max_connections together with this metric, is it exposed somewhere? Otherwise in the same query `current_setting('max_connections')::int`.

central/globaldb/postgres.go
```diff
@@ -231,11 +233,39 @@ func CollectPostgresDatabaseStats(postgresConfig *pgxpool.Config) {
...
 }
 
+// CollectPostgresConnectionStats -- collect connection stats for Postgres
+func CollectPostgresConnectionStats(ctx context.Context, db *pgxpool.Pool) {
+	ctx, cancel := context.WithTimeout(ctx, PostgresQueryTimeout)
+	defer cancel()
+
+	row, err := db.Query(ctx, connectionQuery)
+	if err != nil {
+		log.Errorf("error fetching object counts: %v", err)
```
connorgorman, line 243: not object counts

stehessel (approved):
Agreed with @erthalion comments, otherwise lgtm.

central/globaldb/postgres.go
dashrews78: I left the active one as is.  Figure if we excluded the one making the request we could easily forget that it counts as we approach the limit.  I did add a similar one for total that would include active and idle.  And then as suggested an overarching one for the max connections allowed.

stehessel: > Figure if we excluded the one making the request we could easily forget that it counts as we approach the limit. 

Fair enough 👍

central/globaldb/metrics/metrics.go
```diff
@@ -201,6 +204,27 @@ var (
...
 	})
+
+	PostgresActiveConnections = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: metrics.PrometheusNamespace,
+		Subsystem: metrics.CentralSubsystem.String(),
+		Name:      "postgres_active_db_connections",
+		Help:      "number of active connections to Postgres by database name",
+	}, []string{"database"})
+
+	PostgresTotalConnections = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
connorgorman, line 215: I would suggest that we don't split these, and instead add it as a label so it would be like state: idle, active. This lets you visualize everything more easily

connorgorman: and also takes into accounts other states besides active and not active

erthalion:
Just wanted to reiterate what we were discussing couple of days ago. The PR is fine if it's one-off, but I doubt it's acceptable to implement general db metric collection mechanism this way, it has to be separated from Central.

stehessel:
I'm currently working on https://issues.redhat.com/browse/ROX-14270, which includes exporting AWS RDS metrics via a prometheus exporter. In addition to the ACU metrics, we'll have in principle access to the metrics listed here https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.AuroraMySQL.Monitoring.Metrics.html. @dashrews78 Maybe you can check if the metrics in this PR could be better exported from the list above.

connorgorman (approved):

PR #4511, opened 2023-01-23
Dashrews/ROX-14424 properly set migration yaml on pg upgrade last persisted time

dashrews78:
## Description

Found an issue where a customer tried to go to 3.73.1 R to 3.73.1 P.  There was no real way to rollback because the act was essentially changing the flag to turn Postgres on.  This change and #4475 make sure the version file is adjusted accordingly to allow for that case.  Unfortunately that didn't fix each scenario and we cannot rely singularly on that because for older versions that time will not be updated.  We have to know that we have used RocksDB more recently than Postgres to re-run the migrations.  

Case 1:  RocksDB v1 -> PostgresDB v2 -> RocksDB v1 -> PostgresDB v2
In this case the LastPersistedTime will be older than the LastPersistedTime in Postgres because v1 would not have the code changes to write the LastPersistedTime each iteration.  When the `.previous` moves to `.current` that timestamp is updated so we can use timestamp on `.current` to determine that we went back to RocksDB and thus need to re-run the migrations when we go back to Postgres.

Case 2:  RocksDB v2 -> PostgresDB v2 -> RocksDB v2 -> PostgresDB v2
In this case RocksDB itself was not updated because it did not need to execute any migrations.  Thus the timestamp on `.current` was unchanged.  Thus when we go back to Postgres the 2nd time, the migrations were not executed because we could not tell that RocksDB had been the most recently used database.  Thus the changes to make sure LastPersistedTime and the migration_version.yaml is written each time the migrator is executed, not just when the version changes.

We can use a combination of the timestamp on `.current` for older versions and the LastPersistedTimestmap to determine if RocksDB was used more recently than Postgres or not.

Additionally, I decided to pull out the DecommissionRocksDB code and annotate it for re-enabling later.  My concern is that if we have churn and a couple of patches then we may destroy the rocks database before we would want.

## Checklist

## Testing Performed

UT was updated.  Lots of manual testing of the scenarios outlined below.
3.73 R -> 3.73 P -> 3.73 R -> 3.73P
3.72 R -> 3.73 P -> 3.72 R -> 3.73P
3.72 R -> 3.73 P -> 3.73 R -> 3.73 P -> 3.72 R

Note the case of a rollback to a previous version using helm and the operator will wipe out central-db so that upgrading again should be seamless.

### DIFF
--- a/migrator/clone/db_clone_manager_impl.go
+++ b/migrator/clone/db_clone_manager_impl.go
@@ -81,6 +81,9 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate() (string, string, string, error)
 			rocksVersion = d.dbmRocks.GetVersion(rocksdb.RestoreClone)
 		} else {
 			rocksVersion = d.dbmRocks.GetVersion(rocksdb.CurrentClone)
+			if rocksVersion != nil {
+				rocksVersion.LastPersisted = d.dbmRocks.GetCurrentCloneCreationTime()
+			}
 		}
 
 		pgClone, migrateFromRocks, err = d.dbmPostgres.GetCloneToMigrate(rocksVersion, restoreFromRocks)
--- a/migrator/clone/db_clone_manager_impl.go
+++ b/migrator/clone/db_clone_manager_impl.go
@@ -81,9 +81,6 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate() (string, string, string, error)
 			rocksVersion = d.dbmRocks.GetVersion(rocksdb.RestoreClone)
 		} else {
 			rocksVersion = d.dbmRocks.GetVersion(rocksdb.CurrentClone)
-			if rocksVersion != nil {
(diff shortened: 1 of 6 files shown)


migrator/clone/db_clone_manager_impl.go
```diff
@@ -128,18 +128,20 @@ func (d *dbCloneManagerImpl) Persist(cloneName string, pgClone string, persistBo
 			return err
 		}
 
+		// TODO (ROX-14524):  re-enable this in 3.75.  This will not occur until multiple upgrades of Postgres
```
c-du, line 131: This will occur if you have multiple upgrades including 73 and 74 small upgrade. It will happen.

dashrews78: Right.  any multiple upgrades.  Figured it is best to not do this now and save it for later, probably re-enable when the PVC is deprecated

dashrews78:
No longer going to support going back to Postgres after a downgrade to Rocks in migration code.

PR #4513, opened 2023-01-23
Fix styles in PF network graph filters

vjwilson:
## Description

Fixes a couple of style ooopsies for new PatternFly Network Graph.
1. Classic key-value filter looked dark in dark mode. Used same forced `theme-light` solution as we have used in other new PF sections, until we get the PF dark mode piece in. (Thanks to @dvail for showing me this.)
2. Adding an even higher stacking context for the hierarchy selector dropdowns, so they don't run "under" the toolbar, which has a stacking context "over" the graph itself.

## Checklist

## Testing Performed

Fixed filter search bar in dark mode
Screen Shot 2023-01-23 at 10 38 27 AM

Fixed hierarchy dropdowns stacking order
Screen Shot 2023-01-23 at 10 38 13 AM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
@@ -11,6 +11,12 @@
     height: calc(100vh - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight));
 }
 
+/* Create a stacking context that is higher than the context for the network-graph-toolbar component
+   that is rendered below this toolbar, so that the hierarchy dropdowns will render above everything when open */
+.network-graph-selector-bar {
+    z-index: 350;
+}
+
 /* Create a stacking context that is higher than the context for the react-topology component
    that is rendered below the toolbar, so that the search filter dropdown will render above the graph when open */
 [data-testid="network-graph-toolbar"] {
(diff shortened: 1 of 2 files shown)


dvail (approved):
LGTM

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
```diff
@@ -11,6 +11,12 @@
     height: calc(100vh - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight) - var(--pf-c-page__header--MinHeight));
 }
 
+/* Create a stacking context that is higher than the context for the network-graph-toolbar component
+   that is rendered below this toolbar, so that the hierarchy dropdowns will render above everything when open */
+.network-graph-selector-bar {
+    z-index: 350;
```
pedrottimark, line 17: Too bad that it needs z-index instead of simple static display.

pedrottimark (approved):

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
vjwilson: I did not test other ways of creating a stacking context, in part because z-index is the most explicit way to force it (you can tell what it does even if you're debugging in dev tools, instead of reading the comment in the source code), and in part because of PTSD from the old Internet-Explorer "hasLayout" hacks. 😉

PR #4516, opened 2023-01-23
ROX-14487: Improvements of Group and Identify messages

0x656b694d:
## Description

* Add `*UserAs` methods to `Telemeter` interface, to alter UserID and/or set device context for a particular call. This way a central could Identify a secured cluster from its name.

## Checklist

### Central identity
```json
{
  "context": {
    "Client ID": "6393af03-28ac-4f5f-bf35-b0d10be9f99f",
    "Client Name": "Central",
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    }
  },
  "integrations": {},
  "messageId": "97377717-545e-4a41-bbe4-01d29a536c43",
  "originalTimestamp": "2023-01-23T17:51:39.924036459Z",
  "receivedAt": "2023-01-23T17:52:40.666Z",
  "sentAt": "2023-01-23T17:52:39.745Z",
  "timestamp": "2023-01-23T17:51:40.844Z",
  "traits": {
    "Auth Providers": null,
    "Central version": "3.73.x-505-g211a07a034-dirty",
    "Chart version": "73.0.505-g211a07a034-dirty",
    "Kind": "Central",
    "Kubernetes version": "v1.25.3",
    "Orchestrator": "KUBERNETES_CLUSTER",
    "Total Access Scopes": 2,
    "Total PermissionSets": 9,
    "Total Roles": 9,
    "Total Secured Clusters": 1,
    "Total Signature Integrations": 0
  },
  "type": "identify",
  "userId": "6393af03-28ac-4f5f-bf35-b0d10be9f99f",
  "writeKey": "..."
}
```

### Secured Cluster identity
```json
{
  "context": {
    "Client ID": "6393af03-28ac-4f5f-bf35-b0d10be9f99f",
    "Client Name": "Central",
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    }
  },
  "integrations": {},
  "messageId": "0b6bf669-4361-435b-ad79-9528a8e84684",
  "originalTimestamp": "2023-01-23T18:01:18.199106737Z",
  "receivedAt": "2023-01-23T18:01:40.408Z",
  "sentAt": "2023-01-23T18:01:39.745Z",
  "timestamp": "2023-01-23T18:01:18.861Z",
  "traits": {
    "Admission Controller": true,
    "CPU Capacity": 8,
    "Cluster Type": "KUBERNETES_CLUSTER",
    "Collection Method": "EBPF",
    "Collector Image": "quay.io/stackrox-io/collector",
    "Kind": "Secured Cluster",
    "Main Image": "quay.io/stackrox-io/main:3.73.x-505-g211a07a034-dirty",
    "Managed By": "MANAGER_TYPE_MANUAL",
    "Priority": 1,
    "Slim Collector": true,
    "Total Nodes": 1
  },
  "type": "identify",
  "userId": "5b72c54b-bd81-416a-86fc-15e63069f969",
  "writeKey": "..."
}
```

### DIFF
--- a/central/cluster/datastore/telemetry.go
+++ b/central/cluster/datastore/telemetry.go
@@ -18,7 +18,7 @@ func trackClusterRegistered(cluster *storage.Cluster) {
 			"Cluster ID":   cluster.GetId(),
 			"Managed By":   cluster.GetManagedBy().String(),
 		}
-		cfg.Telemeter().Track("Secured Cluster Registered", cfg.ClientID, props)
+		cfg.Telemeter().Track("Secured Cluster Registered", props)
 	}
 }
 
@@ -37,11 +37,18 @@ func makeClusterProperties(cluster *storage.Cluster) map[string]any {
 
 func trackClusterInitialized(cluster *storage.Cluster) {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
-		cfg.Telemeter().Group(cfg.GroupID, cluster.GetId(), nil)
-		cfg.Telemeter().Identify(cluster.GetId(), "Secured Cluster", makeClusterProperties(cluster))
-		cfg.Telemeter().Track("Secured Cluster Initialized", cluster.GetId(), map[string]any{
-			"Health": cluster.GetHealthStatus().GetOverallHealthStatus().String(),
-		})
+		// Add the secured cluster 'user' to the Tenant group:
+		cfg.Telemeter().GroupUserAs(cluster.GetId(), "", "", cfg.GroupID, nil)
+
+		// Update the secured cluster identity from its name:
+		cfg.Telemeter().IdentifyUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
+			makeClusterProperties(cluster))
+
+		// Issue an event that makes the identity effective:
+		cfg.Telemeter().TrackUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
+			"Secured Cluster Initialized", map[string]any{
+				"Health": cluster.GetHealthStatus().GetOverallHealthStatus().String(),
+			})
 	}
 }
 
@@ -75,7 +82,7 @@ func UpdateSecuredClusterIdentity(ctx context.Context, clusterID string, metrics
 		props := makeClusterProperties(cluster)
 		props["Total Nodes"] = metrics.NodeCount
 		props["CPU Capacity"] = metrics.CpuCapacity
-		cfg.Telemeter().Identify(cluster.GetId(), "Secured Cluster", props)
-		cfg.Telemeter().Track("Updated Secured Cluster Identity", cluster.GetId(), nil)
+		cfg.Telemeter().IdentifyUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster", props)
+		cfg.Telemeter().TrackUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster", "Updated Secured Cluster Identity", nil)
 	}
 }
--- a/central/cluster/datastore/telemetry.go
+++ b/central/cluster/datastore/telemetry.go
@@ -19,6 +19,13 @@ func trackClusterRegistered(cluster *storage.Cluster) {
 			"Managed By":   cluster.GetManagedBy().String(),
 		}
 		cfg.Telemeter().Track("Secured Cluster Registered", props)
+
+		// Add the secured cluster 'user' to the Tenant group:
+		cfg.Telemeter().GroupUserAs(cluster.GetId(), "", "", cfg.GroupID, nil)
+
+		// Update the secured cluster identity from its name:
+		cfg.Telemeter().IdentifyUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
+			makeClusterProperties(cluster))
 	}
 }
 
@@ -37,14 +44,7 @@ func makeClusterProperties(cluster *storage.Cluster) map[string]any {
 
 func trackClusterInitialized(cluster *storage.Cluster) {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
-		// Add the secured cluster 'user' to the Tenant group:
-		cfg.Telemeter().GroupUserAs(cluster.GetId(), "", "", cfg.GroupID, nil)
-
-		// Update the secured cluster identity from its name:
-		cfg.Telemeter().IdentifyUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
-			makeClusterProperties(cluster))
-
-		// Issue an event that makes the identity effective:
+		// Issue an event that makes the secured cluster identity effective:
 		cfg.Telemeter().TrackUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
 			"Secured Cluster Initialized", map[string]any{
 				"Health": cluster.GetHealthStatus().GetOverallHealthStatus().String(),
--- a/central/cluster/datastore/telemetry.go
+++ b/central/cluster/datastore/telemetry.go
@@ -38,7 +38,7 @@ func makeClusterProperties(cluster *storage.Cluster) map[string]any {
 func trackClusterInitialized(cluster *storage.Cluster) {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
 		cfg.Telemeter().Group(cfg.GroupID, cluster.GetId(), nil)
-		cfg.Telemeter().Identify(cluster.GetId(), makeClusterProperties(cluster))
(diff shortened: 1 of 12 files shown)


0x656b694d:
* master
  * **PR #4516** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4516" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4516?utm_source=stack-comment).

pkg/telemetry/phonehome/telemeter.go
```diff
@@ -8,16 +8,27 @@ type Telemeter interface {
...
-func (t *nilTelemeter) Identify(userID string, props map[string]any)       {}
-func (t *nilTelemeter) Track(event, userID string, props map[string]any)   {}
-func (t *nilTelemeter) Group(groupID, userID string, props map[string]any) {}
+func (t *nilTelemeter) Stop()                            {}
+func (t *nilTelemeter) Identify(_ map[string]any)        {}
+func (t *nilTelemeter) Track(_ string, _ map[string]any) {}
+func (t *nilTelemeter) Group(_ string, _ map[string]any) {}
+
+func (t *nilTelemeter) IdentifyUserAs(_, _, _ string, _ map[string]any) {}
+func (t *nilTelemeter) TrackUserAs(_, _, _, _ string, _ map[string]any) {}
```
stehessel, line 33: Shouldn't this be called `TrackAsUser`? Otherwise it sounds like we are tracking the user.

pkg/telemetry/phonehome/telemeter.go
```diff
@@ -8,16 +8,27 @@ type Telemeter interface {
...
-	Track(event, userID string, props map[string]any)
+	Track(event string, props map[string]any)
 	// Group adds the user to a group, supplying group specific properties.
-	Group(groupID, userID string, props map[string]any)
+	Group(groupID string, props map[string]any)
+
+	// IdentifyUserAs updates the user traits.
+	IdentifyUserAs(userID, clientID, clientType string, props map[string]any)
+	// TrackUserAs registers an event, caused by the user.
+	TrackUserAs(userID, clientID, clientType, event string, props map[string]any)
```
stehessel, line 20: Why do you leave out the type annotations of `userID`, `clientID` and `clientType`?

pkg/telemetry/phonehome/telemeter.go
```diff
@@ -8,16 +8,27 @@ type Telemeter interface {
...
 
 type nilTelemeter struct{}
 
-func (t *nilTelemeter) Stop()                                              {}
-func (t *nilTelemeter) Identify(userID string, props map[string]any)       {}
-func (t *nilTelemeter) Track(event, userID string, props map[string]any)   {}
-func (t *nilTelemeter) Group(groupID, userID string, props map[string]any) {}
+func (t *nilTelemeter) Stop()                            {}
+func (t *nilTelemeter) Identify(_ map[string]any)        {}
+func (t *nilTelemeter) Track(_ string, _ map[string]any) {}
```
stehessel, line 29: Why did you remove the parameter names? I'd leave them for better context.

pkg/telemetry/phonehome/telemeter.go
0x656b694d: The second two parameters are client impersonation, so the full name could have been `TrackUserEventAsClient`.
Anyway, what does tracking a user mean? Watching their steps, and that's essentially what this method does: it tracks what a user is doing. If it makes sense, I'd keep the name.

pkg/telemetry/phonehome/telemeter.go
0x656b694d: Would you prefer
```
TrackUserAs(userID, clientID, clientType string, event string, props map[string]any)
```
or
```
TrackUserAs(userID string, clientID string, clientType string, event string, props map[string]any)
```
?

pkg/telemetry/phonehome/telemeter.go
0x656b694d: well, they don't matter here, produce noise, and take a lot of space.

pkg/telemetry/phonehome/telemeter.go
stehessel: I would prefer the second one, but I don't know much about golang style conventions.

stehessel (approved):

pkg/telemetry/phonehome/telemeter.go
0x656b694d, line 20:
```suggestion
	IdentifyUserAs(userID string, clientID string, clientType string, props map[string]any)
	// TrackUserAs registers an event, caused by the user.
	TrackUserAs(userID string, clientID string, clientType string, event string, props map[string]any)
```

0x656b694d: I'd address this issue eventually in a separate PR, as we're struggling with build stability, and I'd prefer to merge this PR while it's in a mergeable state.

central/cluster/datastore/telemetry.go
```diff
@@ -37,11 +44,11 @@ func makeClusterProperties(cluster *storage.Cluster) map[string]any {
 
 func trackClusterInitialized(cluster *storage.Cluster) {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
-		cfg.Telemeter().Group(cfg.GroupID, cluster.GetId(), nil)
-		cfg.Telemeter().Identify(cluster.GetId(), makeClusterProperties(cluster))
-		cfg.Telemeter().Track("Secured Cluster Initialized", cluster.GetId(), map[string]any{
-			"Health": cluster.GetHealthStatus().GetOverallHealthStatus().String(),
-		})
+		// Issue an event that makes the secured cluster identity effective:
+		cfg.Telemeter().TrackUserAs(cluster.GetId(), cluster.GetId(), "Secured Cluster",
```
dhaus67, line 48: Probably want `Secured Cluster` as a `const` here, seeing that it's reused multiple times.

central/role/mapper/telemetry.go
```diff
@@ -9,6 +9,6 @@ import (
 // such users could be segmented by tenant properties.
 func addUserToTenantGroup(user *storage.User) {
 	if cfg := centralclient.InstanceConfig(); cfg.Enabled() {
-		cfg.Telemeter().Group(cfg.GroupID, cfg.HashUserID(user.GetId(), user.GetAuthProviderId()), nil)
+		cfg.Telemeter().GroupUserAs(cfg.HashUserID(user.GetId(), user.GetAuthProviderId()), cfg.GroupID, "", "", nil)
```
dhaus67, line 12: Wait why do we suddenly use `cfg.GroupID` as the client ID, shouldn't it be similar to  `cfg.Telemeter().GroupUserAs(cfg.HashUserID(basic.DefaultUsername, basicAuthProvider.ID()), "", "", cfg.GroupID, nil)` and the second to last argument?

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -89,14 +88,39 @@ func (t *segmentTelemeter) Stop() {
...
+			Type: clientType,
+		},
+	}
+}
+
+func (t *segmentTelemeter) Identify(props map[string]any) {
+	t.IdentifyUserAs("", "", "", props)
+}
+
+func (t *segmentTelemeter) IdentifyUserAs(userID, clientID, clientType string, props map[string]any) {
```
dhaus67, line 114: It's a bit unfortunate that we have the parameters but their usage varies based on the context of callers (i.e. sometimes callers will explicitly set `clientID, clientType`, other times they will leave it empty).

I'm also not quite sure, just reading through the code, when its OK to leave those things empty and when not, I suppose.

For the clarity on which parameters to use, albeit it might become overkill, but will allow us to extend the API further in the future, how about creating `Options` that can be passed to the `*UserAs` methods, so its more explicit what parameters are used and we get rid of the "empty string" parameters (since go doesn't allow for default parameter values)?
As an example, something similar to[ the auth provider options we have](https://github.com/stackrox/stackrox/blob/1b934bf19e1f60a23f3e0f35dbcde15584e29c24/pkg/auth/authproviders/provider_options.go#L20) would make it more explicit which values to use.

central/role/mapper/telemetry.go
0x656b694d: that's a good catch!

central/cluster/datastore/telemetry.go
0x656b694d: makes sense!

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: Any change like this would imply a change in the interface, and Segment implementation doesn't import it, and cannot implement. That's unfortunate.
A possible solution would be to extract the interface to a third package. I'll see if that will look good and go-ish enough.

PR #4520, opened 2023-01-24
Postgres migration to create policy categories, and policy category edges for all policies and corresponding categories

clickboo:
## Description
Migration to create categories and policy <-> category edges

## Checklist

## Testing Performed
Automated only


migrator/migrations/m_169_to_m_170_create_policy_categories_and_edges/migration.go
```diff
@@ -0,0 +1,138 @@
+package m169tom170
```
dashrews78, line 1: @keyallis  already merged `m169tom170`.  You will need to become `m170tom171` unless @charmik-redhat pushes his before you do.  Due to the name differences of the packages these will not show as merge conflicts.

migrator/migrations/n_57_to_n_58_postgres_create_policy_category_edges/migration.go
```diff
@@ -0,0 +1,139 @@
+package n_57_to_n_58_postgres_create_policy_category_edges
```
dashrews78, line 1: I think I would rather this be m_170_to_m_171 and the create edge be m_171_to_m_172.  Or even just a single migration.  ACSCS has previously talked about using more rapid builds.  If they were to have done that then they could have already executed m_168_to_m_169 in theory.  If that were to be the case then their next upgrade would not execute this one.  It also makes the math fuzzy.

dashrews78 (commented):
Don't forget about `pkg/migrations/internal/seq_num.go`

migrator/migrations/n_57_to_n_58_postgres_create_policy_category_edges/migration.go
clickboo: Aah yes, this needs to be removed. This was prior to our m and n discussion.

dashrews78 (changes requested):
Need to bump `CurrentDBVersionSeqNum` in `pkg/migrations/internal/seq_num.go`

clickboo:
> Need to bump `CurrentDBVersionSeqNum` in `pkg/migrations/internal/seq_num.go`

Rebase with conflicts in migrations are a pain :(, done

migrator/migrations/m_170_to_m_171_create_policy_categories_and_edges/migration.go
```diff
@@ -0,0 +1,138 @@
...
+	Categories []string `gorm:"column:categories;type:varchar"`
+}
+
+type Categories struct {
+	Id   string `gorm:"column:id;type:varchar;primaryKey"`
+	Name string `gorm:"column:name;type:varchar"`
+}
+
+func CreatePolicyCategoryEdges(gormDB *gorm.DB, db *pgxpool.Pool) error {
+	// TODO: check
```
dashrews78, line 53: nit:  TODO

migrator/migrations/m_170_to_m_171_create_policy_categories_and_edges/migration.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			return nil
+		},
+	}
+
+	batchSize = 500
+	log       = loghelper.LogWrapper{}
+)
+
+type PolicyIDAndCategories struct {
```
dashrews78, line 42: I think these can be removed since they are in  `policy_category_edges`

clickboo: Aah yes, remnants from pre winter break migration code that I was working on.

dashrews78 (approved):
LGTM.  If you haven't done so already, I would recommend deploying with 3.73.1, making some policy changes, upgrading to this image to make sure it runs how you think.

PR #4523, opened 2023-01-24
ROX-14335: Remediate bug and verify in operator upgrade test

porridge:
## Description

Followup to #4508 that tests the manual remediation instructions.
In a separate PR at least for now, since this can be iterated on as a followup from the fix.

## Checklist
- [x] Checked the debug output in the remediation step to make sure the PVC is as expected before and after the patch

Actually the remediation commands are a no-op, because 3.74.x was still not tagged, and thus the "previous version" is still considered to be 3.72 which does not have the bug. But this will change as soon as the 3.74.x tag is created.

The last two checkboxes are being discussed in the ticket.

## Testing Performed

Ran the script manually and inspected output.
Otherwise CI plus the check mentioned above should be sufficient.

### DIFF
--- a/operator/tests/common/upgrade-assert.envsubst.yaml
+++ b/operator/tests/common/upgrade-assert.envsubst.yaml
@@ -23,6 +23,22 @@ metadata:
 status:
   availableReplicas: 1
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: stackrox-db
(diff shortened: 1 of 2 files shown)


connorgorman (approved):

porridge:
The failure is an unrelated flake tracked in https://issues.redhat.com/browse/ROX-14495

PR #4528, opened 2023-01-25
Network Graph PF edge perf improvement and refactor

alwayshooin:
as it states in the title --
* refactored the data filtering/modifying to be in its own layer (`NetworkGraphContainer`)
* also separated out `TopologyComponent` into its own file and simplified it so that it simply renders whatever the model it is given
* this greatly improves performance of toggling display options when a side panel is open (you should not notice any performance issues)
* display options should persist between selecting a node on the graph and also between edge states

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
@@ -36,6 +36,8 @@ function getFilteredEdges(
                 if (children?.includes(source) || children?.includes(target)) {
                     filteredEdges.push({ ...edge, visible: true });
                 }
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
             } else if (source === selectedNode.data?.id || target === selectedNode.data?.id) {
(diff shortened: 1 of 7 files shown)


vjwilson (approved):
Bob the Builder-level moving stuff around

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
```diff
@@ -0,0 +1,233 @@
...
+            const { source, target } = edge;
+            const { type } = selectedNode.data;
+            if (type === 'NAMESPACE' || type === 'EXTERNAL_GROUP') {
+                // if a namespace is selected, add children's node edges
+                const { children } = selectedNode;
+                if (children?.includes(source) || children?.includes(target)) {
+                    filteredEdges.push({ ...edge, visible: true });
+                }
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
```
alwayshooin, line 40: not sure why ts is failing here since i check against `EXTRANEOUS` above, so i just commented ts-ignore for now

PR #4530, opened 2023-01-25
ROX-14428: Create stub manager for declarative configuration

dhaus67:
## Description

This PR is a pre-factor for future PRs for the declarative config feature.

It will create a coarse structure of a manager that starts a watch on a specific directory.
Additionally, the watch will be conditionally started in the main routine _iff_ the feature flag is enabled.

In future PRs, the manager implementation will be expanded to also cover the reading file contents from the directory, marshalling these YAMLs to structs, transforming the structs to proto messages, and reconciling the proto messages within the stores.

## Checklist

### DIFF
--- a/central/declarativeconfig/manager.go
+++ b/central/declarativeconfig/manager.go
@@ -1,6 +1,6 @@
 package declarativeconfig
 
-// Manager manages reconciling declarative configurations.
+// Manager manages reconciling declarative configuration.
 type Manager interface {
 	WatchDeclarativeConfigDir()
 }
--- a/central/declarativeconfig/manager.go
+++ b/central/declarativeconfig/manager.go
@@ -1,0 +1,6 @@
+package declarativeconfig
+
+// Manager manages reconciling declarative configurations.
+type Manager interface {
+	WatchDeclarativeConfigDir()
+}
--- a/central/declarativeconfig/manager_impl.go
+++ b/central/declarativeconfig/manager_impl.go
@@ -1,0 +1,41 @@
+package declarativeconfig
+
+import (
+	"context"
+	"time"
+
+	"github.com/stackrox/rox/pkg/k8scfgwatch"
+	"github.com/stackrox/rox/pkg/logging"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+const (
+	watchInterval        = 5 * time.Second
+	declarativeConfigDir = "/run/stackrox.io/declarative-configuration"
+)
+
+var (
+	log = logging.LoggerForModule()
+)
+
+type managerImpl struct {
+	once sync.Once
+}
+
+// New creates a new instance of Manager.
+// Note that it will not watch the declarative configuration directories when creaeted, only after
+// WatchDeclarativeConfigDir has been called.
+func New() Manager {
+	return &managerImpl{}
+}
+
+func (m *managerImpl) WatchDeclarativeConfigDir() {
+	m.once.Do(func() {
+		wh := &watchHandler{m: m}
+		// Set Force to true, so we explicitly retry watching the files within the directory and not stop on the first
+		// error occurred.
+		watchOpts := k8scfgwatch.Options{Interval: watchInterval, Force: true}
+		_ = k8scfgwatch.WatchConfigMountDir(context.Background(), declarativeConfigDir,
+			k8scfgwatch.DeduplicateWatchErrors(wh), watchOpts)
+	})
+}
--- a/central/declarativeconfig/manager_impl.go
+++ b/central/declarativeconfig/manager_impl.go
@@ -23,7 +23,7 @@ type managerImpl struct {
 }
 
 // New creates a new instance of Manager.
-// Note that it will not watch the declarative configuration directories when creaeted, only after
+// Note that it will not watch the declarative configuration directories when created, only after
 // WatchDeclarativeConfigDir has been called.
 func New() Manager {
 	return &managerImpl{}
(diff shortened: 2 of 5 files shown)


dhaus67:
* master
  * **PR #4530** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4530" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4530?utm_source=stack-comment).

central/declarativeconfig/manager.go
```diff
@@ -0,0 +1,6 @@
+package declarativeconfig
+
+// Manager manages reconciling declarative configurations.
```
ivan-degtiarenko, line 3:
```suggestion
// Manager manages reconciling declarative configuration.
```

central/declarativeconfig/manager_impl.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+const (
+	watchInterval        = 5 * time.Second
+	declarativeConfigDir = "/run/stackrox.io/declarative-configuration"
+)
+
+var (
+	log = logging.LoggerForModule()
```
ivan-degtiarenko, line 18: `log` is unused

central/declarativeconfig/manager_impl.go
```diff
@@ -0,0 +1,41 @@
...
+	return &managerImpl{}
+}
+
+func (m *managerImpl) WatchDeclarativeConfigDir() {
+	m.once.Do(func() {
+		wh := &watchHandler{m: m}
+		// Set Force to true, so we explicitly retry watching the files within the directory and not stop on the first
+		// error occurred.
+		watchOpts := k8scfgwatch.Options{Interval: watchInterval, Force: true}
+		_ = k8scfgwatch.WatchConfigMountDir(context.Background(), declarativeConfigDir,
```
ivan-degtiarenko, line 38: Shouldn't we at least log this error?

central/declarativeconfig/manager_impl.go
```diff
@@ -0,0 +1,41 @@
...
+var (
+	log = logging.LoggerForModule()
+)
+
+type managerImpl struct {
+	once sync.Once
+}
+
+// New creates a new instance of Manager.
+// Note that it will not watch the declarative configuration directories when creaeted, only after
```
ivan-degtiarenko, line 26:
```suggestion
// Note that it will not watch the declarative configuration directories when created, only after
```

central/declarativeconfig/manager_impl.go
dhaus67: I've used it now within the `OnWatchError`

central/declarativeconfig/manager_impl.go
dhaus67: When `force == true`, no error will be returned by [WatchConfigMountDir](https://github.com/stackrox/stackrox/blob/2bd26143bfbe8de03bd59e9f3bb4a00b264383c0/pkg/k8scfgwatch/watch.go#L36-L40). Instead, the error will be passed to the watch handler's `OnWatchError`. I've added a log there.

central/declarativeconfig/watch_handler.go
```diff
@@ -0,0 +1,18 @@
...
+
+func (w *watchHandler) OnChange(dir string) (interface{}, error) {
+	return nil, nil
+}
+
+func (w *watchHandler) OnStableUpdate(val interface{}, err error) {
+}
+
+func (w *watchHandler) OnWatchError(err error) {
+	log.Errorf("Error watching declarative configuration directory: %+v", err)
```
ivan-degtiarenko, line 17: nit: maybe we can use `log` declaration to the same file as it is used?

central/main.go
```diff
@@ -486,6 +487,10 @@ func startGRPCServer() {
 
 	basicAuthProvider := userpass.RegisterAuthProviderOrPanic(authProviderRegisteringCtx, basicAuthMgr, registry)
 
+	if features.DeclarativeConfiguration.Enabled() {
+		declarativeconfig.Singleton().WatchDeclarativeConfigDir()
```
ivan-degtiarenko, line 491: I don't think I completely get why we need to split the `Singleton` call from `WatchDeclarativeConfigDir()`. Perhaps I miss the bigger picture and there will be more functions added to `declarativeconfig.Manager` interface?

central/declarativeconfig/singleton.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "github.com/stackrox/rox/pkg/sync"
+
+var (
+	once     sync.Once
+	instance Manager
+)
+
+// Singleton provides the instance of Manager to use.
+func Singleton() Manager {
```
ivan-degtiarenko, line 11: Let's rename it to `ManagerSingleton`, otherwise we get usage like `declarativeconfig.Singleton()` which is quite confusing

central/main.go
dhaus67: I'll be honest, I cannot answer yet in what shape / way it will evolve, I suppose we will see this soon.

While I can move the watch start to the `Singleton()`, without an additional function it would look a bit weird `_ = declarativeconfig.Singleton()`, so let's stick with it, see how the interface evolves, and if we don't see the need for anything additional later on revisit this?

ivan-degtiarenko: Sounds good.

ivan-degtiarenko (approved):
🚢 it

PR #4532, opened 2023-01-25
WIP: Create JIRA tasks on CI exit trap

janisz:
## Description

Currently we are creating JIRA tasks when copying reports to artifacts dir. This causes that some steps will not be caught. This PR changes JIRA tasks creation to happen at the end of job so it should handle all failures (except prow failures).


janisz:
OK, this is not the optimal solution as for every failure we will create coresponding step failure task.

PR #4540, opened 2023-01-25
ROX-14495: Fix failing admission controller test.

clickboo:
## Description

- Added code to turn on ROX_NEW_POLICY_CATEGORIES ff when postgres env is turned on.
- Added a UUID to the newPolicy name, to avoid, a duplicate policy name when being added (local runs failed with an error in post policy due to policy name collision)

## Checklist

## Testing Performed
Fixed test hence automated CI run only

### DIFF
--- a/tests/admctrl_configmap_test.go
+++ b/tests/admctrl_configmap_test.go
@@ -10,17 +10,23 @@ import (
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/admissioncontrol"
 	"github.com/stackrox/rox/pkg/booleanpolicy/fieldnames"
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/features"
 	"github.com/stackrox/rox/pkg/gziputil"
 	"github.com/stackrox/rox/pkg/namespaces"
 	"github.com/stackrox/rox/pkg/testutils"
 	"github.com/stackrox/rox/pkg/testutils/centralgrpc"
+	"github.com/stackrox/rox/pkg/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestAdmissionControllerConfigMap(t *testing.T) {
-	t.Skip("ROX-14495: This test is not working.")
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		t.Setenv(features.NewPolicyCategories.Name(), "true")
+	}
+
 	k8sClient := createK8sClient(t)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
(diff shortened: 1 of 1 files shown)


tests/admctrl_configmap_test.go
```diff
@@ -20,6 +21,8 @@ import (
 )
 
 func TestAdmissionControllerConfigMap(t *testing.T) {
+	t.Setenv(features.NewPolicyCategories.Name(), "true")
```
md2119, line 24: Did you add this test through your mammoth feature PR? If not, why cannot we run this test anymore with feature flag off?

theencee: Wait both the postgres and non-postgres will run the same file. So does that mean the one that was working fine will now get the flag and thus might cause other issues?

clickboo: Let's wait for CI, but I don't think so.

md2119: Is the current fix enough on release builds e.g. nightly?

clickboo: Is there a way to find out until a nightly runs with it? Or is there a tag to run the nightlies on this PR?

Also looks like some infra blip causing the e2e tests to fail

clickboo: Alright, CI has other plans. So the wait and watch is longer. 🤷‍♀️

tests/admctrl_configmap_test.go
```diff
@@ -51,10 +57,10 @@ func TestAdmissionControllerConfigMap(t *testing.T) {
 
 	policyServiceClient := v1.NewPolicyServiceClient(cc)
 	newPolicy := &storage.Policy{
-		Name:        "testpolicy_" + t.Name(),
+		Name:        "testpolicy_" + t.Name() + "_" + uuid.NewV4().String(),
 		Description: "test deploy time policy",
 		Rationale:   "test deploy time policy",
-		Categories:  []string{"test"},
+		Categories:  []string{"Test"},
```
theencee, line 63: Why does this make a difference? (or just a holdover from testing?)

tests/admctrl_configmap_test.go
```diff
@@ -10,17 +10,23 @@ import (
...
+	"github.com/stackrox/rox/pkg/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestAdmissionControllerConfigMap(t *testing.T) {
-	t.Skip("ROX-14495: This test is not working.")
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		t.Setenv(features.NewPolicyCategories.Name(), "true")
```
theencee, line 27: so basically if this is a test for postgres, then this test will only run the policy category path. But is that accurate? Should it run whether or not policy categories are enabled?

clickboo: In postgres, new policy categories only, so yes. Am I missing anything?

clickboo: ( we plan to remove the new categories feature flag soon and keep postgres only - then it really won't matter)

clickboo:
@charmik-redhat @md2119 
image

Seeing this failure, randomly at best.

PR #4542, opened 2023-01-25
ROX-14137: Permission Set declarative config Go struct

ivan-degtiarenko:
## Description

This solution is based on re-using existing proto enums within new Go structs.

Alternatives:
1. Introduce a new struct replacing `storage.Access`proto enum within this Go struct + conversion function
2. Convert from YAML to JSON first, then to proto

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/declarativeconfig/permission_set.go
+++ b/pkg/declarativeconfig/permission_set.go
@@ -1,0 +1,33 @@
+package declarativeconfig
+
+import (
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"gopkg.in/yaml.v3"
+)
+
+type PermissionSet struct {
+	Name             string              `yaml:"name,omitempty"`
+	Description      string              ` yaml:"description,omitempty"`
+	ResourceToAccess map[string]DCAccess `yaml:"resource_to_access,omitempty"`
+}
+
+type DCAccess storage.Access
+
+func (a DCAccess) MarshalYAML() ([]byte, error) {
+	protoAccess := storage.Access(a)
+	return []byte(protoAccess.String()), nil
+}
+
+func (a *DCAccess) UnmarshalYAML(value *yaml.Node) error {
+	var v string
+	if err := value.Decode(&v); err != nil {
+		return err
+	}
+	i, ok := storage.Access_value[v]
+	if !ok {
+		return errors.New("not found")
+	}
+	*a = DCAccess(i)
+	return nil
+}
--- a/pkg/declarativeconfig/permission_set.go
+++ b/pkg/declarativeconfig/permission_set.go
@@ -6,24 +6,29 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/declarativeconfig/permission_set.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"gopkg.in/yaml.v3"
+)
+
+type PermissionSet struct {
+	Name             string              `yaml:"name,omitempty"`
+	Description      string              ` yaml:"description,omitempty"`
+	ResourceToAccess map[string]DCAccess `yaml:"resource_to_access,omitempty"`
```
dhaus67, line 12: This is more of a style thing rather than anything else, but:
Currently, if you would specify those resource, you would do:
```yaml
resource_to_access:
  Cluster: READ_ACCESS
  Namespace: READ_ACCESS
```

Making it more like look & feel compared to K8S resources, and looking forward if we were to do this within CRs, we would probably use an array instead of a map, as such:
```
resource_to_access:
- resource: Cluster
  access: READ_ACCESS
```
I personally find the latter one better in line with existing K8S resources + our own CR compared to a `map[string]string`.

dhaus67, line 12: I don't think we should start using `_` in our YAMLs.

I think we can re-use some of the principles / guidelines we have for [operator CR and their naming conventions](https://github.com/stackrox/stackrox/blob/master/operator/EXTENDING_CRDS.md#naming), and would propose to use camel case instead (It also feels awkward writing a YAML with underscores, but that's more of a personal opinion :))

Additionally, we are not bound to re-use proto names or others within our YAMLs, so maybe we want to switch away from `resourceToAccess`, to something more simple, such as `resources`?

pkg/declarativeconfig/permission_set.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"gopkg.in/yaml.v3"
+)
+
+type PermissionSet struct {
+	Name             string              `yaml:"name,omitempty"`
+	Description      string              ` yaml:"description,omitempty"`
```
dhaus67, line 11:
```suggestion
	Description      string              `yaml:"description,omitempty"`
```

pkg/declarativeconfig/permission_set.go
```diff
@@ -0,0 +1,33 @@
...
+	ResourceToAccess map[string]DCAccess `yaml:"resource_to_access,omitempty"`
+}
+
+type DCAccess storage.Access
```
dhaus67, line 15: I think in the scope of this file, you can omit the prefix of `DC` and simply put `Access` here?

dhaus67 (changes requested):

PR #4553, opened 2023-01-26
ROX-14137: access scope YAML format for declarative config

ivan-degtiarenko:
## Description

Go struct which will be used for YAML -> Go struct transformation of declarative Access Scope.

## Checklist

## Testing Performed

Unit test added

### DIFF
--- a/pkg/declarativeconfig/access_scope.go
+++ b/pkg/declarativeconfig/access_scope.go
@@ -19,13 +19,13 @@ type Namespace struct {
 
 type Operator storage.SetBasedLabelSelector_Operator
 
-// MarshalYAML transforms Access to YAML format.
+// MarshalYAML transforms Operator to YAML format.
 func (a Operator) MarshalYAML() ([]byte, error) {
 	protoAccess := storage.SetBasedLabelSelector_Operator(a)
 	return []byte(protoAccess.String()), nil
 }
 
-// UnmarshalYAML makes transformation from YAML to Access.
+// UnmarshalYAML makes transformation from YAML to Operator.
 func (a *Operator) UnmarshalYAML(value *yaml.Node) error {
 	var v string
 	if err := value.Decode(&v); err != nil {
--- a/pkg/declarativeconfig/access_scope.go
+++ b/pkg/declarativeconfig/access_scope.go
@@ -6,17 +6,14 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// AccessScope is representation of storage.AccessScope that supports transformation from YAML.
 type AccessScope struct {
 	Name        string `yaml:"name,omitempty"`
 	Description string `yaml:"description,omitempty"`
 	Rules       Rules  `yaml:"rules,omitempty"`
 }
 
-type Namespace struct {
-	Cluster   string `yaml:"cluster,omitempty"`
-	Namespace string `yaml:"namespace,omitempty"`
-}
-
+// Operator is representation of storage.SetBasedLabelSelector_Operator that supports transformation from YAML.
 type Operator storage.SetBasedLabelSelector_Operator
 
 // MarshalYAML transforms Operator to YAML format.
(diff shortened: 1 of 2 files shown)


pkg/declarativeconfig/access_scope.go
```diff
@@ -0,0 +1,57 @@
...
+	Values   []string `yaml:"values,omitempty"`
+}
+
+type LabelSelector struct {
+	Requirements []Requirement `yaml:"requirements,omitempty"`
+}
+
+type Rules struct {
+	IncludedClusters        []string        `yaml:"includedClusters,omitempty"`
+	IncludedNamespaces      []Namespace     `yaml:"includedNamespaces,omitempty"`
```
dhaus67, line 54: It's a bit unfortunate now that I have to do the following for multiple namespaces from the same cluster:
```yaml
includedClusters:
- cluster A
includedNamespaces:
- cluster: cluster A
  namespace: namespace 1
- cluster: cluster A
  namespace: namespace 2
- cluster: cluster A
  namespace: namespace 2
```

This seems a bit suboptimal. I feel like we could do two things here: either allow to specify multiple namespaces per cluster, avoiding us to specify `cluster` multiple times _or_ we go one step further and merge both `includedClusters/includedNamespaces` together to something like:
```yaml
included:
- cluster: cluster A
  namespace:
  - ns 1
  - ns 2
  - ns 3
```
Then we have just a single setting that holds the included cluster / namespaces (with the usual caveat, that namespaces can be empty for each cluster)

pkg/declarativeconfig/access_scope.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type Namespace struct {
+	Cluster   string `yaml:"cluster,omitempty"`
+	Namespace string `yaml:"namespace,omitempty"`
+}
+
+type Operator storage.SetBasedLabelSelector_Operator
+
+// MarshalYAML transforms Access to YAML format.
```
dhaus67, line 22:
```suggestion
// MarshalYAML transforms Operator to YAML format.
```

pkg/declarativeconfig/access_scope.go
```diff
@@ -0,0 +1,57 @@
...
+func (a Operator) MarshalYAML() ([]byte, error) {
+	protoAccess := storage.SetBasedLabelSelector_Operator(a)
+	return []byte(protoAccess.String()), nil
+}
+
+// UnmarshalYAML makes transformation from YAML to Access.
```
dhaus67, line 28:
```suggestion
// UnmarshalYAML makes transformation from YAML to Operator.
```

dhaus67 (changes requested):

pkg/declarativeconfig/access_scope.go
ivan-degtiarenko: The second option sounds reasonable. It's hard for me to imagine a case where we would want explicitly specify cluster and 0 namespaces within it, so we can assume `cluster + 0 namespaces = the whole cluster`.

pkg/declarativeconfig/access_scope.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"gopkg.in/yaml.v3"
+)
+
+// AccessScope is representation of storage.AccessScope that supports transformation from YAML.
+type AccessScope struct {
+	Name        string `yaml:"name,omitempty"`
+	Description string `yaml:"description,omitempty"`
```
dhaus67, line 12: Not for this PR, more or less a generic question:
I wonder if it makes sense, seeing as this could grow in amount of resources and almost all resources seem to have `name & description` to create a separete `Metadata` struct and embed this within here, i.e.:
```go
type Metadata struct {
	Name        string `yaml:"name,omitempty"`
	Description string `yaml:"description,omitempty"`
}

type AccessScope {
          Metadata
          Rules Rules `yaml:"rules,omitempty"`
}
```

dhaus67 (approved):

pkg/declarativeconfig/access_scope.go
ivan-degtiarenko: Yep, I will create a follow-up JIRA

PR #4555, opened 2023-01-26
Add fallback UI for VM Report configs that could not be migrated to collections

dvail:
## Description

In the migration from access scopes to collections as the method of defining a report scope, it was determined that there are some access scopes that cannot be converted to a collection. When this occurs, the `scopeId` attached to a report will continue to reference an existing access scope, and the database column will contain an ID that could reference two separate entities.

This is done so that we can detect these failed migrations and present a warning to the user that the report will not run without a new configuration.

The method to do this in the UI is the following:

- Upon the initial load of the report config, immediately send a second request to get the collection for `report.scopeId`
- If this returns a collection, we render the form and details page as normal.
- If the request for a collection fails, we make a third request to get the access scope for `report.scopeId`.
- If this third request returns an access scope, we add messages to the UI letting the user know that they need to configure a new collection for this report.
- If the access scope request fails, or returns a 404, we defer to the current error handling and _do not_ display the migration error message.

## Checklist

## Testing Performed

With `ROX_OBJECT_COLLECTIONS` set to `false`:
Visit a report config detail page for a report using a legacy access scope id:
image

Visit the edit page for this report:
image

With `ROX_OBJECT_COLLECTIONS` set to `true`:

Visit a report config detail page for a report using a legacy access scope id:
image

Clicking the link in the alert takes the user to the edit form. The edit form will display an error state and a brief message that the user must configure the collection.
image
If the user doesn't have write access, the edit link is not displayed:
image

Setting a new value and saving the report config successfully updates the report:
image

In the case where a collection request for `scopeId` fails, and the access scope request succeeds without finding a match, the UI will remain as it is today:
image
image

The UI will remain as it is today if there is an error during the report fetch:
image


dvail:
I don't think the last two error cases in the testing section above are ideal, but I opted to not change them because:

- The release code freeze is looming
- We are replacing this section soon as part of VM 2.0 anyway

ui/apps/platform/src/Containers/Collections/CollectionForm.css
```diff
@@ -25,4 +25,4 @@
     growing too large. This likely can be removed when we fully migrate to PF.
   */
   max-height: 2rem;
-}
\ No newline at end of file
+}
```
dvail, line 28: Piggyback style change from another PR

ui/apps/platform/src/Containers/VulnMgmt/Reports/Components/ScopeName.tsx
```diff
@@ -1,64 +1,63 @@
 import React, { ReactElement } from 'react';
```
dvail, line 1: The main structural change in this PR is that the report scope details request was moved from this child component, that was solely used to display the scope name, to being immediately after the request that fetches the main scope details.

This allows us to detect a problem early and display the appropriate error messages outside the `ScopeName` component and then pass that data down to child components without requiring an additional request.

ui/apps/platform/src/hooks/useFetchReport.ts
```diff
@@ -1,31 +1,104 @@
...
+    id: string;
+    name: string;
+    description: string;
+};
 
-const defaultResultState = { report: null, error: null, isLoading: true };
+type Result = {
+    isLoading: boolean;
+    report: ReportConfiguration | null;
+    reportScope: ReportScope | null;
```
dvail, line 20: `reportScope` was added to the `Result` type here, instead of making it a child property of `Result['report']`, since the scope of the `ReportConfiguration` type is much larger and this limits the scope of the change.

ui/apps/platform/src/Containers/VulnMgmt/Reports/Components/ScopeName.tsx
```diff
@@ -1,64 +1,63 @@
...
-
-    if (error) {
-        return <span>Error getting scope info. {getAxiosErrorMessage(error)}</span>;
+    if (reportScope.type === 'AccessControlScope') {
+        return (
+            <Alert
+                isInline
+                variant="danger"
+                title="The report scope for this configuration could not be migrated to a collection"
+                actionLinks={
```
pedrottimark, line 42: Does action link depend on READ_WRITE_ACCESS for WorkflowAdministration resource?

This component is for detail view which requires only READ_ACCESS before this change, correct?

ui/apps/platform/src/hooks/useFetchReport.ts
pedrottimark: Delightful pun to reward the reviewer:

> scope of the change

ui/apps/platform/src/hooks/useFetchReport.ts
```diff
@@ -1,31 +1,104 @@
...
+    return request.then(({ collection: { id, name, description } }) => ({
+        type: 'CollectionScope',
+        id,
+        name,
+        description,
+    }));
+}
+
+function fetchAccessScopeReportScope(scopeId: string): Promise<ReportScope | null> {
+    return fetchAccessScopes().then((scopes) => {
```
pedrottimark, line 43: Oh, good. Plural fetch and find prevents 404 status for a singular fetch!

dvail: To be honest I was looking for a singular fetch and couldn't seem to find one for access scopes. I was going to let the promise reject normally and fall through to the error case.

ui/apps/platform/src/Containers/VulnMgmt/Reports/Components/ScopeName.tsx
dvail: Fixed, added a screenshot as well.

ui/apps/platform/src/Containers/VulnMgmt/Reports/Components/ScopeName.tsx
```diff
@@ -1,64 +1,68 @@
...
-import useFetchScopes from 'hooks/useFetchScopes';
-import useCollection from 'Containers/Collections/hooks/useCollection';
-import { getAxiosErrorMessage } from 'utils/responseErrorUtils';
 import { collectionsBasePath } from 'routePaths';
+import { ReportScope } from 'hooks/useFetchReport';
 
 type ScopeNameProps = {
-    scopeId: string;
+    reportScope: ReportScope | null;
+    // TODO This isn't a prop-type, and it -is- used. Not sure why the lint error needs to be suppressed
```
dvail, line 13: So it looks like this throws a lint error since the prop isn't used in _every_ component that takes this type as a parameter.

Possibly an eslint bug? If not its a questionable rule.

This code will go away next release, so I'm not inclined to investigate to far into it.

pedrottimark (approved):

PR #4558, opened 2023-01-26
Add Group By support to v1.Query

md2119:
## Description

This PR adds support to group query results by fields other than primary keys. For example, 
`select cve, max(severity) from image_cves group by cve` (`cve` is not a primary key, `id` is). This PR is a limited supported with an intention to lay down the foundation for further work. Highly recommend referring to test cases for example queries.

**Note:**

This is the first enhancement to our search framework. In the follow-ups I plan to add:
- Group by support with pagination
- Support to select on derived fields. Currently, we can only retrieve proto field values for search type queries.
- Support to retrieve search results without ID. Currently, search results from our search framework always return results keyed by ID (`search.Result`). Therefore, we currently cannot work with queries like `select count(operatingsystem), max(severity) from image_cves group by severity`

_________

**Update (01/30):**

Reviewers, you may need to expand some files e.g. `pkg/search/postgres/group_by_test.go`.

This PR does the following:
- Updates the `v1.Query` proto to accept group by clause. If provided, it controls how the records should be grouped.
- When grouping by columns other primary keys, selected columns that are not part of group by must be aggregated. This is an SQL constraint. For example, the SQL engine needs to know whether you want to view all results per group, do you want to max/highest/lowest per group etc.
- Currently, all search results coming out of datastore (searcher) are keyed by ID.
```
type Result struct {
	ID      string
	Matches map[string][]string
	Score   float64
	Fields  map[string]interface{}
}
```
Therefore, to match the expectation from the datastore, we unnest the aggregated results. 
- We expect `ID` not aggregated IDs
-  We expect proto fields according to their type and not in their aggregated forms.

3rd bullet in Notes highlights the support to work with any json/struct. 
_________

## Checklist

## Testing Performed
Unit + CI


pkg/search/postgres/common.go
```diff
@@ -169,19 +184,31 @@ func (q *query) getPortionBeforeFromClause() string {
...
+			if field.AggrAndUnnest() {
+				// Alias don't work in SQL `row()` function.
+				field.Alias = ""
+				aggrSelects = append(aggrSelects, field.PathForSelectPortion())
+			} else {
+				nonAggrSelects = append(nonAggrSelects, field.PathForSelectPortion())
+			}
 		}
-		return "select " + stringutils.JoinNonEmpty(", ", selectPaths...)
+		return "select jsonb_agg(row(" + stringutils.JoinNonEmpty(", ", aggrSelects...) + ")), " +
```
erthalion, line 210: `row` result will be converted to `jsonb` using `to_json` function, resulting in field names like "f1","f2", etc. Maybe it would be beneficial to construct the jsonb object directly? In this way one could pass column names for keys, and in the end:

```
select jsonb_build_object('id', 1, 'name', 'test');
    jsonb_build_object
---------------------------
 {"id": 1, "name": "test"}
```

vs

```
select to_json(row(1, 'test'));
       to_json
----------------------
 {"f1":1,"f2":"test"}
```

pkg/search/postgres/common.go
```diff
@@ -268,31 +295,64 @@ func (p *parsedPaginationQuery) AsSQL() string {
...
+			querySoFar.GroupByPrimaryKey = true
 		}
-		switch field.derivedMetadata.DerivationType {
-		case searchPkg.CountDerivationType:
-			applyGroupByPrimaryKeys(querySoFar, schema)
-			return
+
+		var cast string
+		if dbField.SQLType == "uuid" {
+			cast = "::text"
```
erthalion, line 331: Could you pls elaborate, why this cast is needed?

pkg/search/postgres/common_test.go
```diff
@@ -211,3 +212,68 @@ func TestMultiTableQueries(t *testing.T) {
...
+		q             *v1.Query
+		schema        *walker.Schema
+		expectedQuery string
+		expectedError bool
+	}{
+		{
+			desc:          "group by primary key; having w/ derived field",
+			q:             search.NewQueryBuilder().AddStrings(search.TestParent1ValCount, ">1").ProtoQuery(),
+			schema:        schema.TestParent1Schema,
+			expectedQuery: "select test_parent1.Id test_parent1_Id from test_parent1 group by test_parent1.Id having count(test_parent1.Val) > $1",
```
erthalion, line 230: Maybe I'm missing something, but what's the point of generating a query with group by the primary key?

erthalion (commented):
Thanks for the PR.

I have to admit, I could me missing some context, but it would be beneficial to
clarify how exactly grouping supposed to be working at high level. More
precisely it's not clear why nesting/unnesting is needed. Using `jsonb_agg`
would be convenient for aggregating multiple columns, but then it could be
returned to the client in the jsonb format, without unnesting. Few more
comments below.

pkg/search/postgres/common.go
md2119: Yes, I looked at that option as well. I didn't want the query string to be too long hence went with the current approach. Additionally, we do not really use the select path names (except for debugging). 

I will try it but might do it as a polish task in one of the follow-ups.

pkg/search/postgres/common.go
md2119: Because we do not have uuid in go.

pkg/search/postgres/common_test.go
md2119: That is a fair argument.

md2119: Actually, an explicit group by is required because of having clause.

md2119:
> Thanks for the PR.
> 
> I have to admit, I could me missing some context, but it would be beneficial to clarify how exactly grouping supposed to be working at high level. More precisely it's not clear why nesting/unnesting is needed. Using `jsonb_agg` would be convenient for aggregating multiple columns, but then it could be returned to the client in the jsonb format, without unnesting. Few more comments below.

Will add the explanation today.

pkg/search/postgres/common_test.go
erthalion: It's other way around I believe, `having` is operating on groups created by `group by`, so no grouping -- no need having. Otherwise since the group key is unique, count will always be of value 1, right?

https://www.postgresql.org/docs/15/sql-select.html#SQL-HAVING

md2119: The patten of generation is based on the type of query field-derived field or not derived. When we require a join, we need to group by primary key of the correct table. I believe this pattern was done to keep the parser simple enough and not have many if-elses.

Changing the will require a bigger refactor which I will keep outside the scope of this PR. This PR only focuses on non-primary key group bys. I have noted your point, which we can get back to in future. I compared the cost of seq scan vs aggr, there is a bit of extra cost but it is still within an acceptable margin.

md2119:
Closing this in favor of #4688

PR #4567, opened 2023-01-27
ROX-14352: Run dockerized builds as current user, rather than root, to please git.

porridge:
## Description

Recent `git` does not like the repository to be owned by a different user.

To solve this
- pass --user to `docker run`
- make sure the go cache volume source directory is created as the current user before running docker

I think I even managed to find a simple way to fix permissions of existing/newly created volumes. This is especially important on macs, where the volume does not map to an actual directory on disk and cannot be solved by an `rm -rf`.

## Checklist

## Testing Performed

Ran `make clean; rmdir linux-gocache; make sensor-build-dockerized; make image` successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -367,6 +367,7 @@ bin/$(HOST_OS)_$(GOARCH)/admission-control: build-prep
 
 .PHONY: build-volumes
 build-volumes:
+	$(SILENT)mkdir -p $(CURDIR)/linux-gocache
 	$(SILENT)docker volume inspect $(GOPATH_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOPATH_VOLUME_NAME)
 	$(SILENT)docker volume inspect $(GOCACHE_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOCACHE_VOLUME_NAME)
 
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ central-build-nodeps:
 .PHONY: fast-central
 fast-central: deps
 	@echo "+ $@"
-	docker run --rm $(GOPATH_WD_OVERRIDES) $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) make fast-central-build
+	docker run --user "$(shell id -u)" --rm $(GOPATH_WD_OVERRIDES) $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) make fast-central-build
 	$(SILENT)$(BASE_DIR)/scripts/k8s/kill-pod.sh central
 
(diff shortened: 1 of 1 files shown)


porridge:
The postgres e2e test failure seems unrelated to this change:
```
panic: ERROR: duplicate key value violates unique constraint "policy_categories_name_key" (SQLSTATE 23505)
```

Maddosaurus:
Interestingly, the `linux-gocache` is a docker volume instead of a local folder for me, so I tried deleting the volume with `docker volume rm stackrox-rox-gocache`, which leads to permission errors when the sensor build is trying to create folders

Test Environment: MacOS 12.6 (x86) with Colima


```
⇒  docker volume rm stackrox-rox-gocache
stackrox-rox-gocache

 ⇒  make sensor-build-dockerized
stackrox-rox-gocache
+ main-builder-image
# If the next line fails, you need to update the go version in rox-ci-image/images/stackrox-build.Dockerfile
grep -q "go1.18.4" <(docker run --rm "quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.55" go version)
+ sensor-build-dockerized
docker run --user "501" --rm -e CI -e CIRCLE_TAG -e GOTAGS -e DEBUG_BUILD -w /src -e GOPATH=/go -e GOCACHE=/linux-gocache -v/Users/matthias/go/src/github.com/stackrox/stackrox:/src:delegated -v stackrox-rox-gocache:/linux-gocache:delegated -v stackrox-rox-gopath:/go:delegated quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.55 make sensor-build
/src/scripts/go-build.sh sensor/kubernetes sensor/admission-control
failed to initialize build cache at /linux-gocache: mkdir /linux-gocache/00: permission denied
make: *** [Makefile:396: sensor-build] Error 1
make: *** [sensor-build-dockerized] Error 2
```

This is what the container sees if executed directly:

```
⇒  docker run (...) ls -ld .
drwxr-xr-x 1 501 games 1920 Jan 26 17:01 .

⇒  docker run (...) id
uid=0(root) gid=0(root) groups=0(root)
```

Full image call: `docker run -i --rm -w /src -v/Users/matthias/go/src/github.com/stackrox/stackrox:/src:delegated -v stackrox-rox-gocache:/linux-gocache:delegated -v stackrox-rox-gopath:/go:delegated [quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.55](http://quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.55)`

Maddosaurus (approved):
Verified working by deleting both volumes, the GOPATH and the GOCACHE volume and then starting a build.

PR #4571, opened 2023-01-27
Fix passing RHCOS content sets

vikin91:
## Description

We apparently missed the Content Sets when converting between Scanner and Stackrox protos what resulted in always empty content sets and 0 vulnerabilities when scanning.
This PR should fix that and provide a bit more visibility on the matter in the logs.

## Checklist

### N/A

## Testing Performed

- [ ] CI
- [x] OpenShift deployment and checking the logs whether there has been any content sets discovered

```
nodeinventory.go:39: Info: Node inventory has been built with 501 packages and 4 content sets
```

### DIFF
--- a/compliance/collection/nodeinventorizer/nodeinventory.go
+++ b/compliance/collection/nodeinventorizer/nodeinventory.go
@@ -29,13 +29,15 @@ func (n *NodeInventoryCollector) Scan(nodeName string) (*storage.NodeInventory, 
 		log.Errorf("Error scanning node /host inventory: %v", err)
 		return nil, err
 	}
-	log.Info("Finished node inventory")
 	log.Debugf("Components found under /host: %v", componentsHost)
 
 	protoComponents := protoComponentsFromScanComponents(componentsHost)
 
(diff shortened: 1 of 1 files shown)


jschnath (approved):
LGTM

Maddosaurus (approved):
Generally, LGTM!
We could discuss whether it would make sense to `nil` check `c.CertifiedRHELComponents.ContentSets` before accessing it, but besides that, good to go!

vikin91:
[Here](https://github.com/stackrox/stackrox/pull/4492/files#diff-ac541a541bd48a25834817cea7480d43de2d57c6dab8fe1fdf4768953173e2daR36) is the place where we access the `ContentSets` in Central to build a request to Scanner (not merged yet).
 
We do it through the getter so we are safe. Later, this will be accessed in Scanner [here](https://github.com/stackrox/scanner/blob/master/api/v1/nodescan/service.go#L291) and we also do it through a getter.

Ergo: even if `c.CertifiedRHELComponents.ContentSets == nil` we are ready to handle it.

PR #4573, opened 2023-01-27
[ROX-14220] : DeleteCollection service should error out if collection is used by a reportConfiguration

charmik-redhat:
## Description

Same as title

## Checklist

## Testing Performed

Unit tests added

### DIFF
--- a/central/resourcecollection/service/service_impl.go
+++ b/central/resourcecollection/service/service_impl.go
@@ -8,6 +8,7 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
 	deploymentDS "github.com/stackrox/rox/central/deployment/datastore"
+	reportConfigDS "github.com/stackrox/rox/central/reportconfigurations/datastore"
 	"github.com/stackrox/rox/central/resourcecollection/datastore"
 	"github.com/stackrox/rox/central/role/resources"
 	v1 "github.com/stackrox/rox/generated/api/v1"
@@ -67,9 +68,10 @@ type collectionRequest interface {
 type serviceImpl struct {
 	v1.UnimplementedCollectionServiceServer
 
-	datastore     datastore.DataStore
-	queryResolver datastore.QueryResolver
-	deploymentDS  deploymentDS.DataStore
+	datastore             datastore.DataStore
+	queryResolver         datastore.QueryResolver
+	deploymentDS          deploymentDS.DataStore
+	reportConfigDatastore reportConfigDS.DataStore
 }
 
 // RegisterServiceServer registers this service with the given gRPC Server.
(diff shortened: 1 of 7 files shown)


central/resourcecollection/service/service_impl.go
```diff
@@ -155,6 +157,17 @@ func (s *serviceImpl) DeleteCollection(ctx context.Context, request *v1.Resource
...
 	}
+	// error out if collection is in use by a report config
+	err := s.reportConfigDatastore.Walk(ctx, func(reportConfig *storage.ReportConfiguration) error {
+		if reportConfig.GetScopeId() == request.GetId() {
+			return errors.New("Collection is in use in one or more report configurations")
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to delete collection")
```
md2119, line 168: add collection ID to the message

central/resourcecollection/service/service_impl_test.go
```diff
@@ -28,33 +33,49 @@ type CollectionServiceTestSuite struct {
...
 	testutils.SetExampleVersion(suite.T())
 }
 
 func (suite *CollectionServiceTestSuite) TearDownSuite() {
+	suite.testDB.Teardown(suite.T())
 	suite.mockCtrl.Finish()
 }
 
 func (suite *CollectionServiceTestSuite) TestListCollectionSelectors() {
-	if !features.ObjectCollections.Enabled() {
```
md2119, line 54: thanks for cleaning it up!

md2119 (approved):
One minor comment otherwise LGTM

central/resourcecollection/service/service_impl.go
```diff
@@ -155,6 +157,17 @@ func (s *serviceImpl) DeleteCollection(ctx context.Context, request *v1.Resource
 	if request.GetId() == "" {
 		return nil, errors.Wrap(errox.InvalidArgs, "Non empty collection id must be specified to delete a collection")
 	}
+	// error out if collection is in use by a report config
+	err := s.reportConfigDatastore.Walk(ctx, func(reportConfig *storage.ReportConfiguration) error {
+		if reportConfig.GetScopeId() == request.GetId() {
+			return errors.New("Collection is in use in one or more report configurations")
+		}
+		return nil
```
keyallis, line 165: Why don't we just add a search tag to the field in the proto and query to see if any results come up when we search for report configurations with a particular id relating to the collection in question?

keyallis: The other thing here is once we migrate to collections and postgres by default, we could define the scopeid as a foreign key in the proto and then the postgres datastore will handle all of the checks for us

charmik-redhat: Thought about adding a search tag to the report_configuratons proto, but it would change the schema. So @md2119 suggested that it would be better to just walk through the report configs and check for now. It shouldn't take too long either since there won't be too many reportConfigs in the db anyways.

charmik-redhat: We can remove the check once scope_id becomes a foreign key.

md2119: Yes, we have frozen the postgres schema for 3.74. Adding fk to report configuration will change its schema. Although there is a path through the weeds to make it happen, we need to keep the risk minimum for a blotch-free postgres release.

keyallis: Understood, thanks for the clarity

PR #4574, opened 2023-01-27
ROX-14597: Operator bundle, drop version map when replacing rbac-proxy

kurlov:
## Description

The `kube-rbac-proxy` image is replaced when [building a downstream operator image](https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-operator-bundle/render_templates#L73). it is redundantly to specify tag version because only one version is replaced. Thus python script is adjusted.

Additional small changes:

- PEP8 improvements
- Bump pip and PyYAML
- add logs to see what is replaced

**Note**: the version mapping has to be dropped from [downstream building instruction](https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-operator-bundle/render_templates#L73) once this PR is merged.

## Checklist

## Testing Performed
use `bundle-post-process` command locally with following option:
```
./bundle_helpers/patch-csv.py \
		--use-version $(VERSION) \
		--first-version 3.62.0 `# 3.62.0 is the first operator version ever released` \
		--operator-image $(IMG) \
		--no-related-images \
		--add-supported-arch amd64 \
		--replace-rbac-proxy "registry.redhat.io/openshift4/ose-kube-rbac-proxy:v4.10" \
		< bundle/manifests/rhacs-operator.clusterserviceversion.yaml \
		> build/bundle/manifests/rhacs-operator.clusterserviceversion.yaml
```

it replaces `kube-rbac-proxy` image in `operator/build/bundle/manifests/rhacs-operator.clusterserviceversion.yaml`
```
...
                image: registry.redhat.io/openshift4/ose-kube-rbac-proxy:v4.10
                name: kube-rbac-proxy
...
```


operator/bundle_helpers/rewrite.py
```diff
@@ -22,8 +24,10 @@ def rewrite(d, rewriter):
             if res is not None:
                 updates.append((k, res))
         for k, v in updates:
+            logging.info(f'Replaced: {d[k]} with {v}')
```
msugakov, line 27: We redirect the output and save it in the new CSV file. We shouldn't do that, rather we should have command-line option for input and output files.
Would this logging mess the `stdout` and appear in the middle of the output yaml?

operator/bundle_helpers/patch-csv.py
```diff
@@ -98,9 +100,8 @@ def parse_args():
                         help='Which operator image to use in the patched CSV')
     parser.add_argument("--no-related-images", action='store_true',
                         help='Disable passthrough of related images')
-    parser.add_argument("--replace-rbac-proxy", required=False, metavar='original-tag:replacement-image',
-                        nargs='+', help='Replacement directives for the RBAC proxy image',
-                        default=[])
+    parser.add_argument("--replace-rbac-proxy", required=False, metavar='replacement-image:v',
```
msugakov, line 103:
```suggestion
    parser.add_argument("--replace-rbac-proxy", required=False, metavar='replacement-image:tag',
```

operator/Makefile
```diff
@@ -410,7 +410,7 @@ upgrade-dirty-tag-via-olm: kuttl
 # Commands to enter local Python virtual environment and get needed dependencies there.
 ACTIVATE_PYTHON = python3 -m venv bundle_helpers/.venv ;\
 	. bundle_helpers/.venv/bin/activate ;\
-	pip3 install --upgrade pip==21.3.1 setuptools==59.6.0 ;\
```
msugakov, line 413: Omitting versions here may lead to problems when maintainers release broken versions. Omitting versions also leads to non-reproducible builds.

I think, we ran into something like this which forced us put versions. If the current version numbers are too old, you can try put some newer ones from https://pypi.org/

msugakov (commented):
Overall looks good, thank you!

operator/bundle_helpers/rewrite.py
msugakov: Seems like the output is driven to `stderr` by default as the output wasn't redirected here https://github.com/stackrox/stackrox/actions/runs/4025711606/jobs/6919288425#step:11:100 

BTW, I like how this logging looks. Maybe you can also do it for list replacement?

operator/bundle_helpers/patch-csv.py
```diff
@@ -113,6 +114,7 @@ def parse_args():
 
 
 def main():
+    logging.basicConfig(level=logging.INFO)
```
msugakov, line 117: Let's explicitly tell it should log to `stderr` so that I can sleep well at night:
```suggestion
    logging.basicConfig(stream=sys.stderr, level=logging.INFO)
```

operator/bundle_helpers/rewrite.py
kurlov: Yes, `logging` doesn't use the same output as `print` by default. I've adde logging to list replacement and I think there was a bug there. But it did not have any impact because list was used for loop over the doc so no real replacement

operator/Makefile
kurlov: I've bumped the version

kovayur (commented):
LGTM, just a few clarifying questions

operator/bundle_helpers/fix-spec-descriptor-order.py
```diff
@@ -3,6 +3,7 @@
 import sys
 import yaml
 
+
```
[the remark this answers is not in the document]
kurlov: we don't really need it. It's just my IDE formatted it according to [PEP-8](https://peps.python.org/pep-0008/) but I think it's ok to have it

operator/Makefile
```diff
@@ -410,7 +410,7 @@ upgrade-dirty-tag-via-olm: kuttl
...
+	pip3 install --upgrade pip==22.3.1 ;\
```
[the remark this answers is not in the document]
kurlov: I've reverted it because of openshift CI

operator/bundle_helpers/rewrite.py
```diff
@@ -14,16 +16,19 @@ def rewrite(d, rewriter):
         for i, elem in enumerate(d):
             res = rewrite(elem, rewriter)
             if res is not None:
-                list[i] = res
```
msugakov, line 17: Wow. Was there a bug in this `list[i] = res` line previously?

msugakov (approved):

operator/bundle_helpers/rewrite.py
kurlov: I think so. But this line was never used

kovayur (approved):

PR #4575, opened 2023-01-27
ROX-14468: Make new Network Graph a Tech Preview in nav

vjwilson:
## Description

Putting this out there as a simple solution

## Checklist
- [x] Investigated and inspected CI test results (unrelated flakes in Config Management tests)

## Testing Performed

Manual

How it looks when not the active section
Screen Shot 2023-01-27 at 2 35 36 PM

How it looks when it is the active section
Screen Shot 2023-01-27 at 2 35 53 PM

### DIFF
--- a/ui/apps/platform/src/routePaths.js
+++ b/ui/apps/platform/src/routePaths.js
@@ -153,7 +153,7 @@ export const basePathToLabelMap = {
     [networkBasePath]: 'Network Graph',
     [networkBasePathPF]: (
         <Flex>
-            <FlexItem>New Network Graph</FlexItem>
+            <FlexItem>Network Graph 2.0</FlexItem>
             <FlexItem>
                 <Badge style={{ backgroundColor: 'var(--pf-global--palette--gold-400)' }}>
                     Preview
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/routePaths.js
```diff
@@ -148,7 +151,16 @@ const vulnManagementPathToLabelMap = {
 export const basePathToLabelMap = {
     [dashboardPath]: 'Dashboard',
     [networkBasePath]: 'Network Graph',
-    [networkBasePathPF]: 'Network Graph-Patternfly',
+    [networkBasePathPF]: (
+        <Flex>
+            <FlexItem>New Network Graph</FlexItem>
+            <FlexItem>
+                <Badge style={{ backgroundColor: 'var(--pf-global--palette--gold-400)' }}>
```
sachaudh, line 158: Not necessary yet tho, but I imagine we'll probably do something similar for later features. If necessary, we should make a component when we do

sachaudh (commented):
I'll leave the naming choice for the nav item for Alan or Boaz to confirm. LGTM otherwise

alanonthegit (approved):
LGTM!

PR #4577, opened 2023-01-27
Dashrews/ROX-14598 do not open rocks in postgres mode

dashrews78:
## Description

Take advantage of utils.Should to ensure central does not try to open RocksDB or BoltDB in Postgres mode.  This way we will be informed during CI if we have introduced such cases and be able to resolve them.  

I should note that the bleve call for an in memory index `MemOnlyIndex` remains allowed as the compliance standards checks utilize that.

Has to be delivered after #4581 

## Checklist

## Testing Performed
 Ran it before #4581 with Postgres enabled but without Policy Categories off and it failed.  Additionally the unit tests I fixed failed because they were running in postgres mode and didn't actually use postgres.


central/globaldb/bolt.go
```diff
@@ -25,6 +28,10 @@ func initialize() {
 
 // GetGlobalDB returns a pointer to the global db.
 func GetGlobalDB() *bolt.DB {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		utils.Should(errors.New("unexpected call, should not open BoltDB while using Postgres"))
+	}
+
```
c-du, line 34: This has some overlap with the following call: postgres.LogCallerOnPostgres. It should not be called any more. It could be better if we rename/repurpose it.

connorgorman (changes requested):
Does it make sense to just create a function that checks and then otherwise just debug.PrintStack() and call utils.Should on a generic error, like UNEXPECTED opening of BoltDB or RocksDB?

pkg/utils/log.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/stackrox/rox/pkg/logging"
+)
+
+var log = logging.LoggerForModule()
+
+// LogCaller logs the caller of this function
+// This helps trace where calls into deprecated features are coming from.
+// e.g. calls into RocksDB and BoltDB initialization
+func LogCaller(name string) {
```
connorgorman, line 14: I don't think this is actually used anywhere cause you use debug.PrintStack() in the deprecated call

dashrews78: fair point.  Would you prefer a call to this or `debug.PrintStack()`?  Easy enough to wipe this out if we don't need it.

c-du (approved):

connorgorman (approved):

PR #4578, opened 2023-01-27
ROX-14137: auth provider declarative config

ivan-degtiarenko:
## Description

Auth provider struct which will be used to transform from YAML declarative config to Go struct.

## Checklist

## Testing Performed

Unit test added

### DIFF
--- a/pkg/declarativeconfig/auth_provider.go
+++ b/pkg/declarativeconfig/auth_provider.go
@@ -1,0 +1,49 @@
+package declarativeconfig
+
+type RequiredAttribute struct {
+	AttributeKey   string `yaml:"key,omitempty"`
+	AttributeValue string `yaml:"value,omitempty"`
+}
+
+type ClaimMapping struct {
+	PathToClaim    string `yaml:"pathToClaim,omitempty"`
+	MapToClaimName string `yaml:"mapTo,omitempty"`
+}
+
+type OIDCConfig struct {
+	Issuer                    string `yaml:"issuer,omitempty"`
+	CallbackMode              string `yaml:"mode,omitempty"`
+	ClientID                  string `yaml:"clientID,omitempty"`
+	ClientSecret              string `yaml:"clientSecret,omitempty"`
+	DisableOfflineAccessScope bool   `yaml:"disableOfflineAccessScope,omitempty"`
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+type SAMLConfig struct {
+	// TODO: add more fields
+	// "sp_issuer"
+	// "idp_metadata_url"
+	// "idp_cert_pem"
+	// "idp_sso_url"
+	// "idp_nameid_format"
+	SPIssuer string `yaml:"spIssuer,omitempty"`
+}
+
+// AuthProvider is representation of storage.PermissionSet that supports transformation from YAML.
+type AuthProvider struct {
+	Name               string              `yaml:"name,omitempty"`
+	MinimumRoleName    string              `yaml:"minimumRole,omitempty"`
+	UIEndpoint         string              `yaml:"uiEndpoint,omitempty"`
+	ExtraUIEndpoints   []string            `yaml:"extraUIEndpoints,omitempty"`
+	Groups             []Group             `yaml:"groups,omitempty"`
+	RequiredAttributes []RequiredAttribute `yaml:"requiredAttributes,omitempty"`
+	ClaimMappings      []ClaimMapping      `yaml:"claimMappings,omitempty"`
+	OIDCConfig         *OIDCConfig         `yaml:"oidc,omitempty"`
+	SAMLConfig         *SAMLConfig         `yaml:"saml,omitempty"`
+}
+
+type Group struct {
+	AttributeKey   string `yaml:"key,omitempty"`
+	AttributeValue string `yaml:"value,omitempty"`
+	RoleName       string `yaml:"role,omitempty"`
+}
--- a/pkg/declarativeconfig/auth_provider.go
+++ b/pkg/declarativeconfig/auth_provider.go
@@ -12,23 +12,38 @@ type ClaimMapping struct {
 	Name string `yaml:"name,omitempty"`
 }
 
-// OIDCConfig contains config values for OIDC auth provider.
-type OIDCConfig struct {
-	Issuer                    string `yaml:"issuer,omitempty"`
-	CallbackMode              string `yaml:"mode,omitempty"`
-	ClientID                  string `yaml:"clientID,omitempty"`
-	ClientSecret              string `yaml:"clientSecret,omitempty"`
-	DisableOfflineAccessScope bool   `yaml:"disableOfflineAccessScope,omitempty"`
-	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+// Group is representation of storage.Group that supports transformation from YAML.
+type Group struct {
+	AttributeKey   string `yaml:"key,omitempty"`
+	AttributeValue string `yaml:"value,omitempty"`
+	RoleName       string `yaml:"role,omitempty"`
 }
 
-// SAMLConfig contains config values for SAML auth provider.
+// OIDCConfig contains config values for OIDC auth provider.
+type OIDCConfig struct {
+	Issuer string `yaml:"issuer,omitempty"`
+	// Depending on callback mode, different OAuth 2.0 would be preferred.
+	// Possible values are: auto, post, query, fragment.
+	CallbackMode string `yaml:"mode,omitempty"`
+	ClientID     string `yaml:"clientID,omitempty"`
+	ClientSecret string `yaml:"clientSecret,omitempty"`
+	// Disables request for "offline_access" scope from OIDC identity provider.
+	DisableOfflineAccessScope bool `yaml:"disableOfflineAccessScope,omitempty"`
+	// Disables request for "offline_access" scope from OIDC identity provider.
+	DontUseClientSecretConfig bool `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+// SAMLConfig contains config values for SAML 2.0 auth provider.
+// There are two ways to configure SAML: static and dynamic.
+// For dynamic configuration, you only need to specify spIssuer and metadataURL.
+// For static configuration, specify spIssuer, certPEM, ssoURL and nameIdFormat.
 type SAMLConfig struct {
-	SpIssuer     string `yaml:"spIssuer,omitempty"`
-	MetadataURL  string `yaml:"metadataURL,omitempty"`
-	CertPEM      string `yaml:"certPEM,omitempty"`
+	SpIssuer    string `yaml:"spIssuer,omitempty"`
+	MetadataURL string `yaml:"metadataURL,omitempty"`
+	// SAML 2.0 IdP Certificate in PEM format
+	Cert         string `yaml:"cert,omitempty"`
 	SsoURL       string `yaml:"ssoURL,omitempty"`
-	NameidFormat string `yaml:"nameIdFormat,omitempty"`
+	NameIDFormat string `yaml:"nameIdFormat,omitempty"`
 }
 
 // IAPConfig contains config values for IAP auth provider.
@@ -38,7 +53,15 @@ type IAPConfig struct {
 
 // UserpkiConfig contains config values for User Certificates auth provider.
 type UserpkiConfig struct {
-	Keys string `yaml:"keys,omitempty"`
+	// Certificate authorities in PEM format
+	CertificateAuthorities string `yaml:"certificateAuthorities,omitempty"`
+}
+
+// OpenshiftConfig contains config values for Openshift auth provider.
+// The only value "enable" is a flag which can only be set to true.
+// If you don't want auth provider to be Openshift auth provider, don't specify "openshift" section.
+type OpenshiftConfig struct {
+	Enable bool `yaml:"enable,omitempty"`
 }
 
 // AuthProvider is representation of storage.AuthProvider that supports transformation from YAML.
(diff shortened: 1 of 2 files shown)


pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
+package declarativeconfig
+
+type RequiredAttribute struct {
+	AttributeKey   string `yaml:"key,omitempty"`
+	AttributeValue string `yaml:"value,omitempty"`
+}
+
+type ClaimMapping struct {
+	PathToClaim    string `yaml:"pathToClaim,omitempty"`
+	MapToClaimName string `yaml:"mapTo,omitempty"`
```
dhaus67, line 10: These a worded a bit unfortunate IMO, I think we can simplify them with proper documentation (i.e. comments here):
```yaml
claimMappings:
- path: some.claim.path
  name: new-name
```

dhaus67 (changes requested):

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	// "idp_cert_pem"
+	// "idp_sso_url"
+	// "idp_nameid_format"
+	SPIssuer string `yaml:"spIssuer,omitempty"`
+}
+
+// AuthProvider is representation of storage.PermissionSet that supports transformation from YAML.
+type AuthProvider struct {
+	Name               string              `yaml:"name,omitempty"`
+	MinimumRoleName    string              `yaml:"minimumRole,omitempty"`
```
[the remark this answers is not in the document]
ivan-degtiarenko: > Either we handle this within the struct (if miminumRoleName == "" -> do not create default group or within the configuration). 

By handling within the struct you probably mean within the  `declarative config Go struct -> proto struct` transformation logic?

 Right now, whatever we do, I think we should just keep this field as a string and not change YAML -> Go struct transformation logic. I would keep YAML -> declarative config Go struct transformation as simple as possible.

>  I think we should handle it within the struct itself, leave a comment specifying that, if left empty, no default group will be created.

Let's leave a TODO if we have ticket for transformation/validation of declarative config structs.

> (On an unrelated note, it might be good to add some validation to some of the structs in a basic manner, either directly or via things such as [go-playground/validator](https://github.com/go-playground/validator) which we already seem to have as an indirect dependency)

100%

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+type SAMLConfig struct {
+	// TODO: add more fields
+	// "sp_issuer"
+	// "idp_metadata_url"
+	// "idp_cert_pem"
+	// "idp_sso_url"
+	// "idp_nameid_format"
```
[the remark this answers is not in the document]
ivan-degtiarenko: Yep, coming up

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	MapToClaimName string `yaml:"mapTo,omitempty"`
+}
+
+type OIDCConfig struct {
+	Issuer                    string `yaml:"issuer,omitempty"`
+	CallbackMode              string `yaml:"mode,omitempty"`
+	ClientID                  string `yaml:"clientID,omitempty"`
+	ClientSecret              string `yaml:"clientSecret,omitempty"`
+	DisableOfflineAccessScope bool   `yaml:"disableOfflineAccessScope,omitempty"`
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
```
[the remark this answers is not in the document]
ivan-degtiarenko: > disableOfflineAccessScope is currently used by ACSCS only (?) and dontUseClientSecretConfig for clients like the auth0 ones, right?

`dontUseClientSecretConfig` can be selected in the UI for all OIDC auth providers. `disableOfflineAccessScope` is used by ACSCS and we don't display it in the UI, but it can be used for other OIDC auth providers as well via API.

> I'd prefer the clear documentation and "everything-and-the-kitchen-sink" as configuration

Agreed. I would prefer to have a stricter config where every field is documented. I see no need to gatekeep from declarative config users any parameters which are exposed via API.

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	ClaimMappings      []ClaimMapping      `yaml:"claimMappings,omitempty"`
+	OIDCConfig         *OIDCConfig         `yaml:"oidc,omitempty"`
+	SAMLConfig         *SAMLConfig         `yaml:"saml,omitempty"`
+}
+
+type Group struct {
+	AttributeKey   string `yaml:"key,omitempty"`
+	AttributeValue string `yaml:"value,omitempty"`
+	RoleName       string `yaml:"role,omitempty"`
+}
```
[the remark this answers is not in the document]
ivan-degtiarenko: I thought since we bound declarative groups to auth provider in declarative config, we could just re-create them from scratch. Looking at this from the user perspective, it is hard to come up with a proper name for the group compared to other resources - I would avoid that if possible.

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,49 @@
...
+
+type ClaimMapping struct {
+	PathToClaim    string `yaml:"pathToClaim,omitempty"`
+	MapToClaimName string `yaml:"mapTo,omitempty"`
+}
+
+type OIDCConfig struct {
+	Issuer                    string `yaml:"issuer,omitempty"`
+	CallbackMode              string `yaml:"mode,omitempty"`
```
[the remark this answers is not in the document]
ivan-degtiarenko: I agree that guidance should be provided. I wonder if comments written in these files can be compiled into documentation, or should we introduce separate documentation file for that?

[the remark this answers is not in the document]
dhaus67: I think for now it would be best to include the comments here.
I'm not sure how we are going to do the whole documentation for these configs later on in the actual documentation (my current idea would be to propose something similar like [dex documentation](https://dexidp.io/docs/connectors/github/), which I really like), but at the very least we can use the code comments as one of the sources.

pkg/declarativeconfig/auth_provider.go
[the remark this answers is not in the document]
dhaus67: Sounds good to me, with the additional comments in code explaining the field a bit more (especially the `disableOfflineAccessScope` one, as it's the first time we expose it any form of API).

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+// SAMLConfig contains config values for SAML auth provider.
+type SAMLConfig struct {
+	SpIssuer     string `yaml:"spIssuer,omitempty"`
+	MetadataURL  string `yaml:"metadataURL,omitempty"`
+	CertPEM      string `yaml:"certPEM,omitempty"`
+	SsoURL       string `yaml:"ssoURL,omitempty"`
+	NameidFormat string `yaml:"nameIdFormat,omitempty"`
```
dhaus67, line 31:
```suggestion
	NameIDFormat string `yaml:"nameIdFormat,omitempty"`
```

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// IAPConfig contains config values for IAP auth provider.
+type IAPConfig struct {
+	Audience string `yaml:"audience,omitempty"`
+}
+
+// UserpkiConfig contains config values for User Certificates auth provider.
+type UserpkiConfig struct {
+	Keys string `yaml:"keys,omitempty"`
```
dhaus67, line 41: Those are not keys, but rather CA certificates, so should name this something like CAPems?

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+	ClientSecret              string `yaml:"clientSecret,omitempty"`
+	DisableOfflineAccessScope bool   `yaml:"disableOfflineAccessScope,omitempty"`
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+// SAMLConfig contains config values for SAML auth provider.
+type SAMLConfig struct {
+	SpIssuer     string `yaml:"spIssuer,omitempty"`
+	MetadataURL  string `yaml:"metadataURL,omitempty"`
+	CertPEM      string `yaml:"certPEM,omitempty"`
```
dhaus67, line 29: (also relates to the comment below for User PKI): not sure we want to do this, so more of a question:
I'm not sure we want to be so "specific" with our naming, i.e. instead of naming it `certPem`, name it `cert` and have documentation that we expect the cert in PEM format here.

dhaus67, line 29: Isn't it more of a CA instead of a `cert`? We are talking about the `CA` of the SAML issuer, right? So, would rather be explicit and name it `CA` instead.

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+	Issuer                    string `yaml:"issuer,omitempty"`
+	CallbackMode              string `yaml:"mode,omitempty"`
+	ClientID                  string `yaml:"clientID,omitempty"`
+	ClientSecret              string `yaml:"clientSecret,omitempty"`
+	DisableOfflineAccessScope bool   `yaml:"disableOfflineAccessScope,omitempty"`
+	DontUseClientSecretConfig bool   `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+// SAMLConfig contains config values for SAML auth provider.
+type SAMLConfig struct {
```
dhaus67, line 26: Would be also nice to have more documentation about the requirements of the fields, i.e. if `metadataURL` is set, you do not need to provide the `ssoURL, nameIdFormat`.

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+	Name               string              `yaml:"name,omitempty"`
+	MinimumRoleName    string              `yaml:"minimumRole,omitempty"`
+	UIEndpoint         string              `yaml:"uiEndpoint,omitempty"`
+	ExtraUIEndpoints   []string            `yaml:"extraUIEndpoints,omitempty"`
+	Groups             []Group             `yaml:"groups,omitempty"`
+	RequiredAttributes []RequiredAttribute `yaml:"requiredAttributes,omitempty"`
+	ClaimMappings      []ClaimMapping      `yaml:"claimMappings,omitempty"`
+	OIDCConfig         *OIDCConfig         `yaml:"oidc,omitempty"`
+	SAMLConfig         *SAMLConfig         `yaml:"saml,omitempty"`
+	IAPConfig          *IAPConfig          `yaml:"iap,omitempty"`
```
dhaus67, line 57: not sure on this one: is `IAP` generally understood as `Google IAP`, or can we drop the prefix?

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,67 @@
...
+	UserpkiConfig      *UserpkiConfig      `yaml:"userpki,omitempty"`
+	EnableOpenshift    bool                `yaml:"enableOpenshift,omitempty"`
```
dhaus67, line 59: How about keeping it consistent, and doing something like:
```go
type AuthProvider struct {
  OpenShiftConfig *OpenShiftConfig `yaml:"openshift,omitempty"`
}

type OpenShiftConfig struct {
  Enable bool `yaml:"enable,omitempty"`
}
```

This seems more in line with the other settings, and specifying something like:
```yaml
authProvider:
- name: something
  enableOpenshift: true
```
looks a bit weird to me compared to the other methods.

pkg/declarativeconfig/auth_provider.go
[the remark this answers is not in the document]
dhaus67: Hm, yeah that's true actually I think we can skip it here and handle it in the reconciliation by just re-creating all groups.

pkg/declarativeconfig/auth_provider_test.go
```diff
@@ -0,0 +1,66 @@
...
+  role: "Admin"
+- key: "email"
+  value: "someone@stackrox.com"
+  role: "Analyst"
+oidc:
+  issuer: "https://stackrox.com"
+  mode: "auto select"
+  clientID: "some-client-id"
+  clientSecret: "some-client-secret"
+  disableOfflineAccessScope: true
```
dhaus67, line 31: Would be nice to add some more tests for each specific authentication provider.

pkg/declarativeconfig/auth_provider.go
ivan-degtiarenko: When I google IAP, I see only references to Google IAP. We also refer in our codebase to auth provider of this type as `iap`.

Other IAP services can exist though. If you feel strongly, we can change this field name to `googleIAP`.

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,83 @@
...
+type OIDCConfig struct {
+	Issuer string `yaml:"issuer,omitempty"`
+	// Depending on callback mode, different OAuth 2.0 would be preferred.
+	// Possible values are: auto, post, query, fragment.
+	CallbackMode string `yaml:"mode,omitempty"`
+	ClientID     string `yaml:"clientID,omitempty"`
+	ClientSecret string `yaml:"clientSecret,omitempty"`
+	// Disables request for "offline_access" scope from OIDC identity provider.
+	DisableOfflineAccessScope bool `yaml:"disableOfflineAccessScope,omitempty"`
+	// Disables request for "offline_access" scope from OIDC identity provider.
```
dhaus67, line 32: Probably not the comment you wanted there

pkg/declarativeconfig/auth_provider.go
```diff
@@ -0,0 +1,83 @@
...
+	DontUseClientSecretConfig bool `yaml:"dontUseClientSecretConfig,omitempty"`
+}
+
+// SAMLConfig contains config values for SAML 2.0 auth provider.
+// There are two ways to configure SAML: static and dynamic.
+// For dynamic configuration, you only need to specify spIssuer and metadataURL.
+// For static configuration, specify spIssuer, certPEM, ssoURL and nameIdFormat.
```
dhaus67, line 39:
```suggestion
// For static configuration, specify spIssuer, cert, ssoURL and nameIdFormat.
```

pkg/declarativeconfig/auth_provider.go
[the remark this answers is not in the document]
dhaus67: Ticket for transformation is [ROX-14148](https://issues.redhat.com/browse/ROX-14148)

dhaus67 (approved):

PR #4581, opened 2023-01-27
ROX-14634: Miscellaneous policy category fixes: Feature flag removal, upgrade path crashes, roxdb import export workflow fixes

clickboo:
## Description
1. Removing the policy category feature flag both correctness during upgrades.
2. Fixing category migration since it was causing a crash during upgrades
3. Fixing the category datastore usage in the policy import export workflow to be under feature flag to avoid crashes in roxdb/boltdb flow
4. Fixing a bunch of e2e tests
5. Removing extraneous roxdb stuff for policy categories (which was remnant after the decision to make categories postgres only)

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


migrator/migrations/m_170_to_m_171_create_policy_categories_and_edges/policycategorypostgresstore/store.go
```diff
@@ -109,6 +119,58 @@ func (s *storeImpl) Upsert(ctx context.Context, obj *storage.PolicyCategory) err
...
+		defer s.mutex.Unlock()
+
+		if len(objs) < batchAfter {
+			return s.upsert(ctx, objs...)
+		}
+		return s.copyFrom(ctx, objs...)
+	})
+}
+
+// DeleteMany removes the objects associated to the specified IDs from the store.
```
dashrews78, line 138: we don't really need the Delete routines in the migration.  May be cleaner to remove them.

qa-tests-backend/src/test/groovy/UpgradesTest.groovy
```diff
@@ -200,7 +200,7 @@ class UpgradesTest extends BaseSpecification {
     }
 
     @Tag("Upgrade")
-    @IgnoreIf({ Env.CI_JOB_NAME.contains("postgres") })
+    @IgnoreIf({ Env.CI_JOBNAME.contains("postgres") })
```
dashrews78, line 203: Are these tests still a problem since the feature flag has been removed and this is enabled with Postgres?  If it still fails, is it because the upgrade tests is using versions of Postgres software prior to categories?

scripts/ci/jobs/gke_postgres_upgrade_tests.py
```diff
@@ -10,6 +10,8 @@
 from ci_tests import PostgresUpgradeTest
 from post_tests import PostClusterTest, FinalPost
 
+os.environ["ROX_POSTGRES_DATASTORE"] = "true"
```
dashrews78, line 13: This shouldn't be needed because of what this tests does and how it does it.  This test actually starts in Rocks and goes back and forth.  I'm not even certain setting this would make a difference.

migrator/migrations/m_170_to_m_171_create_policy_categories_and_edges/policycategorypostgresstore/store.go
dashrews78: ah, that is right, it does.  carry on then.

scripts/ci/jobs/gke_postgres_upgrade_tests.py
clickboo: Yup, it is not, it seemed really odd that gke_upgrade_tests.py explicitly turns the flag off. But the postgres counterpart doesn't do anything about that flag. In the least, a comment here would help. To keep the test scripts less "magical".

central/policy/datastore/datastore_impl_postgres_test.go
```diff
@@ -187,7 +185,7 @@ func (s *PolicyPostgresDataStoreTestSuite) TestImportPolicy() {
 }
 
 func (s *PolicyPostgresDataStoreTestSuite) TestSearchPolicyCategoryFeatureDisabled() {
```
dashrews78, line 187: nit:  I'm not sure this test is valid as this is explicitly a Postgres test.  It doesn't cause harm but it essentially just tests RocksDB functionality which would be covered in the RocksDB backed tests.

central/policycategoryedge/datastore/singleton.go
```diff
@@ -21,7 +20,7 @@ func initialize() {
 	var storage store.Store
```
dashrews78, line 20: nit: we don't need to declare these outside the if anymore since that is the only time we use them.  It might be cleaner now to do

```
if !env.PostgresDatastoreEnabled.BooleanSetting() {
   return nil
}
```

clickboo: Yes, done.

central/policy/datastore/datastore_impl_postgres_test.go
clickboo: Not really, a good test to have since policy object doesn't get touched in the rocksdb path.

qa-tests-backend/src/test/groovy/UpgradesTest.groovy
```diff
@@ -200,7 +200,7 @@ class UpgradesTest extends BaseSpecification {
     }
 
     @Tag("Upgrade")
-    @IgnoreIf({ Env.CI_JOB_NAME.contains("postgres") })
+    // @IgnoreIf({ Env.CI_JOBNAME.contains("postgres") })
```
dashrews78, line 203: remove?

tests/e2e/bats/fixtures/test-pod-without-oomkilled_object.json
```diff
@@ -175,7 +175,11 @@
                         "value": "false"
                     },
                     {
+<<<<<<< HEAD
```
dashrews78, line 178: missed from rebase

qa-tests-backend/src/test/groovy/UpgradesTest.groovy
clickboo: Yeah trying it out before remove, CI will tell us the truth.

central/main.go
```diff
@@ -401,7 +401,7 @@ func servicesToRegister(registry authproviders.Registry, authzTraceSink observe.
 		servicesToRegister = append(servicesToRegister, cveService.Singleton())
 	}
 
-	if features.NewPolicyCategories.Enabled() && env.PostgresDatastoreEnabled.BooleanSetting() {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
 		servicesToRegister = append(servicesToRegister, policyCategoryService.Singleton())
```
theencee, line 405: Move this to under line 398

central/policy/datastore/datastore_impl.go
```diff
@@ -331,9 +332,15 @@ func (ds *datastoreImpl) ImportPolicies(ctx context.Context, importPolicies []*s
 }
 
 func (ds *datastoreImpl) importPolicy(ctx context.Context, policy *storage.Policy, overwrite bool, policyNameIDMap map[string]string) *v1.ImportPolicyResponse {
+	if policy.GetId() == "" {
```
theencee, line 335: out of curiosity: what was the purpose behind moving this up?

clickboo: the response clone has the ID without it needing to be set

dashrews78 (approved):
I think it looks pretty good now.

PR #4585, opened 2023-01-27
Dashrews/ROX-14614 remove code that drops central active when a rocksDB is present

dashrews78:
## Description

To reduce risk, this PR removes the code that uses a timestamp to determine if a version of RocksDB was used more recently than Postgres.  Once you have data in Postgres, that will be the data of choice if the Postgres flag is set.  This will keep us from blowing away Postgres data in the event something odd happened with the timestamp with RocksDB.

## Checklist

## Testing Performed

Unit tests were updated accordingly.  Manual steps from the test matrix were also completed.  E2E tests will come as part of the larger e2e updates per the test matrix.  Additionally since the upgrades behave differently in early versions we need to wait until this version of software becomes not the most recent so we can step wise add tests such as upgrade from Rocks to this version then on to the next version postgres to postgres upgrade.

### DIFF
--- a/migrator/clone/postgres/db_clone_manager_impl_test.go
+++ b/migrator/clone/postgres/db_clone_manager_impl_test.go
@@ -212,14 +212,7 @@ func (s *PostgresCloneManagerSuite) TestGetCloneMigrateRocks() {
 	pgtest.DropDatabase(s.T(), tempDB)
 	pgtest.DropDatabase(s.T(), migrations.RestoreDatabase)
 	pgtest.DropDatabase(s.T(), migrations.BackupDatabase)
-
-	// Set central_active in the future and have no previous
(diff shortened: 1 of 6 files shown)


connorgorman (approved):

migrator/clone/postgres/db_clone_manager_impl_test.go
```diff
@@ -232,36 +225,37 @@ func (s *PostgresCloneManagerSuite) TestGetCloneMigrateRocks() {
 		LastPersisted: time.Now(),
```
rhybrillou, line 225: It looks like the LastPersisted field in the MigrationVersion is not read anymore.
Would it make sense to remove it from the struct definition and stop setting it ?

rhybrillou (approved):

PR #4586, opened 2023-01-27
ROX-14606: Fix issue with statement timeout crashes on pruning

connorgorman:
## Description

Was running into statement_timeouts that were crashing Central. Should evaluate how we can do this better, but this will avoid crashes

## Checklist

## Testing Performed

Removes the crash


dashrews78 (approved):
Should probably add this under https://issues.redhat.com/browse/ROX-11351

tommartensen:
Removed from milestone as code change was already part of 3.74.x release branch.

PR #4588, opened 2023-01-27
fix: network graph bug - port configuration text fix

sachaudh:
## Description

This PR fixes the following issue in the new network graph:

* When I click on a specific deployment, details popup on the right. I scrolled down and found Port configurations. I see two boxes I can expand, but when collapsed, they seem blank

Screenshot 2023-01-27 at 3 34 10 PM
Screenshot 2023-01-27 at 3 34 31 PM

## Checklist

### DIFF
--- a/ui/apps/platform/src/Components/DeploymentPortConfig.tsx
+++ b/ui/apps/platform/src/Components/DeploymentPortConfig.tsx
@@ -25,7 +25,7 @@ function DeploymentPortConfig({ port }: DeploymentPortConfigProps) {
     };
 
     const toggleText = port.name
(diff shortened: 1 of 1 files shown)


vjwilson (approved):
/lgtm

zhenpesky:
@sachaudh Looks good but just to call out a small thing on the symbol on a use case: For some of the port names we have been using a short hyphen(a minus sign),  we want to differentiate that by using the 'EM dash' (it will look longer than a hyphen(a minus sign)). (option+shift+'mimus sign' on keyboard).

zhenpesky (approved):

PR #4590, opened 2023-01-28
ROX-13431: Network Graph PF edge performance fix

alwayshooin:
so it turns out it was the `ColaGroup` layout all along 😭  
i changed it to `ColaNoForce` and it seems to be working fine. 
i also added some clean up around `useEffect` bc the console was upset we were trying to set the state of unmounted components
it should look/feel/behave the same (even better!) than before 🥲 

image

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -95,7 +95,8 @@ function NetworkGraphPage() {
         const isQueryFilterComplete = isCompleteSearchFilter(remainingQuery);
 
         // only refresh the graph data from the API if both a cluster and at least one namespace are selected
-        const isClusterNamespaceSelected = clusterFromUrl && namespacesFromUrl.length > 0;
+        const isClusterNamespaceSelected =
+            clusterFromUrl && namespacesFromUrl.length > 0 && deploymentCount;
 
         if (isQueryFilterComplete && isClusterNamespaceSelected) {
             if (selectedClusterId) {
@@ -161,13 +162,18 @@ function NetworkGraphPage() {
                     .finally(() => setIsLoading(false));
             }
         }
-    }, [clusterFromUrl, namespacesFromUrl, deploymentsFromUrl, remainingQuery, timeWindow]);
+    }, [
+        clusterFromUrl,
+        namespacesFromUrl,
+        deploymentsFromUrl,
+        remainingQuery,
+        timeWindow,
+        deploymentCount,
+    ]);
 
     function toggleCIDRBlockForm() {
         setIsCIDRBlockFormOpen(!isCIDRBlockFormOpen);
     }
-
-    console.log('NetworkGraphPage');
 
     return (
         <>
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -174,7 +174,8 @@ function NetworkGraphPage() {
     function toggleCIDRBlockForm() {
         setIsCIDRBlockFormOpen(!isCIDRBlockFormOpen);
     }
-    console.log('NetworkGrpahPage');
+
+    console.log('NetworkGraphPage');
 
     return (
         <>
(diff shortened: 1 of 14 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -166,9 +166,9 @@ function NetworkGraphPage() {
         clusterFromUrl,
         namespacesFromUrl,
         deploymentsFromUrl,
-        deploymentCount,
         remainingQuery,
         timeWindow,
+        deploymentCount,
```
alwayshooin, line 171: i thought i could remove this but turns out i need it to ensure the graph gets loaded on refresh 🤷🏼‍♀️

ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
```diff
@@ -126,6 +125,10 @@ const TopologyComponent = ({
         onSelect(ids);
     });
 
+    useEffect(() => {
+        controller.fromModel(model);
+    }, [controller, model]);
+
```
alwayshooin, line 131: this was to prevent a console error ("Cannot update a component (`StyleNode`) while rendering a different component (`TopologyComponent`). To locate the bad setState() call inside `TopologyComponent`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render")

ui/apps/platform/src/Containers/NetworkGraph/hooks/useNetworkPolicySimulator.ts
```diff
@@ -87,22 +87,6 @@ function useNetworkPolicySimulator({ simulation, clusterId }: UseNetworkPolicySi
...
-            });
-        } else {
-            setSimulator({
-                state,
-                modification: null,
-                error: '',
-                isLoading: true,
-            });
-        }
-
```
alwayshooin, line 105: i removed this because it seemed unnecessary since `setSimulator` forces a rerender and also is set again in the switch statement below -- this removed one unnecessary render in the `NetworkGraph` component

alwayshooin: fwiw it seems like there's no impact on the network policy simulator (it worked fine when i tested it out) but @sachaudh should probably double check

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -83,7 +83,7 @@ function NetworkGraphPage() {
     } = getScopeHierarchy(searchFilter);
     const simulation = getSimulation(simulationQueryValue);
 
-    const hasClusterNamespaceSelected = Boolean(clusterFromUrl && namespacesFromUrl.length);
+    const hasClusterNamespaceSelected = !!(clusterFromUrl && namespacesFromUrl.length);
```
vjwilson, line 86: nit: I recently read a compelling argument that casting as a Boolean makes these case easier to pick out when maintaining the code than the old double-negation `!!`

vjwilson (approved):

ui/apps/platform/src/hooks/useFetchNetworkPolicies.ts
```diff
@@ -29,6 +29,8 @@ function useFetchNetworkPolicies(policyIds: string[]): Result {
                     setResult({ networkPolicies: [], error, isLoading: false });
                 });
         }
+
+        return () => setResult(defaultResultState);
```
sachaudh, line 33: Oh, what's the benefit of this?

ui/apps/platform/src/Containers/NetworkGraph/hooks/useNetworkPolicySimulator.ts
sachaudh: Well, it was added for the `isLoading: true` part. When the simulation changes, we want to show a loading indicator until the data returns. That's why it renders twice. Let us keep it if it isn't hindering performance

sachaudh (changes requested):
LGTM. I would suggest keeping the simulator code since it shows a loading indicator when loading data rather than suddenly changing the view.

ui/apps/platform/src/hooks/useFetchNetworkPolicies.ts
alwayshooin: it's useEffect clean up when the component unmounts

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
alwayshooin: ohhh could you share the article??

sachaudh (approved):

PR #4592, opened 2023-01-29
ROX-13958: Enable collections by removing ROX_OBJECT_COLLECTIONS and gate it behind postgres flag

theencee:
## Description

This consolidates the collections flag behind just the postgres flag. While it would've been nicer to just enable the collections flag, in a bunch of areas of the code it was just gated behind that and not the postgres which would lead to errors (e.g. trying to read from collections table when there is no central-db).

## Checklist

## Testing Performed

Ran all the tests in CI and locally. Tested manually by deploying with postgres flag on (see collections) and with it off (no collections). Also did an upgrade from rocks to postgres and saw that collections now appeared.


janisz (approved):

central/resourcecollection/service/service_impl_test.go
```diff
@@ -39,8 +39,8 @@ func (suite *CollectionServiceTestSuite) SetupSuite() {
 	suite.dataStore = datastoreMocks.NewMockDataStore(suite.mockCtrl)
 	suite.queryResolver = datastoreMocks.NewMockQueryResolver(suite.mockCtrl)
 	suite.deploymentDS = deploymentDSMocks.NewMockDataStore(suite.mockCtrl)
-	suite.T().Setenv(features.ObjectCollections.EnvVar(), "true")
 	suite.collectionService = New(suite.dataStore, suite.queryResolver, suite.deploymentDS)
+	suite.T().Setenv(env.PostgresDatastoreEnabled.EnvVar(), "true")
```
charmik-redhat, line 43: Can add skip right after `Setenv` here and avoid calling skip in each test.

central/resourcecollection/service/service_impl_test.go
```diff
@@ -51,7 +51,7 @@ func (suite *CollectionServiceTestSuite) TearDownSuite() {
 }
 
 func (suite *CollectionServiceTestSuite) TestListCollectionSelectors() {
-	if !features.ObjectCollections.Enabled() {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
```
charmik-redhat, line 54: Remove skip from here and skip in `SetupSuite` right after `suite.T().Setenv(env.PostgresDatastoreEnabled.EnvVar(), "true")`

charmik-redhat (commented):
Left a couple of minor comments. Otherwise BE changed look good.

central/main.go
```diff
@@ -395,10 +395,7 @@ func servicesToRegister(registry authproviders.Registry, authzTraceSink observe.
 		servicesToRegister = append(servicesToRegister, clusterCVEService.Singleton())
 		servicesToRegister = append(servicesToRegister, imageCVEService.Singleton())
 		servicesToRegister = append(servicesToRegister, nodeCVEService.Singleton())
-
-		if features.ObjectCollections.Enabled() {
-			servicesToRegister = append(servicesToRegister, collectionService.Singleton())
-		}
+		servicesToRegister = append(servicesToRegister, collectionService.Singleton())
```
md2119, line 398: Put this behind postgres env var. Otherwise, this can panic because the singleton returns nil and we eventually call `RegisterServiceServer` for all registered services, in this call `RegisterServiceServer` on nil.

keyallis (approved):
I'd say LGTM. Ping someone from FE to see if they have any thoughts on places it might also be used in an obfuscated way @dvail

central/resourcecollection/service/service_impl_test.go
md2119: +1 to Charmik's suggestion

central/main.go
theencee: Oh looks like git hides that line by default. [It is behind the flag in line 394](https://github.com/stackrox/stackrox/pull/4592/files#diff-43f1bb62accddc8c479ff1efa4084785789e26508ba52bc30096c6ca5555d070R394)

dvail:
On the FE the flag is always used in a straightforward way that can be searched via direct string search. LGTM too from a FE perspective.

central/main.go
md2119: Oh yes, had to hit expand

md2119 (approved):

PR #4596, opened 2023-01-30
ROX-12948 add port pool in DeploymentExposure tests

lvalerom:
## Description

Changes the tests to retry with a different port if applying the service fails due to the port being already allocated.

Depends on:
* #4729

## Checklist

## Testing Performed

* Tested with a local cluster and an infra cluster. Test passed even with all the services having a port collision.


sensor/tests/resource/service/service_test.go
```diff
@@ -185,71 +187,157 @@ func (s *DeploymentExposureSuite) Test_ClusterIpPermutation() {
 	)
 }
 
+var nextPort int32 = 30007
```
lvalerom, line 190: Set this to 30000.

sensor/tests/resource/service/service_test.go
```diff
@@ -185,71 +187,157 @@ func (s *DeploymentExposureSuite) Test_ClusterIpPermutation() {
...
-							ContainerPort: 80,
-							Exposure:      storage.PortConfig_NODE,
-							ExposureInfos: []*storage.PortConfig_ExposureInfo{
-								{
-									ServiceName: fmt.Sprintf("nginx-svc-node-port-%d", c.port),
-									ServicePort: 80,
-									NodePort:    c.port,
-									Level:       storage.PortConfig_NODE,
+		var err, portErr error
+		port := c.port
```
lvalerom, line 290: Use `getPort` here and generate the service here.

sensor/tests/resource/service/service_test.go
```diff
@@ -450,66 +477,183 @@ func (s *DeploymentExposureSuite) Test_NodePortPermutationWithPod() {
...
+			},
+		},
+	}
+}
+
+func setDynamicFields(svc *v1.Service, name string, port int32, sel map[string]string, serviceFunc func(*v1.Service, string, int32, map[string]string)) {
+	serviceFunc(svc, name, port, sel)
+}
+
+func setDynamicFieldsInSlice(resources []resource.K8sResourceInfo, portConfig []*storage.PortConfig, name string, port int32, sel map[string]string, svcFunc serviceFunc, pcFunc portConfigFunc) {
```
fredrb, line 652: I think naming can be improved here. I suggest using the suffix `fn` for functions passed as parameter. 

Also, the parameter name `pcFunc` is a bit cryptic, could we call it as it is, something like: `portConfigFn`? Same for `svcFunc`.

fredrb (approved):
Nice that you were able to remove the service yaml files. I think this looks better now. I don't see any blockers if the tests are green. I do think some variable names could be improved though.

sensor/tests/resource/service/service_test.go
lvalerom: Done

lvalerom:
> /test gke-nongroovy-e2e-tests

This failure was unrelated to the changes; re-triggering to make sure the changed tests pass.

PR #4598, opened 2023-01-30
Fix inconsistent data state after using embedded collection filter

dvail:
## Description

A logic error in when the embedded collection data was filtered was leading to cases where the client side form state would be invalid.

If the user took the following steps:
1. Attached a collection
2. Filtered the available collection list in a way that the first collection was not visible
3. Attached a second collection, and cleared the search filter

This would result in both of the collections being displayed in the "attached" section of the UI, but the underlying form data would only contain the second collection, which impacted dry-run requests as well as saving the collection.

The fix was to move the search filtering from the `useEmbeddedCollections` hook that fetches the data to the component that displayed the data via a filtering callback.

## Follow ups

- Cover this case in e2e tests

## Checklist

## Testing Performed

Attach a collection and view the updated results in the side panel.
image

Filter the collection lists so that the attach collection is not visible.
image

Attach a second collection and view the updated results in the side panel and ensure that deployments from both embedded collections are displayed.
image

Save the collection, and revisit the collection page to ensure both embedded collections are saved correctly.
image

### DIFF
--- a/ui/apps/platform/cypress/integration/collections/deploymentMatching.test.js
+++ b/ui/apps/platform/cypress/integration/collections/deploymentMatching.test.js
@@ -4,6 +4,7 @@ import {
     assertDeploymentsAreMatched,
     assertDeploymentsAreMatchedExactly,
     assertDeploymentsAreNotMatched,
+    tryCreateCollection,
     tryDeleteCollection,
     visitCollections,
 } from './Collections.helpers';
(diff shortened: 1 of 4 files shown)


dvail:
* master
  * **PR #4598** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4598" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4598?utm_source=stack-comment).

pedrottimark (approved):

PR #4601, opened 2023-01-30
Refactoring: shared Telemeter interface package

0x656b694d:
## Description

A suggestion for a refactoring:
* `Telemeter.With` overrides the User ID for which the following call is addressed;
* `Telemeter.As` overrides the device context to impersonate the device which sends the messages.

## Checklist

## Testing Performed

Unit tests.

### DIFF
--- a/pkg/telemetry/phonehome/segment/segment.go
+++ b/pkg/telemetry/phonehome/segment/segment.go
@@ -91,36 +91,38 @@ func (t *segmentTelemeter) Stop() {
 	}
 }
 
-func makeDeviceContext(clientID, clientType string) *segment.Context {
-	if clientID == "" {
+func (t *segmentTelemeter) overrideUserID(o *telemeter.CallOptions) string {
+	if o.UserID != "" {
+		return o.UserID
+	}
+	return t.userID
+}
+
+func makeDeviceContext(o *telemeter.CallOptions) *segment.Context {
+	if o.ClientID == "" {
 		return nil
 	}
 	return &segment.Context{
 		Device: segment.DeviceInfo{
-			Id:   clientID,
-			Type: clientType,
+			Id:   o.ClientID,
+			Type: o.ClientType,
 		},
 	}
 }
 
-func (t *segmentTelemeter) User(userID string) telemeter.Telemeter {
-	return &segmentTelemeter{client: t.client, userID: userID, ctx: t.ctx}
-}
-
-func (t *segmentTelemeter) As(clientID string, clientType string) telemeter.Telemeter {
-	return &segmentTelemeter{client: t.client, userID: t.userID, ctx: makeDeviceContext(clientID, clientType)}
-}
-
-func (t *segmentTelemeter) Identify(props map[string]any) {
+func (t *segmentTelemeter) Identify(props map[string]any, opts ...telemeter.Option) {
 	if t == nil {
 		return
 	}
+
+	options := telemeter.ApplyOptions(opts)
+
 	traits := segment.NewTraits()
 
 	identity := segment.Identify{
-		UserId:  t.userID,
+		UserId:  t.overrideUserID(options),
 		Traits:  traits,
-		Context: t.ctx,
+		Context: makeDeviceContext(options),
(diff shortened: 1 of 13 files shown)


pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -107,20 +103,24 @@ func makeDeviceContext(clientID, clientType string) *segment.Context {
...
 }
 
-func (t *segmentTelemeter) Identify(props map[string]any) {
-	t.IdentifyUserAs("", "", "", props)
+func (t *segmentTelemeter) User(userID string) telemeter.Telemeter {
+	return &segmentTelemeter{client: t.client, userID: userID, ctx: t.ctx}
+}
+
+func (t *segmentTelemeter) As(clientID string, clientType string) telemeter.Telemeter {
+	return &segmentTelemeter{client: t.client, userID: t.userID, ctx: makeDeviceContext(clientID, clientType)}
```
dhaus67, line 111: I don't think this is how you would typically handle such things in a "go native" way.
Realistically, this could be something like a "builder" of some sorts instead, but I don't think we want to go this route.

What I originally had in mind would be an option pattern instead, which is somewhat similar to what you have shown:
```go
type options struct {
  clientID string
  clientType string
  userID string
  groupID string
}

func (o *option) apply(opts []Option) {
  for _ opt := range opts {
    opt(o)
  }
}

type Option func(*option)

func WithUserID(userID string) Option {
  return func(o *options) {
    o.userID
  }
}
...

func (t *segmentTelemeter) Identify(props map[string]any, opts ...Option) {
  o := &options{}
  o.apply(opts)
  identity := segment.Identify{
    UserId: o.userID,
    ...
  }
}
```

Then, callers would use the API as such, based on their needs:
```go
cfg.Telemeter().Identify(props, telemeter.WithUserID(id), telemeter.WithClientType("Secured Cluster"))
```

dhaus67 (changes requested):

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: I understand that Options might be a better recognized pattern. I don't find it elegant in this case, though. Please take a look at the update.

pkg/telemetry/phonehome/telemeter.go
```diff
@@ -1,34 +1,14 @@
 package phonehome
 
-// Telemeter defines a common interface for telemetry gatherers.
```
dhaus67, line 3: nitpick: I think this change could, in theory, be reverted now, right?

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -6,15 +6,18 @@ import (
...
 var (
-	log = logging.LoggerForModule()
+	log                     = logging.LoggerForModule()
+	_   telemeter.Telemeter = (*segmentTelemeter)(nil)
 )
 
 type segmentTelemeter struct {
 	client segment.Client
 	userID string
+	ctx    *segment.Context
```
dhaus67, line 20: I don't see us setting this value, so let's remove it as a property.

dhaus67 (approved):

pkg/telemetry/phonehome/telemeter.go
0x656b694d: Do you mean extracting the interface in a dedicated package? No, it cannot be reverted.

stehessel:
Looking at the `Telemeter().User(id).As(clientID, clientType).Track(...)` vs `Telemeter().Track(..., telemeter.WithUserID(id), telemeter.WithClient(clientID, clientType))` design choice, I don't have a strong preference either way. Sorry not much of a tie breaker 💀. In cases like this I usually opt to leave the choice to the author, in this case @0x656b694d.

0x656b694d:
Alright, as this question is not something we want to spend time on, I'll accept the Options approach, as it's already implemented, and has got a couple of dependent PRs.

stehessel (approved):

PR #4544, opened 2023-01-25
ROX-14240: Implement dropdown for Add Cluster options

vjwilson:
## Description

Implementation for Version 1 from this Jira comment,
https://issues.redhat.com/browse/ROX-13325?focusedCommentId=21420775&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21420775

Link used for docs from this comment,
https://issues.redhat.com/browse/ROX-13325?focusedCommentId=21608282&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21608282

Note: I tried re-styling the PF toggle to look more like the classic StackRox buttons, but that gave it more of an "uncanny valley" look than it helped, so I left the default PatternFly style for everything.

(I tried to use the Menu PatternFly component, as having better semantics and functionality, but it turned out to have limitations, including not being able to open a link in a new tab. I filed an issue with PatternFly, .) 

## Checklist
- [x] Investigated and inspected CI test results (ui-e2e test failures are unrelated flakes)

## Testing Performed

Closed option list
Screen Shot 2023-01-25 at 2 13 20 PM

Open, ready to click on link
Screen Shot 2023-01-25 at 2 13 27 PM

Linked opened in new tab
Screen Shot 2023-01-25 at 2 13 32 PM

Click on second item acts like former button
Screen Shot 2023-01-25 at 2 13 52 PM

PR #4602, opened 2023-01-30
ROX-13538: Initialize segment analytics

bradr5:
## Description
This PR sets up segment api on the front end for cloud services.

A new endpoint has been implemented to retrieve the write key `/v1/telemetry/config`. The endpoint requires the `ROX_TELEMETRY_STORAGE_KEY_V1` environment variable to be set for successful execution. Upon successful execution, analytics is attached to the window object. In case of a failed call, analytics will not be setup at all which is why the type definition was overwritten in `react-app-env.ts`

## Checklist

## Testing Performed

Manual testing performed using successful and unsuccessful requests.

For success, `ROX_TELEMETRY_STORAGE_KEY_V1` was set using the write key found in segment which resulted in the segment service call returning a response matching the type definition:

```
{
    userId: string,
    endpoint: string,
    storageKeyV1: string   // used to initialize segment
}
```
Once segment was initialized all utility function calls found in `analytics.ts` successfully sent events to segment.

For unsuccessful, `ROX_TELEMETRY_STORAGE_KEY_V1` was cleared out which resulted in our api call to return a status 404. Any util function called here resulted in no data being sent to segment nor any errors/warning in the browser.


ui/apps/platform/src/reducers/telemetryConfig.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+const error = (state = null, action) => {
+    switch (action.type) {
+        case types.FETCH_TELEMETRY_CONFIG.REQUEST:
+        case types.FETCH_TELEMETRY_CONFIG.SUCCESS:
+            return null;
+
+        case types.FETCH_TELEMETRY_CONFIG.FAILURE:
+            return action.error;
```
dvail, line 53: If I'm not mistaken, it looks like the property that contains the error on the action is `payload` based on line 32 above.

ui/apps/platform/src/reducers/telemetryConfig.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-underscore-dangle */
```
dvail, line 1: Is this needed? If so, could it be a single line lint disable instead?

ui/apps/platform/src/Containers/Clusters/ClustersSidePanel.tsx
```diff
@@ -256,6 +257,7 @@ function ClustersSidePanel({ selectedClusterId, setSelectedClusterId }) {
                     */
                     // TODO After saveCluster returns response without normalize,
                     // something like the preceding commented lines should replace the following:
+                    analyticsTrack('New Cluster');
```
dvail, line 260: Nit: could the values being tracked be stored somewhere as a constant instead?

The very little I've done with analytics in the past always led to things like `'New Cluster'` and `'New cluster'` eventually being tracked separately as more strings got added throughout the tracking code.

vjwilson, line 260: This will need to be adapted once this PR is merged.
https://github.com/stackrox/stackrox/pull/4544

vjwilson (approved):

ui/apps/platform/src/reducers/telemetryConfig.js
bradr5: You're right. I did the same thing in another PR.. Updated

ui/apps/platform/src/reducers/telemetryConfig.js
bradr5: Nope not even needed, updated

ui/apps/platform/src/Containers/Clusters/ClustersSidePanel.tsx
bradr5: Yeah I don't like what feels like a random string either. I think a separate file makes more sense which would also allow us to create more of a standard just based on context clues. Not sure where that should live though.. was thinking it might be a good topic for tomorrows meeting but I'll brainstorm a bit and come up with a couple of options

ui/apps/platform/src/Containers/Clusters/ClustersPage.js
```diff
@@ -22,6 +23,10 @@ const ClustersPage = ({
         params: { clusterId: selectedClusterId },
     },
 }) => {
+    useEffect(() => {
+        analyticsPageVisit('Visit', 'Clusters');
+    }, []);
```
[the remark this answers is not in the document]
bradr5: I'll go ahead and make that change then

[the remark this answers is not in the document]
pedrottimark: We discussed and can follow up later the pro and con to redact:
* id values in pathname
* query string

ui/apps/platform/src/utils/analytics.ts
```diff
@@ -0,0 +1,11 @@
+export function analyticsIdentity(userId: string, traits = {}): void {
+    return window.analytics?.identify(userId, traits);
+}
+
+export function analyticsPageVisit(type: string, name: string, additionalProperties = {}): void {
+    return window.analytics?.page(type, name, additionalProperties);
```
pedrottimark, line 6: As we discussed on a call, is optional chaining needed for `.page.?(type, name, additionalProperties)` method call and so on, in case of function calls before the methods are added to `analytics` data structure?

ui/apps/platform/src/global/initializeAnalytics.js
```diff
@@ -0,0 +1,62 @@
...
+                n.parentNode.insertBefore(t, n);
+                analytics._loadOptions = e;
+            };
+            analytics._writeKey = writeKey;
+            analytics.SNIPPET_VERSION = '4.15.3';
+            analytics.load(writeKey);
+            analytics.page();
+        }
+    }
+}
```
pedrottimark, line 62: Not a blocker to merge: can you add a newline at the end of the file?

We can follow up to see why lint does not report it for .js file.

ui/apps/platform/src/types/telemetryConfigService.proto.ts
```diff
@@ -0,0 +1,5 @@
+export type TelemetryConfig = {
+    userId: string;
+    endpoint: string;
+    storageKeyV1: string;
+};
```
pedrottimark, line 5: Not a blocker to merge: maybe just as well to move the type into services/TelemetryConfigService.ts file, because the type is in a special place
https://github.com/stackrox/stackrox/commit/8be5abb209ef162ef5387d3908742775e7117e6e

different from the usual backend proto/storage folder, which has an unrelated type left over from the original telemetry service:
https://github.com/stackrox/stackrox/blob/master/proto/storage/telemetry.proto

ui/apps/platform/src/utils/analytics.ts
bradr5: From my understanding it shouldn't be a concern. I don't think analytics is set on the window until its been fully initialized

ui/apps/platform/src/global/initializeAnalytics.js
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable */
+
+// the code below is generated from segment api with the exception of the writeKey parameter
+// segment intends for the write key to be hardcoded but we are pulling it from the telemetry config service call and adding it using a parameter
+
+export function initializeSegment(writeKey) {
+    const analytics = (window.analytics = window.analytics || []);
```
pedrottimark, line 7: Related to the method call question below in analytics.ts file, is there a possible time interval between this assignment to `window.analytics` and addition of methods in line 45 below?

bradr5: Isn't this part happening synchronously though? So outside of this code the methods would either exist on the object or the object would be undefined. We're calling `window.analytics.identify` immediately after calling `initializeSegment` in `telemetryConfig.js` without issue

pedrottimark: Yes, thank you for thinking it through!

pedrottimark (approved):

PR #4605, opened 2023-01-30
Trim Whitespace on Collections Endpoints Names

keyallis:
## Description

See Title.

## Checklist

## Testing Performed

Added a new unit test on the service implementation

### DIFF
--- a/central/resourcecollection/service/service_impl.go
+++ b/central/resourcecollection/service/service_impl.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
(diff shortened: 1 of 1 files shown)


central/resourcecollection/service/service_impl.go
```diff
@@ -203,7 +204,8 @@ func (s *serviceImpl) UpdateCollection(ctx context.Context, request *v1.UpdateCo
 }
 
 func collectionRequestToCollection(ctx context.Context, request collectionRequest, id string) (*storage.ResourceCollection, error) {
-	if request.GetName() == "" {
+	collectionName := strings.TrimSpace(request.GetName())
```
dvail, line 207: Does this strip the leading/trailing whitespace before attempting to save the collection? Would it be more predictable for API users if this rejected the request with an error instead?

It is probably unlikely, but it seems possible this could break an automation if the name the user submits via script is different from the name that is stored in the database.

keyallis: The function is used by all endpoints as a generalized operation to convert a request object to a storage object. So it would trim leading/trailing whitespace before every operation (get, delete, update, create, dryrun)

md2119: I wouldn't worry about failures due to misconfiguration. The result of APIs is reproducible and not a one-off.

md2119 (approved):

PR #4607, opened 2023-01-30
Fix VM reporting to allow changing between weekly and monthly intervals

dvail:
## Description

This fixes a bug where a vm report is unable to be changed from "monthly" to "weekly" due to multiple mutually-exclusive fields being set on the JSON request object at the same time. The backend proto only supports one of the fields, and will default to using `daysOfMonth` if both it and `daysOfWeek` are set. If this happens when the intervalType is "WEEKLY", the BE will throw an error.

## Checklist

## Testing Performed

Create a VM report config and save it with a "Monthly" interval, setting any value for the days.
image

Reload this config in edit mode and update the interval to "Weekly", verifying that it saves correctly.
image
image

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportForm.tsx
+++ b/ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportForm.tsx
@@ -193,12 +193,15 @@ function VulnMgmtReportForm({
     }
 
     function onScheduledRepeatChange(_id, selection) {
-        // zero out the days selected list if changing interval type
-        if (selection !== values.schedule.intervalType) {
-            void setFieldValue('schedule.interval.days', []);
-        }
+        const intervalDayKey = selection === 'WEEKLY' ? 'daysOfWeek' : 'daysOfMonth';
+        const nextSchedule = {
+            hour: values.schedule.hour,
+            minute: values.schedule.minute,
+            intervalType: selection,
+            [intervalDayKey]: {},
+        };
 
-        void setFieldValue('schedule.intervalType', selection);
+        void setFieldValue('schedule', nextSchedule);
     }
 
     function onScheduledDaysChange(id, selection) {


dvail:
* master
  * **PR #4607** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4607" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4607?utm_source=stack-comment).

ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportForm.tsx
```diff
@@ -193,12 +193,15 @@ function VulnMgmtReportForm({
     }
 
     function onScheduledRepeatChange(_id, selection) {
-        // zero out the days selected list if changing interval type
-        if (selection !== values.schedule.intervalType) {
-            void setFieldValue('schedule.interval.days', []);
```
dvail, line 198: I'm guessing `schedule.interval` is a vestigial property since it doesn't exist on the `Schedule` type or the backend proto, and this line hasn't been edited in ~15 months.

pedrottimark (approved):

ui/apps/platform/src/Containers/VulnMgmt/Reports/VulnMgmtReportForm.tsx
```diff
@@ -193,12 +193,15 @@ function VulnMgmtReportForm({
...
-        }
+        const intervalDayKey = selection === 'WEEKLY' ? 'daysOfWeek' : 'daysOfMonth';
+        const nextSchedule = {
```
pedrottimark, line 197: What do you think are pro and con of follow up to constrain the `Schedule` type so TypeScript compiler can verify the correctness of the conditional logic to build the `nextSchedule` object?

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/types/report.proto.ts#L39-L45

dvail: Yeah absolutely. That is the route I started on but then backtracked in order to keep the change in this PR small. So I suppose the main con is that it would just be a larger change to review.

This section is tricky too, since the Formik `setFieldValue` API is `any`-typed, and most of the function calls throughout VM use implicit-any as well. I'd like to go through and get safer types written, but will likely wait until VM Report 2.0 to see how much code actually gets reused.

PR #4620, opened 2023-01-31
Dashrews/ROX-14599 ROX-14037 ROX-14950 incorporate Rocks to Postgres where Rocks stops at 3.74

dashrews78:
## Description

Made some changes to add some more data to migrate to drive some additional migrations.  So in `postgres_run.sh` we now start a short scale run to pump some data in before we upgrade to Postgres.  The rollback to RocksDB scenarios were removed from that file.  So `postgres_run.sh` goes RocksDB -> Postgres -> Postgres.  The rollback to RocksDB backed central were moved to the `legacy_to_postgres_run.sh`.  This helps separate some concerns and keeps the script files smaller.  Plus we can just delete that file altogether when the time comes.  The tests in there go from Rocks->Postgres->Rocks.  Going back to the sealed Rocks at 74. 

Beefing it up to run a little scale workload solved ROX-14037.  
Updating the helm code to not use path and the roxctl we want directly should solve ROX-14950.

## Checklist

## Testing Performed

it is a test


dashrews78:
ROX-14950

dashrews78:
ROX-14037

tests/upgrade/legacy_to_postgres_run.sh
```diff
@@ -0,0 +1,222 @@
...
+    "$TEST_ROOT/bin/$TEST_HOST_PLATFORM/roxctl" version
+    rm -rf sensor-remote
+    "$TEST_ROOT/bin/$TEST_HOST_PLATFORM/roxctl" -e "$API_ENDPOINT" -p "$ROX_PASSWORD" sensor get-bundle remote
+    [[ -d sensor-remote ]]
+
+    info "Installing sensor"
+    ./sensor-remote/sensor.sh
+    kubectl -n stackrox set image deploy/sensor "*=$REGISTRY/main:$LAST_ROCKS_TAG"
+    kubectl -n stackrox set image deploy/admission-control "*=$REGISTRY/main:$LAST_ROCKS_TAG"
+    kubectl -n stackrox set image ds/collector "collector=$REGISTRY/collector:$(cat COLLECTOR_VERSION)" \
```
connorgorman, line 153: make collector-tag instead of cat version

tests/upgrade/legacy_to_postgres_run.sh
```diff
@@ -0,0 +1,222 @@
...
+
+helm_upgrade_to_latest_postgres() {
+    info "Helm upgrade to Postgres build ${CURRENT_TAG}"
+
+    # Need to push the flag to ci so that the collect scripts pull from
+    # Postgres and not Rocks
+    ci_export ROX_POSTGRES_DATASTORE "true"
+    export CLUSTER="remote"
+
+    pwd
```
connorgorman, line 179: why?

tests/upgrade/lib.sh
```diff
@@ -18,7 +18,7 @@ wait_for_central_reconciliation() {
         numDeployments="$(curl -sSk -u "admin:$ROX_PASSWORD" "https://$API_ENDPOINT/v1/summary/counts" | jq '.numDeployments' -r)"
         echo "Try number ${i}. Number of deployments in Central: $numDeployments"
         [[ -n "$numDeployments" ]]
-        if [[ "$numDeployments" -lt 100 ]]; then
+        if [[ "$numDeployments" -lt 150 ]]; then
```
connorgorman, line 21: interesting, why this change?

connorgorman (commented):
One question is should be also setting the central-db image where applicable? Maybe I missed it but I don't think we're bumping that version as well

tests/upgrade/legacy_to_postgres_run.sh
dashrews78: debug artifact i missed.

tests/upgrade/postgres_run.sh
```diff
@@ -282,61 +241,49 @@ helm_upgrade_to_postgres() {
...
 
-    # use postgres
+    # Use postgres
     export ROX_POSTGRES_DATASTORE="true"
     # Need to push the flag to ci so that the collect scripts pull from
     # Postgres and not Rocks
     ci_export ROX_POSTGRES_DATASTORE "true"
     export CLUSTER="remote"
 
+    pwd
```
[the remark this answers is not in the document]
dashrews78: debug artifact.

dashrews78:
> One question is should be also setting the central-db image where applicable? Maybe I missed it but I don't think we're bumping that version as well

That is a good point, aside from the initial deployment of central-db we have never touched that in these tests.  I will take care of that.

connorgorman (approved):

PR #4622, opened 2023-01-31
Run 'podman run' in user namespace

0x656b694d:
## Description

Podman runs containers as the current user, and uses the user namespace by default.
This PR removes user substitution (`--user ...`) for podman users, but keeps it for docker users.

## Checklist

## Testing Performed

`make image`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ TESTFLAGS=-race -p 4
 BASE_DIR=$(CURDIR)
 
 ifeq (,$(findstring podman,$(shell docker --version 2>/dev/null)))
+# Podman DTRT by running processes unprivileged in containers,
+# but it's UID mapping is more nuanced. Only set user for vanilla docker.
 DOCKER_USER=--user "$(shell id -u)"
 endif
 
@@ -374,6 +376,10 @@ build-volumes:
 	$(SILENT)mkdir -p $(CURDIR)/linux-gocache
 	$(SILENT)docker volume inspect $(GOPATH_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOPATH_VOLUME_NAME)
 	$(SILENT)docker volume inspect $(GOCACHE_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOCACHE_VOLUME_NAME)
+ifneq ($(DOCKER_USER),)
+	@echo "Restoring user's ownership of linux-gocache and go directories after previous runs which could set it to root..."
+	$(SILENT)docker run --rm $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) chown -R "$(shell id -u)" /linux-gocache /go
+endif
 
 .PHONY: main-builder-image
 main-builder-image: build-volumes
--- a/Makefile
+++ b/Makefile
@@ -375,7 +375,6 @@ build-volumes:
 	$(SILENT)mkdir -p $(CURDIR)/linux-gocache
 	$(SILENT)docker volume inspect $(GOPATH_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOPATH_VOLUME_NAME)
 	$(SILENT)docker volume inspect $(GOCACHE_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOCACHE_VOLUME_NAME)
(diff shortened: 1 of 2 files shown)


0x656b694d:
* master
  * **PR #4622** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4622" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4622?utm_source=stack-comment).

Makefile
```diff
@@ -370,7 +377,6 @@ build-volumes:
 	$(SILENT)mkdir -p $(CURDIR)/linux-gocache
 	$(SILENT)docker volume inspect $(GOPATH_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOPATH_VOLUME_NAME)
 	$(SILENT)docker volume inspect $(GOCACHE_VOLUME_NAME) >/dev/null 2>&1 || docker volume create $(GOCACHE_VOLUME_NAME)
-	$(SILENT)docker run --rm $(LOCAL_VOLUME_ARGS) $(BUILD_IMAGE) chown -R "$(shell id -u)" /linux-gocache /go
```
porridge, line 373: This command was here to allow transparent change from root-owned directories (which would typically be the case for anyone who ever ran a dockerized build before `--user ...` flags were added) to regular-user-owned directories. And AFAICT this works for non-podman users.

If this does not work for podman for some reason, and some manual intervention is required, then perhaps you can make this line conditional to not break the user experience for non-podman users? :thinking:

0x656b694d:
If no user options provided to `podman run`, everything works as expected:
```sh
$ podman run -v $PWD:/tmp  --rm quay.io/stackrox-io/apollo-ci:stackrox-build-0.3.55 touch /tmp/abc
$ ls -la abc
-rw-r--r-- 1 mipetrov mipetrov 0 Feb  6 14:31 abc
```
Similarly, there's no problem with git.

Makefile
0x656b694d: @porridge , I reverted this change and added `$(DOCKER_USER)` there, and to a line in `make/protogen.mk`.

0x656b694d: update: `chown` doesn't work for me actually, I'll need to do it conditionally.
```
chown: changing ownership of '/go/pkg/mod/gorm.io/driver/postgres@v1.3.7': Operation not permitted
```

0x656b694d: I'm not sure that the idea of changing ownership of users' GOPATH is a good one. If we corrupted it once, let the user fix it properly.
I suggest removing this `chown`.

Makefile
```diff
@@ -4,6 +4,11 @@ ROX_PROJECT=apollo
 TESTFLAGS=-race -p 4
 BASE_DIR=$(CURDIR)
 
+DOCKER_VERSION=$(shell docker --version)
```
porridge, line 7:
```suggestion
DOCKER_VERSION=$(shell docker --version 2>/dev/null)
```
To hide the errors that you found when `make` is running inside the container.

Makefile
porridge: I also felt a bit uneasy about it, but it seems to have done the trick for everyone so far except for `podman` users and those who happened to run `make` using `sudo`.
How about we make it conditional on `!(,$(findstring podman,$(DOCKER_VERSION)))` for now, and delete in a couple of weeks, once the ownership is fixed for vast majority of engineers?

Makefile
```diff
@@ -4,6 +4,11 @@ ROX_PROJECT=apollo
...
+ifeq (,$(findstring podman,$(DOCKER_VERSION)))
+DOCKER_USER=--user "$(shell id -u)"
```
porridge, line 9:
```suggestion
ifeq (,$(findstring podman,$(DOCKER_VERSION)))
# Podman DTRT by running processes unprivileged in containers, but it's UID mapping is more nuanced. Only set user for vanilla docker.
DOCKER_USER=--user "$(shell id -u)"
```

porridge (changes requested):

Makefile
0x656b694d: Not a big fan of it, I'd prefer engineers to suffer. I added an echo nevertheless.

porridge (approved):
Please update PR description to match.

PR #4624, opened 2023-01-31
Fix race crash when refreshing page, visiting link

vjwilson:
## Description

Building the network flows for the External Entities node, after the user refreshes the page or visits a link with that node selected, results in a race condition where we tried to build the flows for the sidebar _before_ the node with the data was in the topology graph.

This solves it by short-circuiting the flows-building until the node is available.

## Checklist

## Testing Performed

Navigating manually
Screen Shot 2023-01-31 at 11 39 13 AM

Same state, from that URL
Screen Shot 2023-01-31 at 11 39 27 AM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
+++ b/ui/apps/platform/src/Containers/NetworkGraph/utils/flowUtils.ts
@@ -127,44 +127,47 @@ export function getNetworkFlows(
         const sourceNode = controller.getNodeById(edge.source);
         const targetNode = controller.getNodeById(edge.target);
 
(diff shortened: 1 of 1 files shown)


sachaudh:
@vjwilson can you double-check against my PR https://github.com/stackrox/stackrox/pull/4617 to see if that fixes the issue you're trying to fix? I thought I was assigned that bug fix unless this is different

vjwilson:
Closing, because a fix for a different bug report also fixes this bug.

@sachaudh 's PR https://github.com/stackrox/stackrox/pull/4617/files#diff-9756ee85cbf46d97418de998612f74694117ed9419f853f0d737fcb5e4d0c183

PR #4625, opened 2023-01-31
ROX-13754: Trim fields in Policy Wizard form

pedrottimark:
## Description

Although I thought about trim in **pre** or **post** processing, it might cause subtle misunderstandings, especially related to effect of `value` property on dry run results:
* For trim in **pre** processing, what if changes are not saved after verifying preview?
* For trim in **post** processing, there might be changes after verifying preview.

Instead, trim only on change:
* Prevent leading or trailing spaces in fields during create of new policy.
* Remove leading or trailing spaces in changed fields during clone or edit of existing policy.

1. Policy details
    * `name` deferred see **Residue** 3
    * `description`
    * `rationale`
    * `guidance`
2. Policy behavior
3. Policy criteria
    * `sectionName`
    * `value` deferred see **Residue** 4
4. Policy scope
    * `cluster`
    * `name` of deployment
    * `key`
    * `value`
    * `namespace`
5. Review policy

### Residue

1. Investigate annoying interaction for trim space at beginning of text input moves insertion point to the end.
2. Investigate console error for change from uncontrolled to controlled element when `value` prop was undefined.
3. Think more carefully about policy name. For example, trim seems safe for clone or edit.
4. Policy criteria `value` deserves extra special care to make sure it does not regress.

## Checklist

## Testing Performed

See presence or absence of spaces in payload of /v1/policies/dryrunjob request at Step 5

### DIFF
--- a/ui/apps/platform/src/Containers/Policies/Wizard/Step1/PolicyMetadataFormSection.tsx
+++ b/ui/apps/platform/src/Containers/Policies/Wizard/Step1/PolicyMetadataFormSection.tsx
@@ -5,11 +5,21 @@ import { Field, useFormikContext } from 'formik';
 import PolicyCategoriesSelectField from './PolicyCategoriesSelectField';
 
 function PolicyMetadataFormSection(): ReactElement {
-    const { handleChange } = useFormikContext();
+    const { handleChange, setFieldValue } = useFormikContext();
 
     function onChange(_value, event) {
         handleChange(event);
     }
+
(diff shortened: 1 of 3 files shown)


pedrottimark:
Aha, besides **Residue** 1 annoying interaction, onChange trims a trailing space that is intended to be internal:

For example, integration test fails because **NewSection** instead of intended **New Section**

Now I remember this is why we backed it out. Maybe the time to trim is just before dryrun request.

PR #4627, opened 2023-01-31
ROX-13916: Using Search and highlighted fields in the v1/alerts/summary/counts API for better scaled performance and to avoid OOM

clickboo:
## Checklist

## Testing Performed

TODO(replace-me)
Automated + manual on cluster. Scale tests TBD.

### DIFF
--- a/central/alert/service/service_impl.go
+++ b/central/alert/service/service_impl.go
@@ -181,50 +181,46 @@ func (s *serviceImpl) GetAlertsCounts(ctx context.Context, request *v1.GetAlerts
 	}
 
 	request.Request = ensureAllAlertsAreFetched(request.GetRequest())
-	var requestQ *v1.Query
-	if request.GetRequest().GetQuery() == "" {
-		requestQ = search.NewQueryBuilder().AddStringsHighlighted(search.Cluster, search.WildcardString).ProtoQuery()
-	} else {
-		var err error
-		requestQ, err = search.ParseQuery(request.GetRequest().GetQuery())
-		if err != nil {
-			return nil, err
-		}
-		var hasClusterQ, hasSeverityQ, hasCategoryQ bool
-		search.ApplyFnToAllBaseQueries(requestQ, func(bq *v1.BaseQuery) {
-			matchFieldQuery, ok := bq.GetQuery().(*v1.BaseQuery_MatchFieldQuery)
-			if !ok {
-				return
-			}
-
-			if matchFieldQuery.MatchFieldQuery.GetField() == search.Cluster.String() {
-				hasClusterQ = true
-				matchFieldQuery.MatchFieldQuery.Highlight = true
-			}
-			if matchFieldQuery.MatchFieldQuery.GetField() == search.Category.String() {
-				hasCategoryQ = true
-				matchFieldQuery.MatchFieldQuery.Highlight = true
-			}
-			if matchFieldQuery.MatchFieldQuery.GetField() == search.Severity.String() {
-				hasSeverityQ = true
-				matchFieldQuery.MatchFieldQuery.Highlight = true
-			}
-		})
-
-		var conjuncts []*v1.Query
-		if !hasClusterQ {
-			conjuncts = append(conjuncts, search.NewQueryBuilder().AddStringsHighlighted(search.Cluster, search.WildcardString).ProtoQuery())
-		}
-		if !hasSeverityQ {
-			conjuncts = append(conjuncts, search.NewQueryBuilder().AddStringsHighlighted(search.Severity, search.WildcardString).ProtoQuery())
-		}
-		if !hasCategoryQ {
-			conjuncts = append(conjuncts, search.NewQueryBuilder().AddStringsHighlighted(search.Category, search.WildcardString).ProtoQuery())
-		}
-		for _, conjunct := range conjuncts {
-			requestQ = search.ConjunctionQuery(requestQ, conjunct)
-		}
+	requestQ, err := search.ParseQuery(request.GetRequest().GetQuery(), search.MatchAllIfEmpty())
+	if err != nil {
+		return nil, err
 	}
+
+	var hasClusterQ, hasSeverityQ, hasCategoryQ bool
+	search.ApplyFnToAllBaseQueries(requestQ, func(bq *v1.BaseQuery) {
+		matchFieldQuery, ok := bq.GetQuery().(*v1.BaseQuery_MatchFieldQuery)
+		if !ok {
+			return
+		}
+
+		if matchFieldQuery.MatchFieldQuery.GetField() == search.Cluster.String() {
+			hasClusterQ = true
+			matchFieldQuery.MatchFieldQuery.Highlight = true
+		}
+		if matchFieldQuery.MatchFieldQuery.GetField() == search.Category.String() {
+			hasCategoryQ = true
+			matchFieldQuery.MatchFieldQuery.Highlight = true
+		}
+		if matchFieldQuery.MatchFieldQuery.GetField() == search.Severity.String() {
+			hasSeverityQ = true
+			matchFieldQuery.MatchFieldQuery.Highlight = true
+		}
+	})
+
+	var conjuncts []*v1.Query
+	if !hasClusterQ {
+		conjuncts = append(conjuncts, search.NewQueryBuilder().AddStringsHighlighted(search.Cluster, search.WildcardString).ProtoQuery())
+	}
+	if !hasSeverityQ {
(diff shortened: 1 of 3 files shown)


central/alert/service/service_impl.go
```diff
@@ -172,8 +172,29 @@ func (s *serviceImpl) GetAlertsCounts(ctx context.Context, request *v1.GetAlerts
 	if request == nil {
 		request = &v1.GetAlertsCountsRequest{}
 	}
+
+	// TODO: @ksanchet - check correctness. This endpoint could possibly be invoked with any query.
+	// For example, alerts within a specific cluster
+	// in that case, will the code below work? The UI only sends in cluster and or namespaces but
+	// an API client can send in any query
```
md2119, line 179: If there is a cluster, category, or severity input query then you must transform the respective query from `AddStrings` to `AddStringsHighlighted`. If not, add the wildcard.

central/alert/service/service_impl.go
```diff
@@ -172,8 +172,29 @@ func (s *serviceImpl) GetAlertsCounts(ctx context.Context, request *v1.GetAlerts
...
+			return nil, err
+		}
+	}
+
+	highlightedQ := search.NewQueryBuilder().AddStringsHighlighted(search.Cluster, search.WildcardString).
+		AddStringsHighlighted(search.Category, search.WildcardString).
+		AddStringsHighlighted(search.Severity, search.WildcardString).
+		ProtoQuery()
+
+	alerts, err := s.dataStore.SearchListAlerts(ctx, search.ConjunctionQuery(requestQ, highlightedQ))
```
md2119, line 197: use `Search` that returns `[]search.Result` and collect cluster, severity, and category from the `Matches` field.

central/alert/service/service_impl.go
clickboo: I had uncommitted changes I just pushed doing exactly that :-) ...check latest

central/alert/service/service_impl.go
```diff
@@ -62,14 +63,14 @@ var (
...
-	// groupByFunctions provides a map of functions that group slices of ListAlet objects by category or by cluser.
-	groupByFunctions = map[v1.GetAlertsCountsRequest_RequestGroup]func(*storage.ListAlert) []string{
-		v1.GetAlertsCountsRequest_UNSET: func(*storage.ListAlert) []string { return []string{""} },
-		v1.GetAlertsCountsRequest_CATEGORY: func(a *storage.ListAlert) (output []string) {
-			output = append(output, a.GetPolicy().GetCategories()...)
+	// groupByFunctions provides a map of functions that group slices of result objects by category or by cluster.
+	groupByFunctions = map[v1.GetAlertsCountsRequest_RequestGroup]func(result search.Result) []string{
+		v1.GetAlertsCountsRequest_UNSET: func(result search.Result) []string { return []string{""} },
+		v1.GetAlertsCountsRequest_CATEGORY: func(a search.Result) (output []string) {
+			output = append(output, a.Matches[search.Category.String()]...)
```
md2119, line 70: The key in `Matches` is the field path in the object like `.policy.name` and sth.

You can get the field paths from the list alerts options map as such,
```
field, ok := {alertMappings}.OptionsMap.Get(search.Category.String())
fieldPath := field.GetFieldPath()
```

central/alert/service/service_impl.go
```diff
@@ -172,8 +173,35 @@ func (s *serviceImpl) GetAlertsCounts(ctx context.Context, request *v1.GetAlerts
...
+		if err != nil {
+			return nil, err
+		}
+		search.ApplyFnToAllBaseQueries(q, func(bq *v1.BaseQuery) {
+			matchFieldQuery, ok := bq.GetQuery().(*v1.BaseQuery_MatchFieldQuery)
+			if !ok {
+				return
+			}
+
+			if matchFieldQuery.MatchFieldQuery.GetField() == search.Cluster.String() {
```
md2119, line 193: Is the endpoint limited to accepting cluster queries only? You should do the same for Category and Severity.

central/alert/service/service_impl.go
```diff
@@ -510,16 +538,19 @@ func countAlerts(alerts []*storage.ListAlert, groupByFunc func(*storage.ListAler
 	return
 }
 
-func getMapOfAlertCounts(alerts []*storage.ListAlert, groupByFunc func(alert *storage.ListAlert) []string) (groups map[string]map[storage.Severity]int) {
+func getMapOfAlertCounts(alerts []search.Result, groupByFunc func(alert search.Result) []string) (groups map[string]map[storage.Severity]int) {
 	groups = make(map[string]map[storage.Severity]int)
```
md2119, line 542: What is the key of the outermost map?

central/alert/service/service_impl.go
```diff
@@ -510,16 +538,19 @@ func countAlerts(alerts []*storage.ListAlert, groupByFunc func(*storage.ListAler
...
 		for _, g := range groupByFunc(a) {
 			if groups[g] == nil {
 				groups[g] = make(map[storage.Severity]int)
 			}
-
-			groups[g][a.GetPolicy().GetSeverity()]++
+			if len(a.Matches[search.Severity.String()]) == 0 {
+				continue
+			}
+			severity, _ := strconv.Atoi(a.Matches[search.Severity.String()][0])
```
md2119, line 552: same comment as above regarding the field path

central/alert/service/service_impl.go
clickboo: Good point, yes. The UI uses only cluster and namespace but anyone can invoke that API with any query. Done.

central/alert/service/service_impl.go
clickboo: the group (aka the category  name or cluster name  or "") - check the expected test results in the related test file.

central/alert/service/service_impl.go
clickboo: ??

central/alert/service/service_impl.go
clickboo: Got it

md2119 (changes requested):

central/alert/service/service_impl.go
```diff
@@ -510,16 +561,20 @@ func countAlerts(alerts []*storage.ListAlert, groupByFunc func(*storage.ListAler
...
 			if groups[g] == nil {
 				groups[g] = make(map[storage.Severity]int)
 			}
-
-			groups[g][a.GetPolicy().GetSeverity()]++
+			field, _ := mappings.OptionsMap.Get(search.Severity.String())
+			if len(a.Matches[field.GetFieldPath()]) == 0 {
+				continue
+			}
+			severity, _ := strconv.Atoi(a.Matches[field.GetFieldPath()][0])
```
md2119: discussed the fix on slack

md2119 (approved):
Minor comments on logging since we already came across unexpected behavior differences.

central/alert/service/service_impl.go
```diff
@@ -510,16 +561,20 @@ func countAlerts(alerts []*storage.ListAlert, groupByFunc func(*storage.ListAler
...
 	groups = make(map[string]map[storage.Severity]int)
 
 	for _, a := range alerts {
 		for _, g := range groupByFunc(a) {
 			if groups[g] == nil {
 				groups[g] = make(map[storage.Severity]int)
 			}
-
-			groups[g][a.GetPolicy().GetSeverity()]++
+			field, _ := mappings.OptionsMap.Get(search.Severity.String())
```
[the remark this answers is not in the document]
clickboo: actually can use MustGet instead. thoughts?

central/alert/service/service_impl.go
clickboo: That root cause though 😓

central/alert/service/service_impl.go
```diff
@@ -172,8 +179,53 @@ func (s *serviceImpl) GetAlertsCounts(ctx context.Context, request *v1.GetAlerts
 	if request == nil {
 		request = &v1.GetAlertsCountsRequest{}
 	}
+
 	request.Request = ensureAllAlertsAreFetched(request.GetRequest())
-	alerts, err := s.dataStore.ListAlerts(ctx, request.GetRequest())
+	var requestQ *v1.Query
+	if request.GetRequest().GetQuery() == "" {
+		requestQ = search.NewQueryBuilder().AddStringsHighlighted(search.Cluster, search.WildcardString).ProtoQuery()
+	} else {
```
md2119, line 187: Drop the if block altogether. When there is no query only cluster values are pulled out without severity and category.

PR #4632, opened 2023-01-31
ROX-13754: Trim fields in Policy Wizard form before dryrun and save

pedrottimark:
## Description

Trim in `getServerPolicy` function which is called:
* Before POST /v1/dryrunjub request in Review Policy step.
* Before POST or PUT /v1/policies request on click Save button.

### Residue

1. Possibly add frontend validation for `name` length and allowed characters.
2. Policy criteria `value` deserves extra special care to make sure it does not regress.
3. Policy scope `key` and `value` of label deserve extra special care to make sure it does not regress.

## Checklist

## Testing Performed

1. Temporary console.log to compare client and server policy.
2. See presence or absence of spaces:
    * in payload of POST /v1/policies/dryrunjob request at Step 5
    * in response of GET /v1/policies/id after save

### DIFF
--- a/ui/apps/platform/src/Containers/Policies/policies.utils.ts
+++ b/ui/apps/platform/src/Containers/Policies/policies.utils.ts
@@ -532,21 +532,21 @@ function trimPolicyScope(scope: PolicyScope) {
 }
 
 function trimClientWizardPolicy(policyUntrimmed: ClientPolicy): ClientPolicy {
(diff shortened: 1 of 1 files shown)


vjwilson (approved):

pedrottimark:
The test failure is persistent today, but unrelated to these changes, so time to merge.

PR #4635, opened 2023-01-31
ROX-14619: fix metadata retrieval errors in scan time test

RTann:
## Description

Update the DockerHub images to Quay.io images, as these tests may be failing due to DockerHub rate-limiting. The purpose of these tests is to ensure the "scan time" is not "null", so it really is not important which images are used.

## Checklist

## Testing Performed

CI

### DIFF
--- a/qa-tests-backend/src/test/groovy/ImageScanningTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ImageScanningTest.groovy
@@ -655,8 +655,8 @@ class ImageScanningTest extends BaseSpecification {
         where:
         image                                              | registry
         "registry.k8s.io/ip-masq-agent-amd64:v2.4.1"       | "gcr registry"
-        "docker.io/jenkins/jenkins:lts"                    | "docker registry"
-        "docker.io/jenkins/jenkins:2.220-alpine"           | "docker registry"
+        "quay.io/rhacs-eng/qa:alpine-3.16.0"               | "quay registry"
+        "quay.io/stackrox-io/scanner:2.27.3"               | "quay registry"
         "gke.gcr.io/heapster:v1.7.2"                       | "one from gke"
         "mcr.microsoft.com/dotnet/core/runtime:2.1-alpine" | "one from mcr"
     }


RTann:
Failure is unrelated, and I verified the test in question passes

qa-tests-backend/src/test/groovy/ImageScanningTest.groovy
```diff
@@ -655,8 +655,8 @@ class ImageScanningTest extends BaseSpecification {
         where:
         image                                              | registry
         "registry.k8s.io/ip-masq-agent-amd64:v2.4.1"       | "gcr registry"
-        "docker.io/jenkins/jenkins:lts"                    | "docker registry"
-        "docker.io/jenkins/jenkins:2.220-alpine"           | "docker registry"
+        "quay.io/rhacs-eng/qa:alpine-3.16.0"               | "quay registry"
+        "quay.io/stackrox-io/scanner:2.27.3"               | "quay registry"
```
jvdm, line 659: Why do we have two for the same registry (quay)?

gavin-stackrox (approved):

jvdm:
Do you what the original test is designed to capture? I mean, how the presence of scan time depends on the type of registry being used or the image?

Also, could the rate-limiting from Docker Hub impacting us also impact customers? Or are we running this E2E multiple times in different places with the same creds? Do you know if we should follow up on how we are pulling/calling Docker Hub in Scanner to accommodate the rate-limiting errors we are experiencing here?

qa-tests-backend/src/test/groovy/ImageScanningTest.groovy
RTann: doesn't really matter much. I just used Quay because we haven't had problems with it before

RTann:
> Do you know what the original test is designed to capture? I mean, how the presence of scan time depends on the type of registry being used or the image?

Original tests just want to make sure we populate `scanTime` which is actually done via Central, so neither the image nor registry matters

> Also, could the rate-limiting from Docker Hub impacting us also impact customers? Or are we running this E2E multiple times in different places with the same creds? Do you know if we should follow up on how we are pulling/calling Docker Hub in Scanner to accommodate the rate-limiting errors we are experiencing here?

David mentioned it is possible that DockerHub is rate limiting Red Hat IPs in general. It's just a thought. Not 100% sure why this is happening, but that's just our running theory. These images are public, and I don't think we are using any specific creds. Either way, we know more about Quay since we are part of Red Hat, so it is probably best to use it over DockerHub anyway

dcaravel:
> David mentioned it is possible that DockerHub is rate limiting Red Hat IPs in general. It's just a thought. Not 100% sure why this is happening, but that's just our running theory. 

FWIW, I did some searching after our convo and looks like DockerHub responds with the appropriate status code (`429 Too Many Requests`) when rate limiting - so my theory is false - must be something else

tommartensen:
FYI we added this PR to the 3.73.2-rc.3 milestone.

PR #4650, opened 2023-02-01
Assign operator to stackrox/draco

janisz:


kovayur (approved):

porridge:
I'm having second thoughts about this change. Apparently it causes the whole draco team to be CCed on every PR that touches the operator directory. Apparently even explicitly requesting review from a selected draco member does not prevent that. I'm worried that it creates too much noise :-(
Do you know any way to control that behaviour @janisz ?

PR #4651, opened 2023-02-01
ROX-14657 Add inline errors for collection form fields

dvail:
## Description

Note: this is a low risk nice-to-have for 3.74 release, but also relatively low priority.

Adds inline error message to collection form fields that give more information on why the field is in an error state. This wraps existing TextInputs with FormGroups in order to take advantage of the inline error messages provided by the `helperTextInvalid` prop.

Also includes another minor UI fix for the VM report page. The severity labels were all set to red, this updates them to be the correct colors used elsewhere in the app.

## Checklist

## Testing Performed

On the collection form, add a "by name" rule and leave the value empty:
image

Enter a value, and then add another rule line. The second line should be the one containing the error. The error should also apply if the field is set to a regex value.
image

Entering a correct value causes the error to go away:
image

Repeat for label values, adding multiple rules with multiple values. The by label field should give a separate error message for both empty values and invalid label values.
image

---

On the VM reporting page, ensure that the severity labels in the table have the correct color.
image

Same on the VM report detail page:
image

### DIFF
--- a/ui/apps/platform/src/Components/PatternFly/SeverityLabelsList.tsx
+++ b/ui/apps/platform/src/Components/PatternFly/SeverityLabelsList.tsx
@@ -8,10 +8,21 @@ export type SeverityLabelsListProps = {
     severities: VulnerabilitySeverity[];
 };
 
+const vulnerabilitySeverityLabelColors = {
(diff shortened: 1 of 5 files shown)


pedrottimark (approved):
Super attention to detail.

PR #4654, opened 2023-02-01
Update front end check for regex rule message

dvail:
## Description

Updates the FE check to match the new BE error message.

Additionally adds `<` and `>` as supported characters for collection names, as requested by @charmik-redhat due to these characters existing in migrated access scopes.

## Checklist


keyallis (approved):
LGTM

PR #3230, opened 2022-09-28
ROX-12856: roxctl: Make sure that PSPs are not rendered for deployment bundles

mtesseract:
## Description

See https://issues.redhat.com/browse/ROX-12856.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #4655, opened 2023-02-01
ROX-14485: update monitoring psp

Maddosaurus:
## Description
Kubernetes 1.25 deprecated PSP (PodSecurityPolicies) altogether.
We had previously introduced switches to decide whether PSPs are generated for our deployments (see https://github.com/stackrox/stackrox/pull/3230 and https://github.com/stackrox/stackrox/pull/1747).
We did not introduce this change to our Monitoring setup, which this PR fixes.

I introduced a new env var, `MONITORING_ENABLE_PSP`, which decides whether PSPs are generated or not.
The env var defaults to `false` and needs to be explicitly enabled (e.g. for k8s < 1.25).
The ClusterRole was kept, as it is needed for Prometheus discovery features to work ([more information](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/rbac.md#prometheus-rbac)). I only removed the PSP-relevant parts.

## Checklist

Most of the PR items don't apply as the monitoring charts are internal eng use only.

## Testing Performed
Tested on Colima 1.25 and 1.24 locally, as well as an OpenShift 4.12 cluster.
In all cases, I deployed with and without generated PSPs and checked function of Prometheus.


vikin91 (commented):
The change looks good, but I am afraid that we will need to provide manual upgrade steps if this is used by customers.

vikin91 (approved):
Updating based on the answer from the standup. Thanks!

PR #4656, opened 2023-02-01
FOX-14426: Fix CVE exports in Findings sections

vjwilson:
## Description

Earlier work to vary the CSV export of CVEs by type for Postgres missed passing down the type to the export functionality fo the Fixable CVE tables in the Findings sections of individual entities.

1. One change passes that CVE type down when Postgres is enabled.
2. Another change adds a fallback when Postgres is not enabled, so that the best approximate CVE_Type filter is added for an individual entity when exporting. (Note, the non-Postgres API does not support multiple filter values, so Cluster export will just get image CVEs. Asking for an API improvement for a datastore that will be removed in 9 weeks is not worth the effort.)

## Checklist
- [x] Investigated and inspected CI test results (ui-e2e test failures are unrelated flakes)

## Testing Performed

Manually testing on:

1. Postgres (stagingdb)
2. Non-Postgres (staging)

### DIFF
--- a/ui/apps/platform/src/services/VulnerabilitiesService.js
+++ b/ui/apps/platform/src/services/VulnerabilitiesService.js
@@ -93,8 +93,20 @@ export function exportCvesAsCsv(fileName, workflowState, cveType) {
         return { ...{ [key]: fullEntityContext[key] } };
     }, {});
 
+    const currentSearchState = workflowState.getCurrentSearchState();
+
+    // TODO: remove after Postgres is required for all installations
+    // kludge to make Findings sections CSV export of CVEs work until Postgres is on
+    if (cveType === entityTypes.CVE && !currentSearchState['CVE Type']) {
+        if (fullEntityContext.NODE) {
+            currentSearchState['CVE Type'] = 'NODE_CVE';
+        } else {
+            currentSearchState['CVE Type'] = 'IMAGE_CVE';
+        }
+    }
+
     const query = queryService.objectToWhereClause({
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/services/VulnerabilitiesService.js
```diff
@@ -93,8 +93,20 @@ export function exportCvesAsCsv(fileName, workflowState, cveType) {
         return { ...{ [key]: fullEntityContext[key] } };
     }, {});
 
+    const currentSearchState = workflowState.getCurrentSearchState();
+
+    // TODO: remove after Postgres is required for all installations
```
sachaudh, line 98: I have one concern about our TODOs. We add them but I know sometimes we can forget about them later on. My suggestion is that we should couple our TODOs with an actual Jira ticket, under whatever Epic is appropriate, and track clean-up work as well. That way we don't accumulate tech debt

That also makes me think, maybe we should go through our TODOs sometime and try to track some clean up or updates that we might have forgotten about

cc: @vjwilson @alwayshooin @pedrottimark @dvail @bradr5

sachaudh (approved):
LGTM

ui/apps/platform/src/services/VulnerabilitiesService.js
vjwilson: Good ideas.

I this case, I intentionally fashioned the conditional so that it won't run after the transition to Postgres.

PR #4659, opened 2023-02-01
ROX-14620: Fix migrations referring to current schema and add import lint

c-du:
## Description
Add import link analysis for migrator/migrations to ensure migrations do not contain reference to current schema.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


tools/roxvet/analyzers/validateimports/analyzer.go
```diff
@@ -19,27 +20,27 @@ const doc = `check that imports are valid`
 const roxPrefix = "github.com/stackrox/rox/"
 
 var (
-	validRoots = []string{
```
rhybrillou, line 22: Would it make sense to add a comment here requesting the set items to be kept in alphabetic order ?

tools/roxvet/analyzers/validateimports/analyzer.go
```diff
@@ -177,14 +179,15 @@ func checkForbidden(impPath, packageName string) error {
 // only import StackRox code from allowedPackages
 func verifyImportsFromAllowedPackagesOnly(pass *analysis.Pass, imports []*ast.ImportSpec, validImportRoot, packageName string) {
 	allowedPackages := []string{validImportRoot, "generated", "image"}
-	// The migrator is NOT allowed to import all code from pkg except process/id as that pkg is isolated.
-	if validImportRoot != "pkg" && validImportRoot != "migrator" {
+	// The migrator is NOT allowed to import all code from pkg except packages isolated.
```
rhybrillou, line 182: nit: `except isolated packages (list comes below).`

tools/roxvet/analyzers/validateimports/analyzer.go
```diff
@@ -241,6 +243,16 @@ func verifyImportsFromAllowedPackagesOnly(pass *analysis.Pass, imports []*ast.Im
 			"pkg/uuid",
 			"pkg/version",
 		)
+
+		// Migrations shall not depend on current schemas. Each migration can include a copy of the schema before and
+		// after a specific migration.
+		if validImportRoot == "migrator" {
```
rhybrillou, line 249: This exception is there for the migrator to:
- read schema version from an existing postgres database
- apply the schema for initial deployment.

Migrations themselves MUST depend on frozen schemas (possibly before and after each specific migration).

rhybrillou (approved):
Great work. Thank you.
Just minor questions and comments.

tools/roxvet/analyzers/validateimports/analyzer.go
c-du: Yes.

tools/roxvet/analyzers/validateimports/analyzer.go
```diff
@@ -19,27 +20,27 @@ const doc = `check that imports are valid`
...
 		"integration-tests",
 		"migrator",
+		"migrator/migrations",
+		"operator",
 		"pkg",
 		"roxctl",
 		"scale",
-		"sensor/common",
-		"sensor/kubernetes",
-		"sensor/kubernetes/sensor",
```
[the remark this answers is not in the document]
c-du: Yes. Before logic change, "sensor/kubernetes/sensor" is not in actually in use.

migrator/version/convert_versions.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by pg-bindings generator. DO NOT EDIT.
```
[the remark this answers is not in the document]
c-du: It was generated but I think it should evolve by itself from now on. Thanks!

connorgorman (approved):

migrator/migrations/frozenschema/v74/permission_sets.go
```diff
@@ -0,0 +1,36 @@
+// Code generated by pg-bindings generator. DO NOT EDIT.
```
dashrews78, line 1: So I have a fundamental question with this one, Probably for outside the scope of the PR.  The schema did not change from v73.  So why do we need a copy of it in v74?  I think we only need a frozen schema IF the schema changes for the release.  For migrations that do not change the schema, the previous migrations would have been responsible to get the database into the proper schema state and thus a schema that does not change the schema should not try to apply it.  But like I said, just dropping a thought that we should probably take outside the PR.

c-du: Yes, a migration that does not change schema do not need to apply it but it may still need to use it. This is to make migrations independent of each other for the clarity of the codes. In most cases, we do not have the comprehensive knowledge which schema is changed and which is not. One thing we can do but not yet is to have a version per table for each schema change.

PR #4662, opened 2023-02-01
Use go1.19 memory management changes to GC

connorgorman:
## Description

Follow up to moving to go1.19 (to be shipped after 74 in order to ensure no adverse effects)
https://tip.golang.org/doc/gc-guide

## Checklist

## Testing Performed

Will start running on scale tests

### DIFF
--- a/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
+++ b/image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
@@ -92,6 +92,14 @@ spec:
             drop: ["NET_RAW"]
           readOnlyRootFilesystem: true
         env:
+        - name: GOMEMLIMIT
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.memory
+        - name: GOMAXPROCS
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.cpu
         - name: ROX_NAMESPACE
           valueFrom:
             fieldRef:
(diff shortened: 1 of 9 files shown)


image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -92,6 +92,14 @@ spec:
             drop: ["NET_RAW"]
           readOnlyRootFilesystem: true
         env:
+        - name: GOMEMLIMIT
```
janisz, line 95: Is rocksdb included in this limit?

connorgorman: RocksDB will no longer be in the image for 4.0 so this will not be as important

connorgorman: but also in general, this is only for memory that go knows about so it won't be as effective on rocksdb

janisz (approved):

image/templates/helm/stackrox-central/templates/01-central-13-deployment.yaml
```diff
@@ -92,6 +92,14 @@ spec:
...
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.memory
+        - name: GOMAXPROCS
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.cpu
```
janisz, line 102: Is this always integer?

connorgorman:
I'm not sure the go max procs change is actually helping us

PR #4668, opened 2023-02-01
Network graph: PF - adding reset graph callback for smooth edge state change

alwayshooin:
the current experience causes the layout to be merged oddly on edge state change:

Before:
* active state:
* image
* changing to extraneous state:
* image

After:
* active state:
* image
* changing to extraneous state:
* image

everything else should work as expected (layout on init, refresh, display options toggling, selected node toggling) -- the layout should _not_ change on any of these and _only_ on edge state change

i also added clearing the URL when the user has a node selected that only exists in the `extraneous` graph and then toggles to the `active` graph


vjwilson (approved):

sachaudh (approved):
LGTM

vjwilson (approved):

PR #4679, opened 2023-02-02
ROX-14438: Refactor Sensor Event component messages

fredrb:
## Description

This PR removes the helper methods previously added used to create and merge instances of  `component.ResourceEvent`. It also changes the component event messages to use a slice of deployment references rather than a single deployment entry. This allows for pods with multiple owners to trigger the processing of multiple deployments.

## Checklist

## Testing Performed

- CI


sensor/kubernetes/listener/resources/deployments.go
```diff
@@ -153,18 +159,28 @@ func (d *deploymentHandler) processWithType(obj, oldObj interface{}, action cent
...
-	}
 
-	if deploymentWrap == nil {
-		if objAsPod != nil {
+		if deploymentWrap == nil {
 			// It's only a pod event and the pod belongs to another resource (e.g. deployment)
-			events = component.MergeResourceEvents(events, d.maybeUpdateParentsOfPod(objAsPod, oldObj, action))
+			d.maybeUpdateParentsOfPod(objAsPod, oldObj, action, events)
+			if removeEvents != nil {
+				events.AppendMessage(removeEvents)
```
lvalerom, line 174: [nit] Would move this to line 167 (right after calling `processPodEvent`) make sense? It looks like if `removeEvents` is not `nil` we are appending the events always anyway. Unless `maybeUpdateParentsOfPod` should be called before appending, but I don't think the order matters here 🤔

lvalerom (commented):
I want to give it a second look before approving but it looks really good, the functions are way more intuitive to use!

lvalerom (approved):

PR #4681, opened 2023-02-02
Ensure Role and Access Scope migration to postgres and UUIDs also covers references from consumers

rhybrillou:
## Description

The tables used to store the roles and associated elements were designed to leverage UUIDs. The data migration to postgres ensures the object IDs for access scopes and permission sets are valid UUIDs and can be stored as UUIDs in the tables.

In the effort, one consumer of access scopes was overlooked: vulnerability report configurations. In order to ensure the vulnerability report configurations still rely on proper scopes, these get updated as the roles, permission sets and and access scopes get migrated to postgres.

## Checklist

## Testing Performed

CI should cover the most relevant part.

In order to ensure things work properly, manual migration tests will have to cover the creation of dedicated access scopes, as well as the creation of report configurations that reference both default and manually created scopes. The generated reports before and after the migration should point at the same data.

Manual testing performed
- spawn 3.73.0 demo stackrox instance
- create access scope limited to cluster production and namespace operations
- create vulnerability report with the above scope
- edit the central container image to the one built by this PR
- spawn a central-db instance for the deployed stackrox (`roxctl central db generate k8s pvc`)
- edit the central container environment to enable postgres mode (env `ROX_POSTGRES_DATASTORE` set to `true`)
- let central restart
- re-open the vulnerability report
- see that the report is configured for a collection created from the above scope


migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +255,52 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
 }
 
+func getReportConfigurationScopeID(reportConfiguration *storage.ReportConfiguration) (string, error) {
+	reportConfigurationScopeID := strings.TrimPrefix(reportConfiguration.GetScopeId(), accessScopeIDPrefix)
+	if replacement, found := accessScopeIDMapping[reportConfigurationScopeID]; found {
+		reportConfigurationScopeID = replacement
+	}
+	_, accessIDParseErr := uuid.FromString(reportConfigurationScopeID)
+	if accessIDParseErr != nil {
+		log.WriteToStderrf("failed to convert report configuration to postgres format, bad scope ID. Role [%s], error %v", reportConfiguration.GetName(), accessIDParseErr)
```
charmik-redhat, line 265: Report Configuration [%s] instead of Role [%s]

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -91,6 +92,9 @@ func migrateAll(rocksDatabase *rocksdb.RocksDB, gormDB *gorm.DB, postgresDB *pgx
 		return errors.Wrap(err,
 			"moving roles from rocksdb to postgres")
 	}
+	if err := updateReportConfigurationScopeIDs(postgresDB); err != nil {
```
dhaus67, line 95: super-nit: call it `migrate..` instead of `update..`.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
dhaus67: + would be nice to specifically print the `reportConfigurationScopeID` as well.
```suggestion
		log.WriteToStderrf("failed to convert report configuration to postgres format, bad scope ID. Report [%s], error %v", reportConfiguration.GetName(), accessIDParseErr)
```

rhybrillou: Nice catch. Thank you.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +255,52 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+	}
+	_, accessIDParseErr := uuid.FromString(reportConfigurationScopeID)
+	if accessIDParseErr != nil {
+		log.WriteToStderrf("failed to convert report configuration to postgres format, bad scope ID. Report Configuration [%s], error %v", reportConfiguration.GetName(), accessIDParseErr)
+		return "", accessIDParseErr
+	}
+	return reportConfigurationScopeID, nil
+}
+
+func updateReportConfigurationScopeIDs(postgresDB *pgxpool.Pool) error {
```
dhaus67, line 271: nit: would be nice to have a comment specifying that this is has been added at a later point (although, I don't think it's _that_ critical to have).

pkg/postgres/pgtest/postgres.go
```diff
@@ -127,6 +127,12 @@ func ForTCustomPool(t testing.TB, dbName string) *pgxpool.Pool {
 	return pool
 }
 
+// GetGormDB opens a Gorm DB to the Postgres DB
+func (tp *TestPostgres) GetGormDB(t testing.TB) *gorm.DB {
+	source := conn.GetConnectionStringWithDatabaseName(t, tp.database)
+	return OpenGormDB(t, source)
+}
+
```
dhaus67, line 135: I don't see this being used anywhere, why do we need it?

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +255,52 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+	ctx := sac.WithAllAccess(context.Background())
+	store := pgReportConfigurationStore.New(postgresDB)
+	var reportConfigs []*storage.ReportConfiguration
+	err := walkReportConfigurations(ctx, store, func(obj *storage.ReportConfiguration) error {
+		reportConfigurationScopeID, conversionErr := getReportConfigurationScopeID(obj)
+		if conversionErr != nil {
+			return conversionErr
+		}
```
dhaus67, line 279: Do we want to fail open vs fail closed here?
i.e. do we want to abort migration of _all_ report configuration because we have a single entry with an unexpected ID, or do we rather want to log the occurrence of it, and continue with migrating the other report configurations?

I know we have had a mix of those concepts in the rocksdb migrations, not sure how we have converged on the postgres migrations or its still more or less a case-by-case basis.

pkg/postgres/pgtest/postgres.go
rhybrillou: It is needed in the `access scope`/`permission set`/`role` migration test to create the `report_configurations` table, which is pre-populated then updated.
See the `SetupTest` function change.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
rhybrillou: Good idea. Now both ID and name are reported in the log.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +257,60 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+func getReportConfigurationScopeID(reportConfiguration *storage.ReportConfiguration) (string, error) {
+	reportConfigurationScopeID := strings.TrimPrefix(reportConfiguration.GetScopeId(), accessScopeIDPrefix)
+	if replacement, found := accessScopeIDMapping[reportConfigurationScopeID]; found {
+		reportConfigurationScopeID = replacement
+	}
+	_, accessIDParseErr := uuid.FromString(reportConfigurationScopeID)
+	if accessIDParseErr != nil {
+		log.WriteToStderrf("failed to convert report configuration to postgres format, bad scope ID. Report Configuration ID: [%s] Name:[%s], error %v", reportConfiguration.GetId(), reportConfiguration.GetName(), accessIDParseErr)
+		return "", accessIDParseErr
+	}
```
md2119, line 269:
```suggestion
	
	if _, accessIDParseErr := uuid.FromString(reportConfigurationScopeID); accessIDParseErr != nil {
		log.WriteToStderrf("failed to convert report configuration to postgres format, bad scope ID. Report Configuration ID: [%s] Name:[%s], error %v", reportConfiguration.GetId(), reportConfiguration.GetName(), accessIDParseErr)
		return "", accessIDParseErr
	}
```

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +257,60 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+			if err != nil {
+				batchIDs := make([]string, 0, len(reportConfigs))
+				for _, reportConfig := range reportConfigs {
+					batchIDs = append(batchIDs, reportConfig.GetId())
+				}
+				log.WriteToStderrf("failed to persist report configurations IDs [%s] to store %v", strings.Join(batchIDs, " "), err)
+				return err
+			}
+			reportConfigs = reportConfigs[:0]
+		}
```
md2119, line 299:
```suggestion
		if len(reportConfigs) < batchSize {
		   return nil
		}
		err := store.UpsertMany(ctx, reportConfigs)
		if err != nil {
			batchIDs := make([]string, 0, len(reportConfigs))
			for _, reportConfig := range reportConfigs {
				batchIDs = append(batchIDs, reportConfig.GetId())
			}
			log.WriteToStderrf("failed to persist report configurations IDs [%s] to store %v", strings.Join(batchIDs, " "), err)
			return err
		}
		reportConfigs = reportConfigs[:0]
```

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +257,60 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	if len(reportConfigs) > 0 {
+		err = store.UpsertMany(ctx, reportConfigs)
+		if err != nil {
+			return err
```
md2119, line 308: Why not log error here?

md2119 (approved):

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -251,6 +257,60 @@ func migrateRoles(gormDB *gorm.DB, postgresDB *pgxpool.Pool, legacyStore legacyr
...
+		reportConfigs = append(reportConfigs, obj)
+		if len(reportConfigs) == batchSize {
+			err := store.UpsertMany(ctx, reportConfigs)
+			if err != nil {
+				batchIDs := make([]string, 0, len(reportConfigs))
+				for _, reportConfig := range reportConfigs {
+					batchIDs = append(batchIDs, reportConfig.GetId())
+				}
+				log.WriteToStderrf("failed to persist report configurations IDs [%s] to store %v", strings.Join(batchIDs, " "), err)
+				return err
```
charmik-redhat, line 296: Since other places have changed (or suggested to be changed) to just logging error and returning nil and continue to migrate the rest of the configs, we can do the same here as well.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
charmik-redhat: +1

charmik-redhat (approved):
Left a couple of comments. Otherwise good to merge once @md2119's comments are addressed.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/migration.go
```diff
@@ -91,6 +92,11 @@ func migrateAll(rocksDatabase *rocksdb.RocksDB, gormDB *gorm.DB, postgresDB *pgx
 		return errors.Wrap(err,
 			"moving roles from rocksdb to postgres")
 	}
+	// This function call was added between 3.73 and 3.74 in order to cover an overlooked reference to access scope IDs.
```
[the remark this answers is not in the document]
rhybrillou: I'll ask regarding backport in 3.73.

dhaus67 (approved):

PR #4686, opened 2023-02-02
Disable Prom metrics if monitoring is not set

Maddosaurus:
## Description

The prom metrics server is started, even when monitoring is not explicitly exposed.
Therefore, we get port clashes in that case as well.
This PR disables the Compliance port completely if monitoring is not enabled, and sets it to a non-standard port if enabled, eliminating all clashes.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
+++ b/image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
@@ -91,9 +91,11 @@ spec:
       - command:
         - stackrox/compliance
         env:
-        {{- if ._rox.collector.exposeMonitoring }}
         - name: ROX_METRICS_PORT
+        {{- if ._rox.collector.exposeMonitoring }}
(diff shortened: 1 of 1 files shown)


vikin91 (approved):
Code and the idea looks good, I didn't have a chance to test it.
 
Please make sure that it works - ideally with a fresh install & cluster upgrade scenario if possible.

Will we need manual upgrade instructions for users installing from manifests?

Maddosaurus:
> Will we need manual upgrade instructions for users installing from manifests?

I have tested with a fresh OS 4.12 cluster and confirmed this working.
I don't think we need upgrade instructions for customers just yet, as we are not enabling Metrics collection when the feature flag is turned off.
I think we only run into this situation when we enable the FF and deploy to OpenShift, as this leads to our deployment scripts forcing monitoring to be enabled.

PR #4427, opened 2023-01-17
[ROX-14516] Standardize postgres search query parser steps

md2119:
## Description

This PR standardizes a few steps in search query parsing. This is split out from the GROUP BY clause POC that I am working on.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed
Unit tests cover the refactor

PR #4688, opened 2023-02-02
Postgres search enhancements: part 1

md2119:
## Description

This PR adds an enhancement to the Postgres search framework. This PR:
- expands the `v1.Query` proto, our current query object, to receive `select` and ` group by` portion of the query. Note that the `v1.Query` is not exposed at any API. I do not plan to expose the new query format at REST endpoints just yet to give enough soak time. 
- adds support to read free-form resultadds support to directly read proto fields from database into a struct
- adds support to group the responses
- adds aliasing to SQL query to remove confusion because of overlapping names. For example, without aliasing `select table1.id, table2.id from table1 inner join table2 on table1.id = table2.fk` returns the result as 
```
 id | id 
----+----
(0 rows)
```

The framework exposes func `RunSelectRequestForSchema` enabling us to interact with the database with complex queries and retrieve data in free-form. Since the response is free-form (not a stored proto object), the functions takes in a type. The type is a struct representation of the retrieved records. The formula to construct the query is simple. The struct must have fields you requested along with the group by fields. My added aliasing to postgres search framework.

**Example:** 
To query all policies grouped by severity, the query would be:
```
q :=	search.NewQueryBuilder().
		AddSelectFields(search.PolicyName).
		AddGroupBy(search.Severity).ProtoQuery()
```
 and the response struct would be:
```
type Demo struct {
  Policies []string `db:"policy"`
  Severity int `db:"severity"`
}
// where proto tags are essentially the lower case search labels without whitespaces.
// PolicyName = newFieldLabel("Policy")
// Severity = newFieldLabel("Severity")
```

and the call would be:
```
results, err = pkgPG.RunSelectRequestForSchema[Demo](ctx, testDB.Pool, schema.Policies, q)
// where results is []*Demo
```
**Strongly recommend that you look at the test cases for example queries and responses.**

<br>
<br>

----
Before this PR, data from the database could be retrieved only in the following formats OOTB:
- `proto.Message` - the stored objects
- `search.Result` - slimmed version of IDs and matches of stored object matching
```
type Result struct {
	ID      string
	Matches map[string][]string
	Score   float64
	Fields  map[string]interface{}
}
```

Before this PR, it was impossible to get free-form data OOTB from the database. For example, following queries cannot be answered OOTB, but can be answer with the new framework.
- `get all containers names in namespace stackrox`
- `get a count of distinct base operating systems in the environment`
- `count of alerts grouped by severity`

----

## Checklist

## Testing Performed
Unit (this feature is not being used yet)


pkg/search/postgres/query/common.go
```diff
@@ -46,7 +49,17 @@ func (f *SelectQueryField) PathForSelectPortion() string {
 	if f.Alias == "" {
 		return f.SelectPath
 	}
-	return fmt.Sprintf("%s as %s", f.SelectPath, f.Alias)
+	return fmt.Sprintf("%s %s", f.SelectPath, f.Alias)
+}
+
+// MustAggregate returns true if the select field needs to aggregated.
+func (f *SelectQueryField) MustAggregate() bool {
+	return !f.FromGroupBy && !f.DerivedField
```
clickboo, line 57: What is the difference between this func and the AggrAndUnnest? What is the decision point to unnest or not? When is this needed? The check condition is the same in both 🤔

md2119: ironed out the code. Removed unused.

pkg/search/postgres/common.go
```diff
@@ -268,31 +308,102 @@ func (p *parsedPaginationQuery) AsSQL() string {
...
+
+	for _, field := range querySelect.GetFields() {
+		fieldMetadata := queryFields[field]
+		dbField := fieldMetadata.baseField
+		if dbField == nil {
+			return errors.Errorf("field %s in SELECT clause does not exist in table %s or connected tables", field, schema.Table)
+		}
+
+		querySoFar.SelectedFields = append(querySoFar.SelectedFields, pgsearch.SelectQueryField{
+			SelectPath: qualifyColumn(dbField.Schema.Table, dbField.ColumnName, ternary.String(dbField.SQLType == "uuid", "::text", "")),
```
keyallis, line 324: This is a really curious line. My understanding is that idiomatically go as a language doesn't want to support ternary because it's deemed difficult to parse. https://go.dev/doc/faq#Does_Go_have_a_ternary_form

pkg/search/postgres/common.go
```diff
@@ -422,18 +535,72 @@ func standardizeQueryAndPopulatePath(q *v1.Query, schema *walker.Schema, queryTy
...
+
+	queryEntry, err := compileQueryToPostgres(schema, q, dbFields, nowForQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	if queryEntry != nil {
+		parsedQuery.Where = queryEntry.Where.Query
+		parsedQuery.Data = queryEntry.Where.Values
+		// We won't need this once highlights is removed and fields can only be selected when explicitly specified in the query.
```
keyallis, line 587: Might be good to add a TODO here or something else easily identifiable so we know to come back and clean this up

keyallis (commented):
Couple of notes.

go.mod
```diff
@@ -129,6 +129,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/georgysavva/scany v1.2.1
```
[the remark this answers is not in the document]
md2119: seems like the `go get` may not group the deps correctly. Manually collapsed this and the indirect group below.

proto/api/v1/search_service.proto
```diff
@@ -111,14 +119,18 @@ message BooleanQuery {
...
+        ConjunctionQuery    conjunction   = 1;
+        DisjunctionQuery    disjunction   = 2;
+        BaseQuery           base_query    = 3;
+        BooleanQuery        boolean_query = 6;
     }
-    QueryPagination pagination       = 4;
+
+    QuerySelect             select        = 8;
+    QueryGroupBy            group_by      = 7;
+    QueryPagination         pagination    = 4;
```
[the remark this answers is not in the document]
md2119: This one I will keep it as is.  The `one of` are top-level fields and not nested. IMO Re-arranging as such removes that misconception.

pkg/search/postgres/common.go
md2119: I believe this function simply encapsulates if-then-else construct under the hood and nothing more.

pkg/search/postgres/common.go
```diff
@@ -89,7 +92,10 @@ type query struct {
 	Pagination parsedPaginationQuery
 	InnerJoins []innerJoin
 
-	GroupBys []groupByEntry
+	// This indicates if a primary key is present in the group by clause. Unless GROUP BY clause is explicitly provided,
+	// we order the results by the primary key of the schema.
```
clickboo, line 96: It might be worthwhile to explore sql proto tags which indicate default group by field.  This will help in cases where the ID is a uuid but the unique name is almost always what we want to group by (category name, policy name, cluster name)

keyallis (approved):
LGTM. Still not sure ternary is something we should be using, but that's no hill to die on

pkg/search/postgres/common.go
```diff
@@ -268,31 +308,102 @@ func (p *parsedPaginationQuery) AsSQL() string {
...
+			}(),
+			// TODO(mandar): derived fields as field labels or expand QuerySelect to store user-defined aggr functions?
+			// TODO(mandar): Handle array type and map type.
+		})
+	}
+	return nil
+}
+
+func populateGroupBy(querySoFar *query, groupBy *v1.QueryGroupBy, schema *walker.Schema, queryFields map[string]searchFieldMetadata) error {
+	if querySoFar.QueryType == SEARCH && len(groupBy.GetFields()) > 0 {
```
clickboo, line 340: Shouldn't this be querySoFar.QueryType != SELECT

pkg/search/postgres/query/common.go
```diff
@@ -21,6 +21,9 @@ type SelectQueryField struct {
 	FieldType  walker.DataType
 	FieldPath  string // This is the search.Field.FieldPath for this field.
 
+	FromGroupBy  bool
```
clickboo, line 24: comments would be nice

pkg/search/postgres/query/common.go
```diff
@@ -46,7 +49,7 @@ func (f *SelectQueryField) PathForSelectPortion() string {
 	if f.Alias == "" {
 		return f.SelectPath
 	}
-	return fmt.Sprintf("%s as %s", f.SelectPath, f.Alias)
+	return fmt.Sprintf("%s %s", f.SelectPath, f.Alias)
```
clickboo, line 52: I like having the "as" in the query for readability - especially since the select path is the combined fieldpath which is long and unreadable. Any reason to remove it?

pkg/search/postgres/select_query_test.go
```diff
@@ -0,0 +1,309 @@
...
+				{"bcs"},
+			},
+		},
+		{
+			desc: "base schema; select w/ where",
+			q: search.NewQueryBuilder().
+				AddSelectFields(search.TestString).
+				AddExactMatches(search.TestString, "acs").ProtoQuery(),
+			resultStruct:  Struct1{},
+			expectedQuery: "select test_multi_key_structs.String_ teststring from test_multi_key_structs where test_multi_key_structs.String_ = $1",
```
clickboo, line 99: Curious why .String_ vs .Nested (without the _). Should be consistent right?

pkg/search/postgres/select_query_test.go
```diff
@@ -0,0 +1,309 @@
...
+				{"acs"},
+			},
+		},
+		{
+			desc: "child schema; multiple select w/ where",
+			q: search.NewQueryBuilder().
+				AddSelectFields(search.TestNestedString, search.TestNestedBool).
+				AddExactMatches(search.TestNestedString, "nested_acs").ProtoQuery(),
+			resultStruct: Struct2{},
+			expectedQuery: "select test_multi_key_structs_nesteds.Nested testnestedstring, test_multi_key_structs_nesteds.IsNested testnestedbool " +
```
clickboo, line 111: as helps readability

clickboo (commented):
Will be doing another pass in some time, sending comments in anyway to enable processing them in parallel

pkg/search/postgres/common.go
md2119: I don't think we should add query information to protos. Protos carry the schema definition whereas select query is a custom user-defined operation. I would like to keep the contract as transparent as possible. Users must specify a complete query for the response they expect.

pkg/search/postgres/common.go
```diff
@@ -100,11 +106,22 @@ type groupByEntry struct {
...
 func (q *query) ExtraSelectedFieldPaths() []pgsearch.SelectQueryField {
-	if !q.DistinctAppliedOnPrimaryKeySelect() {
+	if !q.DistinctAppliedOnPrimaryKeySelect() && !q.groupByNonPKFields() {
 		return nil
 	}
 
 	var out []pgsearch.SelectQueryField
+	for _, groupBy := range q.GroupBys {
+		var alreadyExists bool
+		for _, selectedField := range q.SelectedFields {
```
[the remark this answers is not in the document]
md2119: We cannot do that. Order matters. Sets do not maintain order.

pkg/search/postgres/select_query_test.go
md2119: not following. `string_` is the table column name.

pkg/search/postgres/select_query_test.go
md2119: not following

pkg/search/postgres/query/common.go
md2119: I added it in a prefactor PR #4427 but realized that it is unnecessary and make string parsing tricky.

pkg/search/postgres/common.go
```diff
@@ -124,14 +141,20 @@ func (q *query) populatePrimaryKeySelectFields() {
 	for _, pk := range q.Schema.PrimaryKeys() {
 		q.PrimaryKeyFields = append(q.PrimaryKeyFields, pgsearch.SelectQueryField{
 			SelectPath: qualifyColumn(pk.Schema.Table, pk.ColumnName, ternary.String(pk.SQLType == "uuid", "::text", "")),
-			Alias:      strings.Join([]string{q.Schema.Table, pk.ColumnName}, "_"),
-			FieldType:  pk.DataType,
+			Alias: func() string {
+				if q.QueryType == SELECT {
```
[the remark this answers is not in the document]
md2119: Get query deterministically gets `serialized` column only.

PR #4694, opened 2023-02-02
ROX-14182: Disable API modifications for declaratively defined permission sets

ivan-degtiarenko:
## Description

First PR in series of PRs that would disable API modifications for declaratively defined:
* Permission sets
* Access scopes
*  Roles
* Auth providers

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/role/datastore/datastore.go
+++ b/central/role/datastore/datastore.go
@@ -23,8 +23,8 @@ type DataStore interface {
 	GetAllPermissionSets(ctx context.Context) ([]*storage.PermissionSet, error)
 	CountPermissionSets(ctx context.Context) (int, error)
 	AddPermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
-	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet, requestOrigin storage.Traits_Origin) error
-	RemovePermissionSet(ctx context.Context, id string, requestOrigin storage.Traits_Origin) error
+	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
+	RemovePermissionSet(ctx context.Context, id string) error
 
 	GetAccessScope(ctx context.Context, id string) (*storage.SimpleAccessScope, bool, error)
 	GetAllAccessScopes(ctx context.Context) ([]*storage.SimpleAccessScope, error)
--- a/central/role/datastore/datastore.go
+++ b/central/role/datastore/datastore.go
@@ -23,8 +23,8 @@ type DataStore interface {
 	GetAllPermissionSets(ctx context.Context) ([]*storage.PermissionSet, error)
 	CountPermissionSets(ctx context.Context) (int, error)
 	AddPermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
-	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
-	RemovePermissionSet(ctx context.Context, id string) error
+	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet, requestOrigin storage.Traits_Origin) error
+	RemovePermissionSet(ctx context.Context, id string, requestOrigin storage.Traits_Origin) error
 
 	GetAccessScope(ctx context.Context, id string) (*storage.SimpleAccessScope, bool, error)
 	GetAllAccessScopes(ctx context.Context) ([]*storage.SimpleAccessScope, error)
--- a/central/role/datastore/datastore_impl.go
+++ b/central/role/datastore/datastore_impl.go
@@ -203,7 +203,7 @@ func (ds *dataStoreImpl) AddPermissionSet(ctx context.Context, permissionSet *st
 	return nil
 }
 
-func (ds *dataStoreImpl) UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error {
+func (ds *dataStoreImpl) UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet, requestOrigin storage.Traits_Origin) error {
 	if err := sac.VerifyAuthzOK(roleSAC.WriteAllowed(ctx)); err != nil {
 		return err
 	}
@@ -218,7 +218,11 @@ func (ds *dataStoreImpl) UpdatePermissionSet(ctx context.Context, permissionSet 
 	defer ds.lock.Unlock()
 
 	// Verify storage constraints.
-	if err := ds.verifyPermissionSetIDExists(ctx, permissionSet.GetId()); err != nil {
+	existingPermissionSet, err := ds.verifyPermissionSetIDExists(ctx, permissionSet.GetId())
+	if err != nil {
+		return err
+	}
+	if err := verifyPermissionSetOriginMatches(existingPermissionSet, requestOrigin); err != nil {
 		return err
 	}
 
@@ -231,7 +235,7 @@ func (ds *dataStoreImpl) UpdatePermissionSet(ctx context.Context, permissionSet 
 	return nil
 }
 
-func (ds *dataStoreImpl) RemovePermissionSet(ctx context.Context, id string) error {
+func (ds *dataStoreImpl) RemovePermissionSet(ctx context.Context, id string, requestOrigin storage.Traits_Origin) error {
 	if err := sac.VerifyAuthzOK(roleSAC.WriteAllowed(ctx)); err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


central/role/datastore/datastore.go
```diff
@@ -22,8 +22,8 @@ type DataStore interface {
 	GetAllPermissionSets(ctx context.Context) ([]*storage.PermissionSet, error)
 	CountPermissionSets(ctx context.Context) (int, error)
 	AddPermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
-	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) error
-	RemovePermissionSet(ctx context.Context, id string) error
+	UpdatePermissionSet(ctx context.Context, permissionSet *storage.PermissionSet, requestOrigin storage.Traits_Origin) error
+	RemovePermissionSet(ctx context.Context, id string, requestOrigin storage.Traits_Origin) error
```
dhaus67, line 26: I wonder how easy it would be as a client of this datastore now.

For people that do not know the constraints on the `Traits_Origin`, they would currently have to be careful when calling this that they pass in `storage.Traits_IMPERATIVELY`, as otherwise they would be allowed to modify declarative configurations.
Seeing as `storage.Traits_IMPERATIVELY` is the default value of the enum, since by default we assume all resources are imperatively created ones, this makes it even more confusing to me.

We might think about generalizing this "optionally setting the request origin" problem. I do like options for this, but I'm not sure whether they are overkill (i.e. shared options for the declarative config that overwrite the default origin used, which would be `storage.Traits_IMPERATIVELY`). Alternatively, we would also add the request origin as a variadic args (`...`), allowing people to _optionally_ specify the origin (with the caveat that we now can allow removals of both imperative / declarative origins for clients, which might be something beneficial at some point).

central/role/datastore/datastore_impl.go
```diff
@@ -269,6 +276,14 @@ func (ds *dataStoreImpl) RemovePermissionSet(ctx context.Context, id string) err
 	return nil
 }
 
+func verifyPermissionSetOriginMatches(ps *storage.PermissionSet, requestOrigin storage.Traits_Origin) error {
+	if ps.GetTraits().GetOrigin() != requestOrigin {
+		return errors.Wrapf(errox.InvalidArgs, "permission set %q is %s, cannot be modified or deleted in %s manner",
```
dhaus67, line 281: This sounds more sound to me?
```suggestion
		return errors.Wrapf(errox.InvalidArgs, "permission set %q's origin is %s and not %s, cannot be modified or deleted",
```

central/role/service/service_impl.go
```diff
@@ -230,7 +230,7 @@ func (s *serviceImpl) PostPermissionSet(ctx context.Context, permissionSet *stor
 }
 
 func (s *serviceImpl) PutPermissionSet(ctx context.Context, permissionSet *storage.PermissionSet) (*v1.Empty, error) {
-	err := s.roleDataStore.UpdatePermissionSet(ctx, permissionSet)
+	err := s.roleDataStore.UpdatePermissionSet(ctx, permissionSet, storage.Traits_IMPERATIVE)
```
dhaus67, line 233: We should add tests on the service level to verify that update / delete is handled correctly, this way we add more than just the manual tests.

central/role/datastore/datastore.go
ivan-degtiarenko: > shared options for the declarative config that overwrite the default origin used, which would be storage.Traits_IMPERATIVELY

What do you mean by "shared options"?

Here are the alternative solutions that I see. All of them would not require any change for the clients who work with imperative resources:
1. Introduce functions like `RemoveDeclarativePermissionSet` and `UpdateDeclarativePermissionSet`
2. Introduce variadic args
3. Pass request origin within the context

central/role/service/service_impl.go
ivan-degtiarenko: What do you mean by Update/Delete handled correctly? Looking at the logic within `service_impl`, I don't think there is much to test(especially if we go with determining client source from context)

pkg/declarativeconfig/context.go
```diff
@@ -0,0 +1,26 @@
+package declarativeconfig
```
dhaus67, line 1: We should add unit tests for this functionality as well.

pkg/declarativeconfig/context.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type originCheckerKey struct{}
+
+// AllowOnlyDeclarativeOperations signals that the context holder is allowed to modify declarative resources.
+const AllowOnlyDeclarativeOperations = true
+
+// WithAllowOnlyDeclarativeOperations returns a context that is a child of the given context and allows to modify
+// declarative resources.
+func WithAllowOnlyDeclarativeOperations(ctx context.Context) context.Context {
```
dhaus67, line 16: From the naming, maybe `WithModifyDeclarativeResource` is better? I think we can drop the only, think it's more confusing then really helpful but up to you.

pkg/declarativeconfig/context.go
```diff
@@ -0,0 +1,26 @@
...
+	return context.WithValue(ctx, originCheckerKey{}, AllowOnlyDeclarativeOperations)
+}
+
+// IsOriginModifiable returns whether context allows to modify declarative resources.
+func IsOriginModifiable(ctx context.Context, origin storage.Traits_Origin) bool {
```
dhaus67, line 21: Maybe it would be worthwhile to do the following changes:

Instead of passing in `Traits_Origin`, since clients won't actually care much about the `Traits` but rather the resource, we can create an interface and let clients pass in their proto message, and we'll receive the traits origin within the function:
```go
type ResourceWithTraits interface {
  GetTraits() *storage.Traits
}

func CanModifyResource(ctx context.Context, resource ResourceWithTraits) bool {
	if ctx.Value(originCheckerKey{}) == AllowOnlyDeclarativeOperations {
		return resource.GetTraits().GetOrigin() == storage.Traits_DECLARATIVE
	}
	return resource.GetTraits().GetOrigin() == storage.Traits_IMPERATIVE
}

// Clients can then call it as such:
func verifyPermissionSetOriginMatches(ctx context.Context, ps *storage.PermissionSet) error {
	if !declarativeconfig.CanModifyResource(ctx, ps) {
		return errors.Wrapf(errox.InvalidArgs, "permission set %q's origin is %s, cannot be modified or deleted within this context",
			ps.GetName(), ps.GetTraits().GetOrigin())
	}
	return nil
}
```

I think that ones more worthwhile to have - it's also _nearly_ in line with all our protos we have in scope for declarative config, as they do not define the `Traits` message nested. Exception would be the `Groups` proto, but I'm willing to make the compromise since its "weird" per se - defining the ID within the `GroupProperties` message.

central/role/datastore/datastore_impl.go
```diff
@@ -269,6 +277,14 @@ func (ds *dataStoreImpl) RemovePermissionSet(ctx context.Context, id string) err
 	return nil
 }
 
+func verifyPermissionSetOriginMatches(ctx context.Context, ps *storage.PermissionSet) error {
+	if !declarativeconfig.IsOriginModifiable(ctx, ps.GetTraits().GetOrigin()) {
+		return errors.Wrapf(errox.InvalidArgs, "permission set %q's origin is %s, cannot be modified or deleted within this context",
```
dhaus67, line 282: Do you think we should leak this "within this context" to any potential clients?
If someone calls this via API, I can imagine that it may be confusing.

Maybe we should just do the following:
- Return `errox.NotAuthorized` to clearly specify that its not invalid args, and in the current context of trying to delete via API there's no "valid" args to give, it's simply not authorized.
- As an error message, swap out the "within this context" with "with the current permission"

central/role/datastore/datastore_test.go
```diff
@@ -386,6 +393,29 @@ func (s *roleDataStoreTestSuite) TestPermissionSetWriteOperations() {
 
 	err = s.dataStore.AddPermissionSet(s.hasWriteCtx, goodPermissionSet)
 	s.NoError(err, "adding a permission set with ID and name that used to exist is not an error")
+
+	err = s.dataStore.AddPermissionSet(s.hasWriteCtx, declarativePermissionSet)
+	s.NoError(err, "adding a permission set declaratively is not an error")
```
dhaus67, line 398: General point:
Just from a consistency point of view, should we also limit _adding_ declarative resources via the context?

Realistically, only the declarative config manager should be the one allowed to add declarative resources - otherwise users may add declarative resources but essentially can never delete them / do anything with them from the API POV.

On the other hand, the declarative resources should be reconciled and will be deleted in an eventual consistent manner, but from the UX point of view these things might be confusing.

Let's add a follow-up to the current story, since it doesn't include the whole "constrain API interactions with declarative resources to a particular context / client" and constrain the ability to add declarative resources within there. WDYT?

ivan-degtiarenko: Sounds good. Created a follow up ticket: https://issues.redhat.com/browse/ROX-14999

dhaus67 (approved):

PR #4695, opened 2023-02-03
ROX-14797: Fix bulk Add CVEs to Policy for Postgres

vjwilson:
## Description

The main purpose of this change is to fix a gap when migrating to Postgres, where CVEs could no longer be added to a policy from the CVE list pages.

In addition, it cleans up some "drift" in functionality, and some bugs that show this hasn't been a much-used feature.

- It filters out default policies from the list of existing VM policies that CVEs could be added to, because at some point subsequent to this default policies became read-only.
- It fixes a bug where existing non-default (i.e., customer-created) polices could not be selected, because of the way the VM data flow "locks" the JS objects that are passed by reference into this dialog. (The solution is to use a utility function to deep clone the an existing policy object selected as a target. The clone does not have the "locked" state, so the new CVEs can be added to it. In this case, the single policy object is a finite size, so the clone is neither slow nor particularly memory intensive.)
- Fixed several browser console warnings about incorrect React type declarations. (This code is from before we started to use TypeScript.)
- Removed the Add to Policy option from lists of Node and Platform CVEs, because those are not applicable to our polices at this time.

## Checklist
- [x] Investigated and inspected CI test results (failing Go unit tests are flakes)

## Testing Performed

Manual testing
1. Confirmed that we can add CVEs to a new policy
2. Confirmed that only non-default policies now appear in the dropdown list of possible target existing policies.
3. Confirmed that we can add CVEs to an existing non-default policy.

Screen Shot 2023-02-07 at 12 13 38 PM

Screen Shot 2023-02-07 at 11 57 46 AM

### DIFF
--- a/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/CveBulkActionDialogue.js
+++ b/ui/apps/platform/src/Containers/VulnMgmt/List/Cves/CveBulkActionDialogue.js
@@ -9,6 +9,7 @@ import CustomDialogue from 'Components/CustomDialogue';
 import InfoList from 'Components/InfoList';
 import Loader from 'Components/Loader';
 import { POLICY_ENTITY_ALL_FIELDS_FRAGMENT } from 'Containers/VulnMgmt/VulnMgmt.fragments';
+import entityTypes from 'constants/entityTypes';
 import queryService from 'utils/queryService';
 import { createPolicy, savePolicy } from 'services/PoliciesService';
 import { truncate } from 'utils/textUtils';
@@ -30,7 +31,7 @@ const findCVEField = (policySections) => {
     return { policySectionIdx, policyGroupIdx };
 };
 
-const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds }) => {
+const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds, cveType }) => {
     const [messageObj, setMessageObj] = useState(null);
     const dialogueRef = useRef(null);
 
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/VulnMgmt/List/Cves/CveBulkActionDialogue.js
```diff
@@ -30,12 +32,15 @@ const findCVEField = (policySections) => {
     return { policySectionIdx, policyGroupIdx };
 };
 
-const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds }) => {
+const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds, cveType }) => {
```
md2119, line 35: Can you additionally pass `data.cve` to the modal and use that instead of parsing the ID? If the ID format changes, this will break and we are back to the repair shop.

vjwilson: Because the labyrinthine way we had to construct the data flow in order to allow for the "infinite drilldown" style of vuln management, we do not have the data at the top level--we only have the query and query vars, which we pass down to the generic VM table component to make its own GraphQL call.

md2119 (approved):
LGTM although is code has become a maze. Let us make sure we simply this workflow in VM2.0

ui/apps/platform/src/Containers/VulnMgmt/List/Cves/CveBulkActionDialogue.js
```diff
@@ -46,16 +52,61 @@ const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds }) => {
...
     // use GraphQL to get the (hopefully cached) cve summaries to display in the dialog
-    const CVES_QUERY = gql`
-        query getCves($query: String) {
-            results: vulnerabilities(query: $query) {
-                id
-                cve
-                summary
-                vulnerabilityTypes
-            }
+    let cveQuery = '';
```
alwayshooin, line 55: can we stick to the all caps convention as we have before for the gqls for easy parsing?

alwayshooin (approved):

ui/apps/platform/src/Containers/VulnMgmt/List/Cves/CveBulkActionDialogue.js
```diff
@@ -46,16 +52,61 @@ const CveBulkActionDialogue = ({ closeAction, bulkActionCveIds }) => {
...
+            cveQuery = gql`
+                query getCves($query: String) {
+                    results: vulnerabilities(query: $query) {
+                        id
+                        cve
+                        summary
+                        vulnerabilityTypes
+                    }
+                }
+            `;
```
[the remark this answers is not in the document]
vjwilson: The files with "reusable" queries has grown unwieldy over the years, and often works at odds where GraphQL allows us to grab just the fields we need in a particular context.

That's the case here, where the dialog really only needs these 4 fields, but this query may never get reused. I opted to leave this one-off custom query here, co-located with the code that needs it.

PR #4706, opened 2023-02-03
ROX-14477: Fix data race in `poller_test.go`

msugakov:
## Description

See the output in the related ticket https://issues.redhat.com/browse/ROX-14477

The case seems to be a very simple example of the data race. Despite that, I wasn't able to reproduce the original failure locally. Maybe the runner machine had some special condition.

The implementation of the `Poller` (struct under test) seems legit, at least I wasn't able to spot anything obviously broken.

## Checklist

None of the following is needed:

## Testing Performed

* Will merge if CI passes.
* Ran it successfully locally with `GOTAGS=release RACE=true scripts/go-test.sh -v -count=1000 -run TestPollWithTimeout ./pkg/concurrency`. However, as I mentioned above, even the original failure did not get reproduced this way.

### DIFF
--- a/pkg/concurrency/poller_test.go
+++ b/pkg/concurrency/poller_test.go
@@ -50,12 +50,11 @@ func TestPollWithTimeout(t *testing.T) {
 	assert.False(t, PollWithTimeout(func() bool {
 		return false
 	}, 5*time.Millisecond, 50*time.Millisecond))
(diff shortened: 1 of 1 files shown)


jschnath (commented):
LGTM

jschnath (approved):

PR #4710, opened 2023-02-03
Add an OCP option for UI e2e tests

gavin-stackrox:
## Description

This will allow UI e2e tests to run in an OSCI OCP 4.12 cluster via `/test ocp-ui-e2e-tests`. At present they will not pass and that is a separate WIP. 

Companion to https://github.com/openshift/release/pull/36201

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/tests/e2e/run-ui-e2e.sh
+++ b/tests/e2e/run-ui-e2e.sh
@@ -32,18 +32,20 @@ test_ui_e2e() {
 run_ui_e2e_tests() {
     info "Running UI e2e tests"
 
-    touch /tmp/hold
-    while [[ -e /tmp/hold ]]; do
-        info "Holding this job for debug"
-        sleep 60
-    done
-
     if [[ "${LOAD_BALANCER}" == "lb" ]]; then
         local hostname
-        hostname=$("$ROOT/tests/e2e/get_hostname.py" "${API_HOSTNAME}")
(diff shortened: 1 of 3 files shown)


tests/e2e/run-ui-e2e.sh
```diff
@@ -32,10 +34,18 @@ run_ui_e2e_tests() {
...
+            hostname=$("$ROOT/tests/e2e/get_hostname.py" "${API_HOSTNAME}")
+        else
+            hostname="${API_HOSTNAME}"
+        fi
+        info "Hostname for central-lb alias: ${hostname}"
         echo "central-lb ${hostname}" > /tmp/hostaliases
         export HOSTALIASES=/tmp/hostaliases
         export UI_BASE_URL="https://central-lb:443"
+    elif [[ "${LOAD_BALANCER}" == "route" ]]; then
+        die "unsupported LOAD_BALANCER ${LOAD_BALANCER}"
```
vjwilson, line 48: I don't know enough about this environment to be sure, but this looks like it this aborts if we are in an OpenShift environment.

I'm going to assume that's not all that's going on, and not block the PR on this. I would like to know why this is here.

gavin-stackrox: We unfortunately do not cover OpenShift routes in CI (the option is there in the deploy scripts to use it but it is not done (added https://issues.redhat.com/browse/ROX-15022)).
I added this check in case someone tried to sneak it in without properly adding support . The same logic for 'lb' would _probably_ work but falling through to 'localhost' would falsely work.

PR #4712, opened 2023-02-03
Add support to select derived fields

md2119:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why
you did not do so. Valid reasons include, for example, "CI is sufficient",
"No testable changes". Feel free to attach JSON snippets, curl commands,
screenshots.

In addition to reviewing your code, reviewers **must** also review your testing
instructions and make sure they are sufficient.

PR #4714, opened 2023-02-03
Postgres search enhancements: part 3

md2119:
## Description

[For part 2 refer here](https://github.com/stackrox/stackrox/pull/4712) 

This PR adds a generic template for writing derived field query. This enables us to write create derived field queries from any regular search fields.

-----
**Example 1:**
To query the count of images grouped by registry, the query would be:
```
q := search.NewQueryBuilder().
	AddSelectFields(
            *v1.QueryField{
               Field: search.ImageName, 
               AggrFunc: "count",
            }
         ).
	AddGroupBy(search.ImageRegistry).ProtoQuery()
```
and the response struct would be:
```
type Demo struct {
  ImageCount int `db:"image_count"`
  ImageRegistry string `db:"image_registry"`
}
// where proto tags are essentially the lower case search labels+aggr func w/ underscore for whitespaces.
// ImageName  = newFieldLabel("Image")
// ImageRegistry = newFieldLabel("Image Registry")
```
and the call would be:
```
results, err = pkgPG.RunSelectRequestForSchema[Demo](ctx, testDB.Pool, schema.ImagesSchema, q)
// where results is []*Demo
```

**Example 2:**
To query the count of images grouped by registry, the query would be:
```
q := search.NewQueryBuilder().
	AddSelectFields(
            *v1.QueryField{
               Field: search.Severity, 
               AggrFunc: "max",
            }
         ).
	AddGroupBy(search.CVE).ProtoQuery()
```
and the response struct would be:
```
type Demo struct {
  TopSeverity int `db:"severity_max"`
  CVE string `db:"cve"`
}
// where proto tags are essentially the lower case search labels+aggrfunc w/ underscore for whitespaces.
// CVE = newFieldLabel("CVE")
// Severity  = newFieldLabel("Severity")
```
and the call would be:
```
results, err = pkgPG.RunSelectRequestForSchema[Demo](ctx, testDB.Pool, schema.ImageCvesSchema, q)
// where results is []*Demo
```
----

Before this, we had to declare derived fields like below, which is hard to maintain and not easy to extend.

```
NamespaceCount  = newDerivedFieldLabel("Namespace Count", NamespaceID, CountDerivationType)
```

## Checklist

## Testing Performed
Unit


pkg/search/postgres/common.go
```diff
@@ -444,10 +421,10 @@ func populatePagination(querySoFar *query, pagination *v1.QueryPagination, schem
 			var descending bool
 			switch fieldMetadata.derivedMetadata.DerivationType {
 			case searchPkg.CountDerivationType:
-				selectField = selectQueryField(so.GetField(), dbField, "count", walker.Integer)
+				selectField = selectQueryField(so.GetField(), dbField, false, "count")
```
clickboo, line 424: use const

pkg/search/postgres/aggr_func.go
```diff
@@ -0,0 +1,27 @@
+package postgres
+
+import "github.com/stackrox/rox/pkg/postgres/walker"
+
+// AggrFunc is the internal enum representation of the SQL aggregate functions.
+type AggrFunc string
+
+// Defines all the SQL aggregate functions.
+const (
+	Count AggrFunc = "count"
```
clickboo, line 10: Define another one? Empty AggrFunc = ""

pkg/search/postgres/common.go
```diff
@@ -1179,19 +1156,28 @@ func qualifyColumn(table, column, cast string) string {
 	return table + "." + column + cast
 }
 
-func selectQueryField(searchField string, field *walker.Field, aggrFunc string, dataType walker.DataType) pgsearch.SelectQueryField {
+func selectQueryField(searchField string, field *walker.Field, selectDistinct bool, aggrFunc string) pgsearch.SelectQueryField {
```
clickboo, line 1159: Use AggrFunc instead of string as param

pkg/search/postgres/common.go
```diff
@@ -1179,19 +1156,28 @@ func qualifyColumn(table, column, cast string) string {
...
+	}
+	if dataType == "" {
+		dataType = field.DataType
 	}
 	return pgsearch.SelectQueryField{
-		SelectPath: selectPath,
-		Alias:      stringutils.JoinNonEmpty("", strings.Fields(searchField)...),
-		FieldType:  dataType,
+		SelectPath:   selectPath,
+		Alias:        strings.Join(strings.Fields(searchField+" "+aggrFunc), ""),
```
clickboo, line 1179: meant a space or a underscore here?

clickboo (commented):
Gave it a skim, sending a few comments

pkg/search/postgres/aggr_func.go
keyallis: +1 to this, defining an empty value and using that for comparison instead of comparing to `""`

md2119: good point

pkg/search/postgres/aggr_func.go
```diff
@@ -0,0 +1,28 @@
+package postgres
+
+import "github.com/stackrox/rox/pkg/postgres/walker"
+
+// AggrFunc is the internal enum representation of the SQL aggregate functions.
+type AggrFunc string
+
+// Defines all the SQL aggregate functions.
+const (
+	Unset AggrFunc = ""
```
dashrews78, line 10: nit:  IMO for readability, I think it would make sense to name these things like `UnsetAggr`, `CountAggr`, etc.  Particularly for the Unset one as that does not correspond to an aggregation.  I think that woulds make it clear when looking through the calls to `selectQueryField`.  I know doing so is sort of redundant.

md2119: that is a fair point

clickboo:
code lgtm, but I think we should add the PR description into a README in the search/postgres directory - with examples on how to use our search framework. Looking at code to figure how to write a query seems 👎

clickboo (approved):

dashrews78 (approved):
LGTM.  +1 to @clickboo README suggestion.

md2119:
Yes, I plan to write the README on the use once the framework is ready to use, so not quite there yet.

PR #4716, opened 2023-02-03
Query cancellation or timeout should not be transient

connorgorman:
## Description

If we run into a particular circumstance where we are hitting statement timeout, then we should protect ourselves from crashing because the issue is most likely not transient.

Follow up will be to add a notification to the user that we're experiencing unexpected query times

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -36,7 +36,9 @@ var transientPGCodes = set.NewFrozenStringSet(
 
 	// Class 57 — Operator Intervention
 	"57000", // operator_intervention
-	"57014", // query_canceled
+
+	// "57014", // query_canceled - this means there is a statement timeout and retrying will most likely not result in greater success
+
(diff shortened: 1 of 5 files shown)


pkg/postgres/pgutils/error.go
```diff
@@ -36,7 +36,9 @@ var transientPGCodes = set.NewFrozenStringSet(
 
 	// Class 57 — Operator Intervention
 	"57000", // operator_intervention
-	"57014", // query_canceled
+
+	// "57014", // query_canceled - this means there is a statement timeout and retrying will most likely not result in greater success
```
c-du, line 40: How do you tell that "retrying will most likely not result in greater success"?

connorgorman: Yeah, this might be a overstatement. It's unlikely, but still possible if there was a high load on the DB then the query may be successful after the initial part

PR #4721, opened 2023-02-06
ROX-14780: Disable re-sync when deploying ACS with deploy.sh

fredrb:
## Description

Disable re-sync when deploying ACS using `deploy/k8s` scripts. 

## Checklist

## Testing Performed

- [x] Deploying ACS using the updated script and check that resync is really disabled.

### DIFF
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -567,6 +567,12 @@ function launch_sensor {
            kubectl -n stackrox patch deploy/sensor --patch '{"spec":{"template":{"spec":{"containers":[{"name":"sensor","resources":{"limits":{"cpu":"500m","memory":"500Mi"},"requests":{"cpu":"500m","memory":"500Mi"}}}]}}}}'
        fi
     fi
+
+    # TODO(ROX-14310): Remove this patch when re-sync is disabled unconditionally
+    if [[ -n "$ROX_RESYNC_DISABLED" ]]; then
+        kubectl -n stackrox set env deploy/sensor ROX_RESYNC_DISABLED="true"
+    fi
+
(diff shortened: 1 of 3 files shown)


Maddosaurus (changes requested):
The change itself looks good to me!
Why I am requesting changes:

Have you tried this on an OpenShift cluster? 
AFAIK even `deploy/openshift/deploy.sh` calls `common/k8sbased.sh`, which would also disable resync on OpenShift clusters, wouldn't it?
The question is whether this is a problem or not.

In case this is not a problem at all, let me know and I'll approve the PR 😸

fredrb:
> AFAIK even deploy/openshift/deploy.sh calls common/k8sbased.sh, which would also disable resync on OpenShift clusters, wouldn't it?

It does, but that's why this is guarded by an env check:
```bash
if [[ -n "$ROX_RESYNC_DISABLED" ]]; then # Only disable if ROX_RESYNC_DISABLED is set 
    kubectl -n stackrox set env deploy/sensor ROX_RESYNC_DISABLED="true"
fi
```

And `ROX_RESYNC_DISABLED="true"` is only set on the k8s specific path: `deploy/k8s/sensor.sh`. If you check `deploy/openshift/deploy.sh` you'll notice that it doesn't set this env. 

However, you comment just made me realise that we're only checking if the env is set, not its content. Meaning that if for some reason you export `ROX_RESYNC_DISABLED="false"` in your env before running Openshift deploy, it will _incorrectly_ disable the re-sync.

lvalerom (approved):
LGTM

Maddosaurus (approved):
Ah, thanks for your comment and the fix!
Completely missed the fact you only set the env var in `/k8s/sensor.sh`, but happy to hear this lead to a fix 😹

PR #4723, opened 2023-02-06
DO NOT REVIEW Dashrews/ROX-14794 updates to handle issues when rocksdb to Postgres migration has issue in the middle

dashrews78:
## Description
Hold off.  I'm working on another solution that is actually much simpler and fits the migrator framework better.

If a migration from RocksDB to Postgres failed in the middle, the migrator would think that Postgres was the database of choice when starting back up and thus have no handles to RocksDB resulting in crashes.  This change will check to see if Postgres has completed the RocksDB to Postgres migration before determining whether to ignore RocksDB or not.  If we have determined that the migration failed in the middle, we will start over with the RocksDB to Postgres migration by moving what we have processed thus far to `central_temp` and starting `central_active` from scratch.  

Initially the plan was to pick up where the migration failed.  This turned out to be riskier than I initially thought so I felt it best to simply re-run the migrations.  The reasons I felt this was safer are:
1. The case where RocksDB migrations have to be executed as well caused additional logic paths to be introduced.  For example going from 3.72 -> 3.74 results in RocksDB migrations being executed but those are not persisted until the end.  So those cases need to start over regardless.
2.  The other case is that in testing I was running into migration issues were Postgres was complaining about unique key violations on the primary key when starting with a partially filled Postgres.  Those would crash central and eventually resolve themselves.  With the number of migrations involved, it felt more prudent to simply start from scratch for this one time migration than to debug and test those cases in each migration.

## Checklist

## Testing Performed

Updated unit and integration tests.  Additionally tested manually but starting 3.72, 3.73, and 3.74 in RocksDB mode with a sizeable workload such that the migration would take a few minutes.  After some time to get data populated, I would upgrade to 3.74 in Postgres mode.  Watching the logs, I would kill the pod at various times during the RocksDB -> Postgres migration to ensure that the migration would start over and complete.

### DIFF
--- a/migrator/clone/postgres/db_clone_manager_impl.go
+++ b/migrator/clone/postgres/db_clone_manager_impl.go
@@ -155,15 +155,12 @@ func (d *dbCloneManagerImpl) GetCloneToMigrate(rocksVersion *migrations.Migratio
 
 	currClone, currExists := d.cloneMap[CurrentClone]
 
-	// If the current Postgres version is less than Rocks version then we need to migrate rocks to postgres
-	// If the versions are the same, but rocks has a more recent update then we need to migrate rocks to postgres
-	// Otherwise we roll with Postgres->Postgres.  We use central_temp as that will get cleaned up if the migration
-	// of Rocks -> Postgres fails so we can start fresh.
+	// If a Rocks database exists, consider if we need to use it to migrate or not.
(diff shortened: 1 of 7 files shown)


dashrews78:
Found a simpler solution.  Will go with #4786 instead

PR #4727, opened 2023-02-06
Check operator generated sources

janisz:
## Description

Migrate operator check generated source to makefile and run it in GHA.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -236,11 +236,20 @@ endef
 ##@ Development
 
 .PHONY: style
-style: generate manifests bundle ## Make sure generated files are up to date
+style: check-manifest check-bundle ## Make sure generated files are up to date
 	@echo 'Checking for diffs after making bundle, generate and manifests...'
-	@echo 'If this fails, check if the invocation of the normalize-metadata.py script in Makefile'
-	@echo 'needs to change due to formatting changes in the generated files.'
-	git diff --exit-code HEAD
+
+
+.PHONY: check-manifest
+check-manifest: manifests generate ## Make sure generated files are up to date
+	@git diff --exit-code HEAD || \
+	 (echo 'If this fails, check generated sources and commit them.' && false)
+
+.PHONY: check-bundle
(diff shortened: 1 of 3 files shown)


janisz:
* master
  * **PR #4727** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4727" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4727?utm_source=stack-comment).

operator/Makefile
```diff
@@ -235,6 +235,12 @@ endef
 
 ##@ Development
 
+.PHONY: style
+style: generate manifests bundle ## Make sure generated files are up to date
+	echo 'Checking for diffs after making bundle, generate and manifests...'
+	echo 'If this fails, check if the invocation of the normalize-metadata.py script in Makefile'
+	echo 'needs to change due to formatting changes in the generated files.'
```
porridge, line 242:
```suggestion
	@echo 'Checking for diffs after making bundle, generate and manifests...'
	@echo 'If this fails, check if the invocation of the normalize-metadata.py script in Makefile'
	@echo 'needs to change due to formatting changes in the generated files.'
```

porridge: Previously this message was only displayed if the `generate` and `manifests` targets caused no diffs *and* the `bundle` target caused some diffs.
Now it is now misleading, since it will steer people towards investigating `normalize-metadata.py` while the error will typically be more benign and caused by not committing changes produced by running `make generate manifests`.

Can you please split the `style` target to `style-manifests` and `style-bundle` with separate checks? It would also be great if the `If this fails, run `make -C operator manifests generate bundle` and commit the result.)` part of the message in the `.sh` file was moved to the target.

In my dream, these message would only be emitted only if the target is about to fail, rather than ever time, but I realize this might be asking for way too much. :-)

porridge (changes requested):

janisz:
It makes no sense. We need another way to migrate static checks out of osci to gha

PR #4729, opened 2023-02-06
sensor integration test helper refactor

lvalerom:
## Description

Refactor the helper functions used in the `sensor-integration-test`

* Enables the helper functions to create resources based on structs.
* Adds the `testRun` struct to hold all the information regarding a test run.
* Adds a `RetryCallback` that will be called (if set) when the creation of a resource fails.

**Note to the reviewers**: Consider reviewing the PR commit by commit.

## Checklist

## Testing Performed

* CI
* `make sensor-integration-test`

### DIFF
--- a/sensor/tests/resource/helper.go
+++ b/sensor/tests/resource/helper.go
@@ -55,16 +55,18 @@ const (
 	defaultTicker = 500 * time.Millisecond
 )
 
-// YamlTestFile is a test file in YAML
-type YamlTestFile struct {
-	Kind string
-	File string
+// K8sResourceInfo is a test file in YAML or a struct
+type K8sResourceInfo struct {
+	Kind     string
+	YamlFile string
+	Obj      interface{}
+	Name     string
 }
 
 // requiredWaitResources slice of resources that need to be awaited
 var requiredWaitResources = []string{"Service"}
 
-func shouldWaitForResource(kind string) bool {
+func shouldRetryResource(kind string) bool {
 	for _, k := range requiredWaitResources {
 		if k == kind {
 			return true
@@ -74,7 +76,7 @@ func shouldWaitForResource(kind string) bool {
 }
 
 // objByKind returns the supported dynamic k8s resources that can be created
-// add new ones here to support adding new resource files
+// add new ones here to support adding new resource types
 func objByKind(kind string) k8s.Object {
 	switch kind {
 	case "Deployment":
@@ -150,6 +152,15 @@ func NewContextWithConfig(t *testing.T, config CentralConfig) (*TestContext, err
 	}, nil
 }
 
+// NewRun runs a test case. Fails the test if the testRun cannot be created.
+func (c *TestContext) NewRun(options ...TestRunFunc) {
+	tr, err := newTestRun(options...)
+	if err != nil {
+		c.t.Fatal(err)
+	}
+	c.run(tr)
+}
+
 // Stop test context and sensor.
 func (c *TestContext) Stop() {
 	c.stopFn()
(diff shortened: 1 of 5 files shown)


sensor/tests/resource/pod/pod_test.go
```diff
@@ -77,115 +77,125 @@ func assertDeploymentContainerImages(images ...string) resource.AssertFunc {
...
-		messages := testC.GetFakeCentral().GetAllMessages()
-		uniquePodNames := resource.GetUniquePodNamesFromPrefix(messages, "sensor-integration", "nginx-")
-		s.Require().Len(uniquePodNames, 3, "Should have received three different pod events")
-	})
+	s.testContext.NewRun(
+		resource.WithResources([]resource.K8sResourceInfo{
+			NginxDeployment,
+		}),
+		resource.WithTestCase(func(t *testing.T, testC *resource.TestContext, objects map[string]k8s.Object) {
+			// wait until pods are created
```
fredrb, line 85: Maybe an idea for another improvement: now that we have these `opts` functions, we could use one `resource.BeforeCase` which provides a function to be executed before the test case and after the resources were created. 

Since there are multiple tests that we wait for `AvailableReplicas == 3` for a resource, this could be a generic function rather than pasted in all test case bodies. 

But this is not for now, just an idea.

sensor/tests/resource/helper.go
```diff
@@ -150,6 +152,43 @@ func NewContextWithConfig(t *testing.T, config CentralConfig) (*TestContext, err
...
+
+// WithRetryCallback sets the RetryCallback function
+func WithRetryCallback(retryCallback RetryCallback) TestRunFunc {
+	return func(t *testRun) {
+		t.retryCallback = retryCallback
+	}
+}
+
+// NewRun runs a test case. Fails the test if the testRun cannot be created.
+func (c *TestContext) NewRun(options ...TestRunFunc) {
```
fredrb, line 184: I think `NewRun` implies the creation of an object (which is the case of `newTestRun`, for example). WDYT about renaming the public function to simply: `Run(...)`? `RunTest` or `Test` sound good also, because they are more imperative.

sensor/tests/resource/role/role_test.go
```diff
@@ -97,37 +99,39 @@ func matchBinding(namespace, id string) resource.MatchResource {
...
+		resource.WithTestCase(func(t *testing.T, testC *resource.TestContext, _ map[string]k8s.Object) {
+			deleteDep, err := testC.ApplyResourceNoObject(context.Background(), "sensor-integration", NginxDeployment, nil)
+			defer utils.IgnoreError(deleteDep)
+			require.NoError(t, err)
 
-		var binding v12.RoleBinding
-		deleteRoleBinding, err := testC.ApplyFile(context.Background(), "sensor-integration", NginxRoleBinding, &binding)
-		defer utils.IgnoreError(deleteRoleBinding)
-		require.NoError(t, err)
+			var binding v12.RoleBinding
```
fredrb, line 108: It's odd that github is showing this `tab` highlighted... Did you accidentally mixed spaces with tabs here or is it just a diff display issue?

sensor/tests/resource/helper.go
```diff
@@ -513,27 +571,45 @@ func createConnectionAndStartServer(fakeCentral *centralDebug.FakeService) (*grp
 	return conn, fakeCentral, closeF
 }
 
-// ApplyFileNoObject - apply a file without an object
-func (c *TestContext) ApplyFileNoObject(ctx context.Context, ns string, file YamlTestFile) (func() error, error) {
-	obj := objByKind(file.Kind)
-	return c.ApplyFile(ctx, ns, file, obj)
+// ApplyResourceNoObject - apply a resource without an object
```
fredrb, line 574: This comments can definitely be improved. I think they were added just to stop lint failures.

Here's my suggestion:
```
// ApplyResourceNoObject creates a Kubernetes resource using `ApplyResource` without requiring
// an object reference.

// ApplyResource creates a Kubernetes resource in namespace `ns` from a resource definition (see 
// `K8sResourceInfo` for more details). Once the resource is applied, the `obj` will be populated
// with the properties from the resource definition. In case the creation fails (due to the client
// API rejecting the definition), a `RetryCallback` function can be provided to manipulate the
// object prior to the retry.
```

fredrb (commented):
Looks good. Thanks for making the change into `opts` functions and moving the retry to the `Apply` functions

sensor/tests/resource/helper.go
```diff
@@ -194,74 +233,93 @@ func (c *TestContext) GetFakeCentral() *centralDebug.FakeService {
...
 	if err != nil {
-		c.t.Fatalf("failed to create namespace: %s", err)
+		return errors.Errorf("failed to create namespace: %s", err)
 	}
 	var removeFunctions []func() error
 	fileToObj := map[string]k8s.Object{}
-	for _, file := range files {
-		obj := objByKind(file.Kind)
-		removeFn, err := c.ApplyFile(context.Background(), DefaultNamespace, file, obj)
+	for i := range resources {
```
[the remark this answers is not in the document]
lvalerom: `resources` is being modified here:

```go
removeFn, err := c.ApplyResource(context.Background(), DefaultNamespace, &resources[i], obj, retryFn)
```

The reason is because we need to set this new `Name` and that depends on whether the resource is defined as a YamlFile or and Obj. If that logic is moved out of `ApplyResource` then it would be possible to use `_, r`. WDYT?

sensor/tests/resource/helper.go
lvalerom: I like `RunTest` or `Test` :)

sensor/tests/resource/helper.go
lvalerom: Agree. I'll update the comment

sensor/tests/resource/role/role_test.go
lvalerom: I think it's because before the changes the indentation was shifted to the left and now we have one more "level" of indentation to the right.

sensor/tests/resource/helper.go
```diff
@@ -194,74 +233,93 @@ func (c *TestContext) GetFakeCentral() *centralDebug.FakeService {
 	return c.fakeCentral
 }
 
-// RunWithResources runs the test case applying files in `files` slice in order.
-func (c *TestContext) RunWithResources(files []YamlTestFile, testCase TestCallback) {
+// run handles the run of a testRun case
```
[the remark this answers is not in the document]
lvalerom: I've updated the comment with a little bit more information. Let me know what you think.

sensor/tests/resource/helper.go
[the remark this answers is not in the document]
fredrb: Today I learned that the 2nd value from `range` is copying the value. Maybe if we use a slice of references you could pass the `r` there. But I think the way you did it right now is fine.

> The reason is because we need to set this new Name and that depends on whether the resource is defined as a YamlFile or and Obj. If that logic is moved out of ApplyResource then it would be possible to use _, r. WDYT?

Up to you, really. I don't have a strong opinion here.

sensor/tests/resource/helper.go
```diff
@@ -194,74 +233,94 @@ func (c *TestContext) GetFakeCentral() *centralDebug.FakeService {
 	return c.fakeCentral
 }
 
-// RunWithResources runs the test case applying files in `files` slice in order.
-func (c *TestContext) RunWithResources(files []YamlTestFile, testCase TestCallback) {
+// run calls the proper test function depending on the configuration of the testRun.
```
fredrb, line 236:
```suggestion
// run calls the test function depending on the configuration of the testRun.
```
Reads better IMO

fredrb (approved):

sensor/tests/resource/helper.go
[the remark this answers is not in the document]
lvalerom: I think I'll keep it as it is right now

PR #4731, opened 2023-02-06
ROX-14777: Update Postgres coming soon message

vjwilson:
## Description

One more round of word-smithing

## Checklist

## Testing Performed

Updated message when using an environment running RocksDB
displayed_when_DB_is_NOT_Postgres

Allowing text wrapping on small screen widths, to account for longer message
displayed_when_DB_is_NOT_Postgres_narrow_screen

No longer a message, after message is dismissed and page is refreshed (this screenshot is just to verify that there was no regression)
not_displayed_when_dismissed

No message when using an environment running Postgres
not_displayed_when_DB_is_Postgres

### DIFF
--- a/ui/apps/platform/src/Containers/MainPage/AnnouncementBanner.tsx
+++ b/ui/apps/platform/src/Containers/MainPage/AnnouncementBanner.tsx
@@ -1,10 +1,13 @@
 import React, { ReactElement, useEffect, useState } from 'react';
 import { AlertVariant, Banner, Button } from '@patternfly/react-core';
 
+import { fetchDatabaseStatus } from 'services/DatabaseService';
+
 const ANNOUNCEMENT_BANNER_KEY = 'postgresAnnouncementBannerDismissed';
 
 function AnnouncementBanner(): ReactElement | null {
     const [isDisplayed, setIsDisplayed] = useState(false);
+    const [databaseType, setDatabaseType] = useState('');
 
     useEffect(() => {
         const localStorageValue = localStorage.getItem(ANNOUNCEMENT_BANNER_KEY);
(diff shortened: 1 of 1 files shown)


kcarmichael08 (commented):
LGTM!

ui/apps/platform/src/Containers/MainPage/AnnouncementBanner.tsx
```diff
@@ -1,35 +1,49 @@
...
                 className="pf-u-display-flex pf-u-justify-content-center pf-u-align-items-center"
                 isSticky
                 variant={AlertVariant.info}
+                style={{ whiteSpace: 'normal' }}
             >
                 <span className="pf-u-text-align-center">
-                    The next version of this product will be version 4.0.0. Central will be using
-                    Postgres for its data store starting in v4.0.0. You must backup your database
-                    before upgrading to 4.0.0.
+                    Red Hat Advanced Cluster Security plans to change its database to PostgreSQL in
```
msugakov, line 44: Nitpicking: "Red Hat Advanced Cluster Security" is valid for RHACS_BRANDING. Could you please use branded product name so that "StackRox" shows up for community builds?

vjwilson: This would be different than our current practice, where the UI does not change for anything at runtime. Only the logo changes at build time, based on ACS vs. opensource.

msugakov: I believe that info is available at build time. Legitimately hardcoded here:
https://github.com/stackrox/stackrox/blob/baada33c57ee29ae0527d470bc31be3f0143cd68/ui/apps/platform/src/constants/productBranding.ts#L33

bradr5 (approved):

knewcomerRH:
LGTM

PR #4734, opened 2023-02-06
Reduce the number of confusing instances of `stackrox.io` in charts

msugakov:
## Description

@BradLugo got confused seeing `registry: stackrox.io` in the generated Helm values.
While that confusion is fair, Helm values files are meant to be examples illustrating usage rather than the set of currently selected values. I myself would prefer helm value files to give **actual** defaults, but this can easily make this change XL-sized.
Instead, I simply get rid of `stackrox.io` mentions where they may be confusing.

Context thread
https://srox.slack.com/archives/CELUQKESC/p1675462004575229?thread_ts=1675459102.811779&cid=CELUQKESC

## Checklist

None of these needed:
- [ ] Unit test and regression tests added - the change does not deserve ones since values files are examples and not executables.

## Testing Performed

* Manually:
```bash
$ make cli
$ rm -rf stackrox-central-services-chart
$ rm -rf stackrox-secured-cluster-services-chart
$ roxctl helm output central-services --image-defaults=rhacs

$ grep -F '#image:' -B1 -A4 stackrox-central-services-chart/values.yaml
## Common settings for all image properties
#image:
#  # The image registry to use. Unless overridden in the more specific configs, this
#  # determines the base registry for each image referenced in this config file.
#  registry: registry.redhat.io/advanced-cluster-security
#

$ grep -F '# image:' -A7 stackrox-central-services-chart/values-public.yaml.example
# image:
#   # The registry relative to which all image references are resolved, unless
#   # a specific registry is provided for particular workloads which takes precedence
#   # (see `central.image`, `db.image`, `scanner.image`, and `scanner.dbImage` below).
#   # This can be just a registry hostname such as `stackrox.io`, or a registry hostname with
#   # a "remote" component such as `us.gcr.io/my-stackrox-mirror`.
#   registry: us.gcr.io/my-stackrox-mirror
#

$ grep -F '`image.registry`:' stackrox-central-services-chart/values-public.yaml.example
# - `image.registry`: if you are pulling images from a registry other than `registry.redhat.io/advanced-cluster-security`.

$ roxctl helm output secured-cluster-services --image-defaults=rhacs
$ grep -F '# Add registry defaults' -A6 stackrox-secured-cluster-services-chart/internal/defaults/50-images.yaml
# Add registry defaults
image:
  main:
    registry: {{ ._rox.image.registry }}
  collector:
    registry: {{ ._rox.image.registry }}
```

### DIFF
--- a/image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
+++ b/image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
@@ -24,7 +24,7 @@ image:
   main:
     registry: {{ ._rox.image.registry }}
   collector:
-    registry: {{ if or (eq ._rox.image.registry "stackrox.io") (eq ._rox.image.registry "registry.connect.redhat.com") }}collector.stackrox.io{{ else }}{{ ._rox.image.registry }}{{ end }}
+    registry: {{ ._rox.image.registry }}
   [< if (not .KubectlOutput) >]
   scanner:
     registry: {{ ._rox.image.registry }}
--- a/image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
+++ b/image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
@@ -24,7 +24,7 @@ image:
   main:
     registry: {{ ._rox.image.registry }}
   collector:
-    registry: {{ ._rox.image.registry }}
+    registry: {{ if or (eq ._rox.image.registry "stackrox.io") (eq ._rox.image.registry "registry.connect.redhat.com") }}collector.stackrox.io{{ else }}{{ ._rox.image.registry }}{{ end }}
   [< if (not .KubectlOutput) >]
   scanner:
     registry: {{ ._rox.image.registry }}
--- a/image/templates/README.md
+++ b/image/templates/README.md
@@ -3,7 +3,7 @@
 The currently maintained helm charts consists of the `stackrox-central-services`
 and `stackrox-secured-cluster-services` charts.
 
-Helm charts are distributed by `https://charts.stackrox.io` and [stackrox/helm-charts](https://github.com/stackrox/helm-charts).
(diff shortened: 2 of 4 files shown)


image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
```diff
@@ -24,7 +24,7 @@ image:
   main:
     registry: {{ ._rox.image.registry }}
   collector:
-    registry: {{ if or (eq ._rox.image.registry "stackrox.io") (eq ._rox.image.registry "registry.connect.redhat.com") }}collector.stackrox.io{{ else }}{{ ._rox.image.registry }}{{ end }}
```
msugakov, line 27: These are old remnants related to `stackrox.io` -> `collector.stackrox.io` which aren't needed. Cleaning this up since I found it too.

clickboo (approved):

BradLugo:
Would it make sense to change the default registry to `quay.io/stackrox-io`? I see that's what we use for the default in our [deploy/common/deploy.sh](https://github.com/stackrox/stackrox/blob/master/deploy/common/deploy.sh#L3). If so, I think we can change the default here: https://github.com/stackrox/stackrox/blob/master/scripts/ci/lib.sh#L602

That way users will have a working installation by default. Open to opinions on this

image/templates/helm/stackrox-secured-cluster/internal/defaults/50-images.yaml.htpl
msugakov: Unfortunately, this broke unit tests and requires deeper cleanups. Let me undo the change in this file and defer to the bigger cleanup https://issues.redhat.com/browse/ROX-11642

msugakov:
@BradLugo I might misunderstand your suggestion.
The way I read the `push_helm_charts()` function: it generates three sets of Helm charts for `stackrox.io`, for `rhacs` and for `opensource` image flavors. The respective image registries will be stackrox.io, registry.redhat.io/advanced-cluster-security and quay.io/stackrox-io.
What you seem to be asking should be covered by `--image-defaults=opensource` lines. Just need to figure out where do these charts get published.
Does this make sense?

BradLugo:
Yes, that makes sense. Thank you for explaining

BradLugo (approved):

PR #4738, opened 2023-02-06
ROX-14175: Wait for deployments to be deleted before trying to delete namespace in RiskTests

keyallis:
## Description

This was the proposed solution, with the assumption being that the namespace failing deletion is causing failures and could be avoided by ensuring that the namespace is empty at the time the deletion is attempted.

## Checklist

## Testing Performed

As this is a flake, reproduction will be difficult. This change will be done and then monitored for any persistent issues. Further solutions will be explored in the case that this does not solve the issue at hand.


theencee (approved):

gavin-stackrox (approved):

keyallis:
Unrelated flake, updated Jira tracker for it

PR #4740, opened 2023-02-06
ROX-14582: Wire up network graph polling

vjwilson:
## Description

This implements the "nodes-were-added-click-to-update" functionality of the old Network Graph, on the new Network Graph.

There were no mocks for how this should look in the new format. On the old graph, there was a floating rectangle near the upper-right corner of the graph, with the timestamp when the graph was last updated. If the number of nodes available for the graph changed, a count of the difference of nodes was added below the timestamp, and the whole thing became a clickable area to force a graph update.

On the new graph layout, the mocks show the timestamp in the far right of the lowest toolbar. There is no mock for the 

## Checklist

## Testing Performed

Steps to test
(shown in the screen recording attached below)
1. I selected the Production cluster, and the backend and default namespaces, on staging.
2. I deployed a new deployment, `nginx`, in that default NS.
3. I waited about 30 seconds for the polling to happen, and then I observed the timestamp changed changed to a button labeled "1 update available".
4. I deployed the Kubernetes `shell-demo` in that default NS.
5. I waited another 30 seconds for the polling to happen, and then I observed the button labeled change from "1 update available" to "2 updates available".
6. I clicked that update, and the graph updated to show the 2 new deployments in the default NS and a new timestamp in place of the button.
7. I deleted the shell-demo pod.
8. I waited about 30 seconds for the polling to happen, and then I observed the timestamp changed changed to a button labeled "1 update available".
9. I clicked that update, and the graph updated to show the shell-demo deployment gone in the default NS and a new timestamp in place of the button.

https://user-images.githubusercontent.com/715729/217105928-4f4f972c-25fa-4977-9ef2-01c4ce25893f.mov

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
     PageSection,
     Title,
@@ -17,12 +17,17 @@ import { timeWindows } from 'constants/timeWindows';
 import useFetchClusters from 'hooks/useFetchClusters';
 import useFetchDeploymentCount from 'hooks/useFetchDeploymentCount';
 import useURLSearch from 'hooks/useURLSearch';
-import { fetchNetworkFlowGraph, fetchNetworkPolicyGraph } from 'services/NetworkService';
+import {
+    fetchNetworkFlowGraph,
+    fetchNetworkPolicyGraph,
+    fetchNodeUpdates,
+} from 'services/NetworkService';
 import queryService from 'utils/queryService';
 import timeWindowToDate from 'utils/timeWindows';
 import { isCompleteSearchFilter } from 'utils/searchUtils';
 
 import PageTitle from 'Components/PageTitle';
+import useInterval from 'hooks/useInterval';
 import useURLParameter from 'hooks/useURLParameter';
 import NetworkGraphContainer, { Models } from './NetworkGraphContainer';
 import EmptyUnscopedState from './components/EmptyUnscopedState';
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
```diff
@@ -101,6 +108,26 @@ function NetworkGraphPage() {
     const selectedCluster = { name: clusterFromUrl, id: selectedClusterId };
     const { deploymentCount } = useFetchDeploymentCount(selectedClusterId || '');
 
+    const [prevEpochCount, setPrevEpochCount] = useState(0);
+    const [currentEpochCount, setCurrentEpochCount] = useState(0);
+
+    // We will update the poll epoch after 30 seconds to update the node count for a cluster
+    useInterval(() => {
+        setPollEpoch(pollEpoch + 1);
+    }, 30000);
```
sachaudh, line 117: I'm curious about the use of interval vs. timeout. In this scenario we could have this possibility:
1. Interval updates
2. API is triggered
3. API still hasn't returned within 30 seconds due to possible network slowdown
4. Interval updates

Would it be better to apply a timeout to call the API again after a successful response rather than use an interval?

sachaudh (commented):
LGTM. Had a question though

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.tsx
vjwilson: While a timeout that re-initializes itself would be safer for short time periods, API calls that don't return within 30 seconds are indicative of a larger problem.

We are adopted this useInterval hook in seven other sections of the app that poll, to replace the redux-saga version of polling, such as in this case `yield call(delay, 30000);`

sachaudh (approved):

PR #4764, opened 2023-02-07
ROX-14953: Fix invalid kubectl files within roxctl generate

dhaus67:
## Description

This PR fixes an issue when trying to generate kubectl files for central and using the default values.

Specifically, the declarative configuration default values are the culprit.

Leaving the values as:
```
Enter list of config maps to add as declarative configuration mounts in central (default: "[]"):
Enter list of secrets to add as declarative configuration mounts in central (default: "[]"):
```

will lead to the following created mounts:
```
name: []
mountPath: /run/stackrox.io/declarative-configuration/[]
readOnly: true
name: []
mountPath: /run/stackrox.io/declarative-configuration/[]
readOnly: true
name: []
mountPath: /run/stackrox.io/declarative-configuration/[]
readOnly: true
name: []
mountPath: /run/stackrox.io/declarative-configuration/[]
readOnly: true
```

These are invalid names, hence applying the bundle fails.

_Note: this only applies to the kubectl format, the Helm format works as expected_

Additionally, during investigation, I noticed that in interactive mode + kubectl format, the values of the array values were duplicated. I've created [ROX-14956](https://issues.redhat.com/browse/ROX-14956) as a follow-up to address this, and for now de-duplicated the values.

Last but not least, the flags should have been behind the feature flag from the beginning, so this was done in this PR as well, including adjusting the `interactive` roxctl tests to reflect this.
 
## Checklist

## Testing Performed
- additional unit tests.
- e2e interactive tests adjusted for the removed flags.

### DIFF
--- a/pkg/renderer/kubernetes.go
+++ b/pkg/renderer/kubernetes.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stackrox/rox/pkg/images/defaults"
 	imageUtils "github.com/stackrox/rox/pkg/images/utils"
 	kubernetesPkg "github.com/stackrox/rox/pkg/kubernetes"
+	"github.com/stackrox/rox/pkg/sliceutils"
 	"github.com/stackrox/rox/pkg/utils"
 	"github.com/stackrox/rox/pkg/zip"
 )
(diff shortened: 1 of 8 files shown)


dhaus67:
* master
  * **PR #4764** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4764" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4764?utm_source=stack-comment).

pkg/renderer/kubernetes.go
```diff
@@ -103,6 +104,15 @@ func postProcessConfig(c *Config, mode mode, imageFlavor defaults.ImageFlavor) e
 		}
 	}
 
+	// Currently, when the K8S config is generated through interactive mode, the configuration flags will be called twice.
+	// This doesn't affect single value configurations, like booleans and strings, but slices.
+	// TODO(ROX-14956):Once the duplication of flag values is removed, this can be removed.
+	c.K8sConfig.DeclarativeConfigMounts.ConfigMaps = sliceutils.Unique(c.K8sConfig.DeclarativeConfigMounts.ConfigMaps)
```
ivan-degtiarenko, line 110: Are `c.K8sConfig.DeclarativeConfigMounts` always non-nil?

ivan-degtiarenko (approved):
Looks good modulo 2 comments. Approving as not to block the release

pkg/renderer/render_test.go
```diff
@@ -145,6 +146,158 @@ func TestRenderWithDeclarativeConfig(t *testing.T) {
 	}
 }
 
+func TestRenderDeclarativeConfigEmpty(t *testing.T) {
+	flavor := testutils.MakeImageFlavorForTest(t)
+	cases := map[string]Config{
+		"empty declarative config mounts": {
```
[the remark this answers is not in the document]
dhaus67: That was one of the issues as well: on one hand, we had empty values (i.e. `array == nil/uninitialized`), on the other hand, cobra specifies an empty array as the string `[]`. Due to the duplication of the values, we ended up with an array that looked like `[]string{"[]"}` for the default value - hence the special case test here to cover this one as well.

pkg/renderer/kubernetes.go
dhaus67: Currently yes.
But, even if they are, `sliceutils.Unique/Without` can both handle nil arrays, so no special casing needed here.

PR #4767, opened 2023-02-07
Default to empty DB Spec if unset in CR and update Kuttl tests

connorgorman:
## Description

This was resulting in is db isEnabled being false

## Checklist

## Testing Performed

The operator e2e tests should pass

### DIFF
--- a/operator/tests/central/basic/76-switch-back-zzz-deletes.yaml
+++ b/operator/tests/central/basic/76-switch-back-zzz-deletes.yaml
@@ -1,0 +1,15 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+# Clean up the external DB password secret, we no longer refer to it in this step.
+- apiVersion: v1
+  kind: Secret
+  name: my-central-db-password
+# The previous step (75-switch-to-external-central-db) has caused the initially generated
+# central-db-password secret to be overwritten with a manually specified "external" password.
+# The DB PCV was left behind, but it's difficult to extract the password from there.
+# Instead, we now remove the central DB PVC, to cause the operator create a new one such that
+# it can be populated with a freshly generated DB password.
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  name: central-db
--- a/operator/tests/central/basic/76-switch-back-zzz-deletes.yaml
+++ b/operator/tests/central/basic/76-switch-back-zzz-deletes.yaml
@@ -7,7 +7,7 @@ delete:
   name: my-central-db-password
 # The previous step (75-switch-to-external-central-db) has caused the initially generated
 # central-db-password secret to be overwritten with a manually specified "external" password.
-# The DB PCV was left behind, but it's difficult to extract the password from there.
+# The DB PVC was left behind, but it's difficult to extract the password from there.
 # Instead, we now remove the central DB PVC, to cause the operator create a new one such that
 # it can be populated with a freshly generated DB password.
 - apiVersion: v1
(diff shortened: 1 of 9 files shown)


gavin-stackrox (approved):

porridge:
I tried making an independent PR, but it turns out that before your changes we were not even spawning a `central-db` deployment :facepalm:

So I took the liberty of pushing a fix commit to your branch @connorgorman 

```
|     Central operator e2e test: fix races and step 76.
|     
|     - in steps 40 and 76, which check whether central deployment *still* has
|       available replicas, introduce a numbered test-step annotation that is asserted
|       to make sure we're not looking at the state of the deployment as it was
|       reported by the deployment controller before it even saw the current step's
|       modifications
|     - in step 40, make sure the central-db deployment is actually there (now that
|       the previous commit actually makes that true), and fetch its logs on failure
|     - in step 75 (that switches to an external db) make sure the central-db
|       deployment is gone
|     - fix step 76: generating a new DB password secret has no effect on the
|       password seen by postgres, since it's only set once, on DB initialization.
|       Drop the DB PVC to pick up the newly generated password.
```


/test openshift-newest-qa-e2e-tests
/test openshift-oldest-qa-e2e-tests
/test openshift-penultimate-qa-e2e-tests

vladbologa:
> |     - fix step 76: generating a new DB password secret has no effect on the
> |       password seen by postgres, since it's only set once, on DB initialization.
> |       Drop the DB PVC to pick up the newly generated password.

Related to the above, but more generally @porridge 

If we change the DB password secret, or the external DB config in the Central CR (in which case the operator picks up the change and updates the ConfigMap), Central will not detect these changes, because it only reads them on startup.

This makes secret rotation problematic, among other things. We should either:
- force a Central restart when this happens, but I don't know if this possible from the operator?
- have Central always re-read this config when making new DB connections. Not sure of the implications of this cc @dashrews78  @c-du

porridge:
> force a Central restart when this happens, but I don't know if this possible from the operator?

It might be a bit complicated to do in a reliable but idempotent way... :thinking:

vladbologa:
> > force a Central restart when this happens, but I don't know if this possible from the operator?
> 
> It might be a bit complicated to do in a reliable but idempotent way... 🤔

I see. Anyway, the better solution that avoids downtime would be for Central to update its external DB config on the fly.

porridge:
This looks good to me, I *think* the failure is unrelated, let's retry, but I wouldn't wait for this to finish before merging if my change LGTY @connorgorman 
/retest

porridge:
Looks like that `RoutesTest` failure is unrelated, also [happened on last nightly](https://github.com/stackrox/stackrox/commit/26d131c9e8d28692a1d3c450441d93da3a74b718).

operator/tests/central/basic/76-switch-back-zzz-deletes.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+# Clean up the external DB password secret, we no longer refer to it in this step.
+- apiVersion: v1
+  kind: Secret
+  name: my-central-db-password
+# The previous step (75-switch-to-external-central-db) has caused the initially generated
+# central-db-password secret to be overwritten with a manually specified "external" password.
+# The DB PCV was left behind, but it's difficult to extract the password from there.
```
connorgorman, line 10:
```suggestion
# The DB PVC was left behind, but it's difficult to extract the password from there.
```

PR #4658, opened 2023-02-01
Bump apollo-ci image to 0.3.56

janisz:
https://github.com/stackrox/rox-ci-image/releases/tag/0.3.56
https://github.com/openshift/release/pull/36092

PR #4768, opened 2023-02-07
Revert "Bump apollo-ci image to 0.3.56"

gavin-stackrox:
Reverts stackrox/stackrox#4658 which added a `-dirty` tag to operator builds.
https://srox.slack.com/archives/CELUQKESC/p1675807797443359

PR #4769, opened 2023-02-08
go1.19.4

RTann:
## Description

https://github.com/stackrox/stackrox/pull/4658 started the bump from go1.18.4 to 1.19.4, but it missed `EXPECTED_GO_VERSION` and generating new proto files which were re-styled due to the version bump. This PR hopefully completes the update to go1.19.4.

That original PR was reverted due to test failures. This has been resolved via https://github.com/openshift/release/pull/36193

## Checklist

## Testing Performed

CI

I updated `EXPECTED_GO_VERSION` and then ran both `make proto-generated-srcs` and `gogen .`


RTann:
https://github.com/openshift/release/blob/master/ci-operator/config/stackrox/stackrox/stackrox-stackrox-master.yaml#L26 will need to be updated for this to work

janisz (approved):

generated/storage/group.pb.go
```diff
@@ -98,10 +98,11 @@ func (m *Group) Clone() *Group {
 
 // GroupProperties defines the properties of a group. Groups apply to users when
 // their properties match. For instance:
-//   * If GroupProperties has only an auth_provider_id, then that group applies
```
rhybrillou, line 101: I suppose the comments in the source proto files should be updated and pushed as well.

BUILD_IMAGE_VERSION
```diff
@@ -1 +1 @@
-stackrox-build-0.3.55
```
rhybrillou, line 1: Has the issue at the origin of the revert https://github.com/stackrox/stackrox/pull/4768 been understood and resolved ?

janisz: `-dirty` was added because `generated/` files were changed

generated/storage/group.pb.go
rhybrillou: Looking at the `grouped-static-checks` tests, I think the styling changes are required in the proto files to ensure the code re-generation does not erase the fixes later.

RTann: sure we can be on the safe side

rhybrillou (approved):
Looks good.
Please make sure all the proto files at the source of the changed files under `generated` also contain the changes.

0x656b694d:
Fixes ROX-12582.

PR #4779, opened 2023-02-08
ROX-14775 Delete the status field if present when encoding NeworkPolicies to YAML

lvalerom:
## Description

Kubernetes added a 'status' field for NetworkPolicies in 1.24. See: [v1.24 Changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#api-change-3)

Using the `NewSerializerWithOptions` from `k8s.io/apimachinery/pkg/runtime/serializer/json` with a `v1.NetworkPolicy` will return the `status` field even if the status is empty. This is because the type `Status` is defined as a value instead of a pointer so it always gets serialized in json.

The problem is: Old cluster using kubernetes < 1.24 will fail to apply NetworkPolicies generated this way. See: [ROX-14775](https://issues.redhat.com/browse/ROX-14775)

Since the `status` field is not handled by ACS, we can delete the field manually if present.

Note: This might not be necessary in the future since the feature was withdrawn by the sig-network. See: [NetworkPolicy Status](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2943-networkpolicy-status#implementation-history)

## Checklist

## Testing Performed

* [x] CI
* [x] Unit tests for this case have been added
* [x] Manual testing `roxctl generate netpol`
* [x] Manual testing `Simulate network policy` in the UI

### DIFF
--- a/pkg/protoconv/networkpolicy/k8s.go
+++ b/pkg/protoconv/networkpolicy/k8s.go
@@ -24,14 +24,6 @@ type KubernetesNetworkPolicyWrap struct {
 
 // ToYaml produces a string holding a JSON formatted yaml for the network policy.
 func (np KubernetesNetworkPolicyWrap) ToYaml() (string, error) {
-	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(np.NetworkPolicy)
-	if err != nil {
-		return "", err
-	}
-	encoder := json.NewSerializerWithOptions(json.DefaultMetaFactory, nil, nil, json.SerializerOptions{
-		Yaml: true,
-	})
-
 	// Kubernetes added a 'status' field for NetworkPolicies in 1.24. See:
 	// * (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#api-change-3)
 	// * (https://github.com/kubernetes/kubernetes/pull/107963)
@@ -41,7 +33,15 @@ func (np KubernetesNetworkPolicyWrap) ToYaml() (string, error) {
 	// This code might not be necessary in the future since the feature was withdrawn by the sig-network. See:
 	// * (https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2943-networkpolicy-status#implementation-history)
 	// * (https://github.com/kubernetes/kubernetes/pull/107963#issuecomment-1400220883)
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(np.NetworkPolicy)
+	if err != nil {
+		return "", err
+	}
 	delete(uObj, "status")
+
+	encoder := json.NewSerializerWithOptions(json.DefaultMetaFactory, nil, nil, json.SerializerOptions{
+		Yaml: true,
+	})
 
 	stringBuilder := &strings.Builder{}
 	err = encoder.Encode(&unstructured.Unstructured{Object: uObj}, stringBuilder)
--- a/pkg/protoconv/networkpolicy/k8s.go
+++ b/pkg/protoconv/networkpolicy/k8s.go
@@ -11,6 +11,8 @@ import (
 	k8sCoreV1 "k8s.io/api/core/v1"
 	networkingV1 "k8s.io/api/networking/v1"
 	k8sMetaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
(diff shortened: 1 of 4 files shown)


msugakov:
@lvalerom I saw @misberner used unstructured objects in https://github.com/stackrox/stackrox/pull/3371 as a way to deal with `status: {}` thingy. I haven't looked there myself but maybe you can check if this would work more elegantly than re-serializing.

lvalerom:
> @lvalerom I saw @misberner used unstructured objects in #3371 as a way to deal with `status: {}` thingy. I haven't looked there myself but maybe you can check if this would work more elegantly than re-serializing.

@msugakov I haven’t looked into the PR in too much detail (I will) but I think the case with that one was to deserialize objects and then validate them against the OpenAPI. The case here is a little bit different because we want to serialize `storage.NetworkPolicy` into yaml, but the idea of using unstructured looks like a viable solution. At this moment, I believe we have three options:
1. Convert the `v1.NetworkPolicy` type into unstructured.
2. Avoid using the `v1.NetworkPolicy` type and just convert `storage.NetworkPolicy` into unstructured.
3. Re-serialize.

I think the first option is my least favorite because we still need to delete the status field manually. Converting a `v1.NetworkPolicy` into unstructured will add the status field and, if we serialize that, we run into the same problem. Also, converting static types into unstructured basically does serialize and deserialize into json (or makes heavy use of reflection) which would be pretty much the same as option 3 under the hood.

The only problem with option 2 is that we need to be careful when we instantiate the unstructured object because it won’t validate invalid fields (e.g. if the apiVersion is empty it will still set it, thus making the yaml have an empty apiVersion which is not correct).

I think option 2 is the way to go but I would like to hear your thoughts.

pkg/protoconv/networkpolicy/k8s.go
```diff
@@ -22,10 +24,27 @@ type KubernetesNetworkPolicyWrap struct {
...
+	// Kubernetes added a 'status' field for NetworkPolicies in 1.24. See:
+	// * (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#api-change-3)
+	// * (https://github.com/kubernetes/kubernetes/pull/107963)
+	// Using the `NewSerializerWithOptions` from `k8s.io/apimachinery/pkg/runtime/serializer/json` with a `v1.NetworkPolicy` will return the `status` field.
+	// The problem is: Old cluster using kubernetes < 1.24 will fail to apply NetworkPolicies generated this way.
+	// Since the `status` field is not handled by ACS, we delete the field manually if present here.
+	// This code might not be necessary in the future since the feature was withdrawn by the sig-network. See:
+	// * (https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2943-networkpolicy-status#implementation-history)
+	// * (https://github.com/kubernetes/kubernetes/pull/107963#issuecomment-1400220883)
+	delete(uObj, "status")
```
msugakov, line 44: All looks correct but I'd suggest to change the order so that comment applies to both Unstructured and removal, and `encoder` stays outside of that.

```go
	// Kubernetes added a 'status' field for NetworkPolicies in 1.24. See:
	// * (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#api-change-3)
	// * (https://github.com/kubernetes/kubernetes/pull/107963)
	// Using the `NewSerializerWithOptions` from `k8s.io/apimachinery/pkg/runtime/serializer/json` with a `v1.NetworkPolicy` will return the `status` field.
	// The problem is: Old cluster using kubernetes < 1.24 will fail to apply NetworkPolicies generated this way.
	// Since the `status` field is not handled by ACS, we delete the field manually if present here.
	// This code might not be necessary in the future since the feature was withdrawn by the sig-network. See:
	// * (https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2943-networkpolicy-status#implementation-history)
	// * (https://github.com/kubernetes/kubernetes/pull/107963#issuecomment-1400220883)
	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(np.NetworkPolicy)
	if err != nil {
		return "", err
	}
	delete(uObj, "status")

	encoder := json.NewSerializerWithOptions(json.DefaultMetaFactory, nil, nil, json.SerializerOptions{
		Yaml: true,
	})
```

msugakov (commented):
The main code looks ok with 1 comment. Haven't checked tests yet.

pkg/protoconv/networkpolicy/k8s.go
lvalerom: Makes sense. I've changed the order as you suggested.

msugakov:
@lvalerom The main code looks ok, but I had few comments about tests that I thought would be quicker to implement myself than explain. I made that and directly pushed onto your branch. Please give it a read, and feel free to edit or even undo that as you like.

msugakov (approved):

lvalerom:
Thanks @msugakov for refactoring the tests. I think you’re right, they are more readable this way. As soon as the CI gives us the green light I’ll merge this.

Maddosaurus (approved):
I agree about the readability and maintainability of mega-tests.
WRT failing Openshift tests, you might need to rebase on lastest master.
AFAIK, there have been quite some changes recently.

lvalerom:
> WRT failing Openshift tests, you might need to rebase on lastest master.
AFAIK, there have been quite some changes recently.

Rebase done. Let's see if the CI likes it 🙂

msugakov:
@lvalerom Just investigate if failing tests have known JIRA tickets and proceed merging despite these.

lvalerom:
> @lvalerom Just investigate if failing tests have known JIRA tickets and proceed merging despite these. No need to rerun "until it succeeds" because it may not succeed.

* `RoutesTest: Verify that routes are detected correctly` -> [ROX-14004](https://issues.redhat.com/browse/ROX-14004)
* `K8sRbacTest: Add Service Account and verify it gets scraped` x2 -> [ROX-15061](https://issues.redhat.com/browse/ROX-15061)

PR #4783, opened 2023-02-08
Update scanner version to 2.28.2

daynewlee:
## Description

Update scanner version to 2.28.2

## Checklist

## Testing Performed

CI is sufficient


tommartensen:
@daynewlee @RTann this PR should be targeting the `release-3.74` branch directly. 
Note that OSCI will not notice if you change the base branch in the GH PR, so you'll need to open a new PR.

RTann:
@tommartensen why does it need to target the release branch? Do we not just want to cherry-pick this?

tommartensen:
My understanding is that the SCANNER_VERSION on master should be the floating tag, e.g. `2.28.x-...`. Merging to master + cherry-picking, and later changing to the floating tag are just additional steps that I thought we could avoid :)

RTann:
> My understanding is that the SCANNER_VERSION on master should be the floating tag, e.g. `2.28.x-...`.

Not necessarily. Just the latest version

> Merging to master + cherry-picking, and later changing to the floating tag are just additional steps that I thought we could avoid :)

That's fine, but I am under the impression cherry-picking is the preferred method when master has not diverged too far from the release branch. I am pretty sure that is what we have been doing for past releases. Direct merges were typically reserved for older releases or commits which would have been difficult to cherry-pick due to master diverging too far from the release.

I will close this PR for @daynewlee since https://github.com/stackrox/stackrox/pull/4797 was merged

PR #4784, opened 2023-02-08
ROX-14987: Fix wrong info for anomalous external flows count

sachaudh:
## Description

This PR fixes an issue with the deployment details tab showing the wrong info for the number of external flows for this deployment

Screenshot 2023-02-08 at 10 30 06 AM

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.test.ts
+++ b/ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.test.ts
@@ -1,0 +1,106 @@
+import { CustomEdgeModel } from '../types/topology.type';
+import { getNumFlowsFromEdge } from './networkGraphUtils';
+
+describe('networkGraphUtils', () => {
+    describe('getNumFlowsFromEdge', () => {
+        it('should get number of flows for a non-bidirectional edge with one port/protocol', () => {
+            const edge: CustomEdgeModel = {
+                id: 'edge-1',
+                type: 'edge',
+                visible: true,
+                source: 'node-1',
+                target: 'node-2',
+                data: {
+                    isBidirectional: false,
+                    portProtocolLabel: '8080 TCP',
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+                            protocol: 'L4_PROTOCOL_TCP',
+                            lastActiveTimestamp: '2023-02-08T18:24:50.470158762Z',
+                        },
+                        {
+                            port: 80,
+                            protocol: 'L4_PROTOCOL_TCP',
+                            lastActiveTimestamp: '2023-02-08T18:24:50.470158762Z',
+                        },
+                    ],
+                    targetToSourceProperties: [
```
sachaudh, line 85: Considering that this field is only present when the `isBidirectional` value is true, maybe we can tighten up our typing for the `CustomEdgeModel` to reflect that. Right now we just have `targetToSourceProperties` as an optional field

alwayshooin: i think that makes sense, maybe have two different edge types to accommodate for `DirectionalEdge` and `BidirectionalEdge`

ui/apps/platform/src/Containers/NetworkGraph/utils/networkGraphUtils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+                    tag: '4',
+                },
+            };
+
+            const numFlows = getNumFlowsFromEdge(edge);
+
+            expect(numFlows).toEqual(4);
+        });
+    });
+});
```
alwayshooin, line 106: tests 🥲

alwayshooin (approved):

PR #4785, opened 2023-02-08
ROX-14766: Add olm-rollback-test

kylape:
Goals:

* Validate an OLM upgrade that has at least one intermediate upgrade
  * Example: 3.72.0 -> 3.74.0 has an intermediate upgrade of 3.73.0
* Figure out how to perform a rollback with the operator

Current high level steps this test performs:

1. Deploy OLM if not present
2. Deploy verion 3.72.0 of the operator
3. Deploy a Central and do very basic API validation
4. Upgrade operator to 3.74.x
5. Wait for Central to upgrade and do very basic API validation
6. Downgrade operator back to 3.72 and wait for Central to get downgraded

A few points:

* Test uses rhacs-eng images to test upgrade to unreleased version
* Install plans are configured to be automatically approved because this is how most customers have it configured (afaik)
* Rollback test is currently incomplete.
* Testing has been performed on default-gke k8s clusters

Questions:

* Does it make sense to keep an upgrade test like this in the repo?
* Should this script be refactored into an existing test suite (e.g. kuttl)?

## Description

## Checklist

### DIFF
--- a/operator/hack/common.sh
+++ b/operator/hack/common.sh
@@ -114,6 +114,42 @@ function approve_install_plan() {
   retry 3 5 kubectl -n "${operator_ns}" patch installplan "${install_plan_name}" --type merge -p '{"spec":{"approved":true}}'
 }
 
+function csv_version_tag() {
+  local -r operator_ns="$1"
+  echo $(kubectl -n "${operator_ns}" get csv  -l operators.coreos.com/rhacs-operator.operators -o json | jq -r '.items[].spec.version' | sort | tail -n 1)
+}
+
+function nurse_olm_upgrade() {
+  local -r operator_ns="$1"
+  local -r initial_version_tag="$2"
+  local -r desired_version_tag="$3"
+
+  echo current_version_tag
+  local current_version_tag="$(csv_version_tag ${operator_ns})"
+  echo ${current_version_tag}
+
+  # Wait until a new CSV has been created
+  while [[ "${current_version_tag}" == "${initial_version_tag}" ]]; do
+      sleep 1
+      current_version_tag="$(csv_version_tag ${operator_ns})"
+  done
+
+  # Nurse the deployment for each intermediate CSV
+  while [[ "${current_version_tag}" != "${desired_version_tag}" ]]; do
+      echo Current: ${current_version_tag}
+      echo Desired: ${desired_version_tag}
+      nurse_deployment_until_available ${operator_ns} ${current_version_tag}
+      while [[ "$(csv_version_tag ${operator_ns})" == "${current_version_tag}" ]]; do
+          # Wait until the new CSV is created
+          sleep 10
+      done
+      current_version_tag="$(csv_version_tag ${operator_ns})"
+  done
+
+  # Nurse the deployment for the final CSV
+  nurse_deployment_until_available ${operator_ns} ${desired_version_tag}
+}
+
 function nurse_deployment_until_available() {
   local -r operator_ns="$1"
   local -r version_tag="$2"
@@ -122,6 +158,9 @@ function nurse_deployment_until_available() {
   retry 30 10 kubectl -n "${operator_ns}" patch clusterserviceversions.operators.coreos.com \
     "rhacs-operator.v${version_tag}" --type json \
     -p '[ { "op": "add", "path": "/spec/install/spec/deployments/0/spec/template/spec/imagePullSecrets", "value": [{"name": "'"${pull_secret}"'"}] } ]'
+
+  retry 30 10 kubectl -n "${operator_ns}" patch deploy rhacs-operator-controller-manager --type json \
+    -p '[ { "op": "add", "path": "/spec/template/spec/imagePullSecrets", "value": [{"name": "'"${pull_secret}"'"}] } ]'
 
   # Just waiting turns out to be the quickest and most reliable way of propagating the change.
   # Deleting the deployment sometimes tends to never get reconciled, with evidence of the
--- a/operator/hack/common.sh
+++ b/operator/hack/common.sh
@@ -114,42 +114,6 @@ function approve_install_plan() {
   retry 3 5 kubectl -n "${operator_ns}" patch installplan "${install_plan_name}" --type merge -p '{"spec":{"approved":true}}'
 }
 
-function csv_version_tag() {
-  local -r operator_ns="$1"
-  echo $(kubectl -n "${operator_ns}" get csv  -l operators.coreos.com/rhacs-operator.operators -o json | jq -r '.items[].spec.version' | sort | tail -n 1)
-}
-
-function nurse_olm_upgrade() {
-  local -r operator_ns="$1"
-  local -r initial_version_tag="$2"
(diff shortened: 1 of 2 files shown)


operator/hack/common.sh
```diff
@@ -123,6 +159,9 @@ function nurse_deployment_until_available() {
     "rhacs-operator.v${version_tag}" --type json \
     -p '[ { "op": "add", "path": "/spec/install/spec/deployments/0/spec/template/spec/imagePullSecrets", "value": [{"name": "'"${pull_secret}"'"}] } ]'
 
+  retry 30 10 kubectl -n "${operator_ns}" patch deploy rhacs-operator-controller-manager --type json \
```
porridge, line 162:
```suggestion
  log "Patching image pull secret also into rhacs-operator-controller-manager to speed things up."
  retry 30 10 kubectl -n "${operator_ns}" patch deploy rhacs-operator-controller-manager --type json \
```
Did I get the intention right?

operator/hack/common.sh
```diff
@@ -114,6 +114,42 @@ function approve_install_plan() {
   retry 3 5 kubectl -n "${operator_ns}" patch installplan "${install_plan_name}" --type merge -p '{"spec":{"approved":true}}'
 }
 
+function csv_version_tag() {
+  local -r operator_ns="$1"
+  echo $(kubectl -n "${operator_ns}" get csv  -l operators.coreos.com/rhacs-operator.operators -o json | jq -r '.items[].spec.version' | sort | tail -n 1)
```
porridge, line 119:
```suggestion
  echo "$(kubectl -n "${operator_ns}" get csv -l operators.coreos.com/rhacs-operator.operators -o json | jq -r '.items[].spec.version' | sort | tail -n 1)"
```
I don't remember off the top of my head what format the `version` is in, but I would just like to point out that using a plain `sort` to sort (partially-)numeric values can go badly wrong. (Think how `sort` orders `1`, `11`, `9` vs. how `sort -n` does it).

operator/hack/olm-rollback-test.sh
```diff
@@ -0,0 +1,176 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: admin-pass
+data:
+  # letmein
+  password: bGV0bWVpbg==
+EOF
+
+function test() {
```
porridge, line 105:
```suggestion
function test_central_functionality() {
```

operator/hack/olm-rollback-test.sh
```diff
@@ -0,0 +1,176 @@
...
+
+  if [ "${test_result}" != "API Reference" ]; then
+      log Failed to query API
+      exit 1
+  fi
+}
+
+test
+
+new_version=3.74.0-13-gc76ca946d4 # perhaps `make tag` would make this dyanmic?
```
porridge, line 122: Yes, though notice the format differs in a slight but critical way depending on whether you run it in root or `operator/` :wink:

operator/hack/olm-rollback-test.sh
```diff
@@ -0,0 +1,176 @@
...
+nurse_deployment_until_available "${operator_ns}" "${old_version}"
+
+"${KUTTL}" assert --timeout 300 --namespace ${central_ns} /dev/stdin <<-END
+apiVersion: platform.stackrox.io/v1alpha1
+kind: Central
+metadata:
+  name: stackrox-central-services
+status:
+  productVersion:  ${old_version}
+END
```
porridge, line 176: We probably also want to call the test function here as well, no? Ah, that's probably what you meant by:
> Rollback test is currently incomplete.

porridge (commented):
To answer your questions with a question: what is the end goal here? Do we intend to run this kind of test in CI? I guess this might be difficult as is, since I don't think it is possible to downgrade central without restoring its old database first?

operator/hack/common.sh
kylape: yes, exactly.

operator/hack/common.sh
kylape: yeah, admittedly I used `sort` as a quick way to get the "latest" version based only on the two versions I had been testing.  I'll see if I can make this a bit less naive.

kylape:
> what is the end goal here?

kylape:
> what is the end goal here? 

The short term goals are what I stated at the top of the PR description, but I am unsure about the long term goal of these particular tests.  Perhaps they can somehow be integrated in to the CI pipeline (either as some shell script or refactored in to e.g. kuttl), or maybe it's okay to throw this away after the Postgres migration.  I think your thoughts/opinions here would be really valuable.

> Do we intend to run this kind of test in CI? 

On the surface, it seems valuable to execute OLM upgrades (and potentially rollbacks) in CI to test unknown unknowns, but I would imagine these kinds of tests would be expensive to run (i.e. take a long time to finish) and maintain (probably very flaky).

> I don't think it is possible to downgrade central without restoring its old database first?

One of the (short term) goals of this work is to answer this question.  I intend to write down a rollback procedure, however convoluted it may be.

porridge:
> > what is the end goal here?
> 
> The short term goals are what I stated at the top of the PR description, but I am unsure about the long term goal of these particular tests. Perhaps they can somehow be integrated in to the CI pipeline (either as some shell script or refactored in to e.g. kuttl), or maybe it's okay to throw this away after the Postgres migration. I think your thoughts/opinions here would be really valuable.

I have very limited knowledge about the rollback capabilities of central in the rocksdb-only world, on the boundary, and in the postgres-only world. My naive understanding is that:
- in the postgres-only world there are two-way DB migrations so having an automated roll-forward-and-back test would be relatively straightforward. 
- in the two former cases, a rollback basically means "restore from (effectively) a snapshot", and additionally in the rocksdb-only case central needs to be supplied a special flag to do that

If the above is correct, then I think this code has two purposes:
1. to document our understanding of the best way to perform a rollback to rocksdb. This is somewhat temporary and does not have to be pretty nor generic (hardcoded versions are OK), although it would be good to at least be somewhat readable (i.e. a single function should approximately fit on a screen rather than be a spaghetti monstrosity) and moderately robust (`set -e` would be a good idea). It would be great to have it integrated into CI, but just as long as 3.x releases are supported.
2. to validate the possibility to roll forward and back in the postgres-only world. I suspect the code to achieve just this, and not worry about the rocksdb cases might be much simpler. I'd keep this going forward, and so should be kept to the same standards as the rest of the codebase.

> > Do we intend to run this kind of test in CI?
> 
> On the surface, it seems valuable to execute OLM upgrades (and potentially rollbacks) in CI to test unknown unknowns, but I would imagine these kinds of tests would be expensive to run (i.e. take a long time to finish) and maintain (probably very flaky).

You're aware of the [operator upgrade test we have already](https://github.com/stackrox/stackrox/blob/da708eb2d53a7a0c17ef8781169efb760afda49e/operator/Makefile#L402), right?

> > I don't think it is possible to downgrade central without restoring its old database first?
> 
> One of the (short term) goals of this work is to answer this question. I intend to write down a rollback procedure, however convoluted it may be.

PR #4796, opened 2023-02-09
Fix mock central and upgrade unit test

c-du:
## Description

This PR:
1. Remove the hack around LastRocksDBVersionSeqNum. This number shall not be changed and update related tests.
2. SetCurrent for RocksDB to Postgres migration should be at the end of migrator. Fix MockCentral to reflect that and fix related tests.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


pkg/migrations/migration_version.go
```diff
@@ -65,7 +66,7 @@ func SetCurrent(dbPath string) {
 		newVersion := &MigrationVersion{
 			dbPath:      dbPath,
 			MainVersion: version.GetMainVersion(),
-			SeqNum:      LastRocksDBVersionSeqNum(), // Most recent possible RocksDB version
+			SeqNum:      mathutil.MinInt(LastRocksDBVersionSeqNum(), CurrentDBVersionSeqNum()),
```
dashrews78, line 69: The math isn't necessary since you do it in the call to `CurrentDBVersionSeqNum()`.

pkg/migrations/seq_num.go
```diff
@@ -12,7 +13,7 @@ func CurrentDBVersionSeqNum() int {
 	if env.PostgresDatastoreEnabled.BooleanSetting() {
 		return internal.CurrentDBVersionSeqNum
 	}
-	return internal.LastRocksDBVersionSeqNum
+	return mathutil.MinInt(internal.LastRocksDBVersionSeqNum, internal.CurrentDBVersionSeqNum)
```
dashrews78, line 16: If we are going to do it this way, do we still need the `LastRocksDBVersionSeqNum()` below?  I'm a little fuzzy on how this makes it more clear what is happening.

pkg/migrations/migration_version.go
c-du: I thought so but ...
The minimum in CurrentDBVersionSeqNum returns the correct sequence number with Postgres knob on/off. But this one is used in case of Postgres on. If we are running on Postgres and we are making the last mark to RocksDB to its maximum possible value.

pkg/migrations/seq_num.go
c-du: Yes, we will still need it. Before we change, we are still using these codes in mockCentral which may have PostgresDatastoreEnabled off.

dashrews78: Got it, mostly for all the test scenarios.

pkg/migrations/migration_version.go
dashrews78: yeah I'm following.  IMO it is clearer to directly set it to `LastRocksDBVersionSeqNum`, but that is just my opinion.

dashrews78 (approved):

PR #4800, opened 2023-02-09
Update migrator README

c-du:
## Description

Small changes to update README regarding the change of Postgres GA and recommend frozenschema instead.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


migrator/README.md
```diff
@@ -18,17 +19,17 @@ A migration can read from any of the databases, make changes to the data or to t
...
    (from `BoltDB` and `RocksDB` to `Postgres`). The migration sequence is the following: first all key-value
    data migrations are applied, then all data move migrations are applied.
 
     - Key-value store data migrations have the form `m_{currentDBVersion}_to_m_{currentDBVersion+1}_{summary_of_migrations}` .
     - Data move migrations have the form `n_{postgresSchemaVersion}_to_n_{postgresSchemaVersion+1}_{moved_data_type}` .
     
-3. After 3.73, the key-value stores are deprecated and `Postgres` becomes the only data store.  
+3. After 4.0, the key-value stores are deprecated and `Postgres` becomes the only data store.
 
    The migration returns to the old scheme, restarting from the database version after all data moves to `Postgres`.
```
dashrews78, line 34: This has been a source of confusion.  Might be worth clarifying and providing an example like after the last "data move migration" `n_56_to_n_57` we resume use of the normal migration pattern at `m_168_to_m_169`

migrator/README.md
```diff
@@ -132,7 +133,7 @@ it manipulates and for converting it.
 #### Create or upgrade the schema of a table.
 
 ```go
-pkgSchema.ApplySchemaForTable(context.Background(), databases.PostgresDB, <schema>)
+pgutils.CreateTableFromModel(context.Background(), gormDB, frozenSchema.CreateTableCollectionsStmt)
```
dashrews78, line 136: If this is the call we are going to recommend we should make it clear that for the frozen version of the schema we need a copy of the entire schema and not just the gorm struct.  Also I think we should call out that for data manipulation migrations we should not need to freeze the schema nor make this call.

migrator/README.md
```diff
@@ -18,17 +19,17 @@ A migration can read from any of the databases, make changes to the data or to t
 
    All migrations were of the form `m_{currentDBVersion}_to_m_{currentDBVersion+1}_{summary_of_migrations}` .
 
-2. Release 3.73 brought the database `Postgres` as Technical preview. This had the consequence that 
+2. Release 3.73 and 3.74 brought the database `Postgres` as Technical preview. This had the consequence that
 `Postgres` was a new potential datastore targeted by the migrator.
 
-   In 3.73, two sets of data migrations were possible: key-value store migrations, and data move migrations
+   In these versions, two sets of data migrations were possible: key-value store migrations, and data move migrations
```
rhybrillou, line 25: Technically, key-value store migrations are not allowed anymore after 3.73.
With 3.74, there are already postgres -> postgres migrations.
I'm not sure how to formulate things clearly here, though.

migrator/README.md
rhybrillou: +1

migrator/README.md
rhybrillou: There should indeed be a section covering the case of data manipulation migrations. I think these will be common migrations.
I agree there is no need to call `pgutils.CreateTableForModel` in data manipulation migrations, but I think these should rely on a frozen schema and on frozen store code, though.

dashrews78: regarding the frozen schema for data migrations, sure, but we should we make a new copy?.  For instance, if we make a data migration in 4.0, should we make a copy of the schema or just use the most recent frozen one?  Such as using the v73 frozen schema for a 4.0 data only migration instead of copying the same v73 schema into the v4.0 frozen schemas.

rhybrillou: Indeed, the last known frozen state of the schema / store should be sufficient for data manipulation migrations.
I'm also not sure release numbering is the right approach for schema versions. I could well imagine the same table changing multiple times during the same release to accommodate for the needs of different features or projects.
IMHO, the schema version should be incremented each time a datamodel change is submitted. Impacted stores could then be pinned to the frozen changed schema (that could avoid duplication like the `permissionset` store present in `m_168_to_m_169` and in `m_169_to_m_170`).

dashrews78: That is a good point.  I agree, aside from the initial v73 frozen set, future schema changes should be frozen in the specific migration package itself.  That will make it easier to find and follow what is happening.

dashrews78: That is a good point.  I agree, aside from the initial v73 frozen set, future schema changes should be frozen in the specific migration package itself.  That will make it easier to find and follow what is happening.

migrator/README.md
c-du: Em, it is at line 35.
 All migrations again have the form `m_{currentDBVersion}_to_m_{currentDBVersion+1}_{summary_of_migrations}`.

c-du:
By data manipulation, do you mean the migration that does not have schema change?

dashrews78:
> By data manipulation, do you mean the migration that does not have schema change?

Yes.  No changes to the table structure.  Only value updates whether to a column or most likely to the contents of the serialized object.

migrator/README.md
dashrews78: It is, but I don't think it is clear especially with regard to the jump in numbers.  I think we should follow up line 35 with an example pointing out the gap.  I think it would help people understand.

migrator/README.md
c-du: I moved the related notes to an separate session and add more details. Thanks for the feedback.

migrator/README.md
```diff
@@ -130,19 +130,30 @@ it manipulates and for converting it.
...
+from Central and keep the codes of migration stable.
+
+For the initial data push to postgres, the generated schemas under `migrator/migrations/frozenschema/v73` were used.
+The auto-generation scripts are removed to freeze the schemas after 3.73.
+
+In 3.74, snapshots of the schemas are taken with an on-demand basis by version. 
+
+Starting from release 4.0, we recommend to keep frozen schemas inside a new migration:
+- If the migration does not change the schema but need to access the data of a table, it needs to freeze its schema in the migration.
+- If the migration changes the schema of a tale, it may need to freeze two versions of its schema before and after the schema change.
```
dashrews78, line 147: typo: "table"

migrator/README.md
```diff
@@ -130,19 +130,30 @@ it manipulates and for converting it.
...
+and is associated with a specific sequence number (aka version). The frozen schema helps to keep migration separated
+from Central and keep the codes of migration stable.
+
+For the initial data push to postgres, the generated schemas under `migrator/migrations/frozenschema/v73` were used.
+The auto-generation scripts are removed to freeze the schemas after 3.73.
+
+In 3.74, snapshots of the schemas are taken with an on-demand basis by version. 
+
+Starting from release 4.0, we recommend to keep frozen schemas inside a new migration:
+- If the migration does not change the schema but need to access the data of a table, it needs to freeze its schema in the migration.
```
dashrews78, line 146: I struggle with this statement.  I'm not sure I understand the why as we need to freeze the schema if we don't need to apply the schema.  Is this for the case if we are accessing the data via the gorm object?  I think what we mainly need to freeze is any store code we are using to access the data here.  I feel like I'm missing something and need clarification on why we need to freeze the schema in this instance.

c-du: When we access the data, we also use the frozen schema. This is not limited for access using gorm object. If we use the freezed store code, the store codes contains reference to the schema also. It is not likely that the schema is not used at all.

dashrews78: that is a good point.  I didn't think the stores needed the schema.  So I took a look and that brings up another potential issue.  It seemed odd to me why the stores in a migration would be using the schema.  We only need the schema for to pass to `search` and we are using `search` for the walk, delete, and get functions in the stores.  Should we be using current search code to access data in an old schema?  Migrations would probably be safer if we did more with sql

rhybrillou: Is there a policy regarding the content of the `frozenschema` directory hierarchy ?

From the files in `v74`, only `policy_category_edges.go` is new. Looking closer at the schemas and where associated `CreateTable...Stmt` variables are used, I think we should have `v74/policy_category_edges.go` and use it in the `m_170_to_m_171` migration, then `v75/collections.go` and use it in the `m_171_to_m_172` migration.
The other uses of the `v74` schema files should actually rely on the frozen versions of `v73`.
Additionally, the postgres store files should be located in a subdirectory of the appropriate `vXX` schema directory.

The above would allow to have the `m_168_to_m_169` and `m_169_to_m_170` migrations rely on the same frozen schema and store code to remove permissions from permissionsets.

The `m_170_to_m_171` migration would then rely on the `v74` schema and on stores located in various schemas.

I have in mind an upcoming change where I want to enable search on a table, and where the columns for the search would have to be populated from the current content of the proto files (addresses rather the next bullet from the list). I imagine there reading from the `v73` store matching the prior frozen version, and writing back using the `v76` store which created the new columns in the table.

I'll open PRs that puts my ideas in practice and reference them, so that the above ideas become clearer.

rhybrillou: I tried to materialize my ideas above in the following PRs:
- https://github.com/stackrox/stackrox/pull/4839 (centralize store files associated to schemas and have DDL migration influence frozen schema versions)
- https://github.com/stackrox/stackrox/pull/4840 (postgres -> postgres migration with datamodel update, read from old table schema and write to new table schema)

dashrews78: @rhybrillou so your proposal would require a migration to essentially be for a release.  group the store and the schema with the release that changed them.  What about if we make multiple changes within a release that would cause a migration or patch releases?  I was starting to like the idea of keeping these things encapsulated within the specific migration, but I wasn't liking all the potential duplication.  Your approach solves a lot of the duplication issues

c-du: Yes, exactly. @dashrews78
I will add the duplicate store option in the doc.

rhybrillou: Sorry for the late answer. My idea was to have one new datamodel version per datamodel change, not per software release. Hence the idea of having `policy_category_edges` added in datamodel `+1`, then `collections` in datamodel `+2` and new columns in `api_tokens` in datamodel `+3`.
I think decoupling the datamodel and migration versions from the software version also simplify the case of multiple changes within a software release. Assuming a table has to undergo multiple schema change in a software release cycle, each migration that changes the datamodel can read in its `old` datamodel and write to its `new` datamodel. The `new` datamodel of a migration can become the `old` datamodel of a later migrations.

dashrews78: This has me wondering if it may be simpler to have the datamodel you need within the migration itself.  if we version the schema separately we are now keeping track of software version, database sequence, and now schema version.  Though as I type this I can see some value in versioning the schema separately but if we do so we should figure out a way to do it in a way that is obvious.  Perhaps renaming v73frozenschema to v1frozenschema?

c-du: The DB sequence number is actually representing all schemas as a whole. We are trying to invent something like GetSchemaAt(table string, seqNum int) and ideally it could pop up a schema(defined in the current master) that fits into the schema at seqNum. It should not be too hard to implement and we would need a special Registry for all frozen schemas. We can get this offline.

Before that happens, we can freeze schema in each migration to be simpler.

c-du: ROX-15197

dashrews78 (approved):
I think it is in good shape based on our current scheme.  We will just have to make sure any changes to how we migrate later get reflected back into the README.

rhybrillou (approved):

PR #4802, opened 2023-02-09
ROX-14974: Tune waitForEdgeUpdate time

Maddosaurus:
## Description

Tuning for test flake in `Verify cluster updates can block flow connections from showing` (`NetworkFlowTest.groovy`)

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/qa-tests-backend/src/test/groovy/NetworkFlowTest.groovy
+++ b/qa-tests-backend/src/test/groovy/NetworkFlowTest.groovy
@@ -626,8 +626,8 @@ class NetworkFlowTest extends BaseSpecification {
 
         then:
         "make sure edge does not get updated"
-        //Use a 20 second buffer to account for additional edges coming in through the data pipeline
-        assert !waitForEdgeUpdate(edges.get(0), 60, 20)
+        //Use one cadence buffer to account for additional edges coming in through the data pipeline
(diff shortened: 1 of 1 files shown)


Maddosaurus:
[This test run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4802/pull-ci-stackrox-stackrox-master-gke-kernel-qa-e2e-tests/1630858897881305088) failed for the syslog notifier, not for Network Flow

vikin91 (approved):
I see that this changes the assertion wait time from 20 to 30s. The changes are of course perfectly fine, but I am curious why adding 10s is enough here (e.g., why not adding 20 or 30s?)

Maddosaurus:
@vikin91 As far as I have understood the test, `waitForEdgeUpdate` used to have a 20 second buffer, but our NetworkGraphUtil would only update results every `NETWORK_FLOW_UPDATE_CADENCE_IN_SECONDS`.
My guess is that this produced flakes when the update wasn't done in time. 

I increased the time only by 10 seconds  to strike a balance between test execution time and flakyness.
We could err on the safe side and increase it, for example even to 2*update cadence, but I was hesitant to do this without further discussion or feedback.

PR #4805, opened 2023-02-09
Dashrews/ROX-14618 external backup test flakes

dashrews78:
## Description

This doesn't happen all that often considering how frequently the tests runs so it is not something I was able to reproduce.  That said, after looking at the logs, it appears that central is simply not ready as it looks like it is in the middle of a restore from a previous test when this is encountered.  So my theory is it just needs time.  I took 2 approaches.  I added some checks between the tests to make sure central is ready.  Additionally I put the first call to test the backups in a retry.  No need to put the others in a retry as if the first one fails the test fails and likely all the other attempts will fail.  I'm open to other ideas.

## Checklist

## Testing Performed

this is a test

### DIFF
--- a/tests/external_backup_test.go
+++ b/tests/external_backup_test.go
@@ -80,10 +80,25 @@ func TestGCSExternalBackup(t *testing.T) {
 		},
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	_, err = service.TestExternalBackup(ctx, externalBackup)
+	// We could be in a situation where central isn't quite ready from the
+	// previous tests.  This will retry a few times until it is if that is the case.
+	// If this first one doesn't work, then the rest are doomed so no need to wrap those
+	// in retries.
+	err = retry.WithRetry(func() error {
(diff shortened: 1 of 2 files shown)


tests/external_backup_test.go
```diff
@@ -80,12 +80,27 @@ func TestGCSExternalBackup(t *testing.T) {
...
+	// previous tests.  This will retry a few times until it is if that is the case.
+	// If this first one doesn't work, then the rest are doomed so no need to wrap those
+	// in retries.
+	err = retry.WithRetry(func() error {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		_, err := service.TestExternalBackup(ctx, externalBackup)
+		cancel()
+		return err
+	},
+		retry.Tries(10),
```
c-du, line 93: Is 10*1 seconds enough? Otherwise LGTM.

c-du (approved):

tests/external_backup_test.go
dashrews78: Fair point, going for 10*10.

PR #4811, opened 2023-02-10
Clarify that second section is not for every SecuredCluster chart change.

porridge:
## Checklist

## Testing Performed

Doc-only change.

### DIFF
--- a/image/templates/CHANGING_CHARTS.md
+++ b/image/templates/CHANGING_CHARTS.md
@@ -37,9 +37,12 @@ metadata:
     stackrox.io/description: "{{ ._rox.clusterDescription }}"
 ```
 
-### Add a cluster config field to SecuredCluster:
+### Add a cluster config field:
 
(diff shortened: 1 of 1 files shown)


image/templates/CHANGING_CHARTS.md
```diff
@@ -37,9 +37,12 @@ metadata:
     stackrox.io/description: "{{ ._rox.clusterDescription }}"
 ```
 
-### Add a cluster config field to SecuredCluster:
+### Add a cluster config field:
 
-Adding a field to the `SecuredCluster` Helm chart is more complex because the Helm Cluster configuration is tracked in Central and needs adjustments to its conversion logic.
+Making a change that affects the cluster configuration (which is persisted in
```
SimonBaeumer, line 42: +1 This is less confusing. I would a bit more context to indicate that the cluster config belongs to the `secured-cluster-chart`.

```suggestion
Making a change that affects the Secured Cluster chart's cluster configuration (which is persisted in
```

SimonBaeumer (approved):

PR #4813, opened 2023-02-10
Set PSP generation disabled by default in deploy scripts

Maddosaurus:
## Description

As PSPs are deprecated as of k8s 1.25, set `POD_SECURITY_POLICIES` to disabled by default to ensure successful deployments on any Kubernetes flavour, not only local k8s

## Checklist

## Testing Performed

Observed deployment to succeed and ACS as well as Monitoring running as expected without setting an extra env var with `deploy.sh` on:
- Local Colima 1.25
- Local Colima 1.23
- Infra OS 4.12 Cluster
- Infra OS 4.11 Cluster


msugakov:
Regarding testing performed: does deployment still succeed on OCP <=4.11 and/or k8s <= 1.24?

Maddosaurus:
I have also tested this on Colima running Kubernetes 1.23.16 as well as an OS 4.11 cluster.
In both cases, I did not set the env var explicitly, but made sure it is set to false in CLI output.
ACS deployed successfully including Monitoring, the Grafana dashboards populated and ACS was confirmed running as expected.

Maddosaurus:
Ignoring failing `openshift-oldest` tests as they seem to be broken right now.
[Connor is investigating](https://srox.slack.com/archives/CELUQKESC/p1676012050043339).

msugakov (approved):

dashrews78 (approved):

clickboo (approved):

gavin-stackrox:
How does this work in relation to @johannes94 change: https://github.com/stackrox/stackrox/blob/e888172948e602045fc92984611bee83c1f010b4/tests/e2e/lib.sh#L315

Maddosaurus:
@gavin-stackrox These changes are definitely related. 
I am not familiar enough with our test setup to make final statements, but from what I understand, Johannes changes focus on the test setup, correct?
We would still run into problems when a human wants to use our deploy scripts for k8s or OpenShift.

IIUC, Johannes changes are needed to fix the CI tests, so we don't need to cherry pick this PR (4813) to fix them.
Do you agree?

msugakov:
@Maddosaurus
Johannes change tries to auto-sense PSP support, but does that in CI scripting. Your change hardcodes the choice (in a way that should not interfere with CI scripting) for local usage.
The problem is that we don't need that many ways of doing the same thing.

PSPs support can be autosensed in the deployment scripts which should apply to CI automatically, and then no need to hardcode the choice anywhere. How? One option is to simply check exit code from `kubectl get podsecuritypolicy.policy > /dev/null 2>&1` command in `deploy/common/k8sbased.sh`.

Definitely don't check k8s version like `if (( "$majorVersion" >= 1 && "$minorVersion" >= 25 ))` - this won't work when k8s 2.0.0 sees the light.

johannes94:
> Definitely don't check k8s version like `if (( "$majorVersion" >= 1 && "$minorVersion" >= 25 ))` - this won't work when k8s 2.0.0 sees the light.
You're right this is a terrible mistake, we should fix it ASAP. The idea with testing the exit code doesn't sound very reliable to me, because the exit code if an api type is not found is `1` which could have a lot of different reasons.

Just for context, I did not want to set the default for `POD_SECURITY_POLICIES` to `false` because I was not sure if that would disable tests for older k8s version that we want to be executed. Also the place where the `setup_podsecuritypolicies_config` function is used was my best guess without ever looking at our CI scripts before. So if there is a better place to call it like `deploy/common/k8sbased.sh` feel free to move it.

PR #4801, opened 2023-02-09
ROX-14947: Define API to list (ID, name) pairs for cluster and namespace in user scope

rhybrillou:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
~~- [ ] Evaluated and added CHANGELOG entry if required~~
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

If any of these don't apply, please comment below.

## Testing Performed

Manual testing:
- cluster API call with admin user and global scope permission (Role) -> returns empty list
- cluster API call with admin user and cluster scope permission (Node) -> returns list of cluster (ID, name) pairs
- cluster API call with admin user and namespace scope permission (Deployment) -> returns list of (ID, name) pairs
- cluster API call with admin user and no permission -> returns list of (ID, name) pairs (expected behaviour is to return scope data for all cluster and narrower permissions)
- namespace API call with admin user and global scope permission (Role) -> returns empty list
- namespace API call with admin user and cluster scope permission (Node) -> returns empty list
- namespace API call with admin user and namespace scope permission (Deployment) -> returns list of (ID, name) pairs in target cluster
- namespace API call with admin user and no permission -> returns list of (ID, name) pairs in target cluster (expected behaviour is to return scope data for all namespace and narrower permissions)
- create role with only 2 namespace-scoped permissions (Deployment, NetworkGraph), one cluster-scoped permission (Node) and a global scoped permission (Integration), and a scope limited to a single namespace
- create an API token for the above role
- cluster API call with above token and global permission not granted (Role) -> returns empty list
- cluster API call with above token and granted global permission (Integration) -> returns empty list
- cluster API call with above token and cluster permission not granted (Cluster) -> returns empty list
- cluster API call with above token and granted cluster permission (Node) -> returns list of cluster (ID, name) pairs allowed by scope
- cluster API call with above token and namespace permission not granted (Namespace) -> returns empty list
- cluster API call with above token and granted namespace permission (Deployment) -> returns list of cluster (ID, name) pairs allowed by scope
[...]

PR #4816, opened 2023-02-10
ROX-14947: Plug UI network page on new API for clusters and namespaces in scope

rhybrillou:
## Description

Currently, the NetworkGraph pages in the UI requires `Cluster`, `Namespace`, `NetworkGraph` and `Deployment` permissions to display.

The `Cluster` permission is problematic as it could expose cluster configuration information that is not relevant to activities around network graphs.
A new API has been created to expose only limited cluster and namespace information for the clusters and namespaces visible by the user access scopes.

This change aims at integrating this new API in the UI NetworkGraph pages.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


rhybrillou:
* master
  * **PR #4801** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4801" target="_blank">Graphite</a> 
    * **PR #4816** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4816" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4816?utm_source=stack-comment).

ui/apps/platform/src/hooks/useFetchClustersForPermissions.ts
```diff
@@ -0,0 +1,57 @@
...
+    clusters: Cluster[];
+    error: string;
+};
+
+function useFetchClustersForPermissions(permissions: string[]): Result {
+    const defaultResultState = {
+        clusters: [],
+        error: '',
+        isLoading: true,
+    };
```
vjwilson, line 18: Same comment about default of true for loading here, as I had in the file `ui/apps/platform/src/hooks/useFetchClusterNamespacesForPermissions.ts` above.

ui/apps/platform/src/services/RolesService.ts
```diff
@@ -116,3 +117,53 @@ export function updatePermissionSet(entity: PermissionSet): Promise<Empty> {
 export function deletePermissionSet(id: string): Promise<Empty> {
     return axios.delete(`${permissionSetsUrl}/${id}`);
 }
+
+const clustersForPermissionsUrl = '/v1/sac/clusters';
```
vjwilson, line 121: As I currently understand this feature, we are hoping to reuse this endpoint for many other data types that need to be retrieved, even though the user may not have full permissions for them.

Should we make the endpoint something besides "clusters" ? 

Otherwise, there will be a proliferation of sub-endpoints under the `sac/` root.

vjwilson (changes requested):
Overall, good job of abstracting this logic, and good JS idiom.

I had a couple of requests, and one major concern about the API endpoint you have been given to work with.

Outside of this particular PR, but something we should consider: the old Network Graph is scheduled to be removed in `4.0.0`, so these changes will not ever be released for customers to use. While there is a chance the new Network Graph could be delayed, it may be a better use of type to redirect this work to the new `Containers/NetworkGraph/` directory code, which already has some of the optimizations that are in this PR (although not the main point, the SAC functionally).

ui/apps/platform/src/hooks/useFetchClusterNamespacesForPermissions.ts
```diff
@@ -0,0 +1,66 @@
...
+    namespaces: Namespace[];
+};
+
+const emptyResponse: NamespaceResponse = {
+    loading: false,
+    error: '',
+    namespaces: [] as Namespace[],
+};
+
+export function useFetchClusterNamespacesForPermissions(
```
ivan-degtiarenko, line 22: I couldn't find usages of this function

rhybrillou: It is now used in the namespace filter (and in the Network Graph 2.0 page).

ui/apps/platform/src/services/RolesService.ts
rhybrillou: In the current state of things, the `/v1/sac/` endpoint should only have two sub-endpoints: `/v1/sac/clusters` and `/v1/sac/clusters/{cluster_id}/namespaces`.
I think the current state of the endpoint and its sub-endpoint reflects the state of access scopes in the product and should not evolve more at this point. Future evolutions of the access scope model may have to be reflected in the sub-endpoint hierarchy.

vjwilson (approved):

ui/apps/platform/src/hooks/useFetchClusterNamespacesForPermissions.ts
```diff
@@ -0,0 +1,67 @@
...
+        setNamespaceResponse({
+            loading: true,
+            error: '',
+            namespaces: [],
+        });
+        if (selectedClusterId) {
+            getNamespacesForClusterAndPermissions(selectedClusterId, requestedPermissions)
+                .then((data) => {
+                    const responseNamespaces = data.namespaces;
+                    const namespaces: Namespace[] = [];
```
ivan-degtiarenko, line 39: nit: Can be simplified using `map` function:
```
                    const namespaces: Namespace[] = data.namespaces.map(
                        (responseNamespace: ScopeObject) => {
                            return {
                                id: responseNamespace.id,
                                name: responseNamespace.name,
                            } as Namespace;
                        }
                    );
```

ui/apps/platform/src/hooks/useFetchClustersForPermissions.ts
```diff
@@ -0,0 +1,65 @@
...
+    const [requestedPermissions] = useState<string[]>(permissions);
+
+    useEffect(() => {
+        setResult(defaultResultState);
+
+        getClustersForPermissions(requestedPermissions)
+            .then((data) => {
+                const responseClusters = data.clusters;
+                const clusters: Cluster[] = [];
+                responseClusters.forEach((responseCluster: ScopeObject) => {
```
ivan-degtiarenko, line 37: nit: can be simplified using `map` function

ui/apps/platform/src/Containers/Network/Header/useNamespaceFilters.ts
```diff
@@ -13,55 +12,19 @@ const selector = createStructuredSelector<SelectorState, SelectorResult>({
...
-    // If the selectedClusterId has not been set yet, do not run the gql query
-    const queryOptions = selectedClusterId
-        ? { variables: { id: selectedClusterId } }
-        : { skip: true };
 
-    const { loading, error, data } = useQuery<NamespaceMetadataResp, { id: string }>(
-        NAMESPACES_FOR_CLUSTER_QUERY,
-        queryOptions
+    const targetClusterId: string | undefined =
+        selectedClusterId == null ? undefined : selectedClusterId;
```
ivan-degtiarenko, line 22: what is the goal of transforming `null` to `undefined`?

ui/apps/platform/src/Containers/Network/Header/ClusterSelect.tsx
```diff
@@ -44,19 +43,16 @@ const ClusterSelect = ({
             placeholderText="Select a cluster"
             onSelect={changeCluster}
         >
-            {clusters
-                .filter((cluster) => networkGraphClusters[cluster.type])
```
ivan-degtiarenko, line 48: why is this filter no longer needed?

ui/apps/platform/src/Containers/Network/Header/useNamespaceFilters.ts
rhybrillou: The goal here is to match the signature of the `useFetchClusterNamespacesForPermissions` function.
I changed the function signature to accept string, null or undefined and removed that code piece.

ui/apps/platform/src/Containers/Network/Header/ClusterSelect.tsx
rhybrillou: The base (and bad) reason is that the piece of data needed by the filter is not available in the new API.
Looking closer at the code, the filter only excludes the `GENERIC_CLUSTER` type (and possibly SWARM clusters, not supported anymore since Jan 19 or DOCKER_EE clusters not supported anymore since Jun 19).
Manual cluster addition in the UI only allows choice between Kubernetes, OpenShift 3.x compatibility or OpenShift 4.x. All these cluster flavours pass the filter.

ivan-degtiarenko: Based on conversation [here](https://redhat-internal.slack.com/archives/CELUQKESC/p1677590535559309) let's create a follow up JIRA to error on `GENERIC_CLUSTER` when met with this value

ivan-degtiarenko (approved):
Approved modulo creating follow-up ticket

ui/apps/platform/src/Containers/Network/Header/ClusterSelect.tsx
rhybrillou: https://issues.redhat.com/browse/ROX-15783

PR #4820, opened 2023-02-10
Fix inability to deploy with helm due to missing certificates

connorgorman:
## Description

We need to generate the central db and central db certs in roxctl which is used in the k8s based install

## Checklist

## Testing Performed

oldest openshift and ui tests

### DIFF
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -253,6 +253,8 @@ function launch_central {
 
     echo "Deploying Central..."
 
+    $(git rev-parse --show-toplevel)/scripts/ci/store-artifacts.sh store_artifacts "${k8s_dir}/central-deploy"
+
     ${KUBE_COMMAND:-kubectl} get namespace "${STACKROX_NAMESPACE}" &>/dev/null || \
       ${KUBE_COMMAND:-kubectl} create namespace "${STACKROX_NAMESPACE}"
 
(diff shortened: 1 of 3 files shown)


deploy/common/k8sbased.sh
```diff
@@ -253,6 +253,8 @@ function launch_central {
 
     echo "Deploying Central..."
 
+    $(git rev-parse --show-toplevel)/scripts/ci/store-artifacts.sh store_artifacts "${k8s_dir}/central-deploy"
```
gavin-stackrox, line 256: nit:
```suggestion
    "$(git rev-parse --show-toplevel)/scripts/ci/store-artifacts.sh" store_artifacts "${k8s_dir}/central-deploy"
```

gavin-stackrox (approved):

pkg/renderer/helm_values.go
```diff
@@ -403,9 +403,6 @@ var (
 	publicValuesTemplate = template.Must(
 		helmTemplate.InitTemplate("values-public.yaml").Parse(publicValuesTemplateStr))
 
-	privateValuesTemplate = template.Must(
```
dashrews78, line 406: should probably remove `privateValuesYamlPostgresTemplateStr` while you are at it.

dashrews78 (approved):

c-du (approved):

PR #4827, opened 2023-02-10
Add examples of new select query for real life use cases.

md2119:
## Description

I am adding some unit tests in the consumer space of the search framework early on for multiple reasons:
- Validate the use of the search framework in the consumer space
- Track regressions, if any, as I continue to evolve the framework

This PR adds some example unit tests for real-life use cases.

Example 1: Violation count by severity (dashboard widget) w. benchmark test. The memory profile shows the mem alloc would be down to 3mb from 500mb.
Example 2: Deployment count by kind

## Checklist

## Testing Performed
Unit


central/deployment/datastore/datastore_impl_postgres_test.go
```diff
@@ -56,35 +52,19 @@ func (s *DeploymentPostgresDataStoreTestSuite) SetupSuite() {
...
-	s.riskDataStore = riskMocks.NewMockDataStore(s.mockCtrl)
+	imageDS, err := imageDataStore.GetTestPostgresDataStore(s.T(), s.testDB.DB)
+	s.Require().NoError(err)
+	s.imageDatastore = imageDS
 
-	s.deploymentDatastore, err = newDataStore(
-		deploymentPostgres.NewFullTestStore(s.T(), deploymentPostgres.CreateTableAndNewStore(s.ctx, s.db, s.gormDB)),
-		nil, s.db, nil, nil, s.imageDatastore, nil, nil, s.riskDataStore,
-		nil, nil, ranking.ClusterRanker(), ranking.NamespaceRanker(), ranking.DeploymentRanker())
+	deploymentDS, err := GetTestPostgresDataStore(s.T(), s.testDB.DB)
```
charmik-redhat, line 61: This would create a new imageDataStore, different from the one created above by `imageDataStore.GetTestPostgresDataStore(s.T(), s.testDB.DB)` . Does that have any consequences to the test?

md2119: It creates new instance of the datastore but both instances of the datastore point to the same Postgres table. This is because they share the same DB config.

The unit test focuses on database search not the internal state of the datastore which the image datastore does not have anyways.

charmik-redhat: Makes sense, thanks for clarification.

charmik-redhat (approved):
Looks good

PR #4828, opened 2023-02-10
ROX-14147: Add reconciliation scaffolding for declarative configurations

dhaus67:
## Description

This PR adds the scaffolding for declarative configurations.

The scaffolding focuses around the parts of ensuring the watch handlers propagate changes in the mount points of the declarative configuration files, so that can later be reused to implement the reconciliation (i.e. upserting resources, deleting resources).

The following has been done specifically:
- Instead of having a single watch handler for the declarative configuration directory, there will be multiple watch handlers. This is due to the nature of the watch handler implementation, which skips updates received within subdirectories.
- Instead of creating the single watch handler, the declarative configuration manager will list all directories within the declarative configuration directory, and subsequently create and start the watch handlers for those.
- Each specific watch handler will:
  - List and unmarshal all files within the `OnUpdate` hook.
  - Cache file contents and only trigger reconciliation within the declarative configuration manager if changes are present (i.e. new files, changed hashes, removed files) in the `OnStableUpdate` hook. While currently the diff is not yet upserted (i.e. only the changed files are sent), it was still worthwhile 
to add it from the beginning.

This will be the pre-req for subsequently adding upserting and deletion in the reconciliation method within the declarative configuration manager.

As additional follow-ups we might think about:
- splitting the work of unmarshalling the file contents (`[][]byte`) to declarative configurations.
- as an optional item, only send the diff of files to the reconciliation method within the declarative configuration manager.

## Checklist

## Testing Performed

- see unit tests that have been added.


dhaus67:
* master
  * **PR #4828** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4828" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4828?utm_source=stack-comment).

central/declarativeconfig/watch_handler.go
```diff
@@ -1,23 +1,177 @@
...
+	return true
+}
+
+func (w *watchHandler) logFileContentsNoLock(contents map[string][]byte) {
+	logMessage := "Found declarative configuration file contents\n"
+
+	w.mutex.RLock()
+	defer w.mutex.RUnlock()
+	for fileName, fileContents := range contents {
+		logMessage += fmt.Sprintf("File %s: %s", fileName, fileContents)
```
ivan-degtiarenko, line 156:
```suggestion
		logMessage += fmt.Sprintf("File %s: %s\n", fileName, fileContents)
```

central/declarativeconfig/watch_handler.go
```diff
@@ -1,23 +1,177 @@
...
+	}
+
+	w.cachedFileHashes = newCachedFileHashes
+	return true
+}
+
+func (w *watchHandler) logFileContentsNoLock(contents map[string][]byte) {
+	logMessage := "Found declarative configuration file contents\n"
+
+	w.mutex.RLock()
```
ivan-degtiarenko, line 153: Function is called `..NoLock` but it seems to me there are locks here :)

central/declarativeconfig/watch_handler.go
```diff
@@ -1,23 +1,177 @@
...
+// Otherwise, it will return false.
+func (w *watchHandler) compareHashesForChanges(fileContents map[string][]byte) bool {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	var changedFiles bool
+	for fileName, fileContent := range fileContents {
+		cachedHash, ok := w.cachedFileHashes[fileName]
+		fileHash := md5.Sum(fileContent)
+		if !ok || !bytes.Equal(cachedHash[:], fileHash[:]) {
+			w.cachedFileHashes[fileName] = fileHash
```
ivan-degtiarenko, line 113: My understanding is that this code works under the assumption that the file is always reconciled error-free after this line. Currently, that's the case(there are no errors returned all until line 90:
https://github.com/stackrox/stackrox/pull/4828/files#diff-e828eface4a18e3746716e6b6ff4135d5773e64d644cdd10eff0ca28664e1394R90

Perhaps, it would be easier not to miss any changes if we update the hash value after it actual reconciliation?

central/declarativeconfig/watch_handler.go
```diff
@@ -1,23 +1,177 @@
...
+// the list of updated files. Otherwise, returns false.
+// In the end, the cached values will be changed to reflect the passed file contents.
+func (w *watchHandler) checkForDeletedFiles(fileContents map[string][]byte) bool {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	cachedFileNames := set.NewStringSet(maputil.Keys(w.cachedFileHashes)...)
+	fileNames := set.NewStringSet(maputil.Keys(fileContents)...)
+
+	// No deleted files if both arrays are equal.
+	if cachedFileNames.Equal(fileNames) {
```
ivan-degtiarenko, line 130: I think this condition can be removed since it will be covered by 
```
if removedFiles.Cardinality() == 0 {
		return false
	}
```

central/declarativeconfig/manager_impl.go
```diff
@@ -30,11 +39,34 @@ func New() Manager {
...
+		// For each directory within the declarative configuration path, create a watch handler.
+		// The reason we need multiple watch handlers and cannot simply watch the root directory is that
+		// changes to directories are ignored within the watch handler.
+		entries, err := os.ReadDir(declarativeConfigDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				log.Info("Declarative configuration directory does not exist, no reconciliation will be done")
+				return
+			}
+			utils.Should(err)
```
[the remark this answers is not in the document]
dhaus67: In development builds, we'll panic, in release builds we'll log it as `Unexpected Error` and not start declarative configuration reconciliation.

central/declarativeconfig/watch_handler.go
dhaus67: So, essentially you would want to have an error returned from `ReconcileDeclarativeConfigs` and only _then_ update the cached file hashes?

Not sure I'd follow the same, since these watch handlers are for me just purely for reading the file content as YAML bytes and passing them to the declarative configuration manager for reconciliation.

What exactly would make the difference in doing it for a _successfuly_ reconciliation, which could take longer and might be asynchronous (right now, we don't actually do anything in `ReconcileDeclarativeConfigs`, but I doubt its going to be synchronous going forward, as this essentially would add the transformed protos to some sort of loop for reconciliation).

central/declarativeconfig/watch_handler.go
```diff
@@ -1,23 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer utils.IgnoreError(f.Close)
+	fileContents, err := io.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := yaml.Unmarshal(fileContents, &map[string]interface{}{}); err != nil {
```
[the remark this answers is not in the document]
dhaus67: I would have to create a separate variable for it (doing inline `interface{}{}` doesn't work, and the actual unmarshalled type is going to be `map[string]interface{}`, so I'd just keep it inline for now, unless you _really_ want it changed.

central/declarativeconfig/watch_handler.go
ivan-degtiarenko: Let's keep it as it is. It's important though to add a comment that there should be no error-prone code between these lines
https://github.com/stackrox/stackrox/pull/4828/files#diff-e828eface4a18e3746716e6b6ff4135d5773e64d644cdd10eff0ca28664e1394R80-R88 -- otherwise, it's possible changes to files will not be picked up.

central/declarativeconfig/watch_handler_test.go
```diff
@@ -0,0 +1,301 @@
...
+	filePath := path.Join(dirToWatch, "role")
+	f, err := os.Create(filePath)
+	require.NoError(t, err)
+	defer utils.IgnoreError(f.Close)
+
+	_, err = f.Write(yamlBytes)
+	require.NoError(t, err)
+
+	// 3. Wait to ensure the watch handler has been triggered.
+	time.Sleep(100 * time.Second)
```
ivan-degtiarenko, line 153: I'm not sure we want to wait 100 seconds or use `Sleep` at all. Perhaps we can wait on a `cachedFileHashes` change with some timeout?

central/declarativeconfig/watch_handler_test.go
```diff
@@ -0,0 +1,301 @@
...
+	// 4. Assert on the cached file hashes.
+	expectedCache := map[string]md5CheckSum{
+		"role": md5.Sum(yamlBytes),
+	}
+
+	wh.mutex.RLock()
+	assert.Equal(t, expectedCache, wh.cachedFileHashes)
+	wh.mutex.RUnlock()
+
+	// 5.Set the expected calls to the updater
```
ivan-degtiarenko, line 287:
```suggestion
	// 5.Set the expected calls to the updater.
```

central/declarativeconfig/watch_handler_test.go
dhaus67: The sleep should be `100 *time.Milisecond`, and I'd rather use it compared to watching for changes. For watching changes, we'd need to acquire a read lock, which could interfere with the watch handler routine and make the test flaky.

ivan-degtiarenko: Hmm, how would acquiring a read lock interfere with the watch handler routine? I thought we put these locks in the first place because we allow for parallel updates.

ivan-degtiarenko (approved):
LGTM modulo discussion on sleeps in tests - however, I'm OK with continuing this discussion after the merge.

central/declarativeconfig/watch_handler_test.go
ivan-degtiarenko: Anyway, in case there are concerns with using `wh.mutex.RLock()` you can use `wh.mutex.TryRLock()`

dhaus67: Yeah, I guess what I initially had in mind we'd have to do within tests would be to create a ticker where we would have to continuously check for changes - which would've been a bit too much for me to implement over simply using time.Sleep.
But, I suppose I forgot about `assert.Eventually` - so there isn't really an argument for the `time.Sleep` - I removed those.

PR #4832, opened 2023-02-13
ROX-14147: Create unmarshalling of configuration from bytes

dhaus67:
## Description

This PR adds unmarshalling of declarative configurations from file contents, i.e. raw bytes. This is required to be used within the declarative configuration manager, which will unmarshal the file contents received from the watch handler watching the declarative configuration mounts to declarative 
configurations. Afterwards, the declarative configurations will be transformed to proto messages and can be used within the reconciliation for upserting / deleting etc.

Within the PR is a caveat of the way YAML unmarshal works:
Since we reuse the `name` and `description` field across all declarative configuration structs (except for auth provider, which only shares name), `yaml.Unmarshal` will not return an error if one tries to unmarshal e.g. a `Role` struct in YAML format into a `Permission Set` struct.
In this case, all other fields related to the specific configuration will be unset.

As a workaround, for the time being I've decided to create small helper functions which verify that the declarative configuration is empty, and if it is, the struct will not be unmarshalled and returned.

As a follow-up, [ROX-15067](https://issues.redhat.com/browse/ROX-15067) has been created to cover the validation on the struct level (at the very least, the static part of it to not increase the duplication of the validation within the data stores).
Additionally, a follow-up of this PR may come as well to create a specific `Unmarshaller` interface implementation for declarative configurations, albeit this one might be more effort than we actually want to spend on.

## Checklist

## Testing Performed

- added unit tests.


dhaus67:
* master
  * **PR #4828** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4828" target="_blank">Graphite</a> 
    * **PR #4832** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4832" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4832?utm_source=stack-comment).

central/declarativeconfig/manager_impl.go
```diff
@@ -68,5 +71,21 @@ func (m *managerImpl) WatchDeclarativeConfigDir() {
...
 // TODO(ROX-14694): Add deletion of resources.
 func (m *managerImpl) ReconcileDeclarativeConfigs(contents [][]byte) {
+	configurations, err := declarativeconfig.ConfigurationFromRawBytes(contents...)
+	if err != nil {
+		log.Errorf("Error during unmarshalling of declarative configuration files: %+v", err)
+		return
+	}
+	transformedConfigurations := make(map[reflect.Type][]proto.Message, len(configurations))
+	for _, configuration := range configurations {
+		transformedConfig, err := m.t.Transform(configuration)
```
ivan-degtiarenko, line 81: nit: perhaps we can rename `t` to `transform` or `transformer`?`m.t` seems kind of confusing

pkg/declarativeconfig/configuration.go
```diff
@@ -15,3 +22,92 @@ const (
...
+			return nil, err
+		}
+		configurations = append(configurations, config)
+	}
+	return configurations, nil
+}
+
+func fromRawBytes(rawConfiguration []byte) (Configuration, error) {
+
+	for _, configurationType := range []ConfigurationType{AuthProviderConfiguration, AccessScopeConfiguration,
```
ivan-degtiarenko, line 73: Let's remove `for` and `switch`

pkg/declarativeconfig/configuration.go
```diff
@@ -15,3 +22,92 @@ const (
...
+			return &permissionSet, nil
+		case RoleConfiguration:
+			var role Role
+			if err := decodeYAMLToConfiguration(rawConfiguration, &role); err != nil {
+				break
+			}
+			return &role, nil
+		}
+	}
+	return nil, errors.Errorf("raw configuration found that didn't match any of the given configurations: %s",
```
ivan-degtiarenko, line 102: I think it would be beneficial to use  `multierror` here and combine all errors to show which values are missing

pkg/declarativeconfig/configuration.go
```diff
@@ -15,3 +22,92 @@ const (
...
+//   - the YAML cannot be unmarshalled into valid configuration type.
+func ConfigurationFromRawBytes(rawConfigurations ...[]byte) ([]Configuration, error) {
+	var configurations []Configuration
+	for _, rawConfiguration := range rawConfigurations {
+		var objects []interface{}
+		err := yaml.Unmarshal(rawConfiguration, &objects)
+		if err == nil {
+			configs, err := fromUnstructuredConfigs(objects)
+			if err != nil {
+				return nil, errors.Wrap(err, "unmarshalling list of raw configuration")
```
ivan-degtiarenko, line 39:
```suggestion
				return nil, errors.Wrap(err, "unmarshalling list of raw configurations")
```

pkg/declarativeconfig/configuration_test.go
```diff
@@ -0,0 +1,138 @@
+package declarativeconfig
+
+import (
+	"testing"
+
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConfigurationFromRawBytes(t *testing.T) {
```
ivan-degtiarenko, line 10: Let's also make sure we have test that every configuration type gets decoded into correct proto type

pkg/declarativeconfig/configuration.go
```diff
@@ -15,3 +23,91 @@ const (
...
+
+	return configurations, nil
+}
+
+func fromUnstructuredConfigs(unstructuredConfigs []interface{}) ([]Configuration, error) {
+	configurations := make([]Configuration, 0, len(unstructuredConfigs))
+	for _, unstructuredConfig := range unstructuredConfigs {
+		rawConfigurationBytes, err := yaml.Marshal(unstructuredConfig)
+		if err != nil {
+			return nil, errors.Wrapf(err, "marshalling configuration from list %+v", unstructuredConfig)
```
ivan-degtiarenko, line 62: super-nit: perhaps mentioning index would be useful as well

ivan-degtiarenko (approved):
🚀

PR #4838, opened 2023-02-13
CI: run tests against a PR in stackrox/collector

ovalenti:
## Description

Temporary PR to run tests against a PR in stackrox/collector


COLLECTOR_VERSION
```diff
@@ -1 +1 @@
-3.13.0
+3.13.x-40-g93913c7b59-latest
```
robbycochran, line 1: You shouldn't not add the `-latest` suffix, it will be added by central if slimMode is not used in central

PR #4839, opened 2023-02-13
Restructure migrations to better version postgres schemas

rhybrillou:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


c-du:
Migration 169 to 172 are introduced in 74. Why moving related schema and store to v75?

rhybrillou:
> Migration 169 to 172 are introduced in 74. Why moving related schema and store to v75?

My idea here is to decorrelate the datamodel versions from the product release versions. The starting point is v73, then each datamodel change (column addition, removal, new table) is a new datamodel version.
It might be a good idea here to also keep track of the last datamodel version in which each schema was last modified, as it would help figure out where to search for the store in case of data manipulation only migrations.

rhybrillou:
In the current state of things, the philosophy is that each migration (data manipulation or data model definition) should embed the files it needs.

PR #4841, opened 2023-02-13
Update Central DB signal handling to default to fast shutdown

connorgorman:
## Description

Default to fast shutdown, which kills all in flight txns and rolls them back. This will allow for faster shutdown and less likely sigkills in the middle of termination

## Checklist

## Testing Performed

Ran the equivalent in a chaos test for several days

### DIFF
--- a/image/postgres/scripts/docker-entrypoint.sh
+++ b/image/postgres/scripts/docker-entrypoint.sh
@@ -346,7 +346,7 @@ _main() {
 
 	"$@" &
 	child=$!
-	echo "Waiting for child process to exit"
+	echo "Waiting for child process $child to exit"
 	wait "$child"
 }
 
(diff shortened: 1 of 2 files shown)


image/postgres/scripts/docker-entrypoint.sh
```diff
@@ -338,7 +344,10 @@ _main() {
 		fi
 	fi
 
-	exec "$@"
+	"$@" &
+	child=$!
+	echo "Waiting for child process to exit"
```
c-du, line 349: Put the process id in this message could be helpful

c-du (approved):

c-du:
With one minor comment

dashrews78 (approved):

PR #4852, opened 2023-02-13
ROX-15103 Correctly handle different alert types on dashboard widget

dvail:
## Description

Correctly handles the response type for the `alerts` graphql queries in the dashboard widget.

The query was incorrectly types as a `DeploymentAlert` when the true response type is `Alert`, which could be a `DeploymentAlert`, `ImageAlert`, or `ResourceAlert`, each of which has different `null` fields. This caused the dashboard to crash when the most recent alert was from a `ResourceAlert` type, which I believe only occurs when the user creates a custom policy for k8s audit logs and sets the severity to critical.

This may be easier to review with "Hide whitespace" on.

## Checklist

## Testing Performed

Load the widget with violations of a custom policy tracking the `configmap` type.
image

Load the widget with violations of a custom policy tracking the `secrets` type.
image

Load the widget with violations that don't match any of our expectations. This should never happen, but like most things, sometimes they do.
image

### DIFF
--- a/ui/apps/platform/src/Components/PatternFly/ResourceIcon/ResourceIcon.css
+++ b/ui/apps/platform/src/Components/PatternFly/ResourceIcon/ResourceIcon.css
@@ -21,3 +21,16 @@
 .resource-icon-deployment {
   background-color: var(--color-resource-deployment);
 }
+
+.resource-icon-configmap {
+  background-color: var(--color-resource-configmap);
+}
+
+.resource-icon-secret {
+  background-color: var(--color-resource-secret);
+}
+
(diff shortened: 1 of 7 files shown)


dvail:
* master
  * **PR #4852** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4852" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4852?utm_source=stack-comment).

pedrottimark (approved):

PR #4853, opened 2023-02-13
ROX-14973: Support CronJobs in a backwards compatible way

connorgorman:
## Description

This uses k8s discovery API to determine if v1/batch CronJob exists. If it does then use the informer for that, if not, then fall back to v1beta1. This does however mean that without more involved changes, that we cannot move forward from k8s client-go 1.24 which is the last version with v1beta1 for cron jobs (there isn't risk of this happening automatically as an API upgrade will cause a compilation error)

## Checklist

## Testing Performed

CI should test the v1 flow, but we'll have to manually test 3.11 and <4.6

### DIFF
--- a/sensor/admission-control/manager/k8s_object.go
+++ b/sensor/admission-control/manager/k8s_object.go
@@ -34,7 +34,11 @@ func unmarshalK8sObject(gvk metav1.GroupVersionKind, raw []byte) (k8sutil.Object
 	case kubernetes.ReplicaSet:
 		obj = &apps.ReplicaSet{}
 	case kubernetes.CronJob:
-		obj = &batchV1beta1.CronJob{}
+		if gvk.Version == "v1beta1" {
+			obj = &batchV1beta1.CronJob{}
+		} else {
+			obj = &batchV1.CronJob{}
+		}
 	case kubernetes.Job:
 		obj = &batchV1.Job{}
 	case kubernetes.DeploymentConfig:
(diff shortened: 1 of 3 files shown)


lvalerom (approved):
LGTM

msugakov:
There's this spot and I was trying to understand if we should make a change.
https://github.com/stackrox/stackrox/blob/46238c9ee8dc4345bf8fc0c13a8104542f525fa0/sensor/admission-control/manager/k8s_object.go#L36-L37

Maddosaurus:
> This does however mean that without more involved changes, that we cannot move forward from k8s client-go 1.24 which is the last version with v1beta1 for cron jobs

Given that we want to upgrade the client some time in the future, what changes would be expected?
I guess we cannot just include different versions of the client libs depending on the discovered k8s version, can we?

IIUC, the upgrade path would be blocked for as long as 1.24 is supported by ACS (which would be quite some time) correct?

msugakov:
After visually comparing
https://github.com/kubernetes/api/blob/master/batch/v1/generated.proto
and
https://github.com/kubernetes/api/blob/master/batch/v1beta1/generated.proto

I believe v1 CronJob is field-wise equivalent to v1beta1 CronJob. I wonder if we could actually modify version and deserialize the v1beta1 as v1.

connorgorman:
@msugakov Yeah they are are equivalent (most of the time v1 vs the last beta is equivalent because it's just a promotion process)

connorgorman:
@Maddosaurus It'll be more involved, but we either vendor multiple client versions or do some magic with deserialization with the dynamic API. All of which are too involved for what we're trying to do right now potentially with respect to 74. I do think that we may need to consider either tightening the support matrix OR creating a generic plan to support these objects going forward

msugakov:
Runtimes and network graph work on OCP 4.7.

msugakov (approved):

msugakov:
Looks like this needs a rebase from `master`. https://srox.slack.com/archives/CELUQKESC/p1676392699199519

PR #4859, opened 2023-02-14
Fix service account test by using a specific query instead of a list

connorgorman:
## Description

On Postgres on Openshift, listing all service accounts can be very slow. We have a different ticket for this: ROX-14666

## Checklist

## Testing Performed

CI

### DIFF
--- a/qa-tests-backend/src/main/groovy/services/ServiceAccountService.groovy
+++ b/qa-tests-backend/src/main/groovy/services/ServiceAccountService.groovy
@@ -28,13 +28,16 @@ class ServiceAccountService extends BaseService {
         }
     }
 
+    static RawQuery getServiceAccountQuery(K8sServiceAccount serviceAccount) {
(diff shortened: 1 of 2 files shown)


theencee (approved):

janisz (approved):

qa-tests-backend/src/main/groovy/services/ServiceAccountService.groovy
```diff
@@ -28,17 +28,17 @@ class ServiceAccountService extends BaseService {
         }
     }
 
+    static RawQuery getServiceAccountQuery(K8sServiceAccount serviceAccount) {
+        def query = "Namespace:\"${serviceAccount.namespace}\"+Service Account:\"${serviceAccount.name}\""
```
clickboo, line 32: Nit: Flip ordering of query params

clickboo (approved):

qa-tests-backend/src/main/groovy/services/ServiceAccountService.groovy
connorgorman: Why

PR #4860, opened 2023-02-14
Wrap pgxpool to allow overriding functions and easier package updates

connorgorman:
## Description

Do not use the pgxpool.Pool directly and instead wrap it so that we can override functions and add metrics, etc

## Checklist

## Testing Performed

CI, should be a no-op


pkg/postgres/pool.go
```diff
@@ -0,0 +1,90 @@
+package postgres
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
```
dashrews78, line 6: If nothing else this might be worth it just for making version changes simpler.

dashrews78 (approved):
Looks good

PR #4870, opened 2023-02-14
ROX-14019: remove policies table from clusters in vuln mgmt

bradr5:
## Description

Remove the policies table from clusters in vulnerability management.

The table on its own doesn't provide as much context as the violations page. The same information can be found on the violations page, which the field team confirmed can be confusing to customers.

## Checklist

## Testing Performed

Screenshot 2023-02-14 at 12 35 33 PM


vjwilson (approved):

md2119 (approved):
Seems like a configuration management test is failing

PR #4871, opened 2023-02-14
ROX-15009 Add navigation item for 'Vulnerabilities' (VM 2.0) section

dvail:
## Description

Adds left hand navigation item for new Vulnerabilities section as well as routes to match the Workload CVE pages:
- Workload CVE Overview
- CVE Single page
- Image Single page
- Deployment Single page

## Checklist

## Testing Performed

The navigation with the feature flag enabled:
image

The navigation with the feature flag disabled:
image

Navigating to any of the Vulnerabilities subpages returns a 404 page, with or without the feature flag. (Route handlers are not implemented in this PR.)
image

### DIFF
--- a/ui/apps/platform/src/Containers/MainPage/Sidebar/NavigationSidebar.tsx
+++ b/ui/apps/platform/src/Containers/MainPage/Sidebar/NavigationSidebar.tsx
@@ -32,6 +32,7 @@ import {
     systemConfigPath,
     systemHealthPath,
     collectionsBasePath,
+    vulnerabilitiesWorkloadCvesPath,
 } from 'routePaths';
 
 import LeftNavItem from './LeftNavItem';
@@ -75,6 +76,8 @@ function NavigationSidebar({
             collectionsBasePath
         );
     }
+
+    const vulnerabilitiesPaths = [vulnerabilitiesWorkloadCvesPath];
 
     const Navigation = (
         <Nav id="nav-primary-simple">
@@ -122,6 +125,33 @@ function NavigationSidebar({
                     path={complianceBasePath}
                     title={basePathToLabelMap[complianceBasePath]}
                 />
+
+                {isFeatureFlagEnabled('ROX_VULN_MGMT_WORKLOAD_CVES') &&
+                    isFeatureFlagEnabled('ROX_POSTGRES_DATASTORE') && (
+                        // TODO We need to designate this as Tech Preview in a more standard way, based on UX guidance
+                        <NavExpandable
+                            id="Vulnerabilities"
+                            title="Vulnerabilities (preview)"
+                            isActive={vulnerabilitiesPaths.some((path) =>
+                                location.pathname.includes(path)
+                            )}
+                            isExpanded={vulnerabilitiesPaths.some((path) =>
+                                location.pathname.includes(path)
+                            )}
+                        >
+                            {vulnerabilitiesPaths.map((path) => {
+                                const isActive = location.pathname.includes(path);
+                                return (
+                                    <LeftNavItem
+                                        key={path}
+                                        isActive={isActive}
+                                        path={path}
+                                        title={basePathToLabelMap[path]}
+                                    />
+                                );
+                            })}
+                        </NavExpandable>
+                    )}
                 <NavExpandable
                     id="VulnerabilityManagement"
                     title="Vulnerability Management"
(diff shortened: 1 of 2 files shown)


dvail:
* master
  * **PR #4871** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4871" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4871?utm_source=stack-comment).

ui/apps/platform/src/Containers/MainPage/Sidebar/NavigationSidebar.tsx
```diff
@@ -122,6 +125,33 @@ function NavigationSidebar({
                     path={complianceBasePath}
                     title={basePathToLabelMap[complianceBasePath]}
                 />
+
+                {isFeatureFlagEnabled('ROX_VULN_MGMT_WORKLOAD_CVES') &&
+                    isFeatureFlagEnabled('ROX_POSTGRES_DATASTORE') && (
+                        // TODO We need to designate this as Tech Preview in a more standard way, based on UX guidance
```
dvail, line 131: UX is working on a consistent way to designate something as a tech preview in the UI, so for now, I just included plain text "(preview)" in the navigation.

This gets the point across, and is ugly enough that we won't forget to change it.

dvail: As a side note, `<NavExpandable>` only accepts a `string` for the `title` prop, unlike `<LeftNavItem>` which accepts a `string | ReactNode`. The PatternFly source renders the title directly as a child though, so there is no technical reason we couldn't pass a component for the `title`.

If we want to display the tech preview in the same way we are for the Network Graph 2.0, we will either need to do some type-casting or reach out to the PF team and see if the prop type can be updated.

pedrottimark: String seems safer and simpler.

Decorating every little detail slows us down and probably slows down visual processing of users too.

ui/apps/platform/src/routePaths.js
```diff
@@ -145,12 +152,18 @@ const vulnManagementPathToLabelMap = {
...
+    [vulnerabilitiesWorkloadCvesPath]: 'Workload CVEs',
+};
+
 export const basePathToLabelMap = {
     [dashboardPath]: 'Dashboard',
     [networkBasePath]: 'Network Graph (1.0)',
     [networkBasePathPF]: 'Network Graph (2.0 preview)',
     [violationsBasePath]: 'Violations',
     [complianceBasePath]: 'Compliance',
+    ...vulnerabilitiesPathToLabelMap,
```
pedrottimark, line 166: If container pages take responsibility to render `PageTitle` element, can we stop this invisible global connection?

ui/apps/platform/src/routePaths.js
```diff
@@ -91,6 +91,13 @@ export const vulnManagementPendingApprovalsPath = `${vulnManagementRiskAcceptanc
 export const vulnManagementApprovedDeferralsPath = `${vulnManagementRiskAcceptancePath}/approved-deferrals`;
 export const vulnManagementApprovedFalsePositivesPath = `${vulnManagementRiskAcceptancePath}/approved-false-positives`;
 
+// VM 2.0 "Vulnerabilities" paths
+export const vulnerabilitiesBasePath = `${mainPath}/vulnerabilities`;
+export const vulnerabilitiesWorkloadCvesPath = `${vulnerabilitiesBasePath}/workload-cves`;
+export const vulnerabilitiesWorkloadCveSinglePath = `${vulnerabilitiesBasePath}/workload-cves/cve/:cveId`;
+export const vulnerabilitiesWorkloadCveImageSinglePath = `${vulnerabilitiesBasePath}/workload-cves/image/:imageId`;
+export const vulnerabilitiesWorkloadCveDeploymentSinglePath = `${vulnerabilitiesBasePath}/workload-cves/deployment/:deploymentId`;
+
```
pedrottimark, line 100: What are pro and con of singular versus plural nouns?

Although classic Vulnerability Management used it to distinguish independent pages from side panels, this feature has a simpler information architecture, correct?

Other containers like collections use plural nouns analogous to REST convention:
* /main/collections for list page
* /main/collections/:collectionIid for individual page

pedrottimark (approved):
A few comments for your consideration

ui/apps/platform/src/routePaths.js
dvail: Correct, I'm not sure why I subconsciously keep gravitating towards the singular version. Agreed that plural is more consistent too, I'll update.

PR #4855, opened 2023-02-13
Delete classic Search modal because superseded by Search page

pedrottimark:
## Description

As a prerequisite to simplified solution to banner layout problems, remove superseded Search modal that depended on absolute position.

### Files changed

1. Remove `ROX_SEARCH_PAGE_UI` feature flag:
    * pkg/features/list.go
    * tests/e2e/lib.sh
    * ui/apps/platform/src/types/featureFlag.ts

2. Remove conditional rendering:
    * src/Containers/MainPage/Body.tsx
    * src/Containers/MainPage/Header/GlobalSearchButton.tsx
    * src/Containers/MainPage/MainPage.tsx

3. Update style:
    * src/css/acs.css add 1 rule
    * src/app.tw.css delete 2 rules

4. Update reducers and sagas
    * src/reducers/index.js
    * src/sagas/index.js

### Files deleted

1. Integration test
    * cypress/integration/search.test.js

2. Classic files in src/Containers/Search folder
    * ReduxSearchInput.js
    * SearchAutoComplete folder
    * SearchModal.js
    * SearchResults.tsx
    * searchCategoryDescriptorMap.ts

3. Reducers
    * src/reducers/globalSearch.js

4. Sagas
    * src/sagas/globalSearchSagas.js
    * src/sagas/searchSagas.js
    * src/sagas/searchSagas.test.js

### Residue

1. Only remaining occurrence of src/reducers/pageSearch.ts is in classic Network Graph
2. Investigate whether the same applies to ui/apps/platform/src/sagas/searchAutocompleteSagas.js

## Checklist
- [x] Investigated and inspected CI test results
- [x] Deleted obsolete integration tests

## Testing Performed

1. Click **Search** on masthead to open Search page without feature flag.
2. See link color override rule in Elements pane and see underline on hover, but decide no reason to override that browser default.

Search_link_inherit_and_underline

PR #4873, opened 2023-02-14
Define PrefixedAction type for reducers in TypeScript

pedrottimark:
## Description

First draft of a pattern for reducers in TypeScript.

Ironically, so much effort on action **creator function** types did nothing for actions in reducers :(

Also to support optional `params` which is only in globalSearch that will be deleted in #4855

`PrefixedAction<'whatever/WHATEVER', Response>` type has better type safety than the following classic function calls:
* `createFetchingActionTypes('whatever/WHATEVER')`
* `createFetchingActions(types.WHATEVER)`

If reducer has `switch` instead of `if` statements, then TypeScript reports errors for incorrect property access:
* `type`
* `response`
* `error`

Note that the change in SystemConfigForm reported a problem because of assumption that response is non-null. Bravo!

Notice an ancient copy-pasto: `'notifiers/FETCH_PUBLIC_CONFIG'`

### Residue

1. Initial attempt to use `PublicConfigAction` with types from redux-thunk failed :(

## Checklist

## Testing Performed

* ui/apps/platform/cypress/integration/systemConfig/systemConfig.test.js

### DIFF
--- a/ui/apps/platform/src/Containers/SystemConfig/Form/SystemConfigForm.tsx
+++ b/ui/apps/platform/src/Containers/SystemConfig/Form/SystemConfigForm.tsx
@@ -26,7 +26,7 @@ import { useFormik } from 'formik';
 
 import ColorPicker from 'Components/ColorPicker';
 import ClusterLabelsTable from 'Containers/Clusters/ClusterLabelsTable';
-import { types } from 'reducers/publicConfig';
+import { PublicConfigAction } from 'reducers/publicConfig';
 import { saveSystemConfig } from 'services/SystemConfigService';
 import { PrivateConfig, PublicConfig, SystemConfig } from 'types/config.proto';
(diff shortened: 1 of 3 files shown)


dvail (approved):
New generic action type is a win, and looks like it is already preventing possible bugs and typos. 🎉 

It looks like you could also prevent action type typos in the thunk with something like:
```typescript
export const fetchPublicConfigThunk = () => {
    return async (dispatch: (action: PublicConfigAction) => void) => {
```
but:
1. Since the consuming code is a mix of plain JS/`any`, it's hard to know for sure that those types are even correct.
2. My gut feeling is that getting full type correction here will require us to eventually convert everything in the reducers/store sections. (Hope I am wrong though 🤞 ).

pedrottimark:
Dave, thank you for helpful suggestion about types for the thunk function as this pattern matures.

PR #4879, opened 2023-02-15
ROX-10999: Fix panic in roxctl central db restore when no file is given

dhaus67:
## Description

This PR fixes a panic that occurs when no file is given within the positional args for `roxctl central db restore`.

The following has changed within the PR:
- Instead of using custom validation of the positional argument, use `cobra.ExactArgs()` instead.
- Mark the previously only _commented_ deprecated `--file` flag as _actually_ deprecated.
- Within the `validate()` step, ensure that its also validated that the file exists.

## Checklist

## Testing Performed

- see added unit tests.

### DIFF
--- a/roxctl/central/db/restore/v2.go
+++ b/roxctl/central/db/restore/v2.go
@@ -44,9 +44,10 @@ type centralDbRestoreCommand struct {
 func V2Command(cliEnvironment environment.Environment) *cobra.Command {
 	centralDbRestoreCmd := &centralDbRestoreCommand{env: cliEnvironment}
 	c := &cobra.Command{
-		Use: "restore <file>",
+		Use:  "restore <file>",
+		Args: cobra.ExactArgs(1),
 		RunE: func(c *cobra.Command, args []string) error {
-			if err := validate(c, args); err != nil {
+			if err := validate(c); err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


roxctl/central/db/restore/v2.go
```diff
@@ -99,6 +95,9 @@ func (cmd *centralDbRestoreCommand) validate() error {
 	if cmd.file == "" {
 		return errox.InvalidArgs.New("file to restore from must be specified")
 	}
+	if _, err := os.Stat(cmd.file); os.IsNotExist(err) {
```
janisz, line 98: What if `cmd.file` is directory?

roxctl/central/db/restore/v2_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/stackrox/rox/pkg/errox"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCentralDBRestore_Validate(t *testing.T) {
+	cmd := &centralDbRestoreCommand{}
+
+	// 1. If file is unset, expect an InvalidArgs error.
```
janisz, line 16: nit: How about wrapping them in subtest with this comment as a name? We can then run this tests in parallel if they do not share cmd

PR #4881, opened 2023-02-15
ROX-3208: Env variables for some roxctl flags

dhaus67:
## Description

This PR adds two environment variables to set roxctl flag values, specifically:
- `ROX_ADMIN_PASSWORD` for the `--password` flag.
- `ROX_ENDPOINT` for the `--endpoint` flag.

The logic of when to use the environment variable and when the flag value is as follows:
- Prefer the flag value over the environment variable, if the flag is explicitly set to a non-default value.
- If the default value for the flag is given, and has not been changed, prefer the environment variable.
- If neither the environment variable is set nor the flag is changed, use the default value.

For now, the current principles of the `flags` package have been re-use, but it may be worthwhile to do the following within a follow-up:
- split up the `flags` package either into separate sub-packages or a more generic package.
- generalize the `Changed` approach for all commands, and move away from package scoped global variables.

## Checklist

## Testing Performed

- added unit tests.

### DIFF
--- a/pkg/env/api.go
+++ b/pkg/env/api.go
@@ -1,6 +1,0 @@
-package env
-
-var (
-	// TokenEnv is the variable that clients can source for commandline operations
-	TokenEnv = RegisterSetting("ROX_API_TOKEN")
-)
--- a/pkg/env/roxctl.go
+++ b/pkg/env/roxctl.go
@@ -1,0 +1,12 @@
+package env
+
+var (
+	// EndpointEnv specifies the central endpoint to use for commandline operations.
+	EndpointEnv = RegisterSetting("ROX_ENDPOINT")
+
+	// PasswordEnv specifies the central admin password to use for commandline operations.
+	PasswordEnv = RegisterSetting("ROX_ADMIN_PASSWORD")
+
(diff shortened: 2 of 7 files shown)


dhaus67:
* master
  * **PR #4881** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4881" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4881?utm_source=stack-comment).

janisz (commented):
For long term solution I was thinking about using viper and allowing user to keep config in file, env or flag. What do you think about it?

dhaus67:
> For long term solution I was thinking about using viper and allowing user to keep config in file, env or flag. What do you think about it?

Yeah, I have this in mind as well. I wanted to split both the work of establishing this "generic" work to specify environment variables for _all_ global flags + the config file in a separate story, so I've created [ROX-15145](https://issues.redhat.com/browse/ROX-15145) for this. Main reason for me was that this is a bigger refactoring effort of the `roxctl/common/flags` pkg.

janisz (approved):

PR #4883, opened 2023-02-15
Dispatch OCP E2E tests via wildcards instead of symlinks

msugakov:
## Description

This should make it possible to configure OSCI job names flexibly for OpenShift Operator E2E and OpenShift E2E tests and drive those to the same Python scripts without having to create symlinks every time.

Now I doubt if we should do like in https://github.com/openshift/release/pull/36289#discussion_r1106928746 instead.

## Checklist

None of these needed:

## Testing Performed

* Relying on CI. Will add `/test` commands to see how all related jobs behave.


msugakov:
This should be done differently.

PR #4884, opened 2023-02-15
Change CI job_script running to simply exec

msugakov:
## Description

Not sure what was the reason for the existing setup, but I guess I'll find out through this PR.

## Checklist

None of these needed:

## Testing Performed

* CI will tell.

### DIFF
--- a/.openshift-ci/ci_tests.py
+++ b/.openshift-ci/ci_tests.py
@@ -149,7 +149,7 @@ class OperatorE2eTest(BaseTest):
 
 
 class QaE2eTestPart1(BaseTest):
-    TEST_TIMEOUT = 240 * 60
+    TEST_TIMEOUT = 400 * 60
 
     def run(self):
         print("Executing qa-tests-backend tests (part I)")
--- a/.openshift-ci/dispatch.sh
+++ b/.openshift-ci/dispatch.sh
@@ -59,13 +59,4 @@ else
     die "ERROR: There is no job script for $ci_job"
 fi
 
-"${job_script}" "$@" &
-job_pid="$!"
-
-forward_sigint() {
-    echo "Dispatch is forwarding SIGINT to job"
-    kill -SIGINT "${job_pid}"
-}
-trap forward_sigint SIGINT
-
-wait "${job_pid}"
+exec "${job_script}" "$@"
--- a/qa-tests-backend/scripts/run-part-1.sh
+++ b/qa-tests-backend/scripts/run-part-1.sh
@@ -16,6 +16,11 @@ set -euo pipefail
 run_part_1() {
     info "Starting test (qa-tests-backend part I)"
 
+    while true; do
+        echo "Sleeping for ten minutes..."
+        sleep 10m
+    done
+
     config_part_1
     test_part_1
 }


gavin-stackrox:
IIRC this is to ensure that the signal is propagated to the test job script. When OpenShift CI reaches its configured timeout it issues a SIGINT. That is used by the GKE cluster wrapper to teardown the cluster. With the exec I don't think the signal was propagated. In order to test you would need to hold the job for > then the 5h configured timeout on OSCI.

msugakov:
I see that in the former setup, the trap for SIGINT was forwarding this signal to the child process. The child process does not receive signals of the parent one that's where that trap helped.

`exec` by definition does not create child processes. `exec` just replaces the current process with the exec-ed one. All signals will go to the replacing one because it becomes _the_ current process.

msugakov:
@gavin-stackrox I induced the test timeout in
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4884/pull-ci-stackrox-stackrox-master-gke-qa-e2e-tests/1628111286099251200
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4884/pull-ci-stackrox-stackrox-master-gke-qa-e2e-tests/1628111286099251200/artifacts/gke-qa-e2e-tests/stackrox-e2e/build-log.txt

Based on my reading of the second log, the `exec` works as expected.
WDYT?

gavin-stackrox:
I'd feel more confident in the change if the wait was below the python layer. e.g. in https://github.com/stackrox/stackrox/blob/b637cb0510eb10f210d0f8ba7c68b167704631ae/qa-tests-backend/scripts/run-part-1.sh#L16

msugakov:
I don't think it's going to be the same.
This `run-part-1.sh` is called from here
https://github.com/stackrox/stackrox/blob/112635540879427384ca379557a4d35795bae5b2/.openshift-ci/ci_tests.py#L157-L159
where `run_with_graceful_kill` starts a child process with `Popen`
https://github.com/stackrox/stackrox/blob/112635540879427384ca379557a4d35795bae5b2/.openshift-ci/ci_tests.py#L17-L28

Any child process does not get signals that are sent to the parent one. In this case, the shell executing `run-part-1.sh` will not get interruption signal that's directed to Python process.
Also, the shell child process should continue running when Python process gets interrupted. It will get killed though when the CI container gets killed.

I'm not really sure what we'll test this way, but let me try and see what happens.

gavin-stackrox:
After looking back at the PR that created this, https://github.com/stackrox/stackrox/pull/2295/files#r914055566, I can explain this as follows: We want the cluster teardown function to execute when ci-operator terminates the job due to a timeout i.e. sends a SIGINT before it gives up and terminates the pod. If the job script were exec'd and sleeping for more then the grace period then the signal handler [here](https://github.com/stackrox/stackrox/blob/3d9b06ccb55e6d04f2b5ab82b25fe268f94a4574/.openshift-ci/clusters.py#L96) does not execute.

I think this was more of an issue in the early days of OSCI adoption before we bumped all the openshift/release timeouts to 5h.

gavin-stackrox:
~I think this was more of an issue in the early days of OSCI adoption before we bumped all the openshift/release timeouts to 5h.~
That is incorrect. This is needed when people push frequently to clear up GKE clusters.

msugakov:
`exec` does not work but wait and signal passing works for a reason that's beyond my understanding. I close this PR for now because I certainly won't pursue it in short term.

PR #4892, opened 2023-02-15
ROX-15113: Add support to suspend a CrobJob while enforcing a deploy time policy

clickboo:
## Description

RHACS falsely claimed that it enforced policies on CronJobs by setting "replicas to 0". However with cron one can only suspend the `CronJob`. Added support to enforce deploy time policies on `CronJob`.

## Checklist

## Testing Performed
Manually, on GKE and Openshift 4.6 by following steps to reproduce in the JIRA (thanks @msugakov !)
```common/enforcer: 2023/02/20 22:01:55.964640 enforcer.go:135: Info: enforcement successful. action enforcement: SCALE_TO_ZERO_ENFORCEMENTSCALE_TO_ZERO_ENFORCEMENT
deployment: <
  deployment_id: "06c8224d-9271-49ad-bc82-25d664d9fda5"
  deployment_name: "hello"
  deployment_type: "CronJob"
  namespace: "boo-test"
  alert_id: "dc23ac02-10f3-4807-b8a9-ae30bd3fe934"
  policy_name: "Boo Test policy"
>
```
```
ksanchet@ksanchet-mac:~/go/src/github.com/stackrox/stackrox$ kc get cronjobs/hello -n boo-test
NAME    SCHEDULE    SUSPEND   ACTIVE   LAST SCHEDULE   AGE
hello   * * * * *   True      0        <none>          9s
```

### DIFF
--- a/sensor/kubernetes/enforcer/cronjob/suspend.go
+++ b/sensor/kubernetes/enforcer/cronjob/suspend.go
@@ -51,8 +51,16 @@ func makePatch(deploymentInfo *central.DeploymentEnforcement, apiVersion string)
 
 // Suspend suspends the cron job
 func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
-	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
+	ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob)
+	if apiErr != nil {
+		return retry.MakeRetryable(apiErr)
+	}
+
+	if ok {
 		patchBytes, patchOptions, err := makePatch(deploymentInfo, batchV1)
+		if err != nil {
+			return err
+		}
 		_, err = client.BatchV1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(),
 			types.ApplyPatchType,
 			patchBytes,
@@ -62,6 +70,9 @@ func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *c
 		}
 	} else {
 		patchBytes, patchOptions, err := makePatch(deploymentInfo, batchV1beta1)
+		if err != nil {
+			return err
+		}
 		_, err = client.BatchV1beta1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(), types.ApplyPatchType,
 			patchBytes,
 			patchOptions)
--- a/sensor/kubernetes/enforcer/cronjob/suspend.go
+++ b/sensor/kubernetes/enforcer/cronjob/suspend.go
@@ -1,0 +1,87 @@
+package cronjob
+
+import (
+	"context"
+
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/pkg/branding"
+	kubernetesPkg "github.com/stackrox/rox/pkg/kubernetes"
+	"github.com/stackrox/rox/pkg/retry"
+	"github.com/stackrox/rox/sensor/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	batchV1      = "batch/v1"
+	batchV1beta1 = "batch/v1beta1"
+)
+
+// Suspend suspends the cron job
+func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+	forcePatch := true
+
+	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
+		patch := map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"name": deploymentInfo.GetDeploymentName(),
+			},
+			"kind":       deploymentInfo.GetDeploymentType(),
+			"apiVersion": batchV1,
+			"spec": map[string]interface{}{
+				"suspend": true,
+			},
+		}
+		patchBytes, err := json.Marshal(patch)
+		if err != nil {
+			return err
+		}
+
+		_, err = client.BatchV1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(),
+			types.ApplyPatchType,
+			patchBytes,
+			metav1.PatchOptions{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       deploymentInfo.GetDeploymentType(),
+					APIVersion: batchV1,
+				},
+				FieldManager: branding.GetProductName(),
+				Force:        &forcePatch,
+			})
+		if err != nil {
+			return retry.MakeRetryable(err)
+		}
+	} else {
+		patch := map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"name": deploymentInfo.GetDeploymentName(),
+			},
+			"kind":       deploymentInfo.GetDeploymentType(),
+			"apiVersion": batchV1beta1,
+			"spec": map[string]interface{}{
+				"suspend": true,
+			},
+		}
+		patchBytes, err := json.Marshal(patch)
+		if err != nil {
+			return err
+		}
+
+		_, err = client.BatchV1beta1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(), types.ApplyPatchType,
+			patchBytes,
+			metav1.PatchOptions{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       deploymentInfo.GetDeploymentType(),
+					APIVersion: batchV1beta1,
+				},
+				FieldManager: branding.GetProductName(),
+				Force:        &forcePatch,
+			})
+		if err != nil {
+			return retry.MakeRetryable(err)
+		}
+	}
+	return nil
+}
--- a/sensor/kubernetes/enforcer/cronjob/suspend.go
+++ b/sensor/kubernetes/enforcer/cronjob/suspend.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/stackrox/rox/generated/internalapi/central"
-	"github.com/stackrox/rox/pkg/branding"
 	kubernetesPkg "github.com/stackrox/rox/pkg/kubernetes"
 	"github.com/stackrox/rox/pkg/retry"
 	"github.com/stackrox/rox/sensor/utils"
@@ -17,6 +16,7 @@ import (
 const (
 	batchV1      = "batch/v1"
 	batchV1beta1 = "batch/v1beta1"
+	fieldManager = "StackRox"
 )
 
 // Suspend suspends the cron job
@@ -47,7 +47,7 @@ func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *c
 					Kind:       deploymentInfo.GetDeploymentType(),
 					APIVersion: batchV1,
 				},
-				FieldManager: branding.GetProductName(),
+				FieldManager: fieldManager,
 				Force:        &forcePatch,
 			})
 		if err != nil {
@@ -76,7 +76,7 @@ func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *c
 					Kind:       deploymentInfo.GetDeploymentType(),
 					APIVersion: batchV1beta1,
 				},
-				FieldManager: branding.GetProductName(),
+				FieldManager: fieldManager,
 				Force:        &forcePatch,
 			})
 		if err != nil {
--- a/sensor/kubernetes/enforcer/cronjob/suspend.go
+++ b/sensor/kubernetes/enforcer/cronjob/suspend.go
@@ -19,66 +19,52 @@ const (
 	fieldManager = "StackRox"
 )
 
-// Suspend suspends the cron job
-func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+func makePatch(deploymentInfo *central.DeploymentEnforcement, apiVersion string) ([]byte, metav1.PatchOptions, error) {
 	forcePatch := true
 
-	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
-		patch := map[string]interface{}{
-			"metadata": map[string]interface{}{
-				"name": deploymentInfo.GetDeploymentName(),
-			},
-			"kind":       deploymentInfo.GetDeploymentType(),
-			"apiVersion": batchV1,
-			"spec": map[string]interface{}{
-				"suspend": true,
-			},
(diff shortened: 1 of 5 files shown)


sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,50 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+// Suspend suspends the cron job
+func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+	depName := deploymentInfo.GetDeploymentName()
+	depType := deploymentInfo.GetDeploymentType()
+
+	patch := "{\"spec\" : {\"suspend\" : true }}"
+	if ok, apiErr := utils.HasAPI(client, "batch/v1", kubernetesPkg.CronJob); ok && apiErr != nil {
```
msugakov, line 22: Looks like this should be
```suggestion
	if ok, apiErr := utils.HasAPI(client, "batch/v1", kubernetesPkg.CronJob); ok && apiErr == nil {
```

sensor/utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"k8s.io/client-go/kubernetes"
+)
+
+// HasAPI checks whether the kubernetes client supports the gv API for the specified kind
+func HasAPI(client kubernetes.Interface, gv, kind string) (bool, error) {
```
msugakov, line 8: Let's please rename `gv` to `groupVersion` because `kind` is not shortened to `k`.
```suggestion
// HasAPI checks whether the kubernetes client supports the groupVersion API for the specified kind
func HasAPI(client kubernetes.Interface, groupVersion, kind string) (bool, error) {
```

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// Suspend suspends the cron job
+func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+	depName := deploymentInfo.GetDeploymentName()
+	depType := deploymentInfo.GetDeploymentType()
+	forcePatch := true
+
+	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
+		patch := fmt.Sprintf("{\"metadata\": {\"name\": %s}, \"kind\": %s, \"apiVersion\": %s, \"spec\": {\"suspend\": true}}", depName, depType, batchV1)
```
msugakov, line 29: It is concerning that any characters in `depName` that have special meaning for JSON will remain unescaped. I think, using `Unstructured` (`import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"`) and serializing that to JSON should be safer choice, although I have not tried that myself.

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,62 @@
...
+	batchBeta1v1 = "batch/beta1v1"
+)
+
+// Suspend suspends the cron job
+func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+	depName := deploymentInfo.GetDeploymentName()
+	depType := deploymentInfo.GetDeploymentType()
+	forcePatch := true
+
+	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
```
msugakov, line 28: Can you think of a way to not query this API each time a CronJob needs to be suspended? This may not be a lot of overhead, and we may drop the support for `batch/v1beta1` altogether in 4.0.0 release, but would be nice to reduce the usage if it does not lead to way more complicated code.

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stackrox/rox/pkg/retry"
+	"github.com/stackrox/rox/sensor/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	batchV1      = "batch/v1"
+	batchBeta1v1 = "batch/beta1v1"
```
msugakov, line 19: You probably mean
```suggestion
	batchBeta1v1 = "batch/v1beta1"
```

Once you're ready with the change please test it on OCP 4.6 or 4.7.
Speaking of testing, would you be able to extend E2E tests for the case of CronJobs?

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,62 @@
...
+		patch := fmt.Sprintf("{\"metadata\": {\"name\": %s}, \"kind\": %s, \"apiVersion\": %s, \"spec\": {\"suspend\": true}}", depName, depType, batchV1)
+
+		_, err = client.BatchV1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, depName, types.ApplyPatchType,
+			[]byte(patch),
+			metav1.PatchOptions{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       depType,
+					APIVersion: batchV1,
+				},
+				FieldManager: branding.GetProductName(),
```
msugakov, line 38: I think it should be sufficient to say `"StackRox"` here instead of `branding.GetProductName()`.
Judging that it was ok to use `StackRox` in https://github.com/stackrox/stackrox/blob/9bdc76a0e18b85b4439cfce1c48844015520b0d5/pkg/detection/deploytime/utils.go#L14-L14

msugakov (commented):
Thanks for finding a way to fix this

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: I'm going to leave it in there if especially `batch/v1beta1` support is going to be drooped.

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: changed to %q

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: Why? I'd say we change that in the "BLockedBy..." to reflect product name. We don't want StackRox to leak through the RH branding right?

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: Can TAL at the e2e test framework - extending that is more often than not a royal PITA.

sensor/kubernetes/enforcer/cronjob/suspend.go
msugakov: I believe `%q` does not escape all characters that matter for JSON (e.g. forward slash) and so it's not a proper replacement for JSON escaping. We can argue whether a forward slash or some other sequence is legit in `depName` in the first place, but why not use JSON escaping in the first place?

sensor/kubernetes/enforcer/cronjob/suspend.go
msugakov: > we change that in the "BLockedBy..."

This was just one nearby instance of many hundreds of others that mention `stackrox`. I don't think we can change that thing exposed in settings, environment variables, k8s objects labels and annotations that quickly.
Hence, I don't see a reason to make this spot any special and use branded name here.

My main concern about usage of branded name as a field manager is the restrictions that are put on the value
https://github.com/kubernetes/apimachinery/blob/6f5782f4e5f2798853afeb749157008669784e83/pkg/apis/meta/v1/types.go#L536-L541

In case the branded name gets too long, in case it gets some special characters (©, ™) or funny emoji, the code starts breaking.

I don't see a point to commit to such a time bomb especially without sufficient test coverage.

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: e2e framework has no cronjob support. maybe later, at best.

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: Changed, though I believe we are regressing. Funny characters and emojis should be within our control in branding.

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,87 @@
...
+		patch := map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"name": deploymentInfo.GetDeploymentName(),
+			},
+			"kind":       deploymentInfo.GetDeploymentType(),
+			"apiVersion": batchV1,
+			"spec": map[string]interface{}{
+				"suspend": true,
+			},
+		}
```
msugakov, line 36: This now is much closer to `Unstructured`. Not sure what prevented you from just adopting that.

The issue is that **very** similar fragment is repeated below. I don't see a reason for copy-paste if it can be avoided.

E.g. define `makePatch`

```go
func makePatch(deploymentInfo *central.DeploymentEnforcement, apiVersion string) ([]byte, metav1.PatchOptions, error) {
	forcePatch := true

	patch := map[string]interface{}{
		"metadata": map[string]interface{}{
			"name": deploymentInfo.GetDeploymentName(),
		},
		"kind":       deploymentInfo.GetDeploymentType(),
		"apiVersion": apiVersion,
		"spec": map[string]interface{}{
			"suspend": true,
		},
	}
	patchBytes, err := json.Marshal(patch)
	if err != nil {
		return nil, metav1.PatchOptions{}, err
	}

	options := metav1.PatchOptions{
		TypeMeta: metav1.TypeMeta{
			Kind:       deploymentInfo.GetDeploymentType(),
			APIVersion: apiVersion,
		},
		FieldManager: fieldManager,
		Force:        &forcePatch,
	}

	return patchBytes, options, nil
}
```

and use it in `Suspend`:

```go
func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {

	ok, err := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob)
	if err != nil {
		return retry.MakeRetryable(err)
	}

	api := batchV1
	if !ok {
		api = batchV1beta1
	}

	patchBytes, patchOptions, err := makePatch(deploymentInfo, api)
	if err != nil {
		return err
	}

	if ok {
		_, err = client.BatchV1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(),
			types.ApplyPatchType,
			patchBytes,
			patchOptions)
	} else {
		_, err = client.BatchV1beta1().CronJobs(deploymentInfo.GetNamespace()).Patch(ctx, deploymentInfo.GetDeploymentName(),
			types.ApplyPatchType,
			patchBytes,
			patchOptions)
	}
	if err != nil {
		return retry.MakeRetryable(err)
	}

	return nil
}
```

sensor/utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"k8s.io/client-go/kubernetes"
+)
+
+// HasAPI checks whether the kubernetes client supports the groupVersion API for the specified kind
```
msugakov, line 7: Spotted just now
```suggestion
// HasAPI checks whether the kubernetes server supports the groupVersion API for the specified kind
```

sensor/kubernetes/enforcer/cronjob/suspend.go
msugakov: I don't believe we, actually the person adjusting branding in the future, will remember about all the nuances of what's allowed or what's in the entire stack should be verified when changing the branding. Especially in the absence of end-to-end tests that would automatically catch regressions.

sensor/kubernetes/enforcer/cronjob/suspend.go
msugakov: I find absence of some code to be a questionable excuse against writing such code.
Do you mind creating a JIRA ticket for end-to-end tests for CronJob enforcement?

clickboo: Indeed, surprising the cron jobs fixes earlier didn't have tests either.
https://issues.redhat.com/browse/ROX-15452

sensor/utils/utils.go
clickboo: Fixed, I had just moved this function as is from elsewhere

sensor/kubernetes/enforcer/cronjob/suspend.go
clickboo: Modified, but not exactly ^

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,73 @@
...
+		FieldManager: fieldManager,
+		Force:        &forcePatch,
+	}
+
+	return patchBytes, options, nil
+}
+
+// Suspend suspends the cron job
+func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
+	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
```
msugakov, line 54: One thing was different in my proposal - handling `apiErr` with retries. Looks like we have different opinions https://github.com/stackrox/stackrox/pull/4892#discussion_r1123493773 , but when I stepped into the code `ServerResourcesForGroupVersion` I saw that it makes api request to k8s api server. Therefore, I suggest explicitly handling `apiErr` rather than assuming v1beta1:

```suggestion
	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); apiErr != nil {
		return retry.MakeRetryable(apiErr)
	} else if ok {
```

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -0,0 +1,73 @@
...
+		patchBytes, patchOptions, err := makePatch(deploymentInfo, batchV1)
```
msugakov, line 55: Please don't forget to handle `err` that comes out of `makePatch` (in both places).

msugakov:
@clickboo please check my comments above. These are from today, on the recent version of code.

clickboo:
Ping @msugakov , gracias!

sensor/kubernetes/enforcer/cronjob/suspend.go
```diff
@@ -51,8 +51,16 @@ func makePatch(deploymentInfo *central.DeploymentEnforcement, apiVersion string)
 
 // Suspend suspends the cron job
 func Suspend(ctx context.Context, client kubernetes.Interface, deploymentInfo *central.DeploymentEnforcement) (err error) {
-	if ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob); ok && apiErr == nil {
+	ok, apiErr := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob)
+	if apiErr != nil {
+		return retry.MakeRetryable(apiErr)
+	}
```
msugakov, line 57: [nit and optional, but...]
Unless there's a good reason, the style which seems dominant is to name `error` as just `err`, not `apiErr` or `somethingErr`.

```suggestion
	ok, err := utils.HasAPI(client, batchV1, kubernetesPkg.CronJob)
	if err != nil {
		return retry.MakeRetryable(err)
	}
```

msugakov (approved):

PR #4894, opened 2023-02-15
ROX-14702: Display origin in permission sets UI

ivan-degtiarenko:
## Description

Display origin(System, User or Declarative) in the permission set list view.

## Checklist

## Testing Performed

Screenshot 2023-02-15 at 21 42 34

### DIFF
--- a/ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
+++ b/ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
@@ -3,7 +3,6 @@ import {
     Alert,
     AlertVariant,
     Button,
-    Label,
     Modal,
     ModalVariant,
     PageSection,
@@ -12,11 +11,12 @@ import {
 } from '@patternfly/react-core';
 import { TableComposable, Tbody, Td, Thead, Th, Tr } from '@patternfly/react-table';
 
-import { getTraitsOriginLabel, isUserResource, PermissionSet, Role } from 'services/RolesService';
+import { PermissionSet, Role } from 'services/RolesService';
 
 import { AccessControlEntityLink, RolesLink } from '../AccessControlLinks';
 import usePermissions from '../../../hooks/usePermissions';
-import { originLabelColours } from '../../../constants/accessControl';
+import { TraitsOriginLabel } from '../TraitsOriginLabel';
+import { isUserResource } from '../traits';
 
 const entityType = 'PERMISSION_SET';
 
@@ -78,10 +78,10 @@ function PermissionSetsList({
                 <TableComposable variant="compact" isStickyHeader>
                     <Thead>
                         <Tr>
-                            <Th width={20}>Name</Th>
+                            <Th width={15}>Name</Th>
                             <Th width={15}>Origin</Th>
-                            <Th width={30}>Description</Th>
-                            <Th width={40}>Roles</Th>
+                            <Th width={25}>Description</Th>
+                            <Th width={35}>Roles</Th>
                             <Th width={10} aria-label="Row actions" />
                         </Tr>
                     </Thead>
@@ -96,9 +96,7 @@ function PermissionSetsList({
                                     />
                                 </Td>
                                 <Td dataLabel="Origin">
-                                    <Label color={originLabelColours[getTraitsOriginLabel(traits)]}>
-                                        {getTraitsOriginLabel(traits)}
-                                    </Label>
+                                    <TraitsOriginLabel traits={traits} />
                                 </Td>
                                 <Td dataLabel="Description">{description}</Td>
                                 <Td dataLabel="Roles">
--- a/ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
+++ b/ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
@@ -3,6 +3,7 @@ import {
     Alert,
     AlertVariant,
     Button,
+    Label,
     Modal,
     ModalVariant,
     PageSection,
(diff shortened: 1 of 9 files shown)


ui/apps/platform/src/services/RolesService.ts
```diff
@@ -79,13 +80,32 @@ export function fetchUserRolePermissions(): Promise<{ response: Role }> {
 
 const permissionSetsUrl = '/v1/permissionsets';
 
+export type Traits = {
+    mutabilityMode: MutabilityMode;
+    origin?: Origin;
+};
+
+export type MutabilityMode = 'ALLOW_MUTATE' | 'ALLOW_MUTATE_FORCED';
+export type Origin = 'IMPERATIVE' | 'DECLARATIVE' | 'DEFAULT';
```
pedrottimark, line 89: Because TypeScript does not scope these types within `Traits` like the corresponding proto file, I often prepend it like a prefix to reduce possible ambiguity in code that import the types:
* `TraitsMutabilityMode`
* `TraitsOrigin`

pedrottimark: Recommended as a followup after you merge, because AuthService and RolesService both depend on these types, please factor out so frontend file structure is more parallel to backend file structure:
* backend: https://github.com/stackrox/stackrox/blob/master/proto/storage/traits.proto
* frontend: src/types/traits.proto.ts if type is in a storage proto file, but in frontend services file if (as is more and more often) backend type is in backend service proto file

pedrottimark: Having recommended to factor out `Traits` on the one side of the coin, here is a question on the other side of the coin. Do the optionality of properties of `traits` property in responses depend on the requests?

For example, I noticed below `traits == null` which seems slightly inconsistent with `traits?: Traits` because that would mean `traits?: Traits | null`

So, if there is a non-nullish `traits` property in response for access control requests, is `origin` property required?

ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
```diff
@@ -77,13 +79,14 @@ function PermissionSetsList({
                     <Thead>
                         <Tr>
                             <Th width={20}>Name</Th>
+                            <Th width={15}>Origin</Th>
```
pedrottimark, line 82: The percents add up to more than 100. Do we need to deduct from one of the other columns?

ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
```diff
@@ -92,6 +95,11 @@ function PermissionSetsList({
                                         entityName={name}
                                     />
                                 </Td>
+                                <Td dataLabel="Origin">
+                                    <Label color={originLabelColours[getTraitsOriginLabel(traits)]}>
+                                        {getTraitsOriginLabel(traits)}
+                                    </Label>
```
pedrottimark, line 101: If all access control entities will render this information on form and table, what do you think about a follow up  to factor out as something like `<TraitsOriginLabel traits=(permissionSet.traits} />`

And then might it be possible to render with initially undefined value even for create case above in form?

pedrottimark (approved):
Most comments are for followup, but please double check the following questions:
* Sum of percents in table needs to be 100
* Does response property need to be `traits?: Traits | null`

ui/apps/platform/src/services/RolesService.ts
ivan-degtiarenko: > Do the optionality of properties of traits property in responses depend on the requests?
> So, if there is a non-nullish traits property in response for access control requests, is origin property required?

Currently, `traits` are treated as an optional API request field for auth providers, groups, roles etc. That means that is possible to create role with empty `traits`, or role with `traits` and no `origin`. For simplicity, the `traits` are not filled with default value on the backend, meaning we should correctly handle resources with empty `traits` or empty `origin`.

> For example, I noticed below traits == null which seems slightly inconsistent with traits?: Traits because that would mean traits?: Traits | null

I don't entirely get what is the inconsistency - perhaps because I'm still struggling to learn how all the different ways of saying "there is no value" in Typescript work :)

ui/apps/platform/src/Containers/AccessControl/PermissionSets/PermissionSetsList.tsx
ivan-degtiarenko: Factored out `TraitsOriginLabel`

ui/apps/platform/src/services/RolesService.ts
```diff
@@ -79,13 +80,32 @@ export function fetchUserRolePermissions(): Promise<{ response: Role }> {
...
 
+export function getTraitsOriginLabel(traits?: Traits): string {
+    return traits && traits.origin && traitsOriginLabels[traits.origin]
+        ? traitsOriginLabels[traits.origin]
+        : 'User';
+}
+
+export function isUserResource(traits?: Traits): boolean {
+    return traits == null || traits.origin === 'IMPERATIVE';
+}
```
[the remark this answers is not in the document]
ivan-degtiarenko: Addressed here

ui/apps/platform/src/services/RolesService.ts
pedrottimark: At first I misread `traits == null` as `traits === null` but then understood the `==` idiom for either
* `undefined` which corresponds to `?` in `traits?: Traits`
* `null` which I did see in the response for some request elsewhere in access control, therefore `traits?: Traits | null` describes that either absence of property or presence of property with `null` value are possible.

Benefit of being as precise as possible on this detail:
* The backend proto in this case is not clear that `null` or `undefined` are possible.
* It reduces the risk that engineers (either backend or frontend) who are not part of the original feature team, who might need to modify the code later, might write less robust predicates than your helper functions. By the way, factoring out the helper functions like you did is a huge help to make the code maintainable.

pedrottimark (approved):
Thank you for the extra effort to make requested changes.

Also for starting with one access control entity. The pattern looks clear for the others.

PR #4895, opened 2023-02-15
ROX-15059: Catch initial collector restarts

gavin-stackrox:
## Description

ROX-15059 highlights unhelpful CI failure patterns when collector cannot load its kernel object. In that situation the e2e tests tend to stagger on to timeout or fail a large set of tests and at no point indicate that collector is crashlooping. This PR checks that collector reaches a 'ready state' by checking for a log line that is currently indicative of such.

## Checklist

## Testing Performed

https://prow.ci.openshift.org/pr-history/?org=stackrox&repo=stackrox&pr=4895

The three test runs above show in right to left order:
- Behavior with a [broken collector](https://github.com/stackrox/collector/pull/1012) prior to these changes.
- Behavior with a [broken collector](https://github.com/stackrox/collector/pull/1012) and these changes. Tests all fail with "Stackrox was not deployed" and the initial error that collector is not ready.
- Behavior with a happy collector.


tests/e2e/lib.sh
```diff
@@ -331,6 +335,40 @@ setup_podsecuritypolicies_config() {
...
+    local readiness_indicator="Successfully established GRPC stream for signals"
+    local start_time
+    start_time="$(date '+%s')"
+    local all_ready="false"
+    while [[ "$all_ready" == "false" ]]; do
+        all_ready="true"
+        for pod in $(kubectl -n stackrox get pods -l app=collector -o json | jq -r '.items[].metadata.name'); do
+            echo "Checking readiness of $pod"
+            if kubectl -n stackrox logs -c collector "$pod" | grep "$readiness_indicator" > /dev/null 2>&1; then
+                echo "ready"
```
tommartensen, line 352: This output is a bit useless, could you invert the condition (and pull up the else branch up)?

tommartensen (commented):
Why are we not relying on the readiness probe status of the collector pod?

tests/e2e/lib.sh
```diff
@@ -331,6 +335,40 @@ setup_podsecuritypolicies_config() {
...
+            else
+                info "$pod is not ready"
+                kubectl -n stackrox logs -c collector "$pod"
+                all_ready="false"
+                break
+            fi
+        done
+        if (( $(date '+%s') - start_time > 300 )); then
```
JoukoVirtanen, line 360: [nit] You could create a variable to contain the timeout period and use that here and in the echo.

tests/e2e/lib.sh
```diff
@@ -331,6 +335,40 @@ setup_podsecuritypolicies_config() {
...
+            echo "ERROR: Collector readiness check timed out after 5m"
+            echo "Not all collector logs contain: $readiness_indicator"
+            exit 1
+        fi
+        if [[ "$all_ready" == "false" ]]; then
+            info "Found at least one unready collector pod, will check again in 10s"
```
JoukoVirtanen, line 366: [nit] Same as above.

JoukoVirtanen (approved):
LGTM!

tests/e2e/lib.sh
```diff
@@ -331,6 +335,40 @@ setup_podsecuritypolicies_config() {
     fi
 }
 
+# wait_for_collectors_to_be_operational() ensures that collector pods are able
+# to load kernel objects and create network connections.
+wait_for_collectors_to_be_operational() {
+    info "Will wait for collectors to reach a ready state"
+
+    local readiness_indicator="Successfully established GRPC stream for signals"
```
JoukoVirtanen, line 343: What if the log statement in collector changes? I think it would be better to check something like if collector reports its own process or something like that.

tests/e2e/lib.sh
gavin-stackrox: > This output is a bit useless, could you invert the condition (and pull up the else branch up)?

I'm not a fan of weird inverted grep syntax for readability and I've tested this path so I'd hate to switch it around at this point :( I'd still echo something in the ready state.

gavin-stackrox:
> Why are we not relying on the readiness probe status of the collector pod?

Good Q. I don't know if that tells us the same thing i.e. that collector has started and loaded its kernel object and made necessary network connections.

tests/e2e/lib.sh
gavin-stackrox: I agree this is a bit fragile (but if someone updates the collector they will have to fix this). I'm happy to check something else if you have a concrete suggestion.

JoukoVirtanen: I suggested checking if collector has reported its own process above, but that might not be the best way as that would rely on sensor and central. You could check collector metrics and make sure that the number of processes or network connections is not 0.

gavin-stackrox:
I'm going to take Jouko's LGTM for the win. If you guys have a better check please note on ROX-15059 and reopen.

PR #4899, opened 2023-02-16
ROX-15101 nack to compliance if central connection is not ready

lvalerom:
## Description

The idea is to have a mechanism to notify `SensorComponents` when the connection with central is established.

I'm thinking of making the `NotifyReady` function more generic and have something like `Notify(msg)` so we can reuse this for other types of notifications and not just notifying when the connection with central is ready.

Note to reviewers: Consider reviewing the PR by commit. The second and forth are the most interesting 🙂 

## Checklist

## Testing Performed

* Tested sending node inventories on sensor's startup with a broken connection.

### DIFF
--- a/sensor/common/component.go
+++ b/sensor/common/component.go
@@ -6,12 +6,20 @@ import (
 	"google.golang.org/grpc"
 )
 
+// SensorComponentEvent represents events about which sensor components can be notified
+type SensorComponentEvent string
+
+const (
+	// SensorComponentEventCentralReachable denotes that Sensor-Central connection is up
+	SensorComponentEventCentralReachable SensorComponentEvent = "central-reachable"
+)
+
 // SensorComponent is one of the components that constitute sensor. It supports for receiving messages from central,
 // as well as sending messages back to central.
 type SensorComponent interface {
 	Start() error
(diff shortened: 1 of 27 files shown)


sensor/common/component.go
```diff
@@ -6,11 +6,20 @@ import (
 	"google.golang.org/grpc"
 )
 
+// SensorComponentEvent represents events about which sensor components can be notified
+type SensorComponentEvent string
+
+const (
+	// SensorComponentEventCentralReachable denotes that Sensor-Central connection is up
+	SensorComponentEventCentralReachable SensorComponentEvent = "central-reachable"
```
Maddosaurus, line 14: Optional nit: Though this is a very expressive var name, it is very long.
I struggle to come up with some shorter ones as well, but how about something like `CentralReached` or `CentralContacted`?

Maddosaurus (approved):
I see you've gone for a more generic Notify function, very cool.
The nit is optional, just a thought that ocured to me.

vikin91 (commented):
Ups, apparently I forgot to submit my review comments. Here they are (all of them are addressed currently).

jschnath (approved):
LGTM

PR #4906, opened 2023-02-16
ROX-13603 use the storeProvider in all components

lvalerom:
## Description

Move all the stores in sensor to the storeProvider. This will allow us to get rid of the singletons for the stores.

## Checklist

## Testing Performed

* CI


sensor/kubernetes/sensor/sensor.go
```diff
@@ -63,7 +62,10 @@ func CreateSensor(cfg *CreateOptions) (*sensor.Sensor, error) {
 		log.Infof("Running sesnor with Kubernetes re-sync enabled. Re-sync time: %s", cfg.resyncPeriod.String())
 	}
 
-	admCtrlSettingsMgr := admissioncontroller.NewSettingsManager(resources.DeploymentStoreSingleton(), resources.PodStoreSingleton())
+	storeProvider := resources.InitializeStore()
+	scan.SetMatchingRegistryFunction(storeProvider.Registries().GetRegistryForImage)
```
fredrb, line 66: It makes sense what you did here, but I think it leads to an API that can be a bit confusing. We're injecting a dependency that is used by a public static function (`EnrichLocalImage`), and if we forget to set it we've introduced a panic in:
```go
// sensor/common/scan/scan.go:54
matchingRegistry, err := getMatchingRegistry(ci.GetName())
```

Since you're at this, my suggestion is to convert `sensor/common/scan/scan.go` in a struct and `EnrichLocalImage` in a method of that struct. You can request a `func(*storage.ImageName) (registryTypes.Registry, error)` parameter or the registry store directly. This makes the dependency injection clearer and less error prone. 

Additionally, you can inject this new struct in the only place where it's used, rather than having a public API that is not used anywhere: `sensor/common/image/service_impl.go`. This service already has a dependency on registry store, so if you want to initialize the dependency in the service constructor that's also possible.

fredrb (approved):
Looks good. But please consider doing the refactor I've mentioned.

sensor/kubernetes/sensor/sensor.go
lvalerom: That's a good suggestion. I've implemented the changes creating a `Scan` struct and wrapping all those functions in it. Let me know what you think.

sensor/common/detector/enricher.go
```diff
@@ -120,7 +123,7 @@ outer:
 	}
 }
 
-func (c *cacheValue) scanAndSet(ctx context.Context, svc v1.ImageServiceClient, req *scanImageRequest) {
+func (c *cacheValue) scanAndSet(ctx context.Context, svc v1.ImageServiceClient, req *scanImageRequest, scan *scan.LocalScan) {
```
fredrb, line 126: I think this parameter is no longer needed?

sensor/common/detector/enricher.go
```diff
@@ -194,11 +198,12 @@ func (e *enricher) runScan(req *scanImageRequest) imageChanResult {
...
 
 	newValue := &cacheValue{
-		signal: concurrency.NewSignal(),
+		signal:    concurrency.NewSignal(),
+		localScan: e.localScan,
 	}
 	value := e.imageCache.GetOrSet(key, newValue).(*cacheValue)
 	if forceEnrichImageWithSignatures || newValue == value {
-		value.scanAndSet(concurrency.AsContext(&e.stopSig), e.imageSvc, req)
+		value.scanAndSet(concurrency.AsContext(&e.stopSig), e.imageSvc, req, e.localScan)
```
fredrb, line 206: Also remove it from here.

fredrb (approved):

sensor/common/detector/enricher.go
lvalerom: Oops! Fixed! 😅

sensor/common/detector/enricher.go
lvalerom: Fixed

PR #4908, opened 2023-02-16
CI: Send a slack message on style or unit test failures

gavin-stackrox:
## Description

Per title. 

## Checklist

## Testing Performed

n/a will run on merge.

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -104,3 +104,18 @@ jobs:
 
       - name: make style
         run: make style
+
+  slack-on-style-failure:
+    if: github.event_name == 'push' && failure()
+    name: Post Style Failure
+    runs-on: ubuntu-latest
+    needs:
+      - check-generated-files
+      - misc-checks
+      - style-check
+    steps:
+      - name: Slack Workflow Notification
+        uses: Gamesight/slack-workflow-status@v1.1.0
+        with:
+          repo_token: ${{secrets.GITHUB_TOKEN}}
+          slack_webhook_url: ${{secrets.TEST_FAILURES_NOTIFY_WEBHOOK}}
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -116,3 +116,19 @@ jobs:
 
     - name: Local roxctl tests
       run: ./scripts/ci/jobs/local-roxctl-tests.sh
+
+  slack-on-style-failure:
+    if: github.event_name == 'push' && failure()
+    name: Post Style Failure
+    runs-on: ubuntu-latest
+    needs:
+      - go
+      - go-postgres
+      - ui
+      - local-roxctl-tests
+    steps:
+      - name: Slack Workflow Notification
+        uses: Gamesight/slack-workflow-status@v1.1.0
+        with:
+          repo_token: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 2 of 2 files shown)


janisz:
How does the message looks like and where it's posted?

gavin-stackrox:
> How does the message looks like and where it's posted?

https://srox.slack.com/archives/CLUNQEEMA/p1676563606462909

.github/workflows/style.yaml
```diff
@@ -104,3 +104,18 @@ jobs:
...
         run: make style
+
+  slack-on-style-failure:
+    if: github.event_name == 'push' && failure()
+    name: Post Style Failure
+    runs-on: ubuntu-latest
+    needs:
+      - check-generated-files
+      - misc-checks
+      - style-check
```
msugakov, line 115: If someone adds a new job to the workflow but forgets to update this list, we'll not have notifications for that new job. I wasn't able to find a way to do something different than `needs:` in order to make this job run last. Do you know if there's a way to maintain this list automatically?

.github/workflows/unit-tests.yaml
```diff
@@ -116,3 +116,19 @@ jobs:
 
     - name: Local roxctl tests
       run: ./scripts/ci/jobs/local-roxctl-tests.sh
+
+  slack-on-style-failure:
```
msugakov, line 120:
```suggestion
  slack-on-unit-tests-failure:
```

gavin-stackrox:
Good points @msugakov. I've limited permission for that token and pinned the version with a SHA.

.github/workflows/style.yaml
gavin-stackrox: Not that I know of either.

msugakov (commented):
Looks good, but I suggest couple further refinements.

.github/workflows/style.yaml
```diff
@@ -107,15 +107,17 @@ jobs:
...
       - check-generated-files
       - misc-checks
       - style-check
+    permissions:
+      read-all
     steps:
       - name: Slack Workflow Notification
-        uses: Gamesight/slack-workflow-status@v1.1.0
+        uses: Gamesight/slack-workflow-status@26a36836c887f260477432e4314ec3490a84f309
         with:
```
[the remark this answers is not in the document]
gavin-stackrox: `include_commit_message` :heavy_check_mark: 
`include_jobs` is already there by default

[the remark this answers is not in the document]
msugakov: `include_jobs: on-failure` should not output succeeding jobs which the default `include_jobs: true` does. Anyway, I'm not sure myself which way is better and definitely don't insist.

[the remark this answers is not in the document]
gavin-stackrox: Ah I see. I think that is better.

.github/workflows/style.yaml
msugakov: A way to achieve this would be to use GH API to get all jobs of the workflow how mentioned here https://stackoverflow.com/a/71241789/484050
And compare that with the jobs mentioned in `needs` context https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

I guess, I'll create a JIRA ticket.

msugakov: https://issues.redhat.com/browse/ROX-15200

msugakov (approved):

PR #4912, opened 2023-02-16
Don't crash on query timeouts and add env var to not crash in case of misattributed error code

connorgorman:
## Description

Statement timeouts should not crash Central as they are unlikely to resolve if we just retry them; however, we may have misattributed an error code so adding an env var to protect against that if necessary

## Checklist

## Testing Performed

CI and will try to cause some cases

### DIFF
--- a/pkg/postgres/pgutils/error.go
+++ b/pkg/postgres/pgutils/error.go
@@ -12,6 +12,8 @@ import (
 	"github.com/stackrox/rox/pkg/set"
 )
 
+const queryCancelledCode = "57014"
+
 var transientPGCodes = set.NewFrozenStringSet(
 	// Class 08 — Connection Exception
 	"08000", // connection_exception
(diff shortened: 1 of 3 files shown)


pkg/postgres/pgutils/error.go
```diff
@@ -47,6 +49,14 @@ var transientPGCodes = set.NewFrozenStringSet(
 	"58030", // io_error
 )
 
+// IsQueryTimeoutError specifies if the error is a statement timeout error
```
dhaus67, line 52: super-nit, or should the func be public:
```suggestion
// isQueryTimeoutError specifies if the error is a statement timeout error
```

pkg/postgres/pgutils/error.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/stackrox/rox/pkg/set"
 )
 
+const queryCancelledCode = "57014"
```
dhaus67, line 15: Maybe reuse this in l.41?

PR #4916, opened 2023-02-16
fix: lint warning fixes

sachaudh:
## Description

This PR fixes a few of the lint warnings from our Lint Bash today

Lint Bash Doc: https://docs.google.com/document/d/1mz8a7Lp058HUIpMgw6FUETvaKVPyegE0bgY4YBDsAFs/edit#heading=h.u0xqdtd6qrsb

## Checklist

### DIFF
--- a/ui/apps/platform/src/Components/Pagination/PaginationInput/PaginationInput.js
+++ b/ui/apps/platform/src/Components/Pagination/PaginationInput/PaginationInput.js
@@ -10,8 +10,8 @@ const TYPING_DELAY = 800;
 const PaginationInput = ({ totalSize, onChange, currentPage, pageSize }) => {
     const [localPage, setLocalPage] = useState(currentPage);
     const delayedSetPage = useCallback(
-        debounce((newPage) => onChange(newPage), TYPING_DELAY),
-        []
+        () => debounce((newPage) => onChange(newPage), TYPING_DELAY),
+        [onChange]
     );
 
     const totalPages = Math.ceil(totalSize / pageSize);
--- a/ui/apps/platform/src/Components/PatternFly/FormMessage.tsx
+++ b/ui/apps/platform/src/Components/PatternFly/FormMessage.tsx
@@ -4,7 +4,6 @@ import { Alert, AlertVariant } from '@patternfly/react-core';
 export type FormResponseMessage = {
     message: string;
     isError: boolean;
-    data?: any;
 } | null;
 
 export type FormMessageProps = {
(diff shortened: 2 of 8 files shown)


ui/apps/platform/src/Components/PatternFly/FormMessage.tsx
```diff
@@ -4,7 +4,6 @@ import { Alert, AlertVariant } from '@patternfly/react-core';
 export type FormResponseMessage = {
     message: string;
     isError: boolean;
-    data?: any;
```
vjwilson, line 7: Curious about this change. Do we not actually use data, or are we just not typing it, so that it can be free-form (basically an implicit typing of any) ?

vjwilson (approved):

ui/apps/platform/src/Components/Pagination/PaginationInput/PaginationInput.js
```diff
@@ -10,8 +10,8 @@ const TYPING_DELAY = 800;
 const PaginationInput = ({ totalSize, onChange, currentPage, pageSize }) => {
     const [localPage, setLocalPage] = useState(currentPage);
     const delayedSetPage = useCallback(
-        debounce((newPage) => onChange(newPage), TYPING_DELAY),
-        []
+        () => debounce((newPage) => onChange(newPage), TYPING_DELAY),
```
pedrottimark, line 13: For my understanding, was original `debounce(…)` incorrect? It does look like a subtle problem.

What are pro and con when arrow function does a side-effect instead of returns a value?
* implicitly `undefined` return value: `(newPage) => onChange(newPage), TYPING_DELAY)`
* explicitly no meaningful return value:

```js
(newPage) => {
    onChange(newPage), TYPING_DELAY);
}
```

ui/apps/platform/src/Components/PatternFly/FormMessage.tsx
pedrottimark: Good eyes!

This component does not use the `data` property.

Saif, can you confirm that `responseData` is the key of the optional property, and here are its uses:
* https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/ApiTokenIntegrationForm/ApiTokenFormMessageAlert.tsx#L15-L32
* https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Integrations/IntegrationForm/Forms/ClusterInitBundleIntegrationForm/ClusterInitBundleFormMessageAlert.tsx#L69-L75

ui/apps/platform/src/Containers/Network/SidePanel/NetworkDeploymentOverlay/Flows/BaselineSettings/useFetchNetworkBaselines.ts
```diff
@@ -89,6 +89,7 @@ function useFetchNetworkBaselines({
             });
         // TODO: Possibly use another value other than selectedDeployment to ensure this logic
         // is executed again. See following comment: https://github.com/stackrox/rox/pull/7254#discussion_r555252326
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
pedrottimark, line 92: Work around warning in classic code, pending its removal, correct?

pedrottimark (approved):
Thank you for attention to these details.
* A few questions for clarification or understanding.
* One comment for your consideration.

PR #4917, opened 2023-02-16
ROX-15086: Junitify pod log checks

gavin-stackrox:
## Description

Failures from the script that checks pod logs for suspicious entries are hidden in CI output. This PR creates a JUnit record for them.

## Checklist
  - how it looks now (testing with a check for stackrox in logs :raised_eyebrow:): https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4917/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests/1626659698134487040
 
## Testing Performed

CI is sufficient

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -458,6 +458,8 @@ check_for_errors_in_stackrox_logs() {
         # shellcheck disable=SC2086
         if ! check_out="$(scripts/ci/logcheck/check.sh $filtered)"; then
             save_junit_failure "SuspiciousLog" "Suspicious entries in log file(s)" "$check_out"
+            sleep 1
+            save_junit_failure "SuspiciousLog" "Suspicious entries in log file(s)" "$check_out"
             die "ERROR: Found at least one suspicious log file entry."
         else
             save_junit_success "SuspiciousLog" "Suspicious entries in log file(s)"
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -454,9 +454,13 @@ check_for_errors_in_stackrox_logs() {
     # shellcheck disable=SC2010,SC2086
     filtered=$(ls $logs | grep -Ev "(previous|_describe).log$" || true)
     if [[ -n "$filtered" ]]; then
+        local check_out=""
         # shellcheck disable=SC2086
-        if ! scripts/ci/logcheck/check.sh $filtered; then
+        if ! check_out="$(scripts/ci/logcheck/check.sh $filtered)"; then
+            save_junit_failure "SuspiciousLog" "Suspicious entries in log file(s)" "$check_out"
             die "ERROR: Found at least one suspicious log file entry."
+        else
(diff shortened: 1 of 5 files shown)


tests/e2e/lib.sh
```diff
@@ -454,9 +454,13 @@ check_for_errors_in_stackrox_logs() {
     # shellcheck disable=SC2010,SC2086
     filtered=$(ls $logs | grep -Ev "(previous|_describe).log$" || true)
     if [[ -n "$filtered" ]]; then
+        local check_out=""
         # shellcheck disable=SC2086
-        if ! scripts/ci/logcheck/check.sh $filtered; then
+        if ! check_out="$(scripts/ci/logcheck/check.sh $filtered)"; then
```
msugakov, line 459: [no-op] TIL this way to both capture the output and exit code.

tests/e2e/lib.sh
```diff
@@ -454,9 +454,13 @@ check_for_errors_in_stackrox_logs() {
...
+            save_junit_failure "SuspiciousLog" "Suspicious entries in log file(s)" "$check_out"
```
msugakov, line 460: Can I ask you to also change this fragment to wrap `${details}` in XML CDATA?
https://github.com/stackrox/stackrox/blob/883967db3a31301a58970af4a909ff8a85499850/scripts/ci/lib.sh#L1401-L1405

This would be:
```xml
<testsuite name="${class}" tests="1" skipped="0" failures="1" errors="0">
    <testcase name="${description}" classname="${class}">
        <failure><![CDATA[${details}]]></failure>
    </testcase>
</testsuite>
```

The reason is that CDATA allows having `<`, `>` and other characters that are otherwise considered as special in XML and which presence may mess up parsing.

---

Additionally, WDYT about including a timestamp in the file name in this line?
https://github.com/stackrox/stackrox/blob/883967db3a31301a58970af4a909ff8a85499850/scripts/ci/lib.sh#L1400

Same about `save_junit_success` function.
This will help us not be afraid that reusing the "class name" will lead to overwriting the file created by some other step.

tests/e2e/lib.sh
gavin-stackrox: yes. Brad and I figured that out pair programming when running the script twice seemed lame.

tests/e2e/lib.sh
gavin-stackrox: It is a bit outside the scope of this PR but I see the value :( I will need to test that having >1 junits for the same class is supported well by prow/spyglass.

gavin-stackrox:
Looks good with CDATA and prow can handle duplicate logs: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_stackrox/4917/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests/1626659698134487040

BradLugo (approved):

PR #4918, opened 2023-02-16
ROX-15166-ROX-15167: Update decode-uri-component library

vjwilson:
## Description

A needed dependency update

## Checklist

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -45,7 +45,8 @@
     },
     "resolutions": {
         "@typescript-eslint/parser": "^4.32.0",
-        "autoprefixer": "10.4.5"
+        "autoprefixer": "10.4.5",
+        "decode-uri-component": "^0.4.1"
     },
     "devDependencies": {
         "chalk-cli": "^5.0.0",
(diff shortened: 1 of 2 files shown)


sachaudh (approved):

vjwilson:
Closing this particular PR, because forcing the `decode-uri-component` to upgrade to the latest causes builds to fail. The latest version of that module is ESM-only, and our current build process tries to reduce its import to CommonJS style.

This will require some chain-reaction upgrades of our app ecosystem that we need to catch up on.

```
$ yarn build
yarn run v1.22.19
$ yarn clean
$ rm -rf build test-results; lerna run clean --stream
/Users/vwilson/go/src/github.com/stackrox/stackrox/ui/node_modules/snapdragon/node_modules/source-map-resolve/lib/decode-uri-component.js:1
var decodeUriComponent = require("decode-uri-component")
                         ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/vwilson/go/src/github.com/stackrox/stackrox/ui/node_modules/decode-uri-component/index.js from /Users/vwilson/go/src/github.com/stackrox/stackrox/ui/node_modules/snapdragon/node_modules/source-map-resolve/lib/decode-uri-component.js not supported.
Instead change the require of index.js in /Users/vwilson/go/src/github.com/stackrox/stackrox/ui/node_modules/snapdragon/node_modules/source-map-resolve/lib/decode-uri-component.js to a dynamic import() which is available in all CommonJS modules.
```

PR #4923, opened 2023-02-17
ROX-15164: Reconciliation loop for declarative configuration

dhaus67:
## Description

This PR adds the reconciliation loop for declarative configurations.

After the contents of the declarative configuration mounts will be watched by multiple watch handlers (introduced within [ROX-14147](https://issues.redhat.com/browse/ROX-14147), the next step is to start with the reconciliation of the transformed proto messages.

This includes the following:
- Creating a "complete" view of all transformed proto messages, keyed by their respective watch handler directory.
- Creating a loop that will continously iterate over the transformed proto messages and allow for upserting proto messages as well as deleting proto messages.

Besides these two points, the following was done as well:
- Instead of specifying the watch handler intervals as well as the reconciliation loop interval as `const`s, exposed environment variables for these two settings for further customization if required in some environments.
- The reconciliation loop includes a short-circuit functionality, bypassing the reconciliation loop interval to trigger the reconciliation when changes are reported by the watch handlers.

As follow-ups, the following will be required:
- Handle the upsertion of proto messages within the reconciliation step, covered by [ROX-14693](https://issues.redhat.com/browse/ROX-14693).
- Handle the deletion of proto messages within the reconciliation step, covered by [ROX-14694](https://issues.redhat.com/browse/ROX-14694).

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/declarativeconfig/manager_impl.go
+++ b/central/declarativeconfig/manager_impl.go
@@ -11,7 +11,6 @@ import (
 	"github.com/stackrox/rox/pkg/concurrency"
 	"github.com/stackrox/rox/pkg/declarativeconfig"
 	"github.com/stackrox/rox/pkg/declarativeconfig/transform"
-	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/k8scfgwatch"
 	"github.com/stackrox/rox/pkg/maputil"
 	"github.com/stackrox/rox/pkg/sync"
@@ -35,20 +34,19 @@ type managerImpl struct {
 	reconciliationTickerDuration time.Duration
 	watchIntervalDuration        time.Duration
 
-	reconciliationTicker     *time.Ticker
-	reconciliationInProgress concurrency.Flag
-	shortCircuitSignal       concurrency.Signal
+	reconciliationTicker *time.Ticker
+	shortCircuitSignal   concurrency.Signal
 }
 
 // New creates a new instance of Manager.
 // Note that it will not watch the declarative configuration directories when created, only after
 // ReconcileDeclarativeConfigurations has been called.
-func New() Manager {
+func New(reconciliationTickerDuration, watchIntervalDuration time.Duration) Manager {
 	return &managerImpl{
 		universalTransformer:         transform.New(),
 		transformedMessagesByHandler: map[string]protoMessagesByType{},
-		reconciliationTickerDuration: env.DeclarativeConfigReconcileInterval.DurationSetting(),
-		watchIntervalDuration:        env.DeclarativeConfigWatchInterval.DurationSetting(),
+		reconciliationTickerDuration: reconciliationTickerDuration,
+		watchIntervalDuration:        watchIntervalDuration,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


central/declarativeconfig/manager_impl.go
```diff
@@ -8,39 +8,51 @@ import (
...
 // New creates a new instance of Manager.
 // Note that it will not watch the declarative configuration directories when created, only after
-// WatchDeclarativeConfigDir has been called.
+// ReconcileDeclarativeConfigurations has been called.
 func New() Manager {
 	return &managerImpl{
-		universalTransformer: transform.New(),
+		universalTransformer:         transform.New(),
+		transformedMessagesByHandler: map[string]protoMessagesByType{},
+		reconciliationTickerDuration: env.DeclarativeConfigReconcileInterval.DurationSetting(),
```
ivan-degtiarenko, line 50: I would expect these parameters to be passed to the `New` method and extracted from `env` in a Singleton

central/declarativeconfig/manager_impl.go
```diff
@@ -87,5 +109,64 @@ func (m *managerImpl) ReconcileDeclarativeConfigs(contents [][]byte) {
 			transformedConfigurations[protoType] = append(transformedConfigurations[protoType], protoMessages...)
 		}
 	}
-	// No-op, see the TODOs within the function comment.
+
+	m.transformedMessagesMutex.Lock()
+	defer m.transformedMessagesMutex.Unlock()
```
ivan-degtiarenko, line 114: there's no reason to hold the lock while executing `m.shortCircuitReconciliationLoop()`

central/declarativeconfig/manager_impl.go
```diff
@@ -87,5 +109,64 @@ func (m *managerImpl) ReconcileDeclarativeConfigs(contents [][]byte) {
...
+			m.shortCircuitSignal.Reset()
+			m.runReconciliation()
+		case <-m.reconciliationTicker.C:
+			m.runReconciliation()
+		}
+	}
+}
+
+func (m *managerImpl) runReconciliation() {
+	// We shouldn't trigger a parallel reconciliation run, hence we should ensure that the flag is not set to true.
```
ivan-degtiarenko, line 150: I just now realised that we can't trigger parallel runs either way - all `runReconciliation` calls are happening in the one goroutine. Perhaps I'm missing something?

central/declarativeconfig/manager_impl.go
```diff
@@ -87,5 +109,64 @@ func (m *managerImpl) ReconcileDeclarativeConfigs(contents [][]byte) {
...
+}
+
+func (m *managerImpl) startReconciliationLoop() {
+	m.reconciliationTicker = time.NewTicker(m.reconciliationTickerDuration)
+
+	go m.reconciliationLoop()
+}
+
+func (m *managerImpl) reconciliationLoop() {
+	// While we currently do not have an exist in the form of "stopping" the reconciliation, still, ensure that
```
ivan-degtiarenko, line 135:
```suggestion
	// While we currently do not have an exit in the form of "stopping" the reconciliation, still, ensure that
```

central/declarativeconfig/manager_impl.go
```diff
@@ -87,5 +109,64 @@ func (m *managerImpl) ReconcileDeclarativeConfigs(contents [][]byte) {
...
+	m.transformedMessagesMutex.Lock()
+	defer m.transformedMessagesMutex.Unlock()
+	m.transformedMessagesByHandler[handlerID] = transformedConfigurations
+	m.shortCircuitReconciliationLoop()
+}
+
+// shortCircuitReconciliationLoop will short circuit the reconciliation loop and trigger a reconciliation loop run.
+// Note that the reconciliation loop will not be run if:
+//   - the short circuit loop signal has not been reset yet and is de-duped.
+//   - a current reconciliation loop run is in progress.
```
ivan-degtiarenko, line 122: I don't think that's the case - my understanding is that `for select` is a single-thread, consecutive execution mechanism. Thus after the reconciliation is finished, we go into a new iteration of the cycle and handle short circuit signal

dhaus67: Addressed within the comment above.

ivan-degtiarenko (approved):
LGTM 👍

PR #4926, opened 2023-02-17
Fix a few linter issues

bradr5:
## Description

Fixes a few linter issues

## Checklist

### DIFF
--- a/ui/apps/platform/src/hooks/useTableSelection.ts
+++ b/ui/apps/platform/src/hooks/useTableSelection.ts
@@ -32,7 +32,7 @@ function useTableSelection<T extends Base>(
 
     React.useEffect(() => {
         setSelected(data.map(preSelectedFunc));
-    }, [data]);
+    }, [data, preSelectedFunc]);
 
     const onClearAll = () => {
         setSelected(data.map(() => false));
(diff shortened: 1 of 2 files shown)


vjwilson (approved):

ui/apps/platform/src/hooks/useTableSelection.ts
```diff
@@ -32,7 +32,7 @@ function useTableSelection<T extends Base>(
 
     React.useEffect(() => {
         setSelected(data.map(preSelectedFunc));
-    }, [data]);
+    }, [data, preSelectedFunc]);
```
pedrottimark, line 35: A question about `preSelectedFunc` optional argument for my understanding.

Does the following have any potential benefit to minimize unneeded re-rendering?
* replace anonymous arrow function, for which every instantiation has different referential identity
* with reference to function assigned in module scope, which never changes

```ts
const preSelectedFuncDefault = () => false;

function useTableSelection<T extends Base>(
    data: T[],
    // determines whether value should be pre-selected or not
    preSelectedFunc: (T) => boolean = preSelectedFuncDefault
): UseTableSelection {
    // and so on
```

Only 1 out of 6 occurrences call with optional argument, the other 5 depend on default value.

pedrottimark (approved):

PR #4927, opened 2023-02-17
Add samples of real full image scan

md2119:
## Description

This PR adds some sample real image scans that can be used in unit tests for datastores, graphQL, API, etc. These images have a couple 1000 vuls.

## Checklist

## Testing Performed
Unit


pkg/fixtures/image/reader.go
```diff
@@ -0,0 +1,51 @@
...
+	dir = "files"
+)
+
+var (
+	//go:embed files/*.json
+	fs embed.FS
+)
+
+// GetTestImages returns a slice of images for testing purposes. These images contain a snapshot of image scan.
+func GetTestImages(_ *testing.T) ([]*storage.Image, error) {
```
charmik-redhat, line 24: In addition to this, would a func that can take a file name and return a single image be useful?

charmik-redhat (approved):
LGTM . Left a minor suggestion.

pkg/fixtures/image/reader.go
md2119: It doesn't hurt to have it. I will club it with other changes in one of the follow-ups.

PR #4929, opened 2023-02-17
ROX-14521: Log user identities which successfully logged in

ivan-degtiarenko:
## Description

Log user identities so that we can later collect them via CloudWatch and thus detect unauthorized access.
We log `ExchangeToken` results as well as it is used in basic auth in the UI(and potentially can be used with other auth providers)

Example of log record:
```
Warn: User successfully logged in with user attributes {"userID": "", "serviceID": "{\"id\":\"a\",\"initBundleId\":\"b\"}", "expires": "2023-02-27T19:22:16.405955Z", "username": "UserID", "friendlyName": "Madonna Slytherin", "roleNames": ["Admin", "Analyst"], "permissions": "{\"id\":\"a\",\"initBundleId\":\"b\"}", "authProvider": {"Id":"","Name":"authProviderName","Type":""}}
```

P.S. The reason to convert `serviceId` and `permissions` to JSON string is to improve readability of these fields. Compare to raw proto formatting:
```
Warn: User successfully logged in with user attributes {"userID": "", "serviceID": "id:\"a\" init_bundle_id:\"b\" ", "expires": "2023-02-27T20:53:27.559186Z", "username": "UserID", "friendlyName": "Madonna Slytherin", "roleNames": ["Admin", "Analyst"], "permissions": "resource_to_access:<key:\"Close Magic Doors\" value:READ_ACCESS > resource_to_access:<key:\"Open Magic Doors\" value:READ_WRITE_ACCESS > ", "authProvider": {"Id":"","Name":"authProviderName","Type":""}}

```

## Checklist

## Testing Performed

1. Run Central
2. Accessed via OIDC, Certificates and basic auth providers
3. Saw corresponding records in the central logs

### DIFF
--- a/central/authprovider/service/service_impl.go
+++ b/central/authprovider/service/service_impl.go
@@ -322,6 +322,9 @@ func (s *serviceImpl) ExchangeToken(ctx context.Context, request *v1.ExchangeTok
 
 	userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
 	if err != nil {
+		if testMode {
+			return nil, errors.Wrap(err, "cannot create role based identity")
+		}
 		log.Warnf("error creating role based identity: %v", err)
 	}
 	userPkg.LogSuccessfulUserLogin(log, userMetadata)
--- a/central/authprovider/service/service_impl.go
+++ b/central/authprovider/service/service_impl.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stackrox/rox/pkg/auth/authproviders/basic"
 	"github.com/stackrox/rox/pkg/auth/authproviders/idputil"
 	"github.com/stackrox/rox/pkg/auth/permissions"
+	userPkg "github.com/stackrox/rox/pkg/auth/user"
 	"github.com/stackrox/rox/pkg/errox"
 	"github.com/stackrox/rox/pkg/grpc/authz"
 	"github.com/stackrox/rox/pkg/grpc/authz/allow"
@@ -319,12 +320,13 @@ func (s *serviceImpl) ExchangeToken(ctx context.Context, request *v1.ExchangeTok
 		Test:        testMode,
 	}
 
+	userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
+	if err != nil {
+		log.Warnf("error creating role based identity: %v", err)
+	}
+	userPkg.LogSuccessfulUserLogin(log, userMetadata)
+
 	if testMode {
-		// We need all access for retrieving roles.
-		userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
-		if err != nil {
-			return nil, errors.Wrap(err, "cannot create role based identity")
-		}
 		response.User = userMetadata
 		return response, nil
 	}
--- a/central/authprovider/service/service_impl.go
+++ b/central/authprovider/service/service_impl.go
@@ -325,7 +325,7 @@ func (s *serviceImpl) ExchangeToken(ctx context.Context, request *v1.ExchangeTok
 		if testMode {
 			return nil, errors.Wrap(err, "cannot create role based identity")
 		}
-		log.Warnf("error creating role based identity: %v", err)
+		log.Warnf("Error creating role based identity: %v", err)
 	}
 	userPkg.LogSuccessfulUserLogin(log, userMetadata)
 
--- a/pkg/auth/user/util.go
+++ b/pkg/auth/user/util.go
@@ -41,10 +41,9 @@ func protoToJSON(log *logging.Logger, message proto.Message) string {
 	result, err := jsonutil.ProtoToJSON(message, jsonutil.OptCompact, jsonutil.OptUnEscape)
 	if err != nil {
 		log.Error("Failed to convert proto to JSON: ", err)
-	} else {
-		return result
+		return ""
 	}
-	return ""
+	return result
 }
 
 // LogSuccessfulUserLogin logs user attributes in the specified logger instance.
@@ -55,7 +54,7 @@ func LogSuccessfulUserLogin(log *logging.Logger, user *v1.AuthStatus) {
 		serviceIdStr = protoToJSON(log, user.GetServiceId())
 	}
 	if user.GetUserInfo().GetPermissions() != nil {
-		permissionsStr = protoToJSON(log, user.GetServiceId())
+		permissionsStr = protoToJSON(log, user.GetUserInfo().GetPermissions())
 	}
 	log.Warnw("User successfully logged in with user attributes",
 		zap.String("userID", user.GetUserId()),
--- a/pkg/auth/user/util.go
+++ b/pkg/auth/user/util.go
@@ -3,9 +3,12 @@ package user
 import (
 	"sort"
 
+	"github.com/gogo/protobuf/proto"
 	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/auth/permissions/utils"
 	"github.com/stackrox/rox/pkg/jsonutil"
 	"github.com/stackrox/rox/pkg/logging"
+	"go.uber.org/zap"
 )
 
 // ConvertAttributes converts a map of user attributes to v1.UserAttribute
@@ -28,13 +31,43 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
 	return result
 }
 
+type loggableAuthProvider struct {
+	Id   string
+	Name string
+	Type string
+}
+
+func protoToJSON(log *logging.Logger, message proto.Message) string {
+	result, err := jsonutil.ProtoToJSON(message, jsonutil.OptCompact, jsonutil.OptUnEscape)
+	if err != nil {
+		log.Error("Failed to convert proto to JSON: ", err)
+	} else {
+		return result
+	}
+	return ""
+}
+
 // LogSuccessfulUserLogin logs user attributes in the specified logger instance.
 func LogSuccessfulUserLogin(log *logging.Logger, user *v1.AuthStatus) {
-	loggableUser := user.Clone()
-	// Auth provider config can contain sensitive data(client secret, certificates etc.) so it shouldn't be logged.
-	if loggableUser != nil {
-		loggableUser.AuthProvider = nil
+	serviceIdStr := ""
+	permissionsStr := ""
+	if user.GetServiceId() != nil {
+		serviceIdStr = protoToJSON(log, user.GetServiceId())
 	}
-	userJSON, _ := jsonutil.ProtoToJSON(loggableUser)
-	log.Warnf("User successfully logged in with user attributes: %s", userJSON)
+	if user.GetUserInfo().GetPermissions() != nil {
+		permissionsStr = protoToJSON(log, user.GetServiceId())
+	}
+	log.Warnw("User successfully logged in with user attributes",
+		zap.String("userID", user.GetUserId()),
+		zap.String("serviceID", serviceIdStr),
+		zap.Any("expires", user.GetExpires()),
+		zap.String("username", user.GetUserInfo().GetUsername()),
+		zap.String("friendlyName", user.GetUserInfo().GetFriendlyName()),
+		zap.Any("roleNames", utils.RoleNamesFromUserInfo(user.GetUserInfo().GetRoles())),
+		zap.String("permissions", permissionsStr),
+		zap.Any("authProvider", &loggableAuthProvider{
+			Id:   user.GetAuthProvider().GetId(),
+			Type: user.GetAuthProvider().GetType(),
+			Name: user.GetAuthProvider().GetName(),
+		}))
 }
--- a/pkg/auth/user/util.go
+++ b/pkg/auth/user/util.go
@@ -4,6 +4,8 @@ import (
 	"sort"
 
 	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/jsonutil"
+	"github.com/stackrox/rox/pkg/logging"
 )
 
 // ConvertAttributes converts a map of user attributes to v1.UserAttribute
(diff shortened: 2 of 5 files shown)


central/authprovider/service/service_impl.go
```diff
@@ -319,12 +320,13 @@ func (s *serviceImpl) ExchangeToken(ctx context.Context, request *v1.ExchangeTok
 		Test:        testMode,
 	}
 
+	userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
+	if err != nil {
+		log.Warnf("error creating role based identity: %v", err)
+	}
+	userPkg.LogSuccessfulUserLogin(log, userMetadata)
+
```
dhaus67, line 328: I'd say we only want to log successful user logins _iff_ we are not in test mode. I'm not sure how helpful it would be to log test mode users.

pkg/auth/user/util.go
```diff
@@ -25,3 +27,14 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
 	})
 	return result
 }
+
+// LogSuccessfulUserLogin logs user attributes in the specified logger instance.
+func LogSuccessfulUserLogin(log *logging.Logger, user *v1.AuthStatus) {
+	loggableUser := user.Clone()
+	// Auth provider config can contain sensitive data(client secret, certificates etc.) so it shouldn't be logged.
+	if loggableUser != nil {
+		loggableUser.AuthProvider = nil
```
stehessel, line 36: Do you know if our logging framework supports obfuscating secrets? If we set the auth provider to nil, it could lead to wrong conclusions when viewing the logs, and we also lose all non-secret data.

central/authprovider/service/service_impl.go
ivan-degtiarenko: From API perspective, the test mode is just a flag. This flag does not prevent API consumers from obtaining the resulting token - so unauthorised access is still possible even when test mode is enabled.

pkg/auth/user/util.go
ivan-degtiarenko: I don't see `zap` supporting secret obfuscating out of the box. We could do this by introducing a custom type for a secret field, but unfortunately, our client secrets are kept in `map[string]string` structure.

Overall, the idea behind my decision is two-fold: the data contained within auth provider config can be sensitive + not needed. If you want to determine which auth provider was used to log in the user, perhaps we can log just auth provider id?

stehessel: I do think that obfuscating sensitive values leads to more insightful logs and is therefore superior to just setting them to `nil`. How about creating a ticket to add such a functionality? Until then we can go with the `nil` solution to unblock the login logs.

pkg/auth/user/util.go
```diff
@@ -25,3 +27,14 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+	}
+	userJSON, _ := jsonutil.ProtoToJSON(loggableUser)
+	log.Warnf("User successfully logged in with user attributes: %s", userJSON)
```
stehessel, line 39: I'd suggest to use a structured log entry here instead of adding the json to the message string directly.

ivan-degtiarenko: Do you suggest going over the fields we want to log individually in the structured log:
```
log.Warnw("User successfully logged in with user attributes",
		zap.String("userID", loggableUser.GetUserId()),
         zap.String("username", loggableUser.GetUserInfo().GetUsername())
```
**or** just to add JSON to the log via `Warnw`:
```
log.Warnw("User successfully logged in with user attributes",
		zap.String("userEntry", userJSON))
```
**or** log proto directly via `Warnw`:
```
log.Warnw("User successfully logged in with user attributes",
		zap.Any("userEntry", loggableUser))
```

The problem with the second approach is that it messes up formatting - we would get all escape characters written down:
```
{"userEntry": "{\n  \"userId\": \"UserID\",\n  \"expires\": \"2023-02-20T21:25:09.770338Z\",\n  \"userInfo\": {\n    \"username\": \"UserID\",\n    \"friendlyName\": \"Madonna Slytherin\",\n    \"permissions\": {\n      \"resourceToAccess\": {\n        \"Close Magic Doors\": \"READ_ACCESS\",\n        \"Open Magic Doors\": \"READ_WRITE_ACCESS\"\n      }\n    }\n  },\n  \"userAttributes\": [\n    {\n      \"key\": \"a\",\n      \"values\": [\n        \"b\"\n      ]\n    }\n  ]\n}"}
```

The problem with the third approach is that also messes up formatting + does not support google protobuf timestamps:
```
 {"userEntry": "user_id:\"UserID\" expires:<seconds:1676928078 nanos:435893000 > user_info:<username:\"UserID\" friendly_name:\"Madonna Slytherin\" permissions:<resource_to_access:<key:\"Close Magic Doors\" value:READ_ACCESS > resource_to_access:<key:\"Open Magic Doors\" value:READ_WRITE_ACCESS > > > user_attributes:<key:\"a\" values:\"b\" > "}
```

pkg/auth/user/util.go
ivan-degtiarenko: I'll create a ticket

pkg/auth/user/util.go
rhybrillou: I think it makes sense to select the fields we want to log rather than logging everything except specific elements.
This is less prone to inadvertently logging a sensitive field that could be added to the data in the future.

pkg/auth/user/util.go
dhaus67: Why would we not be reusing the `[RedactConfig](https://github.com/stackrox/stackrox/blob/21d0873bacc91ae9b27771f09d655e0c8c66e6a9/pkg/auth/authproviders/backend_factory.go#L37)` method from each provider instead of recreating some different way? This is already used to publish config to the UI, so it'd be a good opportunity to use it here as well.

pkg/auth/user/util.go
```diff
@@ -25,3 +27,14 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
 	})
 	return result
 }
+
+// LogSuccessfulUserLogin logs user attributes in the specified logger instance.
+func LogSuccessfulUserLogin(log *logging.Logger, user *v1.AuthStatus) {
```
dhaus67, line 32: Its a bit awkward to have the logger being passed here. Does it really matter that the origin package will be the auth service?

ivan-degtiarenko: I think the reason why it matters is because we call this function from two places, so having those calls distinguished could be helpful in debugging

pkg/auth/user/util.go
ivan-degtiarenko: I switched to logging selected fields instead of the entire `v1.AuthStatus`, which should solve this issue.

pkg/auth/user/util.go
ivan-degtiarenko: I switched to logging selected fields instead of the entire `v1.AuthStatus`, which should solve this issue.
One good argument that persuaded me is that we want to decouple logged structure from `v1.AuthStatus` as future developers might add a new field with sensitive data.

pkg/auth/user/util.go
stehessel: I think this could be separated into a function that just returns the logging context as an array of `zap.Field`. Then the caller calls the logger, and we just pass the logging context.

pkg/auth/user/util.go
```diff
@@ -25,3 +30,44 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+}
+
+func protoToJSON(log *logging.Logger, message proto.Message) string {
+	result, err := jsonutil.ProtoToJSON(message, jsonutil.OptCompact, jsonutil.OptUnEscape)
+	if err != nil {
+		log.Error("Failed to convert proto to JSON: ", err)
+	} else {
+		return result
+	}
+	return ""
```
rhybrillou, line 47: nit:
```
if err != nil {
    log.Error(...)
    return ""
}
return result
```

pkg/auth/user/util.go
```diff
@@ -25,3 +30,44 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+
+// LogSuccessfulUserLogin logs user attributes in the specified logger instance.
+func LogSuccessfulUserLogin(log *logging.Logger, user *v1.AuthStatus) {
+	serviceIdStr := ""
+	permissionsStr := ""
+	if user.GetServiceId() != nil {
+		serviceIdStr = protoToJSON(log, user.GetServiceId())
+	}
+	if user.GetUserInfo().GetPermissions() != nil {
+		permissionsStr = protoToJSON(log, user.GetServiceId())
```
rhybrillou, line 58: Shouldn't it be `user.GetUserInfo().GetPermissions()` that is passed to `protoToJSON` here ?

pkg/auth/user/util.go
ivan-degtiarenko: We agreed with @stehessel that having a special function to call `log.Warnw` makes sense because it allows guaranteeing several callers will log users with the same format, message and log level.
However, it still is beneficial to extract the function that creates logging context - `extractUserLogFields` as it allows for easier unit testing.

pkg/auth/user/util_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"time"
+
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/protoconv"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+)
+
+func TestExtractUserLogFields(t *testing.T) {
```
stehessel, line 14: 👍 for adding unit tests. Can you split this into multiple tests, or at least a golang-style table test? In general I always opt to keep the number of assertions per test at a minimum.

pkg/auth/user/util.go
```diff
@@ -25,3 +32,52 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+}
+
+// The reason this function returns []interface{} instead of []zap.Field
+// is because log.Warnw accepts ...interface{} and []zap.Field does not convert automatically
+// to []interface{}.
+func extractUserLogFields(user *v1.AuthStatus) []interface{} {
+	serviceIdStr := ""
+	permissionsStr := ""
+	if user.GetServiceId() != nil {
+		serviceIdStr = protoToJSON(user.GetServiceId())
```
stehessel, line 63: Maybe check for `nil` in `protoToJSON` and return empty string in that case? Then you can save a few lines here and the function is better protected against nil pointer dereferences.

ivan-degtiarenko: I already tried that and didn't succeed - apparently in the Golang interface "hides" `nil` and check like `proto == nil` wouldn't work. We would need to use reflection there, and I think it's better to avoid that: https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go

stehessel: Hm that is unfortunate.

pkg/auth/user/util_test.go
```diff
@@ -0,0 +1,108 @@
...
+				},
+				{
+					Name: "Analyst",
+				},
+			},
+		},
+	}
+	fields := extractUserLogFields(user)
+	assert.Len(t, fields, 9)
+	assert.True(t, fields[0].(zap.Field).Equals(zap.String("userID", user.GetUserId())))
```
stehessel, line 47: nit: you could use `.Contains` here, such that you don't have to specify the order - which we don't care about.

stehessel (approved):

central/authprovider/service/service_impl.go
```diff
@@ -319,12 +320,16 @@ func (s *serviceImpl) ExchangeToken(ctx context.Context, request *v1.ExchangeTok
...
-	if testMode {
-		// We need all access for retrieving roles.
-		userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
-		if err != nil {
+	userMetadata, err := authproviders.CreateRoleBasedIdentity(sac.WithAllAccess(ctx), provider, authResponse)
+	if err != nil {
+		if testMode {
 			return nil, errors.Wrap(err, "cannot create role based identity")
 		}
+		log.Warnf("error creating role based identity: %v", err)
```
dhaus67, line 328: nit:
```suggestion
		log.Warnf("Error creating role based identity: %v", err)
```

pkg/auth/user/util.go
```diff
@@ -25,3 +32,52 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+}
+
+// The reason this function returns []interface{} instead of []zap.Field
+// is because log.Warnw accepts ...interface{} and []zap.Field does not convert automatically
+// to []interface{}.
+func extractUserLogFields(user *v1.AuthStatus) []interface{} {
+	serviceIDStr := ""
+	permissionsStr := ""
+	if user.GetServiceId() != nil {
+		serviceIDStr = protoToJSON(user.GetServiceId())
```
dhaus67, line 63: nit:
```suggestion
		serviceIDJSON = protoToJSON(user.GetServiceId())
```

pkg/auth/user/util.go
```diff
@@ -25,3 +32,52 @@ func ConvertAttributes(attrMap map[string][]string) []*v1.UserAttribute {
...
+	}
+	if user.GetUserInfo().GetPermissions() != nil {
+		permissionsStr = protoToJSON(user.GetUserInfo().GetPermissions())
```
dhaus67, line 66: nit:
```suggestion
		permissionsJSON = protoToJSON(user.GetUserInfo().GetPermissions())
```

pkg/auth/user/util.go
```diff
@@ -3,9 +3,16 @@ package user
...
 
+	"github.com/gogo/protobuf/proto"
 	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/pkg/auth/permissions/utils"
+	"github.com/stackrox/rox/pkg/jsonutil"
+	"github.com/stackrox/rox/pkg/logging"
+	"go.uber.org/zap"
 )
 
+var utilLog = logging.LoggerForModule()
```
dhaus67, line 14: Rename the parameter from `log` to `logger` within `LogSuccessfulUserLogin` and name this `log`? Since there's no other log within the package, don't see why it needs to be named explicitly here.

dhaus67 (approved):

rhybrillou (approved):

PR #4939, opened 2023-02-17
UI lint errors fix

alwayshooin:
there should be no lint errors in `Containers/NetworkGraph` and `Containers/Policies`

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/components/StyleFakeGroup.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/components/StyleFakeGroup.tsx
@@ -11,6 +11,7 @@ import {
 import AlternateIcon from '@patternfly/react-icons/dist/esm/icons/regions-icon';
 import DefaultIcon from '@patternfly/react-icons/dist/esm/icons/builder-image-icon';
 import useDetailsLevel from '@patternfly/react-topology/dist/esm/hooks/useDetailsLevel';
+import { SVGIconProps } from '@patternfly/react-icons/dist/js/createIcon';
 import DefaultFakeGroup from './DefaultFakeGroup';
 
 const ICON_PADDING = 20;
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
```diff
@@ -71,7 +72,7 @@ const StyleGroup: React.FunctionComponent<StyleGroupProps> = ({
                 delete newData[key];
             }
         });
-        return newData;
+        return newData as CustomGroupNodeData;
```
pedrottimark, line 75: What do you think about `React.useMemo<CustomGroupNodeData>` instead?

https://www.freecodecamp.org/news/react-typescript-how-to-set-up-types-on-hooks/#set-types-on-usememo

ui/apps/platform/src/Containers/NetworkGraph/components/StyleFakeGroup.tsx
```diff
@@ -11,6 +11,7 @@ import {
 import AlternateIcon from '@patternfly/react-icons/dist/esm/icons/regions-icon';
 import DefaultIcon from '@patternfly/react-icons/dist/esm/icons/builder-image-icon';
 import useDetailsLevel from '@patternfly/react-topology/dist/esm/hooks/useDetailsLevel';
+import { SVGIconProps } from '@patternfly/react-icons/dist/js/createIcon';
```
pedrottimark, line 14: Can you search PatternFly repo to see if there is an issue for next major version 5 to improve like:

```ts
import { SVGIconProps } from '@patternfly/react-icons';
```

pedrottimark (approved):
Thank you. Two comments for your consideration.

ui/apps/platform/src/Containers/NetworkGraph/components/StyleFakeGroup.tsx
alwayshooin: looks like there aren't :/

ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
alwayshooin: hm i'm not sure what the benefit of using `useMemo` is here since the function itself already uses the `useMemo` hook. maybe if we run into performance issues in the future we can look into this? i can add it as a comment for now so we don't forget

pedrottimark: Instead of `as` here in `return` statement how about `<CustomGroupNodeData>` useMemo above at line 68?

PR #4947, opened 2023-02-20
Removed ManagedFields at the informer level to save on memory

connorgorman:
## Description

Informers have a SetTransformFunc that mutates an object before it hits the cache. Generically, we can set the managed fields entries to nil

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


lvalerom (approved):

sensor/kubernetes/listener/resource_event_handler.go
```diff
@@ -294,6 +307,11 @@ func handle(
 		missingInitialIDs:          nil,
 	}
 	informer.AddEventHandler(handlerImpl)
+	if !informer.HasSynced() {
+		if err := informer.SetTransform(managedFieldsTransformer); err != nil {
+			utils.Should(err)
```
janisz, line 312: There is no need to guard `utils.Should` with condition. If err is nil it will do nothing.

PR #4952, opened 2023-02-20
Fix lint warning about no explicit any in clusters ProviderMetadata type

pedrottimark:
## Description

1. Import type from frontend proto file.
2. Replace property dot with `in` operator, which TypeScript prefers for case analysis in this situation.
3. Add missing type intersection in frontend proto file.

## Checklist

## Testing Performed

1. `yarn lint` in ui
2. `yarn build` in ui
3. `yarn cypress-open` in ui/apps/platform
    * clusters/clustersHealthStatus.test.js

### DIFF
--- a/ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
+++ b/ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
@@ -12,6 +12,7 @@ import {
     TimesCircleIcon,
 } from '@patternfly/react-icons';
 
+import { ClusterProviderMetadata } from 'types/cluster.proto';
 import { getDate } from 'utils/dateUtils';
 import { CertExpiryStatus } from './clusterTypes';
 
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
```diff
@@ -302,26 +303,19 @@ export function formatBuildDate(orchestratorMetadata) {
...
-    google?: any;
-};
-
-export function formatCloudProvider(providerMetadata: ProviderMetadata) {
+export function formatCloudProvider(providerMetadata: ClusterProviderMetadata) {
     if (providerMetadata) {
         const { region } = providerMetadata;
 
-        if (providerMetadata.aws) {
+        if ('aws' in providerMetadata) {
```
vjwilson, line 310: Is this a TS thing, or just a preference?

vjwilson (approved):

ui/apps/platform/src/Containers/Clusters/cluster.helpers.tsx
pedrottimark: Good question. TypeScript did not distinguish the types with property dot but did with `in` operator.

TypeScript docs site is sadly not very searchable to find an example, let alone a guideline about it.

pedrottimark: By the way, the original `if` statements are my preference.
* TypeScript does understand that idiom for optional properties with `?` or `Whatever | null` types.
* My guess is that TypeScript prefers `in` operator as an explicit attempt to discriminate the type union.

PR #4954, opened 2023-02-21
ROX-10310 Syslog server

ksurabhi91:
Wrote a syslog server for testing syslog notifier for e2e tests.
Syslog server
1.listens on port 514 for syslog messages
2.listens on port 8080n for GET requests
3.server saves incoming syslog messages and returns them in LIFO order

## Checklist
- [ X] Investigated and inspected CI test results
- [ X] Unit test and regression tests added
- [ X] Evaluated and added CHANGELOG entry if required
- [X ] Determined and documented upgrade steps
- [X ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

## Testing Performed
Tested this server for following scenarios:
1.no message sent and send get request
2.send multiple messages and get them all in LIFO order


Makefile
```diff
@@ -649,6 +654,15 @@ webhookserver-image: webhookserver-build
 		-t quay.io/rhacs-eng/webhookserver:1.2 \
 		-f webhookserver/Dockerfile webhookserver
 
+syslog-image: syslog-build
+	-mkdir syslog/bin
+	cp bin/linux_$(GOARCH)/syslog syslog/bin/syslog
+	chmod +w syslog/bin/syslog
+	docker build \
+		-t stackrox/syslog:1.0 \
+		-t quay.io/rhacs-eng/syslog:1.0 \
```
theencee, line 663:
```suggestion
		-t quay.io/rhacs-eng/qa:syslog_server_1_0 \
```

We don't have a repository called `syslog` and instead of creating one I think we should use the `qa` one which is where all testing related stuff should go into. And we can use the tag to say this is syslog.

go.mod
```diff
@@ -367,6 +367,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
 
+require gopkg.in/mcuadros/go-syslog.v2 v2.3.0
```
theencee, line 370: I'm a bit worried that this might impact downstream (i.e. operator) build. But I wasn't able to validate it.

Can you try the following:
1. create a go.mod file in `qa-tests-backend/test-images/syslog`
2. Remove the changes here and an in go.sum
3. Use this in that new go.mod file:
```
module github.com/stackrox/stackrox/qa-tests-backend/test-images/syslog

go 1.18

require gopkg.in/mcuadros/go-syslog.v2 v2.3.0
```
4. cd to `qa-tests-backend/test-images/syslog` and run `go mod tidy`

theencee (approved):

PR #4955, opened 2023-02-21
ROX-15138: Change the links from NWG 1.0 to NWG 2.0

sachaudh:
## Description

This PR changes links from the old Network Graph to the new Network Graph. The following are areas where I found links to the Network Graph:

1. The "Deployments at most risk" widget in the Dashboard section
2. Under the "Risk Indicators" tab in the Risk section

If there are any other areas I might have missed, let me know 

https://user-images.githubusercontent.com/4805485/220225102-402cef2f-ba27-46e5-a32d-5fe8dbac27fc.mov

https://user-images.githubusercontent.com/4805485/220225130-371b5ba5-8226-4295-83f2-1f5f58cfb01a.mov

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/Dashboard/Widgets/DeploymentsAtMostRiskTable.tsx
+++ b/ui/apps/platform/src/Containers/Dashboard/Widgets/DeploymentsAtMostRiskTable.tsx
@@ -4,9 +4,10 @@ import { Truncate } from '@patternfly/react-core';
 import { TableComposable, Thead, Tr, Th, Tbody, Td } from '@patternfly/react-table';
 
 import { ListDeployment } from 'types/deployment.proto';
-import { networkBasePath, riskBasePath } from 'routePaths';
+import { networkBasePathPF, riskBasePath } from 'routePaths';
 import { SearchFilter } from 'types/search';
 import { getUrlQueryStringForSearchFilter } from 'utils/searchUtils';
+import { getQueryString } from 'utils/queryStringUtils';
 
 const columnNames = {
     deployment: 'Deployment',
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/Dashboard/Widgets/DeploymentsAtMostRiskTable.tsx
```diff
@@ -43,31 +44,38 @@ function DeploymentsAtMostRiskTable({
...
-                            width={20}
-                            className="pf-u-pr-0 pf-u-text-align-center-on-md"
-                            dataLabel={columnNames.riskPriority}
-                        >
-                            {priority}
-                        </Td>
-                    </Tr>
-                ))}
+                {deployments.map(({ id, name, cluster, namespace, priority }) => {
+                    const queryString = getQueryString({
```
sachaudh, line 48: What I don't really like about this approach is the fact that if Network Graph's URL architecture changes, constructing links this way would cause issues. I think a more secure way of creating links would be to have some sort of helper function or hook within the Network Graph directory that we could import into other sections and construct links that way. That way we wouldn't have to know details about how to structure these URLs. We could supply the values as parameters to the helper functions or hooks and it would spit out the links we need

sachaudh: Leaving a comment to come back to

vjwilson: I think a helper function makes more sense. I'd go ahead and abstract it out now.

sachaudh: Doing it in a follow-up draft PR makes more sense. I want to ensure everyone agrees on the approach (separately from this bug fix). Refactoring this bug fix should be trivial after

vjwilson (approved):

PR #4965, opened 2023-02-21
Disable frozen schema generation (v73)

rhybrillou:
## Description

Wrapping the generation comments in a bigger comment does not prevent go from executing the generation commands in the generation phase.
The syntax of the frozenschema v73 gen.go file is changed in order not to be picked up by go source generation.

## Checklist
- [ ] Investigated and inspected CI test results
~~- [ ] Unit test and regression tests added~~
~~- [ ] Evaluated and added CHANGELOG entry if required~~
~~- [ ] Determined and documented upgrade steps~~
~~- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

If any of these don't apply, please comment below.

## Testing Performed

Manual change of a proto file, run `make proto-generated-srcs` then `make go-generated-srcs` and make sure no schema file gets changed in the frozenschema/v73 directory.

PR #4968, opened 2023-02-21
Prefactor: enable search on API Tokens

rhybrillou:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


rhybrillou:
* master
    * **PR #4968** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4968" target="_blank">Graphite</a>  👈
      * **PR #4969** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4969" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4968?utm_source=stack-comment).

proto/storage/api_token.proto
```diff
@@ -13,10 +13,11 @@ message TokenMetadata {
     string name                          = 2;
     repeated string roles                = 7;
     google.protobuf.Timestamp issued_at  = 4;
-    google.protobuf.Timestamp expiration = 5;
-    bool revoked                         = 6;
+    google.protobuf.Timestamp expiration = 5 [(gogoproto.moretags) = 'search:"Expiration,store"'];
+    bool revoked                         = 6 [(gogoproto.moretags) = 'search:"Expiration,store"'];
```
dhaus67, line 17: I suppose you want to name it `Revoked` here instead of duplicated search tags.

proto/storage/api_token.proto
```diff
@@ -13,10 +13,11 @@ message TokenMetadata {
...
 
 
     string role = 3 [deprecated = true];
+    // Next available tag: 8
```
dhaus67, line 21: nit: we usually do this at the _start_ of the message, so please move it to l.10.

migrator/migrations/m_172_to_m_173_enable_search_on_api_tokens/oldapitokenpostgresstore/postgres_plugin.go
```diff
@@ -0,0 +1,303 @@
+package postgres
```
dhaus67, line 1: I was under the impression that these files are autogenerated, or are they not?

migrator/migrations/m_172_to_m_173_enable_search_on_api_tokens/migration.go
```diff
@@ -0,0 +1,65 @@
...
+
+	startSeqNum = 172
+)
+
+var (
+	migration = types.Migration{
+		StartingSeqNum: startSeqNum,
+		VersionAfter:   &storage.Version{SeqNum: int32(startSeqNum + 1)},
+		Run: func(database *types.Databases) error {
+			// Migration code comes here
```
dhaus67, line 28: I suppose the call to `migrateAPITokens` is missing here?

dhaus67 (changes requested):

migrator/migrations/m_172_to_m_173_enable_search_on_api_tokens/oldapitokenpostgresstore/postgres_plugin.go
rhybrillou: Not really. The way I did the first postgres -> postgres migration was by copying over the pieces of the store file I needed, and pruning the code I did not need.
In the meantime, the code is supposed to rely on frozen schema data, which in my opinion does make sense. But the code generator support for migration is limited, and does not seem to cover the store part anymore.

proto/storage/api_token.proto
rhybrillou: I'd rather not have this comment be the type description.
I'll move it to the start of the message, in the message body, separated from the fields by a blank line (an alternative is to not add it in the first place, as the message is not that big yet).

rhybrillou: Screenshot 2023-03-01 at 22 47 01
Moving the log to l.10 would result in that entry in the swagger file, which I find a bit misleading.

migrator/migrations/m_172_to_m_173_enable_search_on_api_tokens/oldapitokenpostgresstore/postgres_plugin.go
dhaus67: Ah I understand, was under the initial impression that this was autogenerated.

Do you think it makes sense to add a comment header calling out that the code was copied from the datastore and all unused functions were removed? I know we had that etiquette in the rocksdb migrations, I'd say it makes sense to continue with it here as well?

Makefile
```diff
@@ -481,7 +481,7 @@ go-postgres-unit-tests: build-prep test-prep
...
+		$(shell git ls-files -- '*postgres/*_test.go' '*postgres_test.go' '*datastore_sac_test.go' '*clone_test.go' 'migrator/migrations/n_*/migration_test.go' 'migrator/migrations/m_16?_*/migration_test.go' 'migrator/migrations/m_17?_*/migration_test.go' '*pruning_test.go' '*reprocessor_test.go' '*enricher_impl_test.go' '*v2/parts_test.go' '*version/ensure_test.go' '*version/store/store_impl_test.go' '*activecomponent/updater/updater_impl_test.go' '*role/validate_test.go' '*search/service/service_impl_test.go' '*deployment/service/service_impl_test.go' '*metadata/service/service_impl_test.go' '*systeminfo/listener/listener_test.go' | sed -e 's@^@./@g' | xargs -n 1 dirname | sort | uniq | xargs go list| grep -v '^github.com/stackrox/rox/tests$$' | grep -Ev $(UNIT_TEST_IGNORE)) \
```
dhaus67, line 484: Don't we want to include all `m_*` here, essentially `migrator/migrations/m_*/migration_test.go` or we removed these for a reason? I couldn't find anything in the blame on this

dhaus67 (approved):

Makefile
dashrews78: most the `m_*` current are rocks only.  we would have to update all those to not run in Postgres mode.  unintended consequence of our naming scheme.

central/apitoken/datastore/datastore_impl.go
```diff
@@ -106,3 +136,34 @@ func (b *datastoreImpl) RevokeToken(ctx context.Context, id string) (bool, error
...
+}
+
+func (b *datastoreImpl) SearchRawTokens(ctx context.Context, q *v1.Query) ([]*storage.TokenMetadata, error) {
+	if err := sac.VerifyAuthzOK(integrationSAC.ReadAllowed(ctx)); err != nil {
+		return nil, err
+	}
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return b.storage.GetByQuery(ctx, q)
+	}
+	results, err := b.searcher.Search(ctx, q)
```
dashrews78, line 154: what happens here?  I think `b.searcher` is going to be nil here as it is in Rocks mode and you set the searcher to nil at line 51.

central/apitoken/datastore/datastore_impl.go
```diff
@@ -4,24 +4,54 @@ import (
...
+	storage  store.Store
+	searcher search.Searcher
 
 	sync.Mutex
 }
 
+func newPostgres(pool *postgres.DB) *datastoreImpl {
+	storage := postgresStore.New(pool)
+	indexer := postgresStore.NewIndexer(pool)
+	searcher := blevesearch.WrapUnsafeSearcherAsSearcher(indexer)
```
dashrews78, line 39: curious why we are. using bleve in postgres mode.

migrator/migrations/m_172_to_m_173_enable_search_on_api_tokens/migration.go
```diff
@@ -0,0 +1,64 @@
...
+	migrations.MustRegisterMigration(migration)
+}
+
+// Additional code to support the migration
+
+func migrateAPITokens(postgresDB *postgres.DB, gormDB *gorm.DB) error {
+	ctx := context.Background()
+	oldStore := oldAPITokenStore.New(postgresDB)
+	newStore := newAPITokenStore.New(postgresDB)
+	pgutils.CreateTableFromModel(ctx, gormDB, frozenSchema.CreateTableAPITokensStmt)
```
dashrews78, line 43: nit: logistically I think we should create the table before we create create the new store.  Doesn't really matter since you aren't using the store until after the table is created though.

Makefile
rhybrillou: https://issues.redhat.com/browse/ROX-15801

central/apitoken/datastore/datastore_impl.go
rhybrillou: This `blevesearch` wrapper is a utility from the search package to convert bleve searchers (typically the generated index code) to standard searcher. The difference between the two is the `...blevesearch.SearchOption` parameter in the function signature in the former.

central/apitoken/datastore/datastore_impl.go
rhybrillou: It's likely to panic.
Although the chances that this code gets used in rocks/bleve mode are low, I'll make sure an error is returned when not in postgres mode.

central/apitoken/datastore/datastore_impl.go
dashrews78: fair enough.  If we are going to continue to use that we should probably refactor it at some point.  outside of scope here but worth mentioning.

dashrews78 (approved):

pkg/migrations/internal/seq_num.go
```diff
@@ -4,7 +4,7 @@ var (
 	// CurrentDBVersionSeqNum is the current DB version number.
 	// This must be incremented every time we write a migration.
 	// It is a shared constant between central and the migrator binary.
-	CurrentDBVersionSeqNum = 172
+	CurrentDBVersionSeqNum = 173
```
dashrews78, line 7: 173 was just pushed.

dashrews78 (changes requested):
173 was pushed so you are going to have to do some renaming, unfortunately.

pkg/migrations/internal/seq_num.go
rhybrillou: Re-aligned migration versions, leaving space for https://github.com/stackrox/stackrox/pull/5142

dashrews78 (approved):

PR #4969, opened 2023-02-21
ROX-13519: Create Notification loop for API Tokens about to expire

rhybrillou:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


rhybrillou:
* master
    * **PR #4968** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4968" target="_blank">Graphite</a> 
      * **PR #4969** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4969" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/4969?utm_source=stack-comment).

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/stackrox/rox/pkg/protoconv"
+	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stackrox/rox/pkg/search"
+)
+
+const (
+	notificationInterval = 1 * time.Hour      // 1 hour
+	staleNotificationAge = 24 * time.Hour     // 1 day
+	expirationWindow     = 7 * 24 * time.Hour // 1 week
+	expirationSlice      = 24 * time.Hour     // 1 day
```
janisz, line 24: https://github.com/stackrox/stackrox/blob/1ec040affd21d0927804ae2bebe0dc4304aa18af/pkg/timeutil/date.go#L10-L20

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+	n.stopper.Client().Stop()
+	_ = n.stopper.Client().Stopped().Wait()
+}
+
+func (n *expirationNotifierImpl) runExpirationNotifier() {
+	defer n.stopper.Flow().ReportStopped()
+
+	n.checkAndNotifyExpirations()
+
+	t := time.NewTicker(notificationInterval)
```
janisz, line 81:
```suggestion
	t := time.NewTicker(notificationInterval)
	defer t.Stop()
```

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// ExpiringItemNotifier is the interface to list the tokens about to expire, and to send notifications
+// for items about to expire.
+type ExpiringItemNotifier interface {
+	ListItemsAboutToExpire() ([]search.Result, error)
+	Notify(identifiersToNotify []string) error
+}
+
+type expirationNotifierImpl struct {
```
janisz, line 54: I think we should also inject a "clock" so we will be able to test it with mocked time.

central/apitoken/expiration/expiration_test.go
```diff
@@ -0,0 +1,134 @@
...
+	token := generateToken(&now, &expiration, false)
+	s.Require().NoError(s.datastore.AddToken(ctx, token))
+
+	fetchedTokens, err := s.notifier.listItemsToNotify(now, expiresUntil)
+	s.NoError(err)
+	expectedResults := []*storage.TokenMetadata{}
+	s.ElementsMatch(expectedResults, fetchedTokens)
+}
+
+func (s *apiTokenExpirationNotifierTestSuite) TestDontSelectRevokedToken() {
```
janisz, line 108:
```
=== RUN   TestAPITokenExpirationNotifier/TestDontSelectRevokedToken
    expiration_test.go:119: 
        	Error Trace:	/home/janisz/go/src/github.com/stackrox/stackrox/central/apitoken/expiration/expiration_test.go:119
        	Error:      	elements differ
        	            	
        	            	extra elements in list B:
        	            	([]interface {}) (len=1) {
        	            	 (*storage.TokenMetadata)({
        	            	  Id: (string) (len=36) "08cb277d-33b4-470c-a1f4-9624fc149936",
        	            	  Name: (string) (len=20) "Generated Test Token",
        	            	  Roles: ([]string) (len=1) {
        	            	   (string) (len=5) "Admin"
        	            	  },
        	            	  IssuedAt: (*types.Timestamp)({
        	            	   Seconds: (int64) 1677679569,
        	            	   Nanos: (int32) 702390000,
        	            	   XXX_NoUnkeyedLiteral: (struct {}) {
        	            	   },
        	            	   XXX_unrecognized: ([]uint8) <nil>,
        	            	   XXX_sizecache: (int32) 0
        	            	  }),
        	            	  Expiration: (*types.Timestamp)({
        	            	   Seconds: (int64) 1677686769,
        	            	   Nanos: (int32) 702390000,
        	            	   XXX_NoUnkeyedLiteral: (struct {}) {
        	            	   },
        	            	   XXX_unrecognized: ([]uint8) <nil>,
        	            	   XXX_sizecache: (int32) 0
        	            	  }),
        	            	  Revoked: (bool) true,
        	            	  Role: (string) "",
        	            	  XXX_NoUnkeyedLiteral: (struct {}) {
        	            	  },
        	            	  XXX_unrecognized: ([]uint8) <nil>,
        	            	  XXX_sizecache: (int32) 0
        	            	 })
        	            	}
        	            	
        	            	
        	            	listA:
        	            	([]*storage.TokenMetadata) {
        	            	}
        	            	
        	            	
        	            	listB:
        	            	([]*storage.TokenMetadata) (len=1) {
        	            	 (*storage.TokenMetadata)({
        	            	  Id: (string) (len=36) "08cb277d-33b4-470c-a1f4-9624fc149936",
        	            	  Name: (string) (len=20) "Generated Test Token",
        	            	  Roles: ([]string) (len=1) {
        	            	   (string) (len=5) "Admin"
        	            	  },
        	            	  IssuedAt: (*types.Timestamp)({
        	            	   Seconds: (int64) 1677679569,
        	            	   Nanos: (int32) 702390000,
        	            	   XXX_NoUnkeyedLiteral: (struct {}) {
        	            	   },
        	            	   XXX_unrecognized: ([]uint8) <nil>,
        	            	   XXX_sizecache: (int32) 0
        	            	  }),
        	            	  Expiration: (*types.Timestamp)({
        	            	   Seconds: (int64) 1677686769,
        	            	   Nanos: (int32) 702390000,
        	            	   XXX_NoUnkeyedLiteral: (struct {}) {
        	            	   },
        	            	   XXX_unrecognized: ([]uint8) <nil>,
        	            	   XXX_sizecache: (int32) 0
        	            	  }),
        	            	  Revoked: (bool) true,
        	            	  Role: (string) "",
        	            	  XXX_NoUnkeyedLiteral: (struct {}) {
        	            	  },
        	            	  XXX_unrecognized: ([]uint8) <nil>,
        	            	  XXX_sizecache: (int32) 0
        	            	 })
        	            	}
        	Test:       	TestAPITokenExpirationNotifier/TestDontSelectRevokedToken
    --- FAIL: TestAPITokenExpirationNotifier/TestDontSelectRevokedToken (2.17s)
```

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		return
+	}
+
+	now := time.Now()
+	aboutToExpireDate := now.Add(expirationWindow)
+	staleNotificationDate := now.Add(-staleNotificationAge)
+	staleNotificationTimestamp := protoconv.ConvertTimeToTimestamp(staleNotificationDate)
+
+	scheduleCtx := sac.WithGlobalAccessScopeChecker(context.Background(),
```
janisz, line 103: move to const?

central/apitoken/datastore/datastore_impl.go
```diff
@@ -106,3 +142,54 @@ func (b *datastoreImpl) RevokeToken(ctx context.Context, id string) (bool, error
 	}
 	return true, nil
 }
+
+func (b *datastoreImpl) Search(ctx context.Context, q *v1.Query) ([]search.Result, error) {
+	if ok, err := integrationSAC.ReadAllowed(ctx); err != nil {
+		return nil, err
+	} else if !ok {
+		return nil, nil
+	}
```
dhaus67, line 151: Do we want to use `sac.VerifyAuthZOK(integrationSAC.ReadAllowed(ctx))` here instead? Looking at the references in the current datastore implementation, I don't see a reason why not to (but maybe I'm missing something?)

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+		sac.AllowFixedScopes(
+			sac.AccessModeScopeKeys(storage.Access_READ_ACCESS, storage.Access_READ_WRITE_ACCESS),
+			sac.ResourceScopeKeys(resources.Notifications),
+		),
+	)
+	var notificationSchedule *storage.NotificationSchedule
```
janisz, line 109:
```suggestion
```

migrator/migrations/m_173_to_m_174_create_notification_schedule_table/migration.go
```diff
@@ -0,0 +1,37 @@
...
+	"gorm.io/gorm"
+)
+
+var (
+	startSeqNum = 173
+	migration   = types.Migration{
+		StartingSeqNum: startSeqNum,
+		VersionAfter:   &storage.Version{SeqNum: int32(startSeqNum + 1)},
+		Run: func(database *types.Databases) error {
+			// Migration code comes here
```
dhaus67, line 21:
```suggestion

```

central/apitoken/expiration/singleton.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/sync"
+)
+
+var (
+	notifier Notifier
+	once     sync.Once
+)
+
+// Singleton ...
```
dhaus67, line 14: Please provide a proper comment for this.

central/apitoken/expiration/singleton.go
```diff
@@ -0,0 +1,23 @@
...
+func Singleton() Notifier {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		return nil
+	}
```
dhaus67, line 18: super-nit:
you could move this in the `once.Do` so subsequent calls will be a no-op and simply return nil.

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+		timeUnit := int(expirationSlice.Seconds())
+		ttlSeconds := int(ttl.Seconds())
+		sliceCount := ttlSeconds / timeUnit
+		if ttlSeconds%timeUnit != 0 {
+			sliceCount++
+		}
+		sliceDuration := "hours"
+		if sliceCount == 1 {
+			sliceDuration = "hour"
+		}
```
dhaus67, line 182: How about factoring this out into. separate function, let it return the "log message" so we also actually test the log message generation? Currently, it's not tested at all but AFAICT this is the "essence" of this PR: providing an accurate log statement.

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Warnf("API Token %s (ID %s) will expire in less than %d %s.", token.GetName(), token.GetId(), sliceCount, sliceDuration)
```
dhaus67, line 183: So, are there any plans to add other notification methods as well?

Right now, this is quite simplistic only for the API token one, but as mentioned within the initial design doc it might not be sufficient enough for ACSCS users, as they have limited access to the logs.

In light of that, it might be also good to abstract the `notify` mechanism more than just this simple method, and provide an interface `Notifier`  which provides:
```go
type Notifier interface {
  Notify(token *storage.TokenMetadata) error
}
```
and create a "log notifier" implementation here. This way when we decide to add more "notifiers" (e.g. email, audit logs) the code here doesn't need to be completely refactored.

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+	expirySearchCtx = sac.WithGlobalAccessScopeChecker(context.Background(),
+		sac.AllowFixedScopes(
+			sac.AccessModeScopeKeys(storage.Access_READ_ACCESS),
+			sac.ResourceScopeKeys(resources.Integration),
+		),
+	)
+)
+
+// Notifier is the interface for a background task that notifies about API token expiration
+type Notifier interface {
```
dhaus67, line 42: From other things I've seen, we either name those `...Loop` or `...Manager`.

I might settle for `TokenExpirationLoop` or something like this?

central/apitoken/expiration/expiration_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	fetchedTokens, err := s.notifier.listItemsToNotify(now, expiresUntil)
+	s.NoError(err)
+	expectedResult := token
+	expectedResults := []*storage.TokenMetadata{expectedResult}
+	s.ElementsMatch(expectedResults, fetchedTokens)
+}
+
+func (s *apiTokenExpirationNotifierTestSuite) TestDontSelectTokenNotAboutToExpire() {
+	ctx := sac.WithAllAccess(context.Background())
```
dhaus67, line 95: Might be worthwhile to make this a property of the suite and initialize it in `SetupSuite`.

dhaus67 (changes requested):

central/apitoken/expiration/expiration.go
rhybrillou: The pruning loop ticker does not follow that pattern. Might be worth fixing.
The reprocessor loops match the pattern, in a slightly strange way.

central/apitoken/expiration/expiration_test.go
rhybrillou: @dhaus67 spotted the source of this issue in https://github.com/stackrox/stackrox/pull/4968 (bad search field in proto annotation).
This should be fixed now.

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,186 @@
...
+	Start()
+	Stop()
+}
+
+// ExpiringItemNotifier is the interface to list the tokens about to expire, and to send notifications
+// for items about to expire.
+type ExpiringItemNotifier interface {
```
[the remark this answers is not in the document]
rhybrillou: I think this is a leftover from an attempt at making API Tokens part of a more generic notification framework.
I dropped the generic part to focus on the problem at hand.

This seems unused and will be removed for now.

proto/storage/notification_schedule.proto
```diff
@@ -0,0 +1,13 @@
...
+option go_package = "storage";
+option java_package = "io.stackrox.proto.storage";
+
+import "google/protobuf/timestamp.proto";
+import "gogoproto/gogo.proto";
+
+package storage;
+
+message NotificationSchedule {
+    google.protobuf.Timestamp last_run = 1;
```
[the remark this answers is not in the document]
rhybrillou: Moved to https://github.com/stackrox/stackrox/pull/5067

migrator/migrations/m_173_to_m_174_create_notification_schedule_table/migration.go
rhybrillou: Moved to https://github.com/stackrox/stackrox/pull/5067

migrator/migrations/m_173_to_m_174_create_notification_schedule_table/migration_test.go
```diff
@@ -0,0 +1,8 @@
+package m173tom174
+
+import (
+	"testing"
+)
+
+func TestMigration(t *testing.T) {
```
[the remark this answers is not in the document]
rhybrillou: Addressed in https://github.com/stackrox/stackrox/pull/5067

central/apitoken/expiration/expiration.go
rhybrillou: This depends on the inputs that come from the product side.
I do agree that in the longer run, there could be multiple notification channels, and that an interface for notification would then make sense. I would rather consider the bulk of tokens about to expire rather than individual tokens as input for a notifier (e.g. I'd rather send a single e-mail containing all tokens about to expire rather than an email for each token about to expire).

central/apitoken/expiration/expiration_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/pkg/postgres/pgtest"
+	"github.com/stackrox/rox/pkg/protoconv"
+	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stackrox/rox/pkg/uuid"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestAPITokenExpirationNotifier(t *testing.T) {
```
dhaus67, line 21: We should also cover generating the token log within the suite.

central/apitoken/expiration/expiration.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (n *expirationNotifierImpl) runExpirationNotifier() {
+	defer n.stopper.Flow().ReportStopped()
+
+	n.checkAndNotifyExpirations()
+
+	if n.notificationTicker == nil {
+		n.notificationTicker = time.NewTicker(env.APITokenExpirationNotificationInterval.DurationSetting())
+	}
```
dhaus67, line 94: Don't think we need that extra logic here, since we already ensured this within `Start` and this is an unexported method (within the tests, we could just set the ticker "manually" to a relatively low value).

dhaus67 (approved):
LGTM modulo one comment regarding initializing the ticker within `runExpirationNotifier`.

PR #4971, opened 2023-02-22
ROX-14526: Add exponential backoff to clarify's GetNodeVulnerabilities()

jvdm:
## Description

Add retries with exponential backoff when handling node scans and the Scanner is unavailable.
 
We want to retry this API call because the time for waiting for a new Node Inventory is currently high. Short availability issues on this API could cause customers to wait hours to see the latest Node Scanning components and vulnerabilities.

## Checklist

## Testing Performed

1. Deployed to Openshift-4.
2. Enabled `ROX_RHCOS_NODE_SCANNING=true` in Scanner, Central, Sensor, and Collector (in that order).
3. Observe Node Scanning results in Vulnerability Management.
4. Add a network policy to block ingress on Scanner (see below).
5. Trigger a new node scan in Compliance (by kicking the container: `kubectl -n stackrox exec collector-<id> compliance -- kill 1`.
6. Observe logs.

Logs observed:

```
sensor/service/pipeline/nodeinventory: 2023/02/22 22:04:38.322413 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-88z59-worker-c-fj9wg.c.srox-temp-dev-test.internal' ID='8a9d6ea9-c28d-4b05-ba01-12a804ba90e7'
pkg/scanners/clairify: 2023/02/22 22:04:38.337873 clairify.go:438: Warn: calling GetNodeVulnerabilities() (retrying in 512.055588ms): rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 172.30.111.175:8443: connect: connection refused"
pkg/scanners/clairify: 2023/02/22 22:04:38.851134 clairify.go:438: Warn: calling GetNodeVulnerabilities() (retrying in 435.140532ms): rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 172.30.111.175:8443: connect: connection refused"
pkg/scanners/clairify: 2023/02/22 22:04:39.286958 clairify.go:438: Warn: calling GetNodeVulnerabilities() (retrying in 1.321174225s): rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 172.30.111.175:8443: connect: connection refused"
pkg/scanners/clairify: 2023/02/22 22:04:40.613495 clairify.go:438: Warn: calling GetNodeVulnerabilities() (retrying in 939.716644ms): rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 172.30.111.175:8443: connect: connection refused"
pkg/scanners/clairify: 2023/02/22 22:04:41.554490 clairify.go:438: Warn: calling GetNodeVulnerabilities() (retrying in 2.172302528s): rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 172.30.111.175:8443: connect: connection refused"
sensor/service/pipeline/nodeinventory: 2023/02/22 22:04:48.331949 pipeline.go:86: Warn: enriching node with node inventory secured-cluster-jvdm-argilo:jvdm-argilo-88z59-worker-c-fj9wg.c.srox-temp-dev-test internal: error scanning node secured-cluster-jvdm-argilo:jvdm-argilo-88z59-worker-c-fj9wg.c.srox-tem -dev-test.internal error: Error scanning 'secured-cluster-jvdm-argilo:jvdm-argilo-88z59-worker-c-fj9w .c.srox-temp-dev-test.internal' with scanner "": rpc error: code = DeadlineExceeded desc = context de dline exceeded
```

Meaning: Calls to Scanner were failing to connect, and retries were triggered as expected. Context timed out after 10 seconds, and retries gave up, as expected.

Cluster for inspection: 

```
infractl get jvdm-argilo          
ID:          jvdm-argilo
Flavor:      openshift-4

Status:      READY
Created:     2h15m ago
URL:         https://console-openshift-console.apps.jvdm-argilo.openshift.infra.rox.systems
Lifespan:    2d remaining
```

PS.: The binaries were hotplugged in the cluster, so no image build.

Network Policy:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: deny-scanner
spec:
  podSelector:
    matchLabels:
      app: scanner
  policyTypes:
  - Egress
```


pkg/scanners/clairify/clairify.go
```diff
@@ -411,12 +421,38 @@ func (c *clairify) GetVulnerabilities(image *storage.Image, components *clairGRP
 	return convertImageToImageScan(image.GetMetadata(), resp.GetImage()), nil
 }
 
+func callGRPCWithRetryOnError(ctx context.Context, operation string, f func() error) error {
+	err := wait.ExponentialBackoffWithContext(ctx, scannerDefaultBackoff, func() (bool, error) {
```
msugakov, line 425: Certainly, we're Kubernetes product and integrating with `k8s.io/apimachinery` is not that bad.

However, all current usages of `k8s.io/apimachinery/pkg/util/wait` are in functionality that I believe has problems running test on GHA and needs to be revised.

However, that's not my main point. We already have https://github.com/stackrox/stackrox/blob/46a27ca4f17b15756ee6a8becb78e0da4139497e/pkg/retry/retry.go#L33-L33 and already use `github.com/cenkalti/backoff/v3`. Would either of those fit the use? It is better for maintainability is to have fewer patterns. In this case, I suggest not broaden the usage of `k8s.io/apimachinery/pkg/util/wait` and use either of two other more established patterns.

connorgorman (changes requested):
Doing it this way will have an unintended side effect that in our processing pipelines in Central, it will wait for this to complete, which means we won't get the stubs of any other node. So if you have 16 nodes that all hash into all 16 pipelines, but Scanner is not available, then they will not be handled until one of these either timeouts or completes.

I think that this could be better solved by just saying there is a scanning error and then re-enqueue them or rescanning them via some sort of reprocessing loop that runs periodically. Also assuming that the error received is transient and not some sort of bug (which will need to be distinguished)

pkg/scanners/clairify/clairify.go
jvdm: @msugakov:

>  all current usages of k8s.io/apimachinery/pkg/util/wait are in functionality that I believe has problems running test on GHA and needs to be revised.

Thanks. I was hoping to get pointers on that. I should have mentioned it in the description to make it clear, though.

> We already have pkg/retry/retry.go and use github.com/cenkalti/backoff/v3. Would either of those fit the use? 

I noticed `pkg/retry` when writing this, but [stackrox/pkg/retry/retry.go](https://github.com/stackrox/stackrox/blob/46a27ca4f17b15756ee6a8becb78e0da4139497e/pkg/retry/retry.go#L33-L33) doesn't support out-of-the-box jitter, has hardcoded retry intervals (100ms) and rate (x2), and more importantly can't tie the retries with the timeout context. And I intended to retry within the timeout limit. I did not find references to `cenkalti/backoff/v3` in that package.

I feel you suggest I could onboard cenkalti's `backoff.Backoff()` to `pkg/retry`, which I think is a good idea. But please, kindly confirm if that's indeed your original suggestion.

pkg/scanners/clairify/clairify.go
```diff
@@ -411,12 +421,38 @@ func (c *clairify) GetVulnerabilities(image *storage.Image, components *clairGRP
...
+		err := f()
+		if err != nil {
+			e, _ := status.FromError(err)
+			switch e.Code() {
+			case codes.Internal, codes.Aborted, codes.Unavailable, codes.Unknown:
```
connorgorman, line 430: Hmmm do you know when we get internal or unknown? Is this common, do we do proper error handling from scanner side?

jvdm:
@connorgorman:

> Doing it this way will have an unintended side effect that in our processing pipelines in Central, it will wait for this to complete, which means we won't get the stubs of any other node. So if you have 16 nodes that all hash into all 16 pipelines, but Scanner is not available, then they will not be handled until one of these either timeouts or completes.

Makes sense. The total time for retries is tied to the Timeout Context. I assumed the original timeout was already fine-tuned to accommodate an ongoing request—the [current timeout is 5 minutes](https://github.com/stackrox/stackrox/blob/f14a797ae5f9d166e632ac8b0ce809ca929c8284/pkg/scanners/clairify/clairify.go#L43).

Were you already considering the 5 minutes limit when in your comment? Please, let me know what you think of 5 minutes holding the queue in a worst-case scenario. Maybe we want to reduce the timeout to something shorter, considering Node Scans are faster than image scans?

> I think that this could be better solved by just saying there is a scanning error and then re-enqueue them or rescanning them via some sort of reprocessing loop that runs periodically.

This exponential retry focuses on capturing networking glitches or short (<5 min) availability problems in Scanner. We are concerned about keeping node inventories in memory for too long. It's probably OK to do it, given that it contains only OS components today. But long-term, that might not be true. Hence, we want to drop the request soon and optimistically send a "rescan" message to Sensor -> Compliance. We expect Compliance to send a new inventory (probably from its in-memory cache) randomly sometime in the future if the message is delivered, or a new node scan will appear in the next scan loop in Compliance (every 4hs).

> Also assuming that the error received is transient and not some sort of bug (which will need to be distinguished).

Yes, we want to fail fast if what caused the error is something we know is not recoverable. Please, look at the error codes I selected for retrying. If something there is off let me know so I can adjust.

connorgorman:
@jvdm I added a comment on some of the errors which are largely dependent on scanner actually doing the right thing so would be good to verify. Also, I think 5m is pretty long for the nodes honestly and is inherited from the image scan which actually pulls the layer. I think something like a 10s-30s timeout may be more reasonable. The number of node events should be fairly low

pkg/scanners/clairify/clairify.go
jvdm: I picking up these codes based on [their description](https://pkg.go.dev/google.golang.org/grpc/codes). 

`codes.Unavailable` and `codes.Aborted` are recommended for retries, see `FailedPrecondition` comment on the litmus test. I am in favor of keeping the first two.

`codes.Unknown` happens when the reply cannot be parsed properly, it's a catch all scenario and I agree we might not need to retry on "unknown".

Finally `codes.Internal` is what [scanner actually returns](https://github.com/stackrox/scanner/blob/10e708a2ca778683cee3ede3926cad3581a2ae53/api/v1/nodescan/service.go#L320). But it seems Scanner simply wrap all errors, so invalid input and things that we should not retry will be returned as `codes.Internal`. I vote we should remove it.

jvdm:
@connorgorman:

> @jvdm I added a comment on some of the errors which are largely dependent on scanner actually doing the right thing so would be good to verify. Also, I think 5m is pretty long for the nodes honestly and is inherited from the image scan which actually pulls the layer. I think something like a 10s-30s timeout may be more reasonable. The number of node events should be fairly low

Makes sense. I will update the timeout to 10s and retry 2-3 times within that time. I will also capture only `codes.Unavailable` since that is the only code that can be generated by the gRPC framework and does not depend on Scanner, which today wraps everything under `codes.Internal` (retriable and non-retriable).

@msugakov:

I will look into using `cenkalti/backoff` instead of adding a dep on k8s codebase.

pkg/scanners/clairify/clairify.go
```diff
@@ -411,12 +420,44 @@ func (c *clairify) GetVulnerabilities(image *storage.Image, components *clairGRP
...
+			default:
+				return backoff.Permanent(err)
+			}
+		}
+		return err
+	}
+	notify := func(err error, duration time.Duration) {
+		log.Warnf("calling %s() (retrying in %s): %v", name, duration, err)
+	}
+	eb := backoff.NewExponentialBackOff()
```
msugakov, line 447: Do you want to tweak defaults of this new exponential backoff? Defaults seem to allow 60 seconds of max delay and 15 minutes of max elapsed time. Neither looks too shiny in spite of the current 10 seconds limit.

jvdm: My intention was to use the defaults. In particular, the max elapsed time is irrelevant since we are using `*WithContext()`. Should we give up earlier than 10 seconds? That might be reasonable, specially if Scanner is out:  We would always try for 10s, where we could give up earlier. I can't see a problem with 10s at the moment. Clarify if you see any, please.

pkg/scanners/clairify/clairify.go
```diff
@@ -411,12 +420,44 @@ func (c *clairify) GetVulnerabilities(image *storage.Image, components *clairGRP
...
+	op := func() error {
+		err := f()
+		if err != nil {
+			e, _ := status.FromError(err)
+			switch e.Code() {
+			// Unavailable indicates the service is currently unavailable. The error code is
+			// set by the gRPC framework during failed connections, abrupt shutdown of a
+			// server process or network connection.
+			//
+			// Aborted is not be generated by the gRPC framework, but may be used by server
```
vikin91, line 433:
```suggestion
			// Aborted is not generated by the gRPC framework, but may be used by server
```

pkg/scanners/clairify/clairify.go
```diff
@@ -30,14 +31,22 @@ import (
...
 	// TypeString is the name of the Clairify scanner.
 	TypeString = "clairify"
 
-	clientTimeout             = 5 * time.Minute
+	// defaultClientTimeout default timeout for scanner calls.
+	defaultClientTimeout = 5 * time.Minute
+
+	// nodeScanClientTimeout is used for node scanning operations, it is shorter than
+	// the default because we expect it to have only the database on its path.
+	nodeScanClientTimeout = 10 * time.Second
```
vikin91, line 48: We decreased the max duration of this op from 5m to 10s - this is great!
But isn't 10s too little to include the first call and potential retries?

pkg/scanners/clairify/clairify.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/cenkalti/backoff/v3"
```
vikin91, line 11: Why not v4? https://pkg.go.dev/github.com/cenkalti/backoff/v4

pkg/scanners/clairify/clairify.go
```diff
@@ -411,12 +420,44 @@ func (c *clairify) GetVulnerabilities(image *storage.Image, components *clairGRP
...
+		log.Warnf("calling %s() (retrying in %s): %v", name, duration, err)
+	}
+	eb := backoff.NewExponentialBackOff()
+	return backoff.RetryNotify(op, backoff.WithContext(eb, ctx), notify)
+}
+
 func (c *clairify) GetNodeInventoryScan(node *storage.Node, inv *storage.NodeInventory) (*storage.NodeScan, error) {
 	req := convertNodeToVulnRequest(node, inv)
-	ctx, cancel := context.WithTimeout(context.Background(), clientTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), nodeScanClientTimeout)
```
vikin91, line 453: The timeout counter starts ticking in this line, right? From here, we got only 10s to make the call and potentially retry. I an unfortunate situation, the first call can take 10s and there will be no retries. 

Wouldn't it be better, to:
- Initialize the ctx with timeout of 10s only for a single call `c.nodeScanServiceClient.GetNodeVulnerabilities(ctx, req)` and
- configure the retry mechanism manually to make, e.g., max 3 attempts (or leave defaults and set the absolute max to 30s)?

vikin91 (commented):
The code looks good, but I have some questions in the comments about setting a deadline for all communication attempts using the context.

connorgorman (commented):
Looks reasonable to me now, feel free to dismiss my changes requested

pkg/scanners/clairify/clairify.go
jvdm: @vikin91:

> The timeout counter starts ticking in this line, right?

Yes.

> From here, we got only 10s to make the call and potentiall retries. I an unfortunate situation, the first call can take 10s and there will be no retries.

That's correct.

> Wouldn't it be better, to:

Potentially. It depends on which direction you want to improve.

I will explain why I chose that. Typically, if Scanner is timing out, something in its way is taking too long and is overloaded. In that scenario retrying multiple times might help increase pressure in Scanner. I experimented with Openshift-4, and `GetNodeInventoryScan` typically takes < 4s. Of course, we need more data to adjust the correct timeout. But ~2x seems reasonable to acknowledge Scanner is too busy, and we will release resources and try another time (later, like in 4hs if we don't send a retry message from Central to Compliance).

pkg/scanners/clairify/clairify.go
jvdm: Maybe. Are you concerned with retrying calls that can take more than or close to 10s? If yes, I will address this concern in [your other comment](https://github.com/stackrox/stackrox/pull/4971#discussion_r1119083090). If not, please clarify how 10s can be too little for the first call and retries.

pkg/scanners/clairify/clairify.go
jvdm: I take back that giving up earlier might be possible. Typical calls to Scanner takes 4s according to experiments in an Openshift cluster, so shorter timeouts might increase the chances of us giving up calls that would succeed.

pkg/scanners/clairify/clairify.go
vikin91: Okay, I see that your intention was to squeeze the original call and 1-3 retries within the 10s window. This feels appropriate for our purposes, but does not let the exponential backoff to demonstrate its powers, as there will be no exponential behavior demonstrated for 1-3. 
This is by no means a blocker, I wanted to understand the behavior and design decisions.

vikin91 (approved):

PR #4979, opened 2023-02-22
[VM2.0][ROX-13857]Add graphQL resolver for CVE first discovered in system

md2119:
## Description

GraphQL resolver for workload cve list page. Columns covered in this PR:
- First Discovered in System

Also adjusted the sample image scan that I had added in the earlier PR.

Refer to [API contract](https://docs.google.com/document/d/1c1I9m8cwhdRm3NUwo0D_TxYwAAuIsA-4UMt0VabZg6I/edit#heading=h.wjq0yqnqu4hl) for mocks.

## Checklist

## Testing Performed
Unit


central/views/imagecve/view_test.go
```diff
@@ -78,6 +80,23 @@ func TestGetImageCVECore(t *testing.T) {
...
 
+	// Ensure that the image is stored and constructed as expected.
+	for idx, image := range images {
+		actual, found, err := store.GetImage(ctx, image.GetId())
+		require.NoError(t, err)
+		require.True(t, found)
+
+		cloned := actual.Clone()
+		// Adjust dynamic fields.
+		standardizeImages(image)
```
charmik-redhat, line 91: nit : Can call standardizeImages once and pass both images to it. `standardizeImages(image, cloned)`

central/views/imagecve/view_test.go
```diff
@@ -78,6 +80,23 @@ func TestGetImageCVECore(t *testing.T) {
...
+		standardizeImages(cloned)
+		assert.EqualValues(t, image, cloned)
```
charmik-redhat, line 93: Why do we need this check here? Except for the standardized fields, could the db store image differently than what was passed ?

md2119: We want to make sure that the test is set up as expected. This is just an extra check.

central/views/imagecve/view_test.go
md2119: I refactored this entire test in the next PR which also has your nit addressed.

charmik-redhat (approved):
LGTM

PR #4980, opened 2023-02-22
ROX-12433: Fix layout in MainPage to scroll to end of page when there are banners

pedrottimark:
### Problem

Users cannot scroll to end of page when there are banners.

Serious problem because **Save** button is usually at the end of form or wizard pages.

### Analysis

1. PatternFly `MainPage` style rules include `height: 100%` and its parent `div` does have flex properties to correctly render optional public configuraton header at the top and footer at the bottom of the body (and root `div` element).

2. Banner elements are siblings of `MainPage` in the same `div` element, therefore their height is not deducted from available `MainPage` height.

### Solution

1. Move style from Tailwind classes to rules in src/css/style.css file.

2. Replace unneeded `AppWrapper` level of markup with fragment:
    * Login/LoginPage.js
    * Login/TestLoginResultsPage.js
    * MainPage/MainPage.tsx

3. Rearrange sibling elements in more meaningful order and adjust colors.
    * `Notifications` invisible element.
    * `PublicConfigHeader` has **user** background color.
    * `AnnouncementBanner` has info **blue** background color.
    * `CredentialExpiryBanner` has warning **yellow** background color.
    * `OutdatedVersionBanner` has warning **yellow** background color.
    * `DatabaseStatusBanner` has danger **red** background color.
    * `ServerStatusBanner` has danger **red** background color for UNREACHABLE or success **green** background color for RESURRECTED status.

4. Lift up banner elements so `div` element is parent of `MainPage` only child for its `height: 100%` style rule.

5. Delete unneeded `isSticky` prop from `Banner` elements, because root element provides flex column layout.

### Residue

1. Update login components.
    * Rewrite in TypeScript.
    * Replace simulated style to `LoginPage` PatternFly element.
    * Replace this last occurrence of redux-form with formik.

2. Unify toast notifications.
    * Dispatch actions to notifications reducer.
    * Render PatternFly `AlertGroup` element.

## Checklist

## Testing Performed

Temporarily edit code to render several banners.

### Original cannot scroll to end of page

You cannot see **Next** and **Back** buttons of Policy Wizard.

Even public configuration footer is not visible.

Stop server to see UNREACHABLE status which has warning **yellow** color.
cannot_scroll_UNREACHABLE

Restart server to see RESURRECTED status which also has warning **yellow** color.
cannot_scroll_RESURRECTED

Subtle difference was hard for me to notice even when I expected it.

### Improved can scroll to end of page

You can see **Next** and **Back** buttons of Policy Wizard.

Public configuration footer is visible.

Stop server to see UNREACHABLE status which has danger **red** color.
can_scroll_UNREACHABLE

Restart server to see RESURRECTED status which has success **green** color.

can_scroll-RESURRECTED

Obvious difference is easy for me to notice when the status changes (and meaning is clearer).


ui/apps/platform/src/Containers/Login/TestLoginResultsPage.js
```diff
@@ -93,7 +92,7 @@ function TestLoginResultsPage({ authProviderTestResults }) {
     const { messageBody, messageType } = getMessage(authProviderTestResults);
 
     return (
-        <AppWrapper>
+        <>
```
pedrottimark, line 95: Because `AppWrapper` element had only one child, fragment is redundant but intended as a minimal safe change because this page is very unfamiliar to me.

vjwilson: This page is only used in Access section to check rules, I seem to remember.

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
```diff
@@ -63,22 +62,22 @@ function MainPage(): ReactElement {
...
+                component="CENTRAL"
+                hasServiceIdentityWritePermission={hasServiceIdentityWritePermission}
+            />
+            <CredentialExpiryBanner
+                component="SCANNER"
+                hasServiceIdentityWritePermission={hasServiceIdentityWritePermission}
+            />
+            <OutdatedVersionBanner />
+            <DatabaseStatusBanner />
+            <ServerStatusBanner />
```
vjwilson, line 79: Just out of curiosity, why did you move the ServerStatusBanner to the end of the list?

vjwilson (approved):

ui/apps/platform/src/Containers/Login/TestLoginResultsPage.js
pedrottimark: **Test login** button on Auth provider form?

ui/apps/platform/src/Containers/MainPage/MainPage.tsx
pedrottimark: Good eyes.

Notice in the baseline pair of pictures how the yellow transient status banner gets lost among the certificate expiry banners which might remain for hours or days?

The banners are in order of increasing timeliness of the information.

ui/apps/platform/src/Containers/Login/TestLoginResultsPage.js
vjwilson: Yest that's the one.

PR #4982, opened 2023-02-22
ROX-14436 route handler sends deploymentReference to the resolver

lvalerom:
## Description

The route handler sends related deploymentReferences to the resolver.

* ~Depends on: #5250~

## Checklist

## Testing Performed

* Manual tests in an OCP cluster
* e2e tests running in CI
* e2e tests will be enhanced in #5250

### DIFF
--- a/sensor/kubernetes/listener/resources/route.go
+++ b/sensor/kubernetes/listener/resources/route.go
@@ -3,6 +3,8 @@ package resources
 import (
 	routeV1 "github.com/openshift/api/route/v1"
 	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/pkg/env"
+	"github.com/stackrox/rox/sensor/common/store/resolver"
 	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component"
 )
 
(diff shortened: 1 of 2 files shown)


sensor/kubernetes/listener/resources/route.go
```diff
@@ -42,6 +44,11 @@ func (r *routeDispatcher) ProcessEvent(obj, _ interface{}, action central.Resour
 	if existingService == nil {
 		return nil
 	}
+	if env.ResyncDisabled.BooleanSetting() {
+		event := component.NewEvent()
+		event.AddDeploymentReference(resolver.ResolveDeploymentLabels(existingService.GetNamespace(), existingService.selector), central.ResourceAction_UPDATE_RESOURCE, false)
+		return event
```
fredrb, line 50: Could you just add a comment here explaining that when the re-sync is disabled, we don't want to call `UpdateExposuresForMatchingDeployments` here and instead reprocess all the matching deployments. And also that we only send the event with the deployment reference because services **don't send any resource events** to central. It looks a bit different from the other handlers because of that. 

When I first read the code I forgot that services were not sent, and it would be good to have the context. Probably the comment can be removed once the old path is also removed.

fredrb (approved):

sensor/kubernetes/listener/resources/route.go
lvalerom: Done!

PR #4987, opened 2023-02-22
ROX-15109: Replace tippy with PatternFly tooltip

vjwilson:
## Description

This is the first phase of converting all the product's tooltips to PatternFly versions, even if the sections they appear in are not in PatternFly yet.
The impetus for this is to drop our dependence on the Tippy-JS library, which is would need to be upgraded to unblock us from upgrading several other packages. (I discovered this problem while researching upgrades related to https://issues.redhat.com/browse/ROX-15166)

This phase replaces usage of our old StackRox ui-components `Tooltip` component, which wrapped Tippy.

In one case, that of `ClusteredEventsTooltip` in the process timeline, the PatternFly team recommended that I use their `Popover` component instead of their tooltip, because of the frequent need to overflow and scroll its content. See https://patternfly.slack.com/archives/C4FM977N0/p1677005692935099

Note: I have not yet tackled our direct use of Tippy in graphs and charts. That will be solved in part when we remove the old Network Graph code for release 4.1.0. Then, any remaining direct use of Tippy can be migrated.

## Checklist

## Testing Performed

Manually tested each tooltip as I migrated it.

tooltip_timeline_legend
tooltip_event_list
tooltip_cve_pill
tooltip_numbered_list
tooltip_download_yaml
tooltip_config_secrets_widget
tooltip_compliance_labels
tooltip_sensor_status
tooltip_operator_indicator
tooltip_helm_indicator
tooltip_credential_expiration
tooltip_top_cvss
tooltip_drilldown_button
tooltip_row_action_menu
tooltip_row_action_button
tooltip_table_action_menu
tooltip_table_action_button
tooltip_panel_header
tooltip_table_header
tooltip_process_item
popover_clusteredevents

### DIFF
--- a/ui/apps/platform/cypress/constants/RiskPage.js
+++ b/ui/apps/platform/cypress/constants/RiskPage.js
@@ -118,7 +118,7 @@ export const selectors = {
     eventTimelineOverviewButton: 'button[data-testid="event-timeline-overview"]',
     tooltip: {
         ...tooltipSelectors,
-        legendContents: `${tooltipSelectors.overlay} > div`,
+        legendContents: `${tooltipSelectors.overlay} .pf-c-tooltip__content`,
         legendContent: {
             event: eventSelectors,
         },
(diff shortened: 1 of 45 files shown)


dvail (approved):
Sorry, this one slipped away from me.

This was more straightforward than the "Files changes (45)" suggested.

PR #4999, opened 2023-02-23
Clean up Postgres error logs

connorgorman:
## Description

I think that the error wraps should highlight where the issue is coming from and also errors could be more regular due to Postgres restarts so don't completely log them. Thoughts?

## Checklist

## Testing Performed

Straightforward


pkg/search/postgres/common.go
```diff
@@ -1115,7 +1083,6 @@ func RunDeleteRequestForSchema(ctx context.Context, schema *walker.Schema, q *v1
 	}
 
 	return pgutils.Retry(func() error {
-
 		_, err = db.Exec(ctx, query.AsSQL(), query.Data...)
 		if err != nil {
 			return errors.Wrapf(err, "could not delete from %q", schema.Table)
```
dashrews78, line 1088: query string may be of use here.

pkg/search/postgres/common.go
```diff
@@ -886,19 +874,9 @@ func retryableRunSelectRequestForSchema[T any](ctx context.Context, db *postgres
...
-			return nil, err
-		}
-		if !pgutils.IsTransientError(err) {
-			debug.PrintStack()
-		} else {
-			log.Debugf("%s", debug.Stack())
-		}
-		log.Errorf("Query issue: %s %+v: %v", queryStr, redactedQueryData(query), err)
-		return nil, err
+		return nil, errors.Wrapf(err, "error executing query")
```
dashrews78, line 877: log the query?

dashrews78 (approved):
LGTM

PR #5005, opened 2023-02-23
Add queue length metric for process indicators

connorgorman:
## Description

Adds a metric that tracks how many indicators have yet to be flushed. This is not available in sensor_event_queue because of the batching

## Checklist

## Testing Performed

Will run this on GKE once it builds and should prometheus metric


dashrews78 (approved):
LGTM

PR #5006, opened 2023-02-23
[WIP] ROX-14660: Resolve the overwriting problem

vikin91:
## Description

The idea here is to pull up a bool flag to control whether the NodeScan data should ignored or treated the way that it used to be.
The goal is to force-ignore the updates to NodeScan & Risk when the Node is RHCOS and it arrives to the Node Pipeline.

The major open TODO is to handle the `serialized` field properly when upserting.

## Checklist

## Testing Performed

- None yet

### DIFF
--- a/central/node/datastore/store/postgres/store.go
+++ b/central/node/datastore/store/postgres/store.go
@@ -61,7 +61,7 @@ type Store interface {
 	Count(ctx context.Context) (int, error)
 	Exists(ctx context.Context, id string) (bool, error)
 	Get(ctx context.Context, id string) (*storage.Node, bool, error)
-	Upsert(ctx context.Context, obj *storage.Node) error
+	Upsert(ctx context.Context, node *storage.Node, ignoreScan bool) error
 	Delete(ctx context.Context, id string) error
 	GetMany(ctx context.Context, ids []string) ([]*storage.Node, []int, error)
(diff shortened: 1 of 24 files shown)


central/node/datastore/store/postgres/store.go
```diff
@@ -102,29 +102,48 @@ func (s *storeImpl) insertIntoNodes(
...
+			cloned.GetName(),
+			pgutils.NilOrUUID(cloned.GetClusterId()),
+			cloned.GetClusterName(),
+			cloned.GetLabels(),
+			cloned.GetAnnotations(),
+			pgutils.NilOrTime(cloned.GetJoinedAt()),
+			cloned.GetContainerRuntime().GetVersion(),
+			cloned.GetOsImage(),
+			pgutils.NilOrTime(cloned.GetLastUpdated()),
+			serialized, // TODO: when scan is ignored, we should update it only selectively!
```
jvdm, line 118: Any ideas on how we could update `serialized` selectively?

vikin91: Leverage the existing mechanism maybe?

- Fetch node obj from DB in `isUpdated`
- Clone the node object that is provided as the argument to `upsert`
- Pick what to keep from the clone and what to take from the the arg object
- Serialize the clone
- Upsert

vikin91:
Closing in favor of the #5071

PR #4948, opened 2023-02-20
Update PatternFly to 2023.01

vjwilson:
## Description

Release notes: https://www.patternfly.org/v4/developer-resources/release-notes/#2023.01-release-notes-2023-02-02

Note: There was one breaking change that affected our app

> Search input: ... Reordered parameters in onChange and onSearch callbacks to add event as first parameter. 

I updated our usage of that component to account for the new order of parameters.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

Manual, exploratory testing, and checking the SearchInputs that were specifically modified to support the revised PatternFly parameter order.

PR #5010, opened 2023-02-23
Revert "Update PatternFly to 2023.01"

vjwilson:
Reverts stackrox/stackrox#4948

Ran into issues:
1. checkboxes in table rows no longer check-able
2. errors in browser console about too many renders on pages with tables with many rows


sachaudh (approved):
LGTM

pedrottimark:
Let’s hold this until we confirm whether another recent change is the cause.

vjwilson:
Reverting would not fix the bug, like we thought it might.

PR #5011, opened 2023-02-23
Dashrews/gke postgres upgrade test collector

dashrews78:
## Description

The nature of this tests involves lots of upgrades and restarts.  So pods are coming up and down fairly regularly by design.  So when some of those restarts are slow we can get into situations where something like sensor takes too long to come back up so collector may have an unallowed restart because sensor took too long to come back up.  For this test, things like that are not really an issue.  So for this test only we will allow such restarts of collector.

This solves the issues where the tests gets flagged for collector having a previous.  Not the issue where the test cannot get metadata if it gets rotated to a 1.24 cluster.

## Checklist

## Testing Performed

All about this test.

### DIFF
--- a/tests/upgrade/postgres_run.sh
+++ b/tests/upgrade/postgres_run.sh
@@ -133,12 +133,14 @@ test_upgrade_paths() {
     info "Bouncing central"
     kubectl -n stackrox delete po "$(kubectl -n stackrox get po -l app=central -o=jsonpath='{.items[0].metadata.name}')" --grace-period=0
     wait_for_api
-    # Bounce collectors to avoid restarts on if central is down long enough so sensor restarts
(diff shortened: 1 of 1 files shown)


c-du:
Would you like to have some explanations for the changes?

dashrews78:
> Would you like to have some explanations for the changes?

What happens is if central and/or sensor are down long enough during the upgrades, collector will bounce because it cannot communicate with sensor.  Then collector will be left with a previous and this is a way to clear it out.  That said, I created this PR real fast because I saw 2 failures quickly after pushing the updates to the test in master.  I was waiting to see if we really needed it which is why I don't have a description for it yet.

dashrews78:
Closing this.  Incorporated these changes in #5229

PR #5013, opened 2023-02-23
ROX-15341: Fix for default function prop causing useEffect loop

bradr5:
## Description

Due to defining a default function within the React custom hook, the reference keeps changing and triggers the useEffect, resulting in an infinite loop. Defining the function outside of the hook fixes this issue. The TableModal.tsx file has a similar problem with a passed-in function. However, this issue can be fixed by using the useCallback hook.

## Checklist

## Testing Performed

Manually tested violations and policy management tables

### DIFF
--- a/ui/apps/platform/cypress/integration/vulnmanagement/entitypages.test.js
+++ b/ui/apps/platform/cypress/integration/vulnmanagement/entitypages.test.js
@@ -330,7 +330,6 @@ describe('Entities single views', () => {
             cy.get(`${selectors.tableRows}`).eq(1).click();
         }, entitiesKey);
 
-        cy.get('button:contains("Fixable CVEs")').click();
         cy.wait('@getFixableCvesForEntity');
         cy.get(`${selectors.sidePanel} ${selectors.tableRows}:contains("CVE-2021-20231")`).contains(
             'Active'
--- a/ui/apps/platform/src/Containers/Policies/Wizard/Step3/TableModal.tsx
+++ b/ui/apps/platform/src/Containers/Policies/Wizard/Step3/TableModal.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import {
     Button,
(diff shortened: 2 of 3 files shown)


vjwilson (approved):

pedrottimark (approved):

pedrottimark:
* entitypages test does not have obvious relationship
* policyWizardStep3 test does seem related because it has `preSelectedFunc` so not depending on default value

> Policy wizard, Step 3 Policy Criteria Policy field card table modal should populate table modal select and respect changed values on save

policyWizardStep3

It looks like the preselection behavior might not have worked in the policy criterion.

@bradr5 Better to merge to fix the majority calls of hook and sort out the unique call on Monday.

bradr5:
> * entitypages test does not have obvious relationship
> * policyWizardStep3 test does seem related because it has `preSelectedFunc` so not depending on default value
> 
> > Policy wizard, Step 3 Policy Criteria Policy field card table modal should populate table modal select and respect changed values on save
> 
> policyWizardStep3
> 
> It looks like the preselection behavior might not have worked in the policy criterion.
> 
> @bradr5 Better to merge to fix the majority calls of hook and sort out the unique call on Monday.

So it looks like there is actually an infinite loop occurring again on this page. I'm not entirely sure why the change seemed to fix our issue in other places, but not here. I think our best bet is to just revert the change entirely for now to get tests passing and then revisit this next week.

ui/apps/platform/cypress/integration/vulnmanagement/entitypages.test.js
```diff
@@ -330,7 +330,6 @@ describe('Entities single views', () => {
             cy.get(`${selectors.tableRows}`).eq(1).click();
         }, entitiesKey);
 
-        cy.get('button:contains("Fixable CVEs")').click();
```
bradr5, line 333: This is related to removing policies from the deployments page. Will need to go back and re check tests related to policies in vulnerability management, just wanted to get tests back to a passing state with this merge.

PR #5017, opened 2023-02-24
ROX-15024: Workload CVEs Overview - Default filters modal

alwayshooin:
image
image
image

* The button at the top right corner of the page should be a button
* The button badge should reflect the number of filters set globally
* Clicking on the button should open the Default filters modal
* The modal should look like the above and populate as empty at first
* Clicking the checkboxes and cancelling should not update local storage
* Clicking the checkboxes and hitting the 'Apply filters' button should save the filters to local storage
* Opening the modal on refresh should load the filters saved from local storage

Implementation notes: 
* I saved the empty state to be ```preferences: {
        defaultFilters: {
            Severity: [],
            Fixable: [],
        },
    },``` because those fields will always exist (vs possibly being empty) and thus it didn't make much sense to omit them in the empty state and end up adding conditionals in the code

### DIFF
--- a/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/types.ts
+++ b/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/types.ts
@@ -1,0 +1,13 @@
+export type Severity = 'Critical' | 'Important' | 'Moderate' | 'Low';
+export type FixableStatus = 'Fixable' | 'Not fixable';
+
+export type DefaultFilters = {
+    Severity: Severity[];
+    Fixable: FixableStatus[];
+};
+
+export type VulnMgmtLocalStorage = {
+    preferences: {
+        defaultFilters: DefaultFilters;
+    };
+};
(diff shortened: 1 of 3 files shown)


ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/types.ts
```diff
@@ -0,0 +1,13 @@
+export type Severity = 'Critical' | 'Important' | 'Moderate' | 'Low';
```
dvail, line 1: What do you think about using the `VulnerabilitySeverity` type from `types/cve.proto.ts` as the value here instead? It seems like that is the canonical type for CVE severities on the front end, and the values match what is expected in the search API according to the design doc.

There also looks to be value to display mappings in `messages/common.ts`.

dvail (approved):
One thought on the source of truth for "severity" values but otherwise looks great. 👍

dvail:
Also, it looks like the e2e test failures are false positives but the style check is due to not having the component props typed.

PR #5018, opened 2023-02-24
ROX-14693: reconcile(create/update operation) declarative resources

ivan-degtiarenko:
## Description

Within this PR we create/update declarative resources found within the declarative configuration during reconciliation.
Deletion of declarative resources will be handled in a separate ticket.

The approach taken within this PR is to reconcile each resource individually. Reconciliation is done in order from referenced resources to referring resources(for example, access scope is referenced by role, so access scopes are reconciled first). That allows to avoid resolving references as it is guaranteed that in a valid configuration referenced resources will be reconciled first. 
It is possible that some state of configuration will not be valid(for example, if referenced and referring resources are located in different files), but important thing is that the configuration state in the backend will *eventually* catch on to changes within the files.

**A few other decisions made in this PR:**
1. The alternative to writing `reconcile*` functions in specific order was to introduce `protoTypeUpsertOrder` list of proto types. However, the problem was to generalize the type of reconcile function because of generic limitations in Golang.
2. `ReconciliationErrorReporter` interface decided to be made similar to `integrationHealthReporter`(every resource health and errors are treated separately). However, this is not principal part of this PR and can be changed later
3.  For auth providers `loginURL` field is filled within `registry.CreateProvider` call, `enabled` and `active` is set by default to true during transformation. There is no special handling of `basic` auth providers since declarative auth provider can't be of `basic` type.
4. Every datastore(except auth provider) got `Upsert` function. An alternative to supporting this function would be to execute `Get*` and then conditionally `Update*` or `Add*` datastore methods. There are two reasons not to follow the alternative approach: saving extra `Get*` call and not delegating datastore responsibilities to the datastore client.

## Checklist

## Testing Performed

**Testing plan(TBE - to be executed)**:
1. Manual test(configure declarative config - see if it's being reconciled)
2. Datastore level tests

Manually testing declarative configurations:
1. Create the helm chart:
```bash
roxctl helm output central-services --image-defaults=opensource
```
2. Define the following `values.yaml` file:
```yaml
central:
  declarativeConfiguration:
    mounts:
      configMaps:
        - permission-set
        - access-scope
        - auth-provider
        - role
customize:
  envVars:
    ROX_DECLARATIVE_CONFIGURATION: true
```
3. Install the central services chart:
```bash
helm install stackrox-central-services -n stackrox -f ./values.yaml ./stackrox-central-services-chart
```
4. Verify that central is running and the `ROX_DECLARATIVE_CONFIGURATION` variable is set. The mount points for declarative configuration should exist as well.
5. Create the first config map containing a permission set:
```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: permission-set
  namespace: stackrox
data:
  permission-set: |
    name: hello-world
    description: the first declaratively created permission set
    resources:
    - resource: Integration
      access: READ_ACCESS
    - resource: Administration
      access: READ_ACCESS
    - resource: Access
      access: READ_ACCESS
EOF
```
6. Verify within the a) logs of central that no error occurred during reconciliation b) within the UI / API the permission set `hello-world` exists.
7. Create the second config map containing an access scope:
```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: access-scope
  namespace: stackrox
data:
  access-scope: |
    name: hello-world
    description: the first declaratively created access scope
    rules:
      included:
      - cluster: remote
EOF
```
8. Verify within the a) logs of central that no error occurred during reconciliation b) within the UI / API the access scope `hello-world` exists.
9. Create the third config map containing a role:
```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: role
  namespace: stackrox
data:
  role: |
    name: hello-world
    description: the first declaratively created role
    permissionSet: hello-world
    accessScope: hello-world
EOF
```
10. Verify within the a) logs of central that no error occurred during reconciliation b) within the UI / API the role`hello-world` exists.
11. Create the fourth config map containing an auth provider:
```bash
apiVersion: v1
kind: ConfigMap
metadata:
  name: auth-provider
  namespace: stackrox
data:
  auth-provider: |
    name: hello-world
    minimumRole: "None"
    uiEndpoint: localhost:8000
    groups:
    - key: "email"
      value: "<redacted>"
      role: "Admin"
    oidc:
      issuer: <redacted>
      mode: fragment
      clientID: <redacted>

EOF
```
12. Verify within the a) logs of central that no error occurred during reconciliation b) within the UI / API the auth provider`hello-world` exists and you are able to successfully login, and group mapping works.


central/declarativeconfig/manager_impl.go
```diff
@@ -151,14 +190,97 @@ func (m *managerImpl) runReconciliation() {
 }
 
 func (m *managerImpl) reconcileTransformedMessages(transformedMessagesByHandler map[string]protoMessagesByType) {
```
dhaus67, line 192: We should definitely add tests for upserting.

central/declarativeconfig/manager_impl.go
```diff
@@ -23,6 +30,11 @@ const (
 
 type protoMessagesByType = map[reflect.Type][]proto.Message
 
+// ReconciliationErrorReporter processes declarative resources reconciliation errors.
+type ReconciliationErrorReporter interface {
+	ProcessError(protoValue proto.Message, err error)
+}
+
```
dhaus67, line 37: Albeit its just a WIP and to be changed - let's move this definition to `singleton.go` for now.

central/group/datastore/datastore_impl.go
```diff
@@ -27,6 +27,57 @@ type dataStoreImpl struct {
...
+	if err := ds.validateAndPrepGroupForUpsertNoLock(ctx, group); err != nil {
+		return err
+	}
+
+	return ds.storage.Upsert(ctx, group)
+}
+
+// Validate if the group is allowed to be upserted and prep the group before it is upserted in db.
+// NOTE: This function assumes that the call to this function is already behind a lock.
+func (ds *dataStoreImpl) validateAndPrepGroupForUpsertNoLock(ctx context.Context, newGroup *storage.Group) error {
```
dhaus67, line 50: nit: move the function to the helpers function in the section below.

central/group/datastore/datastore_impl.go
```diff
@@ -27,6 +27,57 @@ type dataStoreImpl struct {
...
+	if err := ValidateGroup(newGroup, true); err != nil {
```
dhaus67, line 51: Shouldn't we only require a ID _iff_ the group didn't exist beforehand? Or do we want to proclaim that for `Upsert`, the ID must be set at all times?

central/group/datastore/datastore_impl.go
```diff
@@ -27,6 +27,57 @@ type dataStoreImpl struct {
...
+		return errox.InvalidArgs.CausedBy(err)
+	}
+
+	oldGroup, exists, err := ds.storage.Get(ctx, newGroup.GetProps().GetId())
+	if err != nil {
+		return err
+	}
+	if exists {
+		if err = verifyGroupOriginMatches(ctx, oldGroup); err != nil {
```
dhaus67, line 60: nit: just reading through this, the name was a bit confusing at first.
Maybe we should get rid of the helper functions and simply have `declarativeconfig.CanModifyResource()` return an `errox.NotAuthorized` with appropriate error message, similar to what we have done with `sac.VerifyAuthZOK`?
This way we have the appropriate HTTP status code the same across datastore and it doesn't require each datastore to redefine the errox type.

central/role/datastore/datastore_impl.go
```diff
@@ -32,6 +32,117 @@ type dataStoreImpl struct {
...
+	permissionSet, accessScope, err := ds.verifyRoleReferencesExist(ctx, newRole)
+	if err != nil {
+		return err
+	}
+	if err := verifyPermissionSetOriginMatches(ctx, permissionSet); err != nil {
+		return err
+	}
+	if err := verifyAccessScopeOriginMatches(ctx, accessScope); err != nil {
+		return err
+	}
```
dhaus67, line 68: So in case someone would use the `UpsertRole` for this datastore, they could only do it in case the role doesn't reference a declaratively created access scope / permission set? Don't think we want to make this constraint just now.

central/role/datastore/datastore_impl.go
```diff
@@ -32,6 +32,117 @@ type dataStoreImpl struct {
...
+	if err != nil {
+		return err
+	}
+	if exists {
+		if err := verifyAccessScopeOriginMatches(ctx, oldScope); err != nil {
+			return err
+		}
+		if err := verifyAccessScopeOriginMatches(ctx, newScope); err != nil {
+			return err
+		}
```
dhaus67, line 133: We should check the access scope's origin also when its a new one - which we currently wouldn't do.

central/role/datastore/datastore_impl.go
```diff
@@ -32,6 +32,117 @@ type dataStoreImpl struct {
...
+		return err
```
dhaus67, line 134: Pretty sure we don't want to return here just yet.

dhaus67 (changes requested):

central/declarativeconfig/manager_impl.go
```diff
@@ -151,14 +190,97 @@ func (m *managerImpl) runReconciliation() {
...
-				protoType.Name(), handler, protoMessages)
+			transformedMessages[protoType] = append(transformedMessages[protoType], protoMessages...)
 		}
+
 	}
+	m.reconcileUpsertAccessScopes(transformedMessages)
+	m.reconcileUpsertPermissionSets(transformedMessages)
+	m.reconcileUpsertRoles(transformedMessages)
+	m.reconcileUpsertAuthProviders(transformedMessages)
+	m.reconcileUpsertGroups(transformedMessages)
```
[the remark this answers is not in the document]
ivan-degtiarenko: I think it's relatively easy to hardcode or parameterize order of reconciliation - for now I decided to go with the first approach by introducing list of types in order of reconciliation.

I think what's good about this approach is that by avoiding extra errors we shorten the time for all declarative resources to be reconciled(because if we fail to create a role on the first run, it will only show up in the system on the second run)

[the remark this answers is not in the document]
ivan-degtiarenko: That's not so useful considering 3/5 of these functions have the same underlying datastore which has the same lock for access scopes, permission sets and roles

central/declarativeconfig/manager_impl.go
ivan-degtiarenko: Do you mean datastore tests?

central/group/datastore/datastore_impl.go
ivan-degtiarenko: I think that depends on whether we want to treat `Upsert` as the function that can be used outside the declarative context or not.

If it were to be used outside the declarative context, then we shouldn't require ID. If we only use this function within declarative context, we know for sure group will have deterministic ID set

central/role/datastore/datastore_impl.go
ivan-degtiarenko: That's a very bad Ctrl-C Ctrl-V 😞

central/role/datastore/datastore_impl.go
ivan-degtiarenko: If I understand correctly, we want to allow imperative reference declarative, but not other way around.
Do you want to introduce this constraint within this PR, or move it to a separate ticket(this one, for example: https://issues.redhat.com/browse/ROX-14999)?

central/group/datastore/datastore_impl.go
ivan-degtiarenko: I think what's good about helper functions is that they allow to:
- specify the type of the resource
- specify the id of the resource
The second one is especially significant since we can't generalize it - the id for the group is accessed via `GetProps.GetId()` while for the rest of resources it's just `GetId`

Regarding helper function names: I agree they are far from ideal :) WDYT about just `verifyGroupOrigin` as an alternative?

central/declarativeconfig/updater/access_scope_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	return &AccessScopeUpdater{
+		roleDS: datastore,
+	}
+}
+
+// Upsert upserts message into the system(datastore, registry if resource has one).
+func (u *AccessScopeUpdater) Upsert(ctx context.Context, m proto.Message) error {
+	accessScope, ok := m.(*storage.SimpleAccessScope)
+	if !ok {
+		return errox.InvariantViolation.Newf("Wrong type passed to access scope updater: %T", accessScope)
```
dhaus67, line 30:
```suggestion
		return errox.InvariantViolation.Newf("wrong type passed to access scope updater: %T", accessScope)
```

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -0,0 +1,44 @@
...
+		authProviderDS:       datastore,
+		authProviderRegistry: registry,
+	}
+}
+
+// Upsert upserts message into the system(datastore, registry if resource has one).
+func (u *AuthProviderUpdater) Upsert(ctx context.Context, m proto.Message) error {
+	authProvider, ok := m.(*storage.AuthProvider)
+	if !ok {
+		return errox.InvariantViolation.Newf("Wrong type passed to auth provider updater: %T", authProvider)
```
dhaus67, line 33:
```suggestion
		return errox.InvariantViolation.Newf("wrong type passed to auth provider updater: %T", authProvider)
```

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/gogo/protobuf/proto"
+	authProviderDatastore "github.com/stackrox/rox/central/authprovider/datastore"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/auth/authproviders"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+// AuthProviderUpdater handles updates of storage.AuthProvider within declarative config reconciliation routine.
+type AuthProviderUpdater struct {
```
dhaus67, line 14: no reason to export those, right?
```suggestion
type authProviderUpdater struct {
```

central/declarativeconfig/updater/access_scope_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/gogo/protobuf/proto"
+	roleDataStore "github.com/stackrox/rox/central/role/datastore"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+// AccessScopeUpdater handles updates of storage.SimpleAccessScope within declarative config reconciliation routine.
+type AccessScopeUpdater struct {
```
dhaus67, line 13: no reason to export those, right?
```suggestion
type accessScopeUpdater struct {
```

central/declarativeconfig/updater/group_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/gogo/protobuf/proto"
+	groupDataStore "github.com/stackrox/rox/central/group/datastore"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+// GroupUpdater handles updates of storage.Group within declarative config reconciliation routine.
+type GroupUpdater struct {
```
dhaus67, line 13: no reason to export those, right?
```suggestion
type groupUpdater struct {
```

central/declarativeconfig/updater/group_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	return &GroupUpdater{
+		groupDS: datastore,
+	}
+}
+
+// Upsert upserts message into the system(datastore, registry if resource has one).
+func (u *GroupUpdater) Upsert(ctx context.Context, m proto.Message) error {
+	group, ok := m.(*storage.Group)
+	if !ok {
+		return errox.InvariantViolation.Newf("Wrong type passed to group updater: %T", group)
```
dhaus67, line 30:
```suggestion
		return errox.InvariantViolation.Newf("wrong type passed to group updater: %T", group)
```

central/declarativeconfig/updater/permission_set_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/gogo/protobuf/proto"
+	roleDataStore "github.com/stackrox/rox/central/role/datastore"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+// PermissionSetUpdater handles updates of storage.PermissionSet within declarative config reconciliation routine.
+type PermissionSetUpdater struct {
```
dhaus67, line 13: no reason to export those, right?
```suggestion
type permissionSetUpdater struct {
```

central/declarativeconfig/updater/permission_set_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	return &PermissionSetUpdater{
+		roleDS: datastore,
+	}
+}
+
+// Upsert upserts message into the system(datastore, registry if resource has one).
+func (u *PermissionSetUpdater) Upsert(ctx context.Context, m proto.Message) error {
+	permissionSet, ok := m.(*storage.PermissionSet)
+	if !ok {
+		return errox.InvariantViolation.Newf("Wrong type passed to permission set updater: %T", permissionSet)
```
dhaus67, line 30:
```suggestion
		return errox.InvariantViolation.Newf("wrong type passed to permission set updater: %T", permissionSet)
```

central/declarativeconfig/updater/role_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/gogo/protobuf/proto"
+	roleDataStore "github.com/stackrox/rox/central/role/datastore"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+// RoleUpdater handles updates of storage.Role within declarative config reconciliation routine.
+type RoleUpdater struct {
```
dhaus67, line 13: no reason to export those, right?
```suggestion
type roleUpdater struct {
```

central/declarativeconfig/updater/role_updater.go
```diff
@@ -0,0 +1,33 @@
...
+	return &RoleUpdater{
+		roleDS: datastore,
+	}
+}
+
+// Upsert upserts message into the system(datastore, registry if resource has one).
+func (u *RoleUpdater) Upsert(ctx context.Context, m proto.Message) error {
+	role, ok := m.(*storage.Role)
+	if !ok {
+		return errox.InvariantViolation.Newf("Wrong type passed to role updater: %T", role)
```
dhaus67, line 30:
```suggestion
		return errox.InvariantViolation.Newf("wrong type passed to role updater: %T", role)
```

central/declarativeconfig/singleton.go
```diff
@@ -10,10 +18,30 @@ var (
...
+		instance = New(
+			env.DeclarativeConfigReconcileInterval.DurationSetting(),
+			env.DeclarativeConfigWatchInterval.DurationSetting(),
+			map[reflect.Type]updater.ResourceUpdater{
+				authProviderType:  updater.NewAuthProviderUpdater(authProviderDatastore.Singleton(), registry),
+				groupType:         updater.NewGroupUpdater(groupDataStore.Singleton()),
+				roleType:          updater.NewRoleUpdater(roleDatastore.Singleton()),
+				permissionSetType: updater.NewPermissionSetUpdater(roleDatastore.Singleton()),
+				accessScopeType:   updater.NewAccessScopeUpdater(roleDatastore.Singleton()),
+			},
```
dhaus67, line 42: How about moving this map initialization call to the `updater` package, in a `DefaultResourceUpdater()` function or something like that?

It's more of a nit, but right now the manager has imports it shouldn't necessarily require / know (talking about the datastore dependencies).

central/group/datastore/datastore_impl.go
```diff
@@ -27,6 +27,24 @@ type dataStoreImpl struct {
 	lock sync.RWMutex
 }
 
+func (ds *dataStoreImpl) Upsert(ctx context.Context, group *storage.Group) error {
+	if ok, err := accessSAC.WriteAllowed(ctx); err != nil {
+		return err
+	} else if !ok {
+		return sac.ErrResourceAccessDenied
+	}
```
dhaus67, line 35: Why not `sac.VerifyAuthZOK(accessSAC.WriteAllowed(ctx))`?

central/role/datastore/datastore_impl.go
dhaus67: I guess we can move it to the separate ticket, fine by me.

central/declarativeconfig/updater/updater.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"context"
+
+	"github.com/gogo/protobuf/proto"
+)
+
+// ResourceUpdater handles updates of proto resources within declarative config reconciliation routine.
+// Each ResourceUpdater is responsible for updates of specific proto type.
+// TODO(ROX-14694): Extend interface with methods necessary for resource deletion.
+type ResourceUpdater interface {
```
dhaus67, line 12: super-nit: I wonder whether just calling it `ResourceHandler` would be more clear, given that it will handle both create / delete, but 🤷 
I don't have too strong of an opinion to block on it, so feel free to ignore that one.

central/declarativeconfig/manager_impl.go
dhaus67: I guess both? It'd be nice to test the flow once, specifically a) the `reconcileTransformedMessages` one regarding transforming messages and calling the updaters (with mocked updaters), ensuring we handle errors correctly as well as call the expected updaters and b) yeah datastore tests as well (seems like for role, permission set, access scope we already have some, but would be great to have one for group as well).

dhaus67 (commented):
Minor comments as well as one comment regarding adding tests, otherwise LGTM 🚢 

Before merging this one, do we want to do some manual testing for this one?
While we are going to add automated once the E2E flow is done (so soon™) I think it'd be worthwhile testing the creation flow?

central/declarativeconfig/manager_impl.go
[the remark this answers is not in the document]
dhaus67: I agree it's not necessarily required for now.

I'm fine with leaving it for now given the limited number of resources, but in the future we might consider not depending on a specific order if the number of resources grows.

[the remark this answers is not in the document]
dhaus67: Yeah, as said more or less optional for the current time being, and only once we have a higher scale.

central/group/datastore/datastore_impl.go
dhaus67: I suppose there's nothing stopping anyone using the datastore's generic interface to use the `Upsert` function currently.

The cleanest solution would be IMO to make it generic, and also usable within the non-declarative context. I don't see why a generic interface has a specific function only usable in a specific context - if we make it part of the generic interface, it should serve a generic purpose.

central/group/datastore/datastore_impl.go
dhaus67: `verifyGroupOrigin` sounds good to me.

Yeah, I can agree with the type of the resource one, the `ID` one isn't really a reason for me since we could extend the `ResourceWithTraits` interface to include `GetId()`, and since we anyways pass `groups.GetProps()` to the `CanModifyResource()` function, it wouldn't really matter.

But, let's keep the helper for now.

central/declarativeconfig/manager_impl.go
```diff
@@ -151,11 +184,27 @@ func (m *managerImpl) runReconciliation() {
...
+
+	}
+	for _, protoType := range protoTypesOrder {
+		messages, hasMessages := transformedMessages[protoType]
+		if !hasMessages {
+			continue
+		}
+		typeUpdater, hasUpdater := m.updaters[protoType]
+		if !hasUpdater {
+			log.Fatalf("Manager does not have updater for type %v", protoType)
```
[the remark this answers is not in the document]
ivan-degtiarenko: Since the customer is not in control of which proto types can be YAML configurations be transformed to and which proto types can be updated, the lack of corresponding updater is an error of us writing incorrect code. I think it is basically an incorrect state of the system, and perhaps we should stop entire `manager_impl` if we see such a type. That's an argument against sending this error down the `ProcessError` pipe.

The argument for sending this error is that `System Health` currently is the only page where this error will be visible to customers. Perhaps the flow should be:
1. If we encounter a such error, we send it to `ProcessError` with the message `FATAL: Manager cannot process proto type {proto type}. Processing declarative configurations will stop now. Please contact support to investigate the issue`
2. After that we immediately stop `manager_impl` from running

central/declarativeconfig/updater/updater.go
ivan-degtiarenko: I guess what prevented me from going for `ResourceHandler` is the fact that we already have a concept of handlers in declarative configuration, as well as a field in `manager_impl` called `transformedMessagesByHandler`

dhaus67 (approved):
LGTM modulo the few comments for proper comments in `central/declarativeconfig/types/accepted_types.go` 🚢

PR #5026, opened 2023-02-27
Dashrews/ROX-15331 batch call to delete by query

dashrews78:
## Description

SQL only allows a parameter list. up to 65535 items due to constraints within SQL.  In the instance with this ticket there were a large number of process indicators deleted and when we sent that large list of ideas to the delete by query it failed.  In order to resolve this quickly I did 2 things:
1.  batch the calls to delete by query when process indicators are removed.
2. Updated the searcher parser to truncate arguments when there are 65535 arguments.

We may want a follow up ticket to update the search framework to handle and batch these scenarios within the framework itself.

## Checklist

## Testing Performed

Added a test to test the batching.

### DIFF
--- a/central/processindicator/datastore/datastore_impl.go
+++ b/central/processindicator/datastore/datastore_impl.go
@@ -152,26 +152,46 @@ func (ds *datastoreImpl) removeIndicators(ctx context.Context, ids []string) err
 		return err
 	}
 
-	if env.PostgresDatastoreEnabled.BooleanSetting() {
-		// Clean up correlated ProcessListeningOnPort objects. Probably could be
-		// done using a proper FK and CASCADE, but it's usually a thing you would
-		// not like to do automatically. search.ProcessID is not a PID, but UUID of
-		// the record in the table
-		deleteQuery := pkgSearch.NewQueryBuilder().
-			AddStrings(pkgSearch.ProcessID, ids...).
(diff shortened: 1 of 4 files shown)


erthalion:
Thanks for the fix.

> Updated the searcher parser to truncate arguments when there are 65535 arguments.

It sound like a good defense measure, but usually it's rather dangerous to have more than several thousands of parameters due to memory pressure (the exact value depends on the amount of available memory of course, the worst case result is OOM of a db backend). Would it make sense to do a warning if the number is over a certain threshold?

erthalion (approved):

central/processindicator/datastore/datastore_impl.go
```diff
@@ -153,26 +153,46 @@ func (ds *datastoreImpl) removeIndicators(ctx context.Context, ids []string) err
 		return err
 	}
 
+	if err := ds.indexer.DeleteProcessIndicators(ids); err != nil {
```
md2119, line 156: I am assuming you changed the order since postgres is turned on by default and this op is essentially a noop.

md2119 (approved):

central/processindicator/datastore/datastore_impl.go
dashrews78: exactly.

PR #5032, opened 2023-02-27
ROX-15345: remove 'Apply network policy' option from UI

sachaudh:
## Description

This PR removes the "Apply network policy" option as suggested by Boaz. See https://issues.redhat.com/browse/ROX-15345

Screenshot 2023-02-27 at 11 49 15 AM

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
@@ -28,7 +28,6 @@ import { Simulation } from './utils/getSimulation';
 import LegendContent from './components/LegendContent';
 
 import {
-    ApplyNetworkPolicyModification,
     NetworkPolicySimulator,
     SetNetworkPolicyModification,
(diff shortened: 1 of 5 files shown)


ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
```diff
@@ -77,7 +77,6 @@ const TopologyComponent = ({
     selectedNode,
     simulator,
     setNetworkPolicyModification,
-    applyNetworkPolicyModification,
```
pedrottimark, line 80: Although style-checks is failing for a different reason, it is possible or even required
* remove from `TopologyComponentProps` above similar to in the following file?
* therefore also last import and type definition itself from userNetworkPolicySimulator hook?

sachaudh: Should be good now

pedrottimark (approved):

PR #5039, opened 2023-02-28
ROX-15158, ROX-15360, ROX-15361: make jira issue check optional, collector released versions idempotent & fix publish-helm-charts

tommartensen:
## Description

* Make check for open Jira issues in ongoing release optional. This was helpful during 3.74.0 release, when there were known open issues for documentation while cutting release candidates
* `mark_collector_release` is now idempotent. Previously, when re-running the automation, it would push the same lines multiple times into the RELEASED_VERSIONS file. Later it could fail because the PR already existed. fyi @Molter73 
* `push_helm_charts` does not request review from the obsolete `release-artifact-publishers` group anymore. Reviewing and merging the Helm charts is task of the upstream release engineer and mentioned in the release tracker, therefore the Slack message is not required anymore as well.

## Checklist

## Testing Performed

CI is sufficient.

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -646,7 +646,7 @@ mark_collector_release() {
     # We need to make sure the file ends with a newline so as not to corrupt it when appending.
     [[ ! -f RELEASED_VERSIONS ]] || sed --in-place -e '$a'\\ RELEASED_VERSIONS
     if grep -q "${tag}" RELEASED_VERSIONS; then
-        echo "Skip RELEASED_VERSIONS file update, already up to date ..." >> "${GITHUB_STEP_SUMMARY}"
+        echo "Skip RELEASED_VERSIONS file change, already up to date ..." >> "${GITHUB_STEP_SUMMARY}"
     else
         echo "Update RELEASED_VERSIONS file ..." >> "${GITHUB_STEP_SUMMARY}"
         echo "${collector_version} ${tag}  # Rox release ${tag} by ${username} at $(date)" \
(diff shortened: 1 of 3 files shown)


scripts/ci/lib.sh
```diff
@@ -645,16 +645,27 @@ mark_collector_release() {
...
+            >>RELEASED_VERSIONS
+        gitbot add RELEASED_VERSIONS
+        gitbot commit -m "Automatic update of RELEASED_VERSIONS file for Rox release ${tag}"
+        gitbot push origin "${branch_name}"
+    fi
+
+    PRs=$(gh pr list -s open \
+    --head "${branch_name}" \
+    --json number \
+    --jq length)
```
gavin-stackrox, line 662: nit:
```suggestion
            --head "${branch_name}" \
            --json number \
            --jq length)
```

scripts/ci/lib.sh
```diff
@@ -645,16 +645,27 @@ mark_collector_release() {
...
-    gitbot add RELEASED_VERSIONS
-    gitbot commit -m "Automatic update of RELEASED_VERSIONS file for Rox release ${tag}"
-    gitbot push origin "${branch_name}"
-
-    echo "Create a PR for collector to add this release to its RELEASED_VERSIONS file" >> "${GITHUB_STEP_SUMMARY}"
-    gh pr create \
-        --title "Update RELEASED_VERSIONS for StackRox release ${tag}" \
-        --body "Add entry into the RELEASED_VERSIONS file" >> "${GITHUB_STEP_SUMMARY}"
+    if grep -q "${tag}" RELEASED_VERSIONS; then
+        echo "Skip RELEASED_VERSIONS file change, already up to date ..." >> "${GITHUB_STEP_SUMMARY}"
```
gavin-stackrox, line 649: It is a pity that this script is now GITHUB dependent. But it preceeds this PR so I'm only going to grumble here. echo() could be changed to something like ci_log() to append to GITHUB_STEP_SUMMARY in the GHA case and echo otherwise.

gavin-stackrox (approved):

PR #5041, opened 2023-02-28
Optimise usage of Go cache in GHA

janisz:
## Description

This pull request aims to improve the caching process for golangci-lint by separating it from the Go cache. Additionally, the keys used to access the Go cache have been modified to prioritize cache entries from the same job, even if they were created later. Previously, unrelated jobs were occasionally downloading unnecessary cache data, which resulted in "no disk space" errors. By implementing this new approach, we expect to make better use of cached entries since there are infrequent instances where the entire cache needs to be invalidated for the same job. 

- https://github.com/golangci/golangci-lint/discussions/1920

## Checklist

## Testing Performed

CI

### DIFF
--- a/.github/actions/cache-go-dependencies/action.yaml
+++ b/.github/actions/cache-go-dependencies/action.yaml
@@ -11,16 +11,12 @@ runs:
       shell: bash
 
     - name: Cache Go Dependencies
-      uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


gavin-stackrox (approved):

porridge (approved):
:see_no_evil:

PR #5043, opened 2023-02-28
ROX-15055 Add Workload CVE Image Single Page shell

dvail:
## Description

Adds the overview section and boilerplate for the Images Single page.

## Checklist

## Testing Performed

To test these pages, manually grab an image id from your local environment and paste it into the browser URL at `/main/vulnerabilities/workload-cves/images/<image_id>`

Test rendering when the initial request results in an error:
image

Test rendering of page when the request is loading:
image

Test rendering of page when the request is complete. The image name and label metadata should be displayed in the header:
image

Clicking on the "SHA" label successfully copies the image sha to the clipboard.
`sha256:864ce567c854b419d8877546814aa2fd3c440c6537dce869a4a40339add0015b`
image

If the image has no deployments, the "Active" label will switch to "Inactive" (simulated)
image


dvail:
* master
  * **PR #5043** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5043" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5043?utm_source=stack-comment).

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
```diff
@@ -1,9 +1,171 @@
...
+import {
+    imageDetailsQuery,
+    ImageDetailsResponse,
+    ImageDetailsVariables,
+} from './queries/imageDetails';
+import ImageSingleVulnerabilities from './ImageSingleVulnerabilities';
+import ImageSingleResources from './ImageSingleResources';
+import useDetailsTabParameter from './hooks/useDetailsTabParameter';
+
+const workloadCveOverviewImagePath = `${vulnerabilitiesWorkloadCvesPath}?cveStatusTab=Observed&entityTab=Image`;
```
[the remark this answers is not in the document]
alwayshooin: oh i started doing that in the modal PR, try looking up `./searchUtils.tsx` -- i mostly copied the logic that exists for the policy management query string manipulation but we can add more to this too

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
```diff
@@ -1,9 +1,205 @@
...
+            variables: { id: imageId },
+        }
+    );
+
+    const [activeTabKey, setActiveTabKey] = useDetailsTabParameter();
+
+    const imageData = data && data.image;
+    const imageName = imageData?.name?.fullName ?? 'NAME UNKNOWN';
+
+    const handleTabClick: TabsProps['onSelect'] = (e, tabKey) => {
```
alwayshooin, line 109: oh neat i wasn't aware we could do that to selectively pull props from the type defs of the library

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
```diff
@@ -1,9 +1,205 @@
...
+                            <ImageDetailBadges imageData={imageData} />
+                        </Flex>
+                    ) : (
+                        <Flex
+                            direction={{ default: 'column' }}
+                            spaceItems={{ default: 'spaceItemsXs' }}
+                            className="pf-u-w-50"
+                        >
+                            <Skeleton fontSize="2xl" />
+                            <Skeleton fontSize="sm" />
```
alwayshooin, line 149: ooh fancy. is this the way we're planning on handling loading sections now instead of spinner/loaders?

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
```diff
@@ -1,9 +1,205 @@
...
+                <Breadcrumb>
+                    <BreadcrumbItemLink to={workloadCveOverviewImagePath}>
+                        Images
+                    </BreadcrumbItemLink>
+                    {!error && (
+                        <BreadcrumbItem isActive>
+                            {imageData ? (
+                                imageName
+                            ) : (
+                                <Skeleton screenreaderText="Loading image name" width="200px" />
```
alwayshooin, line 193: should we have the `screenreaderText` for the `Skeleton` components above as well?

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/queries/imageDetails.ts
```diff
@@ -0,0 +1,50 @@
...
+export type ImageDetailsResponse = {
+    image: {
+        deploymentCount: number;
+        name: {
+            fullName: string;
+        } | null;
+        operatingSystem: string;
+        metadata: {
+            v1: {
+                created: Date | null;
```
alwayshooin, line 16: what's the difference between typing it this way vs ` created?: Date` ?

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/queries/imageDetails.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+            scan {
+                dataSource {
+                    name
+                }
+                scanTime
+            }
+        }
+    }
+`;
```
alwayshooin, line 50: it might be useful to have the queries closer to the components using the queries if we don't have a need to recycle them in other components. if we do need to reuse them, we can probably create a `queries` directory to hold them for easy discovery? 

either way, it feels odd to me to mix the type and the gql strings here :/

alwayshooin (approved):
code looks fine, just a few questions/suggestions!

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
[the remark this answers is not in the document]
dvail: Oh cool, I'll update this once I rebase then.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/queries/imageDetails.ts
dvail: `created?: Date` would allow either `{ created: undefined }` or `{}` in the response object, while GraphQL will explicitly add a `null` for nullable fields `{ created: null }`. In practice it probably doesn't matter a whole lot since we are checking for truthy values in the code, but I figured I'd keep it as accurate as possible.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/queries/imageDetails.ts
dvail: Yeah I still feel like finding a place for the queries and types for GraphQL is a little awkward, compared to our REST services which is pretty clear cut. I'll move the type and query right into the component for now and we can worry about common fragments later.

I'm also going to take a look at GraphQL type generation soon since it seems like we shouldn't have to manually create types for these queries everywhere.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/WorkloadCvesImageSinglePage.tsx
dvail: This was super useful during collections to avoid having to duplicate type signatures. Especially in situations where there was prop-drilling but it didn't really make sense to use a context; we could define the prop type of a component in terms of the prop type of a child component, and when things changed we only needed to change it in two places instead of in every intermediate component.

PR #5044, opened 2023-02-28
ROX-14702: Display origin for access scopes, roles, auth providers and groups

ivan-degtiarenko:
## Description

Following up steps and pattern of changes made here https://github.com/stackrox/stackrox/pull/4894, in this PR I add origin field for other resources in the UI: roles, access scopes, auth providers and groups.

As groups are tightly bound to auth provider and are only rendered on the auth provider page, we do not display groups' origin anywhere in the UI. The assumption is that groups and corresponding auth provider always will be of the same type - either both imperative or declarative. The only change made for groups is that groups with `Declarative` origin are impossible to modify or delete within the UI.

For every other resource next changes are introduced:
* `Origin` field is displayed when listing all resources of a certain type
*  Based on `Origin` field resources might be disabled for modification and deletion
*  `Origin` field is added to the resource page

Integration tests will be added in the follow-up PR.

## Checklist

## Testing Performed

Screenshot 2023-02-28 at 17 30 24
Screenshot 2023-02-28 at 17 30 36

### DIFF
--- a/ui/apps/platform/src/Containers/AccessControl/AccessScopes/AccessScopes.tsx
+++ b/ui/apps/platform/src/Containers/AccessControl/AccessScopes/AccessScopes.tsx
@@ -17,7 +17,6 @@ import {
     accessScopeNew,
     createAccessScope,
     deleteAccessScope,
-    getIsDefaultAccessScopeId,
     fetchAccessScopes,
     updateAccessScope,
 } from 'services/AccessScopesService';
@@ -36,6 +35,7 @@ import AccessControlBreadcrumbs from '../AccessControlBreadcrumbs';
 import AccessControlHeaderActionBar from '../AccessControlHeaderActionBar';
 import AccessControlNoPermission from '../AccessControlNoPermission';
 import usePermissions from '../../../hooks/usePermissions';
+import { isUserResource } from '../traits';
 
 const entityType = 'ACCESS_SCOPE';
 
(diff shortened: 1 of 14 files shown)


ui/apps/platform/src/Containers/AccessControl/AccessScopes/AccessScopes.tsx
```diff
@@ -17,7 +17,6 @@ import {
     accessScopeNew,
     createAccessScope,
     deleteAccessScope,
-    getIsDefaultAccessScopeId,
```
pedrottimark, line 20: For extra credit, can you please add a comment at the definition of `getIsDefaultAccessScopeId` function that the only remaining import is in ResourceScopeSelection.tsx file, which will be deleted when Collections (which depend on postgres) supersede Access Scopes in Vulnerability Reporting.

ui/apps/platform/src/Containers/AccessControl/Roles/Roles.tsx
```diff
@@ -13,7 +13,6 @@ import {
 } from '@patternfly/react-core';
 
 import NotFoundMessage from 'Components/NotFoundMessage';
-import { getIsDefaultRoleName } from 'constants/accessControl';
```
pedrottimark, line 16: Can you verify that this and the other pull request remove all references to `getIsDefaultRoleName` function, and therefore delete it and `defaultRoles` object from src/constants/accessControl.ts file?

pedrottimark (approved):
Very clear and consistent.

2 comments about almost-obsolete and apparently-obsolete helper functions.

PR #5045, opened 2023-02-28
Increase memory resource limit to 3Gi

yocum137:
Large clusters with lots of nodes, pods, network policies, etc. can cause Kubernetes to OOM kill the rhacs-operator during initial startup.  Increasing the memory Limit to 3Gi should alleviate this.

## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/operator/config/manager/manager.yaml
+++ b/operator/config/manager/manager.yaml
@@ -67,7 +67,7 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: 1Gi
+            memory: 3Gi
           requests:
             cpu: 100m
             memory: 200Mi


SimonBaeumer:
@yocum137 Thanks for rising this issue.
Could you provide more information on the scale of your cluster (rough estimates are sufficient)?

- How many nodes?
- How many pods are running?
- How many resources are deployed (deployments, network policies, ...)?

johannes94:
@yocum137 Some additional questions:

- What version of the ACS operator did you use?
- Do I understand correctly that setting the limit to 3GB solved the problem in your specific case?
- We've implemented a feature to dump memory profiles of the operator to debug OOM issues like this. It was first released with `v3.74.0`. Which was released a few days ago. Is it possible for you to upgrade to that version? (#3912) A heap dump would help us with figuring out the root cause of the unexpected memory usage.

yocum137:
@SimonBaeumer 129 nodes. 388K etcd objects. I'm discussing with Johannes internally, as well.

yocum137:
Customer confirms that 3GB is a suitable value to work-around the OOM kill, but ideally this should be configurable.

porridge:
After some deeper investigation @johannes94 found that this OOM was a result of an edge case at which the operator consumes unusual amount of memory. The recommended approach would be to adjust the resources manually in the deployment and not change the default value, which seems sufficient for vast majority of users. Further investigation on why we hit this OOM is tracked in the internal ticket https://issues.redhat.com/browse/ROX-12936

PR #5048, opened 2023-02-28
ROX-15088: Declarative config health

dhaus67:
## Description

This PR extends the existing integration health service available under `/v1/integrationhealth` to offer an additional endpoint for declarative configurations: `/v1/integrationhealth/declarativeconfigs`.

The health status will contain all relevant information about each declarative config and in case any error occurs, these errors will be surfaced within the integration health.

For now, I've opted in to re-use the existing integrationhealth service as well as `storage.IntegrationHealth` since all relevant fields and mechanisms exist. As an alternative, we could have also created a separate API service for this, however due to:
- the scale of declarative configuration not been of concern for the MVP (we currently estimate no more than 100 resources being created declaratively).
- the integration health propagation most likely being moved post-MVP to the status field of the custom resources.
I've opted to re-use the service and not create a separate API service.

Whilst adding the declarative configuration functions to the datastore, I also added a test suite to the datastore + cleaned up the code, specifically w.r.t. the SAC authZ checks.

In a follow-up, the integration health report will be re-used within the declarative config manager, registering each specific declarative configuration, updating their health on errors, as well as removing them in case the declarative config is removed.

## Checklist

## Testing Performed

- see added unit tests.


dhaus67:
* master
  * **PR #5048** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5048" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5048?utm_source=stack-comment).

central/integrationhealth/datastore/datastore_impl.go
```diff
@@ -21,91 +21,81 @@ type datastoreImpl struct {
 }
 
 func (ds *datastoreImpl) GetRegistriesAndScanners(ctx context.Context) ([]*storage.IntegrationHealth, error) {
-	if ok, err := integrationSAC.ReadAllowed(ctx); err != nil {
```
ivan-degtiarenko, line 24: By doing this we(inadvertently?) make these endpoints return errors to FE. In case of no permissions, that means that now UI will show an error instead of showing an empty health widget. I wonder if that's what we want

central/integrationhealth/datastore/datastore.go
```diff
@@ -19,6 +19,7 @@ type DataStore interface {
 	GetRegistriesAndScanners(ctx context.Context) ([]*storage.IntegrationHealth, error)
 	GetNotifierPlugins(ctx context.Context) ([]*storage.IntegrationHealth, error)
 	GetBackupPlugins(ctx context.Context) ([]*storage.IntegrationHealth, error)
+	GetDeclarativeConfigs(ctx context.Context) ([]*storage.IntegrationHealth, error)
 
 	UpdateIntegrationHealth(ctx context.Context, integrationHealth *storage.IntegrationHealth) error
```
ivan-degtiarenko, line 24: Let's use the fact we are touching this part of code and rename to `UpsertIntegrationHealth` - either here or in a separate PR, whatever you prefer.

central/integrationhealth/datastore/datastore_impl.go
```diff
@@ -21,91 +21,81 @@ type datastoreImpl struct {
...
 	return ds.store.Delete(ctx, id)
 }
 
 func (ds *datastoreImpl) GetIntegrationHealth(ctx context.Context, id string) (*storage.IntegrationHealth, bool, error) {
+	if err := sac.VerifyAuthzOK(integrationSAC.ReadAllowed(ctx)); err != nil {
+		return nil, false, err
+	}
+
 	health, found, err := ds.store.Get(ctx, id)
 	if !found || err != nil {
```
ivan-degtiarenko, line 87: I think now we can just `return ds.store.Get(ctx, id)` no matter what

central/integrationhealth/datastore/datastore_impl.go
dhaus67: Hm, yeah. I guess I don't see a reason why it shouldn't have been an error beforehand, but it might break UI in this case.
I reverted the `sac.VerifyAuthZOK` changes.

(unrelated: I feel we should just unify the datastores behavior and be clear: no permission == sac err returned)

ivan-degtiarenko (approved):
LGTM

PR #5050, opened 2023-02-28
Fix baseline stats and avoid unnecessary copies on Postgres

connorgorman:
## Description

The double counting of this function segment makes it very hard to use the segments to show aggregate time taken

## Checklist

## Testing Performed

Pretty straightforward

### DIFF
--- a/central/processindicator/datastore/datastore_impl.go
+++ b/central/processindicator/datastore/datastore_impl.go
@@ -106,7 +106,9 @@ func (ds *datastoreImpl) AddProcessIndicators(ctx context.Context, indicators ..
 	if err != nil {
 		return err
 	}
-
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


central/processindicator/datastore/datastore_impl.go
```diff
@@ -106,7 +106,9 @@ func (ds *datastoreImpl) AddProcessIndicators(ctx context.Context, indicators ..
 	if err != nil {
 		return err
 	}
-
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
```
dashrews78, line 109: nit:  comment here might be useful.

dashrews78 (approved):

central/processindicator/datastore/datastore_impl.go
connorgorman: I just removed this part now that I understand that the baseline stuff is probably the majority

PR #5054, opened 2023-03-01
ROX-14222: CI: Increase retries for pod readiness

gavin-stackrox:
## Description

Test flakes when an image pull does not complete within the 30x3 retries counter (Timer) are regular enough: https://issues.redhat.com/issues/?jql=labels%20%3D%20CI_Failure_Timeout_Pull. Debug shows that often the pull completes soon after the test fails.

This PR extends the retry counter to 60 retries with a the same 3 second wait between tries. Note: This is still within the 800 second global test timeout. (A somewhat imperfect solution but a flake reducer nonetheless and there is no "image pull is progressing albeit slowly" API call that I could find).

## Checklist

## Testing Performed

`all-test` + a few flavors. Bumping timeouts can have knock on effects.


theencee (approved):

BradLugo (approved):

gavin-stackrox:
All green!

PR #5058, opened 2023-03-01
Dashrews/ROX-15331 do not truncate

dashrews78:
## Description

truncating the parameters seems bad.  Best to panic on dev builds.  Not an easy way to return error from here, so the SQL error will occur in release builds and propagate back.    

## Checklist

## Testing Performed

local testing by sending in too many parameters to that call by modifying `central/processindicator/datastore/datastore_impl.go` to have a batch size of 70K to make sure the dev build paniced and crashed.

### DIFF
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -3,11 +3,9 @@ package search
 import (
 	"strings"
 
-	"github.com/pkg/errors"
 	v1 "github.com/stackrox/rox/generated/api/v1"
 	"github.com/stackrox/rox/pkg/logging"
 	"github.com/stackrox/rox/pkg/set"
-	"github.com/stackrox/rox/pkg/utils"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/search/parser.go
```diff
@@ -108,16 +110,12 @@ func parsePair(pair string, allowEmpty bool) (key string, values string, valid b
...
-	}
-	queries := make([]*v1.Query, 0, queryLength)
-	for idx, value := range values {
-		if idx == MaxQueryParameters {
-			log.Warnf("the query contained %d parameters and was truncated to %d parameters", len(values), MaxQueryParameters)
-			break
-		}
+	// A SQL query can have no more than 65535 parameters.
+	if len(values) > MaxQueryParameters {
+		utils.Should(errors.Errorf("too many parameters %d for a query.  No more than %d parameters allowed in single query", len(values), MaxQueryParameters))
```
connorgorman, line 115: Can we just use the UNEXPECTED flag? If we use should then this will blow up on scale tests

connorgorman (approved):

pkg/search/parser.go
```diff
@@ -108,16 +108,12 @@ func parsePair(pair string, allowEmpty bool) (key string, values string, valid b
...
-	}
-	queries := make([]*v1.Query, 0, queryLength)
-	for idx, value := range values {
-		if idx == MaxQueryParameters {
-			log.Warnf("the query contained %d parameters and was truncated to %d parameters", len(values), MaxQueryParameters)
-			break
-		}
+	// A SQL query can have no more than 65535 parameters.
+	if len(values) > MaxQueryParameters {
+		log.Errorf("UNEXPECTED: too many parameters %d for a query.  No more than %d parameters allowed in single query", len(values), MaxQueryParameters)
```
md2119, line 113: How about using `utils.Should(...)` to panic on dev builds?

md2119 (approved):

pkg/search/parser.go
md2119: Just saw Connor's comment, so ignore mine

PR #5066, opened 2023-03-01
Don't register some Bolt and RocksDB metrics when Postgres is enabled

connorgorman:
## Description

Declutter prometheus

Screenshot 2023-03-01 at 12 29 31 PM

## Checklist

## Testing Performed

CI will tell me if I tried to double register something

### DIFF
--- a/central/metrics/init.go
+++ b/central/metrics/init.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stackrox/rox/pkg/env"
 	"github.com/stackrox/rox/pkg/logging"
 )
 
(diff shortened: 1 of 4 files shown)


central/metrics/init.go
```diff
@@ -11,10 +12,16 @@ var (
...
+		)
+	}
+
 	prometheus.MustRegister(
-		boltOperationHistogramVec,
-		rocksDBOperationHistogramVec,
-		dackboxOperationHistogramVec,
 		graphQLOperationHistogramVec,
 		graphQLQueryHistogramVec,
 		indexOperationHistogramVec,
```
dashrews78, line 27: are we using this one for Postgres?

dashrews78 (approved):
LGTM.  just the one question on whether we can axe `indexOperationHistogramVec` or not.

central/metrics/init.go
connorgorman: Yeah, technically search queries go through it

connorgorman: Post RocksDB, let's move this to like a postgres_search_op_duration or something like that

dashrews78: sounds good.

PR #5067, opened 2023-03-01
Prefactor: Add new store for API Token expiration notification schedule

rhybrillou:
## Description

The goal is to introduce the internal datastore (singleton datastore with only one object) needed to later build the Notification loop for API Tokens about to expire

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


proto/storage/notification_schedule.proto
```diff
@@ -0,0 +1,14 @@
...
+option java_package = "io.stackrox.proto.storage";
+
+import "google/protobuf/timestamp.proto";
+import "gogoproto/gogo.proto";
+
+package storage;
+
+message NotificationSchedule {
+    reserved 1; // id, future field
+    google.protobuf.Timestamp last_run = 2;
```
rhybrillou, line 13: From @dhaus67 in https://github.com/stackrox/stackrox/pull/4969

> ~~Should this not have an ID associated with it?~~
> 
> Reading through the code, I realized that only one notification schedule exists currently.
> However, looking at the way forward, it may be that you can group certain items (either types of tokens, or a specific group of tokens) to a notification schedule.
> So, maybe its worthwhile to add an ID field here (and also give it the 1 field tag), and have a static ID defined for the "default" notification schedule, similar to what we do for other default resources (e.g. access scopes)?

central/apitoken/datastore/datastore_impl.go
```diff
@@ -167,3 +173,19 @@ func (b *datastoreImpl) SearchRawTokens(ctx context.Context, q *v1.Query) ([]*st
 	}
 	return objects, nil
 }
+
+func (b *datastoreImpl) GetNotificationSchedule(ctx context.Context) (*storage.NotificationSchedule, bool, error) {
+	if !env.PostgresDatastoreEnabled.BooleanSetting() {
+		log.Warn("Tried to retrieve API Token notification schedule not on Postgres, ignoring.")
+		return nil, false, nil
```
dhaus67, line 180: Same regarding returning an explicit error here as the comment below.

rhybrillou:
* master
    * **PR #4968** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4968" target="_blank">Graphite</a> 
      * **PR #5067** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5067" target="_blank">Graphite</a>  👈
        * **PR #4969** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/4969" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5067?utm_source=stack-comment).

migrator/migrations/frozenschema/v75/notification_schedules.go
```diff
@@ -0,0 +1,33 @@
+package schema
```
dhaus67, line 1: Should rather be v4.0 instead of v75, no?
Or are we just "counting" up and not associating this with the release?

dhaus67, line 1: I'm not sure what the new etiquette will be for the migrations, but the readme mentions post 3.74 that frozen schemas may be kept within the migration and not put into a shared place (currently two migrations follow this), I'm fine with it either way but just wanted to note it might be worthwhile being consistent.

migrator/migrations/m_175_to_m_176_create_notification_schedule_table/migration_test.go
```diff
@@ -0,0 +1,33 @@
...
+	notificationschedulestore "github.com/stackrox/rox/migrator/migrations/m_175_to_m_176_create_notification_schedule_table/notificationschedulestore"
+	pghelper "github.com/stackrox/rox/migrator/migrations/postgreshelper"
+	"github.com/stackrox/rox/pkg/sac"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMigration(t *testing.T) {
+	ctx := sac.WithAllAccess(context.Background())
+	pgTest := pghelper.ForT(t, false)
+	assert.NotNil(t, pgTest)
```
dhaus67, line 18: nit:
```suggestion
	require.NotNil(t, pgTest)
```

migrator/migrations/frozenschema/v75/notification_schedules.go
```diff
@@ -0,0 +1,33 @@
...
+	// NotificationSchedulesSchema is the go schema for table `notification_schedules`.
+	NotificationSchedulesSchema = func() *walker.Schema {
+		schema := walker.Walk(reflect.TypeOf((*storage.NotificationSchedule)(nil)), "notification_schedules")
+		return schema
+	}()
+)
+
+const (
+	// NotificationSchedulesTableName is the name of the table used to store data
+	NotificationSchedulesTableName = "notification_schedules"
```
dhaus67, line 27: Think this isn't used and can just be removed (unless the build tags are ticking me yet again 🥴).

migrator/migrations/m_175_to_m_176_create_notification_schedule_table/migration_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	storage := notificationschedulestore.New(pgTest.DB)
+	// Test get from the not-created-table returns an error
+	_, _, errPre := storage.Get(ctx)
+	assert.ErrorContains(t, errPre, "relation \"notification_schedules\" does not exist")
+
+	// Create the table
+	assert.NoError(t, createNotificationScheduleTable(pgTest.DB, pgTest.GetGormDB()))
+
+	// Test get from the created table returns nil
```
dhaus67, line 28: Let's cover also that you can a) upsert an entry and b) also retrieve said entry.

migrator/migrations/frozenschema/v75/notification_schedules.go
rhybrillou: I tried to push for decorrelation from datamodel version from release version in the past, in order to avoid that kind of confusion (see https://github.com/stackrox/stackrox/pull/4839 ).
In the end the feedback I got was (I think it is the point of the next comment):
> each migration (data manipulation or data model definition) should embed the files it needs.

I moved the files in https://github.com/stackrox/stackrox/pull/4968 and here to have the schema part of the migration itself.

rhybrillou: I made this migration consistent with the recent migrations.

dashrews78: I think this continues to evolve.  Personally I prefer isolating it to the migration itself because it is easier for me to understand and pinpoint exactly when a schema changed.

migrator/migrations/m_175_to_m_176_create_notification_schedule_table/migration_test.go
rhybrillou: c) delete an entry d) retrieve post-delete does not return any data

migrator/migrations/m_175_to_m_176_create_notification_schedule_table/migration.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+)
+
+func init() {
+	migrations.MustRegisterMigration(migration)
+}
+
+func createNotificationScheduleTable(_ *postgres.DB, gormDB *gorm.DB) error {
+	ctx := context.Background()
+	pgutils.CreateTableFromModel(ctx, gormDB, frozenSchema.CreateTableNotificationSchedulesStmt)
```
dashrews78, line 32: Is the sole purpose of this migration to create the table?  No data moved into it, simply create it?

rhybrillou: Correct. The associated store is a singleton store at the moment (single-item table). It will be populated on first use.

dashrews78: Technically based on current implementation a migration is not necessary.  In `migrator/main.go` after the migrations have executed around line 204 we make a call to `pkgSchema.ApplyAllSchemas(context.Background(), gormDB)`.  That would apply such a table.  That said, I'm not opposed to having the migration because honestly I'm not sure I like having that call to apply what's left after the migrations.  I see pros and cons to that.

dhaus67 (approved):

dashrews78 (approved):

PR #5068, opened 2023-03-01
ROX-11612 - fix image scan debug file truncation

dcaravel:
## Description
The contents of [`default-policies-test-struts-app.json`](https://github.com/stackrox/stackrox/blob/ebcfa1d0dde8813aef1d4e94cda4676565d45f3c/qa-tests-backend/src/test/groovy/DefaultPoliciesTest.groovy#L152) were being truncated during CI runs - reducing troubleshooting confidence.

The cause of the truncation was the associated Groovy/Java FileWriter's internal buffer was not being flushed. Calling `.close()` on the FileWriter will force a flush of the internal buffer.

This was noticed while investigating `ROX-11612`

## Checklist

## Testing Performed
Testing locally by creating a Groovy / Spock framework testing sandbox.

A test data file was created via `seq 1 500000 > lotta.txt`

A modified version of the `Helper.collectImageScanForDebug()` was created to reproduce the issue and verify fix

```groovy
static void collectImageScanForDebug() {
    try {
        Path imageScans = Paths.get("./image-scans")

        def cmd = ["/bin/bash", "-c", "cat ./lotta.txt"]
        Process proc = cmd.execute()
        String output = imageScans.resolve("test.txt").toAbsolutePath()
        FileWriter sout = new FileWriter(output)
        StringBuilder serr = new StringBuilder()

        proc.waitForProcessOutput(sout, serr)
        proc.waitFor()
        println serr
        sout.close()
    }
    catch (Exception e) {
        println "Could not collect image details", e
    }
}
```
Without the `sout.close()` the output file `./image-scans/test.txt` was truncated as expected, with `sout.close()` the full contents were written.

### DIFF
--- a/qa-tests-backend/src/main/groovy/util/Helpers.groovy
+++ b/qa-tests-backend/src/main/groovy/util/Helpers.groovy
@@ -148,6 +148,9 @@ class Helpers {
             if (proc.exitValue() != 0) {
                 log.warn "Failed to scan the image. Exit: ${proc.exitValue()}\nStderr: $serr"
             }
+
+            // closing the FileWriter will ensure internal buffer is flushed to file
+            sout.close()
         }
         catch (Exception e) {
             log.error("Could not collect image details", e)


RTann (approved):
LGTM though I wonder if it's worth considering try-with-resources since this is `AutoCloseable`

gavin-stackrox (approved):

dcaravel:
> LGTM though I wonder if it's worth considering try-with-resources since this is `AutoCloseable`

There are a few lines prior to the creation of the `FileWriter` that could `throw` exceptions - I figured this would be "easier to read" than `try-with-resources`, because the latter would require multiple try blocks (either nested or sequential) - but if you feel strongly about it happy to make the change 😉

RTann:
> > LGTM though I wonder if it's worth considering try-with-resources since this is `AutoCloseable`
> 
> There are a few lines prior to the creation of the `FileWriter` that could `throw` exceptions - I figured this would be "easier to read" than `try-with-resources`, because the latter would require multiple try blocks (either nested or sequential) - but if you feel strongly about it happy to make the change 😉

nah it sounds like the answer to "I wonder if it's worth considering try-with-resources" is no :)

PR #5072, opened 2023-03-02
Add default timeouts to Postgres via context and not via statement_timeout

connorgorman:
## Description

Add default timeouts if one doesn't exist on the context

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/pkg/postgres/rows.go
+++ b/pkg/postgres/rows.go
@@ -6,6 +6,19 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
+// Row wraps pgx.Row
+type Row struct {
+	pgx.Row
+	cancelFunc context.CancelFunc
+}
+
+// Scan wraps pgx.Row Scan
+func (r *Row) Scan(dest ...interface{}) error {
+	defer r.cancelFunc()
+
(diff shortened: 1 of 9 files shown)


pkg/postgres/rows.go
```diff
@@ -1,17 +1,35 @@
...
 
+// Row wraps pgx.Row
+type Row struct {
+	pgx.Row
+	cancelFunc context.CancelFunc
+}
+
+// Scan wraps pgx.Row Scan
+func (r *Row) Scan(dest ...interface{}) error {
+	defer r.cancelFunc()
```
dashrews78, line 17: does this cancel the context the first time we call scan?  I think that is OK because once we are calling scan the query is complete and we are processing the results.  I just want to make sure that is what it is doing and that is the intent.

connorgorman: this is just the raw row which does not have next. So the only interface is Scan() which means close will always be called inside scan even if the context has expired

dashrews78 (approved):
LGTM.  I'm going to need this for some migrations.

PR #5077, opened 2023-03-02
Dashrews/rox 15450 query errors

dashrews78:
## Description

Errors occurring during the execution of a query launched by the pgx`Query` function were not being returned.  The call to `Query` will return an error if it cannot execute the query at all BUT it will not if there is an error in the processing of the query.  Such as a math function that results in division by 0 or a timeout, those errors will not be returned in the call to `Query`.  In order to get those errors we have to check `rows.Err()` AFTER the rows are closed.  I found this running a large migration that migrated nothing because the call to Query silently timed out because we were not checking `rows.Err()`.  in this instance it simply returned 0 rows and no error.  So these updates are to use/return `rows.Err()` so that we are aware of errors that occurred during query execution.

For reference: https://pkg.go.dev/github.com/jackc/pgx/v4#Conn.Query

## Checklist

## Testing Performed

CI should tell us what we need to know.

### DIFF
--- a/central/debug/service/db_diagnostics.go
+++ b/central/debug/service/db_diagnostics.go
@@ -54,6 +54,7 @@ func getPostgresExtensions(ctx context.Context, dbPool *postgres.DB) []dbExtensi
 		log.Errorf("error fetching Postgres extensions: %v", err)
 		return nil
 	}
+	defer rows.Close()
 
 	extSlice := make([]dbExtension, 0)
 
@@ -75,7 +76,6 @@ func getPostgresExtensions(ctx context.Context, dbPool *postgres.DB) []dbExtensi
 		extSlice = append(extSlice, dbExt)
 	}
 
-	rows.Close()
 	if rows.Err() != nil {
 		log.Errorf("error getting complete extension information: %v", rows.Err())
 	}
--- a/central/debug/service/db_diagnostics.go
+++ b/central/debug/service/db_diagnostics.go
@@ -76,8 +76,8 @@ func getPostgresExtensions(ctx context.Context, dbPool *postgres.DB) []dbExtensi
 		extSlice = append(extSlice, dbExt)
 	}
 
-	if rows.Err() != nil {
-		log.Errorf("error getting complete extension information: %v", rows.Err())
+	if err := rows.Err(); err != nil {
+		log.Errorf("error getting complete extension information: %v", err)
 	}
 
 	return extSlice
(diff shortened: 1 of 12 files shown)


central/debug/service/db_diagnostics.go
```diff
@@ -76,5 +75,10 @@ func getPostgresExtensions(ctx context.Context, dbPool *postgres.DB) []dbExtensi
 		extSlice = append(extSlice, dbExt)
 	}
 
+	rows.Close()
```
connorgorman, line 78: I'm okay with the defer

connorgorman (changes requested):
So I think we can still use defer in the cases in which Next() is called without a potential to be broken because then next encapsulates the Close(). I think it's extremely odd that Next closes it, but that's okay

central/debug/service/db_diagnostics.go
```diff
@@ -76,5 +76,9 @@ func getPostgresExtensions(ctx context.Context, dbPool *postgres.DB) []dbExtensi
 		extSlice = append(extSlice, dbExt)
 	}
 
+	if rows.Err() != nil {
```
connorgorman, line 79: Nit: if err := rows.Err(); err !=nil

central/globaldb/postgres.go
```diff
@@ -187,6 +187,10 @@ func CollectPostgresStats(ctx context.Context, db *postgres.DB) *stats.DatabaseS
 		statsSlice = append(statsSlice, tableStat)
 	}
 
+	if rows.Err() != nil {
```
connorgorman, line 190: if err := rows.Err(); err != nil

Then can use this instead of rows.Err() below

central/networkgraph/flow/datastore/internal/store/postgres/store.go
```diff
@@ -545,18 +550,18 @@ func (s *flowStoreImpl) retryableRemoveMatchingFlows(ctx context.Context, keyMat
...
+	}
+	defer rows.Close()
+
 	// TODO(ROX-9921) Look at refactoring how these predicates work as an overall refactor of flows.
 	// This operation matches if the either the dest or src deployment no longer exists AND then
 	// if the last seen time is outside a time window.  Since we do not yet know what deployments exist
 	// in Postgres we cannot fully do this work in SQL.  Additionally, there may be issues with the synchronization
 	// of when flow is created vs a deployment deleted that may also make that problematic.
 	if keyMatchFn != nil {
-		rows, err := conn.Query(ctx, walkStmt, s.clusterID)
```
connorgorman, line 554: Actually, this seems like now we've made the query have greater scope? whereas it wasn't running before?

dashrews78: yeah you are right.  Also did that before I understood what Next was doing.  I will move it back.

connorgorman (approved):

PR #5079, opened 2023-03-02
[VM2.0][ROX-13854] GraphQL resolver for resource count by cve severity

md2119:
## Description

- Added a generic resource count by severity resolver
- Added `affectedImagesBySeverity` field to cve list resolver

## Checklist

## Testing Performed

image

### DIFF
--- a/central/views/imagecve/types.go
+++ b/central/views/imagecve/types.go
@@ -29,6 +29,7 @@ type CveView interface {
 	Get(ctx context.Context, q *v1.Query) ([]CveCore, error)
 }
 
+// ResourceCountByCVESeverity is the count of resources affected by cve distributed over severity.
 type ResourceCountByCVESeverity struct {
 	CriticalSeverityCount  int
 	ImportantSeverityCount int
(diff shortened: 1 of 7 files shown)


central/views/imagecve/types.go
```diff
@@ -27,3 +28,11 @@ type CveView interface {
 	Count(ctx context.Context, q *v1.Query) (int, error)
 	Get(ctx context.Context, q *v1.Query) ([]CveCore, error)
 }
+
+// ResourceCountByCVESeverity is the count of resources affected by cve distributed over severity.
+type ResourceCountByCVESeverity struct {
```
charmik-redhat, line 33: A thought for if and when countBySeverity are added for nodeCVE, clusterCVEs. I think this struct could be converted into an interface with `GetCriticalSeverityCount()` etc methods. The `resourceCountBySeverityResolver` can then have this interface as data field. The `imageCVECore.GetImagesBySeverity()` would then return a `ImageCountBySeverity` struct which would implement `ResourceCountBySeverity` . Same for other cveCore types.

charmik-redhat (approved):
LGTM. Added a thought for future.

PR #5080, opened 2023-03-03
ROX-15327: Remove Circle CI references and simplify tagging

gavin-stackrox:
## Description

Less is more! And tag determination was overly complicated.

## Checklist

## Testing Performed

- [x] Regular CI - unrelated flake in postgres upgrade
- [ ] Release simulation (if possible)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,12 +166,6 @@ update-shellcheck-skip:
 	$(SILENT)rm -f scripts/style/shellcheck_skip.txt
 	$(SILENT)$(BASE_DIR)/scripts/style/shellcheck.sh update_failing_list
 
-.PHONY: ci-config-validate
-ci-config-validate:
-	@echo "+ $@"
-	$(SILENT)circleci diagnostic > /dev/null 2>&1 || (echo "Must first set CIRCLECI_CLI_TOKEN or run circleci setup"; exit 1)
-	circleci config validate --org-slug gh/stackrox
-
 .PHONY: fast-central-build
 fast-central-build: central-build-nodeps
 
--- a/Makefile
+++ b/Makefile
@@ -541,8 +541,8 @@ $(CURDIR)/image/rhel/bundle.tar.gz:
 .PHONY: $(CURDIR)/image/rhel/Dockerfile.gen
 $(CURDIR)/image/rhel/Dockerfile.gen:
 	ROX_IMAGE_FLAVOR=$(ROX_IMAGE_FLAVOR) \
-	LABEL_VERSION=$(BUILD_TAG) \
-	LABEL_RELEASE=$(BUILD_TAG) \
+	LABEL_VERSION=$(TAG) \
+	LABEL_RELEASE=$(TAG) \
 	QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION) \
 	envsubst '$${ROX_IMAGE_FLAVOR} $${LABEL_VERSION} $${LABEL_RELEASE} $${QUAY_TAG_EXPIRATION}' \
 	< $(CURDIR)/image/rhel/Dockerfile.envsubst > $(CURDIR)/image/rhel/Dockerfile.gen
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,12 @@ SILENT ?= @
 #  usage: "path/to/ignored|another/path"
 UNIT_TEST_IGNORE := "stackrox/rox/sensor/tests"
 
-ifeq ($(TAG),)
-TAG=$(shell git describe --tags --abbrev=10 --dirty --long --exclude '*-nightly-*')
+ifeq ($(BUILD_TAG),)
+BUILD_TAG=$(shell git describe --tags --abbrev=10 --dirty --long --exclude '*-nightly-*')
 endif
 
 # Set expiration on Quay.io for non-release tags.
-ifeq ($(findstring x,$(TAG)),x)
+ifeq ($(findstring x,$(BUILD_TAG)),x)
 QUAY_TAG_EXPIRATION=13w
 else
 QUAY_TAG_EXPIRATION=never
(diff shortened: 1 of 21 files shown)


make/env.mk
```diff
@@ -35,16 +35,11 @@ ifeq ($(GOBIN),)
 GOBIN := $(GOPATH)/bin
 endif
 
-TAG := # make sure tag is never injectable as an env var
```
msugakov, line 38: I'm failing to understand why it was introduced in https://github.com/stackrox/rox/pull/6521 therefore ok, let's see how it goes without that.

msugakov:
Note that `CIRCLE_TAG` renaming should be mirrored in the downstream. I can help with that.

scripts/ci/lib.sh
```diff
@@ -665,13 +665,11 @@ is_tagged() {
 }
 
 is_nightly_run() {
-    [[ "${CIRCLE_TAG:-}" =~ -nightly- ]] || [[ "${GITHUB_REF:-}" =~ nightly- ]]
+    [[ "${TAG:-}" =~ -nightly- ]] || [[ "${GITHUB_REF:-}" =~ nightly- ]]
 }
 
 is_in_PR_context() {
-    if is_CIRCLECI && [[ -n "${CIRCLE_PULL_REQUEST:-}" ]]; then
```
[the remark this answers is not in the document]
gavin-stackrox: That one is required. See comment above it.

scripts/reference/build-quay-operator-bundles.sh
```diff
@@ -13,9 +13,9 @@ for tag in $(git -C "$workdir" tag | egrep '^3\.(6[3-9]|0\.62)\.\d+$' | sort -V)
 	docker push "quay.io/rhacs-eng/stackrox-operator:${version}"
 
 	git -C "$workdir" checkout "$tag"
-	CI=1 CIRCLE_TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" bundle-build
+	CI=1 TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" bundle-build
```
msugakov, line 16: Probably
```suggestion
	CI=1 BUILD_TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" bundle-build
```

scripts/reference/build-quay-operator-bundles.sh
```diff
@@ -13,9 +13,9 @@ for tag in $(git -C "$workdir" tag | egrep '^3\.(6[3-9]|0\.62)\.\d+$' | sort -V)
...
 	docker push "quay.io/rhacs-eng/stackrox-operator-bundle:v${version}"
 
-	CI=1 CIRCLE_TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" index-build
+	CI=1 TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" index-build
```
msugakov, line 19:
```suggestion
	CI=1 BUILD_TAG="$tag" IMAGE_REPO="quay.io/rhacs-eng" IMAGE_TAG_BASE="quay.io/rhacs-eng/stackrox-operator" make -C "$workdir/operator" index-build
```

Makefile
```diff
@@ -18,12 +18,12 @@ SILENT ?= @
 #  usage: "path/to/ignored|another/path"
 UNIT_TEST_IGNORE := "stackrox/rox/sensor/tests"
 
-ifeq ($(TAG),)
-TAG=$(shell git describe --tags --abbrev=10 --dirty --long --exclude '*-nightly-*')
+ifeq ($(BUILD_TAG),)
+BUILD_TAG=$(shell git describe --tags --abbrev=10 --dirty --long --exclude '*-nightly-*')
 endif
```
msugakov, line 23: After looking at `scripts/check-debugger.sh`, I remember why there was a difference between `BUILD_TAG` (former `CIRCLE_TAG`) and just `TAG` (that was present in this `Makefile` before edits) in the context of `Makefile`.

* `BUILD_TAG`/`CIRCLE_TAG` meant a tag provided externally, i.e. from CI when we need to build something like `3.74.1-rc.1`.
* `TAG` held the current value of the tag.
  * If `BUILD_TAG`/`CIRCLE_TAG` was provided, `TAG` simply got its value.
  * Otherwise, `TAG` held a result of `git describe`, i.e. some `3.74.x-266-gb6315d348b` thingy.

To me this explains the existence of that `TAG=` trick in `make/env.mk` (intent, but not a reasoning).

Without such separation, we'll not be able to differentiate tagged (i.e. release, RC, nightly) and untagged builds. It seems to me that `scripts/check-debugger.sh` is the only place that does that.
Do you know other places?
How broken is the debugger support that we can leave `check-debugger.sh` in incorrect state?

msugakov: > It seems to me that scripts/check-debugger.sh is the only place that does that.
> Do you know other places?

Here's a deeper example. Our `Makefile` calls into itself, e.g. in case of `main-build-dockerized` target
https://github.com/stackrox/stackrox/blob/d8f6253dd83a7e5d03dbcc8ab0060ff3da1e0799/Makefile#L404-L408

The fact that `BUILD_TAG` **environment variable** is provided highlights a nice distinction that:
* Every environment variable will be visible as makefile variable with the same name. `BUILD_TAG=blah make ...` will do `$(BUILD_TAG)` equal to `blah` during the `Makefile` execution.
* However, it does not work the other way. Makefile variable does not become an environment variable with the same name. I.e. when `${BUILD_TAG}` is not defined in the environment, `main-build-dockerized` propagates no `BUILD_TAG` value into `docker run` context.

This means, `main-build-dockerized` should still work correctly w.r.t. actual `GOTAGS=release` (or not) in the resulting binaries, but the part **why** it works correctly is a bit too subtle.

WDYT, should we keep things the way they are now or should we restore `BUILD_TAG`-`TAG` duality in the `Makefile` for better explicitness?

Looking at it from another angle, the "`BUILD_TAG`-`TAG` duality" was to differentiate when we should and shouldn't apply `GOTAGS=release` during Go binaries build. Maybe that's the place that should be changed to make rules more explicit and straightforward?

gavin-stackrox: > WDYT, should we keep things the way they are now or should we restore `BUILD_TAG`-`TAG` duality in the `Makefile` for better explicitness?

I think I will put it back the way it was i.e. `TAG` in the make environment.

msugakov (approved):

PR #5081, opened 2023-03-03
ROX-12578: Ensure no duplicate group entries are created

dhaus67:
## Description

Within the groups API, it is possible to create groups which specify the same `auth provider, key, value, role name` multiple times.

While this doesn't necessarily cause any problems, it doesn't make much sense to allow these duplications from an API point of view.

Hence, this PR ensures that no duplicated groups may be created.

This is done by verifying for each group thats either added or updated whether, for the same properties, a group exists _and_ that group has the same role.
If that's the case, then we disallow adding the group and make it a no-op.

Few caveats / notes on the PR:
- I first tried to check whether its possible to create a unique constraint for all four values (auth provider, key, value, role name), but this doesn't seem to be support OOTB with our current code generation.
- While theres a performance impact by filtering through the groups API for each upsert, it's a low-volume store, and hence it shouldn't lead to performance degredation.

I'm currently undecided whether we want to perform a migration, removing potentially duplicated group entries, happy for feedback on this.

If that will be done, it will be a follow-up PR.

## Checklist

## Testing Performed

- see added unit tests.

### DIFF
--- a/central/group/datastore/datastore_impl.go
+++ b/central/group/datastore/datastore_impl.go
@@ -7,10 +7,12 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/group/datastore/internal/store"
+	"github.com/stackrox/rox/central/group/datastore/serialize"
 	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/declarativeconfig"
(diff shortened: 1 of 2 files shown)


dhaus67:
* master
  * **PR #5081** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5081" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5081?utm_source=stack-comment).

PR #5097, opened 2023-03-03
Added core_bpf to enum of collection methods

JoukoVirtanen:
## Description

Add "CORE_BPF" as a collection method option for Collector configuration in Central helm charts and operator.

## Checklist
- [x] Used helm charts to deploy with CORE_BPF
- [x] Used operator to deploy with CORE_BPF

### Helm Charts

To deploy with helm charts 

settings+=(--set collector.collectionMethod="CORE_BPF")

helm install -n stackrox stackrox-secured-cluster-services rhacs/secured-cluster-services \
    -f perf-bundle.yml \
    "${settings[@]}"

The other settings used here specified the images to use. The image tag used in all cases, except for scanner, was 4.0.x-137-gd98774c7a0. All pods ran. The collector log had the following 

[INFO    2023/04/19 22:51:06] == Collector Startup Diagnostics: ==
[INFO    2023/04/19 22:51:06]  Connected to Sensor?       true
[INFO    2023/04/19 22:51:06]  Kernel driver candidates:
[INFO    2023/04/19 22:51:06]    CO.RE eBPF probe (available)
[INFO    2023/04/19 22:51:06]    collector-ebpf-4.18.0-372.49.1.el8_6.x86_64.o (available)
[INFO    2023/04/19 22:51:06]  Driver loaded into kernel: collector-ebpf-4.18.0-372.49.1.el8_6.x86_64.o
[INFO    2023/04/19 22:51:06] ====================================

### Operator

See https://github.com/stackrox/stackrox/pull/5760


operator/apis/platform/v1alpha1/securedcluster_types.go
```diff
@@ -197,6 +197,8 @@ type CollectionMethod string
 const (
 	// CollectionEBPF means: use EBPF collection.
 	CollectionEBPF CollectionMethod = "EBPF"
+	// CollectionModernBPF means: use modern_bpf collection.
+	CollectionModernBpf CollectionMehtod = "modern_bpf"
```
SimonBaeumer, line 201: `BPF` in caps?

operator/bundle/manifests/platform.stackrox.io_securedclusters.yaml
```diff
@@ -372,6 +372,7 @@ spec:
                           in these section will not have any effect.
                         enum:
                         - EBPF
+                        - modern_bpf
```
SimonBaeumer, line 375: Why using snake case instead of upper camel case here?

SimonBaeumer (changes requested):

operator/apis/platform/v1alpha1/securedcluster_types.go
porridge: BTW I'm not very fond of using adjectives such as `new`, `modern`, `ng`, because they tend to look weird after a few months/years at best, and very confusing at worst, when the next shiny hotness is born. Perhaps just call it `BTF` if that's what you mean? :thinking:

erthalion: BTF itself is also not quite right, as it's still a BPF probe, just of different sort -- so it's technically wrong to call it BTF. But we're open for suggestions, theoretically instead of an exact name it could be a "nick name" for the new collection method. Any other ideas?

porridge: I see. What's different about it? Worst case, perhaps `EBPF2` or pick a cool-sounding name from your favourite sci-fi universe? :wink:

SimonBaeumer: @erthalion How does BTF differentiate from the other collection methods? Is there something the consumer of the API should know about this collection method?

erthalion: > What's different about it?

> How does BTF differentiate from the other collection methods? Is there something the consumer of the API should know about this collection method?

The intrinsic difference is about using "modern" (hence the first ideas for the name) kernel functionality, to allow the probe work on any sufficiently new kernel version. The user -visible API is the same, in theory the consumer shouldn't notice any difference whatsoever on supported platforms. So, what's your favorite sci-fi universe then? :)

porridge: > > What's different about it?
> 
> > How does BTF differentiate from the other collection methods? Is there something the consumer of the API should know about this collection method?
> 
> The intrinsic difference is about using "modern" (hence the first ideas for the name) kernel functionality,

Does this functionality have any feature apart from being modern? :-D

> So, what's your favorite sci-fi universe then? :)

Mine is Lem's but I'm not the implementer here :-)

erthalion: > Does this functionality have any feature apart from being modern? :-D

Yes, it works on *all* new kernels, where the previous was working only on *one* kernel. Maybe a "universal_bpf" then?

> > So, what's your favorite sci-fi universe then? :)
>
> Mine is Lem's but I'm not the implementer here :-)

I would love to see a collection method "ijon_tichy"!

SimonBaeumer: Should a user ever prefer EBPF over BPF if their Kernel is on a minimum version? What is the minimum version?

erthalion: This would be described in the documentation, but the short summary is that the new collection method would be preferable when it become stable enough. The minimal supported kernel version is a fluid thing in this case, and Collector would be responsible to figure out if the new probe is supported. If not it will fallback to the regular ebpf.

porridge, line 201:
```suggestion
	// CollectionUniversalEBPF means: use UNIVERSAL_EBPF collection.
	CollectionUniversalEBPF CollectionMethod = "UNIVERSAL_EBPF"
```

porridge (changes requested):
Please also update lines 193:194.

porridge:
@vladbologa please take over as the reviewer as you pick up the operator hat from tomorrow

operator/apis/platform/v1alpha1/securedcluster_types.go
```diff
@@ -197,6 +197,8 @@ type CollectionMethod string
 const (
 	// CollectionEBPF means: use EBPF collection.
 	CollectionEBPF CollectionMethod = "EBPF"
+	// CollectionModernBPF means: use modern_bpf collection.
```
0x656b694d, line 200: The comment doesn't match the code.

porridge (changes requested):

operator/apis/platform/v1alpha1/securedcluster_types.go
```diff
@@ -190,13 +190,15 @@ type PerNodeSpec struct {
...
-// +kubebuilder:validation:Enum=EBPF;KernelModule;NoCollection
+// CollectionMethod defines the method of collection used by collector. Options are 'EBPF', 'CoreBpf', 'KernelModule' or 'None'.
+// +kubebuilder:validation:Enum=EBPF;CoreBpf;KernelModule;NoCollection
 type CollectionMethod string
 
 const (
 	// CollectionEBPF means: use EBPF collection.
 	CollectionEBPF CollectionMethod = "EBPF"
+	// CollectionCoreBpf means: use CoreBpf collection.
+	CollectionCoreBpf CollectionMethod = "CoreBpf"
```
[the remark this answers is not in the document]
JoukoVirtanen: I went with the second option. @erthalion what do you think?

vladbologa (approved):
LGTM

operator/apis/platform/v1alpha1/securedcluster_types.go
```diff
@@ -190,13 +190,15 @@ type PerNodeSpec struct {
...
-// +kubebuilder:validation:Enum=EBPF;KernelModule;NoCollection
+// CollectionMethod defines the method of collection used by collector. Options are 'EBPF', 'CORE_BPF', 'KernelModule' or 'None'.
+// +kubebuilder:validation:Enum=EBPF;CORE_BPF;KernelModule;NoCollection
 type CollectionMethod string
 
 const (
 	// CollectionEBPF means: use EBPF collection.
 	CollectionEBPF CollectionMethod = "EBPF"
+	// CollectionCoreBpf means: use CORE_BPF collection.
+	CollectionCoreBpf CollectionMethod = "CORE_BPF"
```
porridge, line 201: Please also capitalize the const name.
```suggestion
	CollectionCOREBPF CollectionMethod = "CORE_BPF"
```

porridge (changes requested):

operator/apis/platform/v1alpha1/securedcluster_types.go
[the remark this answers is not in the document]
erthalion: Agree, with a separator it looks better.

erthalion (approved):

porridge (approved):
Please kindly update the PR description too!

operator/bundle/manifests/platform.stackrox.io_securedclusters.yaml
JoukoVirtanen: Using CORE_BPF now

SimonBaeumer (approved):

PR #5099, opened 2023-03-03
ROX-12578: Create unique constraint for groups

dhaus67:
## Description

This PR ensures that the tuple `auth provider id, key, value, role name` is unique across all groups. Previously, this was not the case and not enforce, and it was possible to create duplicated groups.

This had to be done in with workaround and manual changes to the generated code:
- First, each field that should be a separate column had to have a `search` tag associated with it. Otherwise, the schema generation won't pick it up.
- The code generation was run.
- Afterwards, the `gen.go` file was removed, as well as the autogenerated headers from the generated store files. The files were changed to pass linting checks.
- The `group.go` schema file was changed. Since the code generation doesn't support it yet, the `uniqueIndex` struct tag was added manually.

The search fields are both hidden, so they are not exposed to end-users. Once [ROX-13790](https://issues.redhat.com/browse/ROX-13790) lands, the search field workaround, as well as the changes to the autogenerated files can be reverted and hopefully autogeneration can be reused.

Note:
There will be a follow-up with the schema migration.

## Checklist

## Testing Performed

- see added tests for Postgres enforcing the unique_constraint on the database level.


dhaus67:
* master
  * **PR #5099** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5099" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5099?utm_source=stack-comment).

central/group/datastore/datastore_impl_postgres_test.go
```diff
@@ -0,0 +1,118 @@
...
+	s.True(pgutils.IsUniqueConstraintError(err))
+
+	// 3. Adding a different group should work.
+	group.RoleName = "headmaster"
+	group.Props.Id = ""
+	err = s.groupsDatastore.Add(s.ctx, group)
+	s.NoError(err)
+}
+
+func (s *groupsWithPostgresTestSuite) TestUpdateGroups() {
```
ivan-degtiarenko, line 86: Let's also test `Mutate` behaviour as it was specified as problematic in the original ticket

dashrews78:
Can you add a comment to [ROX-13790](https://issues.redhat.com/browse/ROX-13790) or link a ticket to it so we can return this to being generated at that time?

pkg/postgres/schema/groups.go
```diff
@@ -30,11 +33,16 @@ var (
...
 const (
+	// GroupsTableName specifies the name of the table in postgres.
 	GroupsTableName = "groups"
 )
 
 // Groups holds the Gorm model for Postgres table `groups`.
 type Groups struct {
-	PropsId    string `gorm:"column:props_id;type:varchar;primaryKey"`
-	Serialized []byte `gorm:"column:serialized;type:bytea"`
+	PropsID             string `gorm:"column:props_id;type:varchar;primaryKey"`
```
dashrews78, line 42: this is bordering on the case where having a composite primary key vs using PropsId would be an option.  That said the way you have it more closely mimics how we handle most scenarios and keeps it simpler for maintenance.  Just wanted to point out that the ID we generate is not necessarily providing a ton of value.

dhaus67: Yeah, agreed (this was also how it was done beforehand, modulo that the role name was not part of the composite key). We decided to add the ID because this is. what we usually do I suppose.

It definitely has room for improvement in the future.

dhaus67:
@dashrews78 I added the header to `store.go, store_test.go, groups.go`, if there's additional places where the comment should be let me know.

pkg/postgres/pgutils/error.go
```diff
@@ -66,3 +66,15 @@ func IsTransientError(err error) bool {
 	}
 	return errorhelpers.IsAny(err, io.EOF, io.ErrUnexpectedEOF, io.ErrClosedPipe)
 }
+
+const (
+	errCodeUniqueConstraint = "23505"
+)
+
+// IsUniqueConstraintError specifies if the passed error is due to a unique constraint violation.
+func IsUniqueConstraintError(err error) bool {
```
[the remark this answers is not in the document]
dhaus67: (more of a separate PR kind-of-thing): it'd be nice to have a generic `ToRoxError` from `pgError.Error`.

dashrews78:
> @dashrews78 I added the header to `store.go, store_test.go, groups.go`, if there's additional places where the comment should be let me know.

I think that covers it.

dashrews78 (approved):
LGTM.

dashrews78 (commented):
One question though.  It might be worth running the gke-postgres-upgrade-tests  because not doing the migration may or may not break that test.

ivan-degtiarenko (approved):

rhybrillou (approved):

rhybrillou (approved):

PR #5103, opened 2023-03-04
Jv rox 15187 fix processes listening on ports test flake retry plop

JoukoVirtanen:
## Description

See https://github.com/stackrox/stackrox/pull/5159 for an improved version of this. The difference there is that old unmatched listening endpoints are removed in that PR. Also rows are not deleted and inserted back. Instead they are upserted.

Retries to match old plops to process indicators for which the match previously failed, in the same loop in which new plops are added. This fixes the flakiness in the e2e test for plop. There will be another attempted solution in which the retries are not done in the same path as where new plops are added. That PR can be found here https://github.com/stackrox/stackrox/pull/5143

## Checklist
- [x] Ran e2e tests 200 times

## Testing Performed

See above

### DIFF
--- a/central/processlisteningonport/datastore/datastore_impl.go
+++ b/central/processlisteningonport/datastore/datastore_impl.go
@@ -40,37 +40,6 @@ func newDatastoreImpl(
 	}
 }
 
-//type ProcessListeningOnPortStorage struct {
-//        // Ideally it has to be GENERATED ALWAYS AS IDENTITY, which will make it a
-//        // bigint with a sequence. Unfortunately at the moment some bits of store
-//        // generator assume an id has to be a string.
-//        Id                 string           `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty" sql:"pk,type(uuid)"`
-//        Port               uint32           `protobuf:"varint,2,opt,name=port,proto3" json:"port,omitempty" search:"Port,store"`
-//        Protocol           L4Protocol       `protobuf:"varint,3,opt,name=protocol,proto3,enum=storage.L4Protocol" json:"protocol,omitempty" search:"Port Protocol,store"`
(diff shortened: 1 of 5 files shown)


sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -297,10 +297,6 @@ func (m *networkFlowManager) enrichAndSendProcesses() {
 
 	m.enrichedProcessesLastSentState = currentProcesses
 
-	if len(updatedProcesses) == 0 {
```
JoukoVirtanen, line 300: Removing this was necessary so that old plops would be retried even when there are no new plops.

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -40,6 +40,50 @@ func newDatastoreImpl(
...
+		unmatchedIds = append(unmatchedIds, val.Id)
+		plop := convertPlopFromStorageToPlopFromSensor(val)
+		portProcesses = append(portProcesses, plop)
+	}
+
+	// Unmatched plop objects are deleted and added back in to avoid duplicate rows.
+	// Deleting unmatched plops is not the most efficient solution.
+	// Instead plopObjects should be set correctly in AddProcessListeningOnPort.
+	ds.storage.DeleteMany(ctx, unmatchedIds)
+
```
JoukoVirtanen, line 83: Unmatched plops should not be allowed to persist in the table forever. There should probably be another column in the table that specifies when the row was first inserted. Unmatched plops could then be removed say after 90 seconds.

central/sensor/service/pipeline/processlisteningonport/pipeline.go
```diff
@@ -56,12 +56,11 @@ func (s *pipelineImpl) Run(
 	if s.dataStore != nil && update != nil {
 		portProcesses := update.GetProcessesListeningOnPorts()
 
-		if portProcesses != nil {
```
JoukoVirtanen, line 59: Running AddProcessListeningOnPort when portProcesses is nil is necessary so that old plops are retried even when there are no new plops added.

PR #5122, opened 2023-03-07
Add hash index to process baseline ID

connorgorman:
## Description

Testing to see potential speed up

## Checklist

## Testing Performed

Scale test


rhybrillou (commented):
I suppose a migration is needed in order to deploy the new index on existing deployments.
The migration should be pretty basic and simply apply the schema change.

PR #5129, opened 2023-03-07
ROX-15411: De-emphasizing the unconnected nodes

sachaudh:
## Description

This PR adds styling to the nodes to fade out unconnected nodes when a specific node is selected

Screenshot 2023-03-06 at 11 59 15 AM
Screenshot 2023-03-06 at 11 59 25 AM
Screenshot 2023-03-06 at 11 59 33 AM

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
@@ -160,10 +160,14 @@ function fadeOutUnconnectedNodes(
     nodes: CustomNodeModel[],
     edges: CustomEdgeModel[],
     selectedNodeId: string | undefined
-) {
-    const connectedNodeIds = getConnectedNodeIds(nodes, edges, selectedNodeId);
+): CustomNodeModel[] {
+    if (!selectedNodeId) {
+        return nodes;
+    }
+    const connectedNodeIds = getConnectedNodeIds(edges, selectedNodeId);
     const modifiedNodes: CustomNodeModel[] = nodes.map((node) => {
         const { data } = node;
+        // We only want to fade out nodes (not groups), so we target node types specifically
         if (
             data.type === 'DEPLOYMENT' ||
             data.type === 'CIDR_BLOCK' ||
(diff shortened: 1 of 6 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
```diff
@@ -150,6 +155,39 @@ function getDisplayEdges(edges: CustomEdgeModel[], showEdgeLabels: boolean): Cus
...
+            const isFadedOut = isNodeSelected
+                ? !isConnectedToSelectedNode && !isSelectedNode
+                : false;
+            return {
+                ...node,
+                data: {
+                    ...data,
+                    isFadedOut,
+                },
+            } as CustomNodeModel;
```
vjwilson, line 184: Could we short-circuit this sooner? It seems like if there is no `selectedNodeId` given, then there is no reason to calculate `isConnectedToSelectedNode` or `isSelectedNode`

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -605,3 +606,24 @@ export function createExtraneousEdges(selectedNodeId: string): {
     };
     return { extraneousEgressEdge, extraneousIngressEdge };
 }
+
+// This function returns the ids of nodes that are connected to the selected node
+export function getConnectedNodeIds(
+    nodes: CustomNodeModel[],
```
vjwilson, line 612: We don't seem to be using the `nodes` parameter.

Is it an artifact from an earlier version of the code that we could remove?

vjwilson (approved):
Good job tracking this down and implementing it.

Just a few little comments to consider.

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
sachaudh: Oh, yea probably. Good caatch

ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphContainer.tsx
sachaudh: Actually, I think we want to keep it like this because we do want to update the node data correctly whenever we go from selecting a node to unselecting a node

PR #5130, opened 2023-03-07
Don't silently fail on junit2jira

janisz:
## Description

Currently when junit2jira fails we skip it and continue normal flow. Since we now rely on this tool for our day to day job let's not ignore failures and get slca notifications for them as a part of CI trap. 

## Checklist

## Testing Performed

CI

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -1165,18 +1165,18 @@ store_test_results() {
     local from="$1"
     local to="$2"
 
+    info "Copying test results from $from to $to"
+
+    local dest="${ARTIFACT_DIR}/junit-$to"
+
+    cp -a "$from" "$dest" || true # (best effort)
+
     if ! is_in_PR_context; then
         info "Creating JIRA task for failures found in $from"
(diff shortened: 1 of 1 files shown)


gavin-stackrox (commented):
One problem I see with this is that if that curl, junit2jira or jira are having a bad day, every merge job pass and success will result in a slack notice to #test-failures. 

Also, if this allowed to fail it should move after the `cp` so at least the junit files get copied for prow to look at.

porridge (changes requested):
Let's move the `cp` before the `junit2jira` invocation, then. We don't want to skip copying if the tool has problems, right?

janisz:
How about making it like slack notifications but include nightly runs?

gavin-stackrox:
Whatabout only ignoring failures from the curl (as well as moving below the cp)? That is probably where the most 'weakness' lies from github rate limiting. If the tool or jira access is a problem we have more control and can always spin a quick fix. Another option is put junit2jira in the repo or the ci image (with the usual image update pain).

PR #5133, opened 2023-03-07
Dashrews/ROX-15786 migration timeouts

dashrews78:
## Description

Some Postgres to Postgres migrations may take a significant amount of time.  The timeout periods need to be extended for those long running jobs.  It was discovered during this effort that the index on clusterid is no longer needed as the partitions take care of that.  Simply eliminating that index greatly improved the performance of the migration without impacting queries because it was essentially indexing the entire partition for no reason.  We effectively accomplish the same thing simply be using the partitions.

## Checklist

## Testing Performed

Migration tests and lots of scale tests of migrations to see how it performed and if it would succeed at all.  This is how the issue with the clusterid index was discovered.

### DIFF
--- a/migrator/clone/postgres/db_clone_manager_impl.go
+++ b/migrator/clone/postgres/db_clone_manager_impl.go
@@ -281,7 +281,7 @@ func (d *dbCloneManagerImpl) doPersist(cloneName string, prev string) error {
 
 	err = pgadmin.AnalyzeDatabase(d.adminConfig, CurrentClone)
 	if err != nil {
-		log.Warnf("unable to force analyze restore database:  %v", err)
+		log.Warnf("unable to force analyze restore database: %v", err)
 	}
 
 	return nil
--- a/migrator/clone/postgres/db_clone_manager_impl.go
+++ b/migrator/clone/postgres/db_clone_manager_impl.go
@@ -279,6 +279,11 @@ func (d *dbCloneManagerImpl) doPersist(cloneName string, prev string) error {
 		}
 	}
 
+	err = pgadmin.AnalyzeDatabase(d.adminConfig, CurrentClone)
+	if err != nil {
+		log.Warnf("unable to force analyze restore database:  %v", err)
+	}
+
 	return nil
 }
 
--- a/migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
+++ b/migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
@@ -98,6 +98,29 @@ func MigrateToPartitions(gormDB *gorm.DB, db *postgres.DB) error {
 			return err
 		}
 		log.WriteToStderrf("Trimmed network flows to length of %d from %d.", migratedCount, previousCount)
+
+		// Add the index for the cluster
+		for _, index := range updatedSchema.PartitionIndexes {
+			namePart := cluster
+			difference := 64 - len(index.IndexName) - 2 + len(cluster)
+
+			if difference < 0 {
+				namePart = cluster[:64+difference]
+			}
+
+			officialName := fmt.Sprintf(index.IndexName, namePart)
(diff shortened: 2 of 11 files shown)


pkg/postgres/pgadmin/admin_utils.go
```diff
@@ -191,7 +193,9 @@ func AnalyzeDatabase(config *postgres.Config, dbName string) error {
 	// Close the admin connection pool
 	defer connectPool.Close()
 
-	_, err := connectPool.Exec(context.Background(), "ANALYZE")
+	ctx, cancel := context.WithTimeout(context.Background(), analyzeTimeout)
+	defer cancel()
+	_, err := connectPool.Exec(ctx, "ANALYZE")
 
 	log.Debug("Anaylze done")
```
connorgorman, line 200: Analyze

migrator/clone/postgres/db_clone_manager_impl.go
```diff
@@ -279,6 +279,11 @@ func (d *dbCloneManagerImpl) doPersist(cloneName string, prev string) error {
 		}
 	}
 
+	err = pgadmin.AnalyzeDatabase(d.adminConfig, CurrentClone)
+	if err != nil {
+		log.Warnf("unable to force analyze restore database:  %v", err)
```
connorgorman, line 284: extra space after :

migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
```diff
@@ -39,37 +38,44 @@ var (
 
 // MigrateToPartitions updates the btree network flow indexes to be hash
 func MigrateToPartitions(gormDB *gorm.DB, db *postgres.DB) error {
+	parentCtx := context.Background()
```
connorgorman, line 41: just name this ctx?

migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
```diff
@@ -39,37 +38,44 @@ var (
...
+
+	err := analyzeOldTable(parentCtx, db)
+	if err != nil {
+		log.WriteToStderrf("unable to analyze network_flows.  Will continue processing though it may be slow. %v", err)
+	}
+
 	// First get the distinct clusters in the network_flows table
-	clusters, err := getClusters(db)
+	clusters, err := getClusters(parentCtx, db)
```
connorgorman, line 49: again curious if we shouldn't just read clusters here. Seems much faster

migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
```diff
@@ -103,11 +109,49 @@ func MigrateToPartitions(gormDB *gorm.DB, db *postgres.DB) error {
...
+func getPreviousCount(parentCtx context.Context, store previous.FlowStore) (int, error) {
+	// Due to heavy load with network flows, each call to the store could take a long time so we will
+	// pass a context with the migration timeout to them.
+	ctx, cancel := context.WithTimeout(parentCtx, types.DefaultMigrationTimeout)
+	defer cancel()
+
+	return store.Count(ctx)
+}
+
+func getDestinationCount(parentCtx context.Context, store updated.FlowStore) (int, error) {
```
connorgorman, line 121: its typically okay to overwrite the context so this param can be ctx

connorgorman (commented):
added some comments

migrator/migrations/m_172_to_m_173_network_flows_partition/migration.go
dashrews78: I did.  But pulling in the store and all that nonsense seemed like overkill so I just did `getClustersStmt := "select distinct id from clusters;"` in `getClusters`

connorgorman (approved):

PR #5137, opened 2023-03-07
Default periodic process pruning via heuristics to false

connorgorman:
## Description

This optimization is very expensive and based on some analysis, it is not particularly useful in many environments. Default it to off, but have a way to turn it back on if needed

## Checklist

## Testing Performed

CI


janisz (approved):

viswajithiii:
😢

PR #5138, opened 2023-03-07
ROX-15385: Clicking on multi-network cidr block nodes should show the side panel

sachaudh:
## Description

This PR fixes an issue where clicking on a multi-network cidr block node would not show the side panel. The issue happened because the `id` had a `/` in it that confused our URL state. The fix includes encoding/decoding the id

## Checklist

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
@@ -101,7 +101,10 @@ const TopologyComponent = ({
             const queryString = clearSimulationQuery(history.location.search);
             // if found, and it's not the logical grouping of all external sources, then trigger URL update
             if (newDetailId !== 'EXTERNAL') {
-                history.push(`${networkBasePathPF}/${newDetailType}/${newDetailId}${queryString}`);
+                const newURL = `${networkBasePathPF}/${newDetailType}/${encodeURIComponent(
+                    newDetailId
+                )}${queryString}`;
+                history.push(newURL);
             } else {
                 // otherwise, return to the graph-only state
                 history.push(`${networkBasePathPF}${queryString}`);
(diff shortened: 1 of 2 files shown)


ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
```diff
@@ -101,7 +101,10 @@ const TopologyComponent = ({
             const queryString = clearSimulationQuery(history.location.search);
             // if found, and it's not the logical grouping of all external sources, then trigger URL update
             if (newDetailId !== 'EXTERNAL') {
-                history.push(`${networkBasePathPF}/${newDetailType}/${newDetailId}${queryString}`);
+                const newURL = `${networkBasePathPF}/${newDetailType}/${encodeURIComponent(
```
vjwilson, line 104: I think this is fine for our use-case, but I was curious if this was robust.

Turns out there is an edge case where `encodeURIComponent` can throw, if you feed it a "lone surrogate" of a certain range of 2-character Unicode characters.
https://stackoverflow.com/questions/16868415/encodeuricomponent-throws-an-exception 

Really just sharing because I thought that was interesting--I don't think we would use those characters in CIDR block IDs.

vjwilson (approved):

ui/apps/platform/src/Containers/NetworkGraph/TopologyComponent.tsx
sachaudh: Not sure if we'd do anything crazy or if the user would purposefully add characters like that, but I agree we could make this a little better. The following were ideas I had about this:
1. Make sure we don't use an `id` that uses the `/` character 
2. Don't attempt to just use the `useParams` hook directly, but try to create another layer above it. This layer is feature-specific and takes care of getting and setting the URL params
Maybe a proper process for dealing with URL state (search, params, etc) would be a good discussion to have

PR #5141, opened 2023-03-07
[ROX-15811] Support scope context evaluation within select query parser

md2119:
## Checklist

## Testing Performed
Unit

### DIFF
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -623,7 +623,7 @@ func cloneAndCombine(q *v1.Query, scopeQ *v1.Query) *v1.Query {
 	if scopeQ == nil {
 		return q
 	}
-	
+
 	// Select, Group By, and Pagination must be set on the top-level query to be picked up by the query parser.
 	// Therefore, move them to the top-level query.
 
--- a/pkg/search/postgres/common.go
+++ b/pkg/search/postgres/common.go
@@ -612,11 +612,28 @@ func scopeContextToQuery(ctx context.Context, q *v1.Query) (*v1.Query, error) {
 	if scopeQ == nil {
 		return q, nil
 	}
+
+	return cloneAndCombine(q, scopeQ), nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/search/postgres/common.go
```diff
@@ -598,6 +604,30 @@ func standardizeSelectQueryAndPopulatePath(q *v1.Query, schema *walker.Schema, q
...
 
+func scopeContextToQuery(ctx context.Context, q *v1.Query) (*v1.Query, error) {
+	scopeQ, err := scoped.GetQueryForAllScopes(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if scopeQ == nil {
+		return q, nil
+	}
+	cloned := q.Clone()
```
dashrews78, line 615: I feel like this section of code where we make a copy, save off bits, remove bits, add them back in, warrants a comment.  or maybe a `ScopedClone` function that hides some of this.

pkg/search/postgres/common.go
```diff
@@ -546,11 +547,17 @@ func standardizeQueryAndPopulatePath(q *v1.Query, schema *walker.Schema, queryTy
...
 	nowForQuery := time.Now()
+
+	var err error
+	q, err = scopeContextToQuery(ctx, q)
+	if err != nil {
+		return nil, err
+	}
+
 	standardizeFieldNamesInQuery(q)
 	innerJoins, dbFields := getJoinsAndFields(schema, q)
```
charmik-redhat, line 560: Should we check if innerJoins and dbFields are nil/empty here? I think this could happen if selects and matchFields are unrelated to the schema being queried.

charmik-redhat (commented):
One comment. LGTM otherwise

pkg/search/postgres/common.go
md2119: We do not need to. The individual parsers below already tackle it.

charmik-redhat (approved):

PR #5145, opened 2023-03-08
ROX-14660: Fix overwrites to node scans from Node pipeline

jvdm:
## Description

Add an allow list to match Node's `OsImage` attributes and determine if they support full host-level node scanning. If yes:

1. Ignore the node during reprocessing.
2. Handle `SensorEvent_Node` by not enriching the node and skipping risk calculation.
3. Handle `SensorEvent_NodeInventory` by enriching the inventory and attaching a `Node.Scan`, leaving the other metadata intact.

The above ensures both pipelines won't overwrite each other data as long as they don't run concurrently. `SensorEvent_Node` should only update Node, but not `Node.Scan`, while `SensorEvent_NodeInventory` should only update `Node.Scan` but nothing else.

For simplicity, we will continue to allow these pipelines to perform reads before writes to the database to update only parts that are relevant for each. And we will ensure they don't run concurrently by using the same [worker queue when processing them](https://github.com/stackrox/stackrox/pull/5123).

To ensure `Node` is not touched by `SensorEvent_NodeInventory`, we keep `Node.LastUpdated` if already set on upsert. Tests were updated to account for the fact.

## Checklist
- [ ] ~Evaluated and added CHANGELOG entry if required~ N/A
- [ ] ~Determined and documented upgrade steps~ N/A
- [ ] ~Documented user-facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~ N/A

## Testing Performed

Integration tests were added to cover the scenario where `Node` is being upserted with `Node.Scan = nil`.

For manual testing, deploy to Openshift-4.12 (cluster `jvdm-argilo`). and:

1. Reduce reprocessing time: `kubectl -n stackrox set env deployment/central ROX_REPROCESSING_INTERVAL=1m`
1. Reduce node scanning time: `kubectl -n stackrox set env daemonset/collector "ROX_NODE_SCANNING_INTERVAL_DEVIATION=30s" "ROX_NODE_SCANNING_INTERVAL=1m" "ROX_NODE_SCANNING_INTERVAL_MAX_INITIAL_WAIT=0s"`
1. Two experiments:
   a. Run reprocessing and inventories at a high rate (~1m), and trigger node events by kicking Sensor.
   a. Disable inventories with `kubectl -n stackrox set env daemonset/collector "ROX_RHCOS_NODE_SCANNING=false"` and trigger node events.

### UI

image

### Logs

Reprocessing:

```
reprocessor: 2023/03/10 01:09:13.069413 reprocessor.go:420: Info: Successfully reprocessed 87/87 images
reprocessor: 2023/03/10 01:09:13.069609 reprocessor.go:421: Info: Resyncing deployments now that images have been reprocessed...
reprocessor: 2023/03/10 01:09:35.625355 reprocessor.go:273: Info: Found 6 nodes to scan
reprocessor: 2023/03/10 01:09:35.648246 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-master-2.c.srox-temp-dev-test.internal (id: c3d2395f-ed74-455f-b9f6-4507db02c779) does not n
reprocessor: 2023/03/10 01:09:35.650914 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-master-0.c.srox-temp-dev-test.internal (id: 1ce86640-8407-4e01-aed6-665da2362936) does not n
reprocessor: 2023/03/10 01:09:35.652714 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-master-1.c.srox-temp-dev-test.internal (id: 3c995dde-9890-48de-94e9-93bffdce2876) does not n
reprocessor: 2023/03/10 01:09:35.654843 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-worker-b-h7xp9.c.srox-temp-dev-test.internal (id: ea130298-20a1-4bbc-a7ac-04a00a0a21e3) does
reprocessor: 2023/03/10 01:09:35.655099 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-worker-c-6ps8s.c.srox-temp-dev-test.internal (id: 9105af0f-522f-4381-a0f2-39399263e52e) does
reprocessor: 2023/03/10 01:09:35.668693 reprocessor.go:445: Info: node secured-cluster-jvdm-argilo:jvdm-argilo-47bkl-worker-d-lkv4x.c.srox-temp-dev-test.internal (id: 3c73854f-9852-4de8-8a57-d5230c8cfde5) does
reprocessor: 2023/03/10 01:09:35.668803 reprocessor.go:302: Info: Successfully reprocessed 0/6 nodes
```

NodeInventory:

```
sensor/service/pipeline/nodeinventory: 2023/03/10 01:08:02.357826 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-worker-c-6ps8s.c.srox-temp-dev-test.internal' ID='9105af0f-522f-4381-a0f2-39399263e52e'
sensor/service/pipeline/nodeinventory: 2023/03/10 01:08:07.812502 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-master-2.c.srox-temp-dev-test.internal' ID='c3d2395f-ed74-455f-b9f6-4507db02c779'
sensor/service/pipeline/nodeinventory: 2023/03/10 01:11:23.420295 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-master-0.c.srox-temp-dev-test.internal' ID='1ce86640-8407-4e01-aed6-665da2362936'
sensor/service/pipeline/nodeinventory: 2023/03/10 01:11:55.890659 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-worker-b-h7xp9.c.srox-temp-dev-test.internal' ID='ea130298-20a1-4bbc-a7ac-04a00a0a21e3'
sensor/service/pipeline/nodeinventory: 2023/03/10 01:12:17.636049 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-master-1.c.srox-temp-dev-test.internal' ID='3c995dde-9890-48de-94e9-93bffdce2876'
sensor/service/pipeline/nodeinventory: 2023/03/10 01:12:22.904669 pipeline.go:65: Info: Received NodeInventory for Node name='jvdm-argilo-47bkl-worker-d-lkv4x.c.srox-temp-dev-test.internal' ID='3c73854f-9852-4de8-8a57-d5230c8cfde5'
```

## Multi-cluster test

In a multi-cluster setup, one with RHCOS scanning *enabled* the other with RHCOS *disabled*:

image


central/node/datastore/store/postgres/store.go
```diff
@@ -488,22 +488,20 @@ func (s *storeImpl) isUpdated(ctx context.Context, node *storage.Node) (bool, bo
...
 		obj.LastUpdated = iTime
 	}
-	metadataUpdated, scanUpdated, err := s.isUpdated(ctx, obj)
+	scanUpdated, err := s.isUpdated(ctx, obj)
 	if err != nil {
 		return err
 	}
-	if !metadataUpdated && !scanUpdated {
-		return nil
-	}
```
vikin91, line 506: Help me understand this - why we should still do the upsert if there are no changes in both: the metadata and scan?

central/reprocessor/reprocessor_unit_test.go
```diff
@@ -0,0 +1,95 @@
...
+			name: "when OS is RHCOS, nothing is done",
+			setUpMocks: func(t *testing.T, a *args, m *mocks) {
+				node := &storage.Node{
+					OsImage: "Red Hat Enterprise Linux CoreOS 412.86.202302091419-0 (Ootpa)",
+				}
+				m.nodes.EXPECT().GetNode(gomock.Any(), a.id).Return(node, true, nil)
+			},
+		},
+		{
+			name: "when OS is not RHCOS, nothing scanner is called and node is upserted",
```
vikin91, line 43:
```suggestion
			name: "when OS is not RHCOS, scanner is called and node is upserted",
```

central/reprocessor/reprocessor_unit_test.go
```diff
@@ -0,0 +1,95 @@
...
+					m.risk.EXPECT().CalculateRiskAndUpsertNode(gomock.Any()).DoAndReturn(func(node *storage.Node) error {
+						assert.Nil(t, node.LastUpdated, "expected node.LastUpdated = nil when CalculateRiskAndUpsertNode(node) is called")
+						return nil
+					}).Times(1),
+				)
+			},
+			want: true,
+		},
+		{
+			name: "when get node returns err, returns false",
```
vikin91, line 61:
```suggestion
			name: "when get node returns err, reprocessNode returns false",
```
or
```suggestion
			name: "false should be returned when get node returns err",
```

central/reprocessor/reprocessor_unit_test.go
```diff
@@ -0,0 +1,95 @@
...
+			setUpMocks: func(t *testing.T, a *args, m *mocks) {
+				m.nodes.EXPECT().GetNode(gomock.Any(), a.id).Times(1).Return(nil, false, nil)
+			},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
vikin91, line 78: Finish is optional here - see https://pkg.go.dev/github.com/golang/mock/gomock#NewController
We do not call finish in `central/sensor/service/pipeline/nodeinventory/pipeline_test.go`, so maybe let's pick one variant (calling it or not calling it) and apply globally.

central/sensor/service/pipeline/nodeinventory/pipeline_test.go
```diff
@@ -0,0 +1,100 @@
...
+		clusterID string
+		msg       *central.MsgFromSensor
+		injector  common.MessageInjector
+	}
+
+	tests := []struct {
+		name    string
+		mocks   mocks
+		args    args
+		wantErr string
```
vikin91, line 48: It looks like this could be a bool in this test, as the only thing we do with that is checking whether this is empty `tt.wantErr != ""`.

As another option, we could assert using:

```go 
s.ErrorContains(err, tt.wantErr)
```

vikin91 (commented):
This looks good and I found mostly nits, but the main point that I would like understand is why the changes to `isUpdated` are needed - i.e., why we no longer need to know whether the node metadata was updated?

central/node/datastore/store/postgres/store.go
jvdm: Previously, `metadataUpdated` was always set to `true`, meaning this if the statement was redundant.

central/reprocessor/reprocessor_unit_test.go
jvdm: Fixed at 751aa60dc2

central/reprocessor/reprocessor_unit_test.go
jvdm: Thanks! Quoting docs:

> New in go1.14+, if you are passing a *testing.T into NewController function you no longer need to call ctrl.Finish() in your test methods. 

Fixed at 751aa60dc2

central/node/datastore/store/postgres/store.go
vikin91: I see. Shouldn't then we change 

```go
	if !metadataUpdated && !scanUpdated {
		return nil
	}
```

into

```go
	if !scanUpdated {
		return nil
	}
```
?

central/sensor/service/pipeline/nodeinventory/pipeline_test.go
jvdm: At [central/sensor/service/pipeline/nodeinventory/pipeline_test.go:53](https://github.com/stackrox/stackrox/pull/5145/files/9765951442ead153bc3455ae5590ba18b7af0a3f#diff-323ec1a5481b6d118c6a4ba909e66baa7f3f7fc469525cfb4bcb564abefd2a80R53) we check the error's text.

vikin91: I see that we do set it on line 53, but the only place we read it is in https://github.com/stackrox/stackrox/pull/5145/files/9765951442ead153bc3455ae5590ba18b7af0a3f#diff-323ec1a5481b6d118c6a4ba909e66baa7f3f7fc469525cfb4bcb564abefd2a80R95 where we only compare it against `""`.

central/node/datastore/store/postgres/store.go
jvdm: Why?

vikin91: I see that the if-clause would never evaluate to true and we would never quit early here. However, I am thinking about the intentions of the original author of that code, and I think the following was meant: "if there were no changes to metadata and no changes to scan data, then quit early". By removing the entire if, we do keep the same behavior but we dismiss the original intention to quit early if there are no changes. Is there any value in quitting early `if !scanUpdated`?

vikin91 (approved):
It looks good to me! 
I am not fluent with that part of the test code, so I didn't do a thorough review of that part (but I really like the `TestStore_UpsertWithoutScan`), so maybe consider getting another review before merging? (@md2119 is the ideal choice 😄 )

central/node/datastore/store/postgres/store.go
jvdm: There is no value in quitting early `if !scanUpdate`. In fact, that is one of the cases we expect to occur when on `SensorEvent_Node`. The pipeline will update the Node with outdated (in fact `nil`) scan data. If we give up on that scenario, we prevent the update altogether. Unless there is a scenario in which both `last_updated` and `scan_time` are older, I am not aware of any.

central/reprocessor/reprocessor.go
```diff
@@ -433,22 +433,31 @@ func (l *loopImpl) reprocessImagesAndResyncDeployments(fetchOpt imageEnricher.Fe
...
 		return false
 	}
 	if !exists {
-		log.Warnf("node %q does not exist in the database. Skipping...", id)
+		log.Warnf("fetching node id %q from the database: node does not exist", id)
+		return false
+	}
+
+	if nodeEnricher.SupportsNodeScanning(node) {
+		log.Infof("node %s:%s (id: %s) does not need reprocessing (non-legacy node)",
```
md2119, line 445: Consider `cluster/node` convention which is followed in other places as well as UI.

central/reprocessor/reprocessor.go
```diff
@@ -433,22 +433,31 @@ func (l *loopImpl) reprocessImagesAndResyncDeployments(fetchOpt imageEnricher.Fe
...
 
 	err = l.nodeEnricher.EnrichNode(node)
 	if err != nil {
-		log.Errorf("error enriching node %s: %v", node.GetName(), err)
+		log.Errorf("enriching node %s:%s (id: %s): %v",
+			node.GetClusterName(), node.GetName(), id, err)
 		return false
 	}
-
+	node.LastUpdated = nil
```
md2119, line 457: We should avoid setting fields controlled by store elsewhere. Please take a look at my other comment about last updated.

central/reprocessor/reprocessor.go
md2119, line 445: What does non-legacy node mean? If we skip enrichment, how do we make ensure the snoozed state of vulns is updated?

central/node/datastore/store/postgres/store.go
vikin91: @md2119 scans can come now from two sources: nodes pipeline (aka V1 scan) & nodeInventory pipeline (aka V2 scan). The latter has richer set of components and vulnerabilities, but is reported every 4 hours. The Node pipeline has fewer scan data but is reported on every node update from Sensor. We do not want that I fresher V1 scan replaces older but still valid (and reacher) V2.

Is that what was meant here?

md2119: If that is a guarantee, can we move the onus from the low-level store to a separate reconciler? Both node pipelines feed into this reconciler. When central receives node scans from either of the pipelines, the reconciler forwards or drops the message based on what you stated above. Therefore, the reconciler becomes the only channel/entry point from the datastore perspective. This way we move the complexity out of the low-level store.

vikin91: @jvdm what do you think about that idea?

jvdm: @md2119:

> I am not following the discussion entirely. Why would the an outdated scan have scan time later than recent scan time?

The `SensorEvent_Node` is sent when there are updates in the Kubernetes Node resource. That event used to be the only node event. And we would use the metadata (by that, I mean orchestrator, kernel, etc.) from the Node to generate scan components.

With Host-Level Node Scanning, we will not use the metadata from the `SensorEvent_Node` to generate scan components. Instead, with Host-Level Node Scanning, we have a new event, `SensorEvent_NodeInventory`, that arrives at every configurable interval (default ~4hs) with the scan components. We still need to handle `SensorEvent_Node`, but we don't need to populate the `Node.Scan` anymore when handling that event. That is only done when handling `SensorEvent_NodeInventory`.

That's the scenario where `Node.Scan` would be "outdated": When handling `SensorEvent_Node`. When writing the node to the store when handling `SensorEvent_Node`, the scan data should be ignored and pulled from the DB (if there is one). On the other hand, the other event `SensorEvent_NodeInventory` should not touch the Node metadata.

central/node/datastore/datastore.go
```diff
@@ -103,3 +104,8 @@ func GetTestRocksBleveDataStore(t testing.TB, rocksengine *rocksdbBase.RocksDB,
 	nodeComponentRanker := ranking.NodeComponentRanker()
 	return New(dacky, keyFence, bleveIndex, riskStore, nodeRanker, nodeComponentRanker), nil
 }
+
+// NodeString returns a human-readable string representation of a node.
+func NodeString(node *storage.Node) string {
+	return fmt.Sprintf("%s:%s (id: %s)", node.GetClusterName(), node.GetName(), node.GetId())
+}
```
jvdm, line 111: I've added this here because my initial feeling is that this logic should be close to its model, and I could not find a better candidate. Please correct me if you know a better place.

jvdm:
I reverted the `LastUpdated` to allow the field. I intuitively assumed they should be independent, but I don't think it is worth discussing this considering how minimal the impact on customers is, and also @md2119 is advocating keeping the datastore "dump" and "decisionless", which is an agreed goal.

The options considered were, selected (c.):

(a.) Do a Get before calling Upsert.
(b.) Add "options" to Upsert APIs to change its behavior and either overwrite or pull fields before writing.
(c.) Check `Node.Scan == nil` and perform a Get inside the datastore (selected :heavy_check_mark:).

central/node/datastore/datastore.go
md2119: `pkg/node`

central/sensor/service/pipeline/nodes/pipeline.go
```diff
@@ -90,13 +90,24 @@ func (p *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
 		node.ClusterName = clusterName
 	}
 
+	if enricher.SupportsNodeScanning(node) {
+		// Call upsert without any components, only Node metadata. Upsert will read any
+		// components from the database before writing. Safe because NodeInventory and
+		// Node pipelines never run concurrently by the Sensor Event worker queues.
+		if err := p.nodeDatastore.UpsertNode(ctx, node); err != nil {
```
md2119, line 97: I want to confirm my thought about the `if` block:
You are skipping the enrichment and analysis for host-scannable nodes because the inventory pipeline is guaranteed to provide more accurate vuln insights.

If the above is true, it would be a good idea to include it in the comment.

central/sensor/service/pipeline/nodes/pipeline.go
```diff
@@ -90,13 +90,24 @@ func (p *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+			err = errors.Wrapf(err, "upsertting node %s", nodeDatastore.NodeString(node))
```
md2119, line 98: super minor nit: `upserting`

central/sensor/service/pipeline/nodes/pipeline.go
md2119: Additionally, I think you mentioned that you need to handle `SensorEvent_Node` for metadata but don't need to populate the `.scan` from this pipeline. How about setting it .`.scan` to `nil` here? That should guarantee the scan from this pipeline would never disturb scan in db because of the check `scan == nil` in store.

central/sensor/service/pipeline/nodeinventory/pipeline.go
```diff
@@ -56,49 +57,50 @@ func (p *pipelineImpl) Match(msg *central.MsgFromSensor) bool {
...
 
-	log.Debugf("NodeInventory for name='%s' has been scanned and contains %d results", ninv.GetNodeName(),
-		len(node.GetScan().GetComponents()))
-
-	// Here NodeInventory stops to matter. All data required for the DB and UI is in node.NodeScan already
-
-	if err := p.riskManager.CalculateRiskAndUpsertNode(node); err != nil {
-		err = errors.Wrapf(err, "upserting node %s:%s into datastore", node.GetClusterName(), node.GetName())
+	// Update the whole node in the database with the new and previous information.
+	err = p.riskManager.CalculateRiskAndUpsertNode(node)
```
md2119, line 98: By now, we have the correct metadata and scan because metadata is pulled from db, that is given by old pipeline, whereas the scan is provided by the inventory pipeline, correct?

central/node/datastore/store/postgres/store_test.go
```diff
@@ -73,26 +84,57 @@ func (s *NodesStoreSuite) TestStore() {
...
+	node.Scan = nil
+	s.NoError(store.Upsert(s.ctx, node))
+
+	newNode, exists, err := store.Get(s.ctx, node.GetId())
+	s.NoError(err)
+	s.True(exists)
+
+	// We expect only LastUpdated to have changed.
+	foundNode.LastUpdated = newNode.GetLastUpdated()
+	s.Equal(foundNode, newNode)
```
md2119, line 139: Check if nil scan overwrites scan in db.

md2119 (commented):
Dropped a couple of comments, otherwise lgtm. One comment overall, consider putting the pipeline changes behind the feature flag.

central/sensor/service/pipeline/nodeinventory/pipeline.go
jvdm: Correct.

central/node/datastore/store/postgres/store_test.go
md2119: Ignore, my oversight

central/sensor/service/pipeline/nodes/pipeline.go
```diff
@@ -90,13 +90,24 @@ func (p *pipelineImpl) Run(ctx context.Context, clusterID string, msg *central.M
...
+			log.Error(err)
+			return err
+		}
+		return nil
+	}
+
 	err = p.enricher.EnrichNode(node)
 	if err != nil {
-		log.Warnf("enriching node %s:%s: %v", node.GetClusterName(), node.GetName(), err)
+		log.Warnf("enriching node %s failed (but we ignored it): %v", nodeDatastore.NodeString(node), err)
```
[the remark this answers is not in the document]
jvdm: > I see that you would like to communicate more to the users

I want to clarify what this warning means for Customers, Support, and Engineers.

> The error and continued processing?

Yes, both. By ignoring the error, I continue the processing.

> How about something that informs the effect? e.g. risk analysis of the node maybe inaccurate?

Yeah, why not. But "maybe" is not helpful, to be honest. But we can be specific: "enriching node %s failed, the failure was ignored: vulnerability and risk information will not be updated."

central/sensor/service/pipeline/nodeinventory/pipeline.go
```diff
@@ -56,49 +57,50 @@ func (p *pipelineImpl) Match(msg *central.MsgFromSensor) bool {
...
+	if event.GetAction() != central.ResourceAction_UNSET_ACTION_RESOURCE {
+		log.Errorf("node inventory %s with unsupported action: %s", invStr, event.GetAction())
 		return nil
 	}
-
 	ninv = ninv.Clone()
 
-	// TODO(ROX-14484): Resolve the race between pipelines - Start of critical section
+	// Read the node from the database, if not found we fail.
 	node, found, err := p.nodeDatastore.GetNode(ctx, ninv.GetNodeId())
```
md2119, line 77: Here, if you are pulling the node only for the metadata, then you can use `GetNodeMetadata`. It is more performant because no cycles are spent on pulling scan data.

jvdm: But if enriching fails, later on, I will overwrite with an empty scan. I leave it as is.

jvdm: I will follow up on this in a separate PR. I want to add a single API to retrieve one single NodeMetadata instead of using GetMany().

jvdm: Also need to double-check what metadata brings in and see if we are not missing any key piece of information. And write tests around it.

md2119: Sounds like a reasonable follow-up action.

central/sensor/service/pipeline/nodes/pipeline.go
md2119: > Maybe. I don't think there is ambiguity here regarding the role of the two pipelines for host-scannable nodes, are you feeling strong about this?

It is a change in behavior with the launch of this feature, and not many people are aware about it. It is just to save time explaining to individual basis.

> .scan is already always nil. Or, do you mean explicitly setting it to nil, to guard the beavior? That is a good idea. Let me know if you meant something else.

Correct, I meant extra safety valve.

md2119 (approved):

PR #5148, opened 2023-03-08
ROX-15834: Stabilize Nodescan Feature

Maddosaurus:
## Description

This PR will improve stability and robustness of calls to the code that analyzes nodes.

⚠️ For increased protection against excessive node resources consumption, it requires merging https://github.com/stackrox/scanner/pull/1118

## Checklist

### N/A

## Testing Performed

- [x] Unit tests

### DIFF
--- a/pkg/env/node_scan.go
+++ b/pkg/env/node_scan.go
@@ -32,4 +32,7 @@ var (
 
 	// NodeInventoryContainerEnabled is used to tell compliance whether a connection to the node-inventory container should be attempted
 	NodeInventoryContainerEnabled = RegisterBooleanSetting("ROX_CALL_NODE_INVENTORY_ENABLED", true)
+
+	// NodeAnalysisDeadline is a time in which node analyzer should finish filesystem scanning and produce an inventory
+	NodeAnalysisDeadline = registerDurationSetting("ROX_NODE_SCANNING_DEADLINE", 15*time.Second)
 )


compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -65,6 +77,50 @@ func (n *NodeInventoryCollector) Scan(nodeName string) (*storage.NodeInventory,
 	return m, nil
 }
 
+// callAnalyzer isolates the call to Analyze to ensure we recover from any panic that might occur.
+// Analyze runs on a potentially unknown live file system, so it might be prone to unexpected behavior.
```
vikin91, line 81: Add something like: We isolate this function to protect the component against the analyzerFunction : (1) panicking, (2) taking too long to execute (and potentially consuming node resources).

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -65,6 +77,50 @@ func (n *NodeInventoryCollector) Scan(nodeName string) (*storage.NodeInventory,
...
+			close(c)
+		}(results)
+
+		// uncertifiedRHEL is set to false, as scans are only supported on RHCOS for now,
+		// which only exists in certified versions
+		c, err := n.analyzerFunction(nodeName, "/host/", nodes.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: true})
+		results <- analyzerResult{
+			components: c,
+			err:        err,
+		}
```
vikin91, line 110: I see, we have forgotten about one aspect. What we do now, we call the `analyzerFunction` and wait for the result for `analyzerCallDeadline` time. When we miss the deadline, we simply ignore the function's result and return an error. However, the function will still be running and can potentially consume the node resources. 

I think we should pass the context into the `analyzerFunction` (this includes also passing it to `nodes.Analyze`) and make it stop when the context is canceled. This would allow us to protect both:
- the ACS component running this (be it compliance or nodeScanner)
- the customer's node

jvdm: @Maddosaurus Maybe worth looking into https://github.com/stackrox/stackrox/blob/master/pkg/safe/run.go#L38

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -65,6 +77,56 @@ func (n *NodeInventoryCollector) Scan(nodeName string) (*storage.NodeInventory,
...
+					components: nil,
+					err:        fmt.Errorf("call to Analyzer function panicked: %v", r),
+				}:
+				case <-ctx.Done():
+					return
+				}
+
+			}
+			close(c)
+		}(results)
```
jvdm, line 111: @Maddosaurus Maybe worth looking into https://github.com/stackrox/stackrox/blob/master/pkg/safe/run.go#L38

compliance/collection/nodeinventorizer/nodeinventory.go
```diff
@@ -13,30 +16,39 @@ import (
...
 type NodeInventoryCollector struct {
+	analyzerFunction     nodeAnalyzer
+	analyzerCallDeadline time.Duration
+}
+
+// NewNodeInventoryCollector returns a ready to use instance of NodeInventoryCollector
+func NewNodeInventoryCollector() *NodeInventoryCollector {
+	return &NodeInventoryCollector{
+		analyzerFunction:     nodes.Analyze,
+		analyzerCallDeadline: 10 * time.Second,
```
jvdm, line 36: Have we considered making this configurable by an environment variable?

One potential reason is node scanning taking a little longer than we expected for reasons external to our control, but customers want the option to wait a little longer instead of failing.

compliance/collection/nodeinventorizer/nodeinventory.go
vikin91: Yes, we saw that. I am afraid we need some more customization to handle timeouts & panics together, but I will give it a try.

compliance/collection/nodeinventorizer/nodeinventory.go
vikin91: Covered in [31b82a7](https://github.com/stackrox/stackrox/pull/5148/commits/31b82a728cbb1134c9aa0305070fecaed2180812)

compliance/collection/nodeinventorizer/nodeinventory.go
vikin91: I took some parts of that pkg and refactored a bit. Unfortunately, I couldn't reuse it 1:1 as the function we need here was unexported.

compliance/collection/nodeinventorizer/nodeinventory.go
vikin91: To address this issue, we first need to add context support to `nodes.Analyze` - here is a PR for this: https://github.com/stackrox/scanner/pull/1118

vikin91:
ups, rebase went wrong... fixing

pkg/env/node_scan.go
```diff
@@ -23,4 +23,7 @@ var (
 
 	// NodeScanningMaxBackoff is the upper boundary of backoff. Defaults to 5m in seconds, being 50% of Kubernetes restart policy stability timer.
 	NodeScanningMaxBackoff = registerDurationSetting("ROX_NODE_SCANNING_MAX_BACKOFF", 300*time.Second)
+
+	// NodeAnalysisDeadline is a time in which node analyzer should finish filesystem scanning and produce an inventory
+	NodeAnalysisDeadline = registerDurationSetting("ROX_NODE_SCAN_DEADLINE", 15*time.Second)
```
Maddosaurus, line 28: I would be in favour of unifying the env var names. For the caching & backoff I also went with node scanning as introduced by Victor.
I'm open to using node scan as well, but then I'd recommend a rebase on master and changing all env vars including caching & backoff at the same time
```suggestion
	NodeScanningDeadline = registerDurationSetting("ROX_NODE_SCANNING_DEADLINE", 15*time.Second)
```

vikin91:
Closing this as the respective parts need to be added now in the node-inventory container in the scanner repo

PR #5149, opened 2023-03-08
ROX-15812: Propagate `$POD_SECURITY_POLICIES` to `roxctl` when installing sensor.

porridge:
## Description

Looks like this was missed in https://github.com/stackrox/stackrox/pull/3230

## Checklist
- [ ] Investigated and inspected CI test results
- [x] ~Unit test and regression tests added~
- [x] ~Evaluated and added CHANGELOG entry if required~
- [x] ~Determined and documented upgrade steps~
- [x] ~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~

## Testing Performed

CI should be sufficient. Let me figure out how to run this on recent k8s...
/test help

PR #5150, opened 2023-03-08
dummy

porridge:
To check whether #5149 worked...

### DIFF
--- a/qa-tests-backend/Makefile
+++ b/qa-tests-backend/Makefile
@@ -1,4 +1,5 @@
 .PHONY: all
+
 all: style test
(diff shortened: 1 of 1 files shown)


porridge:
```
INFO:	Deployment bundle does not include PodSecurityPolicies (PSPs).
INFO:	This is incompatible with pre-v1.25 Kubernetes installations having the PodSecurityPolicy Admission Controller plugin enabled.
INFO:	Use --enable-pod-security-policies if PodSecurityPolicies are required for your Kubernetes environment.
```

Looks good.

PR #5155, opened 2023-03-08
ROX-15802: Close Postgres on termination

connorgorman:
## Description

Close Postgres on Central shutdown

## Checklist

## Testing Performed

Will try with sigterms once built

### DIFF
--- a/central/globaldb/close.go
+++ b/central/globaldb/close.go
@@ -10,4 +10,7 @@ func Close() {
 	if rocksDB != nil {
 		rocksDB.Close()
 	}
+	if postgresDB != nil {
+		postgresDB.Close()
+	}
 }


central/globaldb/close.go
```diff
@@ -10,4 +10,7 @@ func Close() {
 	if rocksDB != nil {
 		rocksDB.Close()
 	}
+	if postgresDB != nil {
+		postgresDB.Close()
```
dashrews78, line 14: I was reading and had the concern that there is the potential that this call could block until the inflight queries are complete.  Should we send a deadline context the the overarching function so we don't wait forever?  I think it would be a rare problem, but it is possible:
https://github.com/jackc/pgx/issues/802
https://stackoverflow.com/questions/71588402/how-to-force-closing-pgxpool-pool-while-queries-are-still-running

connorgorman: @dashrews78 So based on my read of those two links, basically just having the application die is the method in the worst case which is currently set to 30s by k8s

connorgorman: We don't have access to a global context for all running contexts so there is no way to send a cancel

dashrews78: right.  I wasn't aware of the 30s k8s setting.  That should be more than sufficient.

dashrews78 (approved):

PR #5168, opened 2023-03-09
Simplify retries

janisz:
## Description

`SETUP_FEATURE_CLEANUP` guarantees that test retry will be followed by `cleanup()` and `setup()` with this we don't need our custom code to detect if test run is retry or not. We can simplify our code and move retry logic from helper to base specification as it's only used there.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
```diff
@@ -142,7 +141,7 @@ class AdmissionControllerTest extends BaseSpecification {
 
         assert ClusterService.updateAdmissionController(ac)
         // Maximum time to wait for propagation to sensor
-        Helpers.sleepWithRetryBackoff(5000)
+        sleep(5000)
```
porridge, line 144: I think the goal of `sleepWithRetryBackoff` was to sleep longer on each test retry?
Perhaps we need to evaluate whether getting rid of this is not going to increase the flake rate significantly.

Food for thought: it would be very interesting to see which tests are being retried, how many times and how often. Another use case for the test accounting epic?

qa-tests-backend/src/test/groovy/BaseSpecification.groovy
```diff
@@ -28,14 +30,15 @@ import util.Env
...
+import org.junit.AssumptionViolatedException
 import org.junit.Rule
 import org.junit.rules.TestName
 import org.junit.rules.Timeout
 import spock.lang.Retry
 import spock.lang.Shared
 import spock.lang.Specification
 
-@Retry(condition = { Helpers.determineRetry(failure) })
+@Retry(condition = { Object failure -> determineRetry(failure as Throwable) }, mode = SETUP_FEATURE_CLEANUP)
```
porridge, line 41: Does this mean that any common content of `cleanup:` sections within the test cases can now be moved to the `cleanup()` method?

porridge (commented):
Judging from test results, it seems like groovy is confused about which `determineRetry` you meant? :thinking:

qa-tests-backend/src/test/groovy/BaseSpecification.groovy
janisz: Yep, I think that `cleanup:` block should be used only for specific test cleanup all common things should go in `cleanup()`

PR #5177, opened 2023-03-09
ROX-15434 - [MVP] enable on-prem registry scanning

dcaravel:
## Description
Enables scanning of all images via the secured cluster

A new environment variable,`ROX_FORCE_LOCAL_IMAGE_SCANNING`, was added that will cause Sensor to treat images as if they are accessible by only the secured cluster - image analysis will be performed via the local scanner-slim for any observed deployments, vuln matching is still sent to Central/Scanner

Most dockercfg/dockerconfig secrets will be captured in the existing registry store (in addition to the OCP internal registry secrets)

The enrichment and admission control scanning workflows will leverage the stored secrets to attempt image analysis via the local scanner.  If no stored secrets are found, the workflows assume the registry requires no auth and attempt analysis.

(tech debt) Sensor will now send `EnrichLocalImageInternal` requests to central when local scanning failures occur so that central can store and show the failures.

Scanning workflows via `roxctl`, image watches, Jenkins plugins, etc. (non-sensor) were not modified as part of this MVP and will continue to be processed via Central Services.
image

## Checklist

For this MVP general docs will not be updated, post-MVP docs will be updated. No changes to upgrade steps.

## Testing Performed
Unit tests added where applicable

Testing setup involved creating two clusters:
- `ClusterA` = GKE Central Services (installed via Helm)
- `ClusterB` = OCP Secured Cluster Services (installed via Helm w/ env `ROX_FORCE_LOCAL_IMAGE_SCANNING` = true)

Scanner logs on both clusters were followed while new pods were created on the `ClusterB`. Confirmed via following logs that only scanner-slim was moving / scanning images

A instance of [JFrog Container Registry (JCR)](https://www.jfrog.com/confluence/display/JCR) was installed on ClusterB, images were deployed from this registry using the Artifactory service cluster internal IP to ensure ClusterA would not have access. Observed success in scanning with the environment variable set to true, and failure when environment variable set to false.

## Testing Observations
While feature was turned on observed the following behaviors that need more attention / troubleshooting
- [x] image from new pods/deploys are scanned multiple times and appears twice in central after a single deployment (shows as one active and one in-active)
  - Fixed, caused by using other ids (specifically manifest ID) in place of image ID when image ID not avail. Forced use of image ID in scan requests
- [X] failed deployments (ie: due to `ImagePullBackOff`) trigger a sensor deployment event and a scan - possibly BAU
  - Testing with current versions of ACS - this is normal behavior
- [X] are many upserts into the registry_store, these may be excessive
  - Normal behavior - somewhat reduced by new suppression of storing secrets with service account annotation
- [x] OCP images `ocp-v4.0-art-dev` fail to be scanned by default
  - OCP images required usage of a special 'global pull' secret to access
- [X] scan failures are not reported to central
  - Fixed, modified sensor to always send Enrich request even during failures, modified central to save to DB if not already when errors occur.


sensor/common/image/service_impl.go
```diff
@@ -85,6 +109,32 @@ func (s *serviceImpl) GetImage(ctx context.Context, req *sensor.GetImageRequest)
 	}, nil
 }
 
+// createNoAuthDockerRegistry creates a new registry image integration of type docker that
+// has no credentials
+//
+// TODO: temporary until image integrations from central are used instead
+func createNoAuthDockerRegistry(ctx context.Context, registry string) (registryTypes.Registry, error) {
```
RTann, line 116: what is the difference between this and just using `UpsertNoAuthRegistry`?

sensor/common/registry/registry_store_test.go
```diff
@@ -62,3 +73,83 @@ func TestRegistryStore_same_namespace(t *testing.T) {
...
+
+	reg, err := regStore.UpsertNoAuthRegistry(ctx, "fake-namespace", fakeImgName)
+	require.NoError(t, err)
+	assert.Equal(t, reg.Name(), fakeImgName.GetRegistry())
+	assert.Empty(t, reg.Config().Username)
+	assert.Empty(t, reg.Config().Password)
+}
+
+func TestRegistryStore_GetFirstRegistryForImage(t *testing.T) {
+	logging.SetGlobalLogLevel(zapcore.DebugLevel)
```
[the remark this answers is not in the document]
dcaravel: good catch, I will remove this entirely, that was more for local testing

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -246,11 +247,14 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 
 	sensorEvents := make([]*central.SensorEvent, 0, len(dockerConfig)+1)
 	registries := make([]*storage.ImagePullSecret_Registry, 0, len(dockerConfig))
+
+	serviceAcctName := secret.GetAnnotations()[saAnnotation]
+	hasServiceAcctNameAnnotation := len(serviceAcctName) > 0
```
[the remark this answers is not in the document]
dcaravel: I like that! I had wanted to ensure an 'empty value' annotation was considered, but not seeing that occur for the default SAs, will make the change

sensor/common/image/service_impl.go
dcaravel: `UpsertNoAuthRegistry` requires a namespace to upsert into the registry store - this code is executed by the admission controller workflow that currently does not provide namespace (or any deployment) details. 

If preferred I can move the creation of the no auth registry to a utility method and use it in both places, my thinking was since this is temporary not to over complicate it. Thoughts?

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -276,6 +280,25 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 					newIntegrationSet.Add(ii.GetId())
 				}
 			}
+
+			if env.ForceLocalImageScanning.BooleanSetting() && !hasServiceAcctNameAnnotation {
+				// Store registry secrets to enable downstream scanning of all images
+				//
+				// Ignore secrets with the service-account.name annotation, each auto-generated secret for an OCP service
```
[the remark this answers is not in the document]
dcaravel: I looked into this and wasn't able to find a reference regarding conditions by which the annotation was added, outside of [this](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount) which doesn't apply to dockerconfig type secrets. Are you aware of alternate usages?

This is also temporary until image integrations are used, but given our discussion thinking we may need to change the logic for the long term given the history of autogenerated entries. For post-MVP believe we should use a combination of image integrations (preferred) and dockerconfig secrets (fallback) in the order listed.

sensor/common/registry/registry_store_test.go
```diff
@@ -6,8 +6,19 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
+)
+
+var (
+	fakeImgName = &storage.ImageName{
+		Registry: "quay.io",
+		Remote:   "rhacs-eng/sandbox",
+		Tag:      "noexist:1.2.0",
```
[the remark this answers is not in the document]
dcaravel: nice catch! going to change the tag to just `noexist`, the image is named `sandbox` in that quay repo, so the `tag` was malformed regardless

pkg/env/sensor.go
```diff
@@ -20,6 +20,10 @@ var (
 	// LocalImageScanningEnabled is used to specify if Sensor should attempt to scan images via a local Scanner.
 	LocalImageScanningEnabled = RegisterBooleanSetting("ROX_LOCAL_IMAGE_SCANNING_ENABLED", false)
 
+	// ForceLocalImageScanning expands the scope of `LocalImageScanningEnabled` to have sensor analyze all images
+	// using the local scanner instead of only images from OCP internal registries
+	ForceLocalImageScanning = RegisterBooleanSetting("ROX_FORCE_LOCAL_IMAGE_SCANNING", false)
```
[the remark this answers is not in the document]
dcaravel: Updated CHANGELOG

This being MVP / tech preview - I assume that means by default we do not enable it.  Both the helm charts and the operator allow for environment variables to be set - which would trigger enablement of the feature for those users that would like to try it out, ie:

```
helm install ...
    --set customize.envVars.ROX_FORCE_LOCAL_IMAGE_SCANNING="true"
```

sensor/common/registry/registry_store.go
```diff
@@ -99,6 +100,16 @@ func (rs *Store) UpsertRegistry(ctx context.Context, namespace, registry string,
 	return nil
 }
 
+// UpsertNoAuthRegistry will store a new registry with no user/pass and return the created registry integration
+func (rs *Store) UpsertNoAuthRegistry(ctx context.Context, namespace string, imgName *storage.ImageName) (registryTypes.Registry, error) {
```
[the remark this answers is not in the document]
dcaravel: Sounds good, I modified the signature to `UpsertRegistry` - which had a slight cascade effect.  In the end believe this will be more efficient since the Upsert no longer has to 'search' for the registry it just created and can just return it

sensor/kubernetes/listener/resources/secrets.go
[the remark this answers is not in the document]
dcaravel: back-tracking on this, after testing believe the code will be more robust testing for the empty value

sensor/common/image/service_impl.go
dcaravel: `UpsertNoAuthRegistry` has been removed in latest commits

central/image/service/service_impl.go
```diff
@@ -382,6 +382,13 @@ func (s *serviceImpl) EnrichLocalImageInternal(ctx context.Context, request *v1.
 
 	defer s.internalScanSemaphore.Release(1)
 
+	hasErrors := false
+	if request.Error != "" {
+		hasErrors = true
+		log.Infof("received image enrichment request with errors %q: %v", request.GetImageName().GetFullName(), request.GetError())
```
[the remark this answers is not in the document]
RTann: similarly, would a warning be good or could that blow up logs?

central/image/service/service_impl.go
```diff
@@ -414,31 +421,43 @@ func (s *serviceImpl) EnrichLocalImageInternal(ctx context.Context, request *v1.
 		if exists && !forceScanUpdate && !forceSigVerificationUpdate {
 			return internalScanRespFromImage(existingImg), nil
 		}
+
+		if exists {
```
RTann, line 425: nit: maybe `existingImg, exists, err := s.datastore.GetImage(ctx, imgID)` should be something like:

```
var existingImg <whatever>
var err error
existingImg, imgExists, err = s.datastore.GetImage(ctx, imgID)
```

This way, we don't need to add a new var to mirror the one only exposed inside the `if`

central/image/service/service_impl.go
```diff
@@ -414,31 +421,43 @@ func (s *serviceImpl) EnrichLocalImageInternal(ctx context.Context, request *v1.
...
+		}
 	}
 
 	img := &storage.Image{
-		Id:             imgID,
-		Name:           request.GetImageName(),
+		Id:   imgID,
+		Name: request.GetImageName(),
+		// 'Names' must be populated to enable cache hits in central AND sensor
+		Names:          []*storage.ImageName{request.GetImageName()},
```
RTann, line 434: I'm not as familiar with this field. Does this deprecate `Name` at all?

central/image/service/service_impl.go
```diff
@@ -382,6 +382,13 @@ func (s *serviceImpl) EnrichLocalImageInternal(ctx context.Context, request *v1.
 
 	defer s.internalScanSemaphore.Release(1)
 
+	hasErrors := false
```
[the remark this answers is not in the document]
RTann: can you also leave a comment saying why else this is used? As in, if there are errors in the scan, then why do we still continue for a few lines? I see comments below, but I think it'd be helpful to know why this variable exists up here, too

proto/internalapi/sensor/image_iservice.proto
```diff
@@ -8,8 +8,9 @@ import "storage/image.proto";
 package sensor;
 
 message GetImageRequest {
-    storage.ContainerImage image = 1;
-    bool scan_inline             = 2;
+    storage.ContainerImage  image = 1;
+    bool scan_inline              = 2;
+    storage.Deployment deployment = 3;
```
RTann, line 13: is the entire deployment needed or just specific parts? This is a pretty large proto, potentially, and the image (tag 1) is already part of it, so it's as if we are sending that bit of information twice

sensor/common/image/service_impl.go
```diff
@@ -78,9 +86,19 @@ func (s *serviceImpl) GetImage(ctx context.Context, req *sensor.GetImageRequest)
...
 	}
 
-	img, err := s.localScan.EnrichLocalImage(ctx, s.centralClient, req.GetImage())
+	var err error
+	var img *storage.Image
+	if req.GetImage().GetIsClusterLocal() {
+		img, err = s.localScan.EnrichLocalImage(ctx, s.centralClient, req.GetImage())
+	} else {
+		// ForceLocalImageScanning must be true
+		img, err = s.localScan.EnrichLocalImageInNamespace(ctx, s.centralClient, req.GetImage(), req.GetDeployment().GetNamespace())
```
RTann, line 95: From this call, I think we just need the deployment's namespace, so let's update the proto to just give the namespace instead of the deployment struct

sensor/common/image/service_impl.go
```diff
@@ -58,14 +64,16 @@ func (s *serviceImpl) GetImage(ctx context.Context, req *sensor.GetImageRequest)
 		}
 	}
 
+	log.Debugf("scan request from admission control: %+v", req)
+
 	// Note: The Admission Controller does NOT know if the image is cluster-local,
 	// so we determine it here.
 	// If Sensor's registry store has an entry for the given image's registry,
 	// it is considered cluster-local.
 	req.Image.IsClusterLocal = s.registryStore.HasRegistryForImage(req.GetImage().GetName())
```
RTann, line 73: Can you comment here that this "clsuter-local" stuff is really just about OpenShift Internal registry at this time

sensor/common/image/service_impl.go
```diff
@@ -78,9 +86,19 @@ func (s *serviceImpl) GetImage(ctx context.Context, req *sensor.GetImageRequest)
...
+		img, err = s.localScan.EnrichLocalImage(ctx, s.centralClient, req.GetImage())
+	} else {
+		// ForceLocalImageScanning must be true
+		img, err = s.localScan.EnrichLocalImageInNamespace(ctx, s.centralClient, req.GetImage(), req.GetDeployment().GetNamespace())
+	}
+
 	if err != nil {
-		return nil, errors.Wrap(err, "scanning image via local scanner")
+		err = errors.Wrap(err, "scanning image via local scanner")
+		log.Error(err)
```
RTann, line 100: is this to keep or just debugging?

sensor/common/registry/registry_store.go
```diff
@@ -26,6 +27,9 @@ type Store struct {
 	// store maps a namespace to the names of registries accessible from within the namespace.
 	store map[string]registries.Set
 
+	// globalRegistries holds registries global to openshift (applies to multiple namespaces)
```
RTann, line 30: I don't fully follow the comment. Only OpenShift? Is this the "global pull secret" stuff?

also nit: OpenShift

sensor/common/registry/registry_store.go
```diff
@@ -89,12 +86,27 @@ func (rs *Store) UpsertRegistry(ctx context.Context, namespace, registry string,
...
+	// remove http/https prefixes from registry, matching may fail otherwise, the created registry.url will have
+	// the appropriate prefix
+	registry = urlfmt.TrimHTTPPrefixes(registry)
+	err = regs.UpdateImageIntegration(createImageIntegration(registry, dce, secure))
 	if err != nil {
 		return errors.Wrapf(err, "updating registry store with registry %q", registry)
 	}
 
-	log.Debugf("Upserted registry %q for namespace %q into store", registry, namespace)
+	log.Debugf("upserted registry %q for namespace %q into store", registry, namespace)
```
RTann, line 109: I think for logging, uppercase is fine

sensor/common/scan/scan.go
```diff
@@ -3,62 +3,82 @@ package scan
...
+
+	redhatRegistryEndpoint = "registry.redhat.io"
+
 	log = logging.LoggerForModule()
 )
 
-// LocalScan wraps the functions required in EnrichLocalImage. This allows us to inject different values for testing purposes
+// LocalScan wraps the functions required for enriching local images. This allows us to inject different values for testing purposes
 type LocalScan struct {
-	// Used for testing purposes only to not require setting up registry / scanner.
```
RTann, line 30: how come this was deleted?

dcaravel: I felt it was misleading because its not 'only' used for testing purposes, its used during normal operation as well. The note seems sufficient to indicate tests are impacted by the assignment.

sensor/common/image/service_impl.go
dcaravel: To keep, without it only the admission controller logs would show the error. It was helpful for debugging as well.

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -283,6 +287,24 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 					newIntegrationSet.Add(ii.GetId())
 				}
 			}
+
+			if env.ForceLocalImageScanning.BooleanSetting() {
```
RTann, line 291: how come this is only done when `saName == ""`?

central/image/service/service_impl.go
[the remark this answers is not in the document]
dcaravel: > debug or do we want to keep this around?

Intent was to keep it, will help answer 'why are my scans failing' - today support will search through central logs to find why central initiated scans are failing, this allows them to continue that practice.  Otherwise they would have to know to go to the sensor logs on cluster X to determine why.

Its possible that it could blow up the logs if there are lots of failures yes. In my experience thats better then not having logs and guessing.

> similarly, would a warning be good or could that blow up logs?

Not sure I follow, is the suggestion to change the log level to `warn`? That would make it harder to suppress (if thats the intention) than if at `info` as `warn` is higher level. I would prefer to have more logs than less, is there a reason to log less that I may be missing?

sensor/common/registry/registry_store.go
```diff
@@ -68,15 +72,8 @@ func (rs *Store) getRegistries(namespace string) registries.Set {
...
-// UpsertRegistry upserts the given registry with the given credentials in the given namespace into the store.
-func (rs *Store) UpsertRegistry(ctx context.Context, namespace, registry string, dce config.DockerConfigEntry) error {
-	secure, err := rs.checkTLS(ctx, registry)
-	if err != nil {
-		return errors.Wrapf(err, "unable to check TLS for registry %q", registry)
-	}
-
-	regs := rs.getRegistries(namespace)
-	err = regs.UpdateImageIntegration(&storage.ImageIntegration{
+func createImageIntegration(registry string, dce config.DockerConfigEntry, secure bool) *storage.ImageIntegration {
```
RTann, line 75: I think you've mentioned making all these as Docker integrations does not always works (ex: ECR integrations need periodic token updates). Can we add a comment talking about this?

central/image/service/service_impl.go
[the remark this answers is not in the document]
dcaravel: yup! will add

central/image/service/service_impl.go
dcaravel: I like that, will be cleaner, will make the change

sensor/common/registry/registry_store.go
```diff
@@ -122,12 +153,53 @@ func (rs *Store) GetRegistryForImage(image *storage.ImageName) (registryTypes.Re
...
 }
 
-// HasRegistryForImage returns true when the registry store has the registry
-// for the given image.
-func (rs *Store) HasRegistryForImage(image *storage.ImageName) bool {
-	reg, err := rs.GetRegistryForImage(image)
-	return reg != nil && err == nil
+// getOrInitGlobalRegistries will return global registries if existing, or create a new set
+// mimics `getRegistries`
+func (rs *Store) getOrInitGlobalRegistries() registries.Set {
```
RTann, line 161: I think https://pkg.go.dev/sync#Once.Do may be good here

sensor/common/registry/registry_store.go
```diff
@@ -122,12 +153,53 @@ func (rs *Store) GetRegistryForImage(image *storage.ImageName) (registryTypes.Re
...
+
+	if rs.globalRegistries == nil {
+		rs.globalRegistries = registries.NewSet(rs.factory)
+	}
+
+	return rs.globalRegistries
+}
+
+// UpsertGlobalRegistry will store a new registry with the given credentials into the global registry store
+func (rs *Store) UpsertGlobalRegistry(ctx context.Context, registry string, dce config.DockerConfigEntry) error {
```
RTann, line 173: When would I use this vs `UpsertRegistry`?

central/image/service/service_impl.go
dcaravel: The `Names` field does have a comment indicating that long-term it may replace `Name` (in `proto/storage/image.proto`). I don't know the history or if we're tracking that change elsewhere.

Without setting this the cache is essentially skipped (we store images in the cache but never pull them out).

proto/internalapi/sensor/image_iservice.proto
dcaravel: Immediately whats needed is the namespace, in the next release I plan to use pull secrets as well, past that unknown.  If its preferred we only send the fields that are needed at a given time I can make that change. Lemme know.

sensor/common/image/service_impl.go
dcaravel: will do

sensor/common/image/service_impl.go
dcaravel: will do

sensor/common/registry/registry_store.go
dcaravel: Correct, this is the global pull secret stuff that can apply to any namespace in OpenShift - and is utilized regardless of namespace, specified pull secrets, or service accounts.

sensor/common/registry/registry_store.go
dcaravel: interesting, ok i'll change it back, old habit I suppose. (in my past if a message didn't end in with a `.` or started with a proper noun, then lowercase was preferred) - throughout the codebase I don't see a convention one way or the other, some upper, some lower.

sensor/common/detector/enricher.go
```diff
@@ -114,6 +124,13 @@ outer:
 				}
 			}
 
+			// If cluster local scan is rate-limited, backoff and try again
+			if errors.Is(err, scan.ErrTooManyParallelScans) {
```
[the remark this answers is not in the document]
dcaravel: The section above can only be triggered based on specifics in the GRPC response to the Central ScanImageInternal request. The latter is from the semaphore local to sensor.

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -250,6 +253,7 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 	registries := make([]*storage.ImagePullSecret_Registry, 0, len(dockerConfig))
 
 	saName := secret.GetAnnotations()[saAnnotation]
+	isGlobalPullSecret := secret.GetNamespace() == openshiftConfigNamespace && secret.GetName() == openshiftConfigPullSecret
```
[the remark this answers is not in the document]
dcaravel: good call, I like that better too

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -283,6 +287,24 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 					newIntegrationSet.Add(ii.GetId())
 				}
 			}
+
+			if env.ForceLocalImageScanning.BooleanSetting() {
+				// Store registry secrets to enable downstream scanning of all images
+
+				// TODO: a namespace may contain multiple .dockerconfig* secrets for the same registry (not handled
```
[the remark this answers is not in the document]
dcaravel: ok, will update it

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-15102: new `public_config.telemetry` boolean property of the `/v1/config`
   endpoint request that allows for querying the state, enabling or disabling the
   configured telemetry collection.
+- ROX-15434: new `ROX_FORCE_LOCAL_IMAGE_SCANNING` env var added to sensor which forces all images observed by sensor to be analyzed by the local scanner (OCP only)
```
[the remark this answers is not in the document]
dcaravel: I'll have to think about this, my first thought is no - any image could be scanned (even the OCP platform images) if an image integration was created with creds that had appropriate access.

sensor/kubernetes/listener/resources/secrets.go
dcaravel: I believe the comments that exist already describe that to an extent, perhaps I can be more verbose. TL;DR; thats the only part that changes with this new feature, storing builder, deployer, etc. secrets will overwrite the entries inserted by the 'defaultSA' block above. So it doens't make sense - the TODO: comment describes the limitation and how it will be changed in the future (with the ticket you had me add as well)

sensor/common/registry/registry_store.go
dcaravel: I haven't proven that yet, thats a theory at this point - if I do prove it I'll update the comment accordingly.

sensor/common/registry/registry_store.go
dcaravel: I was mirroring the existing logic from the `getRegistries` method in the same module - it's also protected by a lock - is there a reason why once.Do is preferred? Assuming we'd want to change the existing logic as well if so.

sensor/common/registry/registry_store.go
dcaravel: When the registry being inserted is not namespace bound, and can apply to any workload - the OpenShift global pull secret is the only scenario i'm aware of that this is needed for

sensor/common/registry/registry_store.go
RTann: This is a little bit different. `getRegistries` adds new entries to a map while `getOrInitGlobalRegistries` creates a new set once, lazily. I think using `sync.Once` makes sense here. I (believe) it is more efficient than unconditionally locking/unlocking, and works for this use-case. I don't think it would work for the `getRegistries` use-case because we need to create a new registry set for each key, instead of just once, total.

Another option is to just initialize the "global" registries set in the constructor instead of lazily to remove the need for using any kind of synchronization primitives for this structure.

Also, I just realized this is using the same mutex as used for the "namespaced" registries. I think they should use different mutexes since they are unrelated data structures

sensor/common/registry/registry_store.go
RTann: makes sense. Similar question for `GetGlobalRegistryForImage`. How would a user know when the registry they are looking for is global vs namespaced?

sensor/common/registry/registry_store.go
dcaravel: sounds good, i'll make the change to once and separate the locks

sensor/common/registry/registry_store.go
dcaravel: I believe that depends on the use case / usage - in this case the only usage is in `sensor/common/scan/scan.go` - any matching global registry is added to the 'list' of registries to try

sensor/kubernetes/listener/resources/secrets.go
dcaravel: Added more comments

proto/internalapi/sensor/image_iservice.proto
dcaravel: Changed to only pass namespace

sensor/common/registry/registry_store.go
dcaravel: I updated the comment to be more descriptive

RTann (approved):

central/image/service/service_impl.go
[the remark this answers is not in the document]
dcaravel: will change to warn

proto/internalapi/sensor/image_iservice.proto
```diff
@@ -8,8 +8,9 @@ import "storage/image.proto";
 package sensor;
 
 message GetImageRequest {
-    storage.ContainerImage image = 1;
-    bool scan_inline             = 2;
+    storage.ContainerImage  image = 1;
```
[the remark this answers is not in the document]
dcaravel: fixed

pkg/registries/rhel/rhel.go
```diff
@@ -14,6 +14,10 @@ const (
 
 var (
 	log = logging.LoggerForModule()
+
+	// RedHatRegistryEndpoints represents endpoints for RHEL registries that should
+	// use this registry implementation (Metadata invocations may fail otherwise)
+	RedHatRegistryEndpoints = map[string]bool{"registry.redhat.io": true}
```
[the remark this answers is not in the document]
dcaravel: fixed

central/image/service/service_impl.go
```diff
@@ -382,12 +382,23 @@ func (s *serviceImpl) EnrichLocalImageInternal(ctx context.Context, request *v1.
...
+	if request.Error != "" {
+		// If errors occurred we continue processing so that the failed image scan may be saved in
+		// the central datastore. Without this users would not have an indication that scans from
+		// secured clusters are failing
+		hasErrors = true
+		log.Infof("received image enrichment request with errors %q: %v", request.GetImageName().GetFullName(), request.GetError())
+	}
+
+	var imgExists bool
+	var existingImg *storage.Image
```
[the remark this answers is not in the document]
dcaravel: fixed

central/image/service/service_impl.go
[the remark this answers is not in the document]
dcaravel: fixed

jvdm (commented):
I left some comments asking for clarification. No blockers.

CHANGELOG.md
```diff
@@ -12,9 +12,10 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-15102: new `public_config.telemetry` boolean property of the `/v1/config`
   endpoint request that allows for querying the state, enabling or disabling the
   configured telemetry collection.
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
dcaravel: yes it is, merge conflict - will fix

sensor/kubernetes/listener/resources/secrets.go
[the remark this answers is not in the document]
dcaravel: Thats right, my thought process was to not disturb what was existing. Also the roxctl, image watches, and Jenkins plugin will still request scans via Central and would benefit from the auto-generated integrations.  I created ROX-16349 for post-MVP to stop sending the image integrations when connected to a managed Central - which was a last minute request that couldn't fit into 4.0

sensor/admission-control/manager/images.go
```diff
@@ -80,21 +80,22 @@ func (m *manager) getImageFromSensorOrCentral(ctx context.Context, s *state, img
...
-func (m *manager) fetchImage(ctx context.Context, s *state, resultChan chan<- fetchImageResult, pendingCount *int32, idx int, image *storage.ContainerImage) {
+func (m *manager) fetchImage(ctx context.Context, s *state, resultChan chan<- fetchImageResult, pendingCount *int32, idx int, image *storage.ContainerImage, deployment *storage.Deployment) {
 	defer func() {
 		if atomic.AddInt32(pendingCount, -1) == 0 {
 			close(resultChan)
 		}
 	}()
 
-	scannedImg, err := m.getImageFromSensorOrCentral(ctx, s, image)
+	scannedImg, err := m.getImageFromSensorOrCentral(ctx, s, image, deployment)
```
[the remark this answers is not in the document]
dcaravel: Post MVP will also be sending the list of pull secret names and possibly the service account name - I felt it was cleaner to pass the pointer to the deployment instead of passing the individual fields. I don't mind either way, if you have a strong opinion on it I will change it to the individual fields post MVP

sensor/admission-control/manager/images.go
```diff
@@ -60,7 +60,7 @@ type fetchImageResult struct {
 	img *storage.Image
 }
 
-func (m *manager) getImageFromSensorOrCentral(ctx context.Context, s *state, img *storage.ContainerImage) (*storage.Image, error) {
+func (m *manager) getImageFromSensorOrCentral(ctx context.Context, s *state, img *storage.ContainerImage, deployment *storage.Deployment) (*storage.Image, error) {
 	// Talk to central if we know its endpoint (and the client connection is not shutting down), and if we are not
 	// currently connected to sensor.
 	// Note: Sensor is required to scan images in the local registry.
```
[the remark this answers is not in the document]
dcaravel: Those will indeed fail to be scanned - I mentioned this in the initial design review, it seemed it is was a known unlikely edge case and affects scanning images from OCP internal registries today (admission controller existing without sensor)

I'll bring this up again in the post-MVP design and discuss preferred way to handle - added ROX-16350 to track it

sensor/common/scan/scan.go
```diff
@@ -67,55 +83,171 @@ func (s *LocalScan) EnrichLocalImage(ctx context.Context, centralClient v1.Image
...
+// from the local registry. Afterwards, missing enriched data such as signature verification results and image
+// vulnerabilities will be fetched from central, returning the fully enriched image. A request is always sent
+// to central even if errors occur pulling metadata, scanning, or fetching signatures so that the error may be
+// recorded
+//
+// Will use the first registry from registries that succeeds in pulling metadata, or if registries is empty will
+// assume no auth is required
+//
+// Will return any errors that may occur during scanning, fetching signatures or during reaching out to central.
+func (s *LocalScan) EnrichLocalImageFromRegistry(ctx context.Context, centralClient v1.ImageServiceClient, ci *storage.ContainerImage, registries []registryTypes.Registry) (*storage.Image, error) {
```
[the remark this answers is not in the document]
dcaravel: No it does not in it's current state - will make it private

PR #5178, opened 2023-03-09
ROX-15824: Show disabled active vuln management

vjwilson:
## Description

This is a minimal implementation to hide the Active columns wherever they appear.

## Checklist

## Testing Performed

With flag off,
Screen Shot 2023-03-09 at 1 07 26 PM
Screen Shot 2023-03-09 at 1 15 46 PM
Screen Shot 2023-03-09 at 1 13 57 PM

With flag on,
Screen Shot 2023-03-09 at 1 54 47 PM


ui/apps/platform/src/Containers/VulnMgmt/List/ImageComponents/ListImageComponents.utils.test.js
```diff
@@ -6,6 +6,13 @@ import { WorkflowState } from 'utils/WorkflowState';
 
 import { getFilteredComponentColumns } from './ListImageComponents.utils';
 
+function mockIsFeatureFlagEnabled(flag) {
```
sachaudh, line 9: nit: Can we just call it `isFeatureFlagEnabled` if it's just used within this file

sachaudh (approved):
This looks fine to me. @vjwilson @connorgorman would we need to make modifications to the GraphQL query due to this change? Is the issue that we're querying for that data or is this just about showing/hiding it? Also, is this going to be documented somewhere as well?

connorgorman:
@sachaudh I can make it easy so that we don't need to modify the graphql and instead will have it return disabled or something like that. What do you think?

connorgorman (approved):
lgtm, if you merge it into my branch, we should be good to go

md2119 (approved):
Hopefully, we are not carrying over this code with many tributaries into VM2.0. The backend can return a string `Feature Disabled` or similar if you want to keep things simple.

PR #5182, opened 2023-03-10
Enable alerts to use GetByQuery instead of Search and GetMany

connorgorman:
## Description

Couple changes needed to be made to allow this to happen, mostly around the searcher wrappers and also moving sorting into the service layer where it is expected. 

## Checklist

## Testing Performed

CI + Scale test

### DIFF
--- a/central/alert/datastore/datastore_impl.go
+++ b/central/alert/datastore/datastore_impl.go
@@ -96,6 +96,7 @@ func (ds *datastoreImpl) ListAlerts(ctx context.Context, request *v1.ListAlertsR
 	}
 
 	paginated.FillPagination(q, request.GetPagination(), math.MaxInt32)
+	q = paginated.FillDefaultSortOption(q, paginated.ViolationTimeSortOption)
 
 	alerts, err := ds.SearchListAlerts(ctx, q)
 	if err != nil {
--- a/central/alert/datastore/internal/search/searcher_impl.go
+++ b/central/alert/datastore/internal/search/searcher_impl.go
@@ -24,13 +24,8 @@ import (
 var (
 	log = logging.LoggerForModule()
 
-	defaultSortOption = &v1.QuerySortOption{
-		Field:    search.ViolationTime.String(),
-		Reversed: true,
-	}
-
-	alertSearchHelper           = sac.ForResource(resources.Alert).MustCreateSearchHelper(mappings.OptionsMap)
-	alertPosgresSACSearchHelper = sac.ForResource(resources.Alert).MustCreatePgSearchHelper()
+	alertSearchHelper            = sac.ForResource(resources.Alert).MustCreateSearchHelper(mappings.OptionsMap)
+	alertPostgresSACSearchHelper = sac.ForResource(resources.Alert).MustCreatePgSearchHelper()
 )
 
 // searcherImpl provides an intermediary implementation layer for AlertStorage.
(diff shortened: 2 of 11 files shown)


central/alert/datastore/datastore_impl.go
```diff
@@ -96,6 +96,7 @@ func (ds *datastoreImpl) ListAlerts(ctx context.Context, request *v1.ListAlertsR
 	}
 
 	paginated.FillPagination(q, request.GetPagination(), math.MaxInt32)
```
md2119, line 98: Maybe in the future we combine `FillDefaultSortOption` and `FillPagination`, with the default sort option as an optional parameter, because the sort option is part of the pagination struct.

central/alert/datastore/internal/search/searcher_impl.go
```diff
@@ -55,6 +50,18 @@ func (ds *searcherImpl) SearchAlerts(ctx context.Context, q *v1.Query) ([]*v1.Se
 
 // SearchListAlerts retrieves list alerts from the indexer and storage
 func (ds *searcherImpl) SearchListAlerts(ctx context.Context, q *v1.Query) ([]*storage.ListAlert, error) {
+	if env.PostgresDatastoreEnabled.BooleanSetting() {
```
md2119, line 53: Should we add `ensurePaginated` to getAll/List type of functions in our code? The check will ensure we are not pulling crazy data into memory. (Maybe we can do it as part of search productization epic)

central/alert/datastore/internal/search/searcher_impl.go
```diff
@@ -140,29 +151,17 @@ func formatSearcher(unsafeSearcher blevesearch.UnsafeSearcher) search.Searcher {
...
 	} else {
 		filteredSearcher = alertSearchHelper.FilteredSearcher(unsafeSearcher) // Make the UnsafeSearcher safe.
+		filteredSearcher = sortfields.TransformSortFields(filteredSearcher, mappings.OptionsMap)
+		filteredSearcher = paginated.Paginated(filteredSearcher)
 	}
-	transformedSortFieldSearcher := sortfields.TransformSortFields(filteredSearcher, mappings.OptionsMap)
-	paginatedSearcher := paginated.Paginated(transformedSortFieldSearcher)
-	defaultSortedSearcher := paginated.WithDefaultSortOption(paginatedSearcher, defaultSortOption)
-	withDefaultViolationState := withDefaultActiveViolations(defaultSortedSearcher)
+	withDefaultViolationState := withDefaultActiveViolations(filteredSearcher)
```
md2119, line 160: Is `applyDefaultState` same as this? If so, we can drop it off from one of the places.

central/alert/datastore/internal/search/searcher_impl.go
```diff
@@ -181,30 +180,28 @@ func (ds *defaultViolationStateSearcher) Search(ctx context.Context, q *v1.Query
...
-			querySpecifiesStateField = true
-		}
-	})
+type defaultViolationStateSearcher struct {
+	searcher search.Searcher
+}
 
-	// By default, set stale to false.
-	if !querySpecifiesStateField {
-		cq := search.ConjunctionQuery(q, search.NewQueryBuilder().AddExactMatches(search.ViolationState, storage.ViolationState_ACTIVE.String()).ProtoQuery())
```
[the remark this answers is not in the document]
md2119: Yup, it seems so.

md2119 (approved):
Some minor comments. Otherwise, overall, lgtm.

rhybrillou (approved):

central/alert/datastore/internal/search/searcher_impl.go
connorgorman: I think I'd rather alert on the number of rows returned. Sometimes we have large queries that actually use that without pagination. For example, the alerts in alertmanager don't use pagination. We'd have to offer something like next page to make it really easy

PR #5183, opened 2023-03-10
Index on cve

md2119:
## Description

Indexing on the `cve` field improves the performance of the vm2.0 cve list page. The query for the vm2.0 cve list page uses group by on cve field to aggregate data spanning across multiple distros.

## Checklist

## Testing Performed
Manual testing


dashrews78:
Did you test this via an upgrade to make sure the index was created?  It should be because after the migrations run we make a call to apply all schemas.  Since it is an addition, gorm should apply it.  
My primary concern with relying on that is how long it would take to create these indexes at a scale and if that would be fall into one of the various statement or context timeouts.  I just think we should make sure it works at a scale.

md2119:
> Did you test this via an upgrade to make sure the index was created? It should be because after the migrations run we make a call to apply all schemas. Since it is an addition, gorm should apply it. My primary concern with relying on that is how long it would take to create these indexes at a scale and if that would be fall into one of the various statement or context timeouts. I just think we should make sure it works at a scale.

All insightful points!

I have a cluster with 35000 deployments, 5000 images, 41000 image CVEs in the DB. central nor central_db logs show any errors about statement timeouts related to upserts.

As for the upgrade, I upgraded with `set image` to build from this PR, and everything seems to have been constructed properly.

Do I get a badge? 😃

dashrews78 (approved):
Well done.

PR #5185, opened 2023-03-10
ROX-14694: Prefactor datastores used in declarative config to include filter functions

dhaus67:
## Description

This PR is a prefactor for adding deletion to the declarative configuration reconciliation.

Each datastore that will be used during the reconciliation will be updated to implement a `Get...Filtered` function.

This will allow us later do to the following:
- Filter out all resources that are declarative via their `GetTraits().GetOrigin()`
- Filter out all resources that should exist declaratively via their respective IDs.

Note that the group datastore already provides such functionality, thus it was skipped within this prefactor.

## Checklist

## Testing Performed

- see added unit tests on each datastore.

### DIFF
--- a/central/authprovider/datastore/datastore_impl.go
+++ b/central/authprovider/datastore/datastore_impl.go
@@ -25,7 +25,10 @@ type datastoreImpl struct {
 
 // GetAllAuthProviders retrieves authProviders.
 func (b *datastoreImpl) GetAllAuthProviders(ctx context.Context) ([]*storage.AuthProvider, error) {
-	// No SAC checks here because all users need to be able to read auth providers in order to authenticate.
+	if err := sac.VerifyAuthzOK(accessSAC.ReadAllowed(ctx)); err != nil {
+		return nil, err
+	}
+
 	return b.storage.GetAll(ctx)
 }
 
@@ -34,7 +37,7 @@ func (b *datastoreImpl) GetAuthProvidersFiltered(ctx context.Context,
 	if err := sac.VerifyAuthzOK(accessSAC.ReadAllowed(ctx)); err != nil {
 		return nil, err
 	}
-	// TODO(ROX-XXXXX): The store currently doesn't provide a Walk function. This is mostly due to us supporting the
+	// TODO(ROX-15902): The store currently doesn't provide a Walk function. This is mostly due to us supporting the
 	// old bolt store. Once we deprecate old store solutions with the 4.0.0 release, this should be changed to use
 	// store.Walk.
 	authProviders, err := b.storage.GetAll(ctx)
--- a/central/authprovider/datastore/datastore_impl.go
+++ b/central/authprovider/datastore/datastore_impl.go
@@ -29,12 +29,31 @@ func (b *datastoreImpl) GetAllAuthProviders(ctx context.Context) ([]*storage.Aut
 	return b.storage.GetAll(ctx)
 }
 
+func (b *datastoreImpl) GetAuthProvidersFiltered(ctx context.Context,
+	filter func(provider *storage.AuthProvider) bool) ([]*storage.AuthProvider, error) {
+	if err := sac.VerifyAuthzOK(accessSAC.ReadAllowed(ctx)); err != nil {
+		return nil, err
+	}
+	// TODO(ROX-XXXXX): The store currently doesn't provide a Walk function. This is mostly due to us supporting the
+	// old bolt store. Once we deprecate old store solutions with the 4.0.0 release, this should be changed to use
+	// store.Walk.
+	authProviders, err := b.storage.GetAll(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "retrieving auth providers")
+	}
+	filteredAuthProviders := make([]*storage.AuthProvider, 0, len(authProviders))
+	for _, authProvider := range authProviders {
+		if filter(authProvider) {
+			filteredAuthProviders = append(filteredAuthProviders, authProvider)
+		}
+	}
+	return filteredAuthProviders, nil
+}
+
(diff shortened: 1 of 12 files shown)


dhaus67:
* master
  * **PR #5185** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5185" target="_blank">Graphite</a>  👈
    * **PR #5186** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5186" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5185?utm_source=stack-comment).

central/authprovider/datastore/datastore_impl.go
```diff
@@ -29,12 +29,31 @@ func (b *datastoreImpl) GetAllAuthProviders(ctx context.Context) ([]*storage.Aut
 	return b.storage.GetAll(ctx)
 }
 
+func (b *datastoreImpl) GetAuthProvidersFiltered(ctx context.Context,
+	filter func(provider *storage.AuthProvider) bool) ([]*storage.AuthProvider, error) {
+	if err := sac.VerifyAuthzOK(accessSAC.ReadAllowed(ctx)); err != nil {
+		return nil, err
+	}
+	// TODO(ROX-XXXXX): The store currently doesn't provide a Walk function. This is mostly due to us supporting the
```
ivan-degtiarenko, line 37: Let's replace `XXXXX` with actual JIRA number

central/authprovider/datastore/datastore_impl.go
```diff
@@ -29,12 +29,31 @@ func (b *datastoreImpl) GetAllAuthProviders(ctx context.Context) ([]*storage.Aut
 	return b.storage.GetAll(ctx)
```
ivan-degtiarenko, line 29: I thought we can reuse `GetAuthProvidersFiltered` here, but apparently, the SAC check is skipped here. It's unrelated to changes in this PR, but I think we should actually delegate using the right context to the clients of this datastore, and put a SAC check here. WDYT?

central/role/datastore/datastore_impl.go
```diff
@@ -152,6 +152,27 @@ func (ds *dataStoreImpl) GetAllRoles(ctx context.Context) ([]*storage.Role, erro
 	return ds.getAllRolesNoScopeCheck(ctx)
 }
 
+func (ds *dataStoreImpl) GetRolesFiltered(ctx context.Context, filter func(role *storage.Role) bool) ([]*storage.Role, error) {
+	if ok, err := roleSAC.ReadAllowed(ctx); !ok || err != nil {
```
ivan-degtiarenko, line 156: Why `!ok` results in no error here?

central/authprovider/datastore/datastore_impl_test.go
```diff
@@ -142,6 +142,26 @@ func (s *authProviderDataStoreTestSuite) TestErrorOnAdd() {
...
+			Id:   "some-id-2",
+			Name: "some-name-2",
+		},
+	}
+	s.storage.EXPECT().GetAll(gomock.Any()).Return(authProviders, nil)
+
+	filteredAuthProviders, err := s.dataStore.GetAuthProvidersFiltered(s.hasReadCtx, func(authProvider *storage.AuthProvider) bool {
+		return authProvider.GetName() == "some-name-1"
+	})
+	s.NoError(err)
```
ivan-degtiarenko, line 161: Let's add
```
s.Len(filteredAuthProviders, 1)
```

central/role/datastore/datastore_impl.go
dhaus67: I guess I just copied it from the other role implementations.
There seems to be a mix within the datastore - some use  the construct above, others use `sac.VerifyAuthZOK` (e.g. upserts, remove). While it's a change to the API, it feels more natural to me. I'll move the whole store to use `sac.VerifyAuthZOK`.

central/authprovider/datastore/datastore_impl.go
dhaus67: You mean `GetAllAuthProviders`, right?

I was wondering the same initially, but decided against it due to the comment specifying that users should need this.
However, looking as its usages being both the `telemetry` as well as the registry, and both specify the right access rights, I changed it to also do a SAC check.

central/authprovider/datastore/datastore_impl.go
dhaus67: Will do once JIRA maintenance window is over.

central/role/datastore/datastore_impl.go
```diff
@@ -287,11 +304,32 @@ func (ds *dataStoreImpl) GetAllPermissionSets(ctx context.Context) ([]*storage.P
 	return permissionSets, nil
 }
 
+func (ds *dataStoreImpl) GetPermissionSetsFiltered(ctx context.Context,
+	filter func(permissionSet *storage.PermissionSet) bool) ([]*storage.PermissionSet, error) {
```
ivan-degtiarenko, line 308: I think in the ideal world the filter should be a query that can be transformed into SQL `WHERE` clause. However, looking at how it is currently implemented for policies this would definitely be overkill to reimplement here.

ivan-degtiarenko (approved):
LGTM

central/role/datastore/datastore_impl.go
dhaus67: Yeah I was thinking the same thing. For that to be supported I suppose we have to a) expose search functionality for the datastore b) make the origin field searchable.
Maybe we can use the query select framework for that.

I created https://issues.redhat.com/browse/ROX-16071 to cover that one.

Issue #5126, opened 2023-03-07
Sanitize username and password on SMTP AUTH LOGIN


amedeos:
Currently central for AUTH LOGIN doesn't sanitize "username" and "password" commands coming from smtp server, for example AlertManager before evaluating that commands use strings.ToLower function.

This is central code:
https://github.com/stackrox/stackrox/blob/3.74.0/central/notifiers/email/email.go#L101

instead alertmanager' code:
https://github.com/prometheus/alertmanager/blob/main/notify/email/email.go#L353

amedeos:
just opened pull request #5189

PR #5189, opened 2023-03-10
Sanitize username and password on SMTP AUTH LOGIN (#5126)

amedeos:
Sanitize username and password on SMTP AUTH LOGIN using strings.ToLower before comparing in switch case

## Description
Before compare what from SMTP server we got, it is useful to "sanitize" with ToLower strings function, otherwise we can never intercept "username" and "password" commands.


keyallis (changes requested):
If you could please give this a rebase onto the current master branch, that should resolve the CI failures.

rukletsov (approved):

PR #5194, opened 2023-03-10
ROX-14674: Fix duplicate baselines

fredrb:
## Description

Network Baseline info does not show CIDR blocks when the peer is an external source. This PR aims to add that information to the baseline entry, which can be consumed by the FE.

- [x] Figure out how to upgrade from clusters that already have baseline entries without a CIDR block. 

A: It **does** need a migration. I'm working on it on a separate PR that has to be merged before the release. I'm splitting in different PRs because (1) this PR already unblocks the FE for new installations; (2) not having a migration won't crash central, it just won't show the CIDR block for existing baselines (existing behavior); (3) the migration code could be reviewed by people that have more knowledge on the Postgres migration. 

**Update:** here's the PR with the migration: #5252

For more context on the issue, check [this decision doc](https://docs.google.com/document/d/1Tw3mf6dbd_-fwlKwP55xUK7OWVr4fePp8XkzvV2yWxE/edit?usp=sharing).

## Checklist
- [ ] ~~Evaluated and added CHANGELOG entry if required~~ Maybe this will only be needed after the FE implements the change
- [ ] ~~Determined and documented upgrade steps~~ Depends if `4.0` will only have new installations.
- [ ] ~~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~ Probably not needed now either.

## Testing Performed

- [x] Existing unit tests
- [x] E2E Tests
- [x] Check payload contains CIDR block on newly generated baselines

For the example in the ticket, now the response with baselines shows the `CIDR` blocks when the entity is an external source.

image

### DIFF
--- a/central/networkbaseline/manager/manager_impl.go
+++ b/central/networkbaseline/manager/manager_impl.go
@@ -202,7 +202,12 @@ func (m *manager) sendNetworkBaselinesToSensor(baselines []*storage.NetworkBasel
 	}
 }
 
-func (m *manager) lookUpPeerName(entity networkgraph.Entity) string {
+type peerInfo struct {
+	name      string
+	cidrBlock string
+}
+
+func (m *manager) lookUpPeerInfo(entity networkgraph.Entity) peerInfo {
 	switch entity.Type {
 	case storage.NetworkEntityInfo_DEPLOYMENT:
 		// If the peer is a deployment, just look it up from the baselines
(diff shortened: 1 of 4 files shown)


fredrb:
@sachaudh you can test the FE against the image built in this branch and let me know if that's sufficient to fix the issue.

central/networkbaseline/manager/manager_impl.go
```diff
@@ -247,24 +256,25 @@ func (m *manager) processFlowUpdate(flows map[networkgraph.NetworkConnIndicator]
...
 					DstPort:   conn.DstPort,
 					Protocol:  conn.Protocol,
+					CidrBlock: info.cidrBlock,
 				}, modifiedDeploymentIDs)
 			}
 		}
 		if conn.DstEntity.Type == storage.NetworkEntityInfo_DEPLOYMENT {
-			peerName := m.lookUpPeerName(conn.SrcEntity)
-			if peerName != "" {
+			peerInfo := m.lookUpPeerInfo(conn.SrcEntity)
```
Maddosaurus, line 272: Nit: Is there a reason you are switching between `info` and `peerInfo` as the result var for `lookUpPeerInfo`? 
Here, it is `peerInfo`, but in [L259](https://github.com/stackrox/stackrox/pull/5194/files#diff-460fa46f51a7facf2085f418a1f9b6bbec93f5046f1efb205d68bc702ac871acR259) and [L443](https://github.com/stackrox/stackrox/pull/5194/files#diff-460fa46f51a7facf2085f418a1f9b6bbec93f5046f1efb205d68bc702ac871acR443) it is `info`. 

And one more thought: How about calling it just `peer`, as it is info we're saving about a peer? That would make it `peer.name` and `peer.cidr` or `peer.cidrBlock`, which I (personally) find very pleasing 😄 
I am aware that `peer` as a var name is already taken, so there would be renaming needed, but just throwing that thout out there.

Maddosaurus (approved):
Generally, LGTM!
I do have the feeling that we should add tests for `peer.go`, specifically for its conversion functions, though.
But as you didn't add this PKG file, I'm not blocking the merge because of that :D

Maddosaurus:
I would still recommend we check with Frontend if that fixes their issue (and they implement the needed changes).
Great analysis on the root cause, btw 👍

vikin91 (approved):
The code looks good! 
I do not know about the DB and potential migrations needed for that, but I see this has been discussed in Slack.

Have you been thinking of adding/modifying e2e/integration test to query the API the way you did that manually in the PR description?

central/networkbaseline/manager/manager_impl.go
fredrb: That makes sense. I didn't notice that I named it differently. Changed both of them to `peer`. Thanks!

pkg/networkgraph/networkbaseline/peer.go
```diff
@@ -46,13 +47,21 @@ type Peer struct {
 	Name      string
 	DstPort   uint32
 	Protocol  storage.L4Protocol
+	CidrBlock string
```
md2119, line 50: nit: comment that `CidrBlock` is used for only for external sources.

md2119 (approved):

PR #5195, opened 2023-03-10
Segment: add a delay between 3 Track events after Group

0x656b694d:
## Description

As Segment doesn't guarantee the processing order of the events, and we need to send Group+Track in order, we send 3 consecutive Track events with 2s delay between to raise the chances for the group properties to be captured.

This code is used today only in Fleet Manager when assigning Tenant properties to a new Tenant group.

This change makes sense only if max batch size is set to 1, so that all events are sent immediately and separately.

## Checklist

## Testing Performed

Unit tests.

### DIFF
--- a/pkg/telemetry/phonehome/segment/segment.go
+++ b/pkg/telemetry/phonehome/segment/segment.go
@@ -209,12 +209,25 @@ func (t *segmentTelemeter) Group(props map[string]any, opts ...telemeter.Option)
 		// Track the group properties update with the same device ID
 		// to ensure following events get the properties attached. This is
 		// due to Amplitude partioning by device ID.
-		if err := t.client.Enqueue(segment.Track{
+		track := segment.Track{
 			Event:   "Group Properties Updated",
 			Context: dctx,
-		}); err != nil {
-			log.Error("Cannot enqueue Segment track event: ", err)
 		}
+		go func() {
+			// Segment does not guarantee the processing order of the events,
+			// we need, therefore, to add a delay between Group and Track to
+			// ensure the Track catches the group properties. We do it several
+			// times to raise the chances for the potential events from other
+			// clients coming in between to capture the group properties.
+			for i := 0; i < 3; i++ {
+				if i != 0 {
+					time.Sleep(2 * time.Second)
+				}
+				if err := t.client.Enqueue(track); err != nil {
+					log.Error("Cannot enqueue Segment track event: ", err)
+				}
+			}
+		}()
 	}
 }
 
--- a/pkg/telemetry/phonehome/segment/segment.go
+++ b/pkg/telemetry/phonehome/segment/segment.go
@@ -3,6 +3,7 @@ package segment
 import (
 	"time"
 
+	"github.com/benbjohnson/clock"
 	segment "github.com/segmentio/analytics-go/v3"
 	"github.com/stackrox/rox/pkg/httputil/proxy"
 	"github.com/stackrox/rox/pkg/logging"
@@ -12,6 +13,8 @@ import (
 var (
 	log                     = logging.LoggerForModule()
 	_   telemeter.Telemeter = (*segmentTelemeter)(nil)
+
+	internalClock clock.Clock = clock.New()
 )
 
 type segmentTelemeter struct {
@@ -184,12 +187,11 @@ func (t *segmentTelemeter) Group(props map[string]any, opts ...telemeter.Option)
 	}
 
 	options := telemeter.ApplyOptions(opts)
-	dctx := t.makeContext(options)
 	group := segment.Group{
 		UserId:      t.getUserID(options),
 		AnonymousId: t.getAnonymousID(options),
 		Traits:      props,
-		Context:     dctx,
+		Context:     t.makeContext(options),
 	}
 
 	for _, ids := range options.Groups {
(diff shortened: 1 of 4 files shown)


0x656b694d:
* master
  * **PR #5195** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5195" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5195?utm_source=stack-comment).

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -209,12 +211,27 @@ func (t *segmentTelemeter) Group(props map[string]any, opts ...telemeter.Option)
...
-			Context: dctx,
-		}); err != nil {
-			log.Error("Cannot enqueue Segment track event: ", err)
+		track := segment.Track{
+			UserId:      group.UserId,
+			AnonymousId: group.AnonymousId,
+			Event:       "Group Properties Updated",
+			Context:     group.Context,
 		}
+		go func() {
```
dhaus67, line 220: Instead of adding another dependency, we could do the following instead:
- use a `*timer.Ticker` which will have an interval of two seconds.
- within the spawned go routine, we will have essentially something like this:
```go
var trackCalls int
for trackCalls < 3 {
		select {
		case <-t.ticker.C:
			if err := t.client.Enqueue(track); err != nil {
				log.Error("Cannot enqueue Segment track event: ", err)
			}
                // Theoretically, if we had to respect context cancellation, we would do so here.
		}
}
```
Within the initialization of the tracker we would start the ticker, and within the `Stop()` implementation of the `telemeter.Telemeter` interface we would stop this. I personally don't think it's worth to add a dependency for this.

0x656b694d: True, the same could be done with a timer. Please have a look at the updated code.
I found a use of `github.com/benbjohnson/clock` in operator tools, so thought it's fine to use it here.

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -209,12 +212,31 @@ func (t *segmentTelemeter) Group(props map[string]any, opts ...telemeter.Option)
...
+			if ti == nil {
+				ti = time.NewTicker(2 * time.Second)
+			}
```
dhaus67, line 224: We should probably do this either a) behind a lock or b) outside of the go routine. Currently, this would create a race condition between two go routines accessing this value.

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -209,12 +212,31 @@ func (t *segmentTelemeter) Group(props map[string]any, opts ...telemeter.Option)
...
+			// clients coming in between to capture the group properties.
+			for i := 0; i < 3; i++ {
+				if i != 0 {
+					<-ti.C
+				}
+				if err := t.client.Enqueue(track); err != nil {
+					log.Error("Cannot enqueue Segment track event: ", err)
+				}
+			}
+			ti.Stop()
```
dhaus67, line 238: Hm, we could use a shared ticker throughout the telemeter by having it as a property on the telemeter struct and initialize it in new / start.

If you want to do one on a per-track, maybe factor this inline goroutine into a separate function that takes in a ticker that will be created within the current track function?

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: There wouldn't be a race for this variable, as it's bound to a call.
There is a race though with Segment internal channel, if we close the client while the goroutine is still running.

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: Please take a look at the per-track timer.

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -18,6 +19,8 @@ type segmentTelemeter struct {
 	client     segment.Client
 	clientID   string
 	clientType string
+
+	stopMux *sync.Mutex
```
dhaus67, line 23: Maybe a `sync.RWMutex` is more fitting?

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -155,7 +165,7 @@ func (t *segmentTelemeter) makeContext(o *telemeter.CallOptions) *segment.Contex
 }
 
 func (t *segmentTelemeter) Identify(props map[string]any, opts ...telemeter.Option) {
-	if t == nil {
+	if t == nil || t.client == nil {
```
dhaus67, line 168: Shouldn't this be also guarded behind a mutex then? Theoretically, the same datarace could happen here as well?

pkg/telemetry/phonehome/segment/segment.go
```diff
@@ -241,7 +256,7 @@ func (t *segmentTelemeter) group(props map[string]any, ti *time.Ticker, opts ...
 }
 
 func (t *segmentTelemeter) Track(event string, props map[string]any, opts ...telemeter.Option) {
-	if t == nil {
+	if t == nil || t.client == nil {
```
dhaus67, line 259: Same as above w.r.t. doing this guarded behind a mutex.

dhaus67 (approved):
Just minor comments regarding moving to `RWMutex` and other places where we should use a mutex.

pkg/telemetry/phonehome/segment/segment.go
0x656b694d: Yeah, a different race, but true.

0x656b694d:
@dhaus67 , behold, I removed the mutex. Segment SDK doesn't need to be synchronized, so I refactored the code so that the test execute the synchronous pieces.

dhaus67 (approved):
Sorry for the late review, LGTM 🚢

PR #5201, opened 2023-03-10
Update AWS Identity Doc verification to IMDSv2

RTann:
## Description

IMDSv1 is outdated in favor of IMDSv2. This PR updated our AWS PKCS7 signature verification to use IMDSv2. [As of 1.25.38 of the aws go sdk, IMDSv2 is supported](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#use-a-supported-sdk-version-for-imdsv2).

What's more, the pkcs7 endpoint we used uses DSA with SHA1 which are both not very secure. This PR updates to use RSA-2048 signatures instead.

Note: the certificates may rotate! We may need to periodically check on these...

## Checklist

## Testing Performed

Deployed StackRox to an EKS cluster in `us-west-2`, and I found we were able to validate the cloud provider metadata:

Screenshot 2023-03-15 at 6 45 38 PM

We can see the same result for `us-east-1`:

Screenshot 2023-03-16 at 9 31 16 AM

I also went through the entire list of [ROSA regions](https://docs.openshift.com/rosa/rosa_architecture/rosa_policy_service_definition/rosa-service-definition.html#rosa-sdpolicy-regions-az_rosa-service-definition) and made sure I correctly copied the certificate over from AWS's website.


pkg/cloudproviders/aws/metadata.go
```diff
@@ -2,24 +2,62 @@ package aws
...
+		Transport: proxy.Without(),
+	}
 )
 
 // GetMetadata tries to obtain the AWS instance metadata.
 // If not on AWS, returns nil, nil.
 func GetMetadata(ctx context.Context) (*storage.ProviderMetadata, error) {
-	var verified bool
-
 	errs := errorhelpers.NewErrorList("retrieving AWS EC2 metadata")
```
connorgorman, line 37: let's just move this down to where we start having multiple and default to errors.wrap

clickboo (approved):
lgtm. I had looked at this code before for another JIRA about cloud metadata and like you, I am a bit worried about the embedded certificates and their rotation bit since all we can do is "keep it in mind".

pkg/cloudproviders/aws/certs.go
```diff
@@ -0,0 +1,612 @@
...
+	"github.com/stackrox/rox/pkg/utils"
+)
+
+// awsCerts lists all known AWS certificates as of time of writing.
+// See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-rsa2048.html
+// for the current list of RSA-2048 certificates.
+//
+// Note: THIS CAN BE BRITTLE. THERE IS NO GUARANTEE THESE CERTS WILL LAST FOREVER.
+// See https://github.com/aws/aws-sdk-go/pull/1593#pullrequestreview-70664445.
+// It is probably in everyone's best interest to check on this periodically.
```
keyallis, line 16: I have my reservations about this, but I'm not sure about alternatives to it

RTann: Yeah I don't like it, either, but this seems to be the way Amazon suggests to do it. They just list out a bunch of certificates that you are meant to use to manually check this. This is part of the reason the AWS Go SDK doesn't support this operation, and we need to do it ourselves

PR #5211, opened 2023-03-13
Fix handling empty string matches

erthalion:
## Description

Trimming values and building new string query do not handle empty strings correctly.

The first one is missing out to extract it out of double quotes due to the length condition (an empty string is `""`, which is exactly the same lenght as 2 * `"`, where the `"` is EqualityPrefixSuffix). This leads to an empty string being represented as `\"\"` along the way, producing incorrect search results.

The second one assumes that an empty string is something incorrect, although it's totally fine to search for values equal empty string.

## Checklist

## Testing Performed

Local testing via running a unit-test with an empty string search query.

### DIFF
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -225,7 +225,7 @@ forloop:
 			trimmedValue = strings.ToLower(trimmedValue[len(RegexPrefix):])
 			queryModifiers = append(queryModifiers, Regex)
 			break forloop // Once we see that it's a regex, we don't check for special-characters in the rest of the string.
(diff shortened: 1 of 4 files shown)


md2119 (changes requested):
Add some unit tests that directly tests exported function `GetValueAndModifiersFromString`.
For example,
`test`
`"test"`
`""`
`"`
`"test`

erthalion:
@md2119 now the fix has more tests for modifiers.

connorgorman (approved):

md2119 (approved):
Thanks for the tests and for fixing the panic case from the first commit.

PR #5226, opened 2023-03-14
Add Blob store to Postgres

connorgorman:
## Description

Add a blob store to Postgres that will be used to store probes, external entities and scanner definitions. May need to be tweaked in the future. Will add a feature flag so it can be changed until all those have been properly implemented

## Checklist

## Testing Performed

Unit test


central/blob/datastore/store/store.go
```diff
@@ -0,0 +1,148 @@
...
+
+	los := tx.LargeObjects()
+	lo, err := los.Open(ctx, existingBlob.GetOid(), pgx.LargeObjectModeWrite)
+	if err != nil {
+		err := errors.Wrapf(err, "error opening large object with oid %d", existingBlob.GetOid())
+		return nil, false, wrapRollback(ctx, tx, err)
+	}
+
+	buf := make([]byte, 1024*1024)
+	for {
```
c-du, line 126: What if the io.Writer is a slow writer or intermittent writer? Do we allow them?

PR #5229, opened 2023-03-14
dashrews/ROX-15991-gke postgres upgrade

dashrews78:
## Description

/pragma: gke_cluster_version: 1.24.9-gke.3200

changes for v1.24 kubernetes on GKE causes issues where older releases will timeout trying to get metadata.  For this test we use older releases to do upgrades.  The older central with an old restore will start but not fast enough to suit the call to `wait_for_api`.  The call to `wait_for_api` will fail and thus fail the test.  The simplest fix is to extend the amount of time that `wait_for_api` waits while trying to get a connection.  

Additionally since this test takes sensors and centrals up and down a lot, simply adding some collector exclusions for events when those other components do not start up quickly enough.  Those changes were in #5011 but I consolidated them here and will close that one.

## Checklist

## Testing Performed

This is just a test.


connorgorman (approved):
Nice!

tests/upgrade/postgres_run.sh
```diff
@@ -105,6 +105,13 @@ test_upgrade_paths() {
 
     cd "$TEST_ROOT"
 
+    # This test does a lot of upgrades and bounces.  If things take a little longer to bounce we can get entries in
+    # logs indicating communication problems.  Those need to be allowed in the case of this test ONLY.
+    echo "# postgres was bounced, may see some connection errors" >> scripts/ci/logcheck/allowlist-patterns
+    echo "FATAL: terminating connection due to administrator command \(SQLSTATE 57P01\)" >> scripts/ci/logcheck/allowlist-patterns
+    echo "Unable to connect to Sensor at"
+    echo >> scripts/ci/logcheck/allowlist-patterns
```
gavin-stackrox, line 113: Should this be:
```suggestion
    echo "Unable to connect to Sensor at" >> scripts/ci/logcheck/allowlist-patterns
```

tests/upgrade/postgres_run.sh
```diff
@@ -105,6 +105,13 @@ test_upgrade_paths() {
 
     cd "$TEST_ROOT"
 
+    # This test does a lot of upgrades and bounces.  If things take a little longer to bounce we can get entries in
```
gavin-stackrox, line 108: I like it! But this will result in a `-dirty` tag. Which may not be a problem if the tag is already consumed but could cause future headache. What about allowing `ALLOWLIST_FILE` to be already set in `scripts/ci/logcheck/check.sh`?  And then set it up for that call i.e. `cp scripts/ci/logcheck/allowlist-patterns /tmp/allowlist-patterns; echo ... >> /tmp/allowlist-patterns; export ALLOWLIST_FILE="/tmp/allowlist-patterns"`

gavin-stackrox (changes requested):

PR #5231, opened 2023-03-14
ROX-15780: set file modification date in diagnostic bundle

janisz:
## Description

Currently we do not set any time in diagnostic bundle files. This result with using 1980 as a default which is confusing for users as they expect current date. This PR fixes it by setting current timestamp. 

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -253,7 +253,7 @@ func fetchAndAddJSONToZip(ctx context.Context, zipWriter *zip.Writer, fileName s
 }
 
 func addJSONToZip(zipWriter *zip.Writer, fileName string, jsonObj interface{}) error {
-	w, err := zipWriter.Create(fileName)
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, fileName)
 	if err != nil {
 		return errors.Wrapf(err, "unable to create zip file %q", fileName)
 	}
@@ -265,7 +265,7 @@ func addJSONToZip(zipWriter *zip.Writer, fileName string, jsonObj interface{}) e
 }
 
 func zipPrometheusMetrics(zipWriter *zip.Writer, name string) error {
-	metricsWriter, err := zipWriter.Create(name)
+	metricsWriter, err := zipWriterWithCurrentTimestamp(zipWriter, name)
 	if err != nil {
 		return err
 	}
@@ -273,7 +273,7 @@ func zipPrometheusMetrics(zipWriter *zip.Writer, name string) error {
 }
 
 func getMemory(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("heap.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "heap.tar.gz")
 	if err != nil {
 		return err
 	}
@@ -281,7 +281,7 @@ func getMemory(zipWriter *zip.Writer) error {
 }
 
 func getCPU(ctx context.Context, zipWriter *zip.Writer, duration time.Duration) error {
-	w, err := zipWriter.Create("cpu.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "cpu.tar.gz")
 	if err != nil {
 		return err
 	}
@@ -297,7 +297,7 @@ func getCPU(ctx context.Context, zipWriter *zip.Writer, duration time.Duration) 
 }
 
 func getBlock(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("block.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "block.tar.gz")
 	if err != nil {
 		return err
 	}
@@ -306,7 +306,7 @@ func getBlock(zipWriter *zip.Writer) error {
 }
 
 func getMutex(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("mutex.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "mutex.tar.gz")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


central/debug/service/zip.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"archive/zip"
+	"io"
+	"path"
+	"time"
+
+	"github.com/stackrox/rox/pkg/k8sintrospect"
+)
+
+var now = func() time.Time {
```
ivan-degtiarenko, line 12: Alternative proposal:

```
func zipWriterWithCurrentTimestamp(zipWriter *zip.Writer, fileName string) (io.Writer, error) {
	return zipWriterWithCurrentTimestampUseNowFunc(zipWriter, fileName, func() time.Time { 
		return time.Now() 
	})
}

func zipWriterWithCurrentTimestampUseNowFunc(zipWriter *zip.Writer, fileName string, nowFunc func() time.Time ) (io.Writer, error) {
	header := &zip.FileHeader{
		Name:     fileName,
		Method:   zip.Deflate,
		Modified: now(),
	}
	return zipWriter.CreateHeader(header)
}
```

And using the second function for testing

janisz:
* master
  * **PR #5426** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5426" target="_blank">Graphite</a> 
    * **PR #5231** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5231" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5231?utm_source=stack-comment).

central/debug/service/zip.go
janisz: I believe it's better to have smaller dependencies. Function that returns just a current time gives me better coverage than mocking a whole function.

central/debug/service/service.go
```diff
@@ -265,23 +265,23 @@ func addJSONToZip(zipWriter *zip.Writer, fileName string, jsonObj interface{}) e
 }
 
 func zipPrometheusMetrics(zipWriter *zip.Writer, name string) error {
-	metricsWriter, err := zipWriter.Create(name)
+	metricsWriter, err := zipWriterWithCurrentTimestamp(zipWriter, name)
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 270: Let's be good citizens and  wrap the error similar to
https://github.com/stackrox/stackrox/pull/5231/files#diff-c81b773cf0c8cb97684ec44acb5cb687c550f73d5e24ea4df18d7251c8bcf185R258

central/debug/service/service.go
```diff
@@ -297,7 +297,7 @@ func getCPU(ctx context.Context, zipWriter *zip.Writer, duration time.Duration)
 }
 
 func getBlock(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("block.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "block.tar.gz")
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 302: Same here

central/debug/service/service.go
```diff
@@ -306,7 +306,7 @@ func getBlock(zipWriter *zip.Writer) error {
 }
 
 func getMutex(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("mutex.tar.gz")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "mutex.tar.gz")
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 311: Same here

central/debug/service/service.go
```diff
@@ -315,7 +315,7 @@ func getMutex(zipWriter *zip.Writer) error {
 }
 
 func getGoroutines(zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("goroutine.txt")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "goroutine.txt")
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 320: Same here

central/debug/service/service.go
```diff
@@ -334,7 +334,7 @@ func getLogs(zipWriter *zip.Writer) error {
 }
 
 func getLogFile(zipWriter *zip.Writer, targetPath string, sourcePath string) error {
-	w, err := zipWriter.Create(targetPath)
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, targetPath)
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 339: Same here

central/debug/service/service.go
```diff
@@ -390,7 +390,7 @@ func getCentralDBData(ctx context.Context, zipWriter *zip.Writer) error {
 }
 
 func (s *serviceImpl) getLogImbue(ctx context.Context, zipWriter *zip.Writer) error {
-	w, err := zipWriter.Create("logimbue-data.json")
+	w, err := zipWriterWithCurrentTimestamp(zipWriter, "logimbue-data.json")
 	if err != nil {
 		return err
```
ivan-degtiarenko, line 395: Same here

central/debug/service/zip.go
ivan-degtiarenko: Ideally, I think we would've had a struct calling these methods and that's where we would pass `now()` function, but that is too big of a refactoring for such a case. I won't push - let's keep it as it is

janisz: We can fix it in a separated PR as this will means that most of functions will be struct functions as we need `now()` to be somewhere defined

central/debug/service/service.go
janisz: How about wrapping error at `zipWriterWithCurrentTimestamp` level so we will have full context and don't need to wrap in every method call?

ivan-degtiarenko: sounds good to me 👍

central/debug/service/zip.go
ivan-degtiarenko: Yea, let's not fix it here

ivan-degtiarenko (approved):

PR #5237, opened 2023-03-14
ROX-10998: Verify role references in groups before deletion

dhaus67:
## Description

Previously, it was possible to delete role that were referenced by groups in auth providers (potentially also the minimum access role), leading to breaking auth provider's behavior when deleting such a role.

Now, before deleting a role, it will be first checked whether this role is referenced by any groups. If that's the case, an error will be returned indicating the number of groups that reference the role.

Note that this required a minor refactoring in the group datastore: instead of having the filter in `GetFiltered` filter on `storage.GroupProperties`, the filter now works on `storage.Group` (which it should have in the first place).

## Checklist

## Testing Performed

- see CI (added unit tests, adjusted existing ones).

### DIFF
--- a/central/group/datastore/datastore_impl.go
+++ b/central/group/datastore/datastore_impl.go
@@ -68,7 +68,7 @@ func (ds *dataStoreImpl) GetAll(ctx context.Context) ([]*storage.Group, error) {
 	return ds.storage.GetAll(ctx)
 }
 
-func (ds *dataStoreImpl) GetFiltered(ctx context.Context, filter func(*storage.GroupProperties) bool) ([]*storage.Group, error) {
+func (ds *dataStoreImpl) GetFiltered(ctx context.Context, filter func(*storage.Group) bool) ([]*storage.Group, error) {
 	if ok, err := accessSAC.ReadAllowed(ctx); err != nil {
 		return nil, err
 	} else if !ok {
@@ -79,7 +79,7 @@ func (ds *dataStoreImpl) GetFiltered(ctx context.Context, filter func(*storage.G
 	walkFn := func() error {
 		groups = groups[:0]
 		return ds.storage.Walk(ctx, func(g *storage.Group) error {
-			if filter == nil || filter(g.GetProps()) {
+			if filter == nil || filter(g) {
 				groups = append(groups, g)
 			}
 			return nil
@@ -220,8 +220,8 @@ func (ds *dataStoreImpl) Remove(ctx context.Context, props *storage.GroupPropert
 }
 
 func (ds *dataStoreImpl) RemoveAllWithAuthProviderID(ctx context.Context, authProviderID string, force bool) error {
-	groups, err := ds.GetFiltered(ctx, func(properties *storage.GroupProperties) bool {
-		return authProviderID == properties.GetAuthProviderId()
+	groups, err := ds.GetFiltered(ctx, func(group *storage.Group) bool {
+		return authProviderID == group.GetProps().GetAuthProviderId()
 	})
 	if err != nil {
 		return errors.Wrap(err, "collecting associated groups")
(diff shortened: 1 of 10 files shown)


dhaus67:
* master
  * **PR #5237** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5237" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5237?utm_source=stack-comment).

central/role/datastore/datastore_impl.go
```diff
@@ -721,11 +723,35 @@ func (ds *dataStoreImpl) verifyRoleForDeletion(ctx context.Context, name string)
...
+
+	// Elevate the context to use resources.Access.
+	// TODO(ROX-14398): Once we use Access for this datastore instead of role, we can reuse the same context here.
+	groupCtx := sac.WithGlobalAccessScopeChecker(context.Background(), sac.AllowFixedScopes(
+		sac.AccessModeScopeKeys(storage.Access_READ_ACCESS), sac.ResourceScopeKeys(resources.Access)))
+	return ds.verifyGroupReferences(groupCtx, role)
+}
+
+// Returns errox.ReferencedByAnotherObject if the given role is referenced by a group.
+func (ds *dataStoreImpl) verifyGroupReferences(ctx context.Context, role *storage.Role) error {
```
ivan-degtiarenko, line 742: nit: `verifyNoGroupReferences` corresponds better to what function does

central/role/datastore/datastore_impl.go
```diff
@@ -721,11 +723,35 @@ func (ds *dataStoreImpl) verifyRoleForDeletion(ctx context.Context, name string)
...
+	groups, err := ds.groupStorage.GetFiltered(ctx, func(group *storage.Group) bool {
+		return group.GetRoleName() == role.GetName()
+	})
+	if err != nil {
+		return err
+	}
+
+	if len(groups) > 0 {
+		return errox.ReferencedByAnotherObject.Newf("role %s is referenced by %d groups in auth providers, "+
+			"ensure all references to the role are removed", role.GetName(), len(groups))
```
ivan-degtiarenko, line 752: perhaps specifying group ids in the error would be more helpful for user than just a number

central/role/datastore/datastore_impl.go
```diff
@@ -721,11 +723,35 @@ func (ds *dataStoreImpl) verifyRoleForDeletion(ctx context.Context, name string)
...
 	}
 
-	return verifyNotDefaultRole(role)
+	if err := verifyNotDefaultRole(role); err != nil {
+		return err
+	}
+
+	// Elevate the context to use resources.Access.
+	// TODO(ROX-14398): Once we use Access for this datastore instead of role, we can reuse the same context here.
+	groupCtx := sac.WithGlobalAccessScopeChecker(context.Background(), sac.AllowFixedScopes(
```
ivan-degtiarenko, line 736: Why elevate context instead of using separate, internal `ctx`?

central/role/datastore/datastore_test.go
```diff
@@ -351,6 +357,11 @@ func (s *roleDataStoreTestSuite) TestRoleWriteOperations() {
 	s.ErrorIs(err, errox.InvalidArgs, "invalid scope for Upsert*() yields an error(declarative resource)")
 }
 
+func (s *roleDataStoreTestSuite) TestRoleGroupReferences() {
```
ivan-degtiarenko, line 360: I think this test case was not meant to be pushed :)

ivan-degtiarenko (approved):

central/group/datastore/datastore_impl.go
```diff
@@ -231,11 +231,9 @@ func (ds *dataStoreImpl) RemoveAllWithAuthProviderID(ctx context.Context, authPr
 
 func (ds *dataStoreImpl) RemoveAllWithEmptyProperties(ctx context.Context) error {
 	// Search through all groups and verify whether any group exists with empty properties and attempt to delete them.
-	isEmptyGroupPropertiesF := func(props *storage.GroupProperties) bool {
-		if props.GetAuthProviderId() == "" && props.GetKey() == "" && props.GetValue() == "" {
-			return true
-		}
-		return false
+	isEmptyGroupPropertiesF := func(group *storage.Group) bool {
```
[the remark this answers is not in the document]
dhaus67: The code behind it is quite old and we were unsure whether we even wanted to reintroduce this. So, if anything, I'd rather like to remove it altogether instead of adding another constraint (sadly, the comments on the original PR were quite sparse _why_ this was required to begin with..).

PR #5240, opened 2023-03-15
ROX-15798: Autocomplete filter section for Workload CVEs

alwayshooin:
Adding autocomplete filter input/dropdown section for Workload CVEs.
Ended up going with the UX of https://console.redhat.com/insights/vulnerability/cves?affecting=true&page=1&rule_presence=true&sort=-public_date where the dropdown state affects the autocomplete select/input to the right and the filters below display all filters that have been currently selected.
All filters should persist state within the URL.
Removing chips in the typeahead select AND in the filter chips should work as expected.
image
image
image
image
image
image
image
image


ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/FilterAutocomplete.tsx
```diff
@@ -0,0 +1,168 @@
...
+                categoryName="Namespace"
+                children={undefined}
+            />
+            <ToolbarFilter
+                chips={searchFilter.CLUSTER ? (searchFilter.CLUSTER as string[]) : []}
+                deleteChip={(_, chip) => onDelete('CLUSTER', chip as string)}
+                deleteChipGroup={() => onDeleteGroup('CLUSTER')}
+                categoryName="Cluster"
+                children={undefined}
+            />
```
alwayshooin, line 163: this makes me uncomfortable but i coudln't think of another way to have the filters show without having their respective inputs also show up. i tried even taking out the other filterchips in `WorkloadTableToolbar` but i am running into flex issues :(

dvail (commented):
A handful of things that came up during the sync when you were out that are related to searching/filtering. Not things that we should do in this PR but I wanted to get them written down before I forget.

1. Mansur added updated mocks to the bottom of the sketch file titled "Filtering UX Example" that use a single text input instead of the dropdown for the resource filters. We probably will want to switch to this since it gets a little unwieldy when there are multiple chips inside the text input as well as in the toolbar, and the layout starts to break. (These mocks haven't made it into the eng collab doc yet.)
2. We are going to need to exclude multiple resources from the resource dropdown instead of just a single one. (e.g. on the "Images" single page, the Image, Deployment, Namespace, and Cluster filter types don't really make sense). There is an updated table in the eng collab doc showing which pages should get which filters we can use when we get to the other pages.
3. Although we decided that the autocomplete should _not_ be scoped based on other filters that the user has applied, I still think we might want to scope it on the single pages. e.g. On the images page, autocomplete by CVE gives a list of all CVEs, most of which are not present in the image. I think on these pages we will want to filter by the current page's exact resource at least.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/FilterResourceDropdown.tsx
```diff
@@ -1,44 +1,54 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { Select, SelectOption } from '@patternfly/react-core';
 
-import { SearchFilter } from 'types/search';
+import useSelectToggle from 'hooks/patternfly/useSelectToggle';
 
-export type Resource = 'CVE' | 'Image' | 'Deployment' | 'Namespace' | 'Cluster';
+export type Resource = 'IMAGE_CVE' | 'IMAGE' | 'DEPLOYMENT' | 'NAMESPACE' | 'CLUSTER';
```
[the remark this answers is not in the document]
alwayshooin: so `IMAGE_CVE` is the category we have to use for the autocomplete query, which is why i changed it here. i can look into changing it only for the autocomplete input instead of across the logic here though

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/FilterChips.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+type FilterChipsProps = {
+    defaultFilters: DefaultFilters;
+    searchFilter: SearchFilter;
+    onDeleteGroup: (category) => void;
+    onDelete: (category, chip) => void;
+};
+
+function FilterChips({ defaultFilters, searchFilter, onDeleteGroup, onDelete }: FilterChipsProps) {
+    const deployments = (searchFilter.DEPLOYMENT as string[]) || [];
```
dvail, line 35: Probably unlikely, but if a user enters a query param in the URL in the form of `s[Deployment]=some-value`, or if a bad link is created elsewhere in the app, this entire page will crash. We should make a note or ticket to try and remove the type assertions as a follow up. I've got a couple of functions in the `searchUtils` filter in an in-progress branch that I think would work for this.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/FilterChips.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const images = (searchFilter.IMAGE as string[]) || [];
+    const namespaces = (searchFilter.NAMESPACE as string[]) || [];
+    const clusters = (searchFilter.CLUSTER as string[]) || [];
+    const severities = (searchFilter.Severity as VulnerabilitySeverityLabel[]) || [];
+    const fixables = (searchFilter.Fixable as FixableStatus[]) || [];
+
+    return (
+        <Flex spaceItems={{ default: 'spaceItemsXs' }}>
+            <FlexItem className="pf-u-pt-xs">
+                <ChipGroup
```
dvail, line 46: Small nit: each search category option here will leave an empty div with a 4px margin due to `pf-u-pt-xs` in the chip section that gets pretty noticeable if there are a lot of empty categories. Not worth keeping this PR from being merged though, we can get it during a later round of visual tweaks.
image

dvail (approved):
LGTM, just a couple of comments that we can do as follow ups if we want to get this PR in now.

It looks like using the FilterChips directly instead of the magic PF component actually resulted in a ton less code too! 😄

dvail:
Also, the merge conflict is from a quick fix that needed to apply when we had a broken TypeScript build after a semantic merge conflict.

PR #5243, opened 2023-03-15
ROX-14916: Add support for external configmap with postgresql.conf and pg_hba.conf

connorgorman:
## Description

This allows for folks to override the default configmap used for configuring postgres. This enables us and customers to optimize the config if necessary live at customers. Not sold on the naming of configOverride, open to suggestions

## Checklist

## Testing Performed

Tested specifying the option in the CR


dashrews78 (approved):
We should probably add it to `image/templates/helm/stackrox-central/values-public.yaml.example.htpl` as well.  Otherwise LGTM

operator/apis/platform/v1alpha1/central_types.go
```diff
@@ -202,6 +202,10 @@ type CentralDBSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,order=3
 	Persistence *DBPersistence `json:"persistence,omitempty"`
 
+	// Config map containing postgresql.conf and pg_hba.conf that will be used if modifications need to be applied
```
porridge, line 205:
```suggestion
	// Config map containing postgresql.conf and pg_hba.conf that will be used if modifications need to be applied.
```

image/templates/helm/stackrox-central/templates/01-central-12-central-db.yaml
```diff
@@ -134,7 +134,11 @@ spec:
         {{- toYaml ._rox.central.db.persistence._volumeCfg | nindent 8 }}
       - name: config-volume
         configMap:
+          {{- if ._rox.central.db.configOverride }}
+          name: {{ ._rox.central.db.configOverride }}
```
porridge, line 138: Nit: there's probably some template function which acts like `foo if foo ! "" else bar` that could help avoid the if/else here?

connorgorman: Trying a sprig ternary, let's see. I do find the if/else very annoying as well

connorgorman:
> We should probably add it to image/templates/helm/stackrox-central/values-public.yaml.example.htpl as well. Otherwise LGTM

@dashrews78 This is a bit of a weird case. I feel like helm users should just overwrite the postgresConfig var, but we need a specific var for operator

image/templates/helm/stackrox-central/values.yaml.htpl
```diff
@@ -201,17 +201,22 @@
...
+#      value: null
+#      generate: null
+#
+#    postgresConfig: "@config/centraldb/postgresql.conf|config/centraldb/postgresql.conf.default"
+#    hbaConfig: "@config/centraldb/pg_hba.conf|config/centraldb/pg_hba.conf.default"
+#
+#    # Specifying configOverride mounts the specified config map in the same namespace which must contain
+#    # both pg_hba.conf and postgresql.conf. This should only be used when the default settings are not
+#    # sufficient and manual override is required.
+#    configOverride: null
```
dashrews78, line 219: We need to carry this indention all the way to the scanner section.  For instance, based on this indentation `nodeSelector` at line 221 and at line 121 are at the same level.  Same for persistence.

dashrews78:
> > We should probably add it to image/templates/helm/stackrox-central/values-public.yaml.example.htpl as well. Otherwise LGTM
> 
> @dashrews78 This is a bit of a weird case. I feel like helm users should just overwrite the postgresConfig var, but we need a specific var for operator

They should, but nothing is preventing them from doing it this way.  Thought I suppose leaving it out of the examples file will possibly discourage such behavior.  So I withdraw that suggestion.

connorgorman:
@porridge If you could take another look that'd be great. Thanks!

image/templates/helm/stackrox-central/values.yaml.htpl
```diff
@@ -201,49 +201,54 @@
...
+#    # or left unset (null) and the Helm chart is being installed (as opposed to upgraded).
+#    # Should only be used when utilizing Postgres as central's DB
+#    password:
+#      value: null
+#      generate: null
+#
+#    postgresConfig: "@config/centraldb/postgresql.conf|config/centraldb/postgresql.conf.default"
+#    hbaConfig: "@config/centraldb/pg_hba.conf|config/centraldb/pg_hba.conf.default"
+#
+#    # Specifying configOverride mounts the specified config map in the same namespace which must contain
```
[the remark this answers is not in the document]
connorgorman: No, because I would expect someone using the public config to just modify the postgresql.conf directly

image/templates/helm/stackrox-central/values.yaml.htpl
```diff
@@ -201,49 +201,54 @@
...
+#    # See the documentation for `ca` at the top level for an explanation.
+#    serviceTLS:
+#      cert: null
+#      key: null
+#      generate: null
+#
+#    image:
+#      registry: null
+#      name: central-db
+#      tag: 3.0.48.x-121-g337915cd3d
```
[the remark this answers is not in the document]
connorgorman: I really don't think these are actually used unless specified because helm doesn't default to that version. That being said, I'll move them to 4.0.0 because it looks better

connorgorman:
@c-du @dashrews78 One more pass would be appreciated, need to go through a rebase

dashrews78 (approved):

PR #5247, opened 2023-03-15
Check public config on analytics events

bradr5:
## Description

Whenever analytics methods are called, this option checks the public config. If telemetry is disabled in the public config, events will not be sent to Segment.

Caveat: If telemetry collection is enabled when a user visits the page, Segment is initialized. If the user decides to opt out of analytics, events will no longer be sent to Segment, but Segment will still be set up on the page unless we manually remove the script tags and delete `window.analytics`, which is not recommended.

## Checklist

## Testing Performed

With the flag enabled, you can verify segment is initialized in various ways:
* check segments debugger and look for events
* see if window.analytics exists
* look for segment/amplitude script tags

Disabling telemetry in the system config should result in events not being sent to segment, however `window.analytics` will still exist and the script tags can still be viewed in the html


ui/apps/platform/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { enabled: telemetryEnabled } = telemetry || ({} as Telemetry);
+
+    const analyticsIdentity = useCallback(
+        (userId: string, traits = {}): void => {
+            if (telemetryEnabled) {
+                window.analytics?.identify(userId, traits);
+            }
+        },
+        [telemetryEnabled]
+    );
```
vjwilson, line 18: Is this recommended by Segment or Amplitude, and will we be using it in the near future?

vjwilson (approved):

ui/apps/platform/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback } from 'react';
+import { useSelector } from 'react-redux';
+import { Telemetry } from 'types/config.proto';
+
+import { selectors } from 'reducers';
+
+const useAnalytics = () => {
+    const telemetry = useSelector(selectors.publicConfigTelemetrySelector);
+    const { enabled: telemetryEnabled } = telemetry || ({} as Telemetry);
```
pedrottimark, line 9: Minor point of naming in comparison to the companion pull request where you followed our more common convention `isTelemetryEnabled` for a boolean variable.

ui/apps/platform/src/hooks/useAnalytics.ts
bradr5: The identity is being set up after initializing segment now. My thought was to keep this around as a basic util function but with it not being used I guess I might as well remove it

PR #5248, opened 2023-03-15
[ROX-15808] : Fix for watched images not getting scanned more than once

charmik-redhat:
## Description
In `pkg/images/enricher/enricher_impl.go`, func `EnrichImage` gets called for enriching watched images with option `IgnoreExistingImages`. When this option is set, we should not look for existing image in database to populate the image scan. However, we were not checking for that option in `updateImageFromDatabase` . Due to this, watched images were rescanned only when their SHA changed instead of getting rescanned every 4 hours. The fix is to check for `IgnoreExistingImages` in `updateImageFromDatabase` and return without doing db lookup if the option is set. This will force fresh scan, signature and signature verification of image every 4 hours.

## Checklist

## Testing Performed

Unit tests

Manual tests

FB6370EA-4692-4B42-8F85-4BC88B95C8A5_1_105_c

4A34361A-5102-4678-AEFC-49BB499123F7_1_105_c


md2119 (approved):
When reprocessing doesn't reprocess 😉

PR #5216, opened 2023-03-13
ROX-15886: Skip UI e2e tests that are not relevant for OpenShift

pedrottimark:
## Description

* Cluster/Platform might not have any CVEs let alone fixable CVEs.
* Node might not have any CVEs let alone fixable CVEs.
* Image might not have fixable CVEs: passed here but had failed in #4817

### Files changed

1. Edit cypress/helpers/features.js
    * Delete unused default export.
    * Add `hasOrchestratorFlavor` function.

2. Edit cypress/integration/vulnmanagement/clusterCves.test.js for 3 / 4 tests:
    * should sort the CVSS Score column
    * should display vulnerability descriptions
    * should display links for clusters

3. Edit cypress/integration/vulnmanagement/clusters.test.js for 3 / 9 tests:
    * should display links for fixable image CVEs
    * should display links for all node CVEs
    * should display links for all cluster CVEs

4. Edit cypress/integration/vulnmanagement/deployments.test.js for 1 / 5 tests:
    * should display links for fixable image CVEs

5. Edit cypress/integration/vulnmanagement/entitypages.test.js for 5 / 11 tests:
    * related entities tile links should unset search params upon navigation
    * related entities table header should not say "0 entities" or have "page 0 of 0" if there are rows in the table
    * should scope deployment data based on selected policy from table row click
    * should scope deployment data based on selected policy from entity page tab sublist
    * should not filter cluster entity page regardless of entity context

6. Edit cypress/integration/vulnmanagement/imageComponents.test.js for 1 / 7 tests:
    * should display links for fixable image CVEs

7. Edit cypress/integration/vulnmanagement/images.test.js for 1 / 9 tests:
    * should display links for fixable image CVEs and also Risk Acceptance tabs

8. Edit cypress/integration/vulnmanagement/namespaces.test.js for 1 / 6 tests:
    * should display links for fixable image CVEs

9. Edit cypress/integration/vulnmanagement/nodeComponents.test.js for 3 / 6 tests:
    * should sort the Top CVSS column
    * should display links for all node CVEs
    * should display links for fixable node CVEs

10. Edit cypress/integration/vulnmanagement/nodeCves.test.js for 4 / 6 tests:
    * should sort the CVSS Score column
    * should display vulnerability descriptions
    * should display links for nodes
    * should display links for node-components

11. Edit cypress/integration/vulnmanagement/policies.test.js for 1 / 4 tests:
    * should display links for failing deployments

12. Edit scripts/cypress.sh
[...]

PR #5251, opened 2023-03-15
Replace arrow function in conditional skip for vulnmanagement policies test

pedrottimark:
## Description

Oops, how did I miss the following error in #5216

> TypeError: Cannot read properties of undefined (reading 'skip')

For arrow function, `this` is at module scope, therefore `undefined` here!

```diff
-     it('should display links for failing deployments', () => {
+     it('should display links for failing deployments', function () {
         if (hasOrchestratorFlavor('openshift')) {
             this.skip();
         }
```

P.S. Although TypeScript would have reported this as an error, pardon pun, integration tests are in JavaScript. Sadly, ESLint rule `no-invalid-this` would have reported the correct `function` as an error, because makes such restrictive assumptions about where `this` is valid.

## Checklist
- [x] Edited integration test

## Testing Performed

### DIFF
--- a/ui/apps/platform/cypress/integration/vulnmanagement/policies.test.js
+++ b/ui/apps/platform/cypress/integration/vulnmanagement/policies.test.js
@@ -84,7 +84,7 @@ describe('Vulnerability Management Policies', () => {
 
     // Some tests might fail in local deployment.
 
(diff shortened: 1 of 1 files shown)


gavin-stackrox (approved):
`$self`

PR #5252, opened 2023-03-15
ROX-14674: Add baseline migration 

fredrb:
## Description

This is the follow-up changes needed for fixing the Network Graph issue ROX-14674 where baselines seem duplicate when they are created from two CIDR blocks from the same provider region. 

The code to consider CIDR blocks when creating baselines is here: #5194. 

This PR is here only for the migration part, in case there are already baselines created in the system. We need this migration otherwise any baselines created prior to the upgrade will not display any CIDR blocks in the UI. 

## Checklist

## Testing Performed

- [x] Migration unit test
- [x] E2E Tests


migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/stackrox/rox/migrator/migrations"
+	"github.com/stackrox/rox/migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkbaselinestore"
+	"github.com/stackrox/rox/migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkentitystore"
+	"github.com/stackrox/rox/migrator/types"
+	"github.com/stackrox/rox/pkg/postgres"
+	"gorm.io/gorm"
+)
+
+const (
+	startSeqNum = 175
```
dashrews78, line 17: should be 176

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration.go
```diff
@@ -0,0 +1,137 @@
...
+			ExternalSource: externalSource,
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+func addCIDRBlockToBaselines(postgresDB *postgres.DB, gormDB *gorm.DB) error {
+	ctx := context.Background()
+	networkBaselineStore := networkbaselinestore.CreateTableAndNewStore(ctx, postgresDB, gormDB)
```
dashrews78, line 82: We are not changing the schemas so we do not need to use `CreateTableAndNewStore`.  We simply need to get a new store as the tables will have already been created and populated from all the previous migrations.

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration.go
fredrb: Done

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration.go
```diff
@@ -0,0 +1,137 @@
...
+const (
+	startSeqNum = 175
+
+	batchSize = 500
+)
+
+var (
+	migration = types.Migration{
+		StartingSeqNum: startSeqNum,
+		VersionAfter:   &storage.Version{SeqNum: int32(startSeqNum + 1)},
```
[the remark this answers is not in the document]
fredrb: Done

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration.go
fredrb: That makes sense, thanks!

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration_test.go
```diff
@@ -0,0 +1,197 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	networkBaselines = []*storage.NetworkBaseline{
+		{
+			DeploymentId: dep1.String(),
+			ClusterId:    "c1",
```
dashrews78, line 96: I'm surprised you didn't have to make this one a uuid like deployment id.  FWIW, there are constants in `pkg/fixtures/fixtureconsts/fixture_consts.go` for such things.

migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkbaselinestore/postgres_plugin.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/central/metrics"
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	frozenSchemav73 "github.com/stackrox/rox/migrator/migrations/frozenschema/v73"
+	"github.com/stackrox/rox/pkg/logging"
+	ops "github.com/stackrox/rox/pkg/metrics"
+	"github.com/stackrox/rox/pkg/postgres"
+	"github.com/stackrox/rox/pkg/postgres/pgutils"
+	pkgSchema "github.com/stackrox/rox/pkg/postgres/schema"
```
dashrews78, line 18: Should not reference this package in a migration.  using the `frozenSchema73` would be what you need to use.

migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkentitystore/postgres_plugin.go
```diff
@@ -0,0 +1,407 @@
...
+	"github.com/hashicorp/go-multierror"
+	"github.com/jackc/pgx/v4"
+	"github.com/pkg/errors"
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/logging"
+	ops "github.com/stackrox/rox/pkg/metrics"
+	"github.com/stackrox/rox/pkg/postgres"
+	"github.com/stackrox/rox/pkg/postgres/pgutils"
+	pkgSchema "github.com/stackrox/rox/pkg/postgres/schema"
```
dashrews78, line 15: Shouldn't use this package in a migration.  frozenschema/v73 is probably what you need.

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration_test.go
fredrb: Didn't know about `fixtureconsts`. Updated all entries to use the constants.

migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkbaselinestore/postgres_plugin.go
fredrb: Done

migrator/migrations/m_176_to_m_177_network_baselines_cidr/networkentitystore/postgres_plugin.go
fredrb: Done

migrator/migrations/m_176_to_m_177_network_baselines_cidr/migration_test.go
dashrews78: interesting.  I guess that makes sense as in theory we could change a constant that causes an outdated migration to fail.  So disregard using that for migrations (but at least you know about it for when changing non-migration code)

fredrb:
Nongroovy E2E test failure is unrelated to change. Same test failure flake happened in other PRs: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests

dashrews78 (approved):

rhybrillou (approved):

PR #5253, opened 2023-03-15
ROX-15426: Call remote gRPC NodeScanner service

Maddosaurus:
## Description

This PR replaces the NodeInventorizer in Compliance with a minimal gRPC client.

The NodeInventorizer code will be moved into Scanner and is available as a new CLI flag to the `clair` binary of Scanner.
This PR depends on multiple things to work as intended without manual intervention:

1.  ~~In Scanner, this PR needs to be merged: https://github.com/stackrox/scanner/pull/1116~~
2. ~~Update dependency of Scanner in Stackrox: `go get https://github.com/stackrox/scanner@master`~~
3. Have an image of scanner running `--nodeinventory` available in the `Collector` DaemonSet (ROX-15425)

## Checklist
- [x] ~~Evaluated and added CHANGELOG entry if required~~
- [x] ~~Determined and documented upgrade steps~~
- [x] ~~Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))~~

## Testing Performed
To test this, currently some manual steps are needed.
1. Optionally, get an OpenShift 4 cluster up and running
2. Edit the Collector DaemonSet to add the built image from PR 1116 into it (see changes below for a copy-paste example)

Add this section to the Collector `DaemonSet` in the `containers` section:

```yaml
      - name: node-inventory
        image: quay.io/stackrox-io/scanner:2.28.x-53-gaabe03072a
        command: [ "/scanner", "--nodeinventory" ]
        ports:
        - containerPort: 8444
        env:
        - name: ROX_NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        volumeMounts:
        - mountPath: /host
          name: host-root-ro
          readOnly: true
        - mountPath: /tmp/
          name: tmp-volume
        - mountPath: /cache
          name: cache-volume
```

You can also set additional env variables to shorten the time between node scans to quicker see results:
`kubectl -n stackrox set env daemonsets/collector --containers="compliance" ROX_NODE_SCANNING_INTERVAL="20s" ROX_NODE_SCANNING_INTERVAL_DEVIATION="1s" ROX_NODE_SCANNING_MAX_INITIAL_WAIT="10s" LOGLEVEL="DEBUG" ROX_NODE_SCANNING_CACHE_TIME="40s" ROX_NODE_SCANNING_INITIAL_BACKOFF="2s"`


compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	timestamp "github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/uuid"
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+)
+
+// NodeInventoryResponseToNodeInventory converts a NodeInventory response to a native storage.NodeInventory version
+func NodeInventoryResponseToNodeInventory(response *scannerV1.GetNodeInventoryResponse) *storage.NodeInventory {
```
vikin91, line 11: How about the name: `ToNodeInventory(resp)`?

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		NodeId:     uuid.Nil.String(),
+		NodeName:   response.GetNodeName(),
+		ScanTime:   timestamp.TimestampNow(),
+		Components: inventoryComponentsToStorageComponents(response.GetComponents()),
+		Notes:      convertNotes(response.GetNotes()),
+	}
+	return ni
+}
+
+func inventoryComponentsToStorageComponents(c *scannerV1.Components) *storage.NodeInventory_Components {
```
vikin91, line 22: the same here, you would call that on a components object, so no need for the prefix `inventoryComponents` in the name. I would call it `toStorageComponents`

compliance/collection/main.go
```diff
@@ -156,7 +159,7 @@ func manageSendToSensor(ctx context.Context, cli sensor.ComplianceService_Commun
 	}
 }
 
-func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scanner nodeinventorizer.NodeInventorizer) <-chan *sensor.MsgFromCompliance {
+func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scanner scannerV1.NodeInventoryServiceClient) <-chan *sensor.MsgFromCompliance {
```
vikin91, line 162: If `scanner scannerV1.NodeInventoryServiceClient` is an interface, then all is good. If it is an implementation, I would pass it as a pointer `scanner *scannerV1.NodeInventoryServiceClient`

compliance/collection/main.go
```diff
@@ -256,14 +260,24 @@ func main() {
...
 
+	var niClient scannerV1.NodeInventoryServiceClient
+
 	if features.RHCOSNodeScanning.Enabled() {
 		// Start the prometheus metrics server
 		metrics.NewDefaultHTTPServer().RunForever()
 		metrics.GatherThrottleMetricsForever(metrics.ComplianceSubsystem.String())
+
+		// Set up Compliance <-> NodeInventory connection
+		niConn, err := grpc.Dial("127.0.0.1:8444", grpc.WithTransportCredentials(insecure.NewCredentials()))
```
vikin91, line 271: The port should be configurable over Helm with default `8444`, I think.

compliance/collection/main.go
```diff
@@ -183,14 +186,15 @@ func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scan
 	return sensorC
 }
 
-func scanNode(nodeName string, scanner nodeinventorizer.NodeInventorizer) (*sensor.MsgFromCompliance, error) {
-	result, err := scanner.Scan(nodeName)
+func scanNode(scanner scannerV1.NodeInventoryServiceClient) (*sensor.MsgFromCompliance, error) {
```
vikin91, line 189:
```suggestion
func scanNode(scanner *scannerV1.NodeInventoryServiceClient) (*sensor.MsgFromCompliance, error) {
```
if it is not an interface

compliance/collection/main.go
```diff
@@ -183,14 +186,15 @@ func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scan
...
+	result, err := scanner.GetNodeInventory(context.Background(), &scannerV1.GetNodeInventoryRequest{})
```
vikin91, line 190: Do we need that request param if it is always a zero-value? Could we maybe hide it inside `scanner.GetNodeInventory`?

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		case scannerV1.Note_OS_CVES_STALE:
+			convertedNotes = append(convertedNotes, storage.NodeInventory_OS_CVES_STALE)
+		case scannerV1.Note_LANGUAGE_CVES_UNAVAILABLE:
+			convertedNotes = append(convertedNotes, storage.NodeInventory_LANGUAGE_CVES_UNAVAILABLE)
+		case scannerV1.Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE:
+			convertedNotes = append(convertedNotes, storage.NodeInventory_CERTIFIED_RHEL_SCAN_UNAVAILABLE)
+		}
+	}
+	return convertedNotes
+}
```
vikin91, line 110: @jschnath this looks like a relevant place for your ticket ;)

vikin91 (commented):
Looks good!
I have mainly style comments after a quick look. I didn't have enough time to run that.

compliance/collection/main.go
Maddosaurus: It is an interface, so let's keep it as is 😄

compliance/collection/main.go
Maddosaurus: IIRC, this is a recommendation when defining Protobuf services.
By defining the service with a `Request` from the beginning, we don't break backwards compatibility if we ever introduce request parameters.
`Get NodeInventory` is already the generated protobuf function, so I don't know whether we would like to wrap that in another layer.

compliance/collection/inventory/convert.go
Maddosaurus: Good point, done!

compliance/collection/inventory/convert.go
Maddosaurus: Done!

compliance/collection/main.go
vikin91: Oh, so if `GetNodeInventory` is a proto-generated func, then all looks good.

compliance/collection/main.go
fredrb: > The port should be configurable over Helm with default 8444, I think.

Right now I don't think it can, no. Scanner usually opens two gRPC listeners, one [hardcoded](https://github.com/stackrox/scanner/blob/581c726c4295a330ff71c4a3ddd757e5e92c3286/api/grpc/grpc.go#L23) on port `8444` for internal connections, and another on a configurable port provided in the `config.yaml` file. 

Since both containers will share the same pod, they are able to access each other's ports even if they are not mapped in the container template (AFAIK). So defaulting to `8444` here isn't a problem. 

However, since it isn't too much of effort, I would already extract this into an external config (like an env variable). This way, if scanner gRPC code changes (e.g. due to clair unification) we don't need to change this code in the future.

compliance/collection/main.go
```diff
@@ -256,14 +262,25 @@ func main() {
...
+
 	if features.RHCOSNodeScanning.Enabled() {
 		// Start the prometheus metrics server
 		metrics.NewDefaultHTTPServer().RunForever()
 		metrics.GatherThrottleMetricsForever(metrics.ComplianceSubsystem.String())
+
+		// Set up Compliance <-> NodeInventory connection
+		niConn, err := clientconn.AuthenticatedGRPCConnection(nodeScannerEndpoint, mtls.Subject{}, clientconn.UseInsecureNoTLS(true))
+		if err != nil {
+			log.Fatal(err)
```
fredrb, line 275: This shouldn't happen. Compliance shouldn't go into crash loop if node inventory is not up yet or (more likely) misconfigured. If `ROX_NODE_NAME` is not set for node inventory, it will make compliance crash, which is the last thing we want. 

Even if you don't have the retry loop (yet), this should at least defer the start of the client until the next 4 hour timer kicks in (in manageNodeScanLoop).

compliance/collection/main.go
```diff
@@ -256,14 +262,25 @@ func main() {
 	// Set the random seed based on the current time.
 	rand.Seed(time.Now().UnixNano())
 
+	var niClient scannerV1.NodeInventoryServiceClient
```
fredrb, line 265: nit: `s/niClient/nodeInventoryClient`

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	timestamp "github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/uuid"
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+)
+
+// ToNodeInventory converts a NodeInventory response to a native storage.NodeInventory version
+func ToNodeInventory(resp *scannerV1.GetNodeInventoryResponse) *storage.NodeInventory {
+	ni := &storage.NodeInventory{
```
fredrb, line 12: why not?
```suggestion
	return &storage.NodeInventory{
```

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	var namespace string
+	if c.GetNamespace() == "" {
+		namespace = "unknown"
+	} else {
+		namespace = c.GetNamespace()
+	}
+
+	sc := &storage.NodeInventory_Components{
```
fredrb, line 34: Also here
```suggestion
	return &storage.NodeInventory_Components{
```

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func toStorageComponents(c *scannerV1.Components) *storage.NodeInventory_Components {
+	if c == nil {
+		return nil
+	}
+
+	var namespace string
+	if c.GetNamespace() == "" {
+		namespace = "unknown"
```
fredrb, line 29: Why setting namespace to a hardcoded value such as `unknown`? I'm not sure how these components are bound to namespaced resources, but global resources usually have empty (`""`) namespaces. And that's a valid value for this field in k8s.  

I think leaving the namespace with the exact same value when converting (even if empty) would make more sense. 

If there is a non obvious reason, could you add a comment about it then?

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return convertedComponents
+}
+
+func convertExecutables(e []*scannerV1.Executable) []*storage.NodeInventory_Components_RHELComponent_Executable {
+	if e == nil {
+		return nil
+	}
+	convertedExecutables := make([]*storage.NodeInventory_Components_RHELComponent_Executable, 0)
+	for _, ee := range e {
```
fredrb, line 69: if you make the change suggested above, then you can use the single letter var for the range loop here :-)

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return convertedExecutables
+}
+
+func convertRequiredFeatures(r []*scannerV1.FeatureNameVersion) []*storage.NodeInventory_Components_RHELComponent_Executable_FeatureNameVersion {
+	if r == nil {
+		return nil
+	}
+	convertedFeatures := make([]*storage.NodeInventory_Components_RHELComponent_Executable_FeatureNameVersion, 0)
+	for _, rr := range r {
```
fredrb, line 83: Same here for single letter variables

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		convertedFeatures = append(convertedFeatures, &storage.NodeInventory_Components_RHELComponent_Executable_FeatureNameVersion{
+			Name:    rr.GetName(),
+			Version: rr.GetVersion(),
+		})
+	}
+	return convertedFeatures
+}
+
+func convertNotes(n []scannerV1.Note) []storage.NodeInventory_Note {
```
fredrb, line 92: nit: legibility: `s/n/note`

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+	if n == nil {
+		return nil
+	}
+	convertedNotes := make([]storage.NodeInventory_Note, 0)
+	for _, nn := range n {
```
fredrb, line 97: And again here

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		convertedComponents = append(convertedComponents, &sc)
+	}
+	return convertedComponents
+}
+
+func convertExecutables(e []*scannerV1.Executable) []*storage.NodeInventory_Components_RHELComponent_Executable {
+	if e == nil {
+		return nil
+	}
+	convertedExecutables := make([]*storage.NodeInventory_Components_RHELComponent_Executable, 0)
```
fredrb, line 68: Why not making setting the length of the slice and avoid letting go expand the slice at runtime? 
```suggestion
	convertedExecutables := make([]*storage.NodeInventory_Components_RHELComponent_Executable, len(e))
```

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		})
+	}
+	return convertedExecutables
+}
+
+func convertRequiredFeatures(r []*scannerV1.FeatureNameVersion) []*storage.NodeInventory_Components_RHELComponent_Executable_FeatureNameVersion {
+	if r == nil {
+		return nil
+	}
+	convertedFeatures := make([]*storage.NodeInventory_Components_RHELComponent_Executable_FeatureNameVersion, 0)
```
fredrb, line 82: Use known slice length here as well

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+		})
+	}
+	return convertedFeatures
+}
+
+func convertNotes(n []scannerV1.Note) []storage.NodeInventory_Note {
+	if n == nil {
+		return nil
+	}
+	convertedNotes := make([]storage.NodeInventory_Note, 0)
```
fredrb, line 96: Use known slice length here as well

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return sc
+
+}
+
+func convertRHELComponents(rc []*scannerV1.RHELComponent) []*storage.NodeInventory_Components_RHELComponent {
+	if rc == nil {
+		return nil
+	}
+	convertedComponents := make([]*storage.NodeInventory_Components_RHELComponent, 0)
```
fredrb, line 47: Use known slice length here as well

compliance/collection/inventory/convert_test.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/stackrox/rox/pkg/uuid"
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestInventoryConvert(t *testing.T) {
+	suite.Run(t, new(InventoryConvertTestSuite))
+}
+
+type InventoryConvertTestSuite struct {
```
fredrb, line 16: I think this can be unexported

fredrb (changes requested):
I see two blockers:

1) I'm not confident on the behavior of client calls when the underlying connection was closed. If you've tested that and you're sure that the same client pointer will retry the connection once the server is available, then we should add a comment explaining that where the RPC call is made. Otherwise, I suggest to do what I described in the in-line comment and re-establish the gRPC client if we know it was closed.

2) When starting up compliance, we shouldn't panic if scanner isn't available. There are multiple reasons why it might not be available immediately: container takes longer to start, image not yet available, incorrect config etc... There should be some mechanism to either wait, or defer that the connection creation until the first RPC call has to be made. 

I've added some additional comments to improve the code legibility and other nits I found.

I only skimmed through the tests, I'll check them more thoroughly after these points are addressed.

fredrb:
Oh, and also: the PR (https://github.com/stackrox/scanner/pull/1116) in scanner landed. So you can do the `go.mod` update and no longer need to point to my branch.

compliance/collection/inventory/convert_test.go
fredrb: Not really sure why, but I've always made them unexported. Can't thing of why you'd need to use that struct outside of the package.

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,100 @@
...
+	timestamp "github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/uuid"
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+)
+
+// ToNodeInventory converts a NodeInventory response to a native storage.NodeInventory version
+func ToNodeInventory(resp *scannerV1.GetNodeInventoryResponse) *storage.NodeInventory {
+	return &storage.NodeInventory{
+		NodeId:     uuid.Nil.String(),
```
fredrb, line 13: I assume this is because `NodeId` will be filled only in sensor, with k8s data, right? If that's the case, do you mind adding a single line comment here saying that we add `Nil` uuid as a placeholder and the real ID will be set in sensor?

compliance/collection/main.go
```diff
@@ -252,18 +254,28 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
...
 		// Start the prometheus metrics server
 		metrics.NewDefaultHTTPServer().RunForever()
 		metrics.GatherThrottleMetricsForever(metrics.ComplianceSubsystem.String())
-	}
 
-	clientconn.SetUserAgent(clientconn.Compliance)
+		// Set up Compliance <-> NodeInventory connection
+		niConn, err := clientconn.AuthenticatedGRPCConnection(env.NodeScanningEndpoint.Setting(), mtls.Subject{}, clientconn.UseInsecureNoTLS(true))
+		if err != nil {
+			log.Errorf("Could not initialize connection to NodeInventory service. Node Scanning will be unavailable. Error: %v", err)
```
fredrb, line 272:
```suggestion
			log.Errorf("Could not initialize connection to NodeInventory service. Node Scanning will be unavailable: %v", err)
```

Nit: wrapped errors are formatted as `parent error: child error` without any (Error/Warn/Info) keywords. Makes it more idiomatic to log like IMO.

fredrb (changes requested):
Thanks for making all the changes requested!

I still think there still is a risk of panic when starting the gRPC connection. I left a comment inline here related to the `niConn` variable being nil, and also another comment in the previous review comment, replying to yours. 

Let me know what do you think.

compliance/collection/inventory/convert.go
Maddosaurus: That is correct. The `NodeId` is not available to Compliance or the new NodeScanner.
Sensor is the first component in the message chain that can correlate a `NodeName` to a `NodeID`.
I've added a comment that this is a placeholder, thanks!

compliance/collection/main.go
```diff
@@ -252,18 +254,28 @@ func initializeStream(ctx context.Context, cli sensor.ComplianceServiceClient) (
...
+		}
+		log.Info("Initialized NodeInventory gRPC connection")
+		nodeInventoryClient = scannerV1.NewNodeInventoryServiceClient(niConn)
```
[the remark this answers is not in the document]
Maddosaurus: Thank you very much for checking `Dial` and sharing 😄 
As discussed in the sync, let's err on stability and just disable the feature completely if we run into a `nil`.

compliance/collection/inventory/convert.go
```diff
@@ -0,0 +1,100 @@
...
+	timestamp "github.com/gogo/protobuf/types"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/uuid"
+	scannerV1 "github.com/stackrox/scanner/generated/scanner/api/v1"
+)
+
+// ToNodeInventory converts a NodeInventory response to a native storage.NodeInventory version
+func ToNodeInventory(resp *scannerV1.GetNodeInventoryResponse) *storage.NodeInventory {
+	return &storage.NodeInventory{
+		NodeId:     uuid.Nil.String(), // NodeId is only available to Sensor. Until it arrives there, we add a placeholder
```
fredrb, line 13: nit
```suggestion
		// NodeId is only available to Sensor. Until it arrives there, we add a placeholder
		NodeId:     uuid.Nil.String(),
```

compliance/collection/main.go
```diff
@@ -288,24 +302,8 @@ func main() {
...
 		i := intervals.NewNodeScanIntervalFromEnv()
-		scanner := nodeinventorizer.NewCachingScanner(
-			analyzer,
-			"/cache/inventory-cache",
-			env.NodeScanningCacheDuration.DurationSetting(),
-			env.NodeScanningInitialBackoff.DurationSetting(),
-			env.NodeScanningMaxBackoff.DurationSetting(),
-			func(duration time.Duration) { time.Sleep(duration) })
-		nodeInventoriesC := manageNodeScanLoop(ctx, i, scanner)
+		nodeInventoriesC := manageNodeScanLoop(ctx, i, nodeInventoryClient)
```
fredrb, line 306: Now the issue is here. You guarded the assignment of `nodeInventoryClient`. It will be passed here as `nil` if it wasn't set before. 

I think you need to guard this call (and the goroutine below) if `nodeInventoryClient` is not `nil`.

```go
if nodeInventoryClient != nil {
		nodeInventoriesC := manageNodeScanLoop(ctx, i, nodeInventoryClient)
		go func() { /* ... */ }
}
```

Otherwise the panic will still happen since a `nil` `nodeInventoryClient` will be passed to `manageNodeScanLoop`. 

Alternatively, you can add the guard check inside of `manageNodeScanLoop`. Up to you.

Maddosaurus: Thank you, I completely overlooked the call to `manageNodeScanLoop`.
I have moved the additional checking into the function, taking care to return the channel either case, so it cannot break the `go func` that multiplexes producers.

compliance/collection/main.go
```diff
@@ -156,9 +157,13 @@ func manageSendToSensor(ctx context.Context, cli sensor.ComplianceService_Commun
...
 }
 
-func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scanner nodeinventorizer.NodeInventorizer) <-chan *sensor.MsgFromCompliance {
+func manageNodeScanLoop(ctx context.Context, i intervals.NodeScanIntervals, scanner scannerV1.NodeInventoryServiceClient) <-chan *sensor.MsgFromCompliance {
 	sensorC := make(chan *sensor.MsgFromCompliance)
 	nodeName := getNode()
+	if scanner == nil {
+		log.Error("NodeInventoryServiceClient not provided. Node Scanning will be unavailable.")
+		return sensorC
+	}
```
vikin91, line 166: If we land in this branch, we need to close the chan.

```suggestion
	if scanner == nil {
		close(sensorC)
		log.Error("NodeInventoryServiceClient not provided. Node Scanning will be unavailable.")
		return sensorC
	}
```

compliance/collection/main.go
fredrb: The way this is now we are leaking the go routine. The channel is created, but there are no producers attached to it. Also, we are blocking an entire go routine (multiplexer below) waiting for a message that will never arrive.

vikin91 (approved):
LGTM

fredrb (approved):
🚢

PR #5256, opened 2023-03-15
Fix feature_flag test by accounting for active vuln mgmt

connorgorman:
## Description

Feature flag test broke when I added active vuln mgmt to the flags returned to the UI

## Checklist

## Testing Performed

Test should pass? Not sure how to get it to run actually

### DIFF
--- a/tests/feature_flag_test.go
+++ b/tests/feature_flag_test.go
@@ -47,6 +47,7 @@ func TestFeatureFlagSettings(t *testing.T) {
 
 	// TODO(ROX-14939): Remove this with environment variable
 	delete(actualFlagVals, env.PostgresDatastoreEnabled.EnvVar())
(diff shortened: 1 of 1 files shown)


tests/feature_flag_test.go
```diff
@@ -47,6 +47,7 @@ func TestFeatureFlagSettings(t *testing.T) {
 
 	// TODO(ROX-14939): Remove this with environment variable
```
theencee, line 48: [nit] Can you make the comment generic?

theencee (approved):

tests/feature_flag_test.go
connorgorman, line 48:
```suggestion
// TODO(ROX-14939): Refactor feature flag logic to include environment variables
```

PR #1328, opened 2022-04-18
Edit cve proto and add node and vulnerability proto to types in UI

pedrottimark:
## Description

Related to #1273 and #1275 which adds to proto/storage/node.proto

```proto
    enum Note {
        UNSET              = 0;
        UNSUPPORTED        = 1;
        KERNEL_UNSUPPORTED = 2;
    }
```

So there is a known file where to add the corresponding string union type:

```ts
export NodeScanNote = 'UNSET' | 'UNSUPPORTED' | 'KERNEL_UNSUPPORTED';
```

@sachaudh @vjwilson

TypeScript TS2741 Property 'UNKNOWN_VULNERABILITY_SEVERITY' is missing

* https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/constants/reportConstants.ts#L4-L11
* https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/messages/common.ts#L17-L22

Is there any reason not to include a property for this key?

For example, in the following classic code, `switch` statement has a default case:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Components/CVSSSeverityLabel.js#L7-L20

See below that `"severity": "UNKNOWN_VULNERABILITY_SEVERITY"` can occur in response to /v1/nodes/:clusterId request, although UI uses graphQL queries instead.

```js
          {
            "name": "kernel",
            "version": "5.10.25-linuxkit",
            "vulns": [
              {
                "cve": "CVE-2018-1000026",
                "cvss": 7.7,
                "summary": "Linux Linux kernel version at least v4.8 onwards, probably well before contains a Insufficient input validation vulnerability in bnx2x network card driver that can result in DoS: Network card firmware assertion takes card off-line. This attack appear to be exploitable via An attacker on a must pass a very large, specially crafted packet to the bnx2x card. This can be done from an untrusted guest VM..",
                "link": "https://nvd.nist.gov/vuln/detail/CVE-2018-1000026",
                "scoreVersion": "V3",
                "cvssV2": {
                  "vector": "AV:N/AC:L/Au:S/C:N/I:N/A:C",
                  "attackVector": "ATTACK_NETWORK",
                  "accessComplexity": "ACCESS_LOW",
                  "authentication": "AUTH_SINGLE",
                  "confidentiality": "IMPACT_NONE",
                  "integrity": "IMPACT_NONE",
                  "availability": "IMPACT_COMPLETE",
                  "exploitabilityScore": 8,
[...]

PR #5258, opened 2023-03-15
Add pagination and sorting to image single table

dvail:
## Description

Adds server side sorting and pagination to the image single page table.

## Follow ups

- Add appropriate sorting for all fields on the BE is complete and documented in the collab doc

## Checklist

## Testing Performed

Visit an image single page and verify that a single page worth's of data has loaded. The default sorting of the table should be by Severity, descending.
image

Use the pagination forward/back buttons to verify that pages of data are changed correctly.
image

Update the number of items per page and verify that the number of items is changed correctly.
image

Test that sorting by CVE, Severity, and CVE Status work correctly.
image
image
image
image
image

Verify that the CVEs by severity and CVEs by status cards have a total count matching that displayed at the top of the table.

### DIFF
--- a/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
+++ b/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
@@ -11,6 +11,7 @@ import {
     GridItem,
     Label,
     PageSection,
+    Pagination,
     pluralize,
     Spinner,
     Split,
@@ -28,52 +29,24 @@ import { VulnerabilitySeverity } from 'types/cve.proto';
 import { getAxiosErrorMessage } from 'utils/responseErrorUtils';
 import useURLStringUnion from 'hooks/useURLStringUnion';
 import useURLSearch from 'hooks/useURLSearch';
+import useURLPagination from 'hooks/useURLPagination';
+import useURLSort, { UseURLSortProps } from 'hooks/useURLSort';
 import { getHasSearchApplied } from 'utils/searchUtils';
 import { cveStatusTabValues, FixableStatus } from './types';
 import WorkloadTableToolbar from './WorkloadTableToolbar';
 import BySeveritySummaryCard from './SummaryCards/BySeveritySummaryCard';
 import CvesByStatusSummaryCard from './SummaryCards/CvesByStatusSummaryCard';
 import SingleEntityVulnerabilitiesTable from './Tables/SingleEntityVulnerabilitiesTable';
-import useImageVulnerabilities, {
-    ImageVulnerabilitiesResponse,
-} from './hooks/useImageVulnerabilities';
 import { ImageDetailsResponse } from './hooks/useImageDetails';
+import useImageVulnerabilities from './hooks/useImageVulnerabilities';
 
-function severityCountsFromImageVulnerabilities(
-    imageVulnerabilities: ImageVulnerabilitiesResponse['image']['imageVulnerabilities']
-): Record<VulnerabilitySeverity, number> {
-    const severityCounts = {
-        LOW_VULNERABILITY_SEVERITY: 0,
-        MODERATE_VULNERABILITY_SEVERITY: 0,
-        IMPORTANT_VULNERABILITY_SEVERITY: 0,
-        CRITICAL_VULNERABILITY_SEVERITY: 0,
-    };
-
-    imageVulnerabilities.forEach(({ severity }) => {
-        severityCounts[severity] += 1;
-    });
-
-    return severityCounts;
-}
-
-function statusCountsFromImageVulnerabilities(
-    imageVulnerabilities: ImageVulnerabilitiesResponse['image']['imageVulnerabilities']
-): Record<FixableStatus, number> {
-    const statusCounts = {
-        Fixable: 0,
-        'Not fixable': 0,
-    };
-
-    imageVulnerabilities.forEach(({ isFixable }) => {
-        if (isFixable) {
-            statusCounts.Fixable += 1;
-        } else {
-            statusCounts['Not fixable'] += 1;
-        }
-    });
-
-    return statusCounts;
-}
+const defaultSortOptions: UseURLSortProps = {
+    sortFields: ['CVE', 'Severity', 'Fixable'],
+    defaultSortOption: {
+        field: 'Severity',
+        direction: 'desc',
+    },
+};
 
 export type ImageSingleVulnerabilitiesProps = {
     imageId: string;
@@ -82,12 +55,22 @@ export type ImageSingleVulnerabilitiesProps = {
 
 function ImageSingleVulnerabilities({ imageId, imageData }: ImageSingleVulnerabilitiesProps) {
     const { searchFilter } = useURLSearch();
+    const { page, perPage, setPage, setPerPage } = useURLPagination(50);
+    // TODO Need to reset current page at the same time sorting changes
+    const { sortOption, getSortParams } = useURLSort(defaultSortOptions);
     // TODO Still need to properly integrate search filter with query
-    const { data, loading, error } = useImageVulnerabilities(imageId, {});
+    const pagination = {
+        offset: (page - 1) * perPage,
+        limit: perPage,
+        sortOption,
+    };
+    const { data, previousData, loading, error } = useImageVulnerabilities(imageId, {}, pagination);
 
     const [activeTabKey, setActiveTabKey] = useURLStringUnion('cveStatus', cveStatusTabValues);
 
     let mainContent: ReactNode | null = null;
+
+    const vulnerabilityData = data || previousData;
 
     if (error) {
         mainContent = (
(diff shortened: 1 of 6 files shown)


dvail:
* master
  * **PR #5258** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5258" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5258?utm_source=stack-comment).

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
```diff
@@ -28,65 +29,47 @@ import { VulnerabilitySeverity } from 'types/cve.proto';
...
 import WorkloadTableToolbar from './WorkloadTableToolbar';
 import BySeveritySummaryCard from './SummaryCards/BySeveritySummaryCard';
 import CvesByStatusSummaryCard from './SummaryCards/CvesByStatusSummaryCard';
 import SingleEntityVulnerabilitiesTable from './Tables/SingleEntityVulnerabilitiesTable';
-import useImageVulnerabilities, {
-    ImageVulnerabilitiesResponse,
-} from './hooks/useImageVulnerabilities';
+import useImageVulnerabilities from './hooks/useImageVulnerabilities';
 
-function severityCountsFromImageVulnerabilities(
```
dvail, line 41: This is from a proof-of-concept of client side pagination. Since we are not using client-side pagination, we can get this data directly from the query.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
```diff
@@ -28,65 +29,47 @@ import { VulnerabilitySeverity } from 'types/cve.proto';
...
+        limit: perPage,
+        sortOption,
+    };
+    const { data, previousData, loading, error } = useImageVulnerabilities(imageId, {}, pagination);
 
     const [activeTabKey, setActiveTabKey] = useURLStringUnion('cveStatus', cveStatusTabValues);
 
     let mainContent: ReactNode | null = null;
 
+    const vulnerabilityData = data || previousData;
```
pedrottimark, line 71: Is it possible to replace `||` with `??` nullish coalescing operator (what a name)

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
```diff
@@ -100,34 +83,35 @@ function ImageSingleVulnerabilities({ imageId }: ImageSingleVulnerabilitiesProps
...
-    } else if (data) {
-        const vulnerabilities = data.image.imageVulnerabilities;
-        const severityCounts = severityCountsFromImageVulnerabilities(vulnerabilities);
-        const cveStatusCounts = statusCountsFromImageVulnerabilities(vulnerabilities);
+    } else if (vulnerabilityData) {
+        const vulnerabilities = vulnerabilityData.image.imageVulnerabilities;
+
         // TODO Integrate these with page search filters
         const hiddenSeverities = new Set<VulnerabilitySeverity>([]);
         const hiddenStatuses = new Set<FixableStatus>([]);
```
pedrottimark, line 97: Can you share this at next team meeting?

Map and Set remove a security concern about overwriting Object properties.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/BySeveritySummaryCard.tsx
```diff
@@ -5,19 +5,26 @@ import SeverityIcons from 'Components/PatternFly/SeverityIcons';
...
-    'LOW_VULNERABILITY_SEVERITY',
-] as const;
+const vulnCounterToSeverity: Record<ImageVulnerabilityCounterKey, VulnerabilitySeverity> = {
+    low: 'LOW_VULNERABILITY_SEVERITY',
+    moderate: 'MODERATE_VULNERABILITY_SEVERITY',
+    important: 'IMPORTANT_VULNERABILITY_SEVERITY',
+    critical: 'CRITICAL_VULNERABILITY_SEVERITY',
+} as const;
+
+const severitiesCriticalToLow = ['critical', 'important', 'moderate', 'low'] as const;
```
pedrottimark, line 27: Reader observations, not necessarily change request:

1. Make names more parallel `severitiesCriticalToLow` here and the following in hook?

    ```ts
    export const imageVulnerabilityCounterKeys = ['low', 'moderate', 'important', 'critical'] as const;
   ```

2. Is the LowToCritical order ever used? At least some occurrences work either way, correct?
    * `imageVulnerabilityCounterKeys.forEach(…)`
    * `severitiesCriticalToLow.forEach(…)`

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/BySeveritySummaryCard.tsx
```diff
@@ -34,11 +41,12 @@ function BySeveritySummaryCard({
                 <Grid className="pf-u-pl-sm">
                     {severitiesCriticalToLow.map((severity) => {
                         const count = severityCounts[severity];
-                        const hasNoResults = count === 0;
-                        const isHidden = hiddenSeverities.has(severity);
+                        const hasNoResults = count.total === 0;
+                        const vulnSeverity = vulnCounterToSeverity[severity];
```
pedrottimark, line 45: Can you write a short comment (maybe above at `vulnCounterToSeverity` object) about the uses of the two severity keys?

Or, dare I ask, are the classic `'LOW_VULNERABILITY_SEVERITY'` keys used in any query payloads or responses? At least in this immediate context, the keys seem to be under frontend control:
* `vulnerabilitySeverityLabels`
* `SeverityIcons`

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/CvesByStatusSummaryCard.tsx
```diff
@@ -14,15 +18,25 @@ const statusDisplays = [
...
         Icon: CheckCircleIcon,
         iconColor: 'var(--pf-global--success-color--100)',
-        text: (counts: CvesByStatusSummaryCardProps['cveStatusCounts']) =>
-            `${counts.Fixable} vulnerabilities with available fixes`,
+        text: (counts: ImageVulnerabilityCounter) => {
+            let count = 0;
+            imageVulnerabilityCounterKeys.forEach((key) => {
+                count += counts[key].fixable;
+            });
+            return `${count} vulnerabilities with available fixes`;
```
pedrottimark, line 26: Is `pluralize` is relevant for **vulnerabilities** in these template literals?

pedrottimark (approved):
Another good-sized step forward. Some comments to consider.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/BySeveritySummaryCard.tsx
dvail: The idea here was to create a local variable that enforced the order specific to this component. I had thought about just making the default order in the hook CriticalToLow, but didn't want to have reliance on an order that is pretty much arbitrary in all other use cases.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
dvail: Sure and I can share the `useSet` hook since it might be of use in other places in the app.

What's the security concern you are talking about in this context?

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/BySeveritySummaryCard.tsx
dvail: Also, related, I found that we removed the `UNKNOWN_VULNERABILITY_SEVERITY` value from the front end type a while back. It looks like this is a valid response anywhere we could expect a severity from the BE though, and I've had a couple of cases of this in testing.

Do you see any reason to not add this back in? Possibly a discussion for the team meeting too.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageSingleVulnerabilities.tsx
pedrottimark: Whether possible (see below) or implausible, huge number of reported vulnerabilities are about possible object prototype pollution. Map and Set remove it as even a theoretical risk.

https://github.com/ljharb/qs/pull/428

```js
} else if (cleanRoot !== '__proto__') {
    obj[cleanRoot] = leaf;
}
```

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/SummaryCards/BySeveritySummaryCard.tsx
dvail: There was some light discussion in [this PR](https://github.com/stackrox/stackrox/pull/1328) regarding `UNKNOWN` but I didn't see anything beyond that. I think adding it back will result in very few compile errors, which we can solve by adding empty strings in the worst case.

As for duplications:

It looks like the `cve.proto.ts` type `VulnerabilitySeverity` is referenced 24 times, mostly by other types/utils, which are then imported primarily (if not entirely) in VulnMgmt 1.0, Vuln Reporting 1.0, and Risk Acceptance.

We also have `messages/common.ts` with a const `severityRatings` that is used in Policies. This value uses the severity shorthand version. `vulnSeverityLabels` in the same file maps the long version to the shorthand, which I wouldn't really think is a problem but might be a good canonical value->display mapper.

`severityColors.js` has _many_ different versions that map some severity to a color, I think mixing vuln severity with policy severity. It also contains a comment at the top `//  @TODO: Have one source of truth for severity colors` from 4 years ago 😄 

The new Workload CVEs intermixes the short style with the long style.

There are some .js files that use values like `MODERATE_VULNERABILITY_SEVERITY` and `Moderate` to refer to CVEs, but I'm less worried about those since they aren't typed and will either fall off with time, or become typed if they hang around.

There are also gql queries that reference resolvers like `low`, `moderate`, etc. that semantically seem like the same thing. We could alias those easily client side if we wanted, but we don't right now.

My general feeling is that "CVE Severity" is a concept that is used in enough places that it might be useful to keep it in a global `types` area. I'd be in favor of getting everything aligned with the current `VulnerabilitySeverity` since it is a direct mapping to the typed representation on the backend and would result in the least amount of moving code short-term. (Although too bad the API contract doesn't declare it, and we have to optimistically assume a `string` will be one of these values.)

***

Side note: Is `types/node.proto.ts` dead code? None of the types seem to be imported elsewhere. Transitively it looks like `types/vulnerability.proto.ts` might be dead code as well.

PR #5260, opened 2023-03-15
Run `junit2jira` on unittests

porridge:
## Description

## Checklist
- [ ] Prevent filing issues on PR runs
- [ ] package in a nice action that cares about fetching the binary
- [ ] make sure to bump the version in both places on upgrades
- [ ] fetch the binary to a temporary dir rather than clutter `$PWD`
- [ ] add invocations in all places that need it

## Testing Performed

TODO(replace-me)
screenshots.


janisz (commented):
Maybe it's worth to create GithubAction that could be used after junit job just like `test-summary/action`. Then we don't need artifacts and yaml will be simpler.

porridge:
> Maybe it's worth to create GithubAction that could be used after junit job just like `test-summary/action`. Then we don't need artifacts and yaml will be simpler.

That was my initial approach. But then I realized that if we had multiple `junit2jira` instances running in parallel on very similar jobs (like the ones spawned from the matrix), then they are likely to create duplicate JIRA issues.

This is why I decided to switch to a single invocation that serializes and thus de-duplicates failures.

Of course we can also have an action to encapsulate this, but I'd wait with this part until this is proven to work.

porridge:
FTR, a quick brain dump of things I found so far:
- I think `BaseLink` can be created as `GITHUB_SERVER_URL/GITHUB_REPOSITORY/compare/GITHUB_BASE_REF...GITHUB_HEAD_REF` on GHA.
- we could use `$GITHUB_ACTIONS` which is set to `true` on GHA to detect which CI system we're running on.

janisz:
I think we can close it
- https://github.com/stackrox/stackrox/pull/5675

PR #5264, opened 2023-03-15
ROX-13658: Add first basic e2e tests for PF Network Graph

vjwilson:
## Description

This adds the first basic e2e tests for the new PatternFly Network Graph

1. Navigate to the new graph from the left nav
2. Navigate directly via the URL
3. Select the first cluster in the list and the "stackrox" namespace, and verify that NS group and label are on the graph.

Note: I did not attempt to change the paradigm of the tests from the existing Network Graph test (reusing old selectors and constants where possible). Likewise, I did not try to make the new selectors reusable yet, until we get more tests written.

## Checklist
- [x] Investigated and inspected CI test results (failures in e2e-ui tests are flakes)

## Testing Performed

* Inspect CI results
* Ran test locally

Screen Shot 2023-03-15 at 5 55 41 PM

### DIFF
--- a/ui/apps/platform/cypress/constants/NetworkPage.js
+++ b/ui/apps/platform/cypress/constants/NetworkPage.js
@@ -64,6 +64,10 @@ export const selectors = {
         namespaceSelect: '.namespace-select > button',
         filterSelect: search.multiSelectInput,
     }),
+    selector: scopeSelectors('[data-testid="network-graph-selector-bar"]', {
+        clusterSelect: '.cluster-select > button',
+        namespaceSelect: '.namespace-select > button',
+    }),
     errorOverlay: {
         heading: 'h2:contains("An error has prevented the Network Graph from loading")',
         message: (messageText) =>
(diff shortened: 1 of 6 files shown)


ui/apps/platform/cypress/constants/NetworkPage.js
```diff
@@ -64,6 +64,10 @@ export const selectors = {
         namespaceSelect: '.namespace-select > button',
         filterSelect: search.multiSelectInput,
     }),
+    selector: scopeSelectors('[data-testid="network-graph-selector-bar"]', {
```
sachaudh, line 67: I'm now confused by the interchangeable `selector` vs. `filter` terminology. If we're calling the namespace containers in the Topology `filtered namespaces` should we not use that term for these too? We should eventually revisit this

sachaudh (approved):
Good first step. Thanks for starting it off

ui/apps/platform/cypress/constants/NetworkPage.js
vjwilson: I agree it's confusing. I landed on "selector" for now, because of these 2 reasons.

1. We still have the search bar below it, with our traditional key-value search, which we call our "filter" bar.
2. If we updated to similar functionality using PatternFly paradigms, they refer to those as "filters".

Perhaps we coin a new term for this cluster/namespace/deployment pattern, like for instance "hierarchy" ?

PR #5265, opened 2023-03-16
Jv rox 15995 plop service should report endpoint when process indicator is not found

JoukoVirtanen:
## Description

The processes listening on ports service should report listening endpoints even if there is not matching process indicator. It is better to in these cases provide partial information about the process associated with a listening endpoint than to mislead by not reporting the listening endpoint or any of the available process information.

## Checklist
- [ ] Ran e2e tests 200 times
- [ ] Performance testing
- [ ] Load testing


central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -61,13 +61,20 @@ func (ds *datastoreImpl) AddProcessListeningOnPort(
 	} else if !ok {
 		return sac.ErrResourceAccessDenied
 	}
+	for _, plop := range portProcesses {
+		log.Infof("portProcess: %+v", plop)
+	}
 
 	normalizedPLOPs, completedInBatch := normalizePLOPs(portProcesses)
+	allNormalizedPLOPs := append(normalizedPLOPs, completedInBatch...)
```
JoukoVirtanen, line 69: See https://github.com/stackrox/stackrox/pull/5204 to understand why this and the following change was made.

central/processlisteningonport/store/postgres/full_store.go
```diff
@@ -34,9 +34,9 @@ type fullStoreImpl struct {
 const getByDeploymentStmt = "SELECT plop.id, plop.serialized, " +
 	"proc.serialized as proc_serialized " +
 	"FROM process_listening_on_ports plop " +
-	"JOIN process_indicators proc " +
+	"LEFT OUTER JOIN process_indicators proc " +
 	"ON plop.processindicatorid = proc.id " +
-	"WHERE proc.deploymentid = $1 AND plop.closed = false"
+	"WHERE plop.deploymentid = $1 AND plop.closed = false"
```
JoukoVirtanen, line 39: We need to use the deploymentid from plop as proc may be NULL as a result of the left outer join.

proto/storage/process_listening_on_port.proto
```diff
@@ -57,4 +58,6 @@ message ProcessListeningOnPortStorage {
     // ProcessIndicator will be not empty only for those cases when we were not
     // able to find references process in the database
     ProcessIndicatorUniqueKey process = 7;
+
+    string deployment_id = 8 [(gogoproto.moretags) = 'search:"Cluster ID,store"'];
```
JoukoVirtanen, line 62: deployment_id is needed here as we now need the plop record to have the deployment_id.

erthalion: I guess the name should be `Deployment ID`, fortunately there is a such search category.

central/processlisteningonport/datastore/datastore_impl.go
erthalion: #5204 is about fixing a lookup for process indicators with empty query. Since there is no lookup, why is it needed?

JoukoVirtanen: The query returns an error if it is empty. Without this change we get empty queries in the unit tests.

proto/storage/process_listening_on_port.proto
```diff
@@ -57,4 +58,8 @@ message ProcessListeningOnPortStorage {
     // ProcessIndicator will be not empty only for those cases when we were not
     // able to find references process in the database
     ProcessIndicatorUniqueKey process = 7;
+
+    //  string deployment_id = 8 [(gogoproto.moretags) = 'search:"Deployment ID,store,hidden" policy:",prefer-parent" sql:"index=hash,fk(Deployment:id),no-fk-constraint,type(uuid)"'];
+    //string deployment_id = 8 [(gogoproto.moretags) = 'search:"Deployment ID,store" sql:"index=hash,type(uuid)"'];
+    string deployment_id = 8 [(gogoproto.moretags) = 'search:"Deployment ID,store" sql:"type(uuid)"'];
```
JoukoVirtanen, line 64: I am not confidant about this.

qa-tests-backend/src/test/groovy/BaseSpecification.groovy
```diff
@@ -37,7 +37,7 @@ import spock.lang.Retry
 import spock.lang.Shared
 import spock.lang.Specification
 
-@Retry(condition = { Helpers.determineRetry(failure) })
+//@Retry(condition = { Helpers.determineRetry(failure) })
```
JoukoVirtanen, line 40: Uncomment before merging

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -132,8 +133,15 @@ func (i *processListeningIndicator) toProto(ts timestamp.MicroTS) *storage.Proce
...
 			ProcessArgs:         i.key.process.processArgs,
 		},
+		DeploymentId:   i.key.deploymentID,
 	}
 
+	if ts != timestamp.InfiniteFuture {
+		proto.CloseTimestamp = ts.GogoProtobuf()
+	}
+
+	log.Infof("In toProto %+v", proto)
```
JoukoVirtanen, line 143: Remove before merging

central/processlisteningonport/datastore/datastore_impl.go
```diff
@@ -96,8 +97,11 @@ func (ds *datastoreImpl) AddProcessListeningOnPort(
 
 	plopObjects := []*storage.ProcessListeningOnPortStorage{}
 	for _, val := range normalizedPLOPs {
+		log.Infof("At top of loop val= %+v", val)
 		indicatorID := getIndicatorIDForPlop(val)
+		log.Infof("indicatorID= %s", indicatorID)
 		plopKey := getPlopKeyFromParts(val.GetProtocol(), val.GetPort(), indicatorID)
+		log.Infof("plopKey= %s", plopKey)
```
JoukoVirtanen, line 104: Remove new logging before merging.

proto/storage/process_listening_on_port.proto
JoukoVirtanen: Done, but I am not 100% that the change I made is correct.

proto/storage/process_listening_on_port.proto
dashrews78: hash index might make sense on that column

erthalion: I'm afraid this is not immediately obvious. Hash index maybe good for highly unique values, but here `deployment_id` will most likely be quite repeating. The docs [[1]] are saying:

> As a result of the overflow cases, we can say that hash indexes are most suitable for unique, nearly unique data or data with a low number of rows per hash bucket. One possible way to avoid problems is to exclude highly non-unique values from the index using a partial index condition, but this may not be suitable in many cases.

As a small and dirty experiment I've created two tables with only one uuid column, one table having a hash index, another a btree. Then retail inserting 300k records with only 3 distinct values (so, a lot of repetition) shows this:

```
-- insert into the table with btree index
=# insert into test_index_btree select
   case
     when i % 3 = 0 then '1b7e7c08-4c69-4189-9e7d-7474c872f973'::uuid
     when i % 3 = 1 then '98034214-0ad9-41da-a6f8-fefe37224c7b'::uuid
     when i % 3 = 2 then '2f75ed5a-fdbc-454a-8a0b-d22c441583d0'::uuid
   end
   from generate_series(1, 300000) i;

INSERT 0 300000
Time: 1722.175 ms (00:01.722)

-- insert into the table with hash index, somehow slower
=# insert into test_index_hash select
   case
     when i % 3 = 0 then '1b7e7c08-4c69-4189-9e7d-7474c872f973'::uuid
	 when i % 3 = 1 then '98034214-0ad9-41da-a6f8-fefe37224c7b'::uuid
	 when i % 3 = 2 then '2f75ed5a-fdbc-454a-8a0b-d22c441583d0'::uuid
   end
   from generate_series(1, 300000) i;

INSERT 0 300000
Time: 14444.927 ms (00:14.445)

-- size of the involved relations and indexes,
-- hash is almost the same size as the table,
-- where btree is smaller.

             relation             |    size
----------------------------------+------------
 public.test_index_hash_data_idx  | 14 MB
 public.test_index                | 13 MB
 public.test_index_btree          | 13 MB
 public.test_index_btree_data_idx | 1928 kB

```

I'm not saying it's a definitive proof, but probably makes sense to go with btree first, while doing perf tests to compare with hash on the real data distributions. 

[1]: https://www.postgresql.org/docs/current/hash-intro.html

dashrews78: sounds reasonable

central/processlisteningonport/store/postgres/full_store.go
JoukoVirtanen: Try to optimize query.

proto/storage/process_listening_on_port.proto
```diff
@@ -57,4 +58,8 @@ message ProcessListeningOnPortStorage {
     // ProcessIndicator will be not empty only for those cases when we were not
     // able to find references process in the database
     ProcessIndicatorUniqueKey process = 7;
+
+    //  string deployment_id = 8 [(gogoproto.moretags) = 'search:"Deployment ID,store,hidden" policy:",prefer-parent" sql:"index=hash,fk(Deployment:id),no-fk-constraint,type(uuid)"'];
+    //string deployment_id = 8 [(gogoproto.moretags) = 'search:"Deployment ID,store" sql:"index=hash,type(uuid)"'];
```
JoukoVirtanen, line 63: Ask Mandar about this.

central/processlisteningonport/store/postgres/full_store.go
erthalion: Again, first show there are any concerns exist about this query.

erthalion (approved):
Looks fine. Please update the PR description and the final squash commit to include the reasoning behind this change.

central/processlisteningonport/store/postgres/full_store.go
dashrews78: agree with @erthalion the query isn't horribly complicated though it could involve a lot of data.  But unless you are seeing problems I'm not sure it is worth spending much time on.

dashrews78 (approved):

PR #5273, opened 2023-03-16
Skip reconciliation of declarative config for no changes

dhaus67:
## Description

This PR addresses some improvements mentioned within #5186, specifically regarding the reconciliation.

In short, the following has been improved:
- reconciliation will be skipped if a) no changes to messages compared to the previous run b) the previous upsert did not include an error c) the previous deletion did not include an error.
- if the previous upsert included an error, but no changes are indicated in the reconciled messages, only upserts will be retried, deletions will be skipped.
- if the previous deletion included an error, but no changes are indicated in the reconciled messages, only deletion will be retried, upserts will be skipped.

Regarding the change of the reconciliation:
We will retry reconciliation if no changes are made due to the fact that errors have to reach a certain threshold before they will be propagated to the user. If we were to only execute the reconciliation based off the fact if changes are contained within messages, we would currently not surface any errors within the integration health API.

The deletion part is the "costly" part about reconciliation: within deletions, the datastores will be searched for messages with specific properties (i.e. the origin declarative). We should avoid querying the datastores if possible, hence deletion will only be done when necessary, i.e. the messages to reconcile have changed compared to the previous run or an error occurred during deletion.

There may be an additional improvement we can do in the future:
Instead of upserting / deleting _all_ resources, only attempt to upsert / delete resources which indicated a failure. Currently, when a previous reconciliation included an error _all_ resources will be attempted to be upserted / deleted. While the number of resources is most likely not significant for the MVP, it would be worthwhile in the future to only upsert / delete the required ones.

Additional changes:
- Since the reconciliation has been improved, I lowered the interval for reconciliations to 20 seconds from 1 minute. This will improve UX, since errors will be surface in 1 minute, instead of 3 minutes (given that errors are persistent across reconciliation runs).

## Checklist

## Testing Performed

- see tests passing in CI.

### DIFF
--- a/central/declarativeconfig/manager_impl.go
+++ b/central/declarativeconfig/manager_impl.go
@@ -68,7 +68,7 @@ type managerImpl struct {
 	reconciliationCtx context.Context
 
 	declarativeConfigErrorReporter integrationhealth.Reporter
-	errorsPerDeclarativeConfig     map[string]int
+	errorsPerDeclarativeConfig     map[string]int32
 
 	updaters map[reflect.Type]updater.ResourceUpdater
 }
@@ -100,7 +100,7 @@ func New(reconciliationTickerDuration, watchIntervalDuration time.Duration, upda
 		updaters:                       updaters,
 		reconciliationCtx:              writeDeclarativeRoleCtx,
 		declarativeConfigErrorReporter: reconciliationErrorReporter,
-		errorsPerDeclarativeConfig:     map[string]int{},
+		errorsPerDeclarativeConfig:     map[string]int32{},
 		idExtractor:                    idExtractor,
 		nameExtractor:                  nameExtractor,
 		shortCircuitSignal:             concurrency.NewSignal(),
@@ -319,15 +319,12 @@ func (m *managerImpl) doDeletion(transformedMessagesByHandler map[string]protoMe
 // In case err == nil, the health status will be set to healthy.
 // In case err != nil _and_ the number of errors for this message is >= the given threshold, the health
 // status will be set to unhealthy.
-func (m *managerImpl) updateHealthForMessage(handler string, message proto.Message, err error, threshold int) {
+func (m *managerImpl) updateHealthForMessage(handler string, message proto.Message, err error, threshold int32) {
 	messageID := m.idExtractor(message)
 	integrationHealth := declarativeConfigUtils.IntegrationHealthForProtoMessage(message, handler, err, m.idExtractor, m.nameExtractor)
 
 	if err != nil {
-		// Limit the number of recorded errors to the threshold.
-		if m.errorsPerDeclarativeConfig[messageID] < threshold {
-			m.errorsPerDeclarativeConfig[messageID]++
-		}
+		m.errorsPerDeclarativeConfig[messageID]++
 		if m.errorsPerDeclarativeConfig[messageID] >= threshold {
 			m.declarativeConfigErrorReporter.UpdateIntegrationHealthAsync(integrationHealth)
 		}
--- a/central/declarativeconfig/manager_impl.go
+++ b/central/declarativeconfig/manager_impl.go
@@ -52,8 +52,9 @@ type managerImpl struct {
 	transformedMessagesByHandler map[string]protoMessagesByType
 	transformedMessagesMutex     sync.RWMutex
 
-	lastReconciliationHash   uint64
-	lastReconciliationFailed concurrency.Flag
+	lastReconciliationHash uint64
+	lastUpsertFailed       concurrency.Flag
+	lastDeletionFailed     concurrency.Flag
 
 	reconciliationTickerDuration time.Duration
 	watchIntervalDuration        time.Duration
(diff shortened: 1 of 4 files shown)


dhaus67:
* master
  * **PR #5185** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5185" target="_blank">Graphite</a> 
    * **PR #5186** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5186" target="_blank">Graphite</a> 
      * **PR #5266** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5266" target="_blank">Graphite</a> 
        * **PR #5273** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5273" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5273?utm_source=stack-comment).

central/declarativeconfig/manager_impl.go
```diff
@@ -224,19 +228,37 @@ func (m *managerImpl) runReconciliation() {
...
 	transformedMessagesByHandler := maputil.ShallowClone(m.transformedMessagesByHandler)
 	m.transformedMessagesMutex.RUnlock()
+
 	m.reconcileTransformedMessages(transformedMessagesByHandler)
 }
 
 func (m *managerImpl) reconcileTransformedMessages(transformedMessagesByHandler map[string]protoMessagesByType) {
 	log.Debugf("Run reconciliation for the next handlers: %v", maputil.Keys(transformedMessagesByHandler))
+	hasChanges, hash := calculateHashAndIndicateChanges(transformedMessagesByHandler, m.lastReconciliationHash)
+	m.lastReconciliationHash = hash
```
0x656b694d, line 238: What about little refactoring
```suggestion
	hasChanges := m.updateHash(transformedMessagesByHandler)
```

0x656b694d (approved):
Only minor refactoring suggestion.

central/declarativeconfig/manager_impl.go
```diff
@@ -294,15 +320,16 @@ func (m *managerImpl) handleMissingTypeUpdater(handler string, protoType reflect
...
 	messageID := m.idExtractor(message)
 	integrationHealth := declarativeConfigUtils.IntegrationHealthForProtoMessage(message, handler, err, m.idExtractor, m.nameExtractor)
 
 	if err != nil {
-		currentDeclarativeConfigErrors := m.errorsPerDeclarativeConfig[messageID] + 1
-		m.errorsPerDeclarativeConfig[messageID] = currentDeclarativeConfigErrors
-
-		if currentDeclarativeConfigErrors >= threshold {
+		// Limit the number of recorded errors to the threshold.
+		if m.errorsPerDeclarativeConfig[messageID] < threshold {
```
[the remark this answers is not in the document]
dhaus67: I don't particularly see a benefit of _not_ limiting it, since we do not expose any metrics / the number of times a specific resource encountered any issues. Instead, we can just stop increasing the number once we hit the threshold.

What would be the benefit of keeping it?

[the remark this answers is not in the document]
ivan-degtiarenko: I think  the greatest benefit is that we would have one `if` less :) 
Also, *if we were to expose* `errorsPerDeclarativeConfig` some way(in logs, metrics, etc) I think seeing an increasing counter of errors would be easier to understand compared to the static `threshold` amount of errors

central/declarativeconfig/manager_impl.go
```diff
@@ -251,13 +278,18 @@ func (m *managerImpl) doUpsert(transformedMessagesByHandler map[string]protoMess
 			for _, message := range messages {
 				err := typeUpdater.Upsert(m.reconciliationCtx, message)
 				m.updateHealthForMessage(handler, message, err, consecutiveReconciliationErrorThreshold)
+				if err != nil {
+					failureInUpsert = true
```
ivan-degtiarenko, line 282: nit: `failureInUpsert = err != nil`

central/declarativeconfig/manager_impl.go
```diff
@@ -231,16 +236,38 @@ func (m *managerImpl) runReconciliation() {
...
 	m.transformedMessagesMutex.RUnlock()
+
 	m.reconcileTransformedMessages(transformedMessagesByHandler)
 }
 
 func (m *managerImpl) reconcileTransformedMessages(transformedMessagesByHandler map[string]protoMessagesByType) {
 	log.Debugf("Run reconciliation for the next handlers: %v", maputil.Keys(transformedMessagesByHandler))
-	m.doUpsert(transformedMessagesByHandler)
-	m.doDeletion(transformedMessagesByHandler)
+
```
ivan-degtiarenko, line 245: Let's add unit test covering this newly added logic

central/declarativeconfig/manager_impl.go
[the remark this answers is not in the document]
dhaus67: Don't have a strong enough opinion on this one, I'll revert it back to keep increasing.

central/declarativeconfig/manager_impl.go
dhaus67: Can't make the change like this, this will set `failureInUpsert` to false if the _last_ upsert didn't have any errors, although all previous ones did. So, we would indicate that "no error happened during upserts" which is wrong.

ivan-degtiarenko (approved):

PR #5274, opened 2023-03-16
Enable Dependabot auto-merge for operator

vladbologa:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -80,6 +80,7 @@ updates:
     labels:
     - "dependencies"
     - "area/operator"
+    - "auto-merge"
     reviewers:
     - "stackrox/backend-dep-updaters"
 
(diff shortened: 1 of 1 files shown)


vladbologa:
> I'm afraid some of the dependency updates need to be coupled with operator-sdk scaffolding upgrades. So auto-merge might cause us some surprises... But I'm not against giving it a go and later revert if it really does cause issues.

Would this change fully automate Dependabot? Don't we still need to tag the PRs with `ok-to-test`, and approve them? @janisz 

Operator-sdk updates afaik are backwards compatible, and the migrations only need to be performed if we need the new features. But then we would need to keep track of which migrations we have applied so far.

janisz (approved):

janisz:
> Would this change fully automate Dependabot?

Yes, unless there's a failure in one of the checks. Then you need to return/retest or find other way to fix issues.

The automerge will work after this PR is merged so you need to manually merge Dependabots prs that are already open.

kovayur (approved):

PR #5283, opened 2023-03-16
ROX-15965: Add filtered/derived icons to namespace group labels

sachaudh:
## Description

This PR adds an icon to the labels depending on whether the namespace is filtered or derived.
1. A namespace is `filtered` if it was selected using the dropdown. 
2. A namespace is `derived` if it was not filtered and some deployment within a filtered namespace is communicating with some deployment from that namespace

Screenshot 2023-03-16 at 3 47 19 PM


sachaudh:
@zhenpesky the icons seem to have this circle border around them. I was hoping you could re-upload the SVGs to the jira without those. Let me know if you can, thanks!

zhenpesky (commented):
@sachaudh I just uploaded the new SVG format icon for the derived and filtered in the Jira, please check.

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
```diff
@@ -45,14 +48,16 @@ function getBaseNode(id: string): CustomNodeModel {
...
+    deploymentId: string,
+    isFiltered: boolean
 ): NamespaceNodeModel {
     const namespaceData: NamespaceData = {
         collapsible: true,
         showContextMenu: false,
         type: 'NAMESPACE',
         namespace,
         cluster,
+        labelIconClass: isFiltered ? filteredNamespaceSVG : derivedNamespaceSVG,
```
vjwilson, line 60: That's a confusing property name that I guess the topology library dictates.

vjwilson (approved):

ui/apps/platform/src/Containers/NetworkGraph/utils/modelUtils.tsx
sachaudh: Yes, it is very confusing as it's basically an icon, not a class 😕. Really wish they had some more in-depth docs for these sorts of things

PR #5284, opened 2023-03-16
ROX-16018: Add the filter icon to the deployment node label

sachaudh:
## Description

This PR adds an icon to the labels depending on whether the deployment is filtered or not.
1. A deployment is `filtered` if it was selected using the dropdown. 
2. If not, we don't want to show an icon

Screenshot 2023-03-16 at 4 04 49 PM


zhenpesky (changes requested):
Looks good but we still need to swap the SVGs, I uploaded the new SVGs without the border in Jira.

vjwilson (approved):
TS code looks fine. Approved after replacement icons from Zhenpeng are added.

PR #5285, opened 2023-03-16
[ROX-14919] Scan images that are being tested

md2119:
## Description
Explicitly scan images to make sure the scan data is populated.

## Checklist

## Testing Performed
n/a

### DIFF
--- a/qa-tests-backend/src/test/groovy/DeploymentTest.groovy
+++ b/qa-tests-backend/src/test/groovy/DeploymentTest.groovy
@@ -38,10 +38,14 @@ class DeploymentTest extends BaseSpecification {
 
     def setupSpec() {
         orchestrator.createDeployment(DEPLOYMENT)
+        ImageService.scanImage(DEPLOYMENT_IMAGE_NAME)
     }
 
     def cleanupSpec() {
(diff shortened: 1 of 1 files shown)


qa-tests-backend/src/test/groovy/DeploymentTest.groovy
```diff
@@ -38,10 +38,14 @@ class DeploymentTest extends BaseSpecification {
 
     def setupSpec() {
         orchestrator.createDeployment(DEPLOYMENT)
+        ImageService.scanImage(DEPLOYMENT_IMAGE_NAME)
```
connorgorman, line 41: Why is this necessary? the deployment creation should ensure that images are scanned. If this is not happening, then we should fix the root problem?

md2119: I wonder if the resync changes have to do with the failures we have seen, including policy events not showing up. The coverage of this test is queried correctly. Although you raise a genuine concern, it doesn't seem to be the scope of this test.

If this fixes the failure, it is clear that the issue is not query framework or links etc.

charmik-redhat (approved):

PR #5288, opened 2023-03-17
ROX-16014: Change the label background color to white for derived namespaces

sachaudh:
## Description

This PR changes the background label color to white for the derived namespaces

Screenshot 2023-03-16 at 5 07 21 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
@@ -77,12 +77,21 @@ const StyleGroup: React.FunctionComponent<StyleGroupProps> = ({
         return newData as CustomGroupNodeData;
     }, [data]);
 
+    let className = '';
+
+    if (passedData.type === 'NAMESPACE') {
+        className = `${className} ${
+            passedData.isFilteredNamespace ? 'filtered-namespace' : 'derived-namespace'
+        }`.trim();
+    }
+
     return (
         <DefaultGroup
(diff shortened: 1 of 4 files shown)


zhenpesky (approved):

ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
```diff
@@ -77,12 +77,21 @@ const StyleGroup: React.FunctionComponent<StyleGroupProps> = ({
...
     }, [data]);
 
+    let className = '';
+
+    if (passedData.type === 'NAMESPACE') {
+        className = `${className} ${
+            passedData.isFilteredNamespace ? 'filtered-namespace' : 'derived-namespace'
+        }`.trim();
+    }
+
```
vjwilson, line 87: We do this class name interpolation fairly frequently throughout the app. 

I'm not one to advocate for 3rd-party libraries much, but the `class-names` package would clean up all these interpolations.

https://github.com/JedWatson/classnames#readme

vjwilson (approved):

ui/apps/platform/src/Containers/NetworkGraph/components/StyleGroup.tsx
sachaudh: I think that's a decent suggestion. Maybe we can look into it to see if it's worth adding. We can also see if we can make our own solution. Maybe we can consider this after we start adding a bit more classes. That way we can have more examples to go off of.

PR #5292, opened 2023-03-17
ROX-15425: Add Node Inventory container to collector DaemonSet

fredrb:
## Description

This PR is a follow-up of the work in the scanner repo to have Scanner run in node inventory mode and produce messages that should be consumed by compliance: stackrox/scanner#1116.

## Checklist
- [x] ~~Determined and documented upgrade steps~~ (This will be handled in a separate task, which is already created in the epic: [ROX-14687](https://issues.redhat.com/browse/ROX-14687))

## Testing Performed

- [ ] CI
- [x] Manual Helm deployment on OpenShift cluster 
- [x] Manual local deployment

Screenshot 2023-03-22 at 12 02 18
I deployed that to OpenShift with `MAIN_IMAGE_TAG=3.74.x-459-g8a29de2234 POD_SECURITY_POLICIES=false ./deploy/openshift/deploy.sh` and it looks good!


SCANNER_VERSION
```diff
@@ -1 +1 @@
-2.28.x-39-gb748a2651f
+2.28.x-40-g581c726c42
```
Maddosaurus, line 1: If you like, we could update to `2.28.x-48-g8129118a61` to benefit from the caching scanner 😄

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
```diff
@@ -147,6 +147,28 @@ spec:
...
+              fieldPath: spec.nodeName
+        {{- include "srox.envVars" (list . "daemonset" "collector" "node-inventory") | nindent 8 }}
+        volumeMounts:
+         - mountPath: /host
+           name: host-root-ro
+           readOnly: true
+         - mountPath: /tmp/
+           name: tmp-volume
+         - mountPath: /cache
+           name: cache-volume
```
Maddosaurus, line 171: [No change needed here]
These two mounts are exclusively needed by the Node Scanning code.
As we moved that into this image, we could remove the mounts from Compliance (L#146-149). 
As they don't hurt, we can also leave them for now. Up to you.

Maddosaurus (approved):
Nice set of changes!
I like that you made sure to update tests to reflect the new setup.
The comments are completely optional, feel free to skip them.

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
vikin91: Oh, good point, we should definitely clean this up

go.mod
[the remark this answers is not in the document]
vikin91: I think the only thing that could cause that is pulling the newest scanner dependency.

vikin91:
Screenshot 2023-03-24 at 17 11 41
I investigated the code in deployer.go and all values are okay there (the same pattern as for main image). For some reason the image name of scanner is only `scanner-slim` but should be `rhacs-eng/scanner-slim`. Next idea: search in roxctl.

PR #5306, opened 2023-03-20
ROX-14668: POC new Network Graph API to fetch isolation details on individual deployment IDs

fredrb:
## Description

This is an alternative to significantly changing the API `/v1/networkgraph/cluster` or `/v1/networkpolicies/cluster` to accommodate the needs of Network Graph V2.

The Netwrok Graph V2 requires isolation data (ingress/egress/policy Ids) on the network graph itself rather than on the "allowed" tab. If the edges differ from the allowed graph to the flow graph, the Front-End might be missing data to render the isolation badge on the UI, because one node present in the flow graph isn't being returned in the allowed graph. As described in the issue [ROX-14668](https://issues.redhat.com/browse/ROX-14668). 

This is intended behavior from an API perspective, since the data returned on the `/v1/networkpolicies/cluster` is supposed to show edges where Network Policies allow two nodes to communicate. As opposed to what is shown on the `/v1/networkgraph/cluster`, where it's the observed flow between two nodes. 

 ---

Rather than trying to rebuild these APIs, this PR offers two options to solve this issue:

1) **Enhance flow graph nodes with Network Policy info:** this would be guarded by a URL param (`&include_policies=true`) and will fetch all Network Policies for the deployments found in the flow graph. This makes it easier because the Front-End won't need to make multiple requests. However, it makes the already expensive network graph call even more expensive, as it now also has to fetch Network Policies.

2) ~~**Add a new and smaller API**: `/v1/networkpolicies/deployments` which takes a list of deployment IDs. It returns only the isolation details for each of the deployments in the request. This gives a bit more control to the FE in case they don't need to load partial information. It also leaves the Network Flow Graph request unchanged.~~ (removed in favor of the first option)

## Checklist

## Testing Performed

- [x] Unit tests for the matcher
- [ ] Service tests
- [x] Manual tests
- [ ] Enhance E2E tests

---

Here's an example of `sensor` entity:

`GET v1/networkgraph/cluster/<cluster_id>?query=Namespace:stackrox&since=<now>&includePorts=true&includePolicies=true`
```json
      "entity": {
        "type": "DEPLOYMENT",
        "id": "a50ffcb3-1024-46a3-8fba-31c2996337fa",
        "deployment": {
          "name": "sensor",
          "namespace": "stackrox",
          "cluster": "remote",
          "listenPorts": [
            {
              "port": 9090,
              "l4protocol": "L4_PROTOCOL_TCP"
            },
            {
              "port": 9443,
              "l4protocol": "L4_PROTOCOL_TCP"
            },
            {
              "port": 8443,
              "l4protocol": "L4_PROTOCOL_TCP"
            }
          ]
        }
      },
      "internetAccess": false,
      "policyIds": [
        "8d134e16-6614-4a27-9cfa-f6be1b3c3aa3",
        "42d3dfc4-98be-4d68-8a46-70bf7ff08b9d"
      ],
      "nonIsolatedIngress": false,
      "nonIsolatedEgress": true,
      "queryMatch": true,
      "outEdges": { ... }
```


sachaudh (commented):
I didn't look at the code in-depth (not UI code 😢 ) so I won't approve, but overall looks good to me. With this change, we can simplify the API to be just one with a query parameter to include policy ids (as seen with the second API). This will not break anything once merged even without the UI changes, right? Just double checking

sachaudh:
@fredrb Linda had a good question that came up during our last meeting and I just wanted to double-check with you about it. Apparently, we use the network policy graph as a basis for the network policy simulation. There's an API we use:
```
/v1/networkpolicies/simulate/${clusterId}
vs.
/v1/networkpolicies/cluster/${clusterId}
```
And I think if we're simulating the network graph then we use the first API. Would the issue still persist even with this PR change or would it not?

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
 			node.QueryMatch = true
 		}
 	}
+
+	if request.IncludePolicies {
+		err := s.enhanceWithNetworkPolicyIsolationInfo(ctx, graph)
+		if err != nil {
+			return nil, err
```
connorgorman, line 321: add context to this error

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
+	for _, deployment := range deploymentObjects {
+		clusterNamespaceContext.Add(deploymentMatcher.ClusterNamespace{
+			Cluster:   deployment.GetClusterId(),
+			Namespace: deployment.GetNamespace(),
+		})
+	}
+
+	matcher, err := deploymentMatcher.BuildMatcher(ctx, s.networkPolicy, clusterNamespaceContext.AsSlice())
+	if err != nil {
+		return err
```
connorgorman, line 351: add context to this error

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
+	var deploymentIds []string
+	for _, node := range graph.GetNodes() {
+		if node.GetEntity().GetType() == storage.NetworkEntityInfo_DEPLOYMENT {
+			deploymentIds = append(deploymentIds, node.GetEntity().GetId())
+		}
+	}
+
+	deploymentObjects, err := s.deployments.GetDeployments(ctx, deploymentIds)
+	if err != nil {
+		return err
```
connorgorman, line 338: add context to this error

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
+			EgressIsolated:  details.EgressIsolated,
+			PolicyIDs:       details.PolicyIDs,
+		}
+	}
+
+	for idx, node := range graph.Nodes {
+		if node.GetEntity().GetType() == storage.NetworkEntityInfo_DEPLOYMENT {
+			deploymentID := node.GetEntity().GetId()
+			if isolationDetail, ok := isolationDetails[deploymentID]; ok {
+				graph.Nodes[idx].NonIsolatedEgress = !isolationDetail.EgressIsolated
```
connorgorman, line 368: nodes are pointers so no need to nodes[idx]

central/networkpolicies/deployment/matcher.go
```diff
@@ -0,0 +1,132 @@
...
+
+func buildNetworkPolicies(ctx context.Context, store networkPolicyDS.DataStore, namespace []ClusterNamespace) (map[ClusterNamespace][]selectablePolicy, error) {
+	result := map[ClusterNamespace][]selectablePolicy{}
+	for _, clusterNs := range namespace {
+		policies, err := store.GetNetworkPolicies(ctx, clusterNs.Cluster, clusterNs.Namespace)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get policies for %v", clusterNs)
+		}
+
+		result[clusterNs] = []selectablePolicy{}
```
connorgorman, line 68: any reason we need to initialize this? appending to a nil slice is allowed

central/networkpolicies/deployment/matcher.go
```diff
@@ -0,0 +1,132 @@
...
+
+// GetIsolationDetails will iterate over preloaded Network Policies and return the isolation level of
+// the resource provided as parameter.
+func (m *policyMatcherImpl) GetIsolationDetails(resource LabeledResource) IsolationDetails {
+	// Get Policies from the same cluster and namespace
+	policies, ok := m.netpolMap[ClusterNamespace{
+		Namespace: resource.GetNamespace(),
+		Cluster:   resource.GetClusterId(),
+	}]
+
```
connorgorman, line 92: remove extra line

proto/api/v1/network_graph_service.proto
```diff
@@ -45,6 +45,7 @@ message NetworkGraphRequest {
 
     bool              include_ports = 4;
     NetworkGraphScope scope         = 5;
+    bool include_policies           = 6;
```
connorgorman, line 48: nit realign

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
 
+func (s *serviceImpl) enhanceWithNetworkPolicyIsolationInfo(ctx context.Context, graph *v1.NetworkGraph) error {
+	var deploymentIds []string
+	for _, node := range graph.GetNodes() {
+		if node.GetEntity().GetType() == storage.NetworkEntityInfo_DEPLOYMENT {
+			deploymentIds = append(deploymentIds, node.GetEntity().GetId())
+		}
+	}
+
+	deploymentObjects, err := s.deployments.GetDeployments(ctx, deploymentIds)
```
connorgorman, line 336: Add a TODO and a ticket to make this cheaper by just selecting the required files once Postgres ships. This will make it much cheaper and we will be able to remove the cluster namespace context by select distinct(clusterid, namespace)

central/networkgraph/service/service_impl.go
```diff
@@ -311,9 +314,67 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
+
+	clusterNamespaceContext := set.NewSet[deploymentMatcher.ClusterNamespace]()
+	for _, deployment := range deploymentObjects {
+		clusterNamespaceContext.Add(deploymentMatcher.ClusterNamespace{
+			Cluster:   deployment.GetClusterId(),
+			Namespace: deployment.GetNamespace(),
+		})
+	}
+
+	matcher, err := deploymentMatcher.BuildMatcher(ctx, s.networkPolicy, clusterNamespaceContext.AsSlice())
```
connorgorman, line 349: maybe a minor thing, but is there a reason this needs to be made a slice instead of just passing around the set and iterating on that (which is just a map anyways)

connorgorman (changes requested):
I guess one thing that I'm not super clear on is what isolated means in this context, but nits and other questions just based on the code

central/networkgraph/service/service_impl.go
fredrb: Done

central/networkgraph/service/service_impl.go
fredrb: Done

central/networkgraph/service/service_impl.go
fredrb: Done

central/networkpolicies/deployment/matcher.go
fredrb: Done

central/networkpolicies/deployment/matcher.go
fredrb: Done

proto/api/v1/network_graph_service.proto
fredrb: Done

central/networkgraph/service/service_impl.go
fredrb: Done

central/networkgraph/service/service_impl.go
fredrb: Done

md2119 (commented):
If I am understanding it correctly, the client wants to know if deployments have any network policies applied and what kind which is what translates to isolated or non-isolated?

The approach is fine, given that client has to set an option when making the request.

Separately, have we evaluated the pros and cons of processing this in Sensor and persist the properties in the DB? I believe this is much similar to "Network policy applied" security policy evaluation.

fredrb:
> Separately, have we evaluated the pros and cons of processing this in Sensor and persist the properties in the DB? I believe this is much similar to "Network policy applied" security policy evaluation.

@md2119 I think this would be an alternative. However, it requires a much bigger change. This would mean changing the `storage.Deployment` object to hold a slice of Network Policies (or at least some information on them). It would also mean reverting the augmented object.

What would be the motivation of doing it this way? To avoid querying for Network Policies in the Graph call?

I don't think I can justify a larger change just to avoid a query here. Specially because a this was already being done in the NW 2.0 by calling `v1/networkgraph` and `v1/networkpolicies` sequentially. I'd argue that bundling the work in a single request even might remove the overhead from multiple requests. 

Let me know what do you think.

central/networkgraph/service/service_impl.go
```diff
@@ -309,9 +312,68 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
 			node.QueryMatch = true
 		}
 	}
+
+	if request.IncludePolicies {
+		err := s.enhanceWithNetworkPolicyIsolationInfo(ctx, graph)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to enhance graph nodes with Network Policy information")
```
vikin91, line 319: Shall we maybe return the graph anyway `return graph, errors.Wrap(err,..)`? That would be a bit more backwards compatible (if that makes sense at all). I have personally no opinion about this, but maybe you have?

central/networkpolicies/deployment/matcher.go
```diff
@@ -0,0 +1,131 @@
...
+func hasEgress(types []storage.NetworkPolicyType) bool {
+	return hasPolicyType(types, storage.NetworkPolicyType_EGRESS_NETWORK_POLICY_TYPE)
+}
+
+func hasIngress(types []storage.NetworkPolicyType) bool {
+	if len(types) == 0 {
+		return true
+	}
+	return hasPolicyType(types, storage.NetworkPolicyType_INGRESS_NETWORK_POLICY_TYPE)
+}
```
vikin91, line 122: How about this crazy idea?

```suggestion
func hasIngress(types []storage.NetworkPolicyType) bool {
	return len(types) == 0 || hasPolicyType(types, storage.NetworkPolicyType_INGRESS_NETWORK_POLICY_TYPE)
}
```

vikin91 (approved):
LGTM! I mention here two additional issues to those mentioned in the sync review. It looks really good, so I mark it ✅

md2119:
> @md2119 I think this would be an alternative. However, it requires a much bigger change.

Agreed.

> This would mean changing the storage.Deployment object to hold a slice of Network Policies (or at least some information on them). 

I think this can be decoupled into separate events. One goes to deployment and the other updates the relationship.

> What would be the motivation of doing it this way?  

Offloading the cycles from Central to Sensor. Central spends cycles for all clusters, Sensor computations are confined to one cluster. 

> I don't think I can justify a larger change just to avoid a query here.
Agreed. Maybe you can add the topic to long-term roadmap.

central/networkgraph/service/service_impl.go
```diff
@@ -309,9 +312,68 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
...
+
+	isolationDetails := make(map[string]deploymentMatcher.IsolationDetails, len(deploymentObjects))
+	for _, deployment := range deploymentObjects {
+		details := matcher.GetIsolationDetails(deployment)
+		isolationDetails[deployment.GetId()] = deploymentMatcher.IsolationDetails{
+			IngressIsolated: details.IngressIsolated,
+			EgressIsolated:  details.EgressIsolated,
+			PolicyIDs:       details.PolicyIDs,
+		}
+	}
```
lvalerom, line 361: I think you can avoid this loop if in the previous loop you create a `map[deploymentID]*storage.Deployment` and then in the following loop access this map instead of using the `isolationDetails` map. I think the memory usage should be similar because you are interchanging the new map with `isolationDetails` but this way you avoid one loop.

central/networkgraph/service/service_impl.go
fredrb: Yea, I guess that makes sense. I think it would be better to log a warning and then return the graph without an error object. I think returning an error here would result in a non-200 HTTP status code.

fredrb: Done

central/networkgraph/service/service_impl.go
```diff
@@ -312,9 +315,68 @@ func (s *serviceImpl) getNetworkGraph(ctx context.Context, request *v1.NetworkGr
 			node.QueryMatch = true
 		}
 	}
+
+	if request.IncludePolicies {
```
connorgorman, line 319: Nit: Use the getters when accessing fields. GetIncludePolicies()

connorgorman (approved):

central/networkgraph/service/service_impl.go
lvalerom: 🤔 I think this didn't get pushed.

central/networkgraph/service/service_impl.go
fredrb: Done

lvalerom (approved):
LGTM. Double check Piotr's comment regarding returning the graph in case `enhanceWithNetworkPolicyIsolationInfo` fails, I think the commit didn't get pushed.

central/networkgraph/service/service_impl.go
fredrb: I think I lost the change while restoring/stashing changes. Re-implemented and pushed a new commit. Good catch @lvalerom

PR #5307, opened 2023-03-20
ROX-8277 Fix malformed operator k8s client user-agent

ludydoo:
## Description

In certain cases, the UserAgent string sent by the operator to the kubernetes API is malformed. e.g.

```
"userAgent": "stackrox-operator/v0.0.0 (linux/amd64) kubernetes/$Format/leader-election"
```
^ see the `$Format/...` suffix 

It seems that the operator controller-runtime k8s client config is not used for *all* clients. For example, it seems that there is another k8s client instantiated for the leader election, which does not use the specified userAgent.

### DIFF
--- a/operator/main.go
+++ b/operator/main.go
@@ -94,14 +94,16 @@ func run() error {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	mgr, err := ctrl.NewManager(utils.GetRHACSConfigOrDie(), ctrl.Options{
+	restConfig := utils.GetRHACSConfigOrDie()
+
+	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                 scheme,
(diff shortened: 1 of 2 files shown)


ludydoo:
Closing as it is not currently possible to configure the leader-election user agent. https://github.com/kubernetes-sigs/controller-runtime/issues/2243

porridge:
How about fixing it upstream?

ludydoo:
@porridge yeah I can look at that. I see two issues
1. The client-go `AddUserAgent` not "adding an user agent", but ignoring a pre-existing user agent and using `DefaultKubernetesUserAgent() + "/" + <whatever>` 
2. The `controller-runtime` ignoring the `LeaderElectionConfig: rest.Config.UserAgent` configuration.

ludydoo:
k8s-client PR: https://github.com/kubernetes/kubernetes/pull/116801

PR #5309, opened 2023-03-20
Use multi-arch test images & few changes

pratham-m:
This PR intends to add multi-arch support for the groovy tests.
Test images used in respective test are changed to use alternate multi-arch images which support 3 platforms viz. `x86_64`, `ppc64le` and `s390x`.

Changes are verified manually using `./gradlew test --tests=TestName`

### DIFF
--- a/qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
+++ b/qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
@@ -8,6 +8,7 @@ import objects.Deployment
 import services.GraphQLService
 
 import spock.lang.Tag
+import util.Env
 
 @Tag("BAT")
 @Tag("GraphQL")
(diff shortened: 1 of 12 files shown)


pratham-m:
Hi @gavin-stackrox  @robbycochran can you kindly check the changes ?

qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
```diff
@@ -125,7 +125,7 @@ class DeploymentEventGraphQLTest extends BaseSpecification {
             log.info "return code " + depEvents.getCode()
             assert depEvents.getValue().result != null
             def events = depEvents.getValue().result
-            assert events.numPolicyViolations == 1
```
gavin-stackrox, line 128: IIUC this test expects only "Ubuntu Package Manager Execution" to be triggered. But under P/Z more than that is triggered. That works for other cluster flavors, so I would prefer this to be a:
```
if P/Z
  >= 1
else
  == 1
```
Even better would be to explain what other policies are triggered and why this needs to be >= 1 on P/Z.

gavin-stackrox (changes requested):

qa-tests-backend/src/test/groovy/DeploymentEventGraphQLTest.groovy
pratham-m: ACK - will make the changes

gavin-stackrox (approved):
Looks ready to merge. Thanks @pratham-m

pratham-m:
> Looks ready to merge. Thanks @pratham-m

Thanks @gavin-stackrox

PR #5321, opened 2023-03-20
Fix GitHub actions to allow for external contributions

dhaus67:
## Description

Currently, external contributors won't be able to provide mergable PRs that fulfill all required steps within GitHub actions, due to us currently requiring to push the container image within the build step.

Since external contributors (or, to be more specific, PRs from forks) do not have access to CI secrets, they currently are failing the push step, specifically the docker login part.

Instead of requiring the build step for external contributors to include the push part, I've skipped this for them (for "internal" PRs, i.e. originating from the same repository, the images will still be pushed).

This has the following consequences:
- the workflow will succeed and the PR can be merged.
- all subsequent E2E tests triggered from this PR will fail, due to the image not being available. If the desire is to run these tests, one would have to mirror the branch to an internal branch, build + push the image via CI, and re-trigger E2E tests.

Alternatives considered:
Instead of skipping parts of the current jobs, split out the build and push parts as separate jobs.
I've decided against this due to:
- the build part currently (for most resources) includes doing a docker build. As a consequence, we would have to upload the docker archive as artifact to share between jobs. This would further increase the build time.
- some parts of the build require images being pushed prior. Specifically, the `index-build` step when building the operator images.

## Checklist

## Testing Performed

https://github.com/dhaus67/stackrox/actions/runs/4468859918

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -84,7 +84,6 @@ jobs:
         with:
           name: cli-build
           path: cli-build.tgz
-
   pre-build-go-binaries:
     runs-on: ubuntu-latest
     container:
@@ -123,7 +122,6 @@ jobs:
         with:
           name: go-binaries-build
           path: go-binaries-build.tgz
-
   pre-build-go-binaries-rcd:
     runs-on: ubuntu-latest
     container:
@@ -165,7 +163,6 @@ jobs:
       with:
         name: go-binaries-build-rcd
         path: go-binaries-build-rcd.tgz
-
   pre-build-docs:
     runs-on: ubuntu-latest
     container:
@@ -273,8 +270,11 @@ jobs:
       - name: Generate OSS notice
         run: make ossls-notice
 
+      # Check if we actually need this, or we can skip this and only do if after make docker-build-image.
       # needed for docs ensure_image.sh initial pull with RHACS_BRANDING
       - name: Docker login
+        # Skip for external contributions.
+        if: github.event.pull_request.head.repo.full_name == github.repository
         run: |
             docker login -u "${QUAY_RHACS_ENG_RO_USERNAME}" --password-stdin quay.io <<<"${QUAY_RHACS_ENG_RO_PASSWORD}"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -438,8 +451,10 @@ jobs:
       - name: Setup Go build environment
         if: contains(github.event.pull_request.labels.*.name, 'ci-release-build')
         run: echo "GOTAGS=release" >> "$GITHUB_ENV"
-
+      # Is this required to be run before building the bundle image? If not, let's move it afterwards.
```
janisz, line 454: I think it's requred as it changes the way image is buidl

.github/workflows/build.yaml
```diff
@@ -379,18 +385,25 @@ jobs:
...
     - name: Login to quay.io/rhacs-eng
+      # Skip for external contributions.
+      if: github.event.pull_request.head.repo.full_name == github.repository
       uses: docker/login-action@v2
       with:
         registry: quay.io
         username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
         password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
 
+    # Is this required to be run _after_ docker login? If not, let's move it before doing docker login.
```
janisz, line 400: I think it can be moved up

.github/workflows/build.yaml
```diff
@@ -496,16 +517,22 @@ jobs:
...
+      if: github.event.pull_request.head.repo.full_name == github.repository
       uses: docker/login-action@v2
       with:
         registry: quay.io
         username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
         password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
 
     - name: Push to quay.io/rhacs-eng
+      # Skip for external contributions.
+      if: github.event.pull_request.head.repo.full_name == github.repository
```
janisz, line 535: How about for `build-and-push` action just skip `push` so we can check if image is buildable? Just like`dex` do it but with different condition.

https://github.com/dexidp/dex/blob/master/.github/workflows/docker.yaml#L81

janisz (approved):

PR #5322, opened 2023-03-20
ROX-13190,ROX-14595,ROX-15073,ROX-15311: Admission controller e2e test fixes

clickboo:
## Description

The "Verify CVE snoozing applies to images scanned by admission controller us.gcr.io/stackrox-ci/nginx:1.10.1" was failing because the policy creation itself was failing - because the policy category being used does not exist.
Changed to use a pre existing category for the new policy, and asserting on the policy create being successful, before running the actual testcase  scenario.

A bunch of other inline scan related tests were failing because of possible timing issues with the scan. Added a force scan for the image in question in the setup stage to avoid the scan related timing issues.

## Checklist

## Testing Performed
CI failure, hence CI only.

### DIFF
--- a/qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
+++ b/qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
@@ -1,5 +1,4 @@
 import io.stackrox.proto.api.v1.Common
-import io.stackrox.proto.api.v1.PolicyServiceOuterClass
 import io.stackrox.proto.storage.ClusterOuterClass.AdmissionControllerConfig
 import io.stackrox.proto.storage.PolicyOuterClass
 import io.stackrox.proto.storage.PolicyOuterClass.PolicyGroup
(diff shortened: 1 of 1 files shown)


charmik-redhat (approved):

qa-tests-backend/src/test/groovy/AdmissionControllerTest.groovy
```diff
@@ -202,24 +208,23 @@ class AdmissionControllerTest extends BaseSpecification {
                 .setBooleanOperator(PolicyOuterClass.BooleanOperator.AND)
         policyGroup.addAllValues([PolicyValue.newBuilder().setValue("CVE-2019-3462").build(),])
 
+        String policyName = "Matching CVE (CVE-2019-3462)"
         PolicyOuterClass.Policy policy = PolicyOuterClass.Policy.newBuilder()
-                .setName("Matching CVE (CVE-2019-3462)")
+                .setName(policyName)
                 .addLifecycleStages(PolicyOuterClass.LifecycleStage.DEPLOY)
-                .addCategories("Testing")
+                .addCategories("DevOps Best Practices")
```
theencee, line 215: any reason for this change?

theencee (approved):

PR #5332, opened 2023-03-20
Jv rox 16074 process listening on ports e2e test to check for closed ports

JoukoVirtanen:
## Description

An e2e test where a processes which is listening on a port is killed and there is an assert to check that the listening endpoint no longer appears in the API. Also adds unit tests. These tests revealed some bugs which were fixed by https://github.com/stackrox/stackrox/pull/5358

The above PR was merged into this one.

## Checklist
- [x] Ran e2e test locally (Failed as expected)
- [x] Ran unit tests (Failed as expected)

### Prerequisites

Deploy stackrox locally or connect to a cluster running stackrox

### Steps

```
cd qa-tests-backend
export ROX_POSTGRES_DATASTORE=true
```

That directory should have a file, qa-test-settings.properties with the following contents

```
export REGISTRY_USERNAME=*********@redhat.com
export REGISTRY_PASSWORD="*******************"
```

If this the first time you are running the tests or you made some changes to the tests or generated files you might or might not need to run the following

```
./scripts/migrate_protos.sh
make clean-generated-srcs
make proto-generated-srcs
```

The test is itself run with the following command

```
./gradlew test --tests=ProcessesListeningOnPortsTest
```

You will also need to either comment out 

```
@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
```

in qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy or set 

```
export=RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS =true
```

You may also want to comment out the first test so that you can just test the last two. You may also want to disable retries by commenting out 

```
@Retry(condition = { Helpers.determineRetry(failure) })
```

in qa-tests-backend/src/test/groovy/BaseSpecification.groovy

### DIFF
--- a/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
@@ -4,7 +4,6 @@ import objects.Deployment
 import objects.K8sServiceAccount
 import objects.Service
 import util.Env
-import util.Helpers
 
 import spock.lang.IgnoreIf
 import spock.lang.Shared
@@ -21,6 +20,7 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
     // Deployment names
     static final private String TCPCONNECTIONTARGET1 = "tcp-connection-target-1"
     static final private String TCPCONNECTIONTARGET2 = "tcp-connection-target-2"
+    static final private String TCPCONNECTIONTARGET3 = "tcp-connection-target-3"
 
     // Other namespace
     static final private String OTHER_NAMESPACE = "qa2"
@@ -51,6 +51,16 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
                     .setExposeAsService(true)
                     .setCommand(["/bin/sh", "-c",])
                     .setArgs(["(socat "+SOCAT_DEBUG+" TCP-LISTEN:8081,fork STDOUT)" as String,]),
+            new Deployment()
+                    .setName(TCPCONNECTIONTARGET3)
+                    .setImage("quay.io/rhacs-eng/qa:socat")
+                    .addPort(8082, "TCP")
+                    .addLabel("app", TCPCONNECTIONTARGET3)
+                    .setExposeAsService(true)
+                    .setCommand(["/bin/sh", "-c",])
+                    .setArgs(["(socat "+SOCAT_DEBUG+" TCP-LISTEN:8082,fork STDOUT & " +
+                            "sleep 90 && pkill socat && sleep 3600)" as String,]),
+                    // The 8082 port is opened. 90 seconds later the process is killed. After that we sleep forever
         ]
     }
 
@@ -79,6 +89,7 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
                 )
         )
 
+        destroyDeployments()
         createDeployments()
     }
 
@@ -99,34 +110,17 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
         destroyDeployments()
     }
 
-    def rebuildForRetries() {
-        if (Helpers.getAttemptCount() > 1) {
-            log.info ">>>> Recreating test deployments prior to retest <<<<<"
-            destroyDeployments()
-            createDeployments()
-            log.info ">>>> Done <<<<<"
-        }
-    }
-
     @Tag("BAT")
     def "Verify networking endpoints with processes appear in API at the deployment level"() {
         given:
         "Two deployments that listen on ports are started up"
 
-        rebuildForRetries()
+        setupSpec()
 
-        String deploymentId1 = targetDeployments[0].getDeploymentUid()
-        String deploymentId2 = targetDeployments[1].getDeploymentUid()
+        String deploymentId1 = targetDeployments.find { it.name == TCPCONNECTIONTARGET1 }?.deploymentUid
+        String deploymentId2 = targetDeployments.find { it.name == TCPCONNECTIONTARGET2 }?.deploymentUid
 
-        def gotCorrectNumElements = waitForResponseToHaveNumElements(2, deploymentId1, 240)
-
-        assert gotCorrectNumElements
-
-        def processesListeningOnPorts = evaluateWithRetry(10, 10) {
-                def temp = ProcessesListeningOnPortsService
-                        .getProcessesListeningOnPortsResponse(deploymentId1)
-                return temp
-        }
+        def processesListeningOnPorts = waitForResponseToHaveNumElements(2, deploymentId1, 240)
 
         assert processesListeningOnPorts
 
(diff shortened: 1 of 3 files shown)


sensor/common/networkflow/manager/manager_impl_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	nowTimestamp := timestamp.Now()
+	var sequenceID int64
+	h.connectionsSequenceID = sequenceID
+	err := h.Process(networkInfo, nowTimestamp, sequenceID)
+	suite.NoError(err)
+	suite.Len(h.endpoints, 1)
+}
+
+func (suite *NetworkflowManagerTestSuite) TestAddOpenAndClosed() {
```
JoukoVirtanen, line 80: This unit test is failing due to a bug in getUpdatedContainerEndpoints. Specifically the issue is that ep is used as a key in a map, but processKey is a pointer, which can take different values, even if the process is the same. This was fixed by storing the process pointers in a map, and making sure that identical processes use the same pointer.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -52,6 +53,15 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
...
                     .setArgs(["(socat "+SOCAT_DEBUG+" TCP-LISTEN:8081,fork STDOUT)" as String,]),
+            new Deployment()
+                    .setName(TCPCONNECTIONTARGET3)
+                    .setImage("quay.io/rhacs-eng/qa:socat")
+                    .addPort(8082, "TCP")
+                    .addLabel("app", TCPCONNECTIONTARGET3)
+                    .setExposeAsService(true)
+                    .setCommand(["/bin/sh", "-c",])
+                    .setArgs(["(socat "+SOCAT_DEBUG+" TCP-LISTEN:8082,fork STDOUT & " +
+			"sleep 90 && pkill socat && sleep 3600)" as String,]),
```
JoukoVirtanen, line 64: A listening endpoint is created and 90 seconds later the listening endpoint is destroyed.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -228,6 +238,96 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
...
+        assert endpoint.containerName == TCPCONNECTIONTARGET3
+        assert endpoint.signal.id
+        assert endpoint.signal.containerId
+        assert endpoint.signal.time
+        assert endpoint.signal.name == "socat"
+        assert endpoint.signal.execFilePath == "/usr/bin/socat"
+        assert endpoint.signal.args == "-d -d -v TCP-LISTEN:8081,fork STDOUT"
+        assert endpoint.signal.pid
+
+        gotCorrectNumElements = waitForResponseToHaveNumElements(0, deploymentId3, 180)
```
JoukoVirtanen, line 279: Checks that there are 0 listening endpoints returned by the API at a time when the listening endpoint is closed. Add a comment before merging.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -228,6 +238,96 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
...
+    def "Verify networking endpoints disappear when process is terminated"() {
+        given:
+        "Two deployments that listen on ports are started up"
+
+        rebuildForRetries()
+        def clusterId = ClusterService.getClusterId()
+
+        String deploymentId3 = targetDeployments[2].getDeploymentUid()
+
+        def gotCorrectNumElements = waitForResponseToHaveNumElements(1, deploymentId3, 240)
```
JoukoVirtanen, line 251: First make sure that we see the listening endpoint in the API when it is open. Add a comment here.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -226,6 +238,104 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
...
+
+        // First check that the listening endpoint appears in the API
+        assert gotCorrectNumElements
+
+        def processesListeningOnPorts = evaluateWithRetry(10, 10) {
```
fredrb, line 261: I wonder why didn't you return the `getProcessesListeningOnPortsResponse` response on the `waitForResponseToHaveNumElements` call? 

The test is having to make the request twice. First when running `waitForResponseToHaveNumElements` and then again here to make the content assertions. 

Wouldn't it be simpler to:

```groovy
def processListeningOnPorts = waitForResponseToHaveNumElements(1, deploymentId3, 240)

// make assertions
assert processesListeningOnPorts
def list = processesListeningOnPorts.listeningEndpointsList
assert list.size() == 1
```
You're even making the assertion for the size twice (one inside of `waitForResponseToHaveNumElements` and another here ☝️ )

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -226,6 +238,104 @@ class ProcessesListeningOnPortsTest extends BaseSpecification {
...
+        destroyDeployments()
+    }
+
+    @Tag("BAT")
+    def "Verify networking endpoints disappear when process is terminated"() {
+        given:
+        "When a deployment listening on a port is created and then the process is terminated"
+
+        setupSpec()
+        rebuildForRetries()
```
fredrb, line 251: I might lack some knowledge on the test lifecycle, but it looks to me that the second `rebuildForRetries` is redoing what the first one did? At least the deployment setup part. 

My understanding from what will happen with these two functions calls is:
1. Create test namespace
2. Create image pull secret
3. Create service account
4. Create deployments
5. Delete deployments
6. Create deployments

I'm probably missing something here. Is there something that makes `setupSpec` not run if the test is being retried but `rebuildForRetries` will?

JoukoVirtanen: I have removed rebuildForRetries.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
JoukoVirtanen: I have changed this now to what you suggested. I knew it was better to return processesListeningOnPorts than a boolean, but I am not experienced with groovy and I had some difficulty doing that. I would need to declare processesListeningOnPorts outside of the loop and specify its type. I had some difficulty doing that.

sensor/common/networkflow/manager/manager_impl_test.go
JoukoVirtanen: Should now be passing.

Molter73 (approved):
LGTM!

PR #5333, opened 2023-03-21
Revert "[ROX-15808] : Fix for watched images not getting scanned more than once"

charmik-redhat:
Reverts stackrox/stackrox#5248


clickboo (approved):
Revert looks good. Interesting however, that we are reverting the fix to a legit issue instead of fixing the test. Hoping the revert is an interim solution till CI failures are tackled.

charmik-redhat:
> Revert looks good. Interesting however, that we are reverting the fix to a legit issue instead of fixing the test. Hoping the revert is an interim solution till CI failures are tackled.

Yeah just a temporary revert to investigate ImageScanningTest since it fails frequently.

theencee (approved):

PR #5335, opened 2023-03-21
ROX-14434: Trigger deployment events for Secrets and Service Account updates

fredrb:
## Description

Local scanning and `SourcedAutogeneratedIntegrations` feature will require changes in the deployment object and/or alert detection on secrets and service account updates. 

Local Scanning is only relevant for OpenShift because it's the only case where we have images being scanned locally. In case image pull secrets get updated, we might need to reprocess the deployment and update `IsClusterLocal` and `NotPullable` properties. 

## Checklist


sensor/common/detector/detector.go
```diff
@@ -274,6 +274,13 @@ func (d *detectorImpl) processUpdatedImage(image *storage.Image) error {
 	}
 	d.enricher.imageCache.Add(key, newValue)
 	d.admissionCacheNeedsFlush = true
+	d.deduper.reset()
+	// Take deployment lock and flush
+	concurrency.WithLock(&d.deploymentDetectionLock, func() {
+		for _, deployment := range d.deploymentStore.GetAll() {
```
fredrb, line 280: If we are in the middle of processing a deployment we might send a deployment (i.e. in between handler and resolver) to detector with incorrect permission levels or exposure info. The alerts will be fixed right after since there is a deployment in pipeline, but still, some incorrect alerts will be sent.

deploy/k8s/deploy.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
+# Enable new event pipeline without re-sync for k8s deployments
+export ROX_RESYNC_DISABLED="false"
```
lvalerom, line 5: Is this needed?

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -315,6 +317,11 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 	}}
 	events := component.NewEvent(sensorEvents...)
 	events.AddSensorEvent(secretToSensorEvent(action, protoSecret))
+
+	// When adding new docker config secrets we need to reprocess every deployment in this cluster.
+	// This is because the field `NotPullable` could be updated and hence new image scan results will appear.
+	events.AddDeploymentReference(resolver.ResolveAllDeployments(), central.ResourceAction_UPDATE_RESOURCE, false)
```
lvalerom, line 323: Should we guard this behind the re-sync envVar? It is ignored by the `OutputQueue` if re-sync is enabled, so technically it doesn't matter if it's guarded or not but, since is something we only need to do in the re-sync-less path, it might be clearer if it is.

sensor/kubernetes/listener/resources/serviceaccount.go
```diff
@@ -53,13 +54,23 @@ func (s *serviceAccountDispatcher) ProcessEvent(obj, _ interface{}, action centr
...
 		switch action {
 		case central.ResourceAction_REMOVE_RESOURCE:
 			s.serviceAccountStore.Remove(sa.ServiceAccount)
 		default:
 			s.serviceAccountStore.Add(sa.ServiceAccount)
 		}
+
+		// Since SourcedAutogeneratedIntegrations influences deployment detection, with re-sync disabled we have to make
+		// sure that deployments matching the ServiceAccount get reprocessed whenever a ServiceAccount is created / updated.
+		deploymentReference = resolver.ResolveDeploymentsByMultipleServiceAccounts([]resolver.NamespaceServiceAccount{
```
lvalerom, line 68: Same as above.

lvalerom (approved):
Looks good to me. Just a couple of non-blocking comments.

deploy/k8s/deploy.sh
fredrb: Removed

sensor/kubernetes/listener/resources/secrets.go
fredrb: Done

sensor/kubernetes/listener/resources/serviceaccount.go
fredrb: Done

PR #5338, opened 2023-03-21
Dashrews/ROX-16098 get pool propagate error

dashrews78:
## Description

In getPool we were logging a fatal error if we were unable to get a pool.  There are instances such as getting a pool for statistics where we do not want to fail.  If there is some flake that prevents a connection to get stats we can just log the error and move on.  We only need to fail when trying to getting connections for main workflow items and migrations.  So those errors are now passed back to the callers to choose to fail if necessary.

## Checklist

## Testing Performed

Updated the tests and worked with @vladbologa to verify this resolved the scenario where he encountered the issue running a local central with an external RDS DB.

### DIFF
--- a/central/centralhealth/service/service_impl.go
+++ b/central/centralhealth/service/service_impl.go
@@ -92,11 +92,9 @@ func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v
 				// Get a short-lived connection for the purposes of checking the version of the previous clone.
 				pool, err := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
 				if err != nil {
-					log.Infof("Unable to get previous database, leaving ForceRollbackTo empty.  %v", err)
+					return nil, errors.Wrapf(err, "Failed to retrieve previous database version.  %v", err)
 				}
-				if pool != nil {
-					defer pool.Close()
-				}
+				defer pool.Close()
 
 				// Get rollback to version
 				migVer, err := versionUtils.ReadVersionPostgres(pool)
--- a/central/centralhealth/service/service_impl.go
+++ b/central/centralhealth/service/service_impl.go
@@ -91,9 +91,11 @@ func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v
 
 				// Get a short-lived connection for the purposes of checking the version of the previous clone.
 				pool, err := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
-				defer pool.Close()
 				if err != nil {
 					log.Infof("Unable to get previous database, leaving ForceRollbackTo empty.  %v", err)
+				}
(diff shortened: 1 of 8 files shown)


vladbologa (commented):
Tested this locally and can confirm that Central doesn't restart anymore, and logs the timeouts instead.

central/centralhealth/service/service_impl.go
```diff
@@ -79,15 +79,24 @@ func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v
...
 				if err != nil {
 					return nil, errors.Wrapf(err, "Fail to get database size %s", migrations.PreviousDatabase)
 				}
 
 				// Get a short-lived connection for the purposes of checking the version of the previous clone.
-				pool := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
-				defer pool.Close()
+				pool, err := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
+				if err != nil {
+					log.Infof("Unable to get previous database, leaving ForceRollbackTo empty.  %v", err)
```
connorgorman, line 95: I'm not sure I understand the implication of this

dashrews78: After looking at it, this should probably return an error.  At this point, a previous exists, but if for some reason when we try to connect to it to get the version we cannot.  If we simply log the error and return an empty version for `ForceRollbackTo` that would be very misleading to the caller.  I will change it to return an error

central/centralhealth/service/service_impl.go
```diff
@@ -79,14 +79,21 @@ func (s *serviceImpl) GetUpgradeStatus(ctx context.Context, empty *v1.Empty) (*v
...
 				toBeFreedBytes, err = pgadmin.GetDatabaseSize(adminConfig, migrations.GetPreviousClone())
 				if err != nil {
 					return nil, errors.Wrapf(err, "Fail to get database size %s", migrations.PreviousDatabase)
 				}
 
 				// Get a short-lived connection for the purposes of checking the version of the previous clone.
-				pool := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
+				pool, err := pgadmin.GetClonePool(adminConfig, migrations.GetPreviousClone())
+				if err != nil {
+					return nil, errors.Wrapf(err, "Failed to retrieve previous database version.  %v", err)
```
c-du, line 95: nit: No need to add err to Wrapf since it is already wrapped.

c-du (approved):

connorgorman (changes requested):
For all instances, where we ignore an error and then return nil, can you add a comment about why that is okay?

dashrews78:
> For all instances, where we ignore an error and then return nil, can you add a comment about why that is okay?

That was a valuable exercise.  1 instance should have returned the err and another instance the block of code was totally unnecessary.

connorgorman (approved):

PR #5339, opened 2023-03-21
[ROX-15808] : Existing images should get re-scanned when IgnoreExistingImages option is set when calling EnrichImage

charmik-redhat:
### ROX-15808:
In `pkg/images/enricher/enricher_impl.go`, func `EnrichImage` gets called for enriching images with metadata, scan data and signature data. When `IgnoreExistingImages ` option is set, we should not look for existing image in database to populate the image scan and instead scan the image from scratch.
 
However, we are not checking for that option. Due to this, the image would be rescanned only when its SHA changes instead of getting rescanned each time when `IgnoreExistingImages` option was set. This behavior prevents watched images from getting rescanned every 4 hours. It also affects `ImageService.ScanImage` behavior and prevents a rescan if the image already exists  with a scan in db

The fix is to check for `IgnoreExistingImages` in `updateImageFromDatabase` and return without doing db lookup if the option is set. This will force fresh scan, signature and signature verification of the image.

## Checklist

## Testing Performed

Unit tests

Manual tests

### DIFF
--- a/central/image/datastore/store/postgres/store.go
+++ b/central/image/datastore/store/postgres/store.go
@@ -155,13 +155,6 @@ func (s *storeImpl) insertIntoImages(
 	if err := copyFromImageCves(ctx, tx, iTime, parts.vulns...); err != nil {
 		return err
 	}
-	if cloned.GetId() == "sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e" {
-		cveIDs := make([]string, 0)
-		for _, edge := range parts.imageCVEEdges {
-			cveIDs = append(cveIDs, edge.GetImageCveId())
-		}
-		log.Infof("ROX-16068: CVEs added to ImageCVEEdges = %v", cveIDs)
-	}
 	return copyFromImageCVEEdges(ctx, tx, iTime, false, parts.imageCVEEdges...)
 }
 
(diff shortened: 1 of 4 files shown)


central/image/datastore/store/postgres/store.go
```diff
@@ -483,9 +483,10 @@ func copyFromImageCVEEdges(ctx context.Context, tx *postgres.Tx, iTime *protoTyp
 
 	var err error
 	var oldEdgeIDs set.Set[string]
+	var objsToInsert []*storage.ImageCVEEdge
```
clickboo, line 486: nit: edgesToInsert

central/image/datastore/store/postgres/store.go
```diff
@@ -525,7 +531,7 @@ func copyFromImageCVEEdges(ctx context.Context, tx *postgres.Tx, iTime *protoTyp
 		})
 
 		// if we hit our batch size we need to push the data
```
clickboo, line 533: update comment?

charmik-redhat:
> @charmik-redhat The whole forceAdd logic is quite spaghetti in the `copyFromImageCVEEdges` function - there are at least 3-4 checks for forceAdd. Can you look into a separate func call which handles forceAdd logic for easier readability?

Done

md2119 (approved):

clickboo (approved):

PR #5340, opened 2023-03-21
Fix issues with Deployment Sidebar in Network Graph 2.0

vjwilson:
## Description

While doing manual scale testing, I happened to find 2 issues:

1. Very long deployment names crowd the Close (X) in the PatternFly sidebar layout, so I added a max-width to the deployment name in the title, so that long names wrap.
2. If an error occurs while retrieving deployment details--as happened during the scale test, when a deployment was deleted in the background--the page crashed. I coerced the error to a string, and moved down the position of the error message, so that the Close (X) is still visible.

## Checklist

## Testing Performed

1. Long deployment name wrapping
Screen Shot 2023-03-21 at 5 26 04 PM

2. Error message with Close button, instead of page crash
Screen Shot 2023-03-21 at 4 42 01 PM

### DIFF
--- a/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
+++ b/ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, CSSProperties } from 'react';
 import {
     Alert,
     AlertVariant,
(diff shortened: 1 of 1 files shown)


sachaudh:
@vjwilson this fix made me think of this ticket https://issues.redhat.com/browse/ROX-15410. Would this fix this bug as well? If not, is it worth just tweaking it to get another ticket knocked out too?

ui/apps/platform/src/Containers/NetworkGraph/deployment/DeploymentSideBar.tsx
```diff
@@ -36,6 +36,10 @@ import useSimulation from '../hooks/useSimulation';
 import { EdgeState } from '../components/EdgeStateSelect';
 import { deploymentTabs } from '../utils/deploymentUtils';
 
+const sidebarHeadingStyleConstant = {
+    '--pf-u-max-width--MaxWidth': '26ch',
```
sachaudh, line 40: Is the `26ch` from trial and error or from some system of values?

sachaudh (approved):

vjwilson:
> @vjwilson this fix made me think of this ticket https://issues.redhat.com/browse/ROX-15410. Would this fix this bug as well? If not, is it worth just tweaking it to get another ticket knocked out too?

It appears that the ask in ROX-15410 has already been done, based on sidebar in stagingdb.

Long-term, internationalizing the app may require adding scrolling back in, to account for languages where the tabs need to be wider.

PR #5350, opened 2023-03-22
Revert "Revert "Fix GitHub actions to allow for external contributions""

dhaus67:
Reverts stackrox/stackrox#5349

The previous PR had two issues:
- the `build-push` action failed as the docker buildx was not setup. Instead of using the `context.with`, I've resorted to using just `if` on the step.
- the `push` event when i.e. merging to master didn't trigger builds, since the `event.pull_request` was not set.

Additionally, also moved to `!github.event.pull_request.head.repo.fork` to indicate that the PR is not coming from a fork, since it seemed more readable.

As a sample, this now does the following on a fork with GitHub actions enabled:
- It will detect that the PR is running on a fork, hence the build jobs will be skipped: https://github.com/dhaus67/stackrox/actions/runs/4502247753/jobs/7924325522
- When merging, the push event will be detected, and the jobs will run (this is expected to fail on the fork repo since we do not have secrets configured there, but means the fix works and we do not suddenly "skip" building images): https://github.com/dhaus67/stackrox/actions/runs/4502578300

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -84,7 +84,6 @@ jobs:
         with:
           name: cli-build
           path: cli-build.tgz
-
   pre-build-go-binaries:
     runs-on: ubuntu-latest
     container:
@@ -123,7 +122,6 @@ jobs:
         with:
           name: go-binaries-build
           path: go-binaries-build.tgz
-
   pre-build-go-binaries-rcd:
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


janisz (approved):

.github/workflows/build.yaml
```diff
@@ -507,9 +538,10 @@ jobs:
 
     - name: Push to quay.io/rhacs-eng
       uses: docker/build-push-action@v4
+      if: |
+        github.event_name == 'push' || !github.event.pull_request.head.repo.fork
```
janisz, line 542: Is there any way to extract it to a variable?

dhaus67: Didn't find a way to extract these things to variables (especially since they somewhat rely on the context of the github object) 😞

PR #5354, opened 2023-03-22
ROX-16132 - Only Auto-Generated Image Integrations w/o Service Account Annotation

dcaravel:
## Description

Stops sensor from sending image integrations to central that have the `kubernetes.io/service-account.name` set

## Checklist

Did not find existing documentation stating the conditions by which autogenerated entries are created, therefore nothing to update.

## Testing Performed

Unit tests and infra cluster test


connorgorman (approved):

sensor/kubernetes/listener/resources/secrets.go
```diff
@@ -246,11 +246,15 @@ func (s *secretDispatcher) processDockerConfigEvent(secret, oldSecret *v1.Secret
 
 	sensorEvents := make([]*central.SensorEvent, 0, len(dockerConfig)+1)
 	registries := make([]*storage.ImagePullSecret_Registry, 0, len(dockerConfig))
+
+	saName := secret.GetAnnotations()[saAnnotation]
+	hasSAAnnotation := len(saName) > 0
```
connorgorman, line 251: do we need a separate var for this or can we just check `saName != ""`. I generally prefer this over the len check as a nit

dcaravel: changed, good to know the convention 👍

PR #5355, opened 2023-03-22
Increase Sensor resources by default

connorgorman:
## Description

Sensor has been under-provisioned by default. Bump the Sensor resources to a more reasonable level. If the resources have not been specified then these will be reconciled to these levels, if resources have been specified then they will remain as specified

## Checklist

## Testing Performed

Straightforward, but let's check CI


sthadka (approved):
LGTM!
(though might help to get another pass from any one else with Sensor expertise ;) )

dashrews78 (approved):

fredrb (approved):
Makes sense

lvalerom (approved):
It looks reasonable to me.

PR #5356, opened 2023-03-22
Look at the scope tree to evaluate layer index

md2119:
## Description

Look at the full scope tree and not just the parent to evaluate image component's layer index.

## Checklist

### DIFF
--- a/central/graphql/resolvers/image_components.go
+++ b/central/graphql/resolvers/image_components.go
@@ -244,10 +244,18 @@ func getDeploymentScope(scopeQuery *v1.Query, contexts ...context.Context) strin
 }
 
 func getImageIDFromScope(contexts ...context.Context) string {
+	var scope scoped.Scope
+	var hasScope bool
 	for _, ctx := range contexts {
-		if scope, ok := scoped.GetScope(ctx); ok {
-			if scope.Level == v1.SearchCategory_IMAGES {
+		if features.VulnMgmtWorkloadCVEs.Enabled() {
+			if scope, hasScope = scoped.GetScopeAtLevel(ctx, v1.SearchCategory_IMAGES); !hasScope {
 				return scope.ID
+			}
+		} else {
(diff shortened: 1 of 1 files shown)


charmik-redhat (approved):

dvail:
@md2119 Agreed, the snippet above is incomplete. What I was looking at doing was loading the expanded table async in a separate query, in order to avoid the deep nesting of fields that seems to be required to get all of the correct data.

The query in the document prevented getting data from the following:
- `layerIndex` (fixed by this PR)
- `location` (fixed by this PR)
- `fixedByVersion`, which is used in the queries on other pages, compared to `imageComponent->fixedIn` for this page

What I was attempting was running the following at the top level:
```graphql
query topLevelImageVulnQuery {
        image(id: $id) {
            id
            ...otherFields
            imageVulnerabilities(query: $vulnQuery, pagination: $pagination) {
                id
                severity
                isFixable
                cve
                summary
                discoveredAtImage
                imageComponentCount
            }
        }
}
```
followed by individual requests with the following when the user expands a row:
```graphql
query onDemandTableRowQuery {
        # includes a filter by CVE
        imageComponents(query: $vulnCveQuery) {
            name
            version
            location
            layerIndex
            # includes a filter by CVE
            imageVulnerabilities(query: $vulnCveQuery) {
                severity
                fixedByVersion
            }
        }
}
```

This would allow us to get all the data we need for this table as well as keep the query structure consistent with the other pages. The async loading isn't strictly necessary, but is a precaution to avoid evaluating the deeply nested query for data the user might not even see.
```
imageVulnerabilities {
    imageComponents {
        imageVulnerabilites {
        }
}
```

md2119:
> The async loading isn't strictly necessary, but is a precaution to avoid evaluating the deeply nested query for data the user might not even see.

The actionable data (docker layer and component version that fixes vuln) lies within the nested tables, so I believe users will want to view the nested tables more often. Therefore, consider it in your comparative analysis between load once vs load on click.

PR #5358, opened 2023-03-23
Jv rox 16135 process listening on ports close timestamp should be sent correctly to central

JoukoVirtanen:
## Description

Bugs prevented closed listening endpoints from being correctly sent from sensor to central. This fixes those bugs and gets the e2e tests and unit tests to pass.

## Checklist
- [x] e2e tests pass (Passed 12 times so far)
- [x] Unit tests pass

## Testing Performed

See above

### DIFF
--- a/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
@@ -16,7 +16,7 @@ import services.ProcessesListeningOnPortsService
 
 @Stepwise
 // TODO: Solve the flake and change back to @IgnoreIf({ !Env.get("ROX_POSTGRES_DATASTORE", null) })
-@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
+//@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
 class ProcessesListeningOnPortsTest extends BaseSpecification {
 
     // Deployment names
--- a/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
@@ -16,7 +16,7 @@ import services.ProcessesListeningOnPortsService
 
 @Stepwise
 // TODO: Solve the flake and change back to @IgnoreIf({ !Env.get("ROX_POSTGRES_DATASTORE", null) })
-//@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
+@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
 class ProcessesListeningOnPortsTest extends BaseSpecification {
 
     // Deployment names
--- a/sensor/common/networkflow/manager/manager_impl.go
+++ b/sensor/common/networkflow/manager/manager_impl.go
@@ -46,7 +46,6 @@ type hostConnections struct {
 	connections        map[connection]*connStatus
 	endpoints          map[containerEndpoint]*connStatus
 	lastKnownTimestamp timestamp.MicroTS
-	processes          map[string]*processInfo
 
 	// connectionsSequenceID is the sequence ID of the current active connections state
 	connectionsSequenceID int64
@@ -179,7 +178,7 @@ func (p *processInfo) String() string {
 type containerEndpoint struct {
 	endpoint    net.NumericEndpoint
 	containerID string
-	processKey  *processInfo
+	processKey  processInfo
 }
 
 func (e *containerEndpoint) String() string {
@@ -493,7 +492,7 @@ func (m *networkFlowManager) enrichProcessListening(ep *containerEndpoint, statu
 		key: processUniqueKey{
 			podID:         container.PodID,
 			containerName: container.ContainerName,
-			process:       *ep.processKey,
+			process:       ep.processKey,
 		},
 		port:     ep.endpoint.IPAndPort.Port,
 		protocol: ep.endpoint.L4Proto.ToProtobuf(),
@@ -523,8 +522,6 @@ func (m *networkFlowManager) enrichHostConnections(hostConns *hostConnections, e
 
 func deleteEndpoint(hostConns *hostConnections, ep containerEndpoint) {
 	delete(hostConns.endpoints, ep)
-	processID := getProcessInfoString(ep.processKey)
-	delete(hostConns.processes, processID)
 }
 
 func (m *networkFlowManager) enrichHostContainerEndpoints(hostConns *hostConnections, enrichedEndpoints map[containerEndpointIndicator]timestamp.MicroTS) {
@@ -550,7 +547,9 @@ func (m *networkFlowManager) enrichProcessesListening(hostConns *hostConnections
 
 	prevSize := len(hostConns.endpoints)
 	for ep, status := range hostConns.endpoints {
-		if ep.processKey == nil {
+		if ep.processKey.processArgs == "" &&
+			ep.processKey.processName == "" &&
+			ep.processKey.processExec == "" {
 			// No way to update a process if the data isn't there
 			continue
 		}
@@ -676,7 +675,6 @@ func (m *networkFlowManager) RegisterCollector(hostname string) (HostNetworkInfo
 			hostname:    hostname,
 			connections: make(map[connection]*connStatus),
 			endpoints:   make(map[containerEndpoint]*connStatus),
-			processes:   make(map[string]*processInfo),
 		}
 		m.connectionsByHost[hostname] = conns
 	}
@@ -744,7 +742,7 @@ func (m *networkFlowManager) UnregisterCollector(hostname string, sequenceID int
 
 func (h *hostConnections) Process(networkInfo *sensor.NetworkConnectionInfo, nowTimestamp timestamp.MicroTS, sequenceID int64) error {
 	updatedConnections := getUpdatedConnections(h.hostname, networkInfo)
-	updatedEndpoints := getUpdatedContainerEndpoints(h.hostname, networkInfo, h.processes)
+	updatedEndpoints := getUpdatedContainerEndpoints(h.hostname, networkInfo)
 
 	collectorTS := timestamp.FromProtobuf(networkInfo.GetTime())
 	tsOffset := nowTimestamp - collectorTS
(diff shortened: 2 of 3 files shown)


sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -878,7 +919,7 @@ func getUpdatedContainerEndpoints(hostname string, networkInfo *sensor.NetworkCo
 				IPAndPort: getIPAndPort(endpoint.GetListenAddress()),
 				L4Proto:   net.L4ProtoFromProtobuf(endpoint.GetProtocol()),
 			},
-			processKey: getProcessKey(endpoint.GetOriginator()),
+			processKey: getProcessKey(endpoint.GetOriginator(), processes),
```
JoukoVirtanen, line 922: This was a bug previously as ep was being used as a key, but due to the use of a pointer the same endpoint would appear multiple times in the map.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -58,9 +59,10 @@ type hostConnections struct {
 }
 
 type connStatus struct {
-	firstSeen timestamp.MicroTS
-	lastSeen  timestamp.MicroTS
-	used      bool
+	firstSeen   timestamp.MicroTS
+	lastSeen    timestamp.MicroTS
+	used        bool
+	usedProcess bool
```
JoukoVirtanen, line 65: Don't delete an endpoint unless it has been used by both the regular networkflows path and processes listening on ports paths.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -612,13 +626,16 @@ func computeUpdatedProcesses(current map[processListeningIndicator]timestamp.Mic
 
 	for pl, currTS := range current {
 		prevTS, ok := previous[pl]
-		if !ok || currTS > prevTS {
+		if !ok || currTS > prevTS  || (prevTS == timestamp.InfiniteFuture && currTS != timestamp.InfiniteFuture) {
```
JoukoVirtanen, line 629: nil timestamps are set to timestamp.InfiniteFuture, but we want to treat them as if they are 0 so they are in the past, not the future.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -490,7 +496,7 @@ func (m *networkFlowManager) enrichProcessListening(ep *containerEndpoint, statu
 
 	// Multiple endpoints from a collector can result in a single enriched endpoint,
 	// hence update the timestamp only if we have a more recent endpoint than the one we have already enriched.
-	if oldTS, found := processesListening[indicator]; !found || oldTS < status.lastSeen {
+	if oldTS, found := processesListening[indicator]; !found || oldTS < status.lastSeen || oldTS == timestamp.InfiniteFuture {
```
JoukoVirtanen, line 499: It seems likely that this change is not needed, as if it were a bug it would be negatively impacting other parts of the code that have been in production for a long time. Need to check this. Similar comment for other similar changes.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -510,16 +516,24 @@ func (m *networkFlowManager) enrichHostConnections(hostConns *hostConnections, e
 	flowMetrics.HostConnectionsRemoved.Add(float64(prevSize - len(hostConns.connections)))
 }
 
+func deleteEndpoint(hostConns *hostConnections, ep containerEndpoint) {
+
```
Molter73, line 520:
```suggestion
```

sensor/common/networkflow/manager/manager_impl.go
erthalion: Add this as the code commentary to the structure field, together with the details why keeping an endpoint is needed for PLOP correctness.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -510,16 +516,24 @@ func (m *networkFlowManager) enrichHostConnections(hostConns *hostConnections, e
...
 func (m *networkFlowManager) enrichHostContainerEndpoints(hostConns *hostConnections, enrichedEndpoints map[containerEndpointIndicator]timestamp.MicroTS) {
 	hostConns.mutex.Lock()
 	defer hostConns.mutex.Unlock()
 
 	prevSize := len(hostConns.endpoints)
 	for ep, status := range hostConns.endpoints {
 		m.enrichContainerEndpoint(&ep, status, enrichedEndpoints)
-		if status.used && status.lastSeen != timestamp.InfiniteFuture {
+		if status.used && (status.usedProcess || !features.ProcessesListeningOnPort.Enabled()) && status.lastSeen != timestamp.InfiniteFuture {
```
erthalion, line 533: Please, untangle this condition, it's unreadable. Maybe introduce a local `used` variable computed from the status and the feature enabled or such.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -537,9 +551,10 @@ func (m *networkFlowManager) enrichProcessesListening(hostConns *hostConnections
 		}
 
 		m.enrichProcessListening(&ep, status, processesListening)
-		if status.used && status.lastSeen != timestamp.InfiniteFuture {
+		if status.used && status.usedProcess && status.lastSeen != timestamp.InfiniteFuture {
```
erthalion, line 554: Why there is no need for a condition on PLOP being enabled?

sensor/common/networkflow/manager/manager_impl.go
erthalion: Wouldn't passing by value instead passing by pointer reference solve this problem?

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Done

JoukoVirtanen: Done

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Done

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: I changed `processKey` to be of type `processInfo`.

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Removed

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -792,12 +810,24 @@ func (h *hostConnections) Process(networkInfo *sensor.NetworkConnectionInfo, now
...
+
+func getProcessOriginatorString(originator *storage.NetworkProcessUniqueKey) string {
+	return getProcessString(originator.ProcessName, originator.ProcessArgs, originator.ProcessExecFilePath)
+}
+
+func getProcessInfoString(process *processInfo) string {
+	return getProcessString(process.processName, process.processArgs, process.processExec)
+}
+
+func getProcessKey(originator *storage.NetworkProcessUniqueKey) processInfo {
```
Molter73, line 825: I think we are now ok to remove these.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -510,16 +514,22 @@ func (m *networkFlowManager) enrichHostConnections(hostConns *hostConnections, e
 	flowMetrics.HostConnectionsRemoved.Add(float64(prevSize - len(hostConns.connections)))
 }
 
+func deleteEndpoint(hostConns *hostConnections, ep containerEndpoint) {
+	delete(hostConns.endpoints, ep)
+}
```
Molter73, line 519: I believe this is now not needed.

Molter73 (approved):

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Using the `processes` map could reduce memory if one process opened many endpoints. Then you would just have one string as the key and one pointer to `processInfo` in the map, and each containerEndpoint would have the same pointer instead of having a `procesInfo` for each containerEndpoint. It could save memory or use more of it depending upon how many endpoints the average process has. Anyway it complicates things and probably uses more memory in most cases.

qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -16,7 +16,7 @@ import services.ProcessesListeningOnPortsService
 
 @Stepwise
 // TODO: Solve the flake and change back to @IgnoreIf({ !Env.get("ROX_POSTGRES_DATASTORE", null) })
-@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
+//@IgnoreIf({ Env.get("RUN_PROCESS_LISTENING_ON_PORTS_E2E_TESTS", "false") == "false" })
```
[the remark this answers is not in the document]
JoukoVirtanen: This should be left uncommented for now. I commented it out for local testing and forgot to uncomment before pushing.

sensor/common/networkflow/manager/manager_impl.go
```diff
@@ -531,14 +539,17 @@ func (m *networkFlowManager) enrichProcessesListening(hostConns *hostConnections
 
 	prevSize := len(hostConns.endpoints)
 	for ep, status := range hostConns.endpoints {
-		if ep.processKey == nil {
+		if ep.processKey.processArgs == "" &&
+			ep.processKey.processName == "" &&
+			ep.processKey.processExec == "" {
```
fredrb, line 544: [nitpick] I understand you have to make these checks now because `ep.processKey` is no longer a pointer, but the actual value. But this is can be a problem to maintain in the future in case this is used elsewhere and the same check has to be replicated or if the number of properties in `processInfo` changes. 

I have couple of suggestions:
1) Extract this check to a `func (p processInfo) empty() bool` receiver function.
2) Check only a single property. It seems like that it's either all set or none. Not sure if this is always true though
3) Create an `emptyProcessInfo` var definition that can be compared to. If more properties are added you don't need to update the check. Like:
```go
var (
	emptyProcessInfo = processInfo{}
)

func someFn() {
	// ...
	p := getProcessKey(nil)
	if p == emptyProcessInfo {
		// ...
	}
}
```

fredrb (approved):
Only checked the code since I don't have much knowledge on the feature. 

Looks good to me, only one small optional change.

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Removed

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: Removed

sensor/common/networkflow/manager/manager_impl.go
JoukoVirtanen: I went with option number 3.

PR #5360, opened 2023-03-23
Quickfix: deploy/common/k8sbased.sh not passing loglevel to sensor

fredrb:
## Description

This PR fixes two small issues with our deploy scripts:

1. `LOGLEVEL` was not passed to sensor if exported in the host when deploying. `LOGLEVEL="debug" ./deploy/k8s/deploy.sh` only sets `LOGLEVEL` in Central. With this, it will also inject in Sensor.
2. `ROX_RESYNC_DISABLED` was being set by patching sensor. This is a bit annoying because it will cause sensor to restart right after it's deployed. By injecting with `envVars` it will be deployed with the correct property from the beginning. 

## Checklist

## Testing Performed

- Deployed using `LOGLEVEL="debug" ROX_RESYNC_DISABLED="true" ./deploy/k8s/deploy.sh`

### DIFF
--- a/deploy/common/k8sbased.sh
+++ b/deploy/common/k8sbased.sh
@@ -529,6 +529,20 @@ function launch_sensor {
         helm_args+=(--set "helmManaged=false")
       fi
 
+      if [[ -n "$LOGLEVEL" ]]; then
+        helm_args+=(
+          --set customize.envVars.LOGLEVEL="${LOGLEVEL}"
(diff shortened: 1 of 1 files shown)


vikin91 (approved):

jschnath (approved):

lvalerom (approved):
Yes! 🎉

PR #5361, opened 2023-03-23
ROX-15904: Remove migrated code

Maddosaurus:
## Description

As the caching scanner code was moved to Scanner in https://github.com/stackrox/scanner/pull/1130, we can now remove it from Compliance.
As long as #5292 is not merged, the feature will not work without manual intervention, but Compliance will not crash and continue to function correctly.
The needed manual intervention would be to update the Collector DaemonSet to add the Node Scanning image to it.

## Checklist


jschnath (approved):
LGTM, not much to review here

PR #5362, opened 2023-03-23
Remove unused crypt utils ecdsa and es25519

janisz:
## Description

Looks like ecdsa and es25519 is not used anywhere. Let's remove it.

## Checklist

## Testing Performed

CI


janisz:
@0x656b694d I'm not 100% sure if ecdsa is not used. We can handle it in another PR cc: @SimonBaeumer

0x656b694d:
> @0x656b694d I'm not 100% sure if ecdsa is not used. We can handle it in another PR cc: @SimonBaeumer

But you're removing ecdsa in this PR.

janisz:
Nope, I'm removing verifier and `Signer` and `SignatureVerifier` with all implementations not the key.

PR #5363, opened 2023-03-23
ROX-1137: Add support for GHA in GKE deployment

janisz:
## Description

Handle GH variables in GKE deploy to allow using it in workflows.

## Checklist

## Testing Performed

- https://github.com/stackrox/jenkins-plugin/pull/234

### DIFF
--- a/scripts/ci/gke.sh
+++ b/scripts/ci/gke.sh
@@ -98,24 +98,23 @@ create_cluster() {
     if is_OPENSHIFT_CI; then
         require_environment "JOB_NAME"
         require_environment "BUILD_ID"
-        tags="${tags},stackrox-ci-${JOB_NAME:0:50}"
-        tags="${tags/%-/x}"
-        labels="${labels},stackrox-ci-job=${JOB_NAME:0:63}"
-        labels="${labels/%-/x}"
-        labels="${labels},stackrox-ci-build-id=${BUILD_ID:0:63}"
-        labels="${labels/%-/x}"
+        build_num="${BUILD_ID}"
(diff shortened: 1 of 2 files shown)


gavin-stackrox (approved):

scripts/ci/gke.sh
```diff
@@ -101,6 +104,15 @@ create_cluster() {
...
         labels="${labels/%-/x}"
+    elif is_GITHUB_ACTIONS; then
+        require_environment "GITHUB_JOB"
+        require_environment "GITHUB_RUN_ID"
+        tags="${tags},stackrox-ci-${GITHUB_JOB:0:50}"
+        tags="${tags/%-/x}"
+        labels="${labels},stackrox-ci-job=${GITHUB_JOB:0:63}"
+        labels="${labels/%-/x}"
+        labels="${labels},stackrox-ci-build-id=${GITHUB_RUN_ID:0:63}"
+        labels="${labels/%-/x}"
```
porridge, line 115: How about introducing intermediate variables for job name and run id, common for all branches here, and then have a single common piece of code for squeezing them into tags and labels, after the `fi`? Similar to how we define `$build_num` in `assign_env_variables`.

This way we would not repeat the logic that does truncation and replacement of trailing dash. And then we could maybe even add a comment on what is happening, because `${foo/%-/x}` is not exactly obvious to a non-bash-expert and the reason for truncation similarly (is it because we're making assumptions about the structure of the value, or just GKE has a length limitation).

porridge (approved):
Tentative LGTM, see comment inline.

PR #5364, opened 2023-03-23
ROX-15462: Add sarif output format to roxctl

dhaus67:
## Description

This PR adds the SARIF output format to roxctl.

The [SARIF output format](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html) is a standard for analysis tools to share their results. Most notably, its required by GitHub to support publishing code scanning results ([more info can be found here](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning)).

The following are noteworthy changes within this PR:
- a new custom modifier for gjson has been added: `@text`, which creates a single text from a JSON object. This is used to create a multiline help message within the SARIF report consisting of multiple values (e.g. a component's CVE, name, version, fixed version).
- a new printer that is capable of printing SARIF reports.
- commands which should make use of the SARIF output format have been adjusted to include their relevant JSON path expressions, the commands are: `roxctl image scan, roxctl image check, roxctl deployment check`.

Additional notes for the PR / caveats:
- Currenlty, the location within the SARIF report will be set, but to default values (i.e. line 1 column 1). We cannot leave the location empty, as it is required by the SARIF specification.
- The short and full description of a rule currently show the _same_ output. This is due to us not really having separate informations for the information.
- Currently, no markdown support is added within the report for messages. We may add this at a latter point.
- The library `github.com/owenrumney/go-sarif/v2/sarif` has been used to create the SARIF support. If we want to avoid pulling in the library in the future, we can generate required structs from [SARIF JSON 
schema](https://github.com/oasis-tcs/sarif-spec/blob/main/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json). I decided against it simply for time reasons, but am happy to make the change in a follow-up.

## Checklist

## Testing Performed

- see added unit tests.

Manual tests that the integration works with e.g. GitHub:
- Created a separate repository https://github.com/dhaus67/roxctl-sarif-test
- The [action](https://github.com/dhaus67/roxctl-sarif-test/blob/main/.github/workflows/scan.yaml) currently does the following: a) scan an image via an ACSCS instance b) upload the sarif format
- SARIF format output: https://github.com/dhaus67/roxctl-sarif-test/security/code-scanning

### DIFF
--- a/pkg/gjson/modifiers_test.go
+++ b/pkg/gjson/modifiers_test.go
@@ -124,7 +124,7 @@ func TestTextModifier(t *testing.T) {
 			{
 				Name:    "Hermione Granger",
 				Address: "Heathgate",
-				Age:     18,
+				Age:     19,
 			},
 		},
 	}
@@ -137,15 +137,15 @@ func TestTextModifier(t *testing.T) {
 	}{
 		"without custom column names": {
 			expression: "{textTest.#.name,textTest.#.age,textTest.#.address}.@text",
-			result:     "[\"name Harry Potter\\nage 18\\naddress Privet Drive\",\"name Ron Weasley\\nage 18\\naddress The Burrow\",\"name Hermione Granger\\nage 18\\naddress Heathgate\"]",
+			result:     "[\"name Harry Potter\\nage 18\\naddress Privet Drive\",\"name Ron Weasley\\nage 18\\naddress The Burrow\",\"name Hermione Granger\\nage 19\\naddress Heathgate\"]",
 		},
 		"without modifier should not modify the output": {
 			expression: "{textTest.#.name,textTest.#.age,textTest.#.address}",
-			result:     "{\"name\":[\"Harry Potter\",\"Ron Weasley\",\"Hermione Granger\"],\"age\":[18,18,18],\"address\":[\"Privet Drive\",\"The Burrow\",\"Heathgate\"]}",
+			result:     "{\"name\":[\"Harry Potter\",\"Ron Weasley\",\"Hermione Granger\"],\"age\":[18,18,19],\"address\":[\"Privet Drive\",\"The Burrow\",\"Heathgate\"]}",
 		},
 		"with custom column names": {
 			expression: "{\"Super Cool Name\":textTest.#.name,textTest.#.age,textTest.#.address}.@text",
-			result:     "[\"Super Cool Name Harry Potter\\nage 18\\naddress Privet Drive\",\"Super Cool Name Ron Weasley\\nage 18\\naddress The Burrow\",\"Super Cool Name Hermione Granger\\nage 18\\naddress Heathgate\"]",
+			result:     "[\"Super Cool Name Harry Potter\\nage 18\\naddress Privet Drive\",\"Super Cool Name Ron Weasley\\nage 18\\naddress The Burrow\",\"Super Cool Name Hermione Granger\\nage 19\\naddress Heathgate\"]",
 		},
 		"with singular value": {
 			expression: "{textTest.0.name,textTest.0.age,textTest.0.address}.@text",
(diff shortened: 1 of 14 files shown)


dhaus67:
* master
  * **PR #5364** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5364" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5364?utm_source=stack-comment).

pkg/gjson/modifiers_test.go
```diff
@@ -95,3 +107,138 @@ func TestBoolReplaceModifier(t *testing.T) {
...
+					},
+					{
+						Name:    "Ron Weasley",
+						Address: "The Burrow",
+						Age:     18,
+					},
+					{
+						Name:    "Hermione Granger",
+						Address: "Heathgate",
+						Age:     18,
```
janisz, line 194: nit: 
```suggestion
						Age:     19,
```
Source: https://www.quora.com/Why-is-Hermione-in-the-same-year-as-Harry-and-Ron-if-shes-older-than-them

pkg/printers/sarif.go
```diff
@@ -0,0 +1,212 @@
...
+
+	run.AddResult(sarif.NewRuleResult(entry.ruleID).
+		WithLevel(toSarifLevel(entry.severity)).
+		// Reusing the help here, since the help includes remediation information.
+		WithMessage(sarif.NewMessage().WithText(entry.help)).
+		WithLocations([]*sarif.Location{
+			{
+				PhysicalLocation: &sarif.PhysicalLocation{
+					ArtifactLocation: sarif.NewArtifactLocation().WithUri(entity),
+					Region: sarif.NewRegion().
```
janisz, line 158: `NewSimpleRegion(startLine, endLine int)` ?
Do we need to specify all of it? I know it's required but even github examples does not contain end linte. https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#physicallocation-object

pkg/printers/sarif.go
```diff
@@ -0,0 +1,212 @@
...
+	run.AddRule(entry.ruleID).
+		WithName(name).
+		WithShortDescription(sarif.NewMultiformatMessageString(entry.description)).
+		WithFullDescription(sarif.NewMultiformatMessageString(entry.description)).
+		WithHelp(sarif.NewMultiformatMessageString(entry.help))
+
+	run.AddResult(sarif.NewRuleResult(entry.ruleID).
+		WithLevel(toSarifLevel(entry.severity)).
+		// Reusing the help here, since the help includes remediation information.
+		WithMessage(sarif.NewMessage().WithText(entry.help)).
```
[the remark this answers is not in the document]
janisz: We can handle it in another PR

pkg/printers/sarif.go
dhaus67: Works as well, not sure why GH specifies it as required 🤷

pkg/printers/sarif.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func addEntry(run *sarif.Run, entry sarifEntry, entity string, name string) {
+	run.AddRule(entry.ruleID).
+		WithName(name).
+		WithShortDescription(sarif.NewMultiformatMessageString(entry.description)).
+		WithFullDescription(sarif.NewMultiformatMessageString(entry.description)).
+		WithHelp(sarif.NewMultiformatMessageString(entry.help))
+
+	run.AddResult(sarif.NewRuleResult(entry.ruleID).
```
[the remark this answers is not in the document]
dhaus67: Hm, yeah I guess we could, however GitHub already has the `Tool` information that's filled out, so not sure whether this would be too much, but I'm happy with the change.

dhaus67:
@janisz I've addressed the feedback we discussed today, here's the latest changes:
- SARIF results from vulnerability reports will have the tag `security` and specify the `security-severity` now. Example result: https://github.com/dhaus67/roxctl-sarif-test/security/code-scanning/2743
- SARIF results from policy reports will have the tag `policy-violations` (making it possible to filter only policy-violations). Example result: https://github.com/dhaus67/roxctl-sarif-test/security/code-scanning/2785

I've kept the `Severity` field, since its the "red hat internal" severity, and additionally added it as a tag to make it filterable within GitHub UI.

janisz (approved):

PR #5365, opened 2023-03-23
ROX-16137 Re-enqueue deployments based on received central messages

lvalerom:
## Description

Some events coming from central need to trigger deployment reprocessing (e.g. `UpdatedImage`) This moves the handling of these events from the `Detector` to the `EventPipeline` and triggers the required deployments reprocessing.

## Checklist

## Testing Performed

* CI
* [x] Manual testing


sensor/common/imagecacheutils/common.go
```diff
@@ -15,3 +15,11 @@ func GetImageCacheKey(provider CacheKeyProvider) string {
 	}
 	return provider.GetName().GetFullName()
 }
+
+// CompareImageCacheKey given two CacheKeyProvider, compares if they're equal
+func CompareImageCacheKey(a CacheKeyProvider, b CacheKeyProvider) bool {
+	if a.GetId() != "" && b.GetId() != "" {
+		return a.GetId() == b.GetId()
+	}
+	return a.GetName().GetFullName() == b.GetName().GetFullName()
```
fredrb, line 24: Since `a.GetName()` is a proto message, you can compare them with `proto` package, like:
```suggestion
	return proto.Equal(a.GetName(), b.GetName())
```

sensor/common/imagecacheutils/common.go
```diff
@@ -15,3 +15,11 @@ func GetImageCacheKey(provider CacheKeyProvider) string {
 	}
 	return provider.GetName().GetFullName()
 }
+
+// CompareImageCacheKey given two CacheKeyProvider, compares if they're equal
+func CompareImageCacheKey(a CacheKeyProvider, b CacheKeyProvider) bool {
```
fredrb, line 20:
```suggestion
func CompareImageCacheKey(a, b CacheKeyProvider) bool {
```

sensor/common/store/resolver/resolver.go
```diff
@@ -49,3 +50,10 @@ func ResolveAllDeployments() DeploymentReference {
 		return ids
 	}
 }
+
+// ResolveDeploymentsByImages returns a function that returns a list of deployment ids based on a slice of images
+func ResolveDeploymentsByImages(images ...imagecacheutils.CacheKeyProvider) DeploymentReference {
```
fredrb, line 55: Nit: I'm a little uneasy with using `imagecacheutils.CacheKeyProvider` as the abstraction for the resolver and store functions. Here's my reasoning:

`imagecacheutils.CacheKeyProvider` surfaces an _internal_ design decision of the resolver function: we are using `imagecacheutils` functions to compare the image entries. On the other hand, `storage.*` type object are the canonical model of all ACS entities. Which I think is a much better "common" interface for communicating deployment resolution. Additionally, this ties the method interface to `imagecacheutils`. If in the future we want to index deployment differently using another property of `storage.Image`, we'd have to change the interface or the method signatures. 

With that said, I would prefer to use `storage.Image` in the resolver and store methods.

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -73,13 +77,47 @@ func (e *ResourceEvent) AddDeploymentForReprocessing(ids ...string) *ResourceEve
...
 }
 
-// AddDeploymentReference creates and sets a new deployment reference to this resource event.
-func (e *ResourceEvent) AddDeploymentReference(reference resolver.DeploymentReference, action central.ResourceAction, force bool) {
-	e.DeploymentReferences = append(e.DeploymentReferences, DeploymentReference{
+// DeploymentReferenceOption option function for the DeploymentReference struct
+type DeploymentReferenceOption func(*DeploymentReference)
+
+// DeploymentRefWithParentResourceAction sets the ParentResourceAction
+func DeploymentRefWithParentResourceAction(action central.ResourceAction) DeploymentReferenceOption {
```
fredrb, line 84: I think the prefix is unnecessary. It reads a bit odd on the client call. Let's call it just `WithParentResourceAction`. 

If you don't want to make it confusing for readers that we have different `With*` prefixed functions in the same file, I think it makes sense to move all `DeploymentReference` code into a new file in the same package.

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -73,13 +77,47 @@ func (e *ResourceEvent) AddDeploymentForReprocessing(ids ...string) *ResourceEve
...
+	deploymentReference := DeploymentReference{
 		Reference:            reference,
-		ParentResourceAction: action,
-		ForceDetection:       force,
-	})
+		ParentResourceAction: central.ResourceAction_UPDATE_RESOURCE,
+		ForceDetection:       false,
+		SkipResolving:        false,
+	}
+	for _, o := range options {
```
fredrb, line 112: nit: `s/o/opt`

sensor/kubernetes/eventpipeline/component/message.go
```diff
@@ -73,13 +77,47 @@ func (e *ResourceEvent) AddDeploymentForReprocessing(ids ...string) *ResourceEve
...
+
+// DeploymentRefWithForceDetection sets whether the detection should be forced or not
+func DeploymentRefWithForceDetection(force bool) DeploymentReferenceOption {
+	return func(dr *DeploymentReference) {
+		dr.ForceDetection = force
+	}
+}
+
+// DeploymentRefWithSkipResolving sets whether the deployment should be resolved or not
+func DeploymentRefWithSkipResolving(skip bool) DeploymentReferenceOption {
```
fredrb, line 98: I think the `bool` arg opt functions will read much better without the param. It's defaulted to `false` anyway, so there's no reason to set it to `false` when calling the opt. Just do it:
```go
func DeploymentRefWithSkipResolving() DeploymentReferenceOption {
	return func(dr *DeploymentReference) {
		dr.SkipResolving = true
	}
```

sensor/kubernetes/eventpipeline/pipeline_impl.go
```diff
@@ -90,3 +111,92 @@ func (p *eventPipeline) forwardMessages() {
...
+
+func (p *eventPipeline) processReassessPolicies() error {
+	log.Debug("ReassessPolicies message received from central")
+	if err := p.detector.ProcessReassessPolicies(); err != nil {
+		return err
+	}
+	if env.ResyncDisabled.BooleanSetting() {
+		message := component.NewEvent()
+		message.AddDeploymentReference(resolver.ResolveAllDeployments(),
+			component.DeploymentRefWithForceDetection(true),
```
fredrb, line 128: This will read much better as:
```go
message.AddDeploymentReference(resolver.ResolveAllDeployments(),
			component.WithForceDetection(),
			component.WithSkipResolving())
```

sensor/kubernetes/eventpipeline/pipeline_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/concurrency"
+	"github.com/stackrox/rox/pkg/sync"
+	mocks2 "github.com/stackrox/rox/sensor/common/detector/mocks"
```
fredrb, line 11: `s/mocks2/mockDetector`

sensor/kubernetes/eventpipeline/pipeline_test.go
```diff
@@ -0,0 +1,235 @@
...
+	mocks3 "github.com/stackrox/rox/sensor/common/reprocessor/mocks"
```
fredrb, line 12: `s/mocks3/mockReprocessor`

sensor/kubernetes/eventpipeline/pipeline_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component"
+	"github.com/stackrox/rox/sensor/kubernetes/eventpipeline/component/mocks"
```
fredrb, line 14: maybe add named import here also: `mockComponent`

sensor/common/reprocessor/handler.go
fredrb, line 1: As we discussed, maybe this component shouldn't exist at all. Maybe let's do a follow-up to this change to try and remove this component entirely. The work that is being done here should be done by the pipeline component.

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -181,6 +210,15 @@ func (ds *DeploymentStore) Get(id string) *storage.Deployment {
 	return wrap.GetDeployment()
 }
 
+// GetBuiltDeployment returns a cloned deployment for supplied id
+func (ds *DeploymentStore) GetBuiltDeployment(id string) *storage.Deployment {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+	// TODO(ROX-16138): return nil if the deployment is not built yet
+	wrap := ds.getWrapNoLock(id)
+	return wrap.GetDeployment().Clone()
```
fredrb, line 219: I think this will panic if `wrap` is nil. We should add a `nil` check here before returning.

sensor/kubernetes/eventpipeline/resolver/resolver_impl.go
```diff
@@ -68,6 +68,16 @@ func (r *resolverImpl) processMessage(msg *component.ResourceEvent) {
 					log.Warnf("Deployment with id %s not found", id)
 					continue
 				}
+				// Skip resolving the deployment dependencies
+				if deploymentReference.SkipResolving {
+					d := r.storeProvider.Deployments().GetBuiltDeployment(id)
+					if d == nil {
+						log.Warnf("Deployment with id %s not built", id)
```
fredrb, line 75: This warning makes sense after `ROX-16138` is done. However, we know that we can't take for granted that the improvements will be made soon. So maybe let's change this warning for now into something that makes sense with the current state. 
```suggestion
						log.Warnf("Deployment with id %s not found", id)
```

fredrb (changes requested):
Haven't checked the tests yet.

sensor/common/imagecacheutils/common.go
lvalerom: I kind of want to keep it like that. I’ve seen in other parts of the code that we only compare the ID and, if this one is missing, the FullName, `proto.Equal` will compare all the contents of the `ImageName` but in some cases some other fields might not be set. For example; when we received the `InvalidateImageCache` the content of that request does not have all the fields in the `ImageName` only `Id` and `ImageFullName`.

sensor/common/imagecacheutils/common.go
lvalerom: Done

sensor/common/store/resolver/resolver.go
lvalerom: I had `storage.Image` at first and I changed it to that interface to do it the “go way” (maybe it would’ve been better to define a new interface in the store package). I agree with your reasoning though and since we are using storage structures everywhere else I’ll change it back to `storage.Image`.

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: Done.

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: Done.

sensor/kubernetes/eventpipeline/component/message.go
lvalerom: Done.

sensor/kubernetes/eventpipeline/pipeline_impl.go
lvalerom: Agree.

sensor/kubernetes/eventpipeline/pipeline_test.go
lvalerom: Done.

sensor/kubernetes/eventpipeline/pipeline_test.go
lvalerom: Done.

sensor/kubernetes/eventpipeline/pipeline_test.go
lvalerom: Done.

sensor/common/reprocessor/handler.go
lvalerom: Agree, let’s follow-up on this and remove this component in another PR.

sensor/kubernetes/listener/resources/deployment_store.go
lvalerom: I could be wrong but I don’t think it would panic because `GetDeployment` and `Clone` check whether the caller is nil or not.

sensor/kubernetes/eventpipeline/resolver/resolver_impl.go
lvalerom: Done.

sensor/kubernetes/listener/resources/deployment_store.go
fredrb: You're right, it does. The following test passes:
```go
func (s *deploymentStoreSuite) Test_BuildDeployment() {
	d := s.deploymentStore.GetBuiltDeployment("random")
	s.Require().Nil(d)
}
```

sensor/kubernetes/listener/resources/deployment_store.go
```diff
@@ -167,10 +168,38 @@ func (ds *DeploymentStore) FindDeploymentIDsByLabels(namespace string, sel selec
...
+				// The deployment id is already the set, we can break here
+				break
+			}
+		}
+	}
+	return ids
+}
+
+// FindDeploymentIDsByImages returns a slice of deployment ids based on matching images
+func (ds *DeploymentStore) FindDeploymentIDsByImages(images []*storage.Image) []string {
```
fredrb, line 186: Maybe we could add some tests in deployment store for this function?

fredrb (approved):
Thanks for addressing the review. Looks good now. 
It would be good to have some additional tests in `deployment_store_test` for the new filter function you created.

sensor/kubernetes/listener/resources/deployment_store.go
lvalerom: I've added some tests.

sensor/kubernetes/listener/resources/deployment_store_test.go
```diff
@@ -228,6 +229,114 @@ func (s *deploymentStoreSuite) Test_FindDeploymentIDsByLabels() {
...
+			},
+			expectedIDs: []string{"uuid-2"},
+		},
+		"Match against mixed images": {
+			images: []*storage.Image{
+				newImage("", "docker.io/library/nginx:1.2.3"),
+				newImage("id-private-registry.io/nginx:1.2.3", ""),
+			},
+			expectedIDs: []string{"uuid-1", "uuid-2"},
+		},
```
fredrb, line 329: I'm missing a case where the image ID matches two different deployments with different image names. 

I think this highlights something confusing about your test case. When first reading through them, I thought that `docker.io/library/nginx:1.2.3` and `private-registry.io/nginx:1.2.3` would resolve to the same image ID. And that would make the cases easier to understand IMO. 

I think you should have three deployment/images: 
- `docker.io/library/nginx:1.2.3`
- `private-registry.io/nginx:1.2.3`
- `private-registry.io/main:3.2.1`

Then, additionally to the test cases you have (which should be changed to match `docker.io/library/nginx:1.2.3` and `private-registry.io/main:3.2.1`. You could add the following test case:
```go
"Match against same image ID with different paths": {
		images: []*storage.Image{
			newImage("", imageID1, "docker.io/library/nginx:1.2.3"),
		},
		expectedIDs: []string{"uuid-1", "uuid-2"},
},
```

fredrb (commented):
Thanks for adding the tests. But I think we could improve them by adding a case where image IDs are the same. Let me know what do you think.

sensor/kubernetes/listener/resources/deployment_store_test.go
lvalerom: Makes sense. I've updated the tests to have some deployments with matching image ids.

fredrb (approved):

PR #5382, opened 2023-03-24
Replace unused params with _

janisz:
## Description

This PR enable unused-parameter rule in linter and fixes all issues.
To perform this I enable reporting all issues and json output  in `.golangci.yml`
```yaml
output:
  format: "junit-xml:report.xml,colored-line-number,json:out.json"

issues:
  max-issues-per-linter: 0
  max-same-issues: 0
```
Then with one-liner I prepared a script to replace all unused params with `_`
```bash
cat out.json | jq '.Issues[] | [.Pos.Line, .Text, .Pos.Filename] |@csv' -r | grep unused | sed 's/"unused-parameter: parameter \x27/ /g' | sed 's/\x27 seems to be unused, consider removing or renaming it as _"/ /g' > work.sh
```
I modified `work.sh` manually with multicursor (as I'm not proficient in sed) to make every line in format of 
```bash
sed -i '<line number>s/<param name> /_ /' <file name>
```
Finally I applied the script with `source work.sh`.
This procedure was repeated multiple times until no issue was reported.

## Checklist

## Testing Performed

CI


parametalol (approved):
Nice

ebensh:
I don't want you to have to redo the work if it's not necessary, so this is just a question, not a suggestion: did you consider replacing `param` with `_param` or `_param_unused` or `_unused_param`?

janisz:
@ebensh Nope, as `_` (aka blank identifier) is a workaround for declared but not used params/variables/imports. If I use `_param` it will still be reported as unused as underscore prefix has no special meaning in Go but only in some conventions:
See: 
- https://go.dev/doc/effective_go#blank
- https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

Example:
```
central/activecomponent/datastore/index/indexer_impl.go:30:29: unused-parameter: parameter '_unused' seems to be unused, consider removing or renaming it as _ (revive)
func (b *indexerImpl) Count(_unused context.Context, q *v1.Query, opts ...blevesearch.SearchOption) (int, error) {
```

PR #5386, opened 2023-03-24
Add ImageCVE page summary cards

dvail:
## Description

This adds summary cards to the ImageCVE single page.

## Checklist

## Testing Performed

Load a CVE page:

image

Apply search filters and observe that the total number of images in the ""Environmental impact" card as well as the labels in the "By severity" card are updated. *Note that the actual values here are _wrong_, since the BE query support is not fully merged yet*
image

Query error state:
image

Loading state:
image

Small screen views:

### DIFF
--- a/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageCvePage.tsx
+++ b/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageCvePage.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { gql, useQuery } from '@apollo/client';
 import {
+    Alert,
     Breadcrumb,
     BreadcrumbItem,
     Bullseye,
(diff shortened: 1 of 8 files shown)


ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageCvePage.tsx
```diff
@@ -63,8 +102,36 @@ function ImageCvePage() {
...
             <PageSection className="pf-u-display-flex pf-u-flex-direction-column pf-u-flex-grow-1">
-                <WorkloadTableToolbar />
-                <Divider />
+                <div className="pf-u-background-color-100">
+                    <div className="pf-u-px-sm">
+                        <WorkloadTableToolbar />
+                    </div>
+                    <div className="pf-u-px-lg pf-u-pb-lg">
+                        {summaryRequest.error && (
+                            <Alert
```
dvail, line 111: I used an error-style `Alert` here instead of the error empty state template, mostly because it is a smaller space and felt right. This is inconsistent with the error state in the header if the details query fails to load though...

I'll make a note to discuss this with UX at the next meeting too.

dvail:
There is actually an open question on what exactly the numbers in the "Environmental impact" card mean, so I may need to update this PR later or create a follow up ticket.

alwayshooin (approved):
the changes makes sense -- i think it would be good to bring up the error state to the UX team but otherwise looks good!

PR #5387, opened 2023-03-24
Adjust UI for Origin field in access control resources

ivan-degtiarenko:
## Description

* Show origin as simple text instead of the label in list view
*  show origin for groups + info icon with explanation for unmodifiable groups

## Checklist

## Testing Performed

Screenshot 2023-03-24 at 18 26 51

Screenshot 2023-03-24 at 18 28 59

### DIFF
--- a/ui/apps/platform/src/Containers/AccessControl/AuthProviders/RuleGroups.tsx
+++ b/ui/apps/platform/src/Containers/AccessControl/AuthProviders/RuleGroups.tsx
@@ -174,7 +174,7 @@ function RuleGroups({
                                     </FlexItem>
                                 ) : (
                                     <FlexItem>
-                                        <Tooltip content="This rule is managed declaratively and cannot be edited.">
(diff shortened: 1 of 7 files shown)


ui/apps/platform/src/Containers/AccessControl/AuthProviders/RuleGroups.tsx
```diff
@@ -155,6 +172,18 @@ function RuleGroups({
                                             <TrashIcon />
                                         </Button>
                                     </FlexItem>
+                                ) : (
+                                    <FlexItem>
+                                        <Tooltip content="This rule is managed declaratively and cannot be edited.">
```
dhaus67, line 177: nit: should we specify that this can _only_ be modified within the "declarative" workflow, i.e something like:
`This rule is managed declaratively and can only be edited declaratively`?

dhaus67 (approved):
Minor comment regarding the hint we show, otherwise LGTM.

PR #5401, opened 2023-03-27
ROX-16208: Sensor UpdateImage integration test

fredrb:
## Description

This PR provides a test that captures the issue found in the new re-sync pipeline related to `UpdateImages` not being properly handled when scanner updates image components or wasn't up when a deployment is initially processed. 

The issue is fixed by #5365, however, we didn't have any automated tests failed before. This new test will make sure the issue is fixed and also serve as regression in case the code changes.

## Checklist

## Testing Performed

Just the added tests

### DIFF
--- a/sensor/tests/resource/imagescan/imagescan_test.go
+++ b/sensor/tests/resource/imagescan/imagescan_test.go
@@ -144,7 +144,7 @@ func checkIfAlertsHaveViolation(result *central.AlertResults, name string) bool 
 	if len(alerts) == 0 {
 		return false
 	}
-	for _, alert := range result.GetAlerts() {
+	for _, alert := range alerts {
 		if alert.GetPolicy().GetName() == name {
 			return true
 		}
--- a/sensor/tests/resource/imagescan/imagescan_test.go
+++ b/sensor/tests/resource/imagescan/imagescan_test.go
@@ -1,0 +1,153 @@
+package imagescan
+
+import (
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/generated/internalapi/central"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/uuid"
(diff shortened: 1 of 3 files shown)


sensor/tests/resource/imagescan/imagescan_test.go
```diff
@@ -0,0 +1,153 @@
...
+						},
+					},
+				},
+			})
+
+			tc.GetFakeCentral().StubMessage(&central.MsgToSensor{
+				Msg: &central.MsgToSensor_ReprocessDeployments{
+					ReprocessDeployments: &central.ReprocessDeployments{},
+				},
+			})
```
lvalerom, line 125: If we decide to trigger affected deployments on `UpdatedImage`, then the `ReprocessDeployments` shouldn't be needed here. I think, unless there's a guarantee that after `UpdatedImage` we are going to receive `ReprocessDeployments`, we should trigger affected deployments on `UpdatedImage`, thus I think the tests should test this without sending a `ReprocessDeployments`.

lvalerom (changes requested):
Looks good. I'm just blocking because I think we should agree on sending or not the `ReprocessDeployments` message before merging.

fredrb:
> Looks good. I'm just blocking because I think we should agree on sending or not the ReprocessDeployments message before merging.

I don't agree with this approach. The current behavior (even before your PR) is that Central _has_ to send `ReprocessDeployment` for the deployments to be reprocessed -- even with re-sync enabled. So this test is just validating a constraint we have in the system. I prefer to have this test running in CI and PRs so we can also make sure it passes with your change once is merged.

If we make the change that `UpdateImage` is sufficient to reprocess deployments, we can update the test to reflect that.

sensor/tests/resource/imagescan/imagescan_test.go
fredrb: See comment below

lvalerom (approved):

PR #5408, opened 2023-03-27
Fix errors due to timeouts through net.Error interface

connorgorman:
## Description

The net.Error interface is better because there are a wide variety of errors we may hit. Temporary() comments say not to use it

## Checklist

## Testing Performed

Will run a proxy to make sure this happens and that the error is returned

### DIFF
--- a/central/sensor/service/connection/worker_queue.go
+++ b/central/sensor/service/connection/worker_queue.go
@@ -66,7 +66,7 @@ func (w *workerQueue) runWorker(ctx context.Context, idx int, stopSig *concurren
 	queue := w.queues[idx]
 	for msg := queue.pullBlocking(stopSig); msg != nil; msg = queue.pullBlocking(stopSig) {
 		err := handler(ctx, msg)
-		if err != nil {
+		if err != nil && err != context.Canceled {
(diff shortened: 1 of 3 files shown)


central/sensor/service/connection/worker_queue.go
```diff
@@ -66,7 +66,7 @@ func (w *workerQueue) runWorker(ctx context.Context, idx int, stopSig *concurren
 	queue := w.queues[idx]
 	for msg := queue.pullBlocking(stopSig); msg != nil; msg = queue.pullBlocking(stopSig) {
 		err := handler(ctx, msg)
-		if err != nil {
+		if err != nil && err != context.Canceled {
```
connorgorman, line 69: context.Canceled can occur when Sensor disconnects

dashrews78: is it worth logging that case?

connorgorman: No because they can stop for any variety of reasons

connorgorman: It's a lot of logspam

dashrews78 (approved):

PR #5415, opened 2023-03-28
ROX-14150: Add create command for declarative configs

dhaus67:
## Description

This PR is the first PR in a series of PRs to add roxctl commands to support users with the usage of declarative configurations.

The first command introduced is the `create` command.
The use-case for the `create` command is to offer users with the possibility to generate the YAML format of the declarative configuration via a CLI, instead of manually constructing it themselves (similar to how one can generate a deployment YAML via `kubectl create deployment --replicas=1 --image 
nginx:latest --dry-run-client -o yaml`).

The `create` command exposes subcommands for each resource supported in declarative configuration (`role, auth provider, access scope, permission set`) and each command will a) construct the declarative configuration struct from the flag values b) attempt to transform the struct, as a means of validation.
If everything passed successfully, the YAML will be returned.

There are some caveats within the PR:
- the custom `MarshalYAML` methods added to the `Access` and `Operator` types were fixed, as they had the wrong signature and were never called. Included tests to cover that these work going forward.
- for the `create access-scope` command, a few workarounds had to be taken. This is mainly due to the complex structure of the object, which in turn is quite hard to provide as a CLI:
    - Two custom flag types were added, `included-objects` and `requirements`. The custom flag types allow to add a key with multiple values / three key value pairs within a single flag.

Other alternatives considered within this PR:
While this "manual" work seems OK for now within this PR and for the amount of resources we have, it might be worthwhile generating the flags at the very least going forward. There are a couple of libraries available (such as [go-flags](https://github.com/jessevdk/go-flags)), but either they do not support integration with cobra's `pflag.FlagSet` or are mostly unmaintained, hence I decided against using such a solution.
In the future we might either introduce this as our own / fork one of the libraries, or integrate one if we find a suitable one (the desire for these commands may also be less once we move to CRs).

## Checklist

## Testing Performed

- see added tests.

### DIFF
--- a/pkg/declarativeconfig/access_scope_test.go
+++ b/pkg/declarativeconfig/access_scope_test.go
@@ -9,31 +9,41 @@ import (
 )
 
 func TestAccessScopeYAMLTransformation(t *testing.T) {
-	data := []byte(`
-name: test-name
+	data := []byte(`name: test-name
 description: test-description
 rules:
-  included:
-    - cluster: clusterA
-      namespaces:
-      - namespaceA1
-      - namespaceA2
-    - cluster: clusterB
-  clusterLabelSelectors:
-    - requirements:
-      - key: a
-        operator: IN
-        values: [a, b, c]
-      - key: b
-        operator: NOT_IN
-        values: [b, d]
-    - requirements:
-      - key: c
-        operator: IN
-        values: [d, e, f]
-      - key: d
-        operator: NOT_IN
-        values: [x, y, z]
+    included:
+        - cluster: clusterA
+          namespaces:
+            - namespaceA1
+            - namespaceA2
+        - cluster: clusterB
+    clusterLabelSelectors:
+        - requirements:
+            - key: a
+              operator: IN
+              values:
+                - a
+                - b
+                - c
+            - key: b
+              operator: NOT_IN
+              values:
+                - b
+                - d
+        - requirements:
+            - key: c
+              operator: IN
+              values:
+                - d
+                - e
+                - f
+            - key: d
+              operator: NOT_IN
+              values:
+                - x
+                - "y"
+                - z
 `)
 	as := AccessScope{}
 
@@ -70,4 +80,8 @@ rules:
 	assert.Equal(t, as.Rules.ClusterLabelSelectors[1].Requirements[1].Key, "d")
 	assert.Equal(t, as.Rules.ClusterLabelSelectors[1].Requirements[1].Values, []string{"x", "y", "z"})
 	assert.Len(t, as.Rules.NamespaceLabelSelectors, 0)
+
+	bytes, err := yaml.Marshal(&as)
+	assert.NoError(t, err)
+	assert.Equal(t, string(data), string(bytes))
 }
--- a/roxctl/declarativeconfig/create/access_scope.go
+++ b/roxctl/declarativeconfig/create/access_scope.go
@@ -28,8 +28,8 @@ Each tuple represents a requirement, which will be used to construct the label s
 You may specify this flag multiple times, to create a conjunction of requirements which should apply for a label selector
 to match.
 
-Example of a label selector requiring values: --cluster-label-selector "key=kubernetes.io/hostname;operator=IN;values=nodeA,nodeB"
-Example of a label selector not requiring values: --cluster-label-selector "key=custom-label;operator=EXISTS"
+Example of a label selector requiring values: --%s "key=kubernetes.io/hostname;operator=IN;values=nodeA,nodeB"
+Example of a label selector not requiring values: --%s "key=custom-label;operator=EXISTS"
 
 NOTE: The created access scope will only contain a single label selector, where each specified requirement
 will be in conjunction. If you desire to create multiple label selectors, you have to adjust the YAML output manually.
@@ -66,10 +66,10 @@ In case only a subset of namespace should be included, specify --included cluste
 	// if they wish to do so.
 
 	cmd.Flags().Var(&requirementFlag{requirements: &accessScopeCmd.clusterRequirements}, "cluster-label-selector",
-		labelSelectorUsage)
+		fmt.Sprintf(labelSelectorUsage, "cluster-label-selector", "cluster-label-selector"))
 
 	cmd.Flags().Var(&requirementFlag{requirements: &accessScopeCmd.namespaceRequirements}, "namespace-label-selector",
-		labelSelectorUsage)
+		fmt.Sprintf(labelSelectorUsage, "namespace-label-selector", "namespace-label-selector"))
 
 	utils.Must(cmd.MarkFlagRequired("name"))
 
--- a/roxctl/declarativeconfig/create/access_scope.go
+++ b/roxctl/declarativeconfig/create/access_scope.go
@@ -153,10 +153,10 @@ func (r *requirementFlag) String() string {
 	res := make([]string, 0, len(*r.requirements))
 
 	for _, requirement := range *r.requirements {
-		requirementString := fmt.Sprintf("key=%s;operator=%s", requirement.Key,
+		requirementString := fmt.Sprintf("key=%q;operator=%q", requirement.Key,
 			storage.SetBasedLabelSelector_Operator(requirement.Operator))
 		if len(requirement.Values) != 0 {
-			requirementString = fmt.Sprintf("%s;values=%s",
+			requirementString = fmt.Sprintf("%s;values=%q",
 				requirementString, strings.Join(requirement.Values, ","))
 		}
 		res = append(res, requirementString)
--- a/roxctl/declarativeconfig/create/access_scope.go
+++ b/roxctl/declarativeconfig/create/access_scope.go
@@ -12,6 +12,7 @@ import (
 	"github.com/stackrox/rox/pkg/declarativeconfig"
 	"github.com/stackrox/rox/pkg/declarativeconfig/transform"
 	"github.com/stackrox/rox/pkg/maputil"
+	"github.com/stackrox/rox/pkg/utils"
 	"github.com/stackrox/rox/roxctl/common/environment"
 	"gopkg.in/yaml.v3"
 )
@@ -68,6 +69,8 @@ In case only a subset of namespace should be included, specify --included cluste
 
 	cmd.Flags().Var(&requirementFlag{requirements: &accessScopeCmd.namespaceRequirements}, "namespace-label-selector",
 		labelSelectorUsage)
+
+	utils.Must(cmd.MarkFlagRequired("name"))
 
 	return cmd
 }
@@ -187,7 +190,7 @@ func retrieveRequirement(s string) (*declarativeconfig.Requirement, error) {
 	requirement := &declarativeconfig.Requirement{}
 	for _, kvPair := range kvPairs {
 		if strings.Count(kvPair, "=") != 1 {
-			return nil, fmt.Errorf("%s must specify key=value", s)
+			return nil, fmt.Errorf("%s must specify key=value", kvPair)
 		}
 
 		kv := strings.Split(kvPair, "=")
--- a/roxctl/declarativeconfig/create/access_scope.go
+++ b/roxctl/declarativeconfig/create/access_scope.go
@@ -39,8 +39,9 @@ func accessScopeCommand(cliEnvironment environment.Environment) *cobra.Command {
 	accessScopeCmd := accessScopeCmd{accessScope: &declarativeconfig.AccessScope{}, env: cliEnvironment}
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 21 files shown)


dhaus67:
* master
  * **PR #5415** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5415" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5415?utm_source=stack-comment).

roxctl/declarativeconfig/create/access_scope.go
```diff
@@ -0,0 +1,216 @@
...
+key, value, operator. The key value pairs are expected to be separated with ;.
+Each tuple represents a requirement, which will be used to construct the label selector.
+You may specify this flag multiple times, to create a conjunction of requirements which should apply for a label selector
+to match.
+
+Example of a label selector requiring values: --cluster-label-selector "key=kubernetes.io/hostname;operator=IN;values=nodeA,nodeB"
+Example of a label selector not requiring values: --cluster-label-selector "key=custom-label;operator=EXISTS"
+
+NOTE: The created access scope will only contain a single label selector, where each specified requirement
+will be in conjunction. If you desire to create multiple label selectors, you have to adjust the YAML output manually.
```
janisz, line 35: chat gpt simplify:
```
The flag "cluster-label-selector" is used to create a label selector with one or more requirements, each represented as a tuple with three key-value pairs: "key", "value", and "operator". The tuples are separated by a semicolon (;) and can be specified multiple times to create a conjunction of requirements for the label selector to match.

For example, to create a label selector that requires values, you can use the following format:
--cluster-label-selector "key=kubernetes.io/hostname;operator=IN;values=nodeA,nodeB"

To create a label selector that does not require values, you can use the following format:
--cluster-label-selector "key=custom-label;operator=EXISTS"

Note that the resulting access scope will contain only one label selector with all the specified requirements in conjunction. If you need multiple label selectors, you will need to modify the YAML output manually.
```

roxctl/declarativeconfig/create/access_scope.go
```diff
@@ -0,0 +1,216 @@
...
+// Implementation of pflag.Value to support complex object declarativeconfig.Requirement.
+type requirementFlag struct {
+	requirements *[]declarativeconfig.Requirement
+}
+
+func (r *requirementFlag) String() string {
+	res := make([]string, 0, len(*r.requirements))
+
+	for _, requirement := range *r.requirements {
+		requirementString := fmt.Sprintf("key=%s;operator=%s", requirement.Key,
```
janisz, line 155: should we use `%q` instead of `%s`

roxctl/declarativeconfig/create/access_scope.go
```diff
@@ -0,0 +1,216 @@
...
+
+type accessScopeCmd struct {
+	accessScope *declarativeconfig.AccessScope
+	env         environment.Environment
+
+	clusterRequirements   []declarativeconfig.Requirement
+	namespaceRequirements []declarativeconfig.Requirement
+}
+
+func (a *accessScopeCmd) Validate() error {
```
janisz, line 86: Should we validate if used names are valid? Eg. if labels used in label selectors match k8s requirements?

roxctl/declarativeconfig/create/role_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"testing"
+
+	"github.com/stackrox/rox/roxctl/common/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateRoleCommand_Failures(t *testing.T) {
+	cases := map[string]struct {
```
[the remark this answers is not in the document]
dhaus67: Guess that's not covered by `MarkFlagRequired`.

I'll add the validation to the transformer though, think that's the most appropriate place.

roxctl/declarativeconfig/create/permission_set.go
```diff
@@ -0,0 +1,83 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := permSetCmd.Validate(); err != nil {
+				return errors.Wrap(err, "validating the permission set")
+			}
+			return permSetCmd.PrintYAML()
+		},
+		Short: "Create a declarative configuration for a permission set",
+	}
+
+	cmd.Flags().StringVar(&permSetCmd.permissionSet.Name, "name", "", "name of the permission set")
```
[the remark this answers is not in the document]
dhaus67: Hmm, not sure I'll agree on the usability of it.

Realistically, you won't have the same resource access for _all_ resource, and with the current options you wouldn't be able to modify this.

We could offer just a "dry-run" of creating a permission set with all resources and set `NO_ACCESS` for all and have people modify the YAML file afterwards.

However, let's skip it for now, people can always extend the YAML if they wish to and it doesn't feel like a straightforward UX improvement within the CLI to me.

pkg/declarativeconfig/access_scope_test.go
```diff
@@ -9,31 +9,41 @@ import (
...
+              values:
+                - d
+                - e
+                - f
+            - key: d
+              operator: NOT_IN
+              values:
+                - x
+                - "y"
+                - z
```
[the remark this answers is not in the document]
dhaus67: Hmm, don't really see a reason for it? Its essentially interchangeable to have it either in this format or in the yaml list - the decoder will always provide the YAML list format, so let's keep it like this since its easier to compare.

roxctl/declarativeconfig/create/auth_provider.go
```diff
@@ -0,0 +1,294 @@
...
+
+	t := transform.New()
+	_, err = t.Transform(a.authProvider)
+	return errors.Wrap(err, "validating auth provider")
+}
+
+func (a *authProviderCmd) validateGroups() ([]declarativeconfig.Group, error) {
+	expectedGroups := len(a.groupsKeys)
+
+	if len(a.groupsKeys) != expectedGroups || len(a.groupsValues) != expectedGroups || len(a.groupsRoles) != expectedGroups {
```
ivan-degtiarenko, line 262:
```suggestion
	if len(a.groupsValues) != expectedGroups || len(a.groupsRoles) != expectedGroups {
```

roxctl/declarativeconfig/create/permission_set.go
```diff
@@ -0,0 +1,83 @@
...
+	cmd.Flags().StringVar(&permSetCmd.permissionSet.Description, "description", "",
+		"description of the permission set")
+	cmd.Flags().StringToStringVar(&permSetCmd.resourceWithAccess, "resource-with-access", map[string]string{},
+		`list of resource with the respective access, e.g. --resource-with-access Access=READ_ACCESS,Admin=READ_WRITE_ACCESS
+Note: Capitalization matters!`)
```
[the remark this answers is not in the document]
ivan-degtiarenko: Now we also can delete this note, as capitalization no longer matters

roxctl/declarativeconfig/create/permission_set.go
```diff
@@ -0,0 +1,83 @@
...
+			return permSetCmd.PrintYAML()
+		},
+		Short: "Create a declarative configuration for a permission set",
+	}
+
+	cmd.Flags().StringVar(&permSetCmd.permissionSet.Name, "name", "", "name of the permission set")
+	cmd.Flags().StringVar(&permSetCmd.permissionSet.Description, "description", "",
+		"description of the permission set")
+	cmd.Flags().StringToStringVar(&permSetCmd.resourceWithAccess, "resource-with-access", map[string]string{},
+		`list of resources with the respective access, e.g. --resource-with-access Access=READ_ACCESS,Admin=READ_WRITE_ACCESS
```
ivan-degtiarenko, line 37:
```suggestion
		`list of resources with the respective access, e.g. --resource-with-access Access=READ_ACCESS,Administration=READ_WRITE_ACCESS
```

There is no such resource as `Admin`. Let's have only valid resources as examples

roxctl/declarativeconfig/create/auth_provider_test.go
```diff
@@ -0,0 +1,447 @@
...
+q/I2+0j6dAkOGcK/68z7qQXByeGri3n28a1Kn6o=
+-----END CERTIFICATE-----
+`))
+	assert.NoError(t, err)
+
+	args := []string{
+		"saml",
+		"--name=some-name",
+		"--ui-endpoint=localhost:8000",
+		"--groups-key=email",
```
ivan-degtiarenko, line 213: Let's add a test where there is more than one group

roxctl/declarativeconfig/create/auth_provider_test.go
```diff
@@ -0,0 +1,447 @@
...
+		"--groups-value=example@example.com",
+		"--groups-role=Admin",
+		"--minimum-access-role=Analyst",
+		"--extra-ui-endpoints=localhost:9090",
```
ivan-degtiarenko, line 217: And also let's change somewhere `extra-ui-endpoints` to have several values

roxctl/declarativeconfig/create/auth_provider.go
```diff
@@ -0,0 +1,294 @@
...
+		"Client ID of the OIDC client")
+	cmd.Flags().StringVar(&a.oidcConfig.ClientSecret, "client-secret", "",
+		"Client Secret of the OIDC client")
+	cmd.Flags().BoolVar(&a.oidcConfig.DisableOfflineAccessScope, "disable-offline-access", false,
+		"Disable requesting the scope offline_access from the OIDC identity provider. This should only be set "+
+			"if there are any limitations from your OIDC IdP about the number of sessions with the offline_access scope")
+
+	cmd.Flags().StringToStringVar(&a.claimMapping, "claim-mappings", map[string]string{},
+		`list of non-standard claims from the IdP token that should be available within auth provider rules, e.g.
+--claim-mappings "my_claim_on_the_idp_token=claim_name_on_the_rox_token"`)
```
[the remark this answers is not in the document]
dhaus67: Isn't it quite obvious that this is only applicable for OIDC auth providers if its only a flag for the OIDC auth provider? 😅 
It's not exposed to any other auth provider type, this should be explicit enough.

roxctl/declarativeconfig/create/permission_set.go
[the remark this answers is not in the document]
dhaus67: Capitalization still matters for the resource names, right?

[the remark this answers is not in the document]
ivan-degtiarenko: WDYT about addressing it using `cases.Title` function?

roxctl/declarativeconfig/create/auth_provider.go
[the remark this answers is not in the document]
ivan-degtiarenko: Makes sense :)

roxctl/declarativeconfig/create/auth_provider.go
```diff
@@ -0,0 +1,294 @@
...
+func (a *authProviderCmd) oidcCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "oidc",
+		Args:  cobra.NoArgs,
+		RunE:  a.RunE(),
+		Short: "Create a declarative configuration for an OIDC auth provider",
+	}
+	a.oidcConfig = &declarativeconfig.OIDCConfig{}
+
+	cmd.Flags().StringVar(&a.oidcConfig.Issuer, "issuer", "",
```
ivan-degtiarenko, line 107: Can we require `Issuer`, `CallbackMode` and `ClientID`? I think we always need them to create proper OIDC auth provider

roxctl/declarativeconfig/create/auth_provider.go
```diff
@@ -0,0 +1,294 @@
...
+func (a *authProviderCmd) samlCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "saml",
+		Args:  cobra.NoArgs,
+		RunE:  a.RunE(),
+		Short: "Create a declarative configuration for a SAML auth provider",
+	}
+	a.samlConfig = &declarativeconfig.SAMLConfig{}
+
+	cmd.Flags().StringVar(&a.samlConfig.SpIssuer, "sp-issuer", "", "service provider "+
```
ivan-degtiarenko, line 135: Can we mark `sp-issuer` as required? Also `metadata-url` and other 4 parameters as mutually exclusive, and other 4 parameters as required together?

ivan-degtiarenko (approved):
Some comments regarding marking flags as required + capitalization, otherwise LGTM!

roxctl/declarativeconfig/create/auth_provider.go
dhaus67: I think we can leave `CallbackMode` out since it will default to `auto`.

PR #5416, opened 2023-03-28
ROX-14947: Use deployment search service instead of graphql to fill NetworkGraph 2.0 deployment dropdown

rhybrillou:
## Description

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


ui/apps/platform/src/services/DeploymentsService.ts
```diff
@@ -13,8 +13,8 @@ import { getRequestQueryStringForSearchFilter } from 'utils/searchUtils';
 import { CancellableRequest, makeCancellableAxiosRequest } from './cancellationUtils';
 import axios from './instance';
 
-const deploymentsUrl = '/v1/deploymentswithprocessinfo';
-const deploymentByIdUrl = '/v1/deployments';
+const deploymentUrl = '/v1/deployments';
```
pedrottimark, line 16: Distinguishing variable names for endpoints in this way is very helpful.

For the names to be parallel, what do you think of plural `deploymentsUrl`

ui/apps/platform/src/services/DeploymentsService.ts
```diff
@@ -52,10 +44,46 @@ export function fetchDeployments(
...
+ * Changes from the 'fetchDeploymentsLegacy' function:
+ * - uses the new `SearchFilter` type instead of `RestSearchOption`
+ * - Does not fetch process information linked to the deployment
+ */
+export function listDeployments(
+    searchFilter: SearchFilter,
+    sortOption: Record<string, string>,
+    page: number,
+    pageSize: number
+): Promise<ListDeployment[]> {
```
pedrottimark, line 62: As a follow up after achieving the access control goal, we might test whether `CancellableRequest` is needed to prevent memory leaks if someone selects namespaces, and then immediately clicks a link in left navigation to leave NetworkGraph page.

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -33,7 +44,78 @@ const DEPLOYMENTS_FOR_NAMESPACE_QUERY = gql`
...
+            listDeployments(searchQuery, sortOption, 0, 0)
+                .then((response) => {
+                    const namespacesWithDeployments: NamespaceWithDeployments[] = [];
+                    const deploymentsByNamespace = new Map<string, Deployment[]>();
+                    response.forEach((listDeployment) => {
+                        const deployment: Deployment = {
+                            id: listDeployment.id,
+                            name: listDeployment.name,
+                        };
+                        const { namespace } = listDeployment;
```
pedrottimark, line 78: Here is an alternative that with object destructuring for all 3 variables in argument of `forEach` callback and slight rewriting:

```js
response.forEach({id, name, namespace}) => {
    const deployment = {id, name};
    if (deploymentsByNamespace.has(namespace)) {
        deploymentsByNamespace[namespace].push(deployment);
    } else {
        deploymentsByNamespace.set(namespace, [deployment]);
    }
});
```

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -33,7 +44,78 @@ const DEPLOYMENTS_FOR_NAMESPACE_QUERY = gql`
...
+                        if (!deploymentsByNamespace.has(namespace)) {
+                            deploymentsByNamespace.set(namespace, []);
+                        }
+                        deploymentsByNamespace[namespace].push(deployment);
+                    });
+                    deploymentsByNamespace.forEach((deployments, namespace) => {
+                        const namespaceId = namesToIdMapping.get(namespace);
```
pedrottimark, line 85: What do you think about `namespaceName` parallel with `namespaceId`

ui/apps/platform/src/services/DeploymentsService.ts
```diff
@@ -92,7 +120,7 @@ export function fetchDeploymentsLegacy(
     }
     const params = queryString.stringify(queryObject, { arrayFormat: 'repeat', allowDots: true });
     return axios
-        .get<{ deployments: ListDeploymentWithProcessInfo[] }>(`${deploymentsUrl}?${params}`)
+        .get<{ deployments: ListDeploymentWithProcessInfo[] }>(`${deploymentsWithProcessUrl}?${params}`)
```
pedrottimark, line 123: It looks like style checks reports a problem here with need for wrapping because of line length.

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
rhybrillou: This piece was refactored a bit further. It seems the square bracket accessor breaks the flow.
I changed the conditions to always rely on `get` and `set` functions to interact with the map.

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -1,68 +1,88 @@
...
-        deploymentsByNamespace: availableDeployments,
-    };
+                    setDeploymentResponse({
+                        loading: false,
+                        error: errorMessage,
+                        deploymentsByNamespace: [],
+                    });
+                });
+        }
+    }, [selectedNamespaceIdsAsString]);
```
rhybrillou, line 84: React doc ( https://react.dev/reference/react/useEffect ) states: 
```
React will compare each dependency with its previous value using the [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison.
```
Comparing strings is then a deep compare while array comparison is not. To work around array comparison triggering infinite effect loop, the array is serialized for the effect dependency comparison to compare string contents, then de-serialized for the effect to use the actual array content.

ivan-degtiarenko: Why not use `useDeepCompareEffect` here?

ui/apps/platform/src/services/DeploymentsService.ts
```diff
@@ -52,12 +44,51 @@ export function fetchDeployments(
...
+
+/**
+ * Fetches list of registered deployments.
+ *
+ * Changes from the 'legacy' version of this same function:
+ * - returns a 'cancel' function to abort the request
+ * - uses the new `SearchFilter` type instead of `RestSearchOption`
+ * - Does not implicitly read the value of "shouldHideOrchestratorComponents"
+ */
+export function fetchDeployments(
```
ivan-degtiarenko, line 77: Perhaps this should be named as `fetchDeploymentsWithProcessInfo` - otherwise for user of `DeploymentService` it can be confusing as what to use - `fetchDeployments` or `listDeployments`

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -1,68 +1,88 @@
...
+            });
+            const searchQuery: Record<string, string[]> = {
+                'Namespace ID': namespaceIds,
+            };
+            const sortOption = { field: 'Deployment', reversed: 'false' };
+            listDeployments(searchQuery, sortOption, 0, 0)
+                .then((response) => {
+                    const namespacesWithDeployments: NamespaceWithDeployments[] = [];
+                    const deploymentsByNamespace = new Map<string, Deployment[]>();
+                    response.forEach(({ id, name, namespace }) => {
```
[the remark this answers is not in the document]
rhybrillou: The underscore code would look like this.
```
                    const deploymentsIndexesByNamespace = _.groupBy(response, 'namespace');
                    const namespaces = _.keys(deploymentsIndexesByNamespace);
                    _.each(namespaces, (ns) => {
                        const namespaceDeployments: Deployment[] = [];
                        _.each(deploymentsIndexesByNamespace[ns], ({ id, name }) => {
                            namespaceDeployments.push({ id, name });
                        });
                        const namespaceWithDeployments: NamespaceWithDeployments = {
                            metadata: {
                                name: ns,
                            },
                            deployments: namespaceDeployments,
                        };
                        namespacesWithDeployments.push(namespaceWithDeployments);
                    });
```
I get the following error with the above code:
`TS2686: '_' refers to a UMD global, but the current file is a module. Consider adding an import instead`

ui/apps/platform/src/Containers/Policies/Wizard/Step4/PolicyScopeForm.tsx
```diff
@@ -87,7 +87,7 @@ function PolicyScopeForm() {
         // Move request to exclusion card to add restSearch for cluster or namespace if specified in exclusion scope.
         // Search element to support creatable deployment names.
         const restSort = { field: 'Deployment' }; // ascending by name
-        fetchDeployments([], restSort, 0, 0)
+        fetchDeploymentsWithProcessInfo([], restSort, 0, 0)
```
pedrottimark, line 90: As a follow up, this might become `listDeployments` function, because only `id` and `name` is needed:

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/Policies/Wizard/Step4/PolicyScopeCard.tsx#L150-L157

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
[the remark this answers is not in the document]
pedrottimark: Instead of underscore, UI has lodash as a dependency. If you decide to use its `groupBy` method:

https://lodash.com/docs/4.17.15#groupBy

then here is the import statement:

```js
import groupBy from 'lodash/groupBy';
```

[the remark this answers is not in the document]
pedrottimark: Whether or not you use lodash `groupBy` method, what do you think about factoring out as a function (in this file, but outside of the hook) which returns `NamespaceWithDeployments[]` for `deployments: ListDeployment` argument.

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -1,68 +1,87 @@
...
-        selectedNamespaceIds.length > 0
-            ? { variables: { query: queryService.objectToWhereClause(searchClause) } }
-            : { skip: true };
+    const [deploymentResponse, setDeploymentResponse] = useState<ListDeploymentResponse>({
+        loading: false,
+        error: '',
+        deploymentsByNamespace: [],
+    });
+    useDeepCompareEffect(() => {
+        if (selectedNamespaceIds.length > 0) {
```
pedrottimark, line 31: 1. Although I understand why not to make a request if the id array is empty, it is not clear to me whether or not whether there needs to be a change to value of `deploymentsByNamespace` array. My intuitive impression is that it should become `[]` like the initial value, but I do not see that in the GraphQL nor REST version.
2. If that observation is correct, it suggests that the calling code is written not to depends on the hook clearing the deployments if the user clears the namespace selections.
3. If that hypothesis is correct, then here are 2 questions:
    * Thinking of the hook as reusable for other situations, what should its behavioral contract be for a change from some namespaces selected to no namespaces selected?
    * Would a change in behavior break the existing use of the hook?

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
```diff
@@ -1,68 +1,101 @@
...
-        deploymentsByNamespace: availableDeployments,
-    };
+                    setDeploymentResponse({
+                        loading: false,
+                        error: errorMessage,
+                        deploymentsByNamespace: [],
+                    });
+                });
+        }
+    }, [selectedNamespaceIds]);
```
rhybrillou, line 97: The effect may have to adjust for state cleanup on unmount as well
https://github.com/stackrox/stackrox/commit/764de7a19f46bfbd94764f9859a4c3f7dfc136e2#diff-6e2196a41b13bd09dbca4f6ac768c2685414ea0e02e0226e9ba0f4e90ce76e17

ui/apps/platform/src/Containers/Policies/Wizard/Step4/PolicyScopeForm.tsx
pedrottimark: Thank you for reminding me about that. Policies does not use that option.

ui/apps/platform/src/hooks/useFetchNamespaceDeployments.ts
pedrottimark: That sounds like a good idea. Dave Vail is a good source of advice and review for state cleanup.

pedrottimark (approved):
Thank you for your patience and careful follow through.

Approving to unblock you, in case state cleanup will be a separate contribution.

ivan-degtiarenko (approved):

PR #5422, opened 2023-03-28
ROX-16171: Create node-inventory container only on Openshift 4

vikin91:
## Description

The node-inventory container within the collector pod shall be created only on Openshift clusters in version 4. RHCOS node scanning feature is only supported on such clusters, so it does not make sense to create it on all other types of clusters. This exclusion allows us to save resources.

## Checklist

### N/A

## Testing Performed

- [x] Running `roxctl helm output secured-cluster-services` followed by `helm template ... --set env.openshift=4` and analyzing the `collector.yaml` manifest manually - expecting to see the node-inventory container
- [x] Running `roxctl helm output secured-cluster-services` followed by `helm template` (without `--set env.openshift=4`) and analyzing the `collector.yaml` manifest manually - expecting not to see the node-inventory container
- [x] Running `roxctl sensor generate openshift --insecure-skip-tls-verify --name cluster2 --openshift-version 4` and analyzing the `collector.yaml` manifest manually
- [x] Running `roxctl sensor generate openshift --insecure-skip-tls-verify --name cluster3 --openshift-version 3` and analyzing the `collector.yaml` manifest manually
- [x] CI

### DIFF
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -452,7 +452,7 @@ jobs:
         run: |
           ./oc -n stackrox set image ds/collector compliance="quay.io/rhacs-eng/main:$TAG"
           ./oc -n stackrox set image ds/collector collector="quay.io/rhacs-eng/collector:$TAG"
-          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG"
+          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG" || echo "node-inventory shall be deployed only on Openshift 4 clusters"
           ./oc -n stackrox set image deploy/admission-control admission-control="quay.io/rhacs-eng/main:$TAG"
 
   notify-os4-cluster:
--- a/.github/workflows/cut-rc.yml
+++ b/.github/workflows/cut-rc.yml
@@ -452,7 +452,7 @@ jobs:
         run: |
           ./oc -n stackrox set image ds/collector compliance="quay.io/rhacs-eng/main:$TAG"
           ./oc -n stackrox set image ds/collector collector="quay.io/rhacs-eng/collector:$TAG"
-          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG" || echo "node-inventory shall be deployed only on Openshift 4 clusters"
+          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG"
           ./oc -n stackrox set image deploy/admission-control admission-control="quay.io/rhacs-eng/main:$TAG"
 
   notify-os4-cluster:
--- a/central/clusters/zip/render_test.go
+++ b/central/clusters/zip/render_test.go
@@ -68,11 +68,15 @@ func doTestRenderOpenshif(t *testing.T, clusterType storage.ClusterType) {
 	}
 	assertScannerRemote := func(obj runtime.Object) {
 		ds := obj.(*v1.DaemonSet)
-		assert.Len(t, ds.Spec.Template.Spec.Containers, 3)
-		mainImage := ds.Spec.Template.Spec.Containers[1].Image
-		nodeInvCont := ds.Spec.Template.Spec.Containers[2]
-		expectedScannerParts := strings.Split(strings.ReplaceAll(mainImage, "/main:", "/scanner-slim:"), ":")
-		assert.Truef(t, strings.HasPrefix(nodeInvCont.Image, expectedScannerParts[0]), "scanner-slim image (%q) should be from the same registry as main (%q)", nodeInvCont.Image, mainImage)
+		if clusterType == storage.ClusterType_OPENSHIFT4_CLUSTER {
+			assert.Len(t, ds.Spec.Template.Spec.Containers, 3)
+			mainImage := ds.Spec.Template.Spec.Containers[1].Image
+			nodeInvCont := ds.Spec.Template.Spec.Containers[2]
+			expectedScannerParts := strings.Split(strings.ReplaceAll(mainImage, "/main:", "/scanner-slim:"), ":")
+			assert.Truef(t, strings.HasPrefix(nodeInvCont.Image, expectedScannerParts[0]), "scanner-slim image (%q) should be from the same registry as main (%q)", nodeInvCont.Image, mainImage)
+		} else {
+			assert.Len(t, ds.Spec.Template.Spec.Containers, 2)
+		}
 	}
 
 	cases := map[string]func(object runtime.Object){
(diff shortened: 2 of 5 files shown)


vikin91:
The upgrade tests failed for a reason related to this PR `error: unable to find container named "node-inventory"`. I will give it a look

.github/workflows/cut-rc.yml
```diff
@@ -452,7 +452,7 @@ jobs:
         run: |
           ./oc -n stackrox set image ds/collector compliance="quay.io/rhacs-eng/main:$TAG"
           ./oc -n stackrox set image ds/collector collector="quay.io/rhacs-eng/collector:$TAG"
-          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG"
+          ./oc -n stackrox set image ds/collector node-inventory="quay.io/rhacs-eng/scanner-slim:$TAG" || echo "node-inventory shall be deployed only on Openshift 4 clusters"
```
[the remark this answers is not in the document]
tommartensen: What is the context? The step only runs on Openshift 4 clusters anyway.

The echo would hide `oc` command failures due to other reasons as well, like timeouts, etc. Not sure that is intended.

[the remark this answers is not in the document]
vikin91: Done in [463b132](https://github.com/stackrox/stackrox/pull/5422/commits/463b132b052cba537cbd770e06cee976ec24a8a5)

Maddosaurus (approved):
LGTM!

tests/upgrade/postgres_run.sh
```diff
@@ -234,8 +234,14 @@ test_upgrade_paths() {
     kubectl -n stackrox set image deploy/sensor "*=$REGISTRY/main:$CURRENT_TAG"
     kubectl -n stackrox set image deploy/admission-control "*=$REGISTRY/main:$CURRENT_TAG"
     kubectl -n stackrox set image ds/collector "collector=$REGISTRY/collector:$(make collector-tag)" \
-        "compliance=$REGISTRY/main:$CURRENT_TAG" \
-        "node-inventory=$REGISTRY/scanner-slim:$(make scanner-tag)"
+        "compliance=$REGISTRY/main:$CURRENT_TAG"
+    # Check if collector has 2 or 3 containers - we expect to see 3 only on Openshift 4
+    if kubectl -n stackrox get ds/collector -o=jsonpath='{$.spec.template.spec.containers[2].image}'; then
```
fredrb, line 239: It seems a bit fragile to use the number of containers alone as an indicator for doing this. A better way to check this would be to match container names. You can output all container names and then check as a substring.

Here's an example with JSONPath (untested)
```bash
if [[ $(kubectl -n stackrox get ds/collector -o=jsonpath='{$.spec.template.spec.containers[*].name}') == *"node-inventory"* ]]; then
    # update image
fi
```

tests/upgrade/run.sh
```diff
@@ -207,8 +207,14 @@ test_upgrade_paths() {
     kubectl -n stackrox set image deploy/sensor "*=$REGISTRY/main:$(make --quiet tag)"
     kubectl -n stackrox set image deploy/admission-control "*=$REGISTRY/main:$(make --quiet tag)"
     kubectl -n stackrox set image ds/collector "collector=$REGISTRY/collector:$(cat COLLECTOR_VERSION)" \
-        "compliance=$REGISTRY/main:$(make --quiet tag)" \
-        "node-inventory=$REGISTRY/scanner-slim:$(cat SCANNER_VERSION)"
+        "compliance=$REGISTRY/main:$(make --quiet tag)"
+    # Check if collector has 2 or 3 containers - we expect to see 3 only on Openshift 4
+    if kubectl -n stackrox get ds/collector -o=jsonpath='{$.spec.template.spec.containers[2].image}'; then
```
fredrb, line 212: Same here

central/clusters/zip/render_test.go
```diff
@@ -68,11 +68,15 @@ func doTestRenderOpenshif(t *testing.T, clusterType storage.ClusterType) {
...
 		ds := obj.(*v1.DaemonSet)
-		assert.Len(t, ds.Spec.Template.Spec.Containers, 3)
-		mainImage := ds.Spec.Template.Spec.Containers[1].Image
-		nodeInvCont := ds.Spec.Template.Spec.Containers[2]
-		expectedScannerParts := strings.Split(strings.ReplaceAll(mainImage, "/main:", "/scanner-slim:"), ":")
-		assert.Truef(t, strings.HasPrefix(nodeInvCont.Image, expectedScannerParts[0]), "scanner-slim image (%q) should be from the same registry as main (%q)", nodeInvCont.Image, mainImage)
+		if clusterType == storage.ClusterType_OPENSHIFT4_CLUSTER {
+			assert.Len(t, ds.Spec.Template.Spec.Containers, 3)
+			mainImage := ds.Spec.Template.Spec.Containers[1].Image
+			nodeInvCont := ds.Spec.Template.Spec.Containers[2]
```
fredrb, line 74: Can we be sure that node inventory will always be at index position `2`?

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
```diff
@@ -143,6 +143,7 @@ spec:
         - mountPath: /run/secrets/stackrox.io/certs/
           name: certs
           readOnly: true
+      {{- if eq ._rox.env.openshift 4 }}
```
fredrb, line 146: We have to remember about this when OpenShift 5 comes 🙂 

Jokes aside, it would be good to have a more reliable env. Like `cluster_flavor` or something like that, which is not tied to the version. If that doesn't exist I'm fine with leaving this as is.

central/clusters/zip/render_test.go
vikin91: This is fixed in the template, so if someone would change that, this test would fail.
Let me make it a bit clearer and make an additional assertion that the 3rd container must be named `node-inventory`

image/templates/helm/stackrox-secured-cluster/templates/collector.yaml
vikin91: `{{- if eq ._rox.env.openshift 4 }}` seems to be pretty abundant in our helm templates. If OS 5 comes out, we well have some work to do ;)
I will allow myself to leave it as it is for now

central/clusters/zip/render_test.go
vikin91: Done in [56b7b0d](https://github.com/stackrox/stackrox/pull/5422/commits/56b7b0dd2efc0249d6df4cc04536c448da0d556c)

tests/upgrade/postgres_run.sh
vikin91: Addressed in [53410c9](https://github.com/stackrox/stackrox/pull/5422/commits/53410c9714483a0c7b37e418e04b21c97d25aa2f)

PR #5427, opened 2023-03-28
ROX-14150: Lint command for declarative configs

dhaus67:
## Description

The second PR in a series of PRs to add roxctl commands to support declarative configuration.

This PR introduces a lint command: `roxctl declarative-config lint`.

The user has the ability to pass a file containing declarative configuration(s) in YAML format. The YAML file will be unmarshalled, and each configuration will be attempted to be transformed. Any potential issues that may arise during unmarshalling or transforming will be reported to the user.

Caveats within the PR:
- The `UnmarshalYAML` for `Access, Operator` will now return `errox.InvalidArgs` with a list of allowed values.
- The `ConfigurationsFromRawBytes` was refactored to use a `yaml.Decoder` instead, allowing support for multi-object YAML files with the separator `---` which previously was not working.

## Checklist

## Testing Performed

- see added unit tests.


dhaus67:
* master
  * **PR #5415** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5415" target="_blank">Graphite</a> 
    * **PR #5427** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5427" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5427?utm_source=stack-comment).

pkg/declarativeconfig/configuration.go
```diff
@@ -111,3 +100,37 @@ func decodeYAMLToConfiguration(rawYAML []byte, configuration Configuration) erro
 	}
 	return nil
 }
+
+func parseToConfiguration(contents []byte) ([]Configuration, error) {
+	dec := yaml.NewDecoder(bytes.NewReader(contents))
+	var unstructuredObjs []interface{}
+	for {
```
ivan-degtiarenko, line 107: WDYT about putting some iteration limit on this `for`, like 10000 iterations?

pkg/declarativeconfig/configuration.go
```diff
@@ -111,3 +100,37 @@ func decodeYAMLToConfiguration(rawYAML []byte, configuration Configuration) erro
...
+		var obj interface{}
+		if dec.Decode(&obj) != nil {
```
ivan-degtiarenko, line 109: This code ignores unexpected errors. How about we `break` without an error only on `io.EOF` returned?

roxctl/declarativeconfig/lint/lint_test.go
```diff
@@ -0,0 +1,183 @@
...
+      name: far_away
+oidc:
+    issuer: sample.issuer.com
+    mode: auto
+    clientID: CLIENT_ID
+    clientSecret: CLIENT_SECRET
+`),
+		},
+		"multiple configurations in file": {
+			yaml: []byte(`- name: some-name
```
ivan-degtiarenko, line 94: Looking at this example just now, I wonder if different declarative resource types would be more distinguishable if we forced to add `type: <resource type>` to every declarative resource. Example:
```
- name: ps
  type: PermissionSet:
  resources:
	...
- name: RHSSO
  type: AuthProvider
  ....
```

ivan-degtiarenko (commented):
I think what would be a logical step here is to make `create` call `lint` on the result after YAML is produced - WDYT?

roxctl/declarativeconfig/lint/lint.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type lintCmd struct {
+	env environment.Environment
+
+	file         string
+	fileContents []byte
+}
+
+func (l *lintCmd) Validate() error {
```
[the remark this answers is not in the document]
dhaus67: Called it `Construct()` like the other commands typically do.

pkg/declarativeconfig/configuration.go
dhaus67: What's the reasoning behind this?
In case there's an issue with decoding, we will exit early.

I don't think we can come up with a reasonable number here, since this is correlating with the YAML file you are parsing configuration from, and there usually isn't a hard limit on that.

roxctl/declarativeconfig/lint/lint_test.go
dhaus67: Well, I suppose we could have done this (similar to how we have [ROX-14652](https://issues.redhat.com/browse/ROX-14652) for adding the metadata field, this could include the type field).

However, I think it's not worthwhile to add it anymore. Since post-MVP it's likely we move to CRs for the definition of declarative configuation, they would have the type inferred by their `Kind`.

PR #5430, opened 2023-03-28
ROX-15402: Change to a sticky search bar in the breadcrumb

sachaudh:
## Description

This PR makes the search bar sticky in the namespace/deployment selectors. Because this is not naturally possible to do with the `Select` component, we had to make use of the `customContent` prop to display a `Menu` the same way it functions as it does today

Note: We can refactor the selectors to be more reusable (possibly making an `EntitySelector` component), but in order to keep the PR light, and considering future edge cases, we won't do that right now

https://user-images.githubusercontent.com/4805485/228352245-b2c08378-3b35-4a4e-84a2-2c2274412fc6.mov

### DIFF
--- a/ui/apps/platform/cypress/helpers/networkGraphPF.js
+++ b/ui/apps/platform/cypress/helpers/networkGraphPF.js
@@ -47,12 +47,10 @@ export function selectNamespace(namespace) {
         cy.get(networkGraphSelectors.selector.namespaceSelect).click();
         // Exact match to distinguish stackrox from stackrox-operator namespaces.
         cy.get(
-            `${
-                selectSelectors.patternFlySelect.openMenu
-            } .pf-c-menu__list-item [data-testid="namespace-name"]:contains(${new RegExp(
-                `^${namespace}$`
-            )})`
-        ).click();
+            `${selectSelectors.patternFlySelect.openMenu} .pf-c-menu__list-item [data-testid="namespace-name"]`
+        )
+            .contains(new RegExp(`^${namespace}$`))
+            .click();
         cy.get(networkGraphSelectors.selector.namespaceSelect).click();
     }, routeMatcherMapForClusterInNetworkGraph);
 }
@@ -61,12 +59,10 @@ export function selectDeployment(deployment) {
     interactAndWaitForResponses(() => {
         cy.get(networkGraphSelectors.selector.deploymentSelect).click();
         cy.get(
-            `${
-                selectSelectors.patternFlySelect.openMenu
-            } .pf-c-menu__list-item [data-testid="deployment-name"]:contains(${new RegExp(
-                `^${deployment}$`
-            )})`
-        ).click();
+            `${selectSelectors.patternFlySelect.openMenu} .pf-c-menu__list-item [data-testid="deployment-name"]`
+        )
+            .contains(new RegExp(`^${deployment}$`))
+            .click();
         cy.get(networkGraphSelectors.selector.deploymentSelect).click();
     }, routeMatcherMapForClusterInNetworkGraph);
 }
--- a/ui/apps/platform/cypress/helpers/networkGraphPF.js
+++ b/ui/apps/platform/cypress/helpers/networkGraphPF.js
@@ -47,10 +47,12 @@ export function selectNamespace(namespace) {
         cy.get(networkGraphSelectors.selector.namespaceSelect).click();
         // Exact match to distinguish stackrox from stackrox-operator namespaces.
         cy.get(
-            `${selectSelectors.patternFlySelect.openMenu} .pf-c-select__menu-item [data-testid="namespace-name"]`
-        )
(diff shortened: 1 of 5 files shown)


ui/apps/platform/src/Containers/NetworkGraph/NetworkGraphPage.css
```diff
@@ -73,3 +73,15 @@ div#topology-resize-panel table td {
...
+.pf-c-check__input {
+    height: 16px;
+    width: 16px;
+}
+
+.namespace-select .pf-c-select__menu,
+.deployment-select .pf-c-select__menu {
+    padding-top: 0;
+    padding-bottom: 0;
+    width: 220px;
```
vjwilson, line 86: This causes a lot of longer deployment names to be truncated. Could we get by with just a `min-width` or just min and max widths ?

ui/apps/platform/src/Containers/NetworkGraph/components/DeploymentSelector.tsx
```diff
@@ -67,47 +62,99 @@ function DeploymentSelector({
...
                     </span>
-                </SelectOption>
-            ))}
-        </SelectGroup>
-    ));
+                </MenuItem>
+            ));
+        if (menuItems.length === 0) {
+            return null;
+        }
```
vjwilson, line 86: suggestion: If we are controlling the content anyway, we could substitute an unselectable "No deployments in this namespace match" when this happens.

vjwilson (commented):
Good job of converting this.

While I was testing it locally, it seems to slow down from the same sort of re-render bug that plagued my initial implementation with the PF <Select> component. That's why I had to put the duplicate render return in that filter function, 

https://github.com/stackrox/stackrox/blob/master/ui/apps/platform/src/Containers/NetworkGraph/components/DeploymentSelector.tsx#L27-L51

as in the PF docs.

ui/apps/platform/src/Containers/NetworkGraph/components/DeploymentSelector.tsx
sachaudh: Are you saying to still show the namespace group text and just show the text below that with that message?

Screenshot 2023-03-29 at 11 39 34 AM

I feel like it creates a bit of noise, especially if we have a lot of namespaces. This is what my changes will show in this PR

Screenshot 2023-03-29 at 11 41 12 AM

@zhenpesky maybe you might have opinions about this. Right now we currently show something like this

Screenshot 2023-03-29 at 11 41 56 AM

zhenpesky (approved):
The one in the GIF looks good to me.

ui/apps/platform/src/Containers/NetworkGraph/components/DeploymentSelector.tsx
vjwilson: For the message per namespace, I was thinking about situations like this one I just caused in stagingdb.
Notice the one NS has no matches, while the other still does. Maybe a full message for each NS would be too business--especially if there were 3 or more namespaces--so maybe just a dash `—`.

I agree that if there are no matches at all, it would be better to short-circuit to your one "No deployments found" message.

zhenpesky: I like the dash idea for the use case when there is a deployment found in one of the namespaces while other namespaces have zero deployments, we would just use a dash to represent "no deployment found in this namespace".

sachaudh: Like this? @vjwilson @zhenpesky 

Screenshot 2023-03-29 at 5 14 48 PM

sachaudh: It looks a little odd to me, but if UX thinks its good I'll make the change

vjwilson: I think if there are no deployments in _any_ namespace, then your single "No deployments found" message should replace the NS list with all dashes.

ui/apps/platform/cypress/helpers/networkGraphPF.js
```diff
@@ -47,10 +47,8 @@ export function selectNamespace(namespace) {
         cy.get(networkGraphSelectors.selector.namespaceSelect).click();
         // Exact match to distinguish stackrox from stackrox-operator namespaces.
         cy.get(
-            `${selectSelectors.patternFlySelect.openMenu} .pf-c-select__menu-item [data-testid="namespace-name"]`
-        )
-            .contains(new RegExp(`^${namespace}$`))
```
sachaudh, line 52: @vjwilson I removed this regex and made it an exact match instead. Seems like the regex wasn't working. Do we need to do this anyways? Seems like we should know exactly what we want to match anyways, no?

PR #5440, opened 2023-03-29
Use non-deprecated pointer utils

ludydoo:
Just updating the pointer.* calls to non-deprecated ones


porridge:
Removed myself from reviewers and added current operator 🎩 , @vladbologa

0x656b694d (approved):
LGTM

PR #5442, opened 2023-03-29
ROX-12349: remove scorecard namespace flag

porridge:
## Description

This was a temporary workaround for a problem that appeared after migration to prow. See ticket.

## Checklist
- [ ] Revert the temporary commits

## Testing Performed

CI is sufficient.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -418,10 +418,16 @@ bundle-build: bundle.Dockerfile bundle-post-process ## Build the bundle image.
 
 .PHONY: bundle-test
 bundle-test: operator-sdk bundle-post-process ## Run scorecard tests against bundle files.
+	@echo dumping environment
+	env|sort
+	@echo end of environment
 	$(OPERATOR_SDK) scorecard $(SCORECARD_ARGS) build/bundle/
 
 .PHONY: bundle-test-image
 bundle-test-image: operator-sdk ## Run scorecard tests against bundle image in the registry. Bundle image must be pushed beforehand.
+	@echo dumping environment
+	env|sort
+	@echo end of environment
 	$(OPERATOR_SDK) scorecard $(SCORECARD_ARGS) $(BUNDLE_IMG)
 
 .PHONY: index-build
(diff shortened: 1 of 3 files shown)


operator/Makefile
```diff
@@ -422,10 +418,20 @@ bundle-build: bundle.Dockerfile bundle-post-process ## Build the bundle image.
 
 .PHONY: bundle-test
 bundle-test: operator-sdk bundle-post-process ## Run scorecard tests against bundle files.
+	@echo dumping environment
+	@env|sort
+	@echo end of environment
+	@tail -n 999999 $$HOME/.kube/config || true
+	@tail -n 999999 $$KUBECONFIG || true
```
SimonBaeumer, line 425: Why `-n 999999`? This looks like random number.

operator/tools/go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	sigs.k8s.io/kustomize/kustomize/v4 v4.5.7
 )
 
+replace github.com/operator-framework/operator-sdk v1.24.1 => github.com/porridge/operator-sdk v1.5.1-0.20230329122232-413d752ff864
```
SimonBaeumer, line 13: Why is a fork necessary? I've looked into the change and to me at a first glance looked unrelated.

SimonBaeumer (changes requested):

porridge:
This change is not ready for review, but there is no way to suppress the team-draco review request that gets added manually.
Maybe one could use the draft feature, but (a) I don't know whether that would help and (b) that in turn prevents tests from running.
I considered adding a [WiP] to PR title but I was afraid that the bot will try to be smart and this would have the same effect on tests as the draft status.
I added the `[do-not-merge]` label as last resort but I admit it's a little hidden there on the side...

(╯°□°)╯︵ ┻━┻ 

Suggestions most welcome on how to avoid such confusion in the process.

porridge:
Closing this for now, since https://github.com/kubernetes/client-go/issues/1241 might take a while to be processed.

PR #5448, opened 2023-03-29
Skip slack notification related tests on P/Z

pratham-m:
This PR intends to skip few of the groovy tests related to slack notifications on P/Z.
Changes are verified manually using `./gradlew test --tests=TestName`


gavin-stackrox (approved):
LGTM.
I wonder if we should consider a separate tag for this coverage? i.e. something along the lines of `@Tag("ARCH_VALIDATION")` for tests that we want to run to validate a new architecture.

pratham-m:
> LGTM. I wonder if we should consider a separate tag for this coverage? i.e. something along the lines of `@Tag("ARCH_VALIDATION")` for tests that we want to run to validate a new architecture.

Thanks. I will come up with a POC for the new approach.

PR #5449, opened 2023-03-29
ROX-14479 Adding validation logic for the central.spec.tls.additionalCAs

ludydoo:
## Description

This PR **simply adds the validation logic without hooking it into the operator**. I want to validate with the reviewers if this approach is OK, then go about plugging this logic into the operator in another PR. 

This adds a kubernetes-style validation method for statically validating the `Central` spec. Currently, this only validates the `spec.tls.additionalCAs` field, but it is intended to be used for validating the entire CR. This would be useful when implementing the validation in the admission webhook (ROX-7683).

### DIFF
--- a/operator/apis/platform/v1alpha1/validation/central.go
+++ b/operator/apis/platform/v1alpha1/validation/central.go
@@ -1,0 +1,166 @@
+package validation
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/stackrox/rox/operator/apis/platform/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+const (
+	errAdditionalCANameRequired         = "must specify a name for the additional CA"
+	errAdditionalCAContentRequired      = "must specify a certificate for the additional CA"
+	errAdditionalCAContentNoCertFound   = "no certificates found in the provided CA cert content"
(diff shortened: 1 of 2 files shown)


porridge (changes requested):
The long-term goal was to keep all the validation logic in the [translator](https://github.com/stackrox/stackrox/blob/master/operator/pkg/central/values/translation/translation.go#L59) and then just call the translation in the webhook (and discard the resulting helm value and just look at errors, if any) such that the assumptions made by validation and translation do not diverge with time.

The caveat is that since translation currently also looks at other resources referenced from the CR (I remember at least secrets), and validation probably should *not* do that (since this would impose ordering requirements that could be inconvenient for the user), the translation function would need to grow some kind of "mode" parameter to prevent it from doing that.

porridge:
> Implementing the validation in the operator would only validate installation's via the operator, but the additionalCAs can be installed with Helm, roxctl, Operator and even applied manually.
> 
> In this case, the validation only works for the operator. Customers using Helm and roxctl still face the same problem.
> 
> The validation must happen when Central reads the contents of the mounted secret from its filesystem. The according logic can be found in [tlsconfig.go:GetAdditionalCAs](https://github.com/stackrox/stackrox/blob/master/central/tlsconfig/tlsconfig.go#L27-L27).

I agree with all of the above, but I think this should not prevent us from also validating in the operator. The earlier the better for the user.

PR #5420, opened 2023-03-28
Add images table for CVE single page

dvail:
## Description

This adds the Images data table on the ImageCVE single page.

It also makes a couple of semi-related tweaks on the same page:
- Abstract out the fixability icons to avoid copy/pasting
- Rename the "Environmental impact" summary card

There are many TODOs throughout this section that need to be broken out into subtasks. Most of them are either verifying with the BE team that field usage is correct, or filling in details that will not cause a large structural change.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Load an ImageCVE page. The default entity tab should be "Images" instead of "Deployments". (TBD for the UI component, but is implemented in the logic.)

Request loading state:
image

Request error state:
image

Request complete state:
image

Test that the table does not load when the `entityTab` parameter is set to "Deployment" and the graphql request for images is not sent.
image

Test sorting of "image", "severity", "fix status" and "operating system" columns. (In local data, the latter three are always the same but I can verify that the correct parameter is sent in the request.)
image

Mobile view:
image

PR #5383, opened 2023-03-24
Integrate workload cve image single page table with search filter

dvail:
## Description

Integrates the WorkloadTableToolbar search filter with the images single page table.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Apply a severity filter and see it reflected in the URL, summary cards, and table:
image

Apply another severity filter and a fixable filter and see them reflected in the URL, summary cards, and table:
image

Apply entity filters and see them reflected in the URL (none of the data is impacted by these filters in my environment)
image
Verify that the entity filters are correctly sent as part of the query:
image

PR #5367, opened 2023-03-23
Add CVE single page details

dvail:
## Description

Implements the structure of the ImageCve page and displays available metadata.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Unit test and regression tests added
- [ ] Evaluated and added CHANGELOG entry if required
- [ ] Determined and documented upgrade steps
- [ ] Documented user facing changes (create PR based on [openshift/openshift-docs](https://github.com/openshift/openshift-docs) and merge into [rhacs-docs](https://github.com/openshift/openshift-docs/tree/rhacs-docs))

If any of these don't apply, please comment below.

## Testing Performed

Load a CVE page:
image

In the loading state:
image

If the metadata request results in an error:
image

PR #5451, opened 2023-03-29
Integrate ImageCVE page with EntityTypeToggleGroup

dvail:
## Description

Integrates the EntityTypeToggleGroup component.

I did find an issue unrelated to this change that needs to be fixed as a [follow up](https://github.com/stackrox/stackrox/pull/5458):

The image and deployment queries are only run when their tab is active, by design, to avoid the unnecessary overhead of loading data for both tabs. However, this means that the image/deployment counts are not updated when their tab is not active i.e. when filters change. I'll likely need to restructure these queries a bit.

## Checklist

## Testing Performed

Load the ImageCVE page, and see that the default URL parameter is "Image", and the "Image" tab is selected. The "CVE" tab should not render and the total image count should be visible.

image

Clicking the Deployments tab hides the images tab and correctly highlights the deployments toggle button and updates the URL parameter.
image

Applying a filter correctly updates the text in the toggle button.
image

### DIFF
--- a/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/EntityTypeToggleGroup.tsx
+++ b/ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/EntityTypeToggleGroup.tsx
@@ -2,51 +2,57 @@ import React from 'react';
 import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
 
 import useURLStringUnion from 'hooks/useURLStringUnion';
+import { NonEmptyArray } from 'utils/type.utils';
 import { entityTabValues, EntityTab } from './types';
 
 type EntityTabToggleGroupProps = {
-    entityContext?: EntityTab;
+    className?: string;
+    entityTabs?: Readonly<NonEmptyArray<EntityTab>>;
     cveCount?: number;
     imageCount?: number;
     deploymentCount?: number;
 };
 
 function EntityTabToggleGroup({
-    entityContext,
+    className = '',
+    entityTabs = entityTabValues,
     cveCount = 0,
     imageCount = 0,
     deploymentCount = 0,
 }: EntityTabToggleGroupProps) {
-    const [activeEntityTabKey, setActiveEntityTabKey] = useURLStringUnion(
-        'entityTab',
-        entityTabValues
-    );
+    const [activeEntityTabKey, setActiveEntityTabKey] = useURLStringUnion('entityTab', entityTabs);
 
     return (
-        <ToggleGroup className="pf-u-pl-md pf-u-pt-md">
-            {entityContext !== 'CVE' && (
+        <ToggleGroup className={className}>
+            {entityTabs.includes('CVE') ? (
                 <ToggleGroupItem
                     text={`${cveCount} CVEs`}
                     buttonId="cves"
                     isSelected={activeEntityTabKey === 'CVE'}
                     onChange={() => setActiveEntityTabKey('CVE')}
                 />
+            ) : (
+                <></>
             )}
-            {entityContext !== 'Image' && (
+            {entityTabs.includes('Image') ? (
                 <ToggleGroupItem
                     text={`${imageCount} Images`}
                     buttonId="images"
                     isSelected={activeEntityTabKey === 'Image'}
                     onChange={() => setActiveEntityTabKey('Image')}
                 />
+            ) : (
+                <></>
             )}
-            {entityContext !== 'Deployment' && (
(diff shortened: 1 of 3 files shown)


dvail:
* master
  * **PR #5383** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5383" target="_blank">Graphite</a> 
    * **PR #5367** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5367" target="_blank">Graphite</a> 
      * **PR #5386** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5386" target="_blank">Graphite</a> 
        * **PR #5420** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5420" target="_blank">Graphite</a> 
          * **PR #5451** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5451" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5451?utm_source=stack-comment).

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/EntityTypeToggleGroup.tsx
```diff
@@ -2,51 +2,55 @@ import React from 'react';
 import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core';
 
 import useURLStringUnion from 'hooks/useURLStringUnion';
+import { NonEmptyArray } from 'utils/type.utils';
 import { entityTabValues, EntityTab } from './types';
 
 type EntityTabToggleGroupProps = {
-    entityContext?: EntityTab;
+    entityTabs?: Readonly<NonEmptyArray<EntityTab>>;
```
dvail, line 9: This was added to support setting the default value in cases where `entityContext` was not `undefined`. e.g. If `entityContext="CVE"` the `useURLStringUnion` hook would still set the default parameter to "CVE" even though it was not a valid option.

Initially I left this prop alone and added an explicit `defaultEntity` prop, but ran into issues since the parent component using `EntityTabToggleGroup` had its own copy of the `useURLStringUnion` hook to read the current value and the parameters to each were different.

Doing it this way might make it easier to not forget to pass the same values to the hook and the component when declared as a const, but its probably subjective. 🤷‍♂️ 

The API difference is pretty much:
`<EntityTabToggleGroup entityContext="CVE" defaultEntity="Image" />`
vs
`<EntityTabToggleGroup entityTabs={["Image", "Deployment"] as const} />`

Any strong opinions either way?

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/EntityTypeToggleGroup.tsx
```diff
@@ -2,51 +2,55 @@ import React from 'react';
...
         <ToggleGroup className="pf-u-pl-md pf-u-pt-md">
-            {entityContext !== 'CVE' && (
+            {entityTabs.includes('CVE') ? (
                 <ToggleGroupItem
                     text={`${cveCount} CVEs`}
                     buttonId="cves"
                     isSelected={activeEntityTabKey === 'CVE'}
                     onChange={() => setActiveEntityTabKey('CVE')}
                 />
+            ) : (
```
dvail, line 32: A [PatternFly bug](https://github.com/patternfly/patternfly-react/issues/8877) was causing this to blow up with `&&` when the first condition evaluated to `false`, so the empty fragment is needed here.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/ImageCvePage.tsx
```diff
@@ -236,6 +236,29 @@ function ImageCvePage() {
                 </div>
                 <Divider />
                 <div className="pf-u-background-color-100 pf-u-flex-grow-1">
+                    <Split className="pf-u-px-sm pf-u-pb-md pf-u-align-items-baseline">
```
dvail, line 239: This section was pulled out of the error/loading states below so that if an error occurs in one tab we at least have the ability to change to the other tab.

ui/apps/platform/src/Containers/Vulnerabilities/WorkloadCves/EntityTypeToggleGroup.tsx
alwayshooin: oh i see. i don't have a strong pref either way so that works for me!

alwayshooin (approved):

PR #5457, opened 2023-03-29
ROX-14700: Resolving references in declarative configuration

ivan-degtiarenko:
## Description

Next changes are introduced:
* when declarative permission set referenced by some role is attempted to delete, it moves to status `DECLARATIVE_ORPHANED` and shows as unhealthy in integrations health; deleting referencing roles will result in deletion of the declarative set; re-adding permission set will result in moving its status back to `DECLARATIVE`
* same for access scopes
* same for declarative roles referenced by groups
* declarative and imperative groups belonging to deleted declarative auth provider are deleted with it
* new constraint of group datastore is to allow only groups with valid role name and auth provider id

## Checklist

## Testing Performed

Manual steps:
1. Created helm chart via `roxctl`
2. Updated `values.yaml`:
```
central:
  image:
    tag: 4.0.x-42-g4f8cc2c069
  db:
    image:
      tag: latest
    resources:
      requests:
        memory: "2Gi"
        cpu: "1"
  declarativeConfiguration:
    mounts:
      configMaps:
      - permission-set
      - access-scope
      - auth-provider
      - role
customize:
  envVars:
    ROX_DECLARATIVE_CONFIGURATION: true

imagePullSecrets:
  username: <quay username>
  password: <quay password>

scanner:
  replicas: 1
```
3. Deployed stackrox via helm chart
4. Created declarative resources referencing each other(see https://github.com/stackrox/stackrox/pull/5018 for generai idea of commands)
5. When resource referenced by other is deleted, we see:
* `UNHEALTHY` status for declarative resource we tried to delete
* `Declarative, Orphaned` status:
Screenshot 2023-04-11 at 16 59 06

Screenshot 2023-04-11 at 16 58 53


central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,17 +76,30 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
 	var authProviderIDs []string
 	for _, authProvider := range authProviders {
-		if err := u.authProviderRegistry.DeleteProvider(ctx, authProvider.GetId(), true, true); err != nil {
+		referencingGroups, err := u.groupDS.GetFiltered(ctx, func(group *storage.Group) bool {
+			return group.GetProps().GetAuthProviderId() == authProvider.GetId()
+		})
+		if err != nil || len(referencingGroups) > 0 {
+			if err == nil {
+				err = errox.ReferencedByAnotherObject.Newf("auth provider is still referenced by groups")
+			}
```
dhaus67, line 85: Less indentations and also let's provide the number of groups referenced and the auth provider name in the error message: 
```suggestion
		if len(referencingGroups) > 0 {
			err = errox.ReferencedByAnotherObject.Newf("auth provider %s is still referenced by %d groups", authProvider.GetName(), len(referencingGroups))
		}
		if err != nil {
```

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,17 +76,30 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
 			authProviderDeletionErr = multierror.Append(authProviderDeletionErr, err)
 			authProviderIDs = append(authProviderIDs, authProvider.GetId())
 
 			u.reporter.UpdateIntegrationHealthAsync(utils.IntegrationHealthForProtoMessage(authProvider, "", err,
 				u.idExtractor, u.nameExtractor))
+			authProvider.Traits.Origin = storage.Traits_DECLARATIVE_ORPHANED
+			if err = u.authProviderDS.UpdateAuthProvider(ctx, authProvider); err != nil {
+				authProviderDeletionErr = multierror.Append(authProviderDeletionErr, err)
+			}
```
dhaus67, line 94: Hm, so we'd currently mark an auth provider as `orphaned` if a) the filtering for groups call fails unexpectedly b) it's referenced by either declarative groups or non-declarative groups.

I'd say we shouldn't set this to orphaned if `errors.Is(err, errox.ReferencedByAnotherObject) == false`.

On the other part with the references, I'd agree on the declarative groups, but I'm not too sure on the non-declarative ones: the resolution would be to first go within the UI, delete all of the groups created, and then deletion would succeed.

Do we want to make the distinction within declarative config for imperative groups, when in reality if you'd call the API to delete auth providers the associated groups would be simply deleted?

Maybe my perception at the moment is too harsh, and in reality `DeleteResource` will be called more often than we'd initially expect (e.g. people do really move resources around a lot), but my expectation would be that delete would be something done more explicit and the consequences (in this case deleting the imperative groups) would be understood and acceptable.

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,17 +76,30 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
 		}
-		// TODO(ROX-14700): This currently also deletes imperative groups and should resolve these references instead.
-		if err := u.groupDS.RemoveAllWithAuthProviderID(ctx, authProvider.GetId(), true); err != nil {
-			log.Errorf("Error deleting groups for auth provider id %s: %v", authProvider.GetId(), err)
+		if err := u.authProviderRegistry.DeleteProvider(ctx, authProvider.GetId(), true, true); err != nil {
+			authProviderDeletionErr = multierror.Append(authProviderDeletionErr, err)
+			authProviderIDs = append(authProviderIDs, authProvider.GetId())
+
+			u.reporter.UpdateIntegrationHealthAsync(utils.IntegrationHealthForProtoMessage(authProvider, "", err,
+				u.idExtractor, u.nameExtractor))
```
dhaus67, line 102: super nit: could make this a separate function.

central/declarativeconfig/updater/permission_set_updater.go
```diff
@@ -60,6 +61,12 @@ func (u *permissionSetUpdater) DeleteResources(ctx context.Context, resourceIDsT
 			permissionSetIDs = append(permissionSetIDs, permissionSet.GetId())
 			u.reporter.UpdateIntegrationHealthAsync(utils.IntegrationHealthForProtoMessage(permissionSet, "", err,
 				u.idExtractor, u.nameExtractor))
+			if errors.Is(err, errox.ReferencedByAnotherObject) {
+				permissionSet.Traits.Origin = storage.Traits_DECLARATIVE_ORPHANED
+				if err = u.roleDS.UpsertPermissionSet(ctx, permissionSet); err != nil {
+					permissionSetDeletionErr = multierror.Append(permissionSetDeletionErr, err)
```
dhaus67, line 67:
```suggestion
					permissionSetDeletionErr = multierror.Append(permissionSetDeletionErr, errors.Wrap(err, "setting origin to orphaned")
```

central/declarativeconfig/updater/role_updater.go
```diff
@@ -8,28 +8,32 @@ import (
...
+	reporter       integrationhealth.Reporter
+	idExtractor    types.IDExtractor
+	nameExtractor  types.NameExtractor
+	groupDatastore groupDataStore.DataStore
 }
 
 var _ ResourceUpdater = (*roleUpdater)(nil)
 
-func newRoleUpdater(datastore roleDataStore.DataStore, reporter integrationhealth.Reporter) ResourceUpdater {
+func newRoleUpdater(datastore roleDataStore.DataStore, groupDatastore groupDataStore.DataStore, reporter integrationhealth.Reporter) ResourceUpdater {
```
dhaus67, line 30:
```suggestion
func newRoleUpdater(roleDatastore roleDataStore.DataStore, groupDatastore groupDataStore.DataStore, reporter integrationhealth.Reporter) ResourceUpdater {
```

central/declarativeconfig/updater/role_updater.go
```diff
@@ -55,6 +59,24 @@ func (u *roleUpdater) DeleteResources(ctx context.Context, resourceIDsToSkip ...
 	var roleDeletionErr *multierror.Error
 	var roleNames []string
 	for _, role := range roles {
+		referencingGroups, err := u.groupDatastore.GetFiltered(ctx, func(group *storage.Group) bool {
+			return group.GetRoleName() == role.GetName()
+		})
+		if err != nil || len(referencingGroups) > 0 {
```
dhaus67, line 65: Same as above regarding reducing the amount of indentation.

central/declarativeconfig/updater/role_updater.go
```diff
@@ -55,6 +59,24 @@ func (u *roleUpdater) DeleteResources(ctx context.Context, resourceIDsToSkip ...
...
+			}
+			roleDeletionErr = multierror.Append(roleDeletionErr, err)
+			roleNames = append(roleNames, role.GetName())
+
+			u.reporter.UpdateIntegrationHealthAsync(utils.IntegrationHealthForProtoMessage(role, "", err,
+				u.idExtractor, u.nameExtractor))
+			role.Traits.Origin = storage.Traits_DECLARATIVE_ORPHANED
+			if err = u.roleDS.UpdateRole(ctx, role); err != nil {
+				roleDeletionErr = multierror.Append(roleDeletionErr, err)
+			}
```
dhaus67, line 77: Same as above whether we should set this as orphaned when `errors.Is(err, errox.ReferencedByAnotherObject) == false`

central/declarativeconfig/utils/verify_origin.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/declarativeconfig"
+	"github.com/stackrox/rox/pkg/errox"
+)
+
+func VerifyReferencedResourceOrigin(referencing, referenced declarativeconfig.ResourceWithTraits, referencingName, referencedName string) error {
```
dhaus67, line 9: We can move this to `pkg/declarativeconfig`, no need to keep it central-specific.

dhaus67, line 9: Just from the way my brain thinks: can we switch `referencing` with `referenced` in the order of params? I read this within the tests as `default -> imperative` etc. instead of the other way around.

central/group/datastore/filter/filter.go
```diff
@@ -0,0 +1,50 @@
+package filter
+
+import (
+	"context"
+
+	"github.com/stackrox/rox/central/globaldb"
+	store2 "github.com/stackrox/rox/central/group/datastore/store"
```
dhaus67, line 7:
```suggestion
	groupStore "github.com/stackrox/rox/central/group/datastore/store"
```

central/main.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/stackrox/rox/central/auth/userpass"
 	authProviderDS "github.com/stackrox/rox/central/authprovider/datastore"
 	authProviderSvc "github.com/stackrox/rox/central/authprovider/service"
+	"github.com/stackrox/rox/central/authprovider/telemetry"
```
dhaus67, line 22:
```suggestion
	authProviderTelemetry "github.com/stackrox/rox/central/authprovider/telemetry"
```

central/group/datastore/filter/filter.go
```diff
@@ -0,0 +1,50 @@
...
+	} else {
+		store = bolt.New(globaldb.GetGlobalDB())
+	}
+}
+
+// GroupStoreSingleton returns the singleton providing access to the roles store.
+func GroupStoreSingleton() store2.Store {
+	once.Do(initialize)
+	return store
+}
```
dhaus67, line 50: Instead of this one, we could also have `GetFiltered` call this singleton internally and not expose the `store.Store` interface at all, right?

In this case, you could just inject within role / auth provider datastore a function type and mock it as such, instead of injecting the `store.Store` interface which provides way more access than just the singular function to clients.

pkg/declarativeconfig/context.go
```diff
@@ -24,7 +24,11 @@ type ResourceWithTraits interface {
 // CanModifyResource returns whether context holder is allowed to modify resource.
 func CanModifyResource(ctx context.Context, resource ResourceWithTraits) bool {
 	if ctx.Value(originCheckerKey{}) == allowOnlyDeclarativeOperations {
-		return resource.GetTraits().GetOrigin() == storage.Traits_DECLARATIVE
+		return IsDeclarativeOrigin(resource.GetTraits().GetOrigin())
 	}
 	return resource.GetTraits().GetOrigin() == storage.Traits_IMPERATIVE
 }
+
+func IsDeclarativeOrigin(origin storage.Traits_Origin) bool {
```
dhaus67, line 32: Since we introduced it beforehand, why not reuse `ResourceWithTraits` here instead?

proto/storage/traits.proto
```diff
@@ -32,21 +32,26 @@ message Traits {
...
   // Objects with the IMPERATIVE origin are allowed to be modified via API, not via declarative configuration.
   // They may reference all other objects.
+  // Objects with the DECLARATIVE_ORPHANED origin are not allowed to be modified via either API or declarative configuration.
+  // Objects with this origin will be cleaned up from the system immediately after they will stop be referenced by other resources.
+  // They may be referenced by all other objects.
   enum Origin {
      IMPERATIVE  = 0;
      DEFAULT     = 1;
      DECLARATIVE = 2;
+     DECLARATIVE_ORPHANED = 3;
```
dhaus67, line 54:
```suggestion
     IMPERATIVE           = 0;
     DEFAULT              = 1;
     DECLARATIVE          = 2;
     DECLARATIVE_ORPHANED = 3;
```

dhaus67 (changes requested):

central/declarativeconfig/updater/auth_provider_updater.go
ivan-degtiarenko: Hmm, thinking about how we structure auth provider + groups declarative config:
* they are always defined together
* it's possible to delete group from auth provider
* not possible to delete auth provider without deleting the group

I start to think we should delete all(not only declarative) groups belonging to auth provider if auth provider is deleted. WDYT?

central/declarativeconfig/updater/role_updater.go
ivan-degtiarenko: I realised `RemoveRole` actually checks for group references, so simplified :D

proto/storage/traits.proto
```diff
@@ -32,21 +32,26 @@ message Traits {
...
   // - DECLARATIVE: the object is created via declarative configuration.
+  // - DECLARATIVE_ORPHANED: the object is created via declarative configuration and then unsuccessfully deleted(for example, because it is referenced by another object)
   // Based on the origin, different rules apply to the objects.
   // Objects with the DECLARATIVE origin are not allowed to be modified via API, only via declarative configuration.
   // Additionally, they may not reference objects with the IMPERATIVE origin.
   // Objects with the DEFAULT origin are not allowed to be modified via either API or declarative configuration.
   // They may be referenced by all other objects.
   // Objects with the IMPERATIVE origin are allowed to be modified via API, not via declarative configuration.
   // They may reference all other objects.
+  // Objects with the DECLARATIVE_ORPHANED origin are not allowed to be modified via either API or declarative configuration.
```
dhaus67, line 47: Maybe let's call out that you can make a `DECLARATIVE_ORPHANED` resource `DECLARATIVE` again by re-creating its declarative configuration?

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -10,12 +10,16 @@ import (
...
+	"github.com/stackrox/rox/central/role/resources"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/authproviders"
+	"github.com/stackrox/rox/pkg/declarativeconfig"
 	"github.com/stackrox/rox/pkg/errox"
 	"github.com/stackrox/rox/pkg/integrationhealth"
 	"github.com/stackrox/rox/pkg/logging"
+	"github.com/stackrox/rox/pkg/sac"
 	"github.com/stackrox/rox/pkg/set"
+	utils2 "github.com/stackrox/rox/pkg/utils"
```
dhaus67, line 22: nit: name `declarativeconfig/utils` as `declarativeCfgUtils` and remove the `utils2` here

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,18 +83,37 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
+			if err = u.groupDS.Remove(ctxToUse, group.GetProps(), true); err != nil {
+				authProviderDeletionErr, authProviderIDs = u.processDeletionError(authProviderDeletionErr, err, authProviderIDs, authProvider)
+			}
+		}
+		// If err != nil, that means that one of the groups was not deleted.
+		// Since group is valid only if it references existing auth provider, we can't delete auth provider
+		// until all referencing groups are successfully deleted.
+		if err != nil {
+			continue
+		}
```
dhaus67, line 104: That condition isn't true though, in case you delete 10 groups and the last deletion was successful and all prior deletions failed, then `err = nil`.

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,18 +83,37 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
+		if err := u.authProviderRegistry.DeleteProvider(ctx, authProvider.GetId(), true, true); err != nil {
+			authProviderDeletionErr, authProviderIDs = u.processDeletionError(authProviderDeletionErr, err, authProviderIDs, authProvider)
 		}
 	}
 	return authProviderIDs, authProviderDeletionErr.ErrorOrNil()
 }
+
+func (u *authProviderUpdater) processDeletionError(authProviderDeletionErr *multierror.Error, err error, authProviderIDs []string, authProvider *storage.AuthProvider) (*multierror.Error, []string) {
+	authProviderDeletionErr = multierror.Append(authProviderDeletionErr, err)
+	authProviderIDs = append(authProviderIDs, authProvider.GetId())
```
dhaus67, line 114: Maybe let's change this to a string set and ensure we don't add duplicated auth provider IDs here - with the change on how we delete groups, imagine you'd delete 10 groups associated with a single auth provider, and all of them failed, we'd duplicate the ID here and I'm not entirely sure about the consequences this would have.

central/declarativeconfig/updater/auth_provider_updater.go
```diff
@@ -75,18 +83,37 @@ func (u *authProviderUpdater) DeleteResources(ctx context.Context, resourceIDsTo
...
+		}
+		if err := u.authProviderRegistry.DeleteProvider(ctx, authProvider.GetId(), true, true); err != nil {
+			authProviderDeletionErr, authProviderIDs = u.processDeletionError(authProviderDeletionErr, err, authProviderIDs, authProvider)
 		}
 	}
 	return authProviderIDs, authProviderDeletionErr.ErrorOrNil()
 }
+
+func (u *authProviderUpdater) processDeletionError(authProviderDeletionErr *multierror.Error, err error, authProviderIDs []string, authProvider *storage.AuthProvider) (*multierror.Error, []string) {
+	authProviderDeletionErr = multierror.Append(authProviderDeletionErr, err)
```
dhaus67, line 113: Looking at this, should we create errors _specifically_ for auth providers?

Currently, in case you had two auth providers created declaratively and both would be deleted, this would lead to a single error being reuse for both integration healths, maybe we should ensure that the error will only include actual errors for the specific auth provider?

dhaus67 (approved):
Minor comments otherwise LGTM 🚢 

This can be done in a follow-up: since we have the E2E tests defined for declarative config, let's also create E2E tests that cover this scenario and verifies that deleting things with references leads to unhealthy state.

PR #5458, opened 2023-03-29
Separate image count query from data loading query in image cve page

dvail:
## Description

Restructures the queries a bit so that the counts for `images` and `deployments` are always kept up to date, regardless of the user's current entity tab. The query that requests the full table data will still only request data for the active tab.

## Checklist

## Testing Performed

Load the ImageCVE page and see that the counts for both Images and Deployments are up to date.
image

Change to the Deployments tab, and apply a deployment filter. The counts for both images and deployments should update to match the filter, but the only request that is sent is the `SummaryData` request. The images request is not sent because "Deployments" is the active tab.
image


dvail:
@alwayshooin I'm going to move this back into draft until the `EntityToggle` component owns the entity count parts of the query, then I'll rebase and get this updated to support that design.

alwayshooin (approved):
yeah that makes sense to me, i was leaning towards doing the same as well since it doesn't make much sense to fragment one line of a query. lgtm!

PR #5475, opened 2023-03-30
Dashrews/db anti affinity

dashrews78:
## Description

chaos tests have shown that we will get into states where when central-db is killed sometimes the new pod will start on the same node and thus begin accessing the PVC because RWO on a volume does not enforce access on the same node.  Adding anti-affinity helps ensure that if the pod gets scheduled on the same node, it will remain at Pending until the previous one is gone.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


connorgorman:
@dashrews78 can close this one?

PR #5476, opened 2023-03-30
chore(roxctl): hide non-required global flags where applicable

dhaus67:
## Description

This PR refactors the currently global flags on the `roxctl` command.

Instead of exposing flags related to central connectivity (e.g. endpoint, protocol choice etc.) to _all_ roxctl commands, instead only expose it to commands which use it (or respectively, their subcommands).

The reason for this is that the by default created persistent flags about central connectivity (endpoint, grpc type etc.), may become confusing for clients using subtree of commands where central connectivity is not required.

Additionally, it lead to some of the flags to be less discoverable that are defined as persistent flags (i.e. the config map flags, the shared flags for auth provider types for `roxctl declarative-config create`).

Additionally all scripts have been adjusted which specified `roxctl -e <>`, since the flag now must be passed _after_ the command which provides the global flag has been passed (e.g. `roxctl central -e <>`

## Checklist

## Testing Performed

- see CI

### DIFF
--- a/roxctl/declarativeconfig/create/auth_provider.go
+++ b/roxctl/declarativeconfig/create/auth_provider.go
@@ -14,9 +14,25 @@ import (
 	"github.com/stackrox/rox/pkg/maputil"
 	"github.com/stackrox/rox/pkg/utils"
 	"github.com/stackrox/rox/roxctl/common/environment"
+	"github.com/stackrox/rox/roxctl/common/flags"
 	"github.com/stackrox/rox/roxctl/declarativeconfig/k8sobject"
 	"github.com/stackrox/rox/roxctl/declarativeconfig/lint"
 	"gopkg.in/yaml.v3"
+)
+
+var (
+	persistentFlagsToShow = []string{
+		"name",
+		"minimum-access-role",
+		"ui-endpoint",
+		"extra-ui-endpoints",
+		"required-attributes",
+		"groups-key",
+		"groups-value",
+		"groups-role",
+		k8sobject.ConfigMapFlag,
+		k8sobject.NamespaceFlag,
+	}
 )
 
 func authProviderCommand(cliEnvironment environment.Environment) *cobra.Command {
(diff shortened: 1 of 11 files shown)


0x656b694d:
Wouldn't it be possible to change the flags hierarchy instead of hiding some flags? Like instead of
```
        Base set with central connectivity
        /      \
declarative   central
```
implement this:
```
        Base set without central connectivity
        /      \
declarative   base set with central connectivity
                  \
                  central
```

janisz (approved):
How about adding bats test to verify if flags are not displayed?

dhaus67:
@janisz have a look again, I added:
- bats test which check whether only a single global flag is given (i.e. no-color) or multiple global flags with a partial match to the central flags (I didn't put _all_ global flags, since some command specify i.e. timeout as a global flag, and maintaining this seems also like a whole lot of pain).
- I changed the way the central connectivity flags are added, and only added it to the subcommands that require it (i.e. `central, cluster, collector, deployment, image, scanner, sensor`). Only these will have the connectivity flags and the rest won't get them to begin with, thus ensuring we don't need to hide them like @0x656b694d suggested. I quickly glanced over the commands, and we should be good to leave out the central connectivity flags on `completion, generate, helm, version, declarative config`.

janisz (approved):

roxctl/declarativeconfig/create/auth_provider.go
```diff
@@ -19,6 +19,21 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+var (
+	persistentFlagsToShow = []string{
```
0x656b694d, line 23: Unused?

parametalol (approved):
LGTM

janisz (approved):

dhaus67:
Note for reviewers: the original approach of only adding the connectivity flags to subcommands which require them doesn't work, this would create a breaking change.
Things like `roxctl -e "<central-address>" central ...` wouldn't work anymore. Sadly, this is also the _exact_ example we use within docs. So, reverted the changes back to the original implementation to opt-out / hide the commands for specific subcommands.

PR #5484, opened 2023-03-30
ROX-15448: Add security context to containers displayed in details of NG 2.0

vjwilson:
## Description

Revised requirements:

1. Because the only security context currently available in the Deployment API is at the container level, this ticket will focus on adding security context in each container section of the Deployment Details.
2. In the NG 1.0 implmentation, we limited the Security Context we showed to only Add Capabilities, Drop Capabilities, and Privileged. We will not impose that limit in this implementation–if a security context setting is not empty/null, we will show it.
3. We will not show security context settings that are empty/null.
4. If none of the security context settings are set, we will simply display "No security context

## Checklist

## Testing Performed

Manual testing

Examples (from the three containers within the StackRox `collector` pod

collector container:
Screen Shot 2023-03-30 at 5 04 34 PM

compliance container (a different set of securityContext settings):
Screen Shot 2023-03-30 at 5 04 48 PM

node-inventory (no securityContext settings):
Screen Shot 2023-03-30 at 5 04 55 PM

### DIFF
--- a/ui/apps/platform/src/Components/SecurityContext.tsx
+++ b/ui/apps/platform/src/Components/SecurityContext.tsx
@@ -18,12 +18,10 @@ type SecurityContextProps = {
 };
 
 function SecurityContext({ securityContext }: SecurityContextProps) {
-    // build a map of only those properties that actually have values
-
-    // sort the keys of the prop, so that the map is in alpha order
+    // sort the keys of the security context, so any properties are shown in alpha order
     const sortedKeys = Object.keys(securityContext).sort();
 
-    //
+    // build a map of only those properties that actually have values
     const filteredValues = new Map();
     sortedKeys.forEach((key) => {
         const currentValue = securityContext[key];
--- a/ui/apps/platform/src/Components/SecurityContext.tsx
+++ b/ui/apps/platform/src/Components/SecurityContext.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import { isEmpty } from 'lodash';
+import {
+    Card,
+    CardBody,
+    CardTitle,
+    DescriptionList,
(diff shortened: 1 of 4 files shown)


ui/apps/platform/src/Components/SecurityContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+} from '@patternfly/react-core';
+
+import { ContainerSecurityContext } from 'types/deployment.proto';
+
+type SecurityContextProps = {
+    securityContext: ContainerSecurityContext;
+};
+
+function SecurityContext({ securityContext }: SecurityContextProps) {
+    // build a map of only those properties that actually have values
```
sachaudh, line 21: Is this comment relevant?

ui/apps/platform/src/Components/SecurityContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    // sort the keys of the prop, so that the map is in alpha order
+    const sortedKeys = Object.keys(securityContext).sort();
+
+    //
```
sachaudh, line 26: We can remove this comment line or maybe you wanted to fill it out

ui/apps/platform/src/Components/SecurityContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const filteredValues = new Map();
+    sortedKeys.forEach((key) => {
```
sachaudh, line 28: Proposal: This logic might benefit from being extracted into a util function + tested

sachaudh (approved):
LGTM. A suggestion as well

zhenpesky (approved):

PR #5492, opened 2023-03-31
Disable re-sync for deploy.sh

fredrb:
## Description

Disable re-sync when deploying using deploy scripts. It should work for both k8s and OpenShift.

## Checklist

## Testing Performed

Many for the last 6 months


fredrb:
The same `CSVTest` test seem to be failing for all other PRs in this job.

PR #5493, opened 2023-03-31
ROX-13935: Enable RHCOS FF & add a failsafe switch

vikin91:
## Description

This PR changes the RHCOS NodeScanning feature flag to an env variable, that is *enabled* by default.
We keep the env variable in order to have a possibility to disable the feature in case something unexpected would happen.

Moreover, we remove the no longer used FakeNodeScanner from this repo (it still needs to be removed from the scanner repo).

⚠️ We need a similar PR for the Scanner repo ⚠️ 

## Checklist
- [x] Evaluated and added CHANGELOG entry if required ⚠️ 👉 please review the changelog entry for the feature

### N/A

## Testing Performed

- [ ] CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
 - ROX-15102: new `public_config.telemetry` boolean property of the `/v1/config`
   endpoint request that allows for querying the state, enabling or disabling the
   configured telemetry collection.
+- ROX-10818: vulnerability scanning of node components installed through RPM on
+  OpenShift cluster nodes running Core OS (RHCOS).
+
 
 ### Removed Features
 
--- a/central/reprocessor/reprocessor_unit_test.go
+++ b/central/reprocessor/reprocessor_unit_test.go
@@ -9,7 +9,7 @@ import (
 	nodeDatastoreMocks "github.com/stackrox/rox/central/node/datastore/mocks"
 	riskManagerMocks "github.com/stackrox/rox/central/risk/manager/mocks"
 	"github.com/stackrox/rox/generated/storage"
-	"github.com/stackrox/rox/pkg/features"
+	"github.com/stackrox/rox/pkg/env"
 	nodesEnricherMocks "github.com/stackrox/rox/pkg/nodes/enricher/mocks"
 )
 
@@ -69,7 +69,7 @@ func Test_loopImpl_reprocessNode(t *testing.T) {
 		},
 	}
 	t.Setenv("ROX_RHCOS_NODE_SCANNING", "true")
-	if !features.RHCOSNodeScanning.Enabled() {
+	if !env.RHCOSNodeScanning.BooleanSetting() {
 		t.Log("Assuming this is a release build, so skipping due to ROX_RHCOS_NODE_SCANNING=false")
 		return
 	}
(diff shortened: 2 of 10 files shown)


Maddosaurus (approved):
LGTM!

jvdm (changes requested):
I am not sure we should make this an env variable. Reasons:

 1. Did we test scenarios where we disabled RHCOS scanning after enabling it?
 2. Is it safe for customers to open the door to misconfiguration of this env variable across Pods, Scured Clusters, and Central?
 3. Did we test or assess the impact of (2.)?

Maybe allowing customers only to disable Compliance from calling node-inventory is safer. I acknowledge that an env var in Scanner will bring back the "RHCOS not supported" warning. Is that the reason to allow customers to disable the code paths in Sensor, Central, and Scanner, or is there something else? Do we have reasonable answers to 1., 2., and 3.?

Finally, should we mention the environment variable in the changelog? Where is it going to be documented?

Maddosaurus:
@jvdm I believe the thought emerged from our last check in meeting and the question whether customers can turn off the feature, should it have unintended impact on their systems. 
The idea is to remove this switch after the first release. It is intended as an additional temporary safety. It is not intended to be kept or even turned off and back on regularly. 
I agree that we open the door for possible misconfigurations, but in the event that something unexpected happens because of Node Scanning, having an undocumented env var might be the better option compared to cutting and distributing a patch release to disable/fix it.

With the context that this is a temporary measure: What do you think of having it for now and removing it in the next release?

vikin91:
@jvdm with introduction of this env variable, we want to have a temporal (1 release) possibility to switch off the feature for selected customers if something would lead in to problems. As Matthias mentioned, @sbadve-RH askid for such feature during our recent check-in meeting.
This variable would remain undocumented, as we would only recommend setting it to false when a customer opens a support ticket and we could confirm that the problem might be caused by this feature. In that case, the support engineer would guide the customer and thus the official doc is not required.

I have not tested enabling and then disabling this flag, but I can do this.
We can indeed limit the scope of that variable only to Compliance, but applying it to Sensor and Central would also have positive results, as we will skip starting one sensor component and one central pipeline, which will save some resources.

jvdm (approved):
Thanks for clarifying both @Maddosaurus and @vikin91.

I believe the ask was to disable the node scanning. So, the scope of the env toggle would be Compliance. I think using an env var across clusters increases the blast radius of this change, and the risk might outweigh the drawback of the Sensor and Central pipelines that would be idling -- the impact of idling is very small. This is certainly worth testing to check the customer experience if they disable RHCOS using a var across multiple components. I think not documenting the var and making an internal implementation aspect is a good approach.

Since you plan to test it, I will give you preliminary approval to avoid the cross-timezone back-and-forth. Once you see healthy results from the test, I think it would be OK to merge this.

vikin91:
I have tested the emergency shutdown by executing the following commands:

```
KUBECONFIG="$HOME/.cluster/kubeconfig" ocs set env deploy/scanner ROX_RHCOS_NODE_SCANNING="false"
KUBECONFIG="$HOME/.cluster/kubeconfig" ocs set env deploy/sensor ROX_RHCOS_NODE_SCANNING="false"
KUBECONFIG="$HOME/.cluster/kubeconfig" ocs set env deploy/central ROX_RHCOS_NODE_SCANNING="false"
KUBECONFIG="$HOME/.cluster/kubeconfig" ocs set env daemonsets/collector --containers="compliance" ROX_RHCOS_NODE_SCANNING="false"
KUBECONFIG="$HOME/.cluster/kubeconfig" ocs set env daemonsets/collector --containers="node-inventory" ROX_RHCOS_NODE_SCANNING="false"
```

This resulted in a clean feature shutdown without any problems. The state of the UI after that looks like this
Screenshot 2023-04-04 at 11 26 18
Screenshot 2023-04-04 at 11 26 07
Screenshot 2023-04-04 at 11 26 00

vikin91:
I have enabled the feature after disabling it, and it works correctly too.

vikin91:
The two failing e2e tests (`CSVTest`) are not related - merging this one

PR #5494, opened 2023-03-31
ROX-15337: Bump required OCP version to 4.10+ for the operator

msugakov:
## Description

Note: this is upstream change. The downstream change is done separately, the status can be checked in
https://issues.redhat.com/browse/ROX-15337

From my reading of
- https://redhat-connect.gitbook.io/certified-operator-guide/ocp-deployment/operator-metadata/bundle-directory/managing-openshift-versions
- https://docs.openshift.com/container-platform/4.12/operators/operator_sdk/osdk-working-bundle-images.html

the operator will not be published to OCP operator catalogs that fall outside of range, but it does not seem that OCP with earlier version will reject installing the operator. That's what I'll check when testing manually.

## Checklist
- ~~[ ] Unit test and regression tests added~~ - won't do in the scope of this task. Have https://issues.redhat.com/browse/ROX-16277 to implement automated tests. This time will test manually.

These aren't needed and won't be done:

## Testing Performed

* Relying on CI.
* [x] Tested manually whether OCP 4.8/4.9 does not allow installing the operator.

OCP does not prevent operator from installing (tested manually with v4.8.57). This confirms that the leveraged setting controls only versions of OpenShift catalogs that get the operator published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Please avoid adding duplicate information across this changelog and JIRA/doc inp
   to fetch vulnerability updates from Central. The flag will be removed as part of the 4.2.0 release.
 - ROX-15925: The KernelModule collection method is deprecated in favor of EBPF. This method will be removed in the 4.1 release.
 - Deprecated v1.0 of Network Graph. Please switch to the new 2.0 version for improved functionality and a better user experience.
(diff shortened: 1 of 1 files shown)


jschnath (approved):
LGTM

fredrb (approved):

SimonBaeumer (approved):

msugakov:
Nongroovy e2e tests seem to run against GKE cluster therefore the change here should not cause the observed failure. Merging.

PR #5506, opened 2023-03-31
Jv dont show listening endpoints with no process info

JoukoVirtanen:
## Description

In the following scenario we could end up with a situation where for a brief period of time the only information reported for a listening endpoint is the port, protocol, and deployment Id. Central receives a listening endpoint and matching process indicator. The process indicator is deleted before the message that the port is closed reaches central. In such cases the port is almost certainly closed and should not be reported by the API.

## Checklist
- [x] Added unit test

## Testing Performed

CI is sufficient

### DIFF
--- a/central/processlisteningonport/store/postgres/full_store.go
+++ b/central/processlisteningonport/store/postgres/full_store.go
@@ -128,6 +128,10 @@ func (s *fullStoreImpl) readRows(
 			execFilePath = msg.GetProcess().GetProcessExecFilePath()
 		}
 
+		if podID == "" && containerName == "" && name == "" && args == "" && execFilePath == "" {
+			continue
+		}
+
 		plop := &storage.ProcessListeningOnPort{
 			Endpoint: &storage.ProcessListeningOnPort_Endpoint{
 				Port:     msg.GetPort(),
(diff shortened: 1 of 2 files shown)


central/processlisteningonport/store/postgres/full_store.go
```diff
@@ -128,6 +128,13 @@ func (s *fullStoreImpl) readRows(
 			execFilePath = msg.GetProcess().GetProcessExecFilePath()
 		}
 
+		// If we don't have any of this information from either the process indicator side or
+		// processes listening on ports side, the process indicator has been deleted and the
+		// port has been closed. Central just hasn't gotten the message yet.
+		if podID == "" && containerName == "" && name == "" && args == "" && execFilePath == "" {
```
JoukoVirtanen, line 134: We could remove the process listening on port/listening endpoint in such cases, but that can be done later.

central/processlisteningonport/store/postgres/full_store.go
```diff
@@ -128,6 +128,13 @@ func (s *fullStoreImpl) readRows(
...
+			continue
+		}
+
```
JoukoVirtanen, line 137: We could remove the process listening on port/listening endpoint in such cases, but that can be done later.

erthalion (approved):
Pls, make a follow up later on about simplifying this condition, it looks weird. It probably has to check only `GetProcess` and the result of `procMsg`.

PR #5513, opened 2023-04-03
Store IDs for fake workloads to avoid overly reconciling

connorgorman:
## Description

Every time Central bounces or Sensor OOMs during a scale run then we end up reconciling every deployment and pod, which is unrealistic and ends up causing lots of churn for deployments, pods, alerts, network flows and processes. This should alleviate much of that by storing the active ids locally in Sensor and then recreate the active components based on that if there is a restart

```
sensor/service/pipeline/reconciliation: 2023/04/03 04:27:59.550878 reconciliation.go:148: Info: Reconciliation done. Number of objects deleted by type: map[*central.SensorEvent_Binding:397 *central.SensorEvent_ComplianceOperatorProfile:0 *central.SensorEvent_ComplianceOperatorResult:0 *central.SensorEvent_ComplianceOperatorRule:0 *central.SensorEvent_ComplianceOperatorScan:0 *central.SensorEvent_ComplianceOperatorScanSettingBinding:0 *central.SensorEvent_Deployment:0 *central.SensorEvent_Namespace:0 *central.SensorEvent_NetworkPolicy:0 *central.SensorEvent_Node:0 *central.SensorEvent_Pod:0 *central.SensorEvent_Role:0 *central.SensorEvent_Secret:0 *central.SensorEvent_ServiceAccount:0]
```

## Checklist

## Testing Performed

Will run a scale run


sensor/kubernetes/fake/fake.go
```diff
@@ -123,7 +125,13 @@ func NewWorkloadManager(config *WorkloadManagerConfig) *WorkloadManager {
 		log.Panicf("could not unmarshal workload from file due to error (%v): %s", err, data)
 	}
 
+	db, err := pebble.Open("/var/cache/stackrox/pebble.db", &pebble.Options{})
+	if err != nil {
+		log.Panicf("could not open id storage")
+	}
```
lvalerom, line 131: Can we make this configurable? I think local-sensor will fail to start (unless run as root) if we try to start it with this path.

connorgorman: oh yeah, good point

connorgorman: Done, made it an env var that we apply to the scripts but not to local

dashrews78 (approved):
looks good from my perspective

sensor/kubernetes/fake/fake.go
lvalerom: Awesome! 👍

lvalerom (approved):
LGTM. Thanks for adding the env variable.

PR #5524, opened 2023-04-03
Dashrews/ROX-16326-migration upgrade fix

dashrews78:
## Description

customer found issue migrating with old cluster health statuses that did not have a UUID as the ID.  Basically they didn't have a UUID.  The copyFrom logic should have logged those and moved on, but the check was in the wrong place.  Moved the check before adding to rows and updated tests to demonstrate this.

## Checklist

## Testing Performed

Updated migration tests to drive the condition.


migrator/migrations/n_10_to_n_11_postgres_cluster_health_statuses/postgres/postgres_plugin.go
```diff
@@ -139,6 +139,10 @@ func (s *storeImpl) copyFromClusterHealthStatuses(ctx context.Context, tx *postg
 			return marshalErr
 		}
 
+		if pgutils.NilOrUUID(obj.GetId()) == nil {
```
c-du, line 142: Do we know how does this happen? Why the id is not as expected?

dashrews78: @c-du Turns out it is because `n_94` used the wrong bucket name

`clusterHealthStatusBucket = []byte("cluster_health_status")`
instead of
`bucket = []byte("clusters_health_status")`

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/postgressimpleaccessscopes/postgres_plugin.go
```diff
@@ -121,6 +121,11 @@ func (s *storeImpl) copyFromSimpleAccessScopes(ctx context.Context, tx *postgres
 			return marshalErr
 		}
 
+		if pgutils.NilOrUUID(obj.GetId()) == nil {
+			log.Warnf("Id is not a valid uuid -- %v", obj)
```
c-du, line 125: We still need to use utils.Should.

migrator/migrations/n_52_to_n_53_postgres_simple_access_scopes/postgrespermissionsets/postgres_plugin.go
```diff
@@ -121,6 +121,11 @@ func (s *storeImpl) copyFromPermissionSets(ctx context.Context, tx *postgres.Tx,
 			return marshalErr
 		}
 
+		if pgutils.NilOrUUID(obj.GetId()) == nil {
+			log.Warnf("Id is not a valid uuid -- %v", obj)
```
c-du, line 125: utils.Should

migrator/migrations/n_42_to_n_43_postgres_process_indicators/postgres/postgres_plugin.go
```diff
@@ -153,6 +151,11 @@ func (s *storeImpl) copyFromProcessIndicators(ctx context.Context, tx *postgres.
 			return marshalErr
 		}
 
+		if pgutils.NilOrUUID(obj.GetId()) == nil {
+			log.Warnf("Id is not a valid uuid -- %v", obj)
```
connorgorman, line 155: Don't print a process indicator as it can potentially have secrets in the command line. Perhaps the same applies for secret below even though we don't take the data fields

migrator/migrations/n_23_to_n_24_postgres_image_integrations/postgres/postgres_plugin.go
```diff
@@ -127,6 +125,11 @@ func (s *storeImpl) copyFromImageIntegrations(ctx context.Context, tx *postgres.
 			return marshalErr
 		}
 
+		if pgutils.NilOrUUID(obj.GetId()) == nil {
+			log.Warnf("Id is not a valid uuid -- %v", obj)
```
connorgorman, line 129: Definitely do not log this

connorgorman (changes requested):
Evaluate the log lines for sensitive data

migrator/migrations/n_03_to_n_04_postgres_deployments/postgres/postgres_plugin.go
```diff
@@ -148,8 +146,8 @@ func insertIntoDeploymentsContainers(ctx context.Context, batch *pgx.Batch, obj
 		obj.GetResources().GetMemoryMbRequest(),
 		obj.GetResources().GetMemoryMbLimit(),
 	}
-	if pgutils.NilOrUUID(deployments_Id) == nil {
-		utils.Should(errors.Errorf("deployments_Id is not a valid uuid -- %v", obj))
+	if pgutils.NilOrUUID(deploymentID) == nil {
```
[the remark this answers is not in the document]
dashrews78: since I had to go back in for logging the object, I took care of this while I was there.

connorgorman (approved):

PR #5532, opened 2023-04-03
Refactor image component table to load async

dvail:
## Description

This refactors the current embedded table that contains image component vulnerabilities with two main changes:
1. We now use the following query structure for the Image single page query:
```graphql
# This is the top level CVE list
imageVulnerabilities {
    # Load the components affected by the parent CVE
    imageComponents {
        # Load CVE data scoped to the component, which isn't available at the first level
        imageVulnerabilites {
        }
    }
}
```
2. The nested `imageComponents->imageVulnerabilities` piece of the query is loaded async, per table row, so that we aren't executing this nested query for every row on the page every time the page loads.

## Change overview

The [old ImageComponentsTable](https://github.com/stackrox/stackrox/pull/5532/files#diff-ea040fd3f695436ffdfbce14d18e811e7de2e4a2de17e500329269cc9d83aabb) was renamed to the more appropriate [ComponentVulnerabilitiesTable](https://github.com/stackrox/stackrox/pull/5532/files#diff-74413d8e25b9d439004ed7ee2c643fd1afe38c61d1cf09d8880e6418d7d044ab).

This table now accepts a list of objects that map an [image metadata to it's component vulnerabilities](https://github.com/stackrox/stackrox/pull/5532/files#diff-74413d8e25b9d439004ed7ee2c643fd1afe38c61d1cf09d8880e6418d7d044abR145), which is then used to [flatten](https://github.com/stackrox/stackrox/pull/5532/files#diff-74413d8e25b9d439004ed7ee2c643fd1afe38c61d1cf09d8880e6418d7d044abR103) the data into a single list of rows.

The graphql fragments needed were also pulled into the table component from their previous place in the [query hook](https://github.com/stackrox/stackrox/pull/5532/files#diff-f3e267e981eabd23ada33faf82ad75fcc753b9f2b3a26f07bfa9cf4903e37e68).

Finally, instead of loading the component vulns at the top level when the page loads, an [intermediate component](https://github.com/stackrox/stackrox/pull/5532/files#diff-0b4253edbec8d3bfe7e210a0844a68a17da90c8013987b5febb4a64a29d48e22) was created that handles the async loading of the data. This will make it easy to load the data for the Deployments tab in the future, which is going to require a slightly different query. Plus if it turns out we don't want to run these async, it will be easy to lift the query parts back up to the top level component on each page.

## TODOs/Follow ups

1. One problem with this approach is we can no longer display the component name in the top level "component" column when there is only a single component affected by the vulnerability. IMO this is ok, since this is an edge case for when there is exactly one affected component, but we should confirm with UX. 
2. The loading spinner is very slightly clunky in the table. We could attempt to preload the query on hover, but that might be overthinking things at this point.

## Checklist

## Testing Performed

TODO(replace-me)
screenshots.


dvail:
* master
  * **PR #5532** <a href="https://app.graphite.dev/github/pr/stackrox/stackrox/5532" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/stackrox/5532?utm_source=stack-comment).

PR #5533, opened 2023-04-03
[ROX-16268] : Query parser should skip plus and comma if within double quotes

charmik-redhat:
## Description

Plus `+` is used to indicate a conjunction (`AND`) between match field queries and comma `,` is used to indicate a disjunction (`OR`). But they can also be a part of search value for a field. To allow that if a plus or comma is enclosed in quotes, it should be ignored.

## Checklist

## Testing Performed

Unit tests

### DIFF
--- a/pkg/search/parser.go
+++ b/pkg/search/parser.go
@@ -87,8 +87,15 @@ func GetFieldValueFromQuery(query string, label FieldLabel) (string, bool) {
 func splitQuery(query string) []string {
 	var pairs []string
 	var previousEnd, previousPlusIndex int
-	prevDoubleQuoteCount := 0
+	insideDoubleQuotes := false
 	for i, rune := range query {
+		if rune == '"' {
+			insideDoubleQuotes = !insideDoubleQuotes
+			continue
+		}
+		if insideDoubleQuotes {
+			continue
+		}
 		if rune == ':' && previousPlusIndex != 0 {
 			if previousEnd > previousPlusIndex {
 				continue
@@ -98,10 +105,7 @@ func splitQuery(query string) []string {
 			previousEnd = previousPlusIndex + 1
 			continue
 		}
-		if rune == '"' {
-			prevDoubleQuoteCount++
-		}
-		if rune == '+' && prevDoubleQuoteCount%2 == 0 {
+		if rune == '+' {
 			previousPlusIndex = i
 		}
 	}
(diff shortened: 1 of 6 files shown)


connorgorman (changes requested):
Please add tests

charmik-redhat:
> Please add tests

Added

pkg/search/parser.go
```diff
@@ -84,6 +84,50 @@ func GetFieldValueFromQuery(query string, label FieldLabel) (string, bool) {
 	return "", false
 }
 
+func splitQuery(query string) []string {
```
connorgorman, line 87: I think that it would be beneficial to simply test these functions distinctly. In fact, I think that parser_test.go already has this

pkg/search/parser.go
```diff
@@ -84,6 +84,50 @@ func GetFieldValueFromQuery(query string, label FieldLabel) (string, bool) {
...
+		if rune == ':' && previousPlusIndex != 0 {
+			if previousEnd > previousPlusIndex {
+				continue
+			}
+
+			pairs = append(pairs, query[previousEnd:previousPlusIndex])
+			previousEnd = previousPlusIndex + 1
+			continue
+		}
+		if rune == '"' {
```
connorgorman, line 101: We can just use a flag for inside double quotes right? None of our keywords should matter if we're inside a "

charmik-redhat: Do you mean use a set a bool flag instead of the doubleQuotesCount int ?

pkg/search/parser_test.go
```diff
@@ -52,6 +60,32 @@ func TestSplitQuery(t *testing.T) {
 	}
 }
 
+func TestSplitCommaSeparateValues(t *testing.T) {
```
connorgorman, line 63: Not sure it matters, but this is a deviation from strings.Split
```
		{
			commaSeparatedVals: ",,",
			splitValues:        []string{"", "", ""},
		},
```

connorgorman (approved):

pkg/search/parser_test.go
charmik-redhat: It shouldn't matter because `splitCommaSeparatedvalues` is always called after `parsePair` and it handles strings ending with comma. But I added a fix for that case for consistency.

PR #5535, opened 2023-04-03
Allow empty compliance strings and populate ID

connorgorman:
## Description

Valid proto messages can be 0 bytes, which can happen when you have an uninitialized cluster that you ran compliance against

## Checklist

## Testing Performed

Will run it against this use case


migrator/migrations/n_20_to_n_21_postgres_compliance_strings/legacy/rocksdb_store.go
```diff
@@ -94,7 +93,7 @@ func (r *rocksdbStore) UpsertMany(_ context.Context, objs []*storage.ComplianceS
 	defer batch.Destroy()
 
 	for _, obj := range objs {
-		key := r.createKay()
+		key := r.createKey()
```
c-du: Is the key created here to be consumed in line 57? It looks like it will always end with ":runID"

connorgorman: I just cleaned this up a bit and passed in the id from the compliance strings so that it's testable

c-du (approved):

dashrews78 (approved):

PR #5538, opened 2023-04-03
Replace Th which has empty text with Td

pedrottimark:
### Problem

Solve minor issue from axe DevTools in Search results table:

> Table header text should not be empty

### Analysis

PatternFly examples follow the accessibility work-around to render `Td` instead of `Th` element.

### Solution

Join them if you cannot beat them: an attempt to render **Actions** required additional utility classes in both the table **head** and table **data** cells.

Delete extra `Th` element that does not correspond to a column in `EnableDisableNotificationModal` component.

### Residue

1. Replace `any` in `VulnMgmtReportTablePanel` component.

## Checklist

### Manual testing

1. Visit /main/violations

    * with PatternFly `Th` element that renders as HTML `th` element.
        Th_axe

        Th_inspect

    * with PatternFly `Td` element that renders as HTML `td` element
        Td_axe

        Td_inspect


dvail (approved):
If it works it works. Would be interesting to see how different screen readers handle the different elements, or if the aXe dev tools gave a similar error with `<th style="visibility: hidden">Actions</th>`.

PR #5544, opened 2023-04-04
ROX-16279: Add new flag to diagnostic bundle

fredrb:
## Description

Allows diagnostic bundles to have alert information.

**Context:** the reason why this is being added is to check whether Sensors with re-sync disabled are behaving as expected, without having to re-run sensor with re-sync enabled and compare alerts. The alternative we are proposing here is: when requesting a diagnostic bundle, allow a flag that will request alerts, reprocess all deployments and request alerts again. What we (as team Maple) can do with this info is: check that there is no delta between the two. This gives us confidence that, even after reprocessing everything, Sensor was able to identify all resource changes that led to alerts. 

This flag should be used only through `roxctl` and it's here as a sanity check to test that secured clusters with re-sync disabled are producing the correct alerts. 

Since the new files (`alerts-*.json`) will take a while to generate, we shouldn't have them enabled by default, and should only be used explicitly by customers that are testing the re-sync changes. 

## Checklist

## Testing Performed

- [ ] Request diagnostic bundle via `roxctl`

### DIFF
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -641,7 +641,15 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 
 		// We have to sleep here to make sure that any new alerts that could be generated after requesting all deployments
 		// to be reprocessed will appear in the new JSON file.
-		time.Sleep(opts.withReprocessDelay)
+		select {
+		case <-debugDumpCtx.Done():
+			log.Error("timeout reached trying to create the diagnostic bundle")
+			if err := zipWriter.Close(); err != nil {
+				log.Error(err)
+			}
+			return
+		case <-time.After(opts.withReprocessDelay):
+		}
 		fetchAndAddJSONToZip(debugDumpCtx, zipWriter, "alerts-after.json", s.fetchAlerts)
 	}
 
--- a/central/debug/service/service.go
+++ b/central/debug/service/service.go
@@ -17,6 +17,7 @@ import (
 	"github.com/gogo/protobuf/types"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
+	alertStore "github.com/stackrox/rox/central/alert/datastore"
 	"github.com/stackrox/rox/central/cluster/datastore"
 	configDS "github.com/stackrox/rox/central/config/datastore"
 	"github.com/stackrox/rox/central/globaldb"
@@ -29,6 +30,7 @@ import (
 	"github.com/stackrox/rox/central/sensor/service/connection"
 	"github.com/stackrox/rox/central/telemetry/gatherers"
 	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/internalapi/central"
 	"github.com/stackrox/rox/generated/storage"
 	"github.com/stackrox/rox/pkg/auth/authproviders"
 	"github.com/stackrox/rox/pkg/auth/permissions"
@@ -114,6 +116,8 @@ func New(clusters datastore.DataStore, sensorConnMgr connection.Manager, telemet
 		roleDataStore:        roleDataStore,
 		configDataStore:      configDataStore,
 		notifierDataStore:    notifierDataStore,
+		// TODO: Inject dependency
+		alertDataStore: nil,
 	}
 }
 
@@ -130,6 +134,7 @@ type serviceImpl struct {
 	roleDataStore        roleDS.DataStore
 	configDataStore      configDS.DataStore
 	notifierDataStore    notifierDS.DataStore
+	alertDataStore       alertStore.DataStore
 }
 
 // RegisterServiceServer registers this service with the given gRPC Server.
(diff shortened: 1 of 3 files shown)


roxctl/central/debug/download_diagnostics.go
```diff
@@ -69,6 +73,7 @@ To specify timeout, run  'roxctl' command:
 	c.PersistentFlags().StringVar(&outputDir, "output-dir", "", "output directory in which to store bundle")
 	c.PersistentFlags().StringSliceVar(&clusters, "clusters", nil, "comma separated list of sensor clusters from which logs should be collected")
 	c.PersistentFlags().StringVar(&since, "since", "", "timestamp starting when logs should be collected from sensor clusters")
+	c.PersistentFlags().BoolVar(&withAlerts, "with-alerts", false, "add alert data before/after reassessing policies")
```
0x656b694d, line 76: Not clear to me what does "before/after" mean. Before and/or after?

central/debug/service/service.go
```diff
@@ -598,6 +620,30 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
...
+					ReprocessDeployments: &central.ReprocessDeployments{},
+				},
+			})
+			if err != nil {
+				log.Warnf("Failed to request reprocess deployment for cluster: %s", conn.ClusterID())
+				continue
+			}
+		}
+
+		log.Info("Requested reprocess clusters. Waiting for a minute before finish writing the diagnostic bundle")
```
ivan-degtiarenko, line 639: Why 1 minute? Is there a guarantee it will be enough?

central/debug/service/service.go
```diff
@@ -598,6 +620,30 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
 
 	fetchAndAddJSONToZip(debugDumpCtx, zipWriter, "system-configuration.json", s.getConfig)
 
+	// Fetch alerts and append to the zip
+	if opts.withAlerts {
+		fetchAndAddJSONToZip(debugDumpCtx, zipWriter, "alerts-before.json", s.fetchAlerts)
+		// clusterSet := set.NewFrozenStringSet(opts.clusters...)
+		for _, conn := range s.sensorConnMgr.GetActiveConnections() {
+			err := conn.InjectMessage(debugDumpCtx, &central.MsgToSensor{
+				Msg: &central.MsgToSensor_ReprocessDeployments{
```
ivan-degtiarenko, line 629: I wonder if we are OK with introducing side effects to the diag bundle, and whether we shouldn't rather introduce completely separate flow for that

central/debug/service/service.go
```diff
@@ -598,6 +620,30 @@ func (s *serviceImpl) writeZippedDebugDump(ctx context.Context, w http.ResponseW
...
+				log.Warnf("Failed to request reprocess deployment for cluster: %s", conn.ClusterID())
+				continue
+			}
+		}
+
+		log.Info("Requested reprocess clusters. Waiting for a minute before finish writing the diagnostic bundle")
+
+		// We have to sleep here to make sure that any new alerts that could be generated after requesting all deployments
+		// to be reprocessed will appear in the new JSON file.
+		time.Sleep(time.Minute)
```
ivan-degtiarenko, line 643: During this minute context deadline will be ignored

ivan-degtiarenko (changes requested):
I noticed that the ticket mentions UI changes as well. Is it planned to be done in a separate PR?

roxctl/central/debug/download_diagnostics.go
fredrb: I think "before and after" will be more clear. These are the files that will be output in the diagnostic bundle. Alerts before and after reassessing policies.

central/debug/service/service.go
fredrb: This is pretty arbitrary. The current approach was to set to a time where we're are confident that most alerts would have arrived after reassessing. But we don't have any way to guarantee that all _possible_ alerts were already received. 

We thought also about making this duration configurable. E.g. with a flag `--reprocess-delay 30s`?

central/debug/service/service.go
fredrb: I agree. But adding a new flow might be inconvenient for customers. So, our proposal here is to add a flag that is disabled by default in diag bundles and will only be used for this specific case. 
```
roxctl central debug download-diagnostics --alerts=true --timeout=100
```

Ideally, we should remove this after this release, once re-sync is disabled by default on customer's clusters.  

Would you suggest adding a new `roxctl debug` command altogether for this?

fredrb:
@ivan-degtiarenko @0x656b694d sorry folks, you were automatically added as reviewers without much context on what we were doing. I've updated the description of the PR with the reasoning on why we are adding this. Hope it clarifies. If you have more questions or suggestions let me know. 



> I noticed that the ticket mentions UI changes as well. Is it planned to be done in a separate PR?

We aren't doing any UI changes anymore. That was the original plan, but now we are going `roxctl` only.

fredrb:
Closing in favor of #5559

fredrb:
Closing in favor of #5559

PR #5550, opened 2023-04-04
Revert ProcessIndicator to use generated store

erthalion:
## Description

During PLOP feature development it was needed to make ProcessIndicator to use a custom store. After a round of simplification it's not required any more, to put it back.

Together with that specify DeploymentId as an FK for ProcessListeningOnPort.

## Checklist

## Testing Performed

CI run should be sufficient.


dashrews78 (approved):
LGTM

PR #5554, opened 2023-04-04
ROX-14479 Improve error reporting when handling the defaultTLSCertificate

ludydoo:
## Description

Errors around malformed `defaultTLSCertificate` are not super helpful. This PR aims at providing better error messages for the customer when such errors occur. 

The previous implementation used 2-3 different methods with duplicated logic to load and verify the default TLS certificate. This basically uses a single access point for handling that.

### DIFF
--- a/central/clusters/zip/generate_certs.go
+++ b/central/clusters/zip/generate_certs.go
@@ -59,16 +59,17 @@ func getAdditionalCAs(certs *sensor.Certs) ([]*zip.File, error) {
 }
 
 // maybeCreateZipFileForDefaultTLSCertCA returns a zip file containing the default CA cert if it is not trusted by the system roots.
+// If there is no default CA cert, or if it is already trusted by the system roots, it returns nil.
 func maybeCreateZipFileForDefaultTLSCertCA() (*zip.File, error) {
-	defaultTLSCertificate, err := tlsconfig.MaybeGetDefaultTLSCertificateFromDefaultDirectory()
+	defaultTLSCer, err := tlsconfig.MaybeGetDefaultTLSCertificateFromDefaultDirectory()
 	if err != nil {
 		return nil, errors.Wrap(err, "error getting default TLS certificate from default directory")
 	}
-	if defaultTLSCertificate == nil || len(defaultTLSCertificate.Certificate) == 0 {
+	if defaultTLSCer == nil || len(defaultTLSCer.Certificate) == 0 {
 		return nil, nil
 	}
 
-	lastInChain, err := x509.ParseCertificate(defaultTLSCertificate.Certificate[len(defaultTLSCertificate.Certificate)-1])
+	lastInChain, err := x509.ParseCertificate(defaultTLSCer.Certificate[len(defaultTLSCer.Certificate)-1])
 	if err != nil {
 		return nil, errors.Wrap(err, "error parsing default TLS certificate")
 	}
--- a/central/clusters/zip/generate_certs.go
+++ b/central/clusters/zip/generate_certs.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/stackrox/rox/central/clusters"
 	siDataStore "github.com/stackrox/rox/central/serviceidentities/datastore"
 	"github.com/stackrox/rox/central/tlsconfig"
@@ -48,7 +49,7 @@ func getAdditionalCAs(certs *sensor.Certs) ([]*zip.File, error) {
 	}
 
 	if zipForDefaultTLSCertCA, err := maybeCreateZipFileForDefaultTLSCertCA(); err != nil {
-		log.Errorf("Error obtaining default CA cert: %v", err)
+		log.Errorf("Error obtaining default TLS Certificate: %v", err)
 	} else if zipForDefaultTLSCertCA != nil {
 		files = append(files, zipForDefaultTLSCertCA)
 		certs.Files[fmt.Sprintf("secrets/%s/%s", additionalCAsZipSubdir, centralCA)] = zipForDefaultTLSCertCA.Content
@@ -61,15 +62,15 @@ func getAdditionalCAs(certs *sensor.Certs) ([]*zip.File, error) {
 func maybeCreateZipFileForDefaultTLSCertCA() (*zip.File, error) {
 	defaultTLSCertificate, err := tlsconfig.MaybeGetDefaultTLSCertificateFromDefaultDirectory()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error getting default TLS certificate from default directory")
 	}
-	if defaultTLSCertificate == nil {
+	if defaultTLSCertificate == nil || len(defaultTLSCertificate.Certificate) == 0 {
 		return nil, nil
 	}
 
 	lastInChain, err := x509.ParseCertificate(defaultTLSCertificate.Certificate[len(defaultTLSCertificate.Certificate)-1])
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error parsing default TLS certificate")
 	}
 
 	// Only add cert to bundle if it is not trusted by system roots.
--- a/central/clusters/zip/generate_certs.go
+++ b/central/clusters/zip/generate_certs.go
@@ -1,7 +1,6 @@
 package zip
 
 import (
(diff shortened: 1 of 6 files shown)


central/clusters/zip/generate_certs.go
```diff
@@ -49,32 +48,29 @@ func getAdditionalCAs(certs *sensor.Certs) ([]*zip.File, error) {
...
-	keyFile := filepath.Join(tlsconfig.DefaultCertPath, tlsconfig.TLSKeyFileName)
-
-	if filesExist, err := fileutils.AllExist(certFile, keyFile); err != nil || !filesExist {
-		return nil, err
-	}
-
-	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+// maybeCreateZipFileForDefaultTLSCertCA returns a zip file containing the default CA cert if it is not trusted by the system roots.
+func maybeCreateZipFileForDefaultTLSCertCA() (*zip.File, error) {
+	defaultTLSCertificate, err := tlsconfig.MaybeGetDefaultTLSCertificateFromDefaultDirectory()
```
porridge, line 63: Perhaps this would still be just as readable but make lines a bit shorter?
```suggestion
	defaultTLSCert, err := tlsconfig.MaybeGetDefaultTLSCertificateFromDefaultDirectory()
```

central/clusters/zip/generate_certs.go
```diff
@@ -49,32 +48,29 @@ func getAdditionalCAs(certs *sensor.Certs) ([]*zip.File, error) {
...
-func getDefaultCertCA() (*zip.File, error) {
-	certFile := filepath.Join(tlsconfig.DefaultCertPath, tlsconfig.TLSCertFileName)
-	keyFile := filepath.Join(tlsconfig.DefaultCertPath, tlsconfig.TLSKeyFileName)
-
-	if filesExist, err := fileutils.AllExist(certFile, keyFile); err != nil || !filesExist {
-		return nil, err
-	}
-
-	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+// maybeCreateZipFileForDefaultTLSCertCA returns a zip file containing the default CA cert if it is not trusted by the system roots.
```
porridge, line 61:
```suggestion
// maybeCreateZipFileForDefaultTLSCertCA returns a zip file containing the default CA cert if it is not trusted by the system roots, or nil otherwise.
```

central/tlsconfig/tlsconfig.go
```diff
@@ -56,39 +57,54 @@ func GetAdditionalCAs() ([][]byte, error) {
 	return certDERs, nil
 }
 
-// GetDefaultCertChain reads and parses default cert chain and returns it in DER encoded format
-func GetDefaultCertChain() ([][]byte, error) {
-	certFile := filepath.Join(DefaultCertPath, TLSCertFileName)
-	content, err := os.ReadFile(certFile)
+// MaybeGetDefaultCertChain reads and parses default cert chain and returns it in DER encoded format
```
porridge, line 60:
```suggestion
// MaybeGetDefaultCertChain reads and parses default cert chain and returns it in DER encoded format.
```

central/tlsconfig/tlsconfig.go
```diff
@@ -56,39 +57,54 @@ func GetAdditionalCAs() ([][]byte, error) {
...
 
-	return certDERsFromFile, nil
+// MaybeGetDefaultTLSCertificateFromDefaultDirectory load the default tls certificate from the default directory.
+func MaybeGetDefaultTLSCertificateFromDefaultDirectory() (*tls.Certificate, error) {
+	return MaybeGetDefaultTLSCertificateFromDirectory(DefaultCertPath)
 }
 
-// loadDefaultCertificate load the default tls certificate
-func loadDefaultCertificate(dir string) (*tls.Certificate, error) {
+// MaybeGetDefaultTLSCertificateFromDirectory load the default tls certificate from the given directory.
```
porridge, line 77:
```suggestion
// MaybeGetDefaultTLSCertificateFromDirectory loads the default TLS certificate from the given directory.
```

central/tlsconfig/tlsconfig.go
```diff
@@ -56,39 +57,54 @@ func GetAdditionalCAs() ([][]byte, error) {
...
-	if err != nil {
-		return nil, errors.Wrap(err, "converting additional CA cert to DER")
+	if cert == nil {
+		return nil, nil
 	}
+	return cert.Certificate, nil
+}
 
-	return certDERsFromFile, nil
+// MaybeGetDefaultTLSCertificateFromDefaultDirectory load the default tls certificate from the default directory.
```
porridge, line 72:
```suggestion
// MaybeGetDefaultTLSCertificateFromDefaultDirectory loads the default TLS certificate from the default directory.
```

central/tlsconfig/tlsconfig.go
```diff
@@ -56,39 +57,54 @@ func GetAdditionalCAs() ([][]byte, error) {
...
+		}
+		log.Infof("Default TLS key file %q does not exist. Skipping", keyFile)
+		return nil, nil
 	}
 
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		return nil, err
+		if strings.Contains(err.Error(), "private key does not match public key") {
+			return nil, errors.Wrap(err, "loading default certificate. If the certificate file contains a certificate chain, ensure that the certificate chain is in the correct order (the first certificate should be the server certificate, any following certificates should form the certificate chain).")
```
porridge, line 103: IIRC error messages should not end with a full stop...
```suggestion
			return nil, errors.Wrap(err, "loading default certificate failed; if the certificate file contains a certificate chain, ensure that the certificate chain is in the correct order (the first certificate should be the server certificate, any following certificates should form the certificate chain)")
```

central/tlsconfig/tlsconfig.go
```diff
@@ -56,39 +57,54 @@ func GetAdditionalCAs() ([][]byte, error) {
...
+		}
+		return nil, errors.Wrap(err, "loading default certificate")
```
porridge, line 105:
```suggestion
		return nil, errors.Wrap(err, "loading default certificate failed")
```

central/tlsconfig/tlsconfig.go
porridge: Also, would it be better to say `leaf certificate` instead of `server certificate`? I think there is nothing preventing use of this function to load client certificates, where mentioning "server" would be rather confusing...

central/tlsconfig/tlsconfig_test.go
```diff
@@ -40,3 +48,225 @@ func (s *tlsConfigTestSuite) isCommonNameInCerts(DERs [][]byte, commonName strin
...
+		}
+		cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, nil, err
+		}
+		certs = append(certs, cert)
+		privateKeys = append(privateKeys, privateKey)
+	}
+	// inverse the order of the certs so that the leaf is first
+	for i, j := 0, len(certs)-1; i < j; i, j = i+1, j-1 {
```
porridge, line 224: `<rant>`

:woozy_face: this makes my head ache.
Yeah, after 5 minutes I do understand what this does and that it's probably the most efficient way, and that it DTRT in all corner cases, but I really hope the standard library will grow a generic reverse function at some point...
`</rant>`

central/tlsconfig/tlsconfig_test.go
```diff
@@ -40,3 +48,225 @@ func (s *tlsConfigTestSuite) isCommonNameInCerts(DERs [][]byte, commonName strin
...
+	var certs []*x509.Certificate
+	var privateKeys []*ecdsa.PrivateKey
+	for i := 0; i < length; i++ {
+		var parentCert = &x509.Certificate{}
+		var parentKey = &ecdsa.PrivateKey{}
+		if i > 0 {
+			parentCert = certs[i-1]
+			parentKey = privateKeys[i-1]
+		}
+		serialNumber, err := generateSerialNumber()
```
porridge, line 194: Perhaps just use `i+1` or something?

porridge (approved):
Some nitpicks inline.

central/tlsconfig/tlsconfig_test.go
ludydoo: I agree! Just found out that we already have a method (`sliceutils.ReverseInPlace`). Substituted for that one instead of the ad-hoc reverse.

PR #5559, opened 2023-04-04
ROX-16279: roxctl central debug resync check

fredrb:
## Description

Introduces a temporary debug command: `roxctl central debug resync-check` under the `ROX_RESYNC_DISABLED` tech preview flag.

This command should fetch alerts, reassess policies, and fetch alerts again. It should be used exclusively by users testing this feature as a sanity check to make sure that no alerts were missed.

For more context, also check the discussion in #5544 

## Checklist

This is not a GA feature, and therefore is not documented. 

## Testing Performed

- [x] Unit tests for warning beahvior 
- [x] Manual tests against cluster

Here's an example of running with reduced wait time:
```
$ ROX_RESYNC_DISABLED="true" ./bin/darwin_amd64/roxctl --insecure-skip-tls-verify -e https://localhost:8000 central debug resync-check --wait-for=5s
INFO:	Running re-sync check
INFO:	Found 15 alerts before reassessing
INFO:	Waiting for 5s while reassess policies produces alerts
INFO:	Found 15 alerts after reassessing
INFO:	Storing alerts resync-check-output/alerts-before.json
INFO:	Storing alerts resync-check-output/alerts-after.json
```

And an example with fabricated errors:
```
$ ROX_RESYNC_DISABLED="true" ./bin/darwin_amd64/roxctl --insecure-skip-tls-verify -e https://localhost:8000 central debug resync-check --wait-for=1s
INFO:	Running re-sync check
INFO:	Found 15 alerts before reassessing
INFO:	Waiting for 1s while reassess policies produces alerts
INFO:	Found 15 alerts after reassessing
WARN:	Number of alerts differ in before and after! Reach out to Red Hat ACS support and provide output files generated by this command.
INFO:	Storing alerts resync-check-output/alerts-before.json
INFO:	Storing alerts resync-check-output/alerts-after.json
```


roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Roxctl needs to wait here to realistically get the delta between the violation state before and after reassessing
+	// policies. If the timer is too short, there's a higher change that there will be no delta, because the Sensors
+	// still haven't fully reprocessed the deployments.
+	waitForC := time.After(c.waitFor)
+	select {
+	case <-waitForC:
+		break
+	case <-ctx.Done():
+		return nil, nil, errors.Errorf("context finished before second request could be made: `wait-for` value might be too high: %s", c.waitFor.String())
```
dhaus67, line 104:
```suggestion
		return nil, nil, errox.InvalidArgs.Newf("context finished before second request could be made: `wait-for` value might be too high: %s", c.waitFor.String())
```

roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+func (c *resyncCheckCmd) fetchAlerts(ctx context.Context, svc v1.AlertServiceClient) ([]*storage.ListAlert, error) {
+	response, err := svc.ListAlerts(ctx, &v1.ListAlertsRequest{})
+	if err != nil {
+		return nil, errors.Wrap(err, "could not make ListAlerts request to gRPC server")
+	}
+	return response.GetAlerts(), nil
+}
+
+func (c *resyncCheckCmd) storeFiles(alertsBefore, alertsAfter []*storage.ListAlert) error {
+	if err := os.Mkdir(c.outputDir, 0755); err != nil {
```
dhaus67, line 127: Let's short-circuit here and return an error if the directory couldn't be created and `os.IsExist(err) == false`
```suggestion
	if err := os.Mkdir(c.outputDir, 0755); err != nil && !os.IsExist(err){
```

dhaus67, line 127: super nit (readability, nothing more):
```suggestion
	if err := os.Mkdir(c.outputDir, os.ModePerm); err != nil {
```

roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+
+	sort.SliceStable(before, func(i, j int) bool {
+		return before[i].Id > before[j].Id
+	})
+
+	sort.SliceStable(after, func(i, j int) bool {
+		return after[i].Id > after[j].Id
+	})
+
+	if !cmp.Equal(before, after) {
```
dhaus67, line 168: nit:
```suggestion
	if !protoutils.SlicesEqual(before, after) {
```

dhaus67 (approved):
Minor nits and one suggestion to short-circuit when directory creation fails, otherwise LGTM 🚢

roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+	v1 "github.com/stackrox/rox/generated/api/v1"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/roxctl/common/environment"
+	"github.com/stackrox/rox/roxctl/common/flags"
+	"github.com/stackrox/rox/roxctl/common/util"
+	"google.golang.org/grpc"
+)
+
+const (
+	resyncCheckTimeout = 300 * time.Second
```
lvalerom, line 23: [nit] I would rename this to `defaultResyncCheckTimeout`

roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+	c.PersistentFlags().DurationVar(&waitFor, "wait-for", time.Minute, "how long to wait between before and after alert check")
+
+	return c
+}
+
+type resyncCheckCmd struct {
+	env       environment.Environment
+	conn      *grpc.ClientConn
+	waitFor   time.Duration
+	outputDir string
```
lvalerom, line 59: We should add here a timeout field and set it to `flags.Timeout(c)` in `commandWithConnection`. Then in `run` function the context should be created with this field instead of using `resyncCheckTimeout`. Otherwise I think we cannot configure different timeouts for the command.

```golang
ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
```

roxctl/central/debug/resync_check.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+			before, after, err := cmd.run()
+			if err != nil {
+				return err
+			}
+
+			return cmd.storeFiles(before, after)
+		}),
+	}
+	flags.AddTimeoutWithDefault(c, diagnosticBundleDownloadTimeout)
```
lvalerom, line 48: I think this should be `resyncCheckTimeout`

```suggestion
	flags.AddTimeoutWithDefault(c, resyncCheckTimeout)
```

lvalerom (changes requested):
Looks good. I have a couple of concerns:
* I think the `timeout` flag is not being used.
* If I'm not mistaken `ListAlerts` has a limit of 1000 alerts per query. I'm not sure if we should handle the situation where we might reach this limit and therefore not retrieve all the alerts.

roxctl/central/debug/resync_check.go
fredrb: Done

roxctl/central/debug/resync_check.go
fredrb: Done

roxctl/central/debug/resync_check.go
fredrb: Done

lvalerom (approved):
Looks good. Thanks for making the changes 🙂

PR #5573, opened 2023-04-05
ROX-14960: Fix detection of previous Y-Stream release

msugakov:
## Description

Please see the mentioned ticket and commit messages for the context.

One thing I'm not 100% sure about is whether to keep the script as external shell one or move the logic inside of `patch-csv.py` (or next to it).

Advantages of keeping a separate script:
- It is unit-tested and tests are ran in CI, we don't have unit tests in CI for Python.
- It can be used in other places (e.g. version-compatibility tests, but I'm yet to run them).

Advantages of integrating in `patch-csv.py`:
- Less moving parts like spawning external process from Python.

## Checklist

These aren't needed:

## Testing Performed

* CI here should be green (enough green).
* Testing with `4.0.0-rc.1` tag under https://github.com/stackrox/stackrox/pull/5530

### DIFF
--- a/scripts/get-previous-y-stream.bats
+++ b/scripts/get-previous-y-stream.bats
@@ -23,6 +23,9 @@ setup() {
     test_invalid "3.0.62.x"
     test_invalid "3.0.62.1"
     test_invalid " v4.0.0"
+    test_invalid ".."
+    test_invalid "..-1.2.3"
+    test_invalid "1.2.3-"
 }
 
 test_invalid() {
@@ -51,6 +54,7 @@ test_excessive_args() {
     test_major_unknown "5.0.0" "5.0"
     test_major_unknown "5.0.1" "5.0"
     test_major_unknown "v5.0.x-nightly-12345" "5.0"
+    test_major_unknown "199.0.88" "199.0"
 }
 
 test_major_unknown() {
@@ -75,6 +79,7 @@ test_major_unknown() {
     test_happy "4.1.0" "4.0.0"
     test_happy "5.10.2" "5.9.0"
     test_happy "v3.62.7" "3.61.0"
+    test_happy "45.67.89" "45.66.0"
 }
 
 test_happy() {
--- a/scripts/get-previous-y-stream.bats
+++ b/scripts/get-previous-y-stream.bats
@@ -1,0 +1,84 @@
+#!/usr/bin/env bats
+
+load "./test_helpers.bats"
+
+setup() {
+    program="$BATS_TEST_DIRNAME"/get-previous-y-stream.sh
+}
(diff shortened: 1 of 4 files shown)


msugakov:
As of now the testing is in progress but the change is ready for early review.

scripts/get-previous-y-stream.sh
```diff
@@ -0,0 +1,64 @@
...
+        exit 1
+    fi
+
+    local version="${1-}"
+
+    if [[ -z "$version" || "$version" == "--help" ]]; then
+        usage
+    fi
+
+    if [[ ! "$version" =~ ^v?([0-9]*)\.([0-9]*)\.(x|[0-9]*)(-.*)?$ ]]; then
```
jschnath, line 34: should be
`    if [[ ! "$version" =~ ^v?([0-9]+)\.([0-9]+)\.(x|[0-9]+)(-.+)?$ ]]; then`
`*` also matches 0 repetitions, whereas `+` requires at least 1. A test to differentiate between the two would be to check whether `..-` is a valid version number

scripts/get-previous-y-stream.bats
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bats
+
+load "./test_helpers.bats"
+
+setup() {
```
vikin91, line 5: Nit: This can be `setup_file` as the value of `program` is the same for each test case in this file, and we do not change it during the tests.

scripts/get-previous-y-stream.sh
msugakov: Good catch. This was a result of me over-editing the regex and not actually reviewing it myself.

scripts/get-previous-y-stream.bats
msugakov: I did not think of this before your comment. Tried it and prefer to stay with `setup()`.

The problem with `setup_file` is that it requires `export`-ing variables. While it makes no harm so far, exported variable is seen by subprocesses such as `get-previous-y-stream.sh` itself. This makes the global state seen by that script in test context different than in execution outside of test context.

Since the overhead of current `setup()` is negligible, I'd prefer to stay with it unless you have strong opinion.

vikin91: No, no strong opinion here, it can stay with `setup` as the execution costs nothing.

jschnath (approved):
looks good to me

vikin91 (approved):
Looks good!

I did a quick manual test and found nothing that would worry me

msugakov:
OCP 12 QA E2E completed. I won't wait for OCP 10 ones.

PR #5574, opened 2023-04-05
ROX-14480 Allow PEM files for additional CAs.

ludydoo:
## Description

This PR allows customers to specify additional CAs using the `.pem` file extension, in addition to the currently allowed `.crt` file extension.


central/tlsconfig/tlsconfig.go
```diff
@@ -183,3 +215,32 @@ func validForAllDNSNames(cert *x509.Certificate, dnsNames ...string) bool {
 	}
 	return true
 }
+
+var allowedAdditionalCAExtensions = map[string]bool{
+	".crt": true,
+	".pem": true,
```
SimonBaeumer, line 221: Could the file extension be ignored altogether? 
Just scanning all files, and if a file can't be decoded, log a warning message instead?

central/tlsconfig/tlsconfig.go
```diff
@@ -183,3 +215,32 @@ func validForAllDNSNames(cert *x509.Certificate, dnsNames ...string) bool {
...
+	for ext := range allowedAdditionalCAExtensions {
+		allowedAdditionalCAExtensionsList = append(allowedAdditionalCAExtensionsList, ext)
+	}
+	sort.Strings(allowedAdditionalCAExtensionsList)
+	skipAdditionalCAFileMsg = "skipping additional-ca file %q because it has an invalid extension; allowed file extensions for additional ca certificates are "
+	for i, ext := range allowedAdditionalCAExtensionsList {
+		if i > 0 && i < len(allowedAdditionalCAExtensionsList)-1 {
+			skipAdditionalCAFileMsg += ", "
+		} else if i == len(allowedAdditionalCAExtensionsList)-1 {
+			skipAdditionalCAFileMsg += " and "
```
SimonBaeumer, line 237: You could use a string slice, append the skipped files and later explode the string, writing it to a log message.
No one likes to debug string concatenations (especially not if it results in grammar errors). :(

central/tlsconfig/tlsconfig_test.go
```diff
@@ -257,3 +258,30 @@ func writeKey(dir string, privateKey *ecdsa.PrivateKey) error {
...
+		{"bla.crt.", false},
+		{"bla.pem.", false},
+		{"bla.crt.foo", false},
+		{"", false},
+		{"foo.bar", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := isValidAdditionalCAFileName(tt.name); got != tt.want {
+				t.Errorf("isValidAdditionalCAFileName() = %v, want %v", got, tt.want)
```
SimonBaeumer, line 279: Would this work? It safes an if statement by using the assertion instead.

```suggestion
			assert.Equal(t, tt.want, isValidAdditionalCAFileName(tt.name), "isValidAdditionalCAFileName() = %v, want %v", got, tt.want)

```

central/tlsconfig/tlsconfig.go
```diff
@@ -36,17 +38,47 @@ func GetAdditionalCAs() ([][]byte, error) {
 		return nil, errors.Wrap(err, "reading additional CAs directory")
 	}
 
-	var certDERs [][]byte
+	var files []string
 	for _, certFile := range certFileInfos {
-		if filepath.Ext(certFile.Name()) != ".crt" {
-			log.Infof("Skipping additional-ca file %q, must end with '*.crt'.", certFile.Name())
+		if certFile.IsDir() {
```
SimonBaeumer, line 43: It may not be possible to loop into a directory in a secret mount, but to be safe adding a log statement a directory was skipped is useful.

central/tlsconfig/tlsconfig.go
ludydoo: That would be ideal yes. I would like to do this in another PR, because I would need the symlink logic in #5575 . Perhaps I can add a ticket to the epic, the basically remove any sort of extension restriction for the additional-cas, and address it there. What do you think?

central/tlsconfig/tlsconfig.go
ludydoo: I'm not sure I understand your recommendation fully. Do you mind giving an example ?

central/tlsconfig/tlsconfig.go
SimonBaeumer: I think it is okay. 
To me it sounded like ignoring the file extension reduces the needed logic for error reporting and logging, instead it is reduced to writing encoding errors to the logger.

SimonBaeumer (changes requested):
See comment for `strings.Join` above.

central/tlsconfig/tlsconfig.go
ludydoo: > See comment for `strings.Join` above.

@SimonBaeumer  Yeah, I simplified the building of the error message a lot, basically defining the alllowed extensions in a slice and using this directly for the error message.

porridge (changes requested):
Code looks good, but what about documentation? Hopefully we mention somewhere that only `.pem` are read, we should change this now. If not, we should document this.

ludydoo:
> Code looks good, but what about documentation? Hopefully we mention somewhere that only `.pem` are read, we should change this now. If not, we should document this.

Yes, I'll be in touch with Kerry Carmichael as soon as this is merged to update the documentation. I'll make a PR there https://github.com/openshift/openshift-docs/tree/rhacs-docs

SimonBaeumer:
> > Code looks good, but what about documentation? Hopefully we mention somewhere that only `.pem` are read, we should change this now. If not, we should document this.
> 
> Yes, I'll be in touch with Kerry Carmichael as soon as this is merged to update the documentation. I'll make a PR there https://github.com/openshift/openshift-docs/tree/rhacs-docs

Are you sure you want to add it to the documentation and going that path, considering that we discussed trying to parse all files, regardless of the file extension?

central/tlsconfig/tlsconfig.go
```diff
@@ -36,17 +37,48 @@ func GetAdditionalCAs() ([][]byte, error) {
...
+		files = append(files, path.Join(additionalCADir, certFile.Name()))
+	}
+
+	return files, nil
+
+}
+
+// GetAdditionalCAs reads all additional CAs in DER format.
+func GetAdditionalCAs() ([][]byte, error) {
+
```
SimonBaeumer, line 69: nit empty line

central/tlsconfig/tlsconfig.go
```diff
@@ -183,3 +215,19 @@ func validForAllDNSNames(cert *x509.Certificate, dnsNames ...string) bool {
 	}
 	return true
 }
+
+var allowedAdditionalCAExtensionList = []string{".crt", ".pem"}
+var allowedAdditionalCAExtensionMap = map[string]struct{}{}
```
SimonBaeumer, line 220: nit:
```
var (
  allowedAdditionalCAExtensionList = []string{".crt", ".pem"}
  allowedAdditionalCAExtensionMap = map[string]struct{}{}
)
```

ludydoo:
> Are you sure you want to add it to the documentation and continue on this, considering that we discussed trying to parse all files, regardless of the file extension?

I would update the documentation to reflect those changes when this is merged, and create a ticket to remove the restriction on the extension, with acceptance criteria specifying that the documentation should also be updated

central/tlsconfig/tlsconfig_test.go
```diff
@@ -30,11 +31,11 @@ func (s *tlsConfigTestSuite) TestGetAdditionalCAs() {
 
 	additionalCAs, err := GetAdditionalCAs()
 	s.Require().NoError(err)
-	s.Require().Len(additionalCAs, 3, "Could not decode all certs")
+	s.Require().Len(additionalCAs, 4, "Could not decode all certs")
 	s.True(s.isCommonNameInCerts(additionalCAs, "CENTRAL_SERVICE: Central"), "Could not find cert from multiple crt file")
 
-	// non .crt files should be ignored
-	s.FileExists("testdata/no_ca_cert.pem")
+	// non .crt or .pem files should be ignored
```
SimonBaeumer, line 37: You've invested so much into making the extensions dynamic, but the comment is not dynamically generated from the extensions list :p

central/tlsconfig/tlsconfig_test.go
```diff
@@ -257,3 +258,28 @@ func writeKey(dir string, privateKey *ecdsa.PrivateKey) error {
 	}
 	return pem.Encode(keyFile, &pem.Block{Type: "EC PRIVATE KEY", Bytes: privateKeyBytes})
 }
+
+func Test_isValidAdditionalCAFileName(t *testing.T) {
```
SimonBaeumer, line 262:
```suggestion
func TestIsValidAdditionalCAFileName(t *testing.T) {
```

central/tlsconfig/tlsconfig_test.go
```diff
@@ -257,3 +258,28 @@ func writeKey(dir string, privateKey *ecdsa.PrivateKey) error {
...
+		{"foo.bar", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, isValidAdditionalCAFileName(tt.name))
+		})
+	}
+}
+
+func Test_skipAdditionalCAFileMsg(t *testing.T) {
```
SimonBaeumer, line 283:
```suggestion
func TestSkipAdditionalCAFileMsg(t *testing.T) {
```

image/templates/common/ca-setup.sh
```diff
@@ -35,7 +35,7 @@ function usage {
 	echo
 	echo "The argument may be:"
 	echo "  - a single file"
-	echo "  - a directory (all files ending in .crt will be added)"
+	echo "  - a directory (all files ending in .crt and .pem will be added)"
```
SimonBaeumer, line 38: Good catch adding the extensions to the `ca-setup.sh`!

SimonBaeumer:
> > Are you sure you want to add it to the documentation and continue on this, considering that we discussed trying to parse all files, regardless of the file extension?
> 
> I would update the documentation to reflect those changes when this is merged, and create a ticket to remove the restriction on the extension, with acceptance criteria specifying that the documentation should also be updated

You can do it like this, to me just sounds like doubling the effort. Why do you not want to remove the extensions logic and parse all files directly?

central/tlsconfig/tlsconfig_test.go
ludydoo: lol, I don't think I'll add a `go generate` to update this comment ;)

central/tlsconfig/tlsconfig_test.go
ludydoo: Could you clarify your stance on https://github.com/stackrox/stackrox/pull/5591#discussion_r1165197016

SimonBaeumer (approved):

image/templates/common/ca-setup.sh
```diff
@@ -57,14 +57,14 @@ function create_directory {
 	local dir="$1"
 	echo "The following certificates will be used as additional CAs:"
 	from_file_args=()
-	for f in $dir/*.crt; do
+	for f in $dir/*.crt $dir/*.pem; do
```
porridge, line 60: Since we're touching this line...
```suggestion
	for f in "$dir"/*.crt "$dir"/*.pem; do
```

porridge (approved):

PR #5577, opened 2023-04-05
Disable PLOP E2E tests if the feature is off

erthalion:
## Description

PLOP feature seems to be disabled by default in nightlies. Ignore e2e tests if that's the case.

## Checklist

## Testing Performed

CI run should be sufficient.

### DIFF
--- a/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
+++ b/qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
@@ -13,6 +13,7 @@ import spock.lang.Stepwise
 import services.ProcessesListeningOnPortsService
 
 @Stepwise
+@IgnoreIf({ Env.get("ROX_PROCESSES_LISTENING_ON_PORT", "false") == "false" })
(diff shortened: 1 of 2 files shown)


qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy
```diff
@@ -13,6 +13,7 @@ import spock.lang.Stepwise
 import services.ProcessesListeningOnPortsService
 
 @Stepwise
+@IgnoreIf({ Env.get("ROX_PROCESSES_LISTENING_ON_PORT", "false") == "false" })
 @IgnoreIf({ !Env.get("ROX_POSTGRES_DATASTORE", null) })
```
gavin-stackrox, line 17: It is postgres everywhere now.
```suggestion
```

gavin-stackrox (changes requested):
Also move https://github.com/stackrox/stackrox/blob/b527b6dacbe425c3b0b57b2fc408d8ccc5457277/tests/e2e/lib.sh#L99 inside the `[[ -z "${BUILD_TAG:-" ]]` check and it should be good to go.

